PR #35, opened 2015-01-21
PHOENIX-1601 Fix performance script for binary distribution of phoenix

osakatand:
Change the PHOENIX_TEST_JAR_PATTERN since it was also matching flume tests jars.

The performance.py script wasn't working anymore in the bin distribution of phoenix
because the flume test jar was matching the test jar pattern for phoenix core test jar.
This new pattern fixes this issue.


stoty:
PHOENIX-1601 has already been committed.

PR #37, opened 2015-02-16
Surface partial saves in CommitExcepiton (PHOENIX-900)

elilevine:
https://issues.apache.org/jira/browse/PHOENIX-900


phoenix-core/src/main/java/org/apache/phoenix/execute/MutationState.java
```diff
@@ -488,7 +490,55 @@ public void rollback(PhoenixConnection connection) throws SQLException {
...
+    	public RowMutationState(@NotNull Map<PColumn,byte[]> columnValues, @NotNull Set<Integer> orderOfStatementsInConnection) {
+    		Preconditions.checkNotNull(columnValues);
+    		Preconditions.checkNotNull(orderOfStatementsInConnection);
+    		
+    		this.columnValues = columnValues;
+    		this.orderOfStatementsInConnection = orderOfStatementsInConnection;
+    	}
+    	
+    	public Map<PColumn, byte[]> getColumnValues() {
+			return columnValues;
```
elilevine: Will fix formatting

phoenix-core/src/main/java/org/apache/phoenix/jdbc/PhoenixStatement.java
```diff
@@ -151,6 +151,9 @@
  * @since 0.1
  */
 public class PhoenixStatement implements Statement, SQLCloseable, org.apache.phoenix.jdbc.Jdbc7Shim.Statement {
+	
+	public static final int ORDER_IN_CONNECTION_UNDEFINED = -1;
```
elilevine: Will remove. No longer needed.

phoenix-core/src/main/java/org/apache/phoenix/execute/MutationState.java
```diff
@@ -488,7 +490,55 @@ public void rollback(PhoenixConnection connection) throws SQLException {
         numRows = 0;
     }
     
+    private Set<Integer> getOrderOfUncommittedStatements() {
```
JamesRTaylor: getUncommittedStatementIndexes() instead?

phoenix-core/src/main/java/org/apache/phoenix/execute/MutationState.java
```diff
@@ -134,29 +135,31 @@ public void join(MutationState newMutation) {
...
-                            Map<PColumn,byte[]> newRow = rowEntry.getValue();
+                            Map<PColumn,byte[]> newRow = rowEntry.getValue().getColumnValues();
                             // if new row is PRow.DELETE_MARKER, it means delete, and we don't need to merge it with existing row. 
                             if (newRow != PRow.DELETE_MARKER) {
                                 // Replace existing column values with new column values
                                 for (Map.Entry<PColumn,byte[]> valueEntry : newRow.entrySet()) {
-                                    existingValues.put(valueEntry.getKey(), valueEntry.getValue());
+                                	existingRowMutationState.getColumnValues().put(valueEntry.getKey(), valueEntry.getValue());
                                 }
+                                existingRowMutationState.getOrderOfStatementsInConnection().addAll(rowEntry.getValue().getOrderOfStatementsInConnection());
```
JamesRTaylor: Would be nice if the logic for combining column values set and statement indexes were moved to a method on RowMutationState, like a join method. Also, would it work to replace the for loop with the following:

    existingRowMutationState.getColumnValues().putAll(newRow);

phoenix-core/src/main/java/org/apache/phoenix/execute/MutationState.java
```diff
@@ -488,7 +490,55 @@ public void rollback(PhoenixConnection connection) throws SQLException {
...
+    	return result;
+    }
+    
     @Override
     public void close() throws SQLException {
     }
+    
+    public static class RowMutationState {
+    	private Map<PColumn,byte[]> columnValues;
+    	private Set<Integer> orderOfStatementsInConnection;
```
JamesRTaylor: I'm somewhat worried about the memory bloat of adding another Set object here. An alternative would be a long[] that acts as a bit set where any set bit represents the statement index. It's unlikely that there'd be more than 64 statements in a commit, but if there were, you could always re-allocate the long[]. You could then materialize the set of statement indexes from it. 

Another more flexible alternative would be to make RowMutationState an interface and have a SingleStatementRowMutationState that has a single int index member variable (likely the common case). Then, add a join method that returns a new MultiRowMutationState with a long as a bitset and maybe one more with a long[] as a bitset.

phoenix-core/src/main/java/org/apache/phoenix/jdbc/PhoenixConnection.java
```diff
@@ -128,12 +129,12 @@
     private PMetaData metaData;
     private final PName tenantId;
     private final String datePattern;
-    
+    private int statementExecutionsCount;
```
JamesRTaylor: Should this be set in the PhoenixConnection copy constructor?

phoenix-core/src/main/java/org/apache/phoenix/compile/UpsertCompiler.java
```diff
@@ -156,7 +157,7 @@ private static MutationState upsertSelect(PhoenixStatement statement,
                             column.getMaxLength(), column.getScale(), column.getSortOrder());
                     values[i] = ByteUtil.copyKeyBytesIfNecessary(ptr);
                 }
-                setValues(values, pkSlotIndexes, columnIndexes, table, mutation);
+                setValues(values, pkSlotIndexes, columnIndexes, table, mutation, statement);
```
JamesRTaylor: Note that this statement may have been cloned from the original statement with a clone of the original connection since this upsertSelect is called in parallel by multiple threads in a upsert select statement. Is that going to be ok for linking back to the statement index in your case?

phoenix-core/src/it/java/org/apache/phoenix/execute/PartialCommitIT.java
```diff
@@ -0,0 +1,262 @@
...
+import com.google.common.base.Preconditions;
+import com.google.common.collect.Maps;
+
+@Category(NeedsOwnMiniClusterTest.class)
+public class PartialCommitIT {
+    
+    private static final String TABLE_NAME_TO_FAIL = "b_failure_table".toUpperCase();
+    private static final byte[] ROW_TO_FAIL = Bytes.toBytes("fail me");
+    private static final String UPSERT_TO_FAIL = "upsert into " + TABLE_NAME_TO_FAIL + " values ('" + Bytes.toString(ROW_TO_FAIL) + "', 'boom!')";
+    private static final String DELETE_TO_FAIL = "delete from " + TABLE_NAME_TO_FAIL + "  where k='z'";
```
JamesRTaylor: Can you add some tests around an UPSERT SELECT failing, as this is a somewhat tricky case. There are a few ways this can execute: source and target table are the same with auto commit on and no aggregation (executed server side) and source and target table are different (executed client side). Also, cases in which a given row is updated in both an UPSERT VALUES and an UPSERT SELECT would be good to test.

phoenix-core/src/it/java/org/apache/phoenix/execute/PartialCommitIT.java
```diff
@@ -0,0 +1,262 @@
...
+
+        props.put(PhoenixRuntime.CURRENT_SCN_ATTRIB, 100);
+
+        try (Connection con = driver.connect(url, new Properties())) {
+            con.setAutoCommit(false);
+            Statement sta = con.createStatement();
+            for (String table : newHashSet("a_success_table", TABLE_NAME_TO_FAIL, "c_success_table")) {
+                sta.execute("upsert into " + table + " values ('z', 'z')");
+                sta.execute("upsert into " + table + " values ('zz', 'zz')");
+                sta.execute("upsert into " + table + " values ('zzz', 'zzz')");
```
JamesRTaylor: What happens if the upsert statement is parameterized: UPSERT INTO T VALUES(?, ?) ? And what if it's executed multiple times, as the bind parameters on PhoenixPreparedStatement mutate. How would the client re-construct the statement?

phoenix-core/src/main/java/org/apache/phoenix/execute/MutationState.java
elilevine: Good suggestion. I was struggling with how to name the beast obviously. :o)

phoenix-core/src/main/java/org/apache/phoenix/execute/MutationState.java
elilevine: Good call.

phoenix-core/src/main/java/org/apache/phoenix/execute/MutationState.java
elilevine: Hmm. I guess I'm not too concerned about memory here, since the size of this set is going to be proportional to the number of executed statements on a connection, which is going to be thousands at the high end for realistic use cases IMHO.

phoenix-core/src/main/java/org/apache/phoenix/jdbc/PhoenixConnection.java
elilevine: Yeah, possibly. I'll add more tests around UPSERT SELECT to make sure cloned connections carry state correctly.

phoenix-core/src/main/java/org/apache/phoenix/compile/UpsertCompiler.java
elilevine: It should be OK because I have separated incrementing and getting the statement index. Incrementing should only be done by the original statement/connection. Cloned connections should only be getting the statement index. I think I need to do more testing here, though. Thanks.

phoenix-core/src/it/java/org/apache/phoenix/execute/PartialCommitIT.java
elilevine: The thinking is that the client controls passing in parametrized values, as well as calling execute*() on statements. If the client cares about reconstructing statements, s/he has all the necessary data to do so.

phoenix-core/src/it/java/org/apache/phoenix/execute/PartialCommitIT.java
```diff
@@ -0,0 +1,280 @@
...
+                    fail("Expected a single row from count query");
+                }
+                assertEquals(expectedCountsForVerification.get(i).intValue(), rs.getInt(1));
+            }
+        } catch (SQLException e) {
+            fail(e.toString());
+        }
+    }
+    
+    public static class FailingRegionObserver extends SimpleRegionObserver {
```
elilevine: @jyates, @apurtell FYI, this is how I hooked up a coprocessor to simulate failure of specific rows.

jyates: Yup, that looks like what I was getting at on the mailing list. Good on ya'

phoenix-core/src/it/java/org/apache/phoenix/execute/PartialCommitIT.java
JamesRTaylor: How? All they have is a statement index. They don't know what the bind variables were at that time.

elilevine: Since callers supply bind variables, they can keep track of them if they are interested in handling CommitException.getUncommittedStatementIndexes() correctly. e.g.:
```
try {
    PreparedStatement ps = con.prepareStatement("upsert into t values (?,?)");
    
    ps.setString(1, "a");
    ps.setString(2, "a");
    ps.executeUpdate(); // statement index 0
    
    ps.setString(1, "b");
    ps.setString(2, "b");
    ps.executeUpdate(); // statement index 1

    con.commit();
} catch (CommitException ce) {
    Set<Integer> uncommitedStatementIndexes = ce.getUncommittedStatementIndexes();
    // do something with uncommitedStatementIndexes 
}
```

JamesRTaylor: That's good that the statement index is updated, but seems like we'd need to have a getBindParametersAtIndex(int index) method to track this for them.

elilevine: I like the getBindParametersAtIndex() idea but feel it's a nice-to-have, since customers still have all the info they need to reconstruct a statement with getUncommittedStatementIndexes() and assuming they kept track of bind parameters themselves. You OK going ahead without getBindParametersAtIndex() for version 1? We can always add it later if needed.

JamesRTaylor: Ok, that sounds reasonable. Please file a JIRA for this.

phoenix-core/src/main/java/org/apache/phoenix/execute/MutationState.java
```diff
@@ -78,7 +87,11 @@
     private PhoenixConnection connection;
     private final long maxSize;
     private final ImmutableBytesPtr tempPtr = new ImmutableBytesPtr();
-    private final Map<TableRef, Map<ImmutableBytesPtr,Map<PColumn,byte[]>>> mutations = Maps.newHashMapWithExpectedSize(3); // TODO: Sizing?
+    /**
+     * Use {@link TreeMap} for {@link #mutations} below to enforce a deterministic ordering for easier testing of partial failures.
+     * @see PartialCommitIT
+     */
+    private final Map<TableRef, Map<ImmutableBytesPtr,RowMutationState>> mutations = Maps.newTreeMap(new TableRefComparator());
```
JamesRTaylor: TreeMap perf and memory usage isn't nearly as good as HashMap. Can you just sort the results in your test where we don't care about perf?

phoenix-core/src/main/java/org/apache/phoenix/execute/MutationState.java
```diff
@@ -488,7 +488,62 @@ public void rollback(PhoenixConnection connection) throws SQLException {
...
+        Map<PColumn, byte[]> getColumnValues() {
+            return columnValues;
+        }
+
+        Set<Integer> getOrderOfStatementsInConnection() {
+            return orderOfStatementsInConnection;
+        }
+        
+        void join(RowMutationState newRow) {
+            for (Map.Entry<PColumn,byte[]> valueEntry : newRow.getColumnValues().entrySet()) {
```
JamesRTaylor: Can this for loop be replaced with getColumnValues().putAll(newRow.getColumnValues()) ?

phoenix-core/src/main/java/org/apache/phoenix/execute/MutationState.java
```diff
@@ -488,7 +488,62 @@ public void rollback(PhoenixConnection connection) throws SQLException {
...
+    	return result;
+    }
+    
     @Override
     public void close() throws SQLException {
     }
+    
+    public static class RowMutationState {
+        private Map<PColumn,byte[]> columnValues;
+        private Set<Integer> orderOfStatementsInConnection;
```
JamesRTaylor: Rename for consistency with CommitException to statementIndexes here and getter method. I still think a Set is overkill. Can't it just be an int[] of size 1? Then the join can resize it (in the somewhat unlikely event of a collision).

phoenix-core/src/main/java/org/apache/phoenix/execute/MutationState.java
elilevine: They are being sorted here to allow for consistent partial commit verification in tests. Without a predictable sorting here it is impossible to write robust tests for partial commits.

phoenix-core/src/it/java/org/apache/phoenix/execute/PartialCommitIT.java
elilevine: Done: https://issues.apache.org/jira/browse/PHOENIX-1671

phoenix-core/src/main/java/org/apache/phoenix/execute/MutationState.java
elilevine: Good call. Much better.

phoenix-core/src/it/java/org/apache/phoenix/execute/PartialCommitIT.java
elilevine: Added an UPSERT SELECT test in latest version. For the test for both UPSERT VALUES and UPSERT SELECT, should these two set values on two different columns to the same row and the test would simulate that row's failure? Is that what you are looking for?

phoenix-core/src/main/java/org/apache/phoenix/execute/MutationState.java
JamesRTaylor: I understand, but would it be workable to create a copy of the collection in the test and use Collections.sort() to sort it (since we don't care about perf in tests, as we do at runtime)?

phoenix-core/src/main/java/org/apache/phoenix/execute/MutationState.java
elilevine: Done.

phoenix-core/src/main/java/org/apache/phoenix/execute/MutationState.java
elilevine: Not sure I understand how the collection can be separated from MutationState in an integration test, since this collection is internal and determines the order of mutations before a CommitException is thrown.

A core scenario tested in PartialCommitIT is one where there is a mix of mutations that succeed, mutations that fail, and mutations that are unprocessed (but would have succeed if executed in isolation) done on the same Connection.commit(). This scenario is important because of the way MutationState works: it separates mutations by table and sends each table's mutations to HBase iteratively, aborting with CommitException upon the first exception caught from HTableInteface.batch(). In order to write such a test we need to know upfront the order in which tables will be processed by MutationState, so sorting in the test is not enough IMHO.

As you rightly point out in your previous comments, the number of of statements (and thus tables) will be 1 for most common use cases. Sure, in theory TreeSets are slower than HashSets, but seems unlikely to be an issue in practice.

One more option that might be good would be to revert to using a HashMap and call Collections.sort() once during commit(). This avoid keeping a TreeMap sorted on every MutationState.join() in cases where the ratio of join() to commit() calls might be high.

JamesRTaylor: The Map has one entry per row, so it can in theory grow large (upto 500K in theory which is the default max size). How about addition a test-only copy constructor for PhoenixConnection that takes in MutationState as an argument along with a MutationState constructor that takes a Map argument?

elilevine: @JamesRTaylor, can you take a look now? Switched back to using HashCode in MutationState. PartialCommitIT now has the ability to substitue it for a TreeMap to get consistent mutation sort order.

phoenix-core/src/main/java/org/apache/phoenix/jdbc/PhoenixConnection.java
```diff
@@ -426,6 +430,7 @@ public Void call() throws SQLException {
                 return null;
             }
         }, Tracing.withTracing(this, "committing mutations"));
+        statementExecutionCounter = 0;
     }
```
JamesRTaylor: Instead of mucking with the MutationState methods and having to add a setter for the state, how about this (part 1 of 3)?
{code}
        // In PhoenixConnection constructor, have construction of MutationState
        // go through method on PhoenixConnection to enable tests to specialize
        // the method
        this.mutationState = newMutationState(maxSize);
        this.services.addConnection(this);

        // setup tracing, if its enabled
        this.sampler = Tracing.getConfiguredSampler(this);
        this.customTracingAnnotations = getImmutableCustomTracingAnnotations();
    }
    
    protected MutationState newMutationState(int maxSize) {
        return new MutationState(maxSize, this); 
    }

phoenix-core/src/main/java/org/apache/phoenix/execute/MutationState.java
```diff
@@ -92,7 +100,7 @@ public MutationState(int maxSize, PhoenixConnection connection, long sizeOffset)
         this.sizeOffset = sizeOffset;
     }
```
JamesRTaylor: and this (part 2 of 3):

    // For tests to pass through a Map that maintains order (and keep this.mutations as final)
    public MutationState(int maxSize, PhoenixConnection connection, Map<ImmutableBytesPtr,RowMutationState> mutations) {
        this(maxSize,connection,0, mutations);
    }

    public MutationState(int maxSize, PhoenixConnection connection, long sizeOffset) {
        this.maxSize = maxSize;
        this.connection = connection;
        this.sizeOffset = sizeOffset;
        this.mutations = Maps.newHashMapWithExpectedSize(3); // TODO: size based on something else as this is a bad default
    }
    

    public MutationState(int maxSize, PhoenixConnection connection, long sizeOffset, Map<ImmutableBytesPtr,RowMutationState> mutations) {
        this.maxSize = maxSize;
        this.connection = connection;
        this.sizeOffset = sizeOffset;
        this.mutations = mutations;
    }

phoenix-core/src/it/java/org/apache/phoenix/execute/PartialCommitIT.java
```diff
@@ -0,0 +1,299 @@
...
+                assertEquals(expectedCountsForVerification.get(i).intValue(), rs.getInt(1));
+            }
+        } catch (SQLException e) {
+            fail(e.toString());
+        }
+    }
+    
+    private PhoenixConnection getConnectionWithTableOrderPreservingMutationState() throws SQLException {
+        Connection con = driver.connect(url, new Properties());
+        PhoenixConnection phxCon = new PhoenixConnection(con.unwrap(PhoenixConnection.class));
```
JamesRTaylor: and then this (part 3 of 3):

    // Create new connection with ordered map for verification during testing
    PhoenixConnection pcon = con.unwrap(PhoenixConnection.class);
    pcon = new PhoenixConnection(pcon) {
        @Override
        protected MutationState newMutationState(int maxSize) {
            return new MutationState(maxSize, this, Maps.newTreeMap(new TableRefComparator())); 
        }
    };

elilevine: Thanks for the suggestions. All done now. WRT making MutationState.mutations final, I can either (1) make it final and initialize in all 4 constructions, or (2) keep it non-final, initialize once during declaration and only override in the one new constructor. I went with (2). Let me know if you prefer (1) instead. Other than that it's ready to go.

JamesRTaylor: Thanks for the modifications, Eli. How about option (1), but have the existing constructors funnel through MutationState(int maxSize, PhoenixConnection connection) which would construct the HashMap while chaining through to your new constructor? Then you can make the Map final, but only initialize it in your new constructor? 

+1 after that if that sounds ok to you.

JamesRTaylor: One more request, unrelated to your change, but an improvement (so we can remove the TODO about sizing the Map properly). Would you mind sizing the map based on the mutation batch size, which is a more reasonable initial size, like this?

    public MutationState(int maxSize, PhoenixConnection connection, long sizeOffset) {
       this(maxSize,connection,sizeOffset,
              Maps.newHashMapWithExpectedSize(connection.getMutateBatchSize()));
    }

elilevine:
Pushed to main and 4.0 branches. Closing this PR.

stoty:
PHOENIX-900 has already been committed.

PR #38, opened 2015-02-24
Additional DATE tests

michfr:
In support of the recent feature to support ANSI SQL date literals, here are some tests to add to the existing ones.


michfr:
Here are some additional date tests.

phoenix-core/src/it/java/org/apache/phoenix/end2end/QueryIT.java
```diff
@@ -295,6 +299,130 @@ public void testTimestamp() throws Exception {
     }
     
     @Test
+    public void testDateBetweenLiterals() throws Exception {
```
JamesRTaylor: Tests look good, but let's put them in a new class derived from BaseHBaseManagedTimeIT as we're trying to move away from tests that manage the timestamp through the CURRENT_SCN since it's much less common than just letting HBase manage the timestamps. It's also easier to write your test, because you can just open a single connection and do everything on it.

PR #39, opened 2015-03-05
Phoenix-1580 union all impl

ayingshu:
Implementation using wrapped TupleProjectionPlan. My own implementation of association RowProjector with ResultIterator was commented out in code.

### DIFF
--- a/phoenix-core/src/main/java/org/apache/phoenix/compile/QueryCompiler.java
+++ b/phoenix-core/src/main/java/org/apache/phoenix/compile/QueryCompiler.java
@@ -19,10 +19,12 @@ package org.apache.phoenix.compile;
 
 import java.sql.SQLException;
 import java.sql.SQLFeatureNotSupportedException;
+import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
 import java.util.Set;
 
+import org.apache.hadoop.hbase.HConstants;
 import org.apache.hadoop.hbase.client.Scan;
 import org.apache.hadoop.hbase.util.Pair;
 import org.apache.phoenix.compile.GroupByCompiler.GroupBy;
@@ -30,6 +32,8 @@ import org.apache.phoenix.compile.JoinCompiler.JoinSpec;
 import org.apache.phoenix.compile.JoinCompiler.JoinTable;
 import org.apache.phoenix.compile.JoinCompiler.Table;
 import org.apache.phoenix.compile.OrderByCompiler.OrderBy;
+import org.apache.phoenix.exception.SQLExceptionCode;
+import org.apache.phoenix.exception.SQLExceptionInfo;
 import org.apache.phoenix.execute.AggregatePlan;
 import org.apache.phoenix.execute.ClientAggregatePlan;
 import org.apache.phoenix.execute.ClientScanPlan;
@@ -40,8 +44,10 @@ import org.apache.phoenix.execute.ScanPlan;
 import org.apache.phoenix.execute.SortMergeJoinPlan;
 import org.apache.phoenix.execute.TupleProjectionPlan;
 import org.apache.phoenix.execute.TupleProjector;
+import org.apache.phoenix.execute.UnionPlan;
 import org.apache.phoenix.expression.Expression;
 import org.apache.phoenix.expression.LiteralExpression;
+import org.apache.phoenix.expression.ProjectedColumnExpression;
 import org.apache.phoenix.expression.RowValueConstructorExpression;
 import org.apache.phoenix.hbase.index.util.ImmutableBytesPtr;
 import org.apache.phoenix.iterate.ParallelIteratorFactory;
@@ -63,14 +69,20 @@ import org.apache.phoenix.parse.TableNode;
 import org.apache.phoenix.query.QueryConstants;
 import org.apache.phoenix.schema.AmbiguousColumnException;
 import org.apache.phoenix.schema.ColumnNotFoundException;
+import org.apache.phoenix.schema.PColumn;
+import org.apache.phoenix.schema.PColumnFamily;
+import org.apache.phoenix.schema.PColumnImpl;
 import org.apache.phoenix.schema.PDatum;
+import org.apache.phoenix.schema.PNameFactory;
 import org.apache.phoenix.schema.PTable;
+import org.apache.phoenix.schema.PTableImpl;
+import org.apache.phoenix.schema.TableNotFoundException;
 import org.apache.phoenix.schema.TableRef;
+import org.apache.phoenix.schema.types.PDataType;
 import org.apache.phoenix.util.ScanUtil;
 
 import com.google.common.collect.Lists;
 import com.google.common.collect.Sets;
-
 
 
 /**
@@ -100,6 +112,7 @@ public class QueryCompiler {
     private final boolean projectTuples;
     private final boolean useSortMergeJoin;
     private final boolean noChildParentJoinOptimization;
+    private final boolean isUnionAll;
 
     public QueryCompiler(PhoenixStatement statement, SelectStatement select, ColumnResolver resolver) throws SQLException {
         this(statement, select, resolver, Collections.<PDatum>emptyList(), null, new SequenceManager(statement), true);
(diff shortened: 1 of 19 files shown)


phoenix-core/src/main/java/org/apache/phoenix/compile/QueryCompiler.java
```diff
@@ -147,21 +169,146 @@ public QueryCompiler(PhoenixStatement statement, SelectStatement select, ColumnR
...
-            return compileJoinQuery(context, binds, joinTable, false, false, null);
-        } else {
-            return compileSingleQuery(context, select, binds, false, true);
+            plan = compileSelect(unionAllSelects.get(i));
+            if (containOrderBy) {
+                orderBy = plan.getOrderBy();
+                unionAllSelects.get(i).removeOrderBy();
+                plan = compileSelect(unionAllSelects.get(i));
+            } 
+            plan = buildTupleProjectPlan(plan);
```
maryannxue: Actually the buildTupleProjectPlan() is just a couple lines, maybe you don't even have to write a function for this. The code would like:

TupleProjector projector = new TupleProjector(plan.getRowProjector());
plan = new TupleProjectionPlan(plan, projector, null);

phoenix-core/src/main/java/org/apache/phoenix/compile/QueryCompiler.java
```diff
@@ -147,21 +169,146 @@ public QueryCompiler(PhoenixStatement statement, SelectStatement select, ColumnR
...
+            plans.add(plan);
         }
+        checkProjectionNumAndTypes(plans);
+        plan =  new UnionPlan(context, plan.getStatement(), plan.getTableRef(), plan.getProjector(), plan.getLimit(), orderBy, parallelIteratorFactory, plan.getGroupBy(), plans);
```
maryannxue: Should not pass the inner plan's attributes to UnionPlan, instead you should create its own tableRef, rowProjector, limit, orderBy, groupBy, etc.

For orderBy, groupBy and limit, I think we should just leave them as NULL. And the tableRef and rowProjector is what really matters here.

Since we have already used TupleProjectionPlan earlier, here the tableRef and rowProjector is more like a flat dummy one.

We should create a table like (pretty much what createTempTableForUnionAllResultResolver() does):
List<PColumn> columns = new ArrayList<PColumn>();
// we'll use one of inner plan's row projector just for type information.
for (ColumnProjector colProjector : plan.getProjector().getColumnProjectors()) {
    PColumn column = new PColumnImpl(colProject.getName(), ..., colProjector.getExpression().getType(), ...);
}
PTable tempTable = PTableImpl.makePTable("dummyName", "dummySchemaName", PTableType.SUBQUERY, columns);
TableRef tableRef = new TableRef(null, tempTable, 0);

GroupBy groupBy = GroupBy.EMPTY_GROUP_BY;
OrderBy orderBy = OrderBy.EMPTY_ORDER_BY;
context.setResolver(FromCompiler.getResolver(tableRef));
RowProjector rowProjector = ProjectorCompiler.compile(context, select, groupBy, Collections.<PDatum>emptyList());

plan = new UnionPlan(context, this.statement, tableRef, rowProjector, null, orderBy, parallelIteratorFactory, groupBy, plans);

phoenix-core/src/main/java/org/apache/phoenix/compile/QueryCompiler.java
```diff
@@ -147,21 +169,146 @@ public QueryCompiler(PhoenixStatement statement, SelectStatement select, ColumnR
...
+    private void checkForOrderByLimitInUnionAllSelect(SelectStatement select) throws SQLException {
+        if (select.getOrderBy() != null && !select.getOrderBy().isEmpty()) {
+            throw new SQLExceptionInfo.Builder(SQLExceptionCode.ORDER_BY_IN_UNIONALL_SELECT_NOT_SUPPORTED).setMessage(".").build().buildException();
+        }
+        if (select.getLimit() != null) {
+            throw new SQLExceptionInfo.Builder(SQLExceptionCode.LIMIT_IN_UNIONALL_SELECT_NOT_SUPPORTED).setMessage(".").build().buildException();
+        }
+    }
+
+    private PTable createTempTableForUnionAllResultResolver(QueryPlan plan) throws SQLException {
```
maryannxue: The implementation is pretty good, but should be called elsewhere. Please refer to the comment below.

phoenix-core/src/main/java/org/apache/phoenix/compile/QueryCompiler.java
```diff
@@ -147,21 +169,146 @@ public QueryCompiler(PhoenixStatement statement, SelectStatement select, ColumnR
...
+            PColumnImpl projectedColumn = new PColumnImpl(PNameFactory.newName(colProj.getName().getBytes()), table.getDefaultFamilyName(),
+                    sourceExpression.getDataType(), sourceExpression.getMaxLength(), sourceExpression.getScale(), sourceExpression.isNullable(),
+                    i, sourceExpression.getSortOrder(), 50, new byte[0], true, sourceExpression.toString());
+            projectedColumns.add(projectedColumn);
+        }
+        PTable t = PTableImpl.makePTable(theTable, projectedColumns);
+        return t;
+    }
+
+    private QueryPlan buildTupleProjectPlan(QueryPlan plan) throws SQLException {
```
maryannxue: This method would not be used.

phoenix-core/src/main/java/org/apache/phoenix/compile/QueryCompiler.java
ayingshu: OrderBy is used by union all plan to get Order by information. Cannot be null if it has order by clause.  

When building the RowProjector, which "select" you referred to here. There are many selects in union all. Not matter which one you choose, effectively you pick one of the selects to build the RowProjector. It is the problem I have mentioned. And that is why I pass down RowProjector to result sets!
RowProjector rowProjector = ProjectorCompiler.compile(context, select, groupBy, Collections.emptyList());

ayingshu: OrderBy is used by union all plan to get Order by information. Cannot be null if it has order by clause.

When building the RowProjector, which "select" you referred to here. There are many selects in union all. Not matter which one you choose, effectively you pick one of the selects to build the RowProjector. It is the problem I have mentioned. And that is why I pass down RowProjector to result sets.
RowProjector rowProjector = ProjectorCompiler.compile(context, select, groupBy, Collections.emptyList());

phoenix-core/src/main/java/org/apache/phoenix/compile/QueryCompiler.java
ayingshu: I was wrapping each subselect with TupleProjectionPlan suggested by you. So the method buildTupleProjectPlan(plan) contains both creating the PTable and building the TupleProjectionPlan. It was called by each subselect plan. That is why it is wrapped into a method. 

Look like you do not want to wrap the individual subselects and want to use its plan RowProjector for each sub plan. Are you back to using TupleProjectionPlan for final UnionPlan only? I tested this option before and got the same wrong result.

phoenix-core/src/main/java/org/apache/phoenix/compile/QueryCompiler.java
ayingshu: See above comments about buildTupleProjectPlan(QueryPlan plan).

phoenix-core/src/main/java/org/apache/phoenix/compile/QueryCompiler.java
ayingshu: I was wrapping each subselect with TupleProjectionPlan suggested by you. So the method buildTupleProjectPlan(plan) contains both creating the PTable and building the TupleProjectionPlan. It was called by each subselect plan. That is why it is wrapped into a method.

Look like you do not want to wrap the individual subselects and want to use its plan RowProjector for each sub plan. Are you back to using TupleProjectionPlan for final UnionPlan only? I tested this option before and got the same wrong result.

phoenix-core/src/main/java/org/apache/phoenix/compile/QueryCompiler.java
ayingshu: OrderBy is used by union all plan to get Order by information. Cannot be null if it has order by clause.

When building the RowProjector, which "select" you referred to here. There are many selects in union all. Not matter which one you choose, effectively you pick one of the selects to build the RowProjector. It is the problem I have mentioned. And that is why I pass down RowProjector to result sets.
RowProjector rowProjector = ProjectorCompiler.compile(context, select, groupBy, Collections.emptyList());

phoenix-core/src/main/java/org/apache/phoenix/compile/QueryCompiler.java
ayingshu: See above comments about buildTupleProjectPlan(QueryPlan plan).

stoty:
PHOENIX-1580 has already been committed.

PR #40, opened 2015-03-09
PHOENIX-1715 Implement Build-in Math function sign

shuxiong:
Hi,

I am a student from Peking University, and I'd like to participate GSOC2015.

This patch is an implementation of Build-in math function Sign.

I've already create a subtask[2] under issues [1].

Please help me to check it, and any comments is welcome.

[1] https://issues.apache.org/jira/browse/PHOENIX-1660 
[2] https://issues.apache.org/jira/browse/PHOENIX-1715

### DIFF
--- a/phoenix-core/src/it/java/org/apache/phoenix/end2end/TimezoneOffsetFunctionIT.java
+++ b/phoenix-core/src/it/java/org/apache/phoenix/end2end/TimezoneOffsetFunctionIT.java
@@ -23,7 +23,11 @@ import static org.junit.Assert.fail;
 
 import java.sql.Connection;
 import java.sql.DriverManager;
+import java.sql.PreparedStatement;
 import java.sql.ResultSet;
+import java.sql.Timestamp;
+import java.util.Calendar;
+import java.util.TimeZone;
 
 import org.apache.phoenix.schema.IllegalDataException;
 import org.junit.Test;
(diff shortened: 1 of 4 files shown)


phoenix-core/src/main/java/org/apache/phoenix/expression/function/SignFunction.java
```diff
@@ -0,0 +1,77 @@
...
+ *
+ * @since 4.3.0
+ *
+ */
+@BuiltInFunction(name = SignFunction.NAME,
+                 args = {
+                        @Argument(allowedTypes={PDecimal.class})
+                        }
+                )
+public class SignFunction extends ScalarFunction {
```
JamesRTaylor: Nice work! I think performance of this can be improved by not having to create a BigDecimal in evaluate. One way would be to introduce a PDataType.getSign() method that would work directly from the numeric type of the child (i.e. long, int, short, byte). You'd need to maybe throw in cases where getSign doesn't make sense, like with VARCHAR or CHAR. Another way would be to introduce a PNumericType and introduce the getSign method there. Then reparent all the numeric PDataType to this new class. You'd know you could cast the childExpr.getDataType() method to PNumericType because we guarantee that at compile time.

shuxiong: Hi James,

One question:

In PDataType.java,

  public abstract Object toObject(byte[] bytes, int offset, int length, PDataType actualType,
      SortOrder sortOrder, Integer maxLength, Integer scale);

What does "SortOrder sortOrder, Integer maxLength, Integer scale" mean here?

-------------------

I'd like to take the second solution, which introduces a new type PNumericType. 

So in PNumericType, we have an abstract function getSign, which is like function toObject in PDataType, which we implement the details in each subclass, PInteger, PFloat etc.

How about the solution above?

Thanks.

shuxiong: Hi James,

One question:

In PDataType.java,

  public abstract Object toObject(byte[] bytes, int offset, int length, PDataType actualType,
      SortOrder sortOrder, Integer maxLength, Integer scale);

What does "SortOrder sortOrder, Integer maxLength, Integer scale" mean here?

-------------------

I'd like to take the second solution, which introduces a new type PNumericType. 

So in PNumericType, we have an abstract function getSign, which is like function toObject in PDataType, which we implement the details in each subclass, PInteger, PFloat etc.

How about the solution above?

Thanks.

shuxiong: Sorry, I mistakenly close this request.

PR #42, opened 2015-03-10
PHOENIX-1722 Speedup CONVERT_TZ function

tzolkincz:
Using Joda Time lib instead of java.util.TimeZone. This would speedup this function more than 3 times. I've also updated version of Joda Time to 2.7 cause of some timezone bugfixes and speedups.

### DIFF
--- a/phoenix-core/src/it/java/org/apache/phoenix/end2end/ConvertTimezoneFunctionIT.java
+++ b/phoenix-core/src/it/java/org/apache/phoenix/end2end/ConvertTimezoneFunctionIT.java
@@ -23,8 +23,10 @@ import java.sql.Connection;
 import java.sql.DriverManager;
 import java.sql.ResultSet;
 import java.sql.SQLException;
+import java.sql.Statement;
 
 import org.apache.phoenix.exception.SQLExceptionCode;
+import static org.junit.Assert.assertFalse;
 import org.junit.Test;
 
 /**
@@ -129,7 +131,7 @@ public class ConvertTimezoneFunctionIT extends BaseHBaseManagedTimeIT {
         try {
             ResultSet rs = conn.createStatement().executeQuery(
                     "SELECT k1, dates, CONVERT_TZ(dates, 'UNKNOWN_TIMEZONE', 'America/Adak') FROM TIMEZONE_OFFSET_TEST");
-    
+
             rs.next();
             rs.getDate(3).getTime();
             fail();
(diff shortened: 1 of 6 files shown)


tzolkincz:
@gabrielreid I made changes as you suggested.

gabrielreid:
Patch looks good, but I've got the one remark/question that I put on the commit. Can you take a look at that and let me know what you think?

tzolkincz:
Thank you for review @gabrielreid, i've solved that.
I realized [TIMEZONE_OFFSET](https://github.com/apache/phoenix/blob/master/phoenix-core/src/main/java/org/apache/phoenix/expression/function/TimezoneOffsetFunction.java#L48) have similar issue. I think we should rather extract timezone cache mechanism and make it as part of org.apache.phoenix.cache. Or are there some other use-cases of (instance) cache so we could implement it more generally?

gabrielreid:
I don't think it's the same issue in TIMEZONE_OFFSET because the timezone cache isn't static, so (as far as I know) it will only be accessed by a single thread. That being said, I think it makes sense to make the same change there as well, if you're up for doing that.

tzolkincz:
Yes, it should be thread safe but we could consider to use same lib for timezones and same caching mechanism.  So if you are for duplication of [caching](https://github.com/apache/phoenix/pull/42/files#diff-83862191fb126a769d5fd74be3534d90R98) code, I'll do that. It is small fraction of code, but I'd rather make it more abstract. However it's not major question here.

gabrielreid:
Making it more abstract and not having duplication of code definitely sounds like the way to go, if you're up for doing that it would be great.

JamesRTaylor:
@tzolkincz - we're about to cut a 4.3.1 RC, so if you need this in, we'll need it today. Otherwise, 4.4.0 may be your next release vehicle.

tzolkincz:
Thanks for notification.
I'm done with my changes.

JamesRTaylor:
@gabrielreid - if this looks good to you, please commit to 4.3 and the other branches. Thanks for the contributions and reviews!

gabrielreid:
@tzolkincz I just tried running this on a Phoenix install, and there seems to be an issue with the code now if I use the CONVERT_TZ function over multiple rows.

I'm able to replicate it in ConvertTimezoneFunctionIT with the following test:

    @Test
    public void testConvertMultipleRecords() throws Exception {
        Connection conn = DriverManager.getConnection(getUrl());
        String ddl = "CREATE TABLE IF NOT EXISTS TIMEZONE_OFFSET_TEST (k1 INTEGER NOT NULL, dates DATE CONSTRAINT pk PRIMARY KEY (k1))";
        Statement stmt = conn.createStatement();
        stmt.execute(ddl);
        stmt.execute( "UPSERT INTO TIMEZONE_OFFSET_TEST (k1, dates) VALUES (1, TO_DATE('2014-03-01 00:00:00'))");
        stmt.execute( "UPSERT INTO TIMEZONE_OFFSET_TEST (k1, dates) VALUES (2, TO_DATE('2014-03-01 00:00:00'))");
        conn.commit();

        ResultSet rs = stmt.executeQuery(
                "SELECT k1, dates, CONVERT_TZ(dates, 'UTC', 'America/Adak') FROM TIMEZONE_OFFSET_TEST");

        assertTrue(rs.next());
        assertEquals(1393596000000L, rs.getDate(3).getTime()); //Fri, 28 Feb 2014 14:00:00
        assertTrue(rs.next());
        assertEquals(1393596000000L, rs.getDate(3).getTime()); //Fri, 28 Feb 2014 14:00:00
        assertFalse(rs.next());
    }

Running that test results in the following:

    java.sql.SQLException: ERROR 201 (22000): Illegal data. Unknown timezone �Dx�gAdak
    	at org.apache.phoenix.exception.SQLExceptionCode$Factory$1.newException(SQLExceptionCode.java:362)
    	at org.apache.phoenix.exception.SQLExceptionInfo.buildException(SQLExceptionInfo.java:133)
    	at org.apache.phoenix.schema.IllegalDataException.<init>(IllegalDataException.java:38)
    	at org.apache.phoenix.cache.JodaTimezoneCache.getInstance(JodaTimezoneCache.java:48)
    	at org.apache.phoenix.cache.JodaTimezoneCache.getInstance(JodaTimezoneCache.java:60)
    	at org.apache.phoenix.expression.function.ConvertTimezoneFunction.evaluate(ConvertTimezoneFunction.java:70)
    	at org.apache.phoenix.compile.ExpressionProjector.getValue(ExpressionProjector.java:69)
    	at org.apache.phoenix.jdbc.PhoenixResultSet.getDate(PhoenixResultSet.java:356)
    	at org.apache.phoenix.end2end.ConvertTimezoneFunctionIT.testConvertMultipleRecords(ConvertTimezoneFunctionIT.java:80)

Could you look into what's going wrong?

tzolkincz:
I'm on it, it looks strange.

tzolkincz:
yes my bad. Fixed

samarthjain:
@gabrielreid - does this look good for commit now? We are planning on cutting the RC for 4.3.1 pretty soon.

samarthjain:
Tests ran fine for me locally so I went ahead and committed the patch.

stoty:
This has already been committed.

PR #43, opened 2015-03-10
PHOENIX-1722 Speedup CONVERT_TZ function

tzolkincz:
Using Joda Time lib instead of java.util.TimeZone. This would speedup this function more than 3 times. I've also updated version of Joda Time to 2.7 cause of some timezone bugfixes and speedups.


stoty:
This has already been committed.

PR #45, opened 2015-03-10
PHOENIX-1715 Implement Build-in Math function sign

shuxiong:
1. Add Build-in Math Function Sign
2. Add a new Type, PNumericType, which is superclass of all numeric type, such as PInteger, PFloat etc.
    PNumericType has method getSign

   2.1 All integer types, PInteger, PLong etc., will compute sign result by checking the content bytes.
   2.2 All float types, PFloat, PDouble etc, will construct objects and compare with zero, for their complex content bytes.

I mistakenly close last pull request, sorry for that.

Thanks.

### DIFF
--- a/phoenix-core/src/main/java/org/apache/phoenix/expression/function/SignFunction.java
+++ b/phoenix-core/src/main/java/org/apache/phoenix/expression/function/SignFunction.java
@@ -1,0 +1,88 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.phoenix.expression.function;
+
+import java.sql.SQLException;
+import java.util.List;
+
+import org.apache.hadoop.hbase.io.ImmutableBytesWritable;
+import org.apache.phoenix.expression.Expression;
+import org.apache.phoenix.parse.FunctionParseNode.Argument;
+import org.apache.phoenix.parse.FunctionParseNode.BuiltInFunction;
+import org.apache.phoenix.schema.tuple.Tuple;
+import org.apache.phoenix.schema.types.PDataType;
+import org.apache.phoenix.schema.types.PDecimal;
+import org.apache.phoenix.schema.types.PDouble;
+import org.apache.phoenix.schema.types.PFloat;
+import org.apache.phoenix.schema.types.PInteger;
+import org.apache.phoenix.schema.types.PLong;
+import org.apache.phoenix.schema.types.PNumericType;
+import org.apache.phoenix.schema.types.PSmallint;
+import org.apache.phoenix.schema.types.PTinyint;
+import org.apache.phoenix.schema.types.PUnsignedDouble;
+import org.apache.phoenix.schema.types.PUnsignedFloat;
+import org.apache.phoenix.schema.types.PUnsignedInt;
+import org.apache.phoenix.schema.types.PUnsignedLong;
+import org.apache.phoenix.schema.types.PUnsignedSmallint;
+import org.apache.phoenix.schema.types.PUnsignedTinyint;
+
+/**
+ * Base class for built-in SIGN function.
+ * @since 4.3.0
+ */
+@BuiltInFunction(name = SignFunction.NAME, args = { @Argument(allowedTypes = { PFloat.class,
+        PDouble.class, PDecimal.class, PUnsignedFloat.class, PUnsignedDouble.class, PInteger.class,
+        PUnsignedInt.class, PLong.class, PUnsignedLong.class, PSmallint.class,
+        PUnsignedSmallint.class, PTinyint.class, PUnsignedTinyint.class }) })
+public class SignFunction extends ScalarFunction {
+
+    public static final String NAME = "SIGN";
+
+    private static final byte[][] RESULT = { PInteger.INSTANCE.toBytes(Integer.valueOf(-1)),
+            PInteger.INSTANCE.toBytes(Integer.valueOf(0)),
+            PInteger.INSTANCE.toBytes(Integer.valueOf(1)), };
+
+    public SignFunction() {
+    }
+
+    public SignFunction(List<Expression> children) throws SQLException {
+        super(children);
+    }
+
+    @Override
+    public boolean evaluate(Tuple tuple, ImmutableBytesWritable ptr) {
+        Expression childExpr = children.get(0);
+        PDataType dataType = childExpr.getDataType();
+        if (childExpr.evaluate(tuple, ptr)) {
+            int ret = ((PNumericType) dataType).signum(ptr, childExpr.getSortOrder());
+            ptr.set(RESULT[ret + 1]);
+            return true;
+        }
+        return false;
+    }
+
+    @Override
+    public PDataType getDataType() {
+        return PInteger.INSTANCE;
+    }
+
+    @Override
+    public String getName() {
+        return NAME;
+    }
+}
--- a/phoenix-core/src/main/java/org/apache/phoenix/schema/types/PDecimal.java
+++ b/phoenix-core/src/main/java/org/apache/phoenix/schema/types/PDecimal.java
@@ -31,7 +31,7 @@ import org.apache.phoenix.util.NumberUtil;
 
 import com.google.common.base.Preconditions;
 
-public class PDecimal extends PDataType<BigDecimal> {
+public class PDecimal extends PRealNumber<BigDecimal> {
 
   public static final PDecimal INSTANCE = new PDecimal();
 
@@ -402,4 +402,20 @@ public class PDecimal extends PDataType<BigDecimal> {
   public Object getSampleValue(Integer maxLength, Integer arrayLength) {
     return new BigDecimal((Long) PLong.INSTANCE.getSampleValue(maxLength, arrayLength));
   }
+
+    // take details from org.apache.phoenix.schema.types.PDataType#toBigDecimal(byte[], int, int)
+    @Override
+    public int signum(byte[] bytes, int offset, int length, SortOrder sortOrder, Integer maxLength,
+            Integer scale) {
+        byte signByte;
+        if (sortOrder == SortOrder.DESC) {
+            signByte = SortOrder.invert(bytes[offset]);
+        } else {
+            signByte = bytes[offset];
+        }
+        if (length == 1 && signByte == ZERO_BYTE) {
+            return 0;
+        }
+        return ((signByte & 0x80) == 0) ? -1 : 1;
+    }
 }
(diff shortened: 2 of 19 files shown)


phoenix-core/src/main/java/org/apache/phoenix/schema/types/PDouble.java
```diff
@@ -308,4 +309,28 @@ public PhoenixArray newArray(PDataType type, Object[] elements) {
...
+    public int getSign(byte[] bytes, int offset, int length, SortOrder sortOrder,
+            Integer maxLength, Integer scale) {
+        Preconditions.checkNotNull(sortOrder);
+        checkForSufficientLength(bytes, offset, Bytes.SIZEOF_LONG);
+        if (sortOrder == SortOrder.DESC) {
+            bytes =
+                    SortOrder.invert(bytes, offset, new byte[Bytes.SIZEOF_DOUBLE], 0,
+                        Bytes.SIZEOF_DOUBLE);
+            offset = 0;
+        }
```
JamesRTaylor: To prevent the creation of a Long, you can use the following and remove your sortOrder logic to get the double here:
    double d = getCodec().decodeDouble(bytes, offset, sortOrder);

Then you can use Math.signum(d) to get the sign for simplification. Note that this code will work for PFloat, PDouble, PUnsignedFloat, and PUnsignedDouble, so it's probably worth having another intermediate class here, something like PRealNumber.

phoenix-core/src/main/java/org/apache/phoenix/schema/types/PInteger.java
```diff
@@ -272,4 +273,10 @@ public PhoenixArray newArray(PDataType type, Object[] elements) {
       };
     }
   }
+
+  @Override
+  public int getSign(byte[] bytes, int offset, int length, SortOrder sortOrder,
+          Integer maxLength, Integer scale) {
+      return SignUtils.getSignForSignedInteger(bytes, offset, length, sortOrder, Bytes.SIZEOF_INT);
```
JamesRTaylor: Same idea here. Create an intermediate class, called PWholeNumber or PWholeNumeric and derive PLong, PInteger, PShort, PByte and all the unsigned variations from this one. Then you can do something like this:

    long l = getCodec().decodeLong(bytes, offset, sortOrder);
    return l == 0 ? 0 : l < 0 ? -1 : 1; // Or maybe a mask of the most significant bit is better?

shuxiong:
Hi @JamesRTaylor ,
I update the code, adding two more types, PWholeNumber and PRealNumber. and rename the function from getSign to signum.

Thanks.

phoenix-core/src/main/java/org/apache/phoenix/expression/function/SignFunction.java
```diff
@@ -0,0 +1,85 @@
...
+        super(children);
+    }
+
+    @Override
+    public boolean evaluate(Tuple tuple, ImmutableBytesWritable ptr) {
+        Expression childExpr = children.get(0);
+        PDataType dataType = childExpr.getDataType();
+        if (childExpr.evaluate(tuple, ptr)) {
+            byte[] retBytes;
+            if (dataType instanceof PWholeNumber) {
```
JamesRTaylor: Remove the if statement, declare int signum(ImmutableBytesWritable, SortOrder) at the common base class of PNumericType, and implement differently for PWholeNumber and PRealNumber. This code should look the way you had it before.

phoenix-core/src/main/java/org/apache/phoenix/schema/types/PDecimal.java
```diff
@@ -402,4 +402,25 @@ public String toStringLiteral(Object o, Format formatter) {
   public Object getSampleValue(Integer maxLength, Integer arrayLength) {
     return new BigDecimal((Long) PLong.INSTANCE.getSampleValue(maxLength, arrayLength));
   }
+
+    // take details from org.apache.phoenix.schema.types.PDataType#toBigDecimal(byte[], int, int)
+    @Override
+    public double signum(byte[] bytes, int offset, int length, SortOrder sortOrder,
```
JamesRTaylor: return int here

phoenix-core/src/main/java/org/apache/phoenix/schema/types/PDouble.java
```diff
@@ -308,4 +309,10 @@ public PhoenixArray newArray(PDataType type, Object[] elements) {
       };
     }
   }
+
```
JamesRTaylor: Not needed - just use PDouble.INSTANCE

phoenix-core/src/main/java/org/apache/phoenix/schema/types/PFloat.java
```diff
@@ -305,4 +306,10 @@ public PhoenixArray newArray(PDataType type, Object[] elements) {
       };
     }
   }
+
```
JamesRTaylor: Same - remove.

phoenix-core/src/main/java/org/apache/phoenix/schema/types/PInteger.java
```diff
@@ -272,4 +273,10 @@ public PhoenixArray newArray(PDataType type, Object[] elements) {
       };
     }
   }
+
```
JamesRTaylor: remove

phoenix-core/src/main/java/org/apache/phoenix/schema/types/PLong.java
```diff
@@ -328,4 +329,10 @@ public PhoenixArray newArray(PDataType type, Object[] elements) {
       };
     }
   }
+
```
JamesRTaylor: remove

phoenix-core/src/main/java/org/apache/phoenix/schema/types/PRealNumber.java
```diff
@@ -0,0 +1,44 @@
...
+import org.apache.phoenix.schema.SortOrder;
+
+public abstract class PRealNumber<T> extends PNumericType<T> {
+
+    protected PRealNumber(String sqlTypeName, int sqlType, Class clazz,
+            org.apache.phoenix.schema.types.PDataType.PDataCodec codec, int ordinal) {
+        super(sqlTypeName, sqlType, clazz, codec, ordinal);
+    }
+
+    public final double signum(byte[] bytes, int offset, int length, SortOrder sortOrder) {
```
JamesRTaylor: declare these as returning an int

phoenix-core/src/main/java/org/apache/phoenix/schema/types/PSmallint.java
```diff
@@ -256,4 +257,10 @@ public PhoenixArray newArray(PDataType type, Object[] elements) {
           };
       }
     }
+
```
JamesRTaylor: remove

phoenix-core/src/main/java/org/apache/phoenix/schema/types/PUnsignedRealNumber.java
```diff
@@ -0,0 +1,39 @@
...
+import org.apache.phoenix.schema.SortOrder;
+
+public abstract class PUnsignedRealNumber<T> extends PRealNumber<T> {
+
+    protected PUnsignedRealNumber(String sqlTypeName, int sqlType, Class clazz,
+            org.apache.phoenix.schema.types.PDataType.PDataCodec codec, int ordinal) {
+        super(sqlTypeName, sqlType, clazz, codec, ordinal);
+    }
+
+    @Override
```
JamesRTaylor: not necessary

phoenix-core/src/main/java/org/apache/phoenix/schema/types/PUnsignedWholeNumber.java
```diff
@@ -0,0 +1,39 @@
...
+import org.apache.phoenix.schema.SortOrder;
+
+public abstract class PUnsignedWholeNumber<T> extends PWholeNumber<T> {
+
+    protected PUnsignedWholeNumber(String sqlTypeName, int sqlType, Class clazz,
+            org.apache.phoenix.schema.types.PDataType.PDataCodec codec, int ordinal) {
+        super(sqlTypeName, sqlType, clazz, codec, ordinal);
+    }
+
+    @Override
```
JamesRTaylor: unnecessary

phoenix-core/src/main/java/org/apache/phoenix/schema/types/PWholeNumber.java
```diff
@@ -0,0 +1,45 @@
...
+
+public abstract class PWholeNumber<T> extends PNumericType<T> {
+
+    protected PWholeNumber(String sqlTypeName, int sqlType, Class clazz,
+            org.apache.phoenix.schema.types.PDataType.PDataCodec codec, int ordinal) {
+        super(sqlTypeName, sqlType, clazz, codec, ordinal);
+    }
+
+
+    public final int signum(byte[] bytes, int offset, int length, SortOrder sortOrder) {
```
JamesRTaylor: These can be defined at PNumericType.

phoenix-core/src/main/java/org/apache/phoenix/expression/function/SignFunction.java
shuxiong: 1) According to [1], signum's return type should be as same as argument type.
2) Math#signum has a special case: it will return Nan, if the argument is Nan.

So I choose to return different types for whole and real number.

When getting the result, we don't know its types. I come up with two ways to construct the object,
1) try each type one by one, or
2) achieve as it is.

[1] http://www.postgresql.org/docs/current/static/functions-math.html

JamesRTaylor:
@shuxiong - you're doing a great job so far. Just revert SignFunction back to your original version, keep the PWholeNumber and PRealNumber classes you created but make signum polymorphic, and throw an exception in the case of NaN for double (though I don't think it's possible for that to occur in Phoenix), and you'll be done with this one and ready to move on to other bigger issues. Appreciate your work here.

shuxiong:
Hi @JamesRTaylor  , I already revert the changes and combine all commits into a single one.

phoenix-core/src/main/java/org/apache/phoenix/expression/function/SignFunction.java
```diff
@@ -0,0 +1,71 @@
...
+        super(children);
+    }
+
+    @Override
+    public boolean evaluate(Tuple tuple, ImmutableBytesWritable ptr) {
+        Expression childExpr = children.get(0);
+        PDataType dataType = childExpr.getDataType();
+        if (childExpr.evaluate(tuple, ptr)) {
+            int ret = ((PNumericType) dataType).signum(ptr, childExpr.getSortOrder());
+            byte[] retBytes = PInteger.INSTANCE.toBytes(Integer.valueOf(ret));
```
JamesRTaylor: Can you restore back the way you did an RESULT[ret] to get the bytes of the return value here please?

phoenix-core/src/main/java/org/apache/phoenix/schema/types/PDouble.java
```diff
@@ -17,15 +17,16 @@
  */
 package org.apache.phoenix.schema.types;
 
-import com.google.common.base.Preconditions;
-import com.google.common.primitives.Doubles;
```
JamesRTaylor: Minor, but can you import into Eclipse our phoenix/dev/phoenix.importorder and re-save these files so that the import order doesn't change?

JamesRTaylor:
Couple of very minor items, but otherwise we're good to go. Excellent work, @shuxiong. Nice first contribution.

phoenix-core/src/main/java/org/apache/phoenix/schema/types/PDouble.java
shuxiong: I've already imported the Import Order Config Files, and the rule is:

java
org
com

That's why my eclipse changes the order.

JamesRTaylor: Ah, ok. My bad. Must have been the someone else before you. Thanks for correcting. I'll get this committed. Great work!

shuxiong: Should I combine all the commits into a single one?
That is one issue for one commit.

PR #46, opened 2015-03-15
Draft for PHOENIX-1287 subtask, ByteBasedLikeExpression

shuxiong:
Hi @JamesRTaylor ,

This is a draft for
1. Abstraction for Regex Engine, byte-based one(JONI) and string-based one(j.u.regex)
2. ByteBasedLikeExpression and StringBasedLikeExpression based on 1.

Comment if you think there is anything to be improved.
Then I will work on other expressions(RegexpReplaceFunction, RegexpSplitFunction, RegexpSubstrFunction) next.

Thanks.

### DIFF
--- a/phoenix-core/src/main/java/org/apache/phoenix/expression/LikeExpression.java
+++ b/phoenix-core/src/main/java/org/apache/phoenix/expression/LikeExpression.java
@@ -21,11 +21,12 @@ import java.io.DataInput;
 import java.io.DataOutput;
 import java.io.IOException;
 import java.util.List;
-import java.util.regex.Pattern;
 
 import org.apache.hadoop.hbase.io.ImmutableBytesWritable;
+import org.apache.phoenix.expression.util.regex.AbstractBasePattern;
 import org.apache.phoenix.expression.visitor.ExpressionVisitor;
 import org.apache.phoenix.parse.LikeParseNode.LikeType;
+import org.apache.phoenix.schema.SortOrder;
 import org.apache.phoenix.schema.tuple.Tuple;
 import org.apache.phoenix.schema.types.PBoolean;
 import org.apache.phoenix.schema.types.PDataType;
@@ -49,7 +50,7 @@ import com.google.common.collect.Lists;
  * 
  * @since 0.1
  */
-public class LikeExpression extends BaseCompoundExpression {
+public abstract class LikeExpression extends BaseCompoundExpression {
     private static final Logger logger = LoggerFactory.getLogger(LikeExpression.class);
 
     private static final String ZERO_OR_MORE = "\\E.*\\Q";
@@ -195,10 +196,6 @@ public class LikeExpression extends BaseCompoundExpression {
 //        return sb.toString();
 //    }
 
-    public static LikeExpression create(List<Expression> children, LikeType likeType) {
-        return new LikeExpression(addLikeTypeChild(children,likeType));
-    }
-    
     private static final int LIKE_TYPE_INDEX = 2;
     private static final LiteralExpression[] LIKE_TYPE_LITERAL = new LiteralExpression[LikeType.values().length];
     static {
@@ -206,12 +203,12 @@ public class LikeExpression extends BaseCompoundExpression {
             LIKE_TYPE_LITERAL[likeType.ordinal()] = LiteralExpression.newConstant(likeType.name());
         }
     }
-    private Pattern pattern;
+    private AbstractBasePattern pattern;
 
     public LikeExpression() {
     }
 
-    private static List<Expression> addLikeTypeChild(List<Expression> children, LikeType likeType) {
+    protected static List<Expression> addLikeTypeChild(List<Expression> children, LikeType likeType) {
         List<Expression> newChildren = Lists.newArrayListWithExpectedSize(children.size()+1);
         newChildren.addAll(children);
         newChildren.add(LIKE_TYPE_LITERAL[likeType.ordinal()]);
@@ -247,11 +244,14 @@ public class LikeExpression extends BaseCompoundExpression {
         }
     }
 
-    protected Pattern compilePattern (String value) {
-        if (likeType == LikeType.CASE_SENSITIVE)
-            return Pattern.compile(toPattern(value));
-        else
-            return Pattern.compile("(?i)" + toPattern(value));
+    protected abstract AbstractBasePattern compilePatternSpec(String value);
+
+    protected AbstractBasePattern compilePattern(String value) {
+        if (likeType == LikeType.CASE_SENSITIVE) {
+            return compilePatternSpec(toPattern(value));
+        } else {
+            return compilePatternSpec("(?i)" + toPattern(value));
+        }
     }
 
     private Expression getStrExpression() {
@@ -264,36 +264,40 @@ public class LikeExpression extends BaseCompoundExpression {
 
     @Override
     public boolean evaluate(Tuple tuple, ImmutableBytesWritable ptr) {
-        Pattern pattern = this.pattern;
+        AbstractBasePattern pattern = this.pattern;
         if (pattern == null) { // TODO: don't allow? this is going to be slooowwww
             if (!getPatternExpression().evaluate(tuple, ptr)) {
-                if (logger.isDebugEnabled()) {
-                    logger.debug("LIKE is FALSE: pattern is null");
+                if (logger.isTraceEnabled()) {
+                    logger.trace("LIKE is FALSE: pattern is null");
                 }
                 return false;
             }
             String value = (String) PVarchar.INSTANCE.toObject(ptr, getPatternExpression().getSortOrder());
             pattern = compilePattern(value);
-            if (logger.isDebugEnabled()) {
-                logger.debug("LIKE pattern is expression: " + pattern.pattern());
+            if (logger.isTraceEnabled()) {
+                logger.trace("LIKE pattern is expression: " + pattern.pattern());
             }
         }
 
-        if (!getStrExpression().evaluate(tuple, ptr)) {
-            if (logger.isDebugEnabled()) {
-                logger.debug("LIKE is FALSE: child expression is null");
+        Expression strExpression = getStrExpression();
+        SortOrder strSortOrder = strExpression.getSortOrder();
+        PVarchar strDataType = PVarchar.INSTANCE;
+        if (!strExpression.evaluate(tuple, ptr)) {
+            if (logger.isTraceEnabled()) {
+                logger.trace("LIKE is FALSE: child expression is null");
             }
             return false;
         }
-        if (ptr.getLength() == 0) {
-            return true;
-        }
 
-        String value = (String) PVarchar.INSTANCE.toObject(ptr, getStrExpression().getSortOrder());
-        boolean matched = pattern.matcher(value).matches();
-        ptr.set(matched ? PDataType.TRUE_BYTES : PDataType.FALSE_BYTES);
-        if (logger.isDebugEnabled()) {
-            logger.debug("LIKE(value='" + value + "'pattern='" + pattern.pattern() + "' is " + matched);
+        String value = null;
+        if (logger.isTraceEnabled()) {
+            value = (String) strDataType.toObject(ptr, strSortOrder);
+        }
+        strDataType.coerceBytes(ptr, strDataType, strSortOrder, SortOrder.ASC);
+        pattern.matches(ptr, ptr);
+        if (logger.isTraceEnabled()) {
+            boolean matched = ((Boolean) PBoolean.INSTANCE.toObject(ptr)).booleanValue();
+            logger.trace("LIKE(value='" + value + "'pattern='" + pattern.pattern() + "' is " + matched);
         }
         return true;
     }
@@ -348,4 +352,6 @@ public class LikeExpression extends BaseCompoundExpression {
     public String toString() {
         return (children.get(0) + " LIKE " + children.get(1));
     }
+
+    abstract public LikeExpression clone(List<Expression> children);
 }
--- a/phoenix-core/src/main/java/org/apache/phoenix/expression/util/regex/AbstractBasePattern.java
+++ b/phoenix-core/src/main/java/org/apache/phoenix/expression/util/regex/AbstractBasePattern.java
@@ -1,0 +1,33 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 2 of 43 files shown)


phoenix-core/src/main/java/org/apache/phoenix/expression/util/regex/JavaRegexWrapper.java
```diff
@@ -0,0 +1,89 @@
...
+            if (matcherSourceStr == null) {
+                isLastMatcherStringNull = true;
+                return null;
+            }
+            isLastMatcherStringNull = false;
+            return new JavaMatcher(pattern.matcher(matcherSourceStr));
+        }
+
+        @Override
+        public boolean isPatternStringNull() {
```
JamesRTaylor: Do we need this method and isMatcherSourceStrNull here, or can we do this check in the LikeExpression.evaluate method?

phoenix-core/src/main/java/org/apache/phoenix/expression/util/regex/AbstractBasePattern.java
```diff
@@ -0,0 +1,32 @@
...
+ * limitations under the License.
+ */
+package org.apache.phoenix.expression.util.regex;
+
+import org.apache.hadoop.hbase.io.ImmutableBytesWritable;
+import org.apache.phoenix.schema.SortOrder;
+
+public abstract class AbstractBasePattern {
+
+    public abstract AbstractBaseMatcher macher(ImmutableBytesWritable ptr, SortOrder sortOrder);
```
JamesRTaylor: typo here: macher -> matcher

phoenix-core/src/main/java/org/apache/phoenix/expression/LikeExpression.java
```diff
@@ -289,8 +288,11 @@ public boolean evaluate(Tuple tuple, ImmutableBytesWritable ptr) {
             return true;
         }
 
-        String value = (String) PVarchar.INSTANCE.toObject(ptr, getStrExpression().getSortOrder());
-        boolean matched = pattern.matcher(value).matches();
+        String value = null;
+        if (logger.isDebugEnabled()) {
+            value = (String) PVarchar.INSTANCE.toObject(ptr, getStrExpression().getSortOrder());
```
JamesRTaylor: This is just for debugging I'm assuming, as we don't want to turn ptr into a String. Also, we should rename getStrExpression().

phoenix-core/src/main/java/org/apache/phoenix/expression/util/regex/AbstractBasePattern.java
```diff
@@ -0,0 +1,32 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.phoenix.expression.util.regex;
+
+import org.apache.hadoop.hbase.io.ImmutableBytesWritable;
+import org.apache.phoenix.schema.SortOrder;
+
+public abstract class AbstractBasePattern {
```
JamesRTaylor: If possible, can we have only one abstract class, AbstractBaseMatcher? We can always have an abstract method in LikeExpression that gives you back an AbstractBaseMatcher.

phoenix-core/src/main/java/org/apache/phoenix/expression/util/regex/JONIRegexWrapper.java
```diff
@@ -0,0 +1,181 @@
...
+            }
+            cur = cur.EOF(sb);
+            return sb.toString();
+        }
+
+        @Override
+        public AbstractBaseMatcher macher(ImmutableBytesWritable ptr, SortOrder sortOrder) {
+            Preconditions.checkNotNull(ptr);
+            Preconditions.checkNotNull(sortOrder);
+            byte[] matcherSourceBytes = Utils.immutableBytesWritableToBytes(ptr, sortOrder);
```
JamesRTaylor: I hope there's a pattern.matcher(byte[] buf, int offset) method we can use instead below, as this will cause a copy of the underlying byte[]. 

If this is necessary, can you use ByteUtil.copyKeyBytesIfNecessary() as it prevents a copy unless necessary. I think this logic may be able to be pulled out of here and into LikeExpression.evaluate as it should be the same in both cases.

phoenix-core/src/main/java/org/apache/phoenix/expression/LikeExpression.java
shuxiong: 1. The original code has debugging lines, so I try to remains them. We just turn ptr into String when debugs.
2. I don't quite understand what "rename getStrExpression()" means here?

phoenix-core/src/main/java/org/apache/phoenix/expression/util/regex/JavaRegexWrapper.java
shuxiong: 1. isPatternStringNull seems to be useless. I will remove it later.
2. isMatcherSourceStrNull is necessary, Because function matcher of AbstractBasePattern takes ImmutableBytesWritable as arguments, and we don't know if the parameters are null in LikeExpression, until we turn the arguments to Bytes or Strings in the regex engine(the JONI one and the Java one) in BytesBased or StringsBased LikeExpression.

phoenix-core/src/main/java/org/apache/phoenix/expression/util/regex/JONIRegexWrapper.java
shuxiong: All the expressions take ImmutableBytesWritable as parameters and return ImmutableBytesWritable, so I thinks the arguments type for function matcher of AbstractBasePattern should be ImmutableBytesWritable, neither string nor byte.

But We can have a pattern.matcher(byte[] buf, int offset) function in JONI one(bytes based one), for in some cases we might like to use it like this.

phoenix-core/src/main/java/org/apache/phoenix/expression/util/regex/AbstractBasePattern.java
shuxiong: It seems we can have only one abstract class, AbstractBaseMatcher. I will check it in details later.

shuxiong: I thinks AbstractBasePattern is necessary, for the following reasons:

1. Both j.u.regex and JONI has a pattern and match class.  AbstractBasePattern and AbstractBaseMatcher are abstraction of them. It is also quite easy to update existing code, simply replacing corresponding pattern and match to the abstract ones.

2. We might want to compile one pattern, but use this pattern to match multiple strings, and in this case, it would be easy to use.

phoenix-core/src/main/java/org/apache/phoenix/expression/LikeExpression.java
JamesRTaylor: Please lower the logging level to isTraceEnabled().

phoenix-core/src/main/java/org/apache/phoenix/expression/util/regex/AbstractBaseMatcher.java
```diff
@@ -0,0 +1,22 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.phoenix.expression.util.regex;
+
+public abstract class AbstractBaseMatcher {
```
JamesRTaylor: This class adds no value. Remove it and have your match method below return a boolean:

    boolean matcher(ImmutableBytesWritable ptr, SortOrder sortOrder)

phoenix-core/src/main/java/org/apache/phoenix/expression/util/regex/AbstractBasePattern.java
JamesRTaylor: Make this an interface instead of an abstract class.

phoenix-core/src/main/java/org/apache/phoenix/expression/util/regex/JONIRegexWrapper.java
```diff
@@ -0,0 +1,181 @@
...
+            } else {
+                pattern = null;
+            }
+            isLastMatcherStringNull = false;
+        }
+
+        public enum ReplaceQuoteMachine {
+            STOP, ZERO, ONE_BACKSLASH, IN_QUOTE, IN_QUOTE_ONE_BACKSLASH;
+
+            ReplaceQuoteMachine next(StringBuilder sb, char ch) {
```
JamesRTaylor: Why is all this required? Doesn't JONI understand standard regex patterns?

phoenix-core/src/main/java/org/apache/phoenix/expression/util/regex/JONIRegexWrapper.java
JamesRTaylor: Don't turn ptr into byte[] as it's expensive (it makes a copy of the byte[]). Instead do this below:

    new JONIMatcher(pattern.matcher(ptr.get(), ptr.getOffset()), ptr.getLength());

phoenix-core/src/main/java/org/apache/phoenix/expression/util/regex/JavaRegexWrapper.java
JamesRTaylor: If the ImmutableBytesWritable length is zero, the parameter is null. We don't need isMatcherSourceStrNull.

phoenix-core/src/main/java/org/apache/phoenix/expression/util/regex/RegexPatternFactory.java
```diff
@@ -0,0 +1,44 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.phoenix.expression.util.regex;
+
+import com.google.common.base.Preconditions;
+
+public class RegexPatternFactory {
```
JamesRTaylor: This should not be necessary. You'll have dual expressions, one for JONI, one for Java. Within that class you'll know which way to compile the pattern.

phoenix-core/src/main/java/org/apache/phoenix/expression/util/regex/Utils.java
```diff
@@ -0,0 +1,47 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.phoenix.expression.util.regex;
+
+import org.apache.hadoop.hbase.io.ImmutableBytesWritable;
+import org.apache.phoenix.schema.SortOrder;
+import org.apache.phoenix.schema.types.PVarchar;
+
+class Utils {
```
JamesRTaylor: Remove this util class as it's not needed.

phoenix-core/src/main/java/org/apache/phoenix/expression/util/regex/Utils.java
```diff
@@ -0,0 +1,47 @@
...
+
+    public static byte[] immutableBytesWritableToBytes(ImmutableBytesWritable ptr,
```
JamesRTaylor: An ImmutableBytesWritable is a pointer (by offset and length) into potentially a huge byte array. You're assuming that the ImmutableBytesWritable starts with an offset of 0 and you'd potentially be inverting a huge amount of data needlessly.

phoenix-core/src/main/java/org/apache/phoenix/expression/visitor/CloneExpressionVisitor.java
```diff
@@ -107,7 +109,9 @@ public Expression visitLeave(ComparisonExpression node, List<Expression> l) {
 
     @Override
     public Expression visitLeave(LikeExpression node, List<Expression> l) {
-        return Determinism.PER_INVOCATION.compareTo(node.getDeterminism()) > 0 ? node :  new LikeExpression(l);
```
JamesRTaylor: No instanceof checks. Add an abstract clone method to LikeExpression and call that here instead.

    abstract public LikeExpression clone(List<Expression> children);

phoenix-core/src/main/java/org/apache/phoenix/expression/util/regex/AbstractBasePattern.java
JamesRTaylor: I agree, having a Pattern interface (i.e. AbstractBasePattern) is a good idea. AbstractBaseMatcher is not needed, though (please see comments above). Simpler is better.

phoenix-core/src/main/java/org/apache/phoenix/expression/util/regex/Utils.java
JamesRTaylor: The way to invert is like this:

    dataType.coerceBytes(ptr, dataType, sortOrder, SortOrder.ASC);

If the sortOrder is already ascending, it's a noop. Otherwise, it'll create a new byte array, invert only the bytes pointed to by ptr, and set ptr to point to it.

phoenix-core/src/main/java/org/apache/phoenix/expression/util/regex/AbstractBaseMatcher.java
JamesRTaylor: Actually, simplify this by not passing in SortOrder. Instead use the dataType.coerceBytes() call I mentioned already (in your Utils class) to invert before calling matcher. That way the matcher doesn't need to know anything about SortOrder. So just have this:

    boolean matcher(ImmutableBytesWritable ptr);

phoenix-core/src/main/java/org/apache/phoenix/expression/LikeExpression.java
```diff
@@ -289,11 +289,16 @@ public boolean evaluate(Tuple tuple, ImmutableBytesWritable ptr) {
             return true;
         }
 
-        String value = (String) PVarchar.INSTANCE.toObject(ptr, getStrExpression().getSortOrder());
-        boolean matched = pattern.matcher(value).matches();
+        String value = null;
+        if (logger.isTraceEnabled()) {
+            value = (String) PVarchar.INSTANCE.toObject(ptr, getStrExpression().getSortOrder());
+        }
+        PVarchar.INSTANCE.coerceBytes(ptr, PVarchar.INSTANCE, getStrExpression().getSortOrder(),
```
JamesRTaylor: Minor, but add a local variables at the beginning for strExpression, strSortOrder, strDataType, patternExpression, as they're repeated quote often. Also, here, change to:

    strDataType.coerceBytes(ptr, strDataType, strSortOrder);

JamesRTaylor:
This looks very good. Do the tests pass?

shuxiong:
All tests for LikeExpression are passed.

I update the code and have a draft for RegexpReplaceFunction. But I can't find any tests or end-to-end test for ReplaceFunction. I will write some testcases later.

stoty:
PHOENIX-1287 has already been committed.

PR #47, opened 2015-03-17
PHOENIX-1712 Adding INSTR Function

naveenmadhire:
PHOENIX-1712
I've added the INSTR function. These are initial changes.
Please review if anyone has some time.

Thanks.


naveenmadhire:
Adding some more changes related to Test.

gabrielreid:
This code looks pretty good (with the exception of a few code-formatting issues). I think that the test coverage should be extended a bit though, with the following things being added:
* tests using both sort orders (ascending and descending)
* a test where a the input value is null
* a test where the search string isn't found to verify the return value in that situation
* a test where the function is used in a filter (as this ensures that the function is serialized/deserialized)

Specifically around the serialization/deserialization, there will be a no-args constructor needed, as well as an init method to initialize the search string. See the implementation of RegexpSplitFunction for an example of how this is done.

phoenix-core/src/main/java/org/apache/phoenix/expression/function/InstrFunction.java
```diff
@@ -0,0 +1,128 @@
...
+    @Override
+    public boolean evaluate(Tuple tuple, ImmutableBytesWritable ptr) {
+        Expression child = getStringExpression();
+        if (!child.evaluate(tuple, ptr)) {
+            return false;
+        }
+        
+        int position;
+        //Logic for Empty string search
+        if(strToSearch.length()==0){
```
gabrielreid: Minor code formatting issue: should be a space between "if" and opening parenthesis

phoenix-core/src/main/java/org/apache/phoenix/expression/function/InstrFunction.java
```diff
@@ -0,0 +1,128 @@
...
+    public PDataType getDataType() {
+        return PInteger.INSTANCE;
+    }
+
+    @Override
+    public String getName() {
+        return NAME;
+    }
+
+    public String getStrToSearch() {
```
gabrielreid: The getter and setter for strToSearch probably isn't necessary (this value is only used internally) so it can probably removed.

phoenix-core/src/main/java/org/apache/phoenix/expression/function/InstrFunction.java
```diff
@@ -0,0 +1,128 @@
...
+        
+        int textOffset = ptr.getOffset();
+        int textLength = ptr.getLength();
+        int strLength  = strbytes.getLength();
+        byte [] text = ptr.get();
+        byte[] target = new byte[strLength];
+        int position = 0;
+        
+        //Logic to compare Byte arrays
+        while((textLength - textOffset) > 0){
```
gabrielreid: Minor code formatting issue: there should be a space between the "while" and the opening parenthesis

phoenix-core/src/main/java/org/apache/phoenix/expression/function/InstrFunction.java
```diff
@@ -0,0 +1,128 @@
...
+            int nBytes = StringUtil.getBytesInChar(text[textOffset], sortOrder);
+            
+            //Checking for the boundary condition
+            if(textLength - textOffset < strLength )
+                return -1;
+            
+            System.arraycopy(text, textOffset, target, textOffset, strLength);
+            textOffset += nBytes;
+            position++;
+            if(Arrays.equals(strPattern, target))
```
gabrielreid: Missing space after if, and there should be braces around the code block after this if statement

gabrielreid:
A couple of additional test cases which would also be good:
* testing with strings that contains multi-byte characters
* testing with empty (but not null) strings

naveenmadhire:
Thanks Gabriel. I will work on adding some more test cases to this and changes mentioned.

naveenmadhire:
@gabrielreid I added all the changes which you mentioned. Also I created testcases for all types of combinations, including multibytes and empty strings.

gabrielreid:
@naveenmadhire the added test cases look good, but it looks like your most-recent commit also removed the integration test.

The integration test should include a test like before (basically checking that the function works in a real use case), as well as checking that it works when used as a filter, something like:

    select * from table t where instr(t.value, 'test') > 5

naveenmadhire:
I am having some configuration issues when running a end-2-end test cases. I will figure out today and run that.

naveenmadhire:
@gabrielreid I wrote the required test cases in InstrFunctionIT.java. All ASC test cases are working fine. But for DESC I am getting inverted value as output. For expected value 5 the select statement is giving -6. I confirmed that INSTR function is returning the correct value both for DESC and ASC however the output of select is returning different values for DESC. Is there anything else which I need to handle for DESC scenarios. 
Interestingly, all the testcases of InstrFunctionTest are working fine.

gabrielreid:
The tests you've added look good. It looks like the reason that you're getting issues with the DESC tests is because you're returning the SortOrder of the underlying column in the getSortOrder method. Just removing the getSortOrder method completely should resolve this.

It looks like the filter tests also won't work. This will be because the function instance isn't correctly being deserialized/initialized on the server side. Adding in the following method will ensure that the initialization happens correctly:

    @Override
        super.readFields(input);
        init();
    }

I also took a closer look at the actual testing of the string searching functionality. It looks to me like this could be a lot more simple, although maybe I'm missing something. However, wouldn't it work to just do the following in the evaluate method:

    @Override

        if (!getStringExpression().evaluate(tuple, ptr)) {
            return false;
        }

        String sourceStr = (String) PVarchar.INSTANCE.toObject(ptr, getStringExpression().getSortOrder());

        int position = sourceStr.indexOf(strToSearch);
        ptr.set(PInteger.INSTANCE.toBytes(position));
        return true;
    }

Could you see if that works ok for all test cases, etc? It looks like the code in the `searchString` method is doing a lot of unnecessary array copying otherwise.

naveenmadhire:
@gabrielreid I didn't realize we have pre defined method already available in stringUtil. I've modified the code accordingly and all the test cases are working fine.

Do you see any other test cases need to be added?

gabrielreid:
Looks good, and I think that the test cases cover what needs to be covered. 

Could you rebase the patch on the master branch, and squash the commits into a single commit? If you're not familiar with how to do this, just let me know and I'll do it when I commit the code.

naveenmadhire:
@gabrielreid I tried to do this, but messed it up and created a new Pull request 61. I am closing this request.

PR #48, opened 2015-03-18
PHOENIX-1744 Allow Integer, UnsignedInt and UnsignedLong to be Cast to T...

dhacker1341:
...IMESTAMP

### DIFF
--- a/phoenix-core/src/main/java/org/apache/phoenix/schema/types/PUnsignedLong.java
+++ b/phoenix-core/src/main/java/org/apache/phoenix/schema/types/PUnsignedLong.java
@@ -95,6 +95,11 @@ public class PUnsignedLong extends PWholeNumber<Long> {
   }
 
   @Override
+    public boolean isCastableTo(PDataType targetType) {
+      return super.isCastableTo(targetType) || targetType.isCoercibleTo(PTimestamp.INSTANCE);
+    }
+
+  @Override
   public boolean isCoercibleTo(PDataType targetType) {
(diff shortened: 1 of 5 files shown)


phoenix-core/src/main/java/org/apache/phoenix/schema/types/PUnsignedLong.java
```diff
@@ -95,6 +95,11 @@ public Object toObject(byte[] b, int o, int l, PDataType actualType, SortOrder s
   }
 
   @Override
+    public boolean isCastableTo(PDataType targetType) {
```
JamesRTaylor: Was thinking to only make this change so that an UNSIGNED_LONG is castable to a TIMESTAMP. Allowing an INTEGER to be cast to a TIMESTAMP seems weird to me. How about we just start with this change for now and you adjust your tests accordingly?

dhacker1341: The problem with that is the smaller. Numbers in the jira like 22847 wind up being integer and not long so without doing both you have an arbitrarily small number where things just start working so cast(12949 as timestamp) will not work but cast(284484849 as timestamp) will.

JamesRTaylor: You'd never define a column with a type of integer and cast it to a DATE/TIME. The "constant to DATE/TIME/TIMESTAMP" isn't a real use case. Given there's a workaround already even for unsigned timestamp (casting it to a BIGINT first), I'm somewhat reluctant to do anything.

PR #49, opened 2015-03-19
Index qos

twdsilva:


JamesRTaylor:
Somewhere in Indexer surrounding the point at which a batch of mutations is send over the wire, maybe in doPostWithExceptions, we should call IndexQosCompat.addIndexQosForTable(tableName, batchUUID) which would create a key out of tableName together with batchUUID and add it to the concurrent map. Then, when the batch has been sent, it'd call IndexQosCompat.removeIndexQosForTable(tableName, batchUUID), constructing the same key to remove it from the map. The IndexQosRpcControllerFactory would do a get on the concurrent map using only a key with only the tableName (that matches any batchUUID) in order to know it should use the special queue. That way, the map still gets cleaned up, but works when many different callers are writing index updates at the same time (as the tableName + batch UUID would always be unique). This also avoids the corner case of an index being dropped and subsequently a table being created with the same name. The current Config based approach would find that table name for the table and use the high priority queue.

phoenix-core/src/main/java/org/apache/hadoop/hbase/ipc/PhoenixIndexRpcScheduler.java
```diff
@@ -117,4 +120,5 @@ public int getActiveRpcHandlerCount() {
     public void setExecutorForTesting(RpcExecutor executor) {
         this.callExecutor = executor;
     }
+
```
jyates: nit: extra whitespace

phoenix-core/src/main/java/org/apache/phoenix/hbase/index/IndexQosRpcControllerFactory.java
```diff
@@ -60,27 +60,23 @@ public PayloadCarryingRpcController newController(List<CellScannable> cellIterab
...
-        private Configuration conf;
         private int priority;
 
         public IndexQosRpcController(PayloadCarryingRpcController delegate, Configuration conf) {
             super(delegate);
-            this.conf = conf;
             this.priority = PhoenixIndexRpcSchedulerFactory.getMinPriority(conf);
         }
-
+
```
jyates: nit: whitespace

jyates:
At first blush, it seems like we need to a way to age off index tables or the set of tables could grow out of control, especially if people are creating and deleting a lot of index tables with only minimal writes. On the other hand, even if we had 1 megabyte of table names, that would be likely ~ 50,000 table names (20bytes each). So not really a big problem with cleanup, but James' point above:

>  corner case of an index being dropped and subsequently a table being created with the same name

Is still a concern.

Otherwise, lgtm

phoenix-core/src/main/java/org/apache/phoenix/hbase/index/IndexQosRpcControllerFactory.java
```diff
@@ -60,27 +61,22 @@ public PayloadCarryingRpcController newController(List<CellScannable> cellIterab
...
             super(delegate);
-            this.conf = conf;
             this.priority = PhoenixIndexRpcSchedulerFactory.getMinPriority(conf);
         }
-
         @Override
         public void setPriority(final TableName tn) {
             // if its an index table, then we override to the index priority
-            if (isIndexTable(tn)) {
+            if (!tn.isSystemTable() &&  !PhoenixDatabaseMetaData.SYSTEM_TABLE_NAMES.contains(tn.getNameAsString())) {
```
JamesRTaylor: Use !SchemaUtil.isSystemTable(tn.getName()) instead of this.

phoenix-core/src/main/java/org/apache/phoenix/hbase/index/table/CoprocessorHTableFactory.java
```diff
@@ -42,14 +38,9 @@ public CoprocessorHTableFactory(CoprocessorEnvironment e) {
 
     @Override
     public HTableInterface getTable(ImmutableBytesPtr tablename) throws IOException {
-        Configuration conf = e.getConfiguration();
-
-        // make sure we use the index priority writer for our rpcs
-        IndexQosCompat.setPhoenixIndexRpcController(conf);
 
         // make sure we include the index table in the tables we need to track
         String tableName = Bytes.toString(tablename.copyBytesIfNecessary());
```
JamesRTaylor: No need to convert from byte[] -> String and then later back to byte[].

phoenix-core/src/main/java/org/apache/phoenix/hbase/index/IndexQosRpcControllerFactory.java
JamesRTaylor: Actually, is it only SYSTEM.CATALOG that comes through here? How about if we just check for that with Schema.isMetaTable(tn.getName())? It's possible that we'd add an index to a system table down-the-road, so we may not want to be too general with this check.

phoenix-core/src/main/java/org/apache/phoenix/hbase/index/table/CoprocessorHTableFactory.java
```diff
@@ -42,19 +38,14 @@ public CoprocessorHTableFactory(CoprocessorEnvironment e) {
...
-        IndexQosCompat.enableIndexQosForTable(conf, tableName);
 
         if (LOG.isDebugEnabled()) {
             LOG.debug("Creating new HTable: " + tableName);
         }
-        return this.e.getTable(TableName.valueOf(tablename.copyBytesIfNecessary()));
+        return this.e.getTable(TableName.valueOf(tableName));
```
twdsilva: Changed to use tableName

phoenix-core/src/main/java/org/apache/phoenix/hbase/index/IndexQosRpcControllerFactory.java
twdsilva: SYSTEM.CATALOG and SYSTEM.STATS come through this code path.

phoenix-core/src/main/java/org/apache/phoenix/hbase/index/IndexQosRpcControllerFactory.java
```diff
@@ -60,27 +61,22 @@ public PayloadCarryingRpcController newController(List<CellScannable> cellIterab
...
             super(delegate);
-            this.conf = conf;
             this.priority = PhoenixIndexRpcSchedulerFactory.getMinPriority(conf);
         }
-
         @Override
         public void setPriority(final TableName tn) {
             // if its an index table, then we override to the index priority
-            if (isIndexTable(tn)) {
+            if (!tn.isSystemTable() &&  !SchemaUtil.isSystemTable(tn.getName())) {
```
JamesRTaylor: Sorry to flip/flop on you, but I think the way you did it before was better as if we end up adding indexes to system tables, they wouldn't set the priority like we'd want them to.

phoenix-core/src/main/java/org/apache/phoenix/hbase/index/table/CoprocessorHTableFactory.java
JamesRTaylor: Declare tableName as byte[] instead. I don't see any value in that debug log line, but if you do I guess you can leave it. Otherwise, just ditch it.

PR #54, opened 2015-03-24
PHOENIX-1684 Functional Index using REGEXP_SUBSTR doesn't work correctly

twdsilva:


phoenix-core/src/main/java/org/apache/phoenix/parse/IndexExpressionParseNodeRewriter.java
```diff
@@ -51,6 +51,12 @@ public Void visit(ColumnParseNode node) throws SQLException {
             return null;
         }
         
+        @Override
+        public boolean visitEnter(FunctionParseNode node) throws SQLException {
+            isParseNodeCaseSensitive = true;
```
JamesRTaylor: Is there a downside to always surrounding column names with double quotes when we run the IndexExpressionParseNodeRewriter?

phoenix-core/src/main/java/org/apache/phoenix/parse/IndexExpressionParseNodeRewriter.java
```diff
@@ -51,6 +51,12 @@ public Void visit(ColumnParseNode node) throws SQLException {
             return null;
         }
         
+        @Override
+        public boolean visitEnter(FunctionParseNode node) throws SQLException {
```
JamesRTaylor: Is the intent of this visitor to determine if the indexed column is solely a column name that's not case sensitive? Seems like you'd want to switch the starting value in the constructor of isParseNodeCaseSensitive to true and then only set it false in the visit(ColumnParseNode) if isTopLevel() is true (i.e. it's the only expression in the tree) and the node isn't already case sensitive.

phoenix-core/src/main/java/org/apache/phoenix/parse/IndexExpressionParseNodeRewriter.java
twdsilva: I don't think there is a downside, in fact I think this is the correct behavior because when we create the index table we set we use a case sensitive version of the expressions string, so I think we always have to include the double quotes

String name = expressionStr.replaceAll("\"", "'");
                        colName = ColumnName.caseSensitiveColumnName(IndexUtil.getIndexColumnName(null, name));

phoenix-core/src/main/java/org/apache/phoenix/parse/IndexExpressionParseNodeRewriter.java
twdsilva: I have removed ColumnParseNodeVisitor as its not required.

PR #55, opened 2015-03-25
PHOENIX-1457 Use high priority queue for metadata endpoint calls

twdsilva:


phoenix-core/src/main/java/org/apache/phoenix/hbase/index/PhoenixRpcControllerFactory.java
```diff
@@ -26,52 +26,57 @@
...
+            this.indexPriority = PhoenixRpcSchedulerFactory.getIndexMinPriority(conf);
+            this.metadataPriority = PhoenixRpcSchedulerFactory.getMetadataMinPriority(conf);
         }
         @Override
         public void setPriority(final TableName tn) {
-            // if its an index table, then we override to the index priority
-            if (!tn.isSystemTable() &&  !SchemaUtil.isSystemDataTable(tn.getNameAsString())) {
-                setPriority(this.priority);
+            // this is function is called for hbase system tables, phoenix system tables and index tables 
+            if (SchemaUtil.isSystemDataTable(tn.getNameAsString())) {
```
JamesRTaylor: Minor nit: how about moving SchemaUtil.isSystemDataTable() here and calling it useHighPriorityQueue() instead as I don't think it'd be called elsewhere (since it's somewhat random subset of our system tables).

phoenix-core/src/main/java/org/apache/phoenix/hbase/index/ipc/PhoenixRpcSchedulerFactory.java
```diff
@@ -62,30 +62,59 @@ public RpcScheduler create(Configuration conf, RegionServerServices services) {
...
-                + ",  " + maxPriority + ") must be outside HBase priority range ("
-                + HConstants.REPLICATION_QOS + ", " + HConstants.HIGH_QOS + ")");
+        int indexMinPriority = getIndexMinPriority(conf);
+        int indexMaxPriority = conf.getInt(QueryServices.MAX_INDEX_PRIOIRTY_ATTRIB, QueryServicesOptions.DEFAULT_INDEX_MAX_PRIORITY);
+        validatePriority(indexMinPriority, indexMaxPriority);
+        
+        // get the metadata priority configs
+        int metadataHandlerCount = conf.getInt(QueryServices.INDEX_HANDLER_COUNT_ATTRIB, QueryServicesOptions.DEFAULT_METADATA_HANDLER_COUNT);
+        int metadataMinPriority = getMetadataMinPriority(conf);
+        int metadataMaxPriority = conf.getInt(QueryServices.MAX_INDEX_PRIOIRTY_ATTRIB, QueryServicesOptions.DEFAULT_METADATA_MAX_PRIORITY);
```
JamesRTaylor: Do we really need a range? Might be simpler if we just had a single config attribute for "SYSTEM_RPC_PRIORITY".

phoenix-core/src/main/java/org/apache/phoenix/hbase/index/PhoenixRpcControllerFactory.java
twdsilva: Sure I will move and rename.

phoenix-core/src/main/java/org/apache/phoenix/hbase/index/ipc/PhoenixRpcSchedulerFactory.java
twdsilva: I spoke with @jyates, I added the min/max range in case we want to support setting priorities of specific tables in the future, but if this is not required I can remove it and just have a single priority. Should I make the same change for the index priorities?

JamesRTaylor: That'd be my vote. It'd cut down on what you'd need to configure too, right? If we have a default value for the priority, you wouldn't even need to specify this in hbase-sites.xml. If we need more down the road, we can always add it (but I don't think we will).

jyates: I can see us wanting a reserved range for metadata updates - say 10 different numbers - so we don't have to worry about conflicting with anything else that needs its own priority in the future. Its easy to get into situations, especially if everyone is carefully watching these numbers and fully know whats going on, to get into situations where you have META = 1, SOME_FEATURE=2, OTHER_META=3, etc.

A reserved range is just safer for the future and is likely to never be changed in the configuration, unless people already to priority changing, which is unlikely given how hard it was for us :). However, having the configuration is necessary, regardless of range or a single value, to ensure people can make it work with their existing installations (and as proof against HBase invading our range space by accident).

phoenix-core/src/it/java/org/apache/phoenix/end2end/index/PhoenixRpcIT.java
```diff
@@ -0,0 +1,264 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more contributor license agreements. See the NOTICE
```
jyates: Nit: looks like you formatted the header :-/

phoenix-core/src/it/java/org/apache/phoenix/end2end/index/PhoenixRpcIT.java
```diff
@@ -0,0 +1,264 @@
...
+        setUpConfigForMiniCluster(conf);
+        conf.set(HRegionServer.REGION_SERVER_RPC_SCHEDULER_FACTORY_CLASS,
+                TestPhoenixIndexRpcSchedulerFactory.class.getName());
+        conf.set(RpcControllerFactory.CUSTOM_CONTROLLER_CONF_KEY, PhoenixRpcControllerFactory.class.getName());
+        util = new HBaseTestingUtility(conf);
+        // start cluster with 2 region servers
+        util.startMiniCluster(NUM_SLAVES);
+        admin = util.getHBaseAdmin();
+        String clientPort = util.getConfiguration().get(QueryServices.ZOOKEEPER_PORT_ATTRIB);
+        url = JDBC_PROTOCOL + JDBC_PROTOCOL_SEPARATOR + LOCALHOST + JDBC_PROTOCOL_SEPARATOR + clientPort
```
jyates: isn't there a helper method for this somewhere?

phoenix-core/src/it/java/org/apache/phoenix/end2end/index/PhoenixRpcIT.java
```diff
@@ -0,0 +1,264 @@
...
+
+    @After
+    public void tearDown() throws Exception {
+        try {
+            destroyDriver(driver);
+            if (admin!=null) {
+            	admin.close();
+            }
+        } finally {
+            util.shutdownMiniCluster();
```
jyates: Do we need a whole new cluster for every method?

phoenix-core/src/it/java/org/apache/phoenix/end2end/index/PhoenixRpcIT.java
```diff
@@ -0,0 +1,264 @@
...
+    private static final String DATA_TABLE_FULL_NAME = SchemaUtil.getTableName(SCHEMA_NAME, "T");
+    private static final String INDEX_TABLE_FULL_NAME = SchemaUtil.getTableName(SCHEMA_NAME, "I");
+    private static final int NUM_SLAVES = 2;
+
+    private static String url;
+    private static PhoenixTestDriver driver;
+    private HBaseTestingUtility util;
+    private HBaseAdmin admin;
+    private Configuration conf;
+    private static RpcExecutor indexRpcExecutor = Mockito.spy(new BalancedQueueRpcExecutor("test-index-queue", 30, 1, 300));
```
jyates: Does the spys need to be reset for mockito?

phoenix-core/src/it/java/org/apache/phoenix/end2end/index/PhoenixRpcIT.java
```diff
@@ -0,0 +1,264 @@
...
+import org.apache.phoenix.util.ReadOnlyProps;
+import org.apache.phoenix.util.SchemaUtil;
+import org.junit.After;
+import org.junit.Before;
+import org.junit.Test;
+import org.junit.experimental.categories.Category;
+import org.mockito.Mockito;
+
+
+@Category(NeedsOwnMiniClusterTest.class)
```
jyates: This ensures it runs the methods single-threaded, right?

phoenix-core/src/main/java/org/apache/hadoop/hbase/ipc/PhoenixRpcScheduler.java
```diff
@@ -40,28 +40,34 @@
...
-    private RpcExecutor callExecutor;
+    private int indexPriority;
+    private int metadataPriority;
+    private RpcExecutor indexCallExecutor;
+    private RpcExecutor metadataCallExecutor;
     private int port;
 
-    public PhoenixIndexRpcScheduler(int indexHandlerCount, Configuration conf,
-            RpcScheduler delegate, int minPriority, int maxPriority) {
+    public PhoenixRpcScheduler(int indexHandlerCount, int metadataHandlerCount, Configuration conf,
```
jyates: Can we do layered handlers instead of clumping them all into one? HBase did this with its meta, regular, and replication queues and it is terribly ugly.

Instead, maybe have this class instantiate the necessary handler and then chain through them until one of the handlers range (min/max priority) applies to the incoming message and it then passes it onto the queue. You could also do some checking to ensure the ranges don't overlap, etc.

At the very least, this will get rid of the layers of if/else below.

phoenix-core/src/main/java/org/apache/phoenix/hbase/index/ipc/PhoenixRpcSchedulerFactory.java
```diff
@@ -62,30 +62,59 @@ public RpcScheduler create(Configuration conf, RegionServerServices services) {
...
+        validatePriority(indexMinPriority, indexMaxPriority);
+        
+        //validate index and metadata priorities do not overlap
+        Preconditions.checkArgument(doesNotOverlap(indexMinPriority, indexMaxPriority, metadataMinPriority, metadataMaxPriority), "Priority ranges ("
+                + indexMinPriority + ",  " + indexMaxPriority + ") and  (" + metadataMinPriority + ", " + metadataMaxPriority
```
jyates: super nit: Might be cleaner if you just use a format string here

phoenix-core/src/main/java/org/apache/hadoop/hbase/ipc/PhoenixRpcScheduler.java
JamesRTaylor: Would we need separate entries for each RpcScheduler in hbase-sites.xml then?

jyates: No, we can have standard ones and then the ability to overwrite them in configuration. Each scheduler would likely need its own limits config though.

phoenix-core/src/it/java/org/apache/phoenix/end2end/index/PhoenixRpcIT.java
twdsilva: Yes there is getLocalClusterUrl, I will use that.

phoenix-core/src/main/java/org/apache/phoenix/hbase/index/ipc/PhoenixRpcSchedulerFactory.java
JamesRTaylor: The range is implied and documented by having a config for PHOENIX_META priority and config for PHOENIX_INDEX priority.

phoenix-core/src/main/java/org/apache/hadoop/hbase/ipc/PhoenixRpcScheduler.java
twdsilva: I modified the controller factory to use  create a chain of controllers, which was easy I could just have them extend DelegatingPayloadCarryingRpcController. 
However for the PhoenixRpcScheduler needs to implement getGeneralQueueLength and getActiveRpcHandlerCount which need queue lengths and counts of the indexCallExecutor and metadataCallExecutor . 
Do you think it would be easier to have a map from prioirity to queue and then in dispatch() we just lookup the queue based on the priority (and if its not found in the map call  delegate.dispatch(callTask) )?

phoenix-core/src/main/java/org/apache/phoenix/query/ConnectionQueryServicesImpl.java
```diff
@@ -290,7 +290,7 @@ private void openConnection() throws SQLException {
     @Override
     public HTableInterface getTable(byte[] tableName) throws SQLException {
         try {
-            return HBaseFactoryProvider.getHTableFactory().getTable(tableName, connection, null);
+            return HBaseFactoryProvider.getHTableFactory().getTable(tableName, connection, getExecutor());
```
JamesRTaylor: Revert this change please - it was part of our experiment to get the RpcScheduler set on the client-side.

phoenix-core/src/main/java/org/apache/phoenix/query/QueryServices.java
```diff
@@ -123,9 +123,10 @@
     // Index will be partially re-built from index disable time stamp - following overlap time
     public static final String INDEX_FAILURE_HANDLING_REBUILD_OVERLAP_TIME_ATTRIB =
         "phoenix.index.failure.handling.rebuild.overlap.time";
-    public static final String MIN_INDEX_PRIOIRTY_ATTRIB = "phoenix.regionserver.index.priority.min";
-    public static final String MAX_INDEX_PRIOIRTY_ATTRIB = "phoenix.regionserver.index.priority.max";
+    public static final String INDEX_PRIOIRTY_ATTRIB = "phoenix.regionserver.index.priority";
     public static final String INDEX_HANDLER_COUNT_ATTRIB = "phoenix.regionserver.index.handler.count";
+    public static final String METADATA_PRIOIRTY_ATTRIB = "phoenix.regionserver.metadata.priority";
```
JamesRTaylor: This is a client side and region server param, isn't it? How about phoenix.index.rpc.priority and phoenix.metadata.rpc.priority?

phoenix-core/src/main/java/org/apache/hadoop/hbase/ipc/PhoenixRpcScheduler.java
JamesRTaylor: I don't think you need int indexHandlerCount, int metadataHandlerCount anymore, nor do we need the config params. I believe this was based on the range of priorities we were reserving. Both are just 1 now.

jyates: > DelegatingPayloadCarryingRpcController. 

Huh? That's an odd one to pick.

It seemed to me that you wanted to do something that looked like an RpcScheduler (maybe drop all the priority queue, replication queue methods) and change dispatch to look like ```boolean dispatch(CallRunner callTask)``` where boolean indicates handled or not.

> needs to implement getGeneralQueueLength and getActiveRpcHandlerCount

If you do the 'kinda like an RpcScheduler' route then you just loop through all of them and sum the counts. They don't need to be instantaneously accurate as its just being used for metrics, not for anything critical and need to 100% taken at the same point (really impossible if you have more than 1 queue).

> would be easier to have a map from prioirity to queue and then in dispatch() we just lookup the queue based on the priority

Not necessarily, as priority isn't the only thing that determines which queue an RPC should be sent to. For instance, QosPriority annotations can be used to determine which queue an RPC should be sent to as well:
```
  @QosPriority(priority=HConstants.HIGH_QOS)
  public GetRegionInfoResponse getRegionInfo(final RpcController controller,
      final GetRegionInfoRequest request) throws ServiceException {
```

Now that you have implemented something like what I recommended, do you think its cleaner? easier to reason about? I was recommending based on what seemed like an anti-pattern in favor of a pretty reasonable abstraction from unix kernel hooks (and adopted by HBase for coprocessors)...however, I'm known to be a little heavy on generalizations, so this could be overkill to. What do you think?

jyates: @JamesRTaylor 
>  I believe this was based on the range of priorities we were reserving. Both are just 1 now.

Why don't you agree with the future proofing of a range of priorities? I think I put forward a good case for them elsewhere.. maybe I didn't see your comments otherwise?

JamesRTaylor: I don't think declaring a range of priorities future proofs this as no one will know we've reserved that range - folks don't even read docs in general. If if they did happen upon it, there's nothing that'll enforce that they respect it. Better IMO to declare that we're using two explicit priorities. If folks need to change them, it pretty straightforward as it's two documented config parameters.

The current implementation is simpler, removes a number of configs (we already have too many), and meets the needs.

jyates: And if we ever want to add different priorities for different indexes? Or for different metadata tables? Or some other special tables (e.g. higher priority tables as a logical construct). But ok, we can just assign each a number as it comes up... that's never been weird for any other use case ::sarcasm::

> removes a number of configs 
The number needs to be configurable to support installations where people are using priorities for other things that aren't necessarily phoenix (as I mentioned elsewhere), so having an upper and a lower for each isn't really that big of a deal, IMHO.

> If if they did happen upon it, there's nothing that'll enforce that they respect it
Except their cluster becoming unusable.

>  Better IMO to declare that we're using two explicit priorities
Which is again, buried somewhere in the docs. Its fine to say that they claim a range in the same place and then add a link to a 'you can configure it here', but maybe I'm mistaken in what users want?

jyates: I've stated my case - if you still think I'm being overkill @JamesRTaylor then I'm happy to go with your proposal. Just wanted to call out the full implications.

JamesRTaylor: bq. And if we ever want to add different priorities for different indexes? Or for different metadata tables?
I don't anticipate that based on what I've seen our users do and what's talked about on the mailing list. I'd rather keep things simpler for now as it's easier to understand. As new committers come onboard, it's easier to get them up to speed if that's the case.

PR #62, opened 2015-04-01
PHOENIX-1798 change Collections.empytList() to Lists.newArrayList()

StormAll:
This change related to PHOENIX-1798.
If the select SQL query a lot of data, the issue will occur.


stoty:
This has been committed as PHOENIX-1798

PR #63, opened 2015-04-06
PHOENIX-1815 Use Spark Data Source API in phoenix-spark module

jmahonin:
This allows using the SQLContext.load() functionality to create
a Phoenix data frame, which also supports push-down on column
or predicate filtering from Spark SQL.

As well, DataFrame.save() is supported for persisting DataFrames
back to Phoenix.

This may work with Spark's standalone SQL server mode, but it
hasn't been tested.

ref:
https://spark.apache.org/docs/latest/sql-programming-guide.html#data-sources

https://databricks.com/blog/2015/01/09/spark-sql-data-sources-api-unified-data-access-for-the-spark-platform.html


mravi:
Thanks @jmahonin  for the quick turnaround on this.
 I will review this today and get back.

jmahonin:
Merged in 'master' to update with new integration tests

mravi:
@jmahonin 
   One minor change.  Can you please replace the following line in PhoenixRDD.scala and ProductRDDFunctions. This will ensure we load hbase-site.xml so all configuration parameters set for HBase / Phoenix can be applied.
 
---------------------------

val config = new Configuration(conf)  . 
to
val config = HBaseConfiguration.create(conf)

mravi:
@jmahonin 
      Would it be good to have ProductRDDFunctions renamed to  PhoenixRDDFunctions? I believe you had a good reason to name it as that in the first instance. Thoughts?

jmahonin:
Thanks for the review @mravi 

That HBaseConfiguration.create() step is a great idea, I'll make that change ASAP.

Re: naming scheme, I'd attempted to follow Cassandra-Spark connector, since there's not yet too much available for reference code, but also the feature sets would be relatively closely aligned:
https://github.com/datastax/spark-cassandra-connector/tree/master/spark-cassandra-connector/src/main/scala/com/datastax/spark/connector

Although I'm not completely married to the idea, both Datastax (Cassandra) and Databricks (Spark) seem to follow a _Functions.scala scheme, where _ is the class to which implicit helper parameters are being attached. In this case, the new 'ProductRDDFunctions' applies the implicit helper function 'saveToPhoenix' to objects of type RDD[Product], or an RDD of tuples.

mravi:
Thanks @jmahonin  for making the necessary changes .  Below are the final set of tasks that need to be done before we publish documentation.

a) In phoenix/pom.xml , Add the dependency of phoenix-spark module under <dependencyManagement>.  
https://github.com/apache/phoenix/blob/master/pom.xml#L427
b) In phoenix/phoenix-assembly/pom.xml , Add the dependency of phoenix-spark module .    https://github.com/apache/phoenix/blob/master/phoenix-assembly/pom.xml#L148
c) In phoenix/phoenx-assembly/build/all-common-jars.xml , register spark artifacts.
   https://github.com/apache/phoenix/blob/master/phoenix-assembly/src/build/components/all-common-jars.xml#L73
d) Generating HTML  from the README . 

Is it possible for you to create patches for both the master the 4.4.X HBase 0.98 branch and attach to the JIRA? I can take care of the d) task and push to SVN . 

Thanks!!

jmahonin:
Re: step c, I would lean towards not including either the spark or scala library JARs. They are provided by the Spark runtime itself, so I'm not sure it makes sense to bundle them within the phoenix assembly JAR. Does that make sense to you?

ref:
https://spark.apache.org/docs/latest/submitting-applications.html
https://github.com/sbt/sbt-assembly#-provided-configuration

After those, I think the only other runtime dependency not already in the all-common-jars file is one for snappy-java, which I'm not sure is explicitly needed anymore that we're part of a multi-module build. I will double-check.

jmahonin:
I misread your review at first and had thought you asked to update all-common-dependencies, rather than all-common JARS.

a), b) and c) are all addressed in the latest commit here, and final patchfiles are attached to the JIRA ticket.

Thanks!

PR #68, opened 2015-04-15
Phoenix 1759

7shurik:
This is a fix for https://issues.apache.org/jira/browse/PHOENIX-1759. There's a set of unit tests as well as an integration test.

Main code change is adding a PhoenixArray.toString() implementation and toString on each primitive array type. There're also a couple of fixes for toStringLiteral not to barf on nulls.

Please let me know if you see any problems.

### DIFF
--- a/phoenix-core/src/main/java/org/apache/phoenix/schema/types/PDataType.java
+++ b/phoenix-core/src/main/java/org/apache/phoenix/schema/types/PDataType.java
@@ -1095,14 +1095,14 @@ public abstract class PDataType<T> implements DataType<T>, Comparable<PDataType<
     return toStringLiteral(o, formatter);
   }
   
-  public String toStringLiteral(Object o, Format formatter) {
-      if (formatter != null) {
-          return formatter.format(o);
-      } else if(null == o) {
-          return String.valueOf(o);
-      }
-      return o.toString();
-  }
+    public String toStringLiteral(Object o, Format formatter) {
+        if (formatter != null) {
+            return formatter.format(o);
+        } else if (null == o) {
+            return String.valueOf(o);
+        }
+        return o.toString();
+    }
 
   private static final PhoenixArrayFactory DEFAULT_ARRAY_FACTORY = new PhoenixArrayFactory() {
     @Override public PhoenixArray newArray(PDataType type, Object[] elements) {
--- a/phoenix-core/src/main/java/org/apache/phoenix/schema/types/PDataType.java
+++ b/phoenix-core/src/main/java/org/apache/phoenix/schema/types/PDataType.java
@@ -1104,6 +1104,11 @@ public abstract class PDataType<T> implements DataType<T>, Comparable<PDataType<
         return o.toString();
     }
 
+    public String toStringLiteral(Object o) {
+        // use default formatter when one is unspecified
+        return toStringLiteral(o, null);
+    }
+
   private static final PhoenixArrayFactory DEFAULT_ARRAY_FACTORY = new PhoenixArrayFactory() {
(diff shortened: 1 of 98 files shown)


phoenix-core/src/main/java/org/apache/phoenix/schema/types/PDataType.java
```diff
@@ -1098,8 +1098,10 @@ public String toStringLiteral(byte[] b, int offset, int length, Format formatter
   public String toStringLiteral(Object o, Format formatter) {
       if (formatter != null) {
           return formatter.format(o);
-        }
-        return o.toString();
+      } else if(null == o) {
```
samarthjain: Minor nit: space between if and (

7shurik: I'll fix it.

phoenix-core/src/main/java/org/apache/phoenix/schema/types/PhoenixArray.java
```diff
@@ -561,6 +666,23 @@ public int estimateByteSize(int pos) {
...
+			StringBuilder sb = new StringBuilder(TO_STRING_BEGIN);
+			boolean isFirst = true;
+			final Format defaultFormat = null;
+			for (Byte o : byteArr) {
+				if (isFirst) {
+					isFirst = false;
+				} else {
+					sb.append(TO_STRING_SEPARATOR);
+				}
+				sb.append(this.baseType.toStringLiteral(o, defaultFormat));
```
samarthjain: Looks like the defaultFormat is always null. How about having a new method toStringLiteral(o) which delegates to toStringLiteral(o, defaultFormat) where defaultFormat is null?

samarthjain: Also it looks like toStringLiteral() is taking Object. So is it possible to just use the toString() method defined in the super class PhoenixArray. This way you won't have to duplicate code for toString() method for every array type.

7shurik: Good suggestion on toStringLiteral().

On the second comment.. I started to use PhoenixArray.toString() but that doesn't work. Arrays in primitive array subclasses contain primitives and there's an exception if 
for(Object o : arr) { ... } is used. If arr consists of primitives (e.g. ints), then for(Integer o : ints) works because type boxing/unboxing automatically does the right cast.

I guess another way would be to have a method that converts from primitive to object. Eg. int to Integer which would make toString generic.

JamesRTaylor: Use these methods instead of relying on unboxing:

    PhoenixArray a;
    for (int i = 0; i < a.getDimensions(); i++) {
      Object o = a.getElement(i);
    }

7shurik: Thank you, @JamesRTaylor! That's the silver bullet I was looking for. I hate duplicating the code like it did.

7shurik:
@samarthjain @JamesRTaylor are you still reviewing? I've uploaded the version with suggested changes, including a single implementation of PhoenixArray.toString().
Serhiy Bilousov on the JIRA said he might review over the weekend as well.

samarthjain:
+1, LGTM.

7shurik:
I'm fixing the Apache license header in PrimitiveIntPhoenixArrayToStringTest.java.

stoty:
The final version of this patch is a license header format update.
That license header has since been reformatted anyway.

PR #70, opened 2015-04-17
JIRA Phoenix 1872

7shurik:
I'm fixing https://issues.apache.org/jira/browse/PHOENIX-1872. The fix is to use PrimitiveBooleanArray of boolean[] instead of PhoenixArray of Boolean[]. 

I considered using BitSet and not sure we'd get the benefit vs. using boolean[]. Also the code is much easier to maintain.

I've ran and fixed unit tests and appropriate integration tests as well.


JamesRTaylor:
@twdsilva  - would you mind reviewing this?

stoty:
This (or a later version) has already been merged.

PR #74, opened 2015-04-23
Phoenix 331- Phoenix-Hive initial commit 

nmaillard:


phoenix-hive/src/main/java/org/apache/phoenix/hive/HivePhoenixInputFormat.java
```diff
@@ -0,0 +1,161 @@
...
+
+import com.google.common.base.Preconditions;
+import com.google.common.collect.Lists;
+
+
+/**
+* HivePhoenixInputFormat
+* Need to extend the standard PhoenixInputFormat but also implement the mapred inputFormat for Hive compliance
+*
+* @version 1.0
```
ddraj: Please remove these tags

phoenix-hive/src/main/java/org/apache/phoenix/hive/HivePhoenixInputSplit.java
```diff
@@ -0,0 +1,111 @@
...
+    }
+
+    public int hashCode() {
+        int prime = 31;
+        int result = 1;
+        result = 31 * result + (this.keyRange == null ? 0 : this.keyRange.hashCode());
+        return result;
+    }
+
+    public boolean equals(Object obj) {
```
ddraj: Shouldn't the equals and hashCode take into account the 'path' field?

phoenix-hive/src/main/java/org/apache/phoenix/hive/HivePhoenixRecordReader.java
```diff
@@ -0,0 +1,182 @@
...
+        super(inputClass,configuration,queryPlan);
+        Preconditions.checkNotNull(configuration);
+        Preconditions.checkNotNull(queryPlan);
+        this.inputClass = inputClass;
+        this.configuration = configuration;
+        this.queryPlan = queryPlan;
+    }
+
+
+   /* public NullWritable getCurrentKey() throws IOException, InterruptedException {
```
ddraj: Please remove all commented out code

phoenix-hive/src/main/java/org/apache/phoenix/hive/HivePhoenixInputFormat.java
```diff
@@ -0,0 +1,161 @@
...
+* @since   2015-02-08 
+*/
+
+public class HivePhoenixInputFormat<T extends DBWritable> extends org.apache.phoenix.mapreduce.PhoenixInputFormat<T> implements org.apache.hadoop.mapred.InputFormat<NullWritable, T>{
```
mravi: How about the idea we have an abstract BasePhoenixInputFormat that provides the implementation for both the old and the new mapreduce API and provide an abstract method to convert the KeyRange to a PhoenixInputSplit or HivePhoenixInputSplit.
On the other hand, if we wanted to completely avoid support for mapred api, we can push the code in getQueryPlan() and part of generateSplits into a Util class that way we avoid a lot of code duplication.  

@ddraj , @JamesRTaylor  What is your recommendation?

phoenix-hive/src/main/java/org/apache/phoenix/hive/HivePhoenixOutputCommitter.java
```diff
@@ -0,0 +1,65 @@
...
+* HivePhoenixOutputCommitter
+*
+* @version 1.0
+* @since   2015-02-08 
+*/
+
+public class HivePhoenixOutputCommitter extends PhoenixOutputCommitter{
+    public final Log LOG = LogFactory.getLog(HivePhoenixOutputCommitter.class);
+    
+    public void commit(Connection connection)
```
mravi: Are we using this class anywhere? I don't see any reference.

phoenix-hive/src/main/java/org/apache/phoenix/hive/HivePhoenixOutputFormat.java
```diff
@@ -0,0 +1,81 @@
...
+* HivePhoenixOutputFormat
+* Need to extend the standard PhoenixOutputFormat but also implement the mapred OutputFormat for Hive compliance
+*
+* @version 1.0
+* @since   2015-02-08 
+*/
+
+
+public class HivePhoenixOutputFormat<T extends DBWritable> extends org.apache.phoenix.mapreduce.PhoenixOutputFormat<T> implements
+org.apache.hadoop.mapred.OutputFormat<NullWritable, T> {
```
mravi: We can avoid extending PhoenixOutputFormat here as it isn't necessary.

phoenix-hive/src/main/java/org/apache/phoenix/hive/HivePhoenixInputFormat.java
JamesRTaylor: +1 to avoid code duplication. We can also have the code be different in our different branches if that helps.

phoenix-hive/src/main/java/org/apache/phoenix/hive/HivePhoenixOutputFormat.java
```diff
@@ -0,0 +1,81 @@
...
+            throw new IOException(e);
+        }
+    }
+
+    public void checkOutputSpecs(FileSystem ignored, JobConf job) throws IOException {
+        LOG.debug("checkOutputSpecs");
+        
+    }
+    
+    synchronized Connection getConnection(Configuration configuration) throws IOException {
```
mravi: Lets push this to HivePhoenixRecordWriter so we can initialize a connection and close it in that class itself.

phoenix-hive/src/main/java/org/apache/phoenix/hive/HivePhoenixRecordReader.java
```diff
@@ -0,0 +1,182 @@
...
+/**
+* HivePhoenixRecordReader
+* implements mapred as well for hive needs
+*
+* @version 1.0
+* @since   2015-02-08 
+*/
+
+
+public class HivePhoenixRecordReader<T extends DBWritable> extends PhoenixRecordReader<T> implements
```
mravi: Can you please share your thoughts on why we need to extend PhoenixRecordReader here?

phoenix-hive/src/main/java/org/apache/phoenix/hive/HivePhoenixRecordWriter.java
```diff
@@ -0,0 +1,85 @@
...
+    public void write(NullWritable n, T record) throws IOException {
+        try {
+            record.write(this.statement);
+            this.numRecords += 1L;
+            this.statement.addBatch();
+
+            if (this.numRecords % this.batchSize == 0L) {
+                LOG.info("log commit called on a batch of size : " + this.batchSize);
+                this.statement.executeBatch();
+                //this.conn.commit();
```
mravi: Can you please uncomment this line to ensure we commit each batch

phoenix-hive/src/main/java/org/apache/phoenix/hive/PhoenixMetaHook.java
```diff
@@ -0,0 +1,210 @@
...
+  static Log LOG = LogFactory.getLog(PhoenixMetaHook.class.getName());
+
+  /**
+   *commitCreateTable creates a Phoenix table after the hive table has been created 
+   * incoming hive types.
+   * @param tbl the table properties
+   * 
+   */
+  //Too much logic in this function must revisit and dispatch
+  public void commitCreateTable(Table tbl)
```
mravi: Should we support creating Phoenix tables from Hive? My vote would be to restrict the code to do validations against the existence of phoenix table.

phoenix-hive/src/main/java/org/apache/phoenix/hive/util/HiveConfigurationUtil.java
```diff
@@ -0,0 +1,275 @@
...
+     * is BINARY and makes a call to {@link #castBytes(Object, PDataType)} to convert bytes to
+     * targetPhoenixType
+     * @param o
+     * @param targetPhoenixType
+     * @return Object
+     */
+    public static Object castHiveTypeToPhoenix(Object o, byte objectType,
+            PDataType targetPhoenixType) {
+        /*
+         * PDataType inferredPType = getType(o, objectType); if (inferredPType == null) { return
```
mravi: please remove the comments

phoenix-hive/src/main/java/org/apache/phoenix/hive/util/HiveConnectionUtil.java
```diff
@@ -0,0 +1,161 @@
...
+import org.apache.phoenix.hive.util.HiveConfigurationUtil;
+
+import com.google.common.base.Preconditions;
+
+/**
+ * Describe your class here.
+ *
+ * @since 138
+ */
+public final class HiveConnectionUtil {
```
mravi: We can avoid this class by reusing the functionality in ConnectionUtil to get connections.

phoenix-hive/src/main/java/org/apache/phoenix/hive/util/PhoenixHiveConfiguration.java
```diff
@@ -0,0 +1,106 @@
...
+import org.apache.phoenix.mapreduce.util.PhoenixConfigurationUtil;
+import org.apache.phoenix.util.QueryUtil;
+
+import com.google.common.base.Preconditions;
+
+/**
+ * Created by nmaillard on 6/23/14.
+ */
+public class PhoenixHiveConfiguration {
+    private static final Log LOG = LogFactory.getLog(PhoenixHiveConfiguration.class);
```
mravi: I don't see any reference to this class in the code.  Are we using it ?

mravi:
Thanks a lot for the patch @nmaillard . Couple of additional minor fixes 
1) Couple of classes have your name as the author. Please remove it.
2) Certain classes have lot of commented code .

phoenix-hive/src/main/java/org/apache/phoenix/hive/HivePhoenixRecordWriter.java
```diff
@@ -0,0 +1,85 @@
...
+    public HivePhoenixRecordWriter(Connection conn, Configuration config) throws SQLException {
+        this.conn = conn;
+        this.batchSize = PhoenixConfigurationUtil.getBatchSize(config);
+        String upsertQuery = PhoenixConfigurationUtil.getUpsertStatement(config);
+        this.statement = this.conn.prepareStatement(upsertQuery);
+    }
+
+    public void write(NullWritable n, T record) throws IOException {
+        try {
+            record.write(this.statement);
```
ddraj: Who does the datatype conversion from Hive to Phoenix?

phoenix-hive/src/main/java/org/apache/phoenix/hive/HivePhoenixRecordWriter.java
ddraj: Did you mean to do this or not? What happens if tasks fail? The data in the table is partially written I guess, right?

nmaillard: You are correct did not mean to do this, uncommenting to make sure we commit each batch

phoenix-hive/src/main/java/org/apache/phoenix/hive/PhoenixMetaHook.java
nmaillard: Actually the idea came from the use of the ES connector; I thought it made life at lot easier when creating the 80% of simple tables. Otherwise you need to go to the 2 systems create each table, nothing hard just adds work and no real value. In a second iteration I wanted to add some more logic here and extract table name from Hive table name, extract PK from the first column or from partition columns.

phoenix-hive/src/main/java/org/apache/phoenix/hive/HivePhoenixRecordWriter.java
nmaillard: The type conversion is all done in  HiveTypeUtil, HiveType2PdataType does Hive to phoenix.
In PhoenixHiveDBWritable the write class leverage that information to create the prepared statement. T extends DBWritable and is a PhoenixHiveDBWritable in a this instance.
Some complex types like Array are still missing, this will be follow up work

phoenix-hive/src/main/java/org/apache/phoenix/hive/HivePhoenixInputSplit.java
nmaillard: The Path part is a workaround because Hive in the storage handler part will expect everything to be HDFS file based

phoenix-hive/src/main/java/org/apache/phoenix/hive/HivePhoenixInputFormat.java
nmaillard: +1 to avoid code duplication as well. I do believe this is more a Hive issue and should be fixed there and not use the mapred API anymore an we could get rid of this whole class, and a couple others. This might be a longer battle so open to put this code where seems to be the best place.

phoenix-hive/src/main/java/org/apache/phoenix/hive/HivePhoenixOutputCommitter.java
nmaillard: You are correct this is a relica of the first implementation.

phoenix-hive/src/main/java/org/apache/phoenix/hive/HivePhoenixRecordReader.java
nmaillard: I could extend RecordReader directly, the idea here it is to reuse as much a possible and only add hive specific parst

phoenix-hive/src/main/java/org/apache/phoenix/hive/util/HiveConnectionUtil.java
nmaillard: Yes very true, however currently the getConnections does not allow to get another znode, always using hbase, in my test I hd base-unsecure. This being said I should maybe fix that instead of extending

phoenix-hive/src/main/java/org/apache/phoenix/hive/PhoenixSerde.java
```diff
@@ -0,0 +1,169 @@
...
+            StructField structField = fields.get(i);
+            if (structField != null) {
+                Object field = structInspector.getStructFieldData(row, structField);
+                ObjectInspector fieldOI = structField.getFieldObjectInspector();
+                switch (fieldOI.getCategory()) {
+                case PRIMITIVE:
+                    Writable value =
+                            (Writable) ((PrimitiveObjectInspector) fieldOI)
+                                    .getPrimitiveWritableObject(field);
+                    phrecord.add(value);
```
JamesRTaylor: Will this write the data in a Phoenix serialized format? This is for data coming in from Hive?

phoenix-hive/src/main/java/org/apache/phoenix/hive/util/HiveConfigurationUtil.java
```diff
@@ -0,0 +1,275 @@
...
+            // hiveTypeToPDataType(((PrimitiveObjectInspector)hiveType).getPrimitiveCategory());
+            return hiveTypeToPDataType(hiveType.getTypeName());
+        default:
+
+            throw new SerDeException("Unsupported column type: " + hiveType.getCategory().name());
+        }
+    }
+
+    public static PDataType hiveTypeToPDataType(String hiveType) throws SerDeException {
+        final String lctype = hiveType.toLowerCase();
```
JamesRTaylor: Switch statement? Does Hive have constants for these?

phoenix-hive/src/main/java/org/apache/phoenix/hive/util/HiveTypeUtil.java
```diff
@@ -0,0 +1,151 @@
...
+
+    /**
+     * This method returns the most appropriate PDataType associated with the incoming hive type
+     * name.
+     * @param hiveType
+     * @return PDataType
+     */
+    public static PDataType HiveType2PDataType(String hiveType) throws SerDeException {
+        final String lctype = hiveType.toLowerCase();
+        if ("string".equals(lctype)) {
```
JamesRTaylor: Looks like the same as the earlier one?

phoenix-hive/src/main/java/org/apache/phoenix/hive/PhoenixSerde.java
nmaillard: this is data coming from hive and writes out the writable value to the PhoenixRecordObject. This object will serialize it to a Phoenix Format at write time in the write method

phoenix-hive/src/main/java/org/apache/phoenix/hive/util/HiveTypeUtil.java
nmaillard: yes, correct

phoenix-hive/src/main/java/org/apache/phoenix/hive/HivePhoenixRecordReader.java
```diff
@@ -0,0 +1,148 @@
...
+        this.inputClass = inputClass;
+        this.configuration = configuration;
+        this.queryPlan = queryPlan;
+    }
+
+    public float getProgress() {
+        return 0.0F;
+    }
+
+    public void init(InputSplit split) throws IOException,
```
JamesRTaylor: Why is this code repeated, as it's in the base MR integration already?

phoenix-hive/src/main/java/org/apache/phoenix/hive/util/HiveConnectionUtil.java
```diff
@@ -0,0 +1,161 @@
...
+                String autocommit = configuration.get(HiveConfigurationUtil.AUTOCOMMIT);
+                if (autocommit != null && autocommit.equalsIgnoreCase("true")) {
+                    conn.setAutoCommit(true);
+                } else {
+                    conn.setAutoCommit(false);
+                }
+                connection = conn;
+            } catch (ClassNotFoundException e) {
+                // TODO Auto-generated catch block
+                e.printStackTrace();
```
JamesRTaylor: Throw here - don't swallow this.

phoenix-hive/src/main/java/org/apache/phoenix/hive/PhoenixStorageHandler.java
```diff
@@ -0,0 +1,147 @@
...
+     * parameter tableDesc tabledescription, jobProperties
+     * TODO this avoids any pushdown must revisit
+     */
+    private void configureJobProperties(TableDesc tableDesc, Map<String, String> jobProperties)
+    {
+      Properties tblProps = tableDesc.getProperties();
+      tblProps.getProperty("phoenix.hbase.table.name");
+      HiveConfigurationUtil.setProperties(tblProps, jobProperties);
+
+      //TODO this avoids any pushdown must revisit and extract meaningful parts
```
JamesRTaylor: Can you elaborate on this? What is this code trying to do?

phoenix-hive/src/main/java/org/apache/phoenix/hive/util/HiveConnectionUtil.java
```diff
@@ -0,0 +1,161 @@
...
+            String autocommit = TblParams.get(HiveConfigurationUtil.AUTOCOMMIT);
+            if(autocommit!=null && autocommit.equalsIgnoreCase("true")){
+                conn.setAutoCommit(true);
+            }else{
+                conn.setAutoCommit(false);
+            }
+            
+            return conn;
+        } catch (ClassNotFoundException e) {
+            e.printStackTrace();
```
JamesRTaylor: Here too. Please make sure there aren't other occurrences of swallowing exceptions. Just throw a SQLException that wraps the ClassNotFoundException.

phoenix-hive/src/main/java/org/apache/phoenix/hive/util/PhoenixUtil.java
```diff
@@ -0,0 +1,163 @@
...
+            query.append(" CONSTRAINT pk PRIMARY KEY (");
+            Joiner joiner = Joiner.on(" , ");
+            query.append(" " + joiner.join(pks) + " )");
+        }
+        query.append(" )");
+        if (salt_buckets > 0) {
+            query.append(" SALT_BUCKETS = " + salt_buckets);
+        }
+        if (compression != null) {
+            query.append(" ,COMPRESSION='GZ'");
```
JamesRTaylor: Is there a reason that GZ is the default compression? Why not just let Phoenix default it? Same for VERSIONS below.

phoenix-hive/src/main/java/org/apache/phoenix/hive/util/PhoenixUtil.java
```diff
@@ -0,0 +1,163 @@
...
+        }
+        if (versions_num > 0) {
+            query.append(" ,VERSIONS="+versions_num);
+        }
+        System.out.println("CREATED QUERY " +query.toString());
```
JamesRTaylor: Remove all System.out.println calls and replace with logging instead.

phoenix-hive/src/main/java/org/apache/phoenix/hive/util/PhoenixUtil.java
```diff
@@ -0,0 +1,163 @@
...
+            LOG.info("found the table " + rs.getString("TABLE_NAME"));
+            while (rs.next()) {
+                LOG.info("found the table " + rs.getString("TABLE_NAME"));
+            }
+            return true;
+        }
+        return false;
+    }
+
+    public static boolean testTable(Connection conn, String name, Map<String, String> fields)
```
JamesRTaylor: Move out any test-specific methods to test util classes instead of having them in runtime classes.

JamesRTaylor:
This is looking good, @nmaillard. Still some copy/paste code that we should aim to cleanup. Also, take care that no exceptions are swallowed. Wrap checked exceptions in a SQLException if it's declared in the  method and otherwise wrap in a RuntimeException and throw it. Also, make sure to remove or convert all System.out.println calls to logging instead.

JamesRTaylor:
@mravi  - would you mind giving this a review?

phoenix-hive/src/main/java/org/apache/phoenix/hive/util/HiveConnectionUtil.java
mravi: If you override HiveStorageHandler#configureJobConf in PhoenixStorageHandler and provide a way to add HBase properties to the configuration, I believe we can avoid this class and leverage ConnectionUtil

phoenix-hive/src/main/java/org/apache/phoenix/hive/HivePhoenixRecordReader.java
mravi: I agree. But, I don't see any code of PhoenixRecordReader used in this class except for the call to super  in the constructor.  Am I missing anything.

phoenix-hive/src/main/java/org/apache/phoenix/hive/PhoenixStorageHandler.java
```diff
@@ -0,0 +1,147 @@
...
+      jobProperties.put("phoenix.select.stmt", "select * from " + (String)jobProperties.get("phoenix.hbase.table.name"));
+      if(((String)jobProperties.get("phoenix.hbase.table.name")).contains("limit")==true){
+          
+      }
+      LOG.debug("ConfigurationUtil.SELECT_STATEMENT " + (String)jobProperties.get("phoenix.select.stmt"));
+    }
+    
+    /**
```
mravi: Can you provide an implementation of HiveStorageHandler#configureJobConf and have HBaseConfiguration passed to the JobConf. This way, the InputFormat classes and RecordReaders have the necessary HBase configurations.

phoenix-hive/src/main/java/org/apache/phoenix/hive/util/HiveConfigurationUtil.java
```diff
@@ -0,0 +1,111 @@
...
+    public static final String SALT_BUCKETS = "saltbuckets";
+    public static final String COMPRESSION = "compression";
+    public static final String VERSIONS = "versions";
+    public static final int VERSIONS_NUM = 5;
+    public static final String SPLIT = "split";
+    public static final String REDUCE_SPECULATIVE_EXEC =
+            "mapred.reduce.tasks.speculative.execution";
+    public static final String MAP_SPECULATIVE_EXEC = "mapred.map.tasks.speculative.execution";
+
+    public static void setProperties(Properties tblProps, Map<String, String> jobProperties) {
```
mravi: Since this method is being called from PhoenixHiveStorageHandler, can you 
a) Ensure the HBase configuration is loaded in PhoenixHiveStorageHandler#setConf.
   this.con = HBaseConfiguration.create(conf);
b) Pass the configuration object to the method setProperties() from PhoenixHiveStorageHandler#configureJobProperties .   

That way, we avoid setting default values.

phoenix-hive/src/main/java/org/apache/phoenix/hive/util/PhoenixHiveConfiguration.java
```diff
@@ -0,0 +1,103 @@
...
+            String tableName, String dbname, String columns) {
+        Quorum = quorum != null ? quorum : HiveConfigurationUtil.ZOOKEEPER_QUORUM_DEFAULT;
+        Port = port != null ? port : HiveConfigurationUtil.ZOOKEEPER_PORT_DEFAULT;
+        Parent = parent != null ? parent : HiveConfigurationUtil.ZOOKEEPER_PARENT_DEFAULT;
+        // BatchSize = batchSize!=null?batchSize:ConfigurationUtil.DEFAULT_UPSERT_BATCH_SIZE;
+    }
+
+    static class PhoenixHiveConfigurationUtil {
+
+        public Connection getConnection(final Configuration configuration) throws SQLException {
```
mravi: Can we remove this. I don't see any caller to this method.

phoenix-hive/src/main/java/org/apache/phoenix/hive/util/PhoenixHiveConfiguration.java
```diff
@@ -0,0 +1,103 @@
...
+                        .get(HiveConfigurationUtil.TABLE_NAME) : tbl.getTableName();
+
+        String mapping = mps.get(HiveConfigurationUtil.COLUMN_MAPPING);
+    }
+
+    public void configure(String server, String tableName, long batchSize) {
+        // configure(server, tableName, batchSize, null);
+    }
+
+    public void configure(String quorum, String port, String parent, long batchSize,
```
mravi: Are we using this configure method from any code ?

nmaillard:
Have made the changes per the first feedback, let me know if there is anything else

ddraj:
Ping @JamesRTaylor  @mravi .. Any more inputs for @nmaillard ? Thanks!

JamesRTaylor:
@ddraj - I left feedback and it was never responded to or implemented.

nmaillard:
thanks everyone, sorry I have had little connectivity the last couple of days. I will adress new feedback right away

stoty:
Already merged.
(And since has been factored out to the phoenix-connectors repo)

PR #75, opened 2015-04-23
Support native JSON data type and json_extract_path function PHOENIX-174...

AakashPradeep:
...3 and PHOENIX-1710


AakashPradeep:
@JamesRTaylor  Can you please review Phoenix Json.

phoenix-core/src/it/java/org/apache/phoenix/end2end/PhoenixJsonE2ETest.java
```diff
@@ -0,0 +1,246 @@
...
+import org.apache.phoenix.schema.json.PhoenixJson;
+import org.apache.phoenix.schema.types.PJsonDataType;
+import org.apache.phoenix.util.PropertiesUtil;
+import org.junit.Test;
+
+/**
+ * End to end test for JSON data type for {@link PJsonDataType} and
+ * {@link PhoenixJson}.
+ */
+public class PhoenixJsonE2ETest extends BaseHBaseManagedTimeIT {
```
JamesRTaylor: Rename PhoenixJsonIT instead so it gets run with mvn verify

phoenix-core/src/main/java/org/apache/phoenix/expression/function/JsonExtractPathFunction.java
```diff
@@ -0,0 +1,110 @@
...
+
+    @Override
+    public boolean evaluate(Tuple tuple, ImmutableBytesWritable ptr) {
+
+        Expression jsonExpression = this.children.get(0);
+        if (!jsonExpression.evaluate(tuple, ptr)) {
+            return false;
+        }
+        if (ptr.getLength() == 0) {
+            return false;
```
JamesRTaylor: Return true here. Only return false if the child expression evaluation returns false.

phoenix-core/src/main/java/org/apache/phoenix/expression/function/JsonExtractPathFunction.java
```diff
@@ -0,0 +1,110 @@
...
+                (PhoenixJson) PJsonDataType.INSTANCE.toObject(ptr.get(), ptr.getOffset(),
+                    ptr.getLength());
+
+        Expression jsonPathArrayExpression = children.get(1);
+        if (!jsonPathArrayExpression.evaluate(tuple, ptr)) {
+            return false;
+        }
+
+        if (ptr.getLength() == 0) {
+            return false;
```
JamesRTaylor: Same here: return true

phoenix-core/src/main/java/org/apache/phoenix/expression/function/JsonExtractPathFunction.java
```diff
@@ -0,0 +1,110 @@
...
+    }
+
+    @Override
+    public String getName() {
+        return NAME;
+    }
+
+    @Override
+    public boolean isNullable() {
+        return PJsonDataType.INSTANCE.isNullable();
```
JamesRTaylor: Should be based on the children being nullable. Ask yourself, when can this built-in function return null? If either the first or the second child is nullable (based on the evaluate impl). So maybe this instead:

    return this.children.get(0).isNullable() || this.children.get(0).isNullable()

phoenix-core/src/main/java/org/apache/phoenix/expression/function/JsonExtractPathFunction.java
```diff
@@ -0,0 +1,110 @@
...
+    }
+
+    @Override
```
JamesRTaylor: Remove these implementations, as this looks like the default impl.

phoenix-core/src/main/java/org/apache/phoenix/expression/function/JsonExtractPathTextFunction.java
```diff
@@ -0,0 +1,108 @@
...
+            PhoenixJson phoenixJson2 = phoenixJson.getNullablePhoenixJson(jsonPaths);
+            if (phoenixJson2 == null) {
+                ptr.set(ByteUtil.EMPTY_BYTE_ARRAY);
+            } else {
+                byte[] json = PVarchar.INSTANCE.toBytes(phoenixJson2.serializeToString());
+                ptr.set(json);
+            }
+
+        } catch (SQLException sqe) {
+            new IllegalDataException(new SQLExceptionInfo.Builder(SQLExceptionCode.ILLEGAL_DATA)
```
JamesRTaylor: You can just do a throw here with the sqe in the constructor and it'll get un-nested for you (don't forget to throw it, though! :-) ):
    throw new IllegalDataException(sqe);

phoenix-core/src/main/java/org/apache/phoenix/expression/function/JsonExtractPathTextFunction.java
```diff
@@ -0,0 +1,108 @@
...
+import org.apache.phoenix.schema.types.PJsonDataType;
+import org.apache.phoenix.schema.types.PVarchar;
+import org.apache.phoenix.schema.types.PVarcharArray;
+import org.apache.phoenix.schema.types.PhoenixArray;
+import org.apache.phoenix.util.ByteUtil;
+
+@BuiltInFunction(name = JsonExtractPathTextFunction.NAME, args = {
+        @Argument(allowedTypes = { PJsonDataType.class }),
+        @Argument(allowedTypes = { PVarcharArray.class }) })
+public class JsonExtractPathTextFunction extends ScalarFunction {
```
JamesRTaylor: Same feedback as with previous built-in

phoenix-core/src/main/java/org/apache/phoenix/schema/json/PhoenixJson.java
```diff
@@ -0,0 +1,216 @@
...
+     * should be wrapped in {@link SQLException} by caller.
+     * @param data Buffer that contains data to parse
+     * @param offset Offset of the first data byte within buffer
+     * @param length Length of contents to parse within buffer
+     * @return {@link PhoenixJson}.
+     * @throws JsonParseException
+     * @throws IOException
+     */
+    public static PhoenixJson getPhoenixJson(byte[] jsonData, int offset, int length)
+            throws JsonParseException, IOException {
```
JamesRTaylor: Probably best to throw a SQLException here with the JsonParseException or IOException nested inside. Perhaps create an INVALID_JSON_DATA exception in SQLExceptionCode that you can use for the JsonParseException.

phoenix-core/src/main/java/org/apache/phoenix/schema/json/PhoenixJson.java
```diff
@@ -0,0 +1,216 @@
...
+            return getJsonAsString();
+        }
+
+        return null;
+
+    }
+
+    @Override
+    public String toString() {
+        if (this.jsonAsString == null && this.node != null) {
```
JamesRTaylor: Will this be called often? Not sure if there's a potential threading issue. If not called often, might not want to cache the string.

phoenix-core/src/main/java/org/apache/phoenix/schema/types/PDataType.java
```diff
@@ -1144,17 +1144,18 @@ public static PDataType fromLiteral(Object value) {
...
-        if ((type.getSqlType() == arr.baseType.sqlType + PDataType.ARRAY_TYPE_BASE)
-            && type.getJavaClass().isInstance(value)) {
-          return type;
-        }
-      } else {
-        if (type.getJavaClass().isInstance(value)) {
-          return type;
-        }
-      }
+			if (type.isArrayType() && type.getJavaClass().isInstance(value)) {
```
JamesRTaylor: Formatting looks off here (but it's mistakenly using 2 spaces for tabs which isn't correct). You can reformat the entire class if you want.

JamesRTaylor: Actually, if there are no changes here but whitespace, then please just revert this change.

phoenix-core/src/main/java/org/apache/phoenix/schema/types/PJsonDataType.java
```diff
@@ -0,0 +1,258 @@
...
+ * value contains the same key more than once, all the key/value pairs are kept.
+ * It stores the data as string in single column of HBase and it has same data
+ * size limit as Phoenix's Varchar.
+ * <p>
+ * JSON data types are for storing JSON (JavaScript Object Notation) data, as
+ * specified in RFC 7159. Such data can also be stored as text, but the JSON
+ * data types have the advantage of enforcing that each stored value is valid
+ * according to the JSON rules.
+ */
+public class PJsonDataType extends PDataType<String> {
```
JamesRTaylor: Did you consider deriving from PVarchar?

phoenix-core/src/main/java/org/apache/phoenix/schema/types/PVarchar.java
```diff
@@ -30,134 +31,145 @@
 
 public class PVarchar extends PDataType<String> {
```
JamesRTaylor: Any non whitespace changes here - if not, please revert.

JamesRTaylor:
Very nice, @AakashPradeep! @twdsilva - would you mind reviewing too? Also, looks like it needs to be rebased.

phoenix-core/src/main/java/org/apache/phoenix/expression/function/JsonExtractPathTextFunction.java
AakashPradeep: Yeah I missed that. I will add them.

phoenix-core/src/main/java/org/apache/phoenix/schema/types/PDataType.java
AakashPradeep: I will fix the formatting error here. But it is required to check the isInstance() in first IF itself otherwise it throw ClassCastException. For any value whose data type ordinal is less than PhoenixArray will try to be type casted to PhoenixArray at 1148 which is wrong as it is iterated in the order of ordinal where PhoenixArray type will be iterated before PJsonDataType.

phoenix-core/src/main/java/org/apache/phoenix/schema/types/PVarchar.java
AakashPradeep: I will revert this.

phoenix-core/src/main/java/org/apache/phoenix/schema/json/PhoenixJson.java
AakashPradeep: Nice suggestion, will make this change.

phoenix-core/src/main/java/org/apache/phoenix/schema/types/PDataType.java
JamesRTaylor: But a JSON type shouldn't pass the type.isArrayType() right? How about modifying the type.isArrayType() instead?

AakashPradeep: Default implementation of isArrayType() returns false, hence I have not modified it. But in ordered list PVarcharArray comes before PJsonDataType and so value is first testified against PVarcharArraay. Now since PVarcharArray is ArrayType it gets inside the if() and type cast the value which throws ClassCastException since value is of type PJsonDataType. After casting the value, in next if (Line 1150) it validates whether this isInstance() which is not correct. Instead before casting value we should check whether it is instance of current type or not to avoid any ClassCastException.

JamesRTaylor:
@AakashPradeep - this pull request is associated with the wrong JIRA, as it seems to be adding comments to PHOENIX-174. Can you close this pull and open two new ones? One for PHOENIX-628 and then one for PHOENIX-1710 (for the new built-in functions)? Make sure to use the JIRA in the commit message so it get associated correctly.

phoenix-core/src/main/java/org/apache/phoenix/schema/types/PJsonDataType.java
```diff
@@ -0,0 +1,258 @@
...
+
+	@Override
+	public boolean isCoercibleTo(
+			@SuppressWarnings("rawtypes") PDataType targetType) {
+		return equalsAny(targetType, this, PVarchar.INSTANCE);
+
+	}
+
+	@Override
+	public boolean isCoercibleTo(
```
JamesRTaylor: This is the default impl, so no need to repeat this here.

phoenix-core/src/main/java/org/apache/phoenix/schema/types/PJsonDataType.java
```diff
@@ -0,0 +1,258 @@
...
+			@SuppressWarnings("rawtypes") PDataType targetType, Object value) {
+		return isCoercibleTo(targetType);
+	}
+
+	@Override
+	public boolean isSizeCompatible(ImmutableBytesWritable ptr, Object value,
```
JamesRTaylor: Call PVarchar.isSizeCompatible() instead.

phoenix-core/src/main/java/org/apache/phoenix/schema/types/PJsonDataType.java
```diff
@@ -0,0 +1,258 @@
...
+
+	@Override
+	public boolean isFixedWidth() {
+		return false;
+	}
+
+	@Override
+	public int estimateByteSize(Object o) {
+		PhoenixJson phoenixJson = (PhoenixJson) o;
+		return phoenixJson.toString().length();
```
JamesRTaylor: Add a estimateByteSize() on PhoenixJson as you don't want to have to string-ify the JSON just to get an estimate of the size.

phoenix-core/src/main/java/org/apache/phoenix/schema/types/PJsonDataType.java
```diff
@@ -0,0 +1,258 @@
...
+	@Override
+	public int compareTo(Object lhs, Object rhs,
+			@SuppressWarnings("rawtypes") PDataType rhsType) {
+		if (PJsonDataType.INSTANCE != rhsType) {
+			throwConstraintViolationException(rhsType, this);
+		}
+		PhoenixJson phoenixJsonLHS = (PhoenixJson) lhs;
+		PhoenixJson phoenixJsonRHS = (PhoenixJson) rhs;
+		return PVarchar.INSTANCE.compareTo(phoenixJsonLHS.toString(),
+				phoenixJsonRHS.toString(), rhsType);
```
JamesRTaylor: Doesn't Jackson provide a way of comparing two JSONs? Create a compareTo method on PhoenixJson.

phoenix-core/src/main/java/org/apache/phoenix/schema/types/PJsonDataType.java
```diff
@@ -0,0 +1,258 @@
...
+	}
+
+	@Override
+	public Object toObject(String value) {
+		byte[] jsonData = Bytes.toBytes(value);
+		return getPhoenixJson(jsonData, 0, jsonData.length);
```
JamesRTaylor: You don't want to go from String->byte[]->String here. Create a getPhoenixJson(String) method.

phoenix-core/src/main/java/org/apache/phoenix/schema/types/PJsonDataType.java
```diff
@@ -0,0 +1,258 @@
...
+		return PVarchar.INSTANCE.toStringLiteral(phoenixJson.toString(),
+				formatter);
+	}
+
+	@Override
+	public void coerceBytes(ImmutableBytesWritable ptr, Object o,
+			@SuppressWarnings("rawtypes") PDataType actualType,
+			Integer actualMaxLength, Integer actualScale,
+			SortOrder actualModifier, Integer desiredMaxLength,
+			Integer desiredScale, SortOrder expectedModifier) {
```
JamesRTaylor: Same as PVarchar? Or even the default? Don't repeat here.

phoenix-core/src/main/java/org/apache/phoenix/schema/types/PJsonDataType.java
```diff
@@ -0,0 +1,258 @@
...
+		StringBuilder sbr = new StringBuilder();
+		sbr.append("{").append(key).append(":").append(value).append("}");
+
+		byte[] bytes = Bytes.toBytes(sbr.toString());
+		return getPhoenixJson(bytes, 0, bytes.length);
+	}
+
+	private Object getJsonFromVarchar(Object object,
+			@SuppressWarnings("rawtypes") PDataType actualType) {
+		String s = (String) object;
```
JamesRTaylor: Same here - don't go from String->byte[]->String.

phoenix-core/src/main/java/org/apache/phoenix/schema/types/PDataType.java
JamesRTaylor: If type.isArrayType() is true, then you should be able to cast it to a PhoenixArray. Is it because a PVarcharArray is storing the PJsonDataType instead of PVarchar? I think we'll need a PJsonArray.

phoenix-core/src/main/java/org/apache/phoenix/schema/types/PJsonDataType.java
JamesRTaylor: Also, let's just call this PJson as we don't append "DataType" to any of our other PDataType concrete types.

phoenix-core/src/main/java/org/apache/phoenix/schema/types/PDataType.java
AakashPradeep: IMHO Array is a complex data type and so does the json, we should not allow array of json as we dont allow array of array.  @twdsilva  whats your take on this.

phoenix-core/src/main/java/org/apache/phoenix/schema/types/PJsonDataType.java
AakashPradeep: I am ignoring the order of json here since inversion of order will make json invalid.

phoenix-core/src/main/java/org/apache/phoenix/schema/types/PDataType.java
JamesRTaylor: I'm still not clear what is the root of the issue, so let's start there. If the type.isArrayType() is true, then the data type should be coercible to a PhoenixArray. Why isn't it? Is it because it's a PhoenixJson?

phoenix-core/src/main/java/org/apache/phoenix/schema/types/PJsonDataType.java
JamesRTaylor: You can't ignore the sortOrder. It'll be inverted back before any attempts to change it back into JSON. It's the same for other types. For example, an inverted String cannot be converted back until it's re-inverted. Same with Decimal.

phoenix-core/src/main/java/org/apache/phoenix/schema/types/PDataType.java
twdsilva: @AakashPradeep  You had to make this change because there was a test that was failing can you point us to the test ?

PR #76, opened 2015-04-24
PHOENIX-628 Support native JSON data type

AakashPradeep:
This pull request has following changes:

1. Renames
     a)  PJsonDataType to PJson
     b)  PJsonDataTypeTest to PJsonTest 
     c) PhoenixJsonE2ETest to PhoenixJsonIT

2. Added new SqlException Code INVALID_JSON_DATA
3. Some changes in PhoenixJson
4. Removed overridden method from PJson for which default implementation is fine.
        a) coerceBytes()
        b) isCoercibleTo().

@JamesRTaylor and @twdsilva  please review the changelist.  

Note : Some of the classes have space formatting problem, I am trying to fix my eclipse for that. Hopefully it will be resolved in next pull request.


phoenix-core/src/main/java/org/apache/phoenix/schema/json/PhoenixJson.java
```diff
@@ -0,0 +1,280 @@
...
+            phoenixJson.setJsonAsString(jsonData);
+            return phoenixJson;
+        } catch (IOException x) {
+            throw new SQLExceptionInfo.Builder(SQLExceptionCode.INVALID_JSON_DATA).setRootCause(x)
+                    .setMessage(x.getMessage()).build().buildException();
+        }
+
+    }
+
+    private static PhoenixJson getPhoneixJson(JsonParser jsonParser) throws IOException,
```
twdsilva: minor typo getPhoenixJson

AakashPradeep: Nice catch.. I will fix it.

phoenix-core/src/main/java/org/apache/phoenix/schema/json/PhoenixJson.java
```diff
@@ -0,0 +1,280 @@
...
+     * <p>
+     * for this source json, if we want to know the json at path {'f4','f6'} it will return
+     * {@link PhoenixJson} object for json {"f7":"2"}. It always returns the last key if same key
+     * exist more than once.
+     * <p>
+     * If the given path is unreachable then it return null.
+     * @param paths {@link String []} of path in the same order as they appear in json.
+     * @return {@link PhoenixJson} for the json against @paths.
+     */
+    public PhoenixJson getNullablePhoenixJson(String[] paths) {
```
twdsilva: Since getPhoenixJson and getNullablePhoenixJson are almost the same, maybe have a helper function that takes a nullable argument in order to prevent code duplication.

AakashPradeep: getPhoenixJson() throws Exception if path is not found. It can be used on those case where we want to throw exception to user if path is not correct but in case we just want to return null if path is not valid then I dont want my logic to be catching the exception and then ignore it since Exception are always costly so I provided getNullablePhoenixJson() which return null if path is not valid. 

I will try to put them in a helper and have a boolean argument to specify whether to throw exception or not.

phoenix-core/src/main/java/org/apache/phoenix/schema/json/PhoenixJson.java
```diff
@@ -0,0 +1,280 @@
...
+            return getJsonAsString();
+        }
+
+        return null;
+
+    }
+
+    @Override
+    public String toString() {
+        if (this.jsonAsString == null && this.node != null) {
```
twdsilva: can jsonAsString ever be null, seems like you create a PhoenixJson object you it will always have be from a string

AakashPradeep: when we call getPhoenixJson() from PhoenixJson object after applying path (String [] paths) then it will be null.

phoenix-core/src/main/java/org/apache/phoenix/schema/types/PDataType.java
```diff
@@ -1144,17 +1144,18 @@ public static PDataType fromLiteral(Object value) {
...
-        if ((type.getSqlType() == arr.baseType.sqlType + PDataType.ARRAY_TYPE_BASE)
-            && type.getJavaClass().isInstance(value)) {
-          return type;
-        }
-      } else {
-        if (type.getJavaClass().isInstance(value)) {
-          return type;
-        }
-      }
+		if (type.isArrayType() && type.getJavaClass().isInstance(value)) {
```
twdsilva: might be more readable if you just move isInstance outside 

if (type.getJavaClass().isInstance(value)) {
   if (type.isArrayType()) {
......
   }
  else {
     return type;
  }
}

JamesRTaylor: I get it (finally :-) ). I like Thomas' suggestion, though.

phoenix-core/src/main/java/org/apache/phoenix/schema/json/PhoenixJson.java
JamesRTaylor: Typical pattern is to have a method method like getPhoenixJsonOrNull(). Calls call that one if they're ok getting null back. Then you have a getPhoenixJson() which calls getPhoenixJsonOrNull() and throws if it returns null. Will something like that work?

AakashPradeep: Nice suggestion that will work. I will change the code.

phoenix-core/src/main/java/org/apache/phoenix/schema/types/PDataType.java
AakashPradeep: isInstance() is costly it has a native implementation, I will try to avoid it as much as possible.

AakashPradeep: @twdsilva , your placement of isInstance() is good since in current case if it is of Array type but not compatible class then it will call isInstance() twice (once in if () + once in else part )  whereas in your case it will be called only once.

phoenix-core/src/main/java/org/apache/phoenix/expression/ExpressionType.java
```diff
@@ -204,7 +204,7 @@
     SecondFunction(SecondFunction.class),
     WeekFunction(WeekFunction.class),
     HourFunction(HourFunction.class),
-    NowFunction(NowFunction.class)
+    NowFunction(NowFunction.class),
```
JamesRTaylor: No change, please revert.

phoenix-core/src/main/java/org/apache/phoenix/schema/json/PhoenixJson.java
```diff
@@ -0,0 +1,266 @@
...
+            JsonFactory jsonFactory = new JsonFactory();
+            JsonParser jsonParser = jsonFactory.createJsonParser(jsonData);
+            PhoenixJson phoenixJson = getInstanceInternal(jsonParser);
+            /*
+             * input data has been stored as it is, since some data is lost when json parser runs,
+             * for example if a JSON object within the value contains the same key more than once
+             * then only last one is stored rest all of them are ignored, which will defy the
+             * contract of PJsonDataType of keeping user data as it is.
+             */
+            phoenixJson.setJsonAsString(jsonData);
```
JamesRTaylor: Pass the String through the PhoenixJson constructor. PhoenixJson should be immutable.

phoenix-core/src/main/java/org/apache/phoenix/schema/types/PJson.java
```diff
@@ -0,0 +1,223 @@
...
+        value[k++] = 'v';
+        for (int i = 2; i < value.length - 1; i++) {
+            value[k++] = (char) ('0' + RANDOM.get().nextInt(Byte.MAX_VALUE) % 10);
+        }
+        value[k] = '"';
+        StringBuilder sbr = new StringBuilder();
+        sbr.append("{").append(key).append(":").append(value).append("}");
+
+        byte[] bytes = Bytes.toBytes(sbr.toString());
+        return getPhoenixJson(bytes, 0, bytes.length);
```
JamesRTaylor: Get rid of this method: getPhoenixJson(byte[], offset, length). Instead, just have getPhoenixJson(String), since you're relying on having a String. Converting from a String -> byte[] -> String which is a waste.

phoenix-core/src/main/java/org/apache/phoenix/schema/types/PJson.java
```diff
@@ -0,0 +1,223 @@
...
+
+    }
+
+    @Override
+    public byte[] toBytes(Object object) {
+        if (object == null) {
+            return ByteUtil.EMPTY_BYTE_ARRAY;
+        }
+        PhoenixJson phoenixJson = (PhoenixJson) object;
+        return PVarchar.INSTANCE.toBytes(phoenixJson.toString());
```
JamesRTaylor: Better if you delegate the toBytes() call to PhoenixJson. That'll give you the flexibility to cache the byte[] if you wanted to.

phoenix-core/src/main/java/org/apache/phoenix/schema/types/PJson.java
```diff
@@ -0,0 +1,223 @@
...
+    public Object toObject(byte[] bytes, int offset, int length,
+            @SuppressWarnings("rawtypes") PDataType actualType, SortOrder sortOrder,
+            Integer maxLength, Integer scale) {
+
+        if (!actualType.isCoercibleTo(this)) {
+            throwConstraintViolationException(actualType, this);
+        }
+        if (length == 0) {
+            return null;
+        }
```
JamesRTaylor: Here you want to call through to PVarchar to get a String, then call getPhoenixJson(String). This will take care of the maxLength check and the inversion if necessary based on sortOrder.

    String jsonStr = PVarChar.toObject(bytes, offset, length, actualType, sortOrder, maxLength, scale);
    return getPhoenixJson(jsonStr);

phoenix-core/src/main/java/org/apache/phoenix/schema/json/PhoenixJsonException.java
```diff
@@ -0,0 +1,32 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.phoenix.schema.json;
+
+public class PhoenixJsonException extends Exception {
```
JamesRTaylor: Get rid of this and instead declare your checked exception as SQLException

phoenix-core/src/main/java/org/apache/phoenix/schema/json/PhoenixJson.java
```diff
@@ -0,0 +1,266 @@
...
+import com.google.common.base.Preconditions;
+
+/**
+ * The {@link PhoenixJson} wraps json and uses Jackson library to parse and traverse the json. It
+ * should be used to represent the JSON data type and also should be used to parse Json data and
+ * read the value from it. It always conside the last value if same key exist more than once.
+ */
+public class PhoenixJson implements Comparable<PhoenixJson> {
+    private final JsonNode node;
+    private String jsonAsString;
```
AakashPradeep: Needs suggestions to decide whether json data should be cached as byte [] or String. 

byte[]
-------

If it is byte[] following methods would not involve any String to byte [] or byte [] to String conversion 
1. toBytes(Object object, byte[] bytes, int offset)
2. toBytes(Object object)
3. toObject(byte[] bytes, int offset, int length,
			PDataType actualType, SortOrder sortOrder, Integer maxLength,
			Integer scale)


But for following, we require byte [] -> String 

1. toStringLiteral(Object o, Format formatter)

And for toObject(Object object, @SuppressWarnings("rawtypes") PDataType actualType) if actualType is Varchar we require String -> byte []


String
-----

Where as in case of String, following methods will require String -> byte []

1. toBytes(Object object, byte[] bytes, int offset)
2. toBytes(Object object)
3. toObject(byte[] bytes, int offset, int length,
			PDataType actualType, SortOrder sortOrder, Integer maxLength,
			Integer scale)

No conversion required for 

toStringLiteral(Object o, Format formatter) and toObject(Object object, @SuppressWarnings("rawtypes") PDataType actualType) if actualType is Varchar.



byte[] seems to be more efficient here but I am not sure which is most used method and will help in performance or not. 

@JamesRTaylor  @twdsilva

JamesRTaylor: Cache String as you always need a String for parsing and comparison.

>

AakashPradeep:
@twdsilva  @JamesRTaylor  I have made changes as per the comments. Please let me know if there is anything more.

phoenix-core/src/main/java/org/apache/phoenix/schema/types/PVarchar.java
```diff
@@ -30,134 +31,145 @@
...
-    int length = maxLength != null ? maxLength : 1;
-    if (length != sampleChars.length) {
-      sampleChars = new char[length];
-    }
-    for (int i = 0; i < length; i++) {
-      sampleChars[i] = (char) RANDOM.get().nextInt(Byte.MAX_VALUE);
-    }
-    return new String(sampleChars);
-  }
+	public static final PVarchar INSTANCE = new PVarchar();
```
JamesRTaylor: Minor nit: indenting looks off here. Should be no tabs and 4 space chars.

phoenix-core/src/main/java/org/apache/phoenix/schema/json/PhoenixJson.java
```diff
@@ -0,0 +1,255 @@
...
+                nodeTemp = node.path(index);
+            } else {
+                nodeTemp = node.path(path);
+            }
+            if (nodeTemp == null || nodeTemp.isMissingNode()) {
+                return null;
+            }
+            node = nodeTemp;
+        }
+        return new PhoenixJson(node, null);
```
JamesRTaylor: Might be better to just call node.toString() here instead of passing in null. Otherwise, the getJsonAsString() and toString() methods will allocate a big string on each call.

phoenix-core/src/main/java/org/apache/phoenix/schema/json/PhoenixJson.java
```diff
@@ -0,0 +1,255 @@
...
+        result = prime * result + ((this.rootNode == null) ? 0 : this.rootNode.hashCode());
+        return result;
+    }
+
+    @Override
+    public boolean equals(Object obj) {
+        if (this == obj) return true;
+        if (obj == null) return false;
+        if (getClass() != obj.getClass()) return false;
+        PhoenixJson other = (PhoenixJson) obj;
```
JamesRTaylor: Why do we need to compare both String and the JsonNode here? Seems like we should just compare the JsonNode instead.

phoenix-core/src/main/java/org/apache/phoenix/schema/json/PhoenixJson.java
```diff
@@ -0,0 +1,255 @@
...
+
+    public byte[] toBytes() {
+        return Bytes.toBytes(getJsonAsString());
+    }
+
+    @Override
+    public int compareTo(PhoenixJson o) {
+        if (o == null) {
+            return 1;
+        } else if (this.equals(o)) {
```
JamesRTaylor: Checking both JsonNode.equals and doing a string compare on the string representation is not so good. Does JSON have the concept of < and > comparison? Should we always rely on string equality to mean node equality? When would that break down? If that works, we should perhaps consistently check that for equality and hashcode, etc. If that doesn't work, we may need to introduce the concept of a type that doesn't support comparison, but only = and !=.

phoenix-core/src/main/java/org/apache/phoenix/schema/json/PhoenixJson.java
AakashPradeep: I think String match will faster since JsonNode.equals() traverse the entire Tree Model of Json.

phoenix-core/src/main/java/org/apache/phoenix/schema/json/PhoenixJson.java
AakashPradeep: Good that you brought it up. I was not happy with it and about to discuss. Jackson library does not have any Json Comparator.  I thought of writing own comparator but there are too many unknowns and did not find any specification around it. I will have to assume too many things which is not good. 

Probably we should not support '=' and '!=' with Json data type. Postgres  also throw an error on using '=' operator with json column

# select * from t1 where c1 ='[1,2,3]';
ERROR:  operator does not exist: json = unknown
LINE 1: select * from t1 where c1 ='[1,2,3]';
                                  ^
HINT:  No operator matches the given name and argument type(s). You might need to add explicit type casts.

where as this works 
# select * from t1 where c1::varchar ='[1,2,3]';
   c1    
---------
 [1,2,3]
(1 row)

AakashPradeep: Is throwing an exception from compareTo(), will help us?

Since postgres does not allow '=' and order by with json.

AakashPradeep: Or should we do something like PArrayDataType.compareTo() 

public int compareTo(Object lhs, Object rhs) {
		PhoenixArray lhsArr = (PhoenixArray) lhs;
		PhoenixArray rhsArr = (PhoenixArray) rhs;
		if(lhsArr.equals(rhsArr)) {
			return 0;
		}
		return 1;
	}

JamesRTaylor: That just seems wrong. The compare(Object, Object, PDataType) method is only called in a few places. For example, with COUNT DISTINCT. We rely on the compareTo() method that takes byte[] more. I think we should just do the String comparison as that'll match what the byte[] based compare method does.

AakashPradeep: ok. that make sense. I will change it to String comparison.

phoenix-core/src/main/java/org/apache/phoenix/schema/json/PhoenixJson.java
```diff
@@ -0,0 +1,252 @@
...
+            if (nodeTemp == null || nodeTemp.isMissingNode()) {
+                return null;
+            }
+            node = nodeTemp;
+        }
+        return new PhoenixJson(node, node.toString());
+    }
+    
+    private String getJsonAsString() {
+        if (this.jsonAsString != null) {
```
JamesRTaylor: Should just be the following, as it's always passed in:
    return this.jsonAsString;

phoenix-core/src/main/java/org/apache/phoenix/schema/json/PhoenixJson.java
```diff
@@ -0,0 +1,252 @@
...
+    @Override
+    public boolean equals(Object obj) {
+        if (this == obj) return true;
+        if (obj == null) return false;
+        if (getClass() != obj.getClass()) return false;
+        PhoenixJson other = (PhoenixJson) obj;
+        if ((this.jsonAsString != null) && (other.jsonAsString != null)) {
+            if (this.jsonAsString.equals(other.jsonAsString)) return true;
+        }
+        if (this.rootNode == null) {
```
JamesRTaylor: Remove references to rootNode.equals(). We can just rely on the jsonString being equal. We can do more for the JSONB data type, but this will be the most consistent way to go for now.

phoenix-core/src/main/java/org/apache/phoenix/schema/json/PhoenixJson.java
```diff
@@ -0,0 +1,252 @@
...
+    @Override
+    public String toString() {
+        return getJsonAsString();
+    }
+
+    @Override
+    public int hashCode() {
+        final int prime = 31;
+        int result = 1;
+        result = prime * result + ((this.rootNode == null) ? 0 : this.rootNode.hashCode());
```
JamesRTaylor: Change to use this.jsonAsString.hashCode().

phoenix-core/src/main/java/org/apache/phoenix/schema/json/PhoenixJson.java
AakashPradeep: String equals wont be able to match this kind of json equality 

{"k1":"v1","k2":"v2"}  against {"k2":"v2","k1":"v1"}.

JamesRTaylor: If a JSON type is used in the row key (where the comparison is done at the byte level), this won't be considered equals either. Better in my mind to be consistent. We should normalize the JSON on input if order of tags doesn't matter.

AakashPradeep: We have two options:
1) We should not allow json column as primary key as in Postgres.
     create table json_pk(c1 json primary key); 
    ERROR:  data type json has no default operator class for access method "btree"
    HINT:  You must specify an operator class for the index or define a default operator class for the   data type

2) We should normalize it. 

Which one you suggest ?

JamesRTaylor: I already suggested it: (2). We don't have a declarative way of doing (1) right now, so (2) is the only viable option.

AakashPradeep: I looked into json specification and it says "Json: An object is an unordered set of name/value pairs". So order of key value does not matter.  I looked into postgres and this what i found  

1) does not support = and != operator
2) not allow to group by or use count distinct with json column basically they do not allow any equality operator on json. 

Here is the error message in Postgres 

select count(distinct(c1)) from t1;
ERROR:  could not identify an equality operator for type json
LINE 1: select count(distinct(c1)) from t1;
                              ^

select count(*) from t1 group by c1 ;
ERROR:  could not identify an equality operator for type json
LINE 1: select count(*) from t1 group by c1 ;
                                         ^


IMHO we should do the same for JSON in Phoenix. And allow followings:

1. select count(*) from json_table group by json_extract_path_text(json_column, ARRAY['k1']) ;
2. select * from json_table where json_extract_path_text(json_column, ARRAY['k1'] )='v1';

AakashPradeep: Sorry did not read your last comment. So do you want to normalize it as part of this JIRA ?

PR #77, opened 2015-04-24
PHOENIX-538 Support UDFs

chrajeshbabu:
Patch to support UDFs. It mainly includes
- create temporary/permanent function query parsing
- storing function info
- dynamically loading udf jars.
- resolve functions 
- making use of udfs in different queries
- drop function.
- it tests


phoenix-core/src/main/java/org/apache/phoenix/cache/GlobalCache.java
```diff
@@ -157,4 +159,18 @@ public TenantCache getChildTenantCache(ImmutableBytesWritable tenantId) {
...
     }
+
+    public static class FunctionBytesPtr extends ImmutableBytesPtr {
+
+        public FunctionBytesPtr(byte[] key) {
+            super(key);
+        }
+
+        @Override
+        public boolean equals(Object obj) {
```
samarthjain: Should you be overriding the hashCode() method too?

phoenix-core/src/main/java/org/apache/phoenix/compile/CreateFunctionCompiler.java
```diff
@@ -0,0 +1,84 @@
...
+    }
+
+    public MutationPlan compile(final CreateFunctionStatement create) throws SQLException {
+        if(create.isReplace()) {
+            throw new SQLFeatureNotSupportedException();
+        }
+        final PhoenixConnection connection = statement.getConnection();
+        PhoenixConnection connectionToBe = connection;
+        Scan scan = new Scan();
+        final StatementContext context = new StatementContext(statement, FromCompiler.EMPTY_TABLE_RESOLVER, scan, new SequenceManager(statement));
```
samarthjain: Minor nit: You can use the other constructor here public StatementContext(PhoenixStatement statement)

phoenix-core/src/main/java/org/apache/phoenix/compile/FromCompiler.java
```diff
@@ -383,6 +448,85 @@ protected TableRef createTableRef(NamedTableNode tableNode, boolean updateCacheI
             return tableRef;
         }
 
+        @Override
+        public List<PFunction> getFunctions() {
+            return functions;
+        }
+
+        protected List<PFunction> createFunctionRef(List<String> functionNames, boolean updateCacheImmediately) throws SQLException {
```
samarthjain: Minor nit: is this function overriden somewhere? If not, it could be made private

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/MetaDataEndpointImpl.java
```diff
@@ -700,6 +1007,29 @@ private PTable loadTable(RegionCoprocessorEnvironment env, byte[] key,
...
+            PFunction function = (PFunction)metaDataCache.getIfPresent(cacheKey);
+            // We always cache the latest version - fault in if not in cache
+            if (function != null) {
+                return function;
+            }
+            ArrayList<byte[]> arrayList = new ArrayList<byte[]>(1);
+            arrayList.add(key);
+            List<PFunction> functions = buildFunctions(arrayList, region, asOfTimeStamp);
+            if(functions != null) return functions.get(0);
+            // if not found then check if newer table already exists and add delete marker for timestamp
```
samarthjain: newer function?

phoenix-core/src/main/java/org/apache/phoenix/expression/function/UDFExpression.java
```diff
@@ -0,0 +1,217 @@
...
+            Path path = new Path(jarPath);
+            if(jarPath.endsWith(".jar")) {
+                parent = path.getParent().toString();
+            } else {
+                parent = path.toString();
+            }
+        }
+        if (jarPath == null || jarPath.isEmpty() || config.get(DYNAMIC_JARS_DIR_KEY) != null
+                && (parent != null && parent.equals(config.get(DYNAMIC_JARS_DIR_KEY)))) {
+            Lock lock = locker.acquireLock(tenantId.getString());
```
samarthjain: Rajesh says: Just want to clarify one thing here the map is concurrent map so get and putifAbsent will be atomic but there is a chance like between get and put if other client already puts a class loader for the tenant then the classloader and configuration object created become garbage. That's why using locker. If that's ok I can do as you suggested. What do u say?

Samarth says: How expensive is creating a new classloader and config? To me it doesn't look too expensive. So I wouldn't worry about creating and discarding the two. Besides that race condition of another thread winning the race between get() and putIfAbsent is not very likely. So to sum it up, yes, I would get rid of the KeyLocker altogether.

phoenix-core/src/test/java/org/apache/phoenix/parse/QueryParserTest.java
```diff
@@ -289,24 +289,6 @@ public void testNegativeCountStar() throws Exception {
     }
 
     @Test
-    public void testUnknownFunction() throws Exception {
```
samarthjain: Is there a reason why this test was removed? Are we no longer throwing an error? Or has the exception code changed? I would think if the user doesn't have load udf enabled then this error is still valid.

samarthjain:
Almost there Rajesh. Just some minor nits and questions, otherwise LGTM.

phoenix-core/src/main/java/org/apache/phoenix/compile/FromCompiler.java
chrajeshbabu: no will change it to private.

phoenix-core/src/main/java/org/apache/phoenix/cache/GlobalCache.java
chrajeshbabu: I will override the hashcode and just call super.hashCode() then we won't get checkstyle or findbug warning.

phoenix-core/src/main/java/org/apache/phoenix/compile/CreateFunctionCompiler.java
chrajeshbabu: That would be great. Will change.

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/MetaDataEndpointImpl.java
chrajeshbabu: yes

phoenix-core/src/main/java/org/apache/phoenix/expression/function/UDFExpression.java
chrajeshbabu: Sure then will remove it.

phoenix-core/src/test/java/org/apache/phoenix/parse/QueryParserTest.java
chrajeshbabu: Now the exception will be thrown during compilation not while parsing because while parsing not able check whether udf enabled or not. That's why removed it.

chrajeshbabu:
Thanks Samarth for reviews. Will update patch addressing the comments.

phoenix-core/src/main/java/org/apache/phoenix/expression/function/UDFExpression.java
```diff
@@ -0,0 +1,217 @@
...
+import org.apache.phoenix.parse.PFunction;
+import org.apache.phoenix.schema.PName;
+import org.apache.phoenix.schema.PNameFactory;
+import org.apache.phoenix.schema.tuple.Tuple;
+import org.apache.phoenix.schema.types.PDataType;
+
+import com.google.common.annotations.VisibleForTesting;
+import com.google.common.collect.MapMaker;
+
+public class UDFExpression extends ScalarFunction {
```
JamesRTaylor: Having getDeterminism() return PER_INVOCATION is the only way to get each thread to have its own copy of the expression tree. Can  you file a JIRA with the failing test and exception? Is it only for functional indexing that you run into an issue?

chrajeshbabu: if (expression.getDeterminism() != Determinism.ALWAYS) {
                        throw new SQLExceptionInfo.Builder(SQLExceptionCode.NON_DETERMINISTIC_EXPRESSION_NOT_ALLOWED_IN_INDEX).build().buildException();
                    }

Here it's the reason for failure James...getting problem with functional indexes only.

chrajeshbabu: java.sql.SQLException: ERROR 521 (42898): Non-deterministic expression not allowed in an index
	at org.apache.phoenix.exception.SQLExceptionCode$Factory$1.newException(SQLExceptionCode.java:386)
	at org.apache.phoenix.exception.SQLExceptionInfo.buildException(SQLExceptionInfo.java:145)
	at org.apache.phoenix.schema.MetaDataClient.createIndex(MetaDataClient.java:1179)
	at org.apache.phoenix.compile.CreateIndexCompiler$1.execute(CreateIndexCompiler.java:95)
	at org.apache.phoenix.jdbc.PhoenixStatement$3.call(PhoenixStatement.java:303)
	at org.apache.phoenix.jdbc.PhoenixStatement$3.call(PhoenixStatement.java:1)
	at org.apache.phoenix.call.CallRunner.run(CallRunner.java:53)
	at org.apache.phoenix.jdbc.PhoenixStatement.executeMutation(PhoenixStatement.java:294)
	at org.apache.phoenix.jdbc.PhoenixStatement.execute(PhoenixStatement.java:1247)
	at org.apache.phoenix.end2end.UserDefinedFunctionsIT.testFunctionalIndexesWithUDFFunction(UserDefinedFunctionsIT.java:509)

JamesRTaylor: I see - forgot about that, and we shouldn't change that. Instead, in the RowProjector constructor we'll want to set cloneRequired to true if any of the expressions use a UDF. The easiest way is probably to add a boolean hasUDFs to the RowProjector, add a ColumnResolver.hasUDFs() that returns true if !functions.isEmpty(). This will still clone if a UDF is used in the WHERE clause, but not in the SELECT expressions, but that's ok (it won't harm anything).

phoenix-core/src/main/antlr3/PhoenixSQL.g
```diff
@@ -114,6 +114,15 @@ tokens
     ASYNC='async';
     SAMPLING='sampling';
     UNION='union';
+    FUNCTION='function';
+    AS='as';
+    REPLACE='replace';
```
JamesRTaylor: If support for REPLACE won't make this pull request (which is fine b/c it's already big enough), we should just remove it from the grammar for now.

phoenix-core/src/main/java/org/apache/phoenix/expression/function/UDFExpression.java
chrajeshbabu: Added this James. Thanks for alternative soln.

phoenix-core/src/main/antlr3/PhoenixSQL.g
chrajeshbabu: Removed replace currently.

chrajeshbabu:
Thanks @JamesRTaylor  @samarthjain 
I have addressed the review comments and added to pull request. If it's ok I will commit this tomorrow morning IST and work on subtasks.

chrajeshbabu:
It's committed. Hence closing.

PR #79, opened 2015-04-27
PHOENIX-1875 ARRAY_PREPEND function implemented 

dumindux:
PR for the ARRAY_PREPEND function with tests.
Here an issue with the ARRAY_APPEND function is also fixed(I have found it while doing array prepend. sorry about that. I haven't added ptr offset while obtaining array element offsets with getOffset())


phoenix-core/src/main/java/org/apache/phoenix/schema/types/PArrayDataType.java
```diff
@@ -543,6 +544,156 @@ private static void writeEndBytes(byte[] array, int newOffsetArrayPosition, int
...
+                    } else {
+                        //gets the offset of the first element after nulls at the beginning
+                        firstNonNullElementPosition = currOffset;
+                        break;
+                    }
+                }
+                nulls++;
+
+                int nMultiplesOver255 = nulls / 255;
+                endElementPosition = getOffset(arrayBytes, arrayLength - 1, !useInt, offsetArrayPosition + offset) + nMultiplesOver255 + 2 * Bytes.SIZEOF_BYTE;
```
ramkrish86: Why overwriting endElementPosition - better do this only once where you want it.

dumindux: You mean without assigning it to a variable directly using it in PArrayDataType.useShortForOffsetArray(maxOffset) method?

phoenix-core/src/main/java/org/apache/phoenix/schema/types/PArrayDataType.java
```diff
@@ -543,6 +544,151 @@ private static void writeEndBytes(byte[] array, int newOffsetArrayPosition, int
...
+            if (elementLength == 0) {
+                int nulls = 1;
+                //counts the number of nulls which are already at the beginning of the array
+                for (int index = 0; index < arrayLength; index++) {
+                    int currOffset = getOffset(arrayBytes, index, !useInt, offsetArrayPosition + offset);
+                    if (arrayBytes[offset + currOffset] == QueryConstants.SEPARATOR_BYTE) {
+                        nulls++;
+                    } else {
+                        //gets the offset of the first element after nulls at the beginning
+                        firstNonNullElementPosition = currOffset;
```
ramkrish86: Why this firstNonNullElementPosition is needed and that is getting subtracted in forming the newArray's length?

dumindux: firstNonNullElementPosition is used to know how many bytes are there before the first non null element in the original array bytes(that is number of bytes used to represent nulls at the beginning). when calculating the new array bytes length, bytes required for storing nulls at the beginning is calculated(nMultiplesOver255 + bytesForRemainingNulls + Bytes.SIZEOF_BYTE) and added to the original array bytes length(length with all the bytes) then firstNonNullElementPosition is subtracted from it to subtract the number of bytes used to represent nulls at the beginning in the original array(since those are included in the original array bytes length)

ramkrish86: Yes I get that reading the code, but am just seeing if this logic can be simplified.  There can be two cases, one when the prepended element is null and another is not null. for now we will talk about the null case alone.  Unless the number of nulls is not going to be > 255 most of the case appending a  null is not going to change the length of the array right?  It is only going to increment the count of the array that would have been in offset 1.  When ever the null count reaches 255 and more ( or it is was already above 255) there could be a change in the length.  
Just trying to simplify here.

ramkrish86: I verified the logic once again.  The array size is going to change when ever the array is going to increase just above the multiples of 255.  In all other cases the size is going to be the same including the offset shift.  Better to simplify the code.

phoenix-core/src/main/java/org/apache/phoenix/schema/types/PArrayDataType.java
```diff
@@ -543,6 +544,151 @@ private static void writeEndBytes(byte[] array, int newOffsetArrayPosition, int
...
+        int currentPosition = newOffsetArrayPosition;
+        int offsetArrayElementSize = useShortNew ? Bytes.SIZEOF_SHORT : Bytes.SIZEOF_INT;
+        if (useShortNew) {
+            Bytes.putShort(newArray, currentPosition, (short) (0 - Short.MAX_VALUE));
+        } else {
+            Bytes.putInt(newArray, currentPosition, 0);
+        }
+
+        currentPosition += offsetArrayElementSize;
+        boolean isFirstNullGroup = true;
```
ramkrish86: Better name for this?  Says 'nullsAtBeginning' (may be something better)?

phoenix-core/src/test/java/org/apache/phoenix/expression/ArrayPrependFunctionTest.java
```diff
@@ -0,0 +1,492 @@
...
+        Object element = null;
+        PDataType baseType = PChar.INSTANCE;
+
+        PhoenixArray arr = new PhoenixArray(baseType, o);
+        PhoenixArray expected = new PhoenixArray(baseType, o2);
+        test(arr, element, PDataType.fromTypeId(baseType.getSqlType() + PDataType.ARRAY_TYPE_BASE), 4, null, baseType, 1, null, expected, SortOrder.ASC, SortOrder.DESC);
+    }
+
+    @Test
+    public void testArrayPrependFunction23() throws Exception {
```
ramkrish86: Good

phoenix-core/src/test/java/org/apache/phoenix/expression/ArrayPrependFunctionTest.java
```diff
@@ -0,0 +1,492 @@
...
+
+        PhoenixArray arr = new PhoenixArray(baseType, o);
+        PhoenixArray expected = new PhoenixArray(baseType, o2);
+        test(arr, element, PDataType.fromTypeId(baseType.getSqlType() + PDataType.ARRAY_TYPE_BASE), null, null, baseType, 1, null, expected, SortOrder.ASC, SortOrder.DESC);
+    }
+
+    @Test
+    public void testArrayPrependFunction25() throws Exception {
+        //254 nulls at the beginning
+        Object[] o = new Object[256];
```
ramkrish86: Nice.  Just add one more where we go beyond 510 to show we are able to handle all cases and this logic works fine.

ramkrish86: Also try to names these tests that is specifically for nulls as testForNullWith255Nullsinbeginning, testforNullsWith510NullsinBeginning.
Also add a test for all null case.  Just to see how it works.

phoenix-core/src/main/java/org/apache/phoenix/expression/function/ArrayPrependFunction.java
```diff
@@ -0,0 +1,96 @@
...
+import org.apache.hadoop.hbase.io.ImmutableBytesWritable;
+import org.apache.phoenix.expression.Expression;
+import org.apache.phoenix.parse.FunctionParseNode;
+import org.apache.phoenix.schema.SortOrder;
+import org.apache.phoenix.schema.TypeMismatchException;
+import org.apache.phoenix.schema.tuple.Tuple;
+import org.apache.phoenix.schema.types.*;
+
+@FunctionParseNode.BuiltInFunction(name = ArrayPrependFunction.NAME, args = {
+        @FunctionParseNode.Argument(allowedTypes = {PVarbinary.class}),
```
ramkrish86: Will it better to unify even this?  Let children(0)  be the array and children(1) be the element to be appended/prepended.  Its all about the function as to where it gets added in the array?

dumindux: Here I just followed postgreSQL definition of the function(array_prepend(anyelement, anyarray)). So is it Ok to unify this?

ramkrish86: That's a good point.  Will see what James has to say before we are good to go for commit.

dumindux: Ok great. I will work on the other changes.

stoty:
Already merged.

PR #81, opened 2015-04-30
PHOENIX-1938 Like operator should throw proper exception when it is used with data type other then Varcahr and Char

AakashPradeep:
LIKE operator should be supported only for Varchar and Char and it should throw SQLException with SQLExceptionCode.TYPE_NOT_SUPPORTED_FOR_OPERATOR for any other data type. 

Postgres specification : http://www.postgresql.org/docs/8.3/static/functions-matching.html
Oracle specification: http://docs.oracle.com/cd/B12037_01/server.101/b10759/conditions016.htm

@JamesRTaylor  @twdsilva  Can you please review this.

### DIFF
--- a/phoenix-core/src/it/java/org/apache/phoenix/end2end/ExpressionCompilerIT.java
+++ b/phoenix-core/src/it/java/org/apache/phoenix/end2end/ExpressionCompilerIT.java
@@ -1,0 +1,86 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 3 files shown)


AakashPradeep:
@JamesRTaylor and @twdsilva  can you please review it. It is required for Json.

AakashPradeep:
As per the SQL specification, LIKE only works with String data type (http://www.contrib.andrew.cmu.edu/~shadow/sql/sql1992.txt search for "Like predicate") . 
If it has been set to PVarchar.isCoercible() that means anything which can be coerced to Varchar but is not varchar for example VarBinary and JSON, can also be used with "Like". 

IMHO Like should only work with Varhcar and char datatype and throw illegal data type error with other datatype arguments.

JamesRTaylor:
Actually, change it to rhs.getDataType().isCoercibleTo(PVarChar.INSTANCE). That's more or less as you've described, but done in a way that doesn't hard code type checks. Our PDataType does not necessarily map 1:1 with SQL types, so down the road we might add a PUTF16EncodedType which would still map to a VARCHAR is SQL. This type of check makes a change like this more manageable.

AakashPradeep:
Agreed, but then for all data type like JSON which is coercible to Varchar or Char but does not want to support "LIKE" will have to add an if statement.

Do you think thats ok ?

JamesRTaylor:
I'd make JSON castable to VARCHAR instead of coercible. That way you can do an explicit CAST, but it would not be auto-coerced.

PR #82, opened 2015-05-01
Update CsvBulkLoadTool.java by changing the default delimiter to '\t' from ',' to support Bulkload better

twinklestart:
default delimiter was ',' , in real using ,The delimiter of data is always '\t',But the option '-d' can only support single char as delimiter,so change the default delimiter to '\t' to support it .


gabrielreid:
I'm not really in favor of this change, as it would break the default working of the loader, and because it actually *is* possible to specify a tab (or other escape-based character) as the separator, although the way of doing this is not well known.

You can supply a tab character on the command line in the current loader in two ways:
1. after typing the "-d" and a leading single quote, type Ctrl+v and then hit the tab button, and a closing single quote. This provides a tab literal on the command line
2. enter the -d parameter as -d $'\t' 

This general issue (and both of these solutions) are not Phoenix-specific, they're just the way in which you provide escape character values on the command line.

I'm going to update the documentation on http://phoenix.apache.org/bulk_dataload.html to include this information.

twinklestart:
ok,I have tried  this two ways to load data ,it works well,thanks,I will close this pull request

PR #87, opened 2015-05-22
PBoolean.toBytes(object, sortOrder) negates the value

fiserro:
Same issue in older releases.
Try:
System.out.println(PDataType.BOOLEAN.toObject(PDataType.BOOLEAN.toBytes(true, SortOrder.ASC), SortOrder.ASC));


fiserro:
Unfortunately it's still wrong

PR #89, opened 2015-06-18
Mixed R/W workload support

codymarcel:


codymarcel:
@mujtabachohan @twdsilva

codymarcel:
@samarthjain

phoenix-pherf/src/main/java/org/apache/phoenix/pherf/workload/QueryExecutor.java
```diff
@@ -18,227 +18,271 @@
...
+                Map<String, String> phoenixProperty = conf.getValByRegex("phoenix");
+                try {
+
+                    for (Scenario scenario : scenarios) {
+                        ScenarioResult scenarioResult = new ScenarioResult(scenario);
+                        scenarioResult.setPhoenixProperties(phoenixProperty);
+                        dataModelResult.getScenarioResult().add(scenarioResult);
+                        WriteParams writeParams = scenario.getWriteParams();
+
+                        if (writeParams != null) {
```
codymarcel: This is where writers are started during read scenarios. This is what ties it all together.

phoenix-pherf/src/it/java/org/apache/phoenix/pherf/DataIngestIT.java
```diff
@@ -18,70 +18,122 @@
...
-    protected static PhoenixUtil util = new PhoenixUtil(true);
-    static final String matcherScenario = ".*scenario/.*test.*xml";
-    static final String matcherSchema = ".*datamodel/.*test.*sql";
 
-    @Test
-    public void generateData() throws Exception {
-        util.setZookeeper("localhost");
-        SchemaReader reader = new SchemaReader(util, matcherSchema);
-        XMLConfigParser parser = new XMLConfigParser(matcherScenario);
+    @org.junit.Before
```
samarthjain: Extremely minor - do an import so this becomes @Before

phoenix-pherf/src/it/java/org/apache/phoenix/pherf/PherfMainIT.java
```diff
@@ -0,0 +1,37 @@
...
+public class PherfMainIT extends ResultBaseTestIT {
+
+    @Test
+    public void testPherfMain() throws Exception{
+        String[] args = {"-q", "--scenarioFile", ".*prod_test_unsalted_scenario.*" };
+        try {
+            Pherf.main(args);
+        } catch (Exception e) {
+            fail("Test encountered an exception.");
+            throw e;
```
samarthjain: throw e will never be executed.

phoenix-pherf/src/main/java/org/apache/phoenix/pherf/jmx/MonitorManager.java
```diff
@@ -135,10 +154,7 @@ public synchronized boolean shouldStop() {
         return shouldStop;
     }
 
-    public synchronized void stop() {
-        this.shouldStop = true;
-    }
-
+    @SuppressWarnings("unused")
```
samarthjain: Why is this annotation needed here? If the method if not being used, then we should just remove it?

phoenix-pherf/src/it/java/org/apache/phoenix/pherf/PherfMainIT.java
codymarcel: yep, will remove the try. I think junit will automatically fail if an exception is thrown. This is just wrong.

phoenix-pherf/src/main/java/org/apache/phoenix/pherf/jmx/MonitorManager.java
codymarcel: I found it useful for testing and debugging. I just didn't leave any of the calling code in. I didn't want to have to re-add it later. I can remove it you don't like it though.

phoenix-pherf/src/main/java/org/apache/phoenix/pherf/workload/WorkloadExecutor.java
```diff
@@ -19,95 +19,92 @@
...
 
-        this.pool = Executors.newFixedThreadPool(this.poolSize);
-        if (monitor) {
-            initMonitor(Integer.parseInt(properties.getProperty("pherf.default.monitorFrequency")));
+    /**
+     * Blocks on waiting for all workloads to finish. If a
+     * {@link org.apache.phoenix.pherf.workload.Workload} Requires complete() to be called, it must
+     * be called prior to using this method. Otherwise it will block infinitely.
+     */
+    public synchronized void get() {
```
samarthjain: The synchronization is probably not needed here unless member variable "jobs" is being modified concurrently. In that case you want to have synchronization everywhere "jobs" is accessed (both read and modified) (for ex - complete() ). It would be also good to add a @GuardedBy("this") annotation for "jobs".

phoenix-pherf/src/main/java/org/apache/phoenix/pherf/Pherf.java
```diff
@@ -146,49 +173,77 @@ public void run() throws Exception {
...
             }
 
-            // Schema and Data Load
-            if (loadData) {
+            if (monitor) {
+                MonitorManager
+                        monitorManager =
+                        new MonitorManager(Integer.parseInt(
+                                properties.getProperty("pherf.default.monitorFrequency")));
+                workloadExecutor.add(monitorManager);
```
samarthjain: Should we wait for monitor job to be executed or kicked off by doing a workloadExecutor.get() or workloadExecutor.get(monitorManager)? Or is the intent to rely on subsequent workloadExecutor.get() calls to have them wait on MonitorManager.

phoenix-pherf/src/main/java/org/apache/phoenix/pherf/workload/WorkloadExecutor.java
codymarcel: Yep, I can remove synchronized.

phoenix-pherf/src/main/java/org/apache/phoenix/pherf/configuration/WriteParams.java
```diff
@@ -0,0 +1,72 @@
...
+        this.writerThreadCount = Integer.MIN_VALUE;
+        this.threadSleepDuration = Long.MIN_VALUE;
+        this.executionDurationInMs = Long.MAX_VALUE;
+    }
+
+    public long getThreadSleepDuration() {
+        return threadSleepDuration;
+    }
+
+    @SuppressWarnings("unused")
```
samarthjain: Not sure if I am a fan of @SuppressWarnings("unused") especially for setters. I would strive to make all member state final and get rid of the unused setter methods.

samarthjain: Same comment for setBatchSize(), setWriterThreadCount(), setExecutionDurationInMs()

phoenix-pherf/src/main/java/org/apache/phoenix/pherf/jmx/MonitorManager.java
```diff
@@ -51,13 +58,11 @@
     private volatile boolean shouldStop = false;
```
samarthjain: shouldStop doesn't need to be volatile. You are already synchronizing accesses to it. Better would be to make shouldStop final and an AtomicBoolean.

phoenix-pherf/src/main/java/org/apache/phoenix/pherf/jmx/MonitorManager.java
```diff
@@ -135,10 +155,8 @@ public synchronized boolean shouldStop() {
         return shouldStop;
     }
 
-    public synchronized void stop() {
-        this.shouldStop = true;
-    }
-
+    // Convenience method for testing.
+    @SuppressWarnings("unused")
     public synchronized long getRowCount() {
```
samarthjain: No extra synchronization needed here because rowCount is already AtomicLong. Unless the intent is to not let shouldStop, rowCount, isRunning get modified concurrently since you are guarding accesses to them by doing synchronizing on "this" .

phoenix-pherf/src/main/java/org/apache/phoenix/pherf/Pherf.java
codymarcel: We want the monitor to run through the duration of the workloads. If we call get here, it would block on the monitor until it completed then exit before the other workloads finish. 

I am glad you commented on this though. I just realized I am not calling complete() on the monitor at the end. So it can block indefinitely right now or leave the monitor dangling. I'll make that change.

phoenix-pherf/src/main/java/org/apache/phoenix/pherf/configuration/WriteParams.java
codymarcel: No choice here. The setters are being called by the xml bindings. You just don't see the references to them because they are called dynamically. They are being called though.

phoenix-pherf/src/main/java/org/apache/phoenix/pherf/jmx/MonitorManager.java
codymarcel: Atomic is better. I will change.

phoenix-pherf/src/main/java/org/apache/phoenix/pherf/result/DataModelResult.java
```diff
@@ -18,61 +18,57 @@
...
-	private List<ScenarioResult> scenarioResult = new ArrayList<ScenarioResult>();
-	private String zookeeper;
+    public List<ScenarioResult> getScenarioResult() {
+        return scenarioResult;
+    }
 
-	public List<ScenarioResult> getScenarioResult() {
-		return scenarioResult;
-	}
+    @SuppressWarnings("unused") public void setScenarioResult(List<ScenarioResult> scenarioResult) {
```
samarthjain: Same comment about using SuppressWarnings("unused") here. I would remove the method and make scenarioResult final.

phoenix-pherf/src/main/java/org/apache/phoenix/pherf/result/QuerySetResult.java
```diff
@@ -18,31 +18,30 @@
...
+    }
+
+    public QuerySetResult() {
+    }
+
+    public List<QueryResult> getQueryResults() {
+        return queryResults;
+    }
+
+    @SuppressWarnings("unused") public void setQueryResults(List<QueryResult> queryResults) {
```
samarthjain: I would make queryResults final and remove the method.

phoenix-pherf/src/main/java/org/apache/phoenix/pherf/result/DataModelResult.java
codymarcel: All the xml data bindings require the setters

phoenix-pherf/src/main/java/org/apache/phoenix/pherf/result/RunTime.java
```diff
@@ -18,104 +18,91 @@
...
-	@XmlAttribute()
-	public Long getResultRowCount() {
-		return resultRowCount;
-	}
-
-    @SuppressWarnings("unused")
-	public void setResultRowCount(Long resultRowCount) {
-		this.resultRowCount = resultRowCount;
-	}
+    private Date startTime;
```
samarthjain: I would make all these member variables final and remove the corresponding setters.

phoenix-pherf/src/main/java/org/apache/phoenix/pherf/result/ScenarioResult.java
```diff
@@ -18,31 +18,31 @@
...
-	}
-	
-	public ScenarioResult(Scenario scenario) {
-		this.setDataOverride(scenario.getDataOverride());
-		this.setPhoenixProperties(scenario.getPhoenixProperties());
-		this.setRowCount(scenario.getRowCount());
-		this.setTableName(scenario.getTableName());
-		this.setName(scenario.getName());
-	}
+    private List<QuerySetResult> querySetResult = new ArrayList<QuerySetResult>();
```
samarthjain: Make queryResult final and remove the setter method on it since it is unused.

phoenix-pherf/src/main/java/org/apache/phoenix/pherf/result/QuerySetResult.java
codymarcel: Good point. I think we can remove this one.

phoenix-pherf/src/main/java/org/apache/phoenix/pherf/result/DataModelResult.java
samarthjain: I see. In that case maybe remove the annotation then? Because it is a public method, I don't think IDEs will complain either that the method is unused.

phoenix-pherf/src/main/java/org/apache/phoenix/pherf/result/RunTime.java
samarthjain: Unless XML bindings need them. In that case I would remove the suppressedwarnings annotations from the setter methods.

phoenix-pherf/src/main/java/org/apache/phoenix/pherf/workload/WriteWorkload.java
```diff
@@ -0,0 +1,403 @@
...
+                        PreparedStatement stmt = connection.prepareStatement(sql);
+                        stmt = buildStatement(scenario, columns, stmt, simpleDateFormat);
+                        start = System.currentTimeMillis();
+                        rowsCreated += stmt.executeUpdate();
+                        stmt.close();
+                        if ((i % getBatchSize()) == 0) {
+                            connection.commit();
+                            duration = System.currentTimeMillis() - start;
+                            logger.info("Writer (" + Thread.currentThread().getName()
+                                    + ") committed Batch. Total " + tableName
```
samarthjain: Did you mean getBatchSize() and not tableName here?

phoenix-pherf/src/main/java/org/apache/phoenix/pherf/result/DataModelResult.java
codymarcel: James barked at them the last time I left it off.

phoenix-pherf/src/main/java/org/apache/phoenix/pherf/workload/WriteWorkload.java
```diff
@@ -0,0 +1,403 @@
...
+                    resultUtil.write(dataLoadThreadTime);
+
+                } catch (Exception e) {
+                    logger.warn("", e);
+                }
+            }
+        };
+    }
+
+    private synchronized void exec(DataLoadTimeSummary dataLoadTimeSummary,
```
samarthjain: No synchronization needed here since no state is being guarded unless you want to prevent more than one thread from executing exec which I am not sure you really need.

phoenix-pherf/src/main/java/org/apache/phoenix/pherf/workload/WriteWorkload.java
```diff
@@ -0,0 +1,403 @@
...
+            this.scenario = null;
+            threadSleepDuration = 0;
+        }
+
+        int size = Integer.parseInt(properties.getProperty("pherf.default.dataloader.threadpool"));
+
+        this.threadPoolSize = (size == 0) ? Runtime.getRuntime().availableProcessors() : size;
+
+        // TODO Move pool management up to WorkloadExecutor
+        this.pool = Executors.newFixedThreadPool(this.threadPoolSize);
```
samarthjain: Is the intent here to have a separate pool for write work load? I am not sure if the MonitorManager is running on this pool too. Would be good to monitor write activity too.

stoty:
Already merged as PHOENIX-1920

PR #98, opened 2015-07-10
PHOENIX-2098 - Udf that given a number bulk reserves sequences.

siddhimehta:


siddhimehta:
@JamesRTaylor  Pull request for PHOENIX-2098

phoenix-pig/src/main/java/org/apache/phoenix/pig/udf/ReserveNSequence.java
```diff
@@ -0,0 +1,88 @@
...
+    public static final String EMPTY_ZK_MESSAGE = "ZKQuorum should be not null";
+    public static final String INVALID_NUMBER_MESSAGE = "NUmber of Sequences to Reserve should be greater than 0";
+    public static final String SEQUENCE_NAME_CONF_KEY = "phoenix.sequence.name";
+
+    /**
+     * Reserve N next sequences for a sequence name. N is the first field in the tuple. Sequence name is the second
+     * field in the tuple zkquorum is the third field in the tuple
+     */
+    @Override
+    public Long exec(Tuple input) throws IOException {
```
siddhimehta: UDF that taken in tuple with number,sequenceName,ZkEndpoint and bulk reserves the sequence

phoenix-pig/src/it/java/org/apache/phoenix/pig/udf/ReserveNSequenceTestIT.java
```diff
@@ -0,0 +1,284 @@
...
+import org.junit.Rule;
+import org.junit.Test;
+import org.junit.experimental.categories.Category;
+import org.junit.rules.ExpectedException;
+
+/**
+ * Test class to run all the Pig Sequence UDF integration tests against a virtual map reduce cluster.
+ */
+@Category(NeedsOwnMiniClusterTest.class)
+public class ReserveNSequenceTestIT {
```
JamesRTaylor: Minor item: please take at Ravi's latest changes for BaseHBaseManagedTimeIT [here](https://git-wip-us.apache.org/repos/asf?p=phoenix.git;a=commit;h=984e62223c8aa507e9c044cecfc7fc92ffa42522). It'd be good if you did that as well - it's easier to manage tests that derive from our base classes as the framework changes. Also, they run faster if you don't need to spin up your own cluster for each test.

phoenix-pig/src/main/java/org/apache/phoenix/pig/udf/ReserveNSequence.java
```diff
@@ -0,0 +1,88 @@
...
+ * 
+ * @note The way this UDF is invoked we open a new connection for every tuple row. The UDF will not perform well on
+ *       large datasets as it involves creating a new connection for every tuple row
+ */
+public class ReserveNSequence extends EvalFunc<Long> {
+
+    public static final String INVALID_TUPLE_MESSAGE = "Tuple should have correct fields(NumtoReserve,SequenceName,zkquorum.";
+    public static final String EMPTY_SEQUENCE_NAME_MESSAGE = "Sequence name should be not null";
+    public static final String EMPTY_ZK_MESSAGE = "ZKQuorum should be not null";
+    public static final String INVALID_NUMBER_MESSAGE = "NUmber of Sequences to Reserve should be greater than 0";
```
JamesRTaylor: Typo: NUmber -> Number

JamesRTaylor:
Looks very good, @siddhimehta. Thanks for the quick turnaround. Couple of minor nits and then I'll commit it on your behalf.

phoenix-pig/src/it/java/org/apache/phoenix/pig/udf/ReserveNSequenceTestIT.java
siddhimehta: @JamesRTaylor  Sounds good.Updating the tests.

phoenix-pig/src/main/java/org/apache/phoenix/pig/udf/ReserveNSequence.java
```diff
@@ -0,0 +1,88 @@
...
+import org.apache.pig.EvalFunc;
+import org.apache.pig.data.Tuple;
+import org.apache.pig.impl.util.UDFContext;
+
+import com.google.common.base.Preconditions;
+
+/**
+ * UDF to Reserve a chunk of numbers for a given sequence
+ * 
+ * @note The way this UDF is invoked we open a new connection for every tuple row. The UDF will not perform well on
```
anilgupta84: I am not an expert at Pig or its UDF's. 
Isnt there anyway to pass Connection object in UDF rather than creating it for every row? If you cant pass Connection, can you make Connection an instance variable of UDF class?
Is creating new connection only reason for its non-scalability?
@JamesRTaylor Do you think, this non-scalability might lead to bad user experience in production clusters?

JamesRTaylor: Prashant would know if that's possible, I suspect, but I'm not sure (as you're definitely more of a Pig expert than me). @mravi may know too. It's not too bad from a Phoenix perspective, as we cache the underlying HConnection, so making a new connection (on the same JVM) is cheap.

anilgupta84: i c.. In that case, this UDF is scalable? Or It is non-scalable due to some other reason?

JamesRTaylor: It should be alright, but might be able to be improved. As always, we'll want to perf test this and the rest of the Phoenix/Pig loader integration.

siddhimehta: @JamesRTaylor I corrected the nits. 
@anilgupta84 I can investigate more to see if its possible to pass in the connection object to the UDF. If  we make can make Connection an instance variable of UDF class how do we ensure that the connection is correctly released/closed after all the rows.

stoty:
This has already been merged.

PR #104, opened 2015-08-05
PHOENIX-1673 Allow TenantId to be of any integral data type

JeffreyLyonsD2L:
Hopefully everything here is in order, if there are any questions or requests don't hesitate to ask.

### DIFF
--- a/phoenix-core/src/main/java/org/apache/phoenix/compile/WhereOptimizer.java
+++ b/phoenix-core/src/main/java/org/apache/phoenix/compile/WhereOptimizer.java
@@ -99,21 +99,23 @@ public class WhereOptimizer {
      * @param whereClause the where clause expression
      * @return the new where clause with the key expressions removed
      */
-    public static Expression pushKeyExpressionsToScan(StatementContext context, FilterableStatement statement, Expression whereClause) {
+    public static Expression pushKeyExpressionsToScan(StatementContext context, FilterableStatement statement, Expression whereClause)
+            throws SQLException{
         return pushKeyExpressionsToScan(context, statement, whereClause, null);
     }
 
     // For testing so that the extractedNodes can be verified
     public static Expression pushKeyExpressionsToScan(StatementContext context, FilterableStatement statement,
-            Expression whereClause, Set<Expression> extractNodes) {
+            Expression whereClause, Set<Expression> extractNodes) throws SQLException {
         PName tenantId = context.getConnection().getTenantId();
+        byte[] tenantIdBytes = null;
         PTable table = context.getCurrentTable().getTable();
     	Integer nBuckets = table.getBucketNum();
     	boolean isSalted = nBuckets != null;
     	RowKeySchema schema = table.getRowKeySchema();
     	boolean isMultiTenant = tenantId != null && table.isMultiTenant();
     	if (isMultiTenant) {
-    		tenantId = ScanUtil.padTenantIdIfNecessary(schema, isSalted, tenantId);
+            tenantIdBytes = ScanUtil.getTenantIdBytes(schema, isSalted, tenantId);
     	}
 
         if (whereClause == null && (tenantId == null || !table.isMultiTenant()) && table.getViewIndexId() == null) {
@@ -167,7 +169,7 @@ public class WhereOptimizer {
         boolean hasViewIndex = table.getViewIndexId() != null;
         if (hasMinMaxRange) {
             int minMaxRangeSize = (isSalted ? SaltingUtil.NUM_SALTING_BYTES : 0)
-                    + (isMultiTenant ? tenantId.getBytes().length + 1 : 0) 
+                    + (isMultiTenant ? tenantIdBytes.length + 1 : 0)
                     + (hasViewIndex ? MetaDataUtil.getViewIndexIdDataType().getByteSize() : 0);
             minMaxRangePrefix = new byte[minMaxRangeSize];
         }
@@ -188,7 +190,6 @@ public class WhereOptimizer {
         
         // Add tenant data isolation for tenant-specific tables
         if (isMultiTenant) {
-            byte[] tenantIdBytes = tenantId.getBytes();
             KeyRange tenantIdKeyRange = KeyRange.getKeyRange(tenantIdBytes);
             cnf.add(singletonList(tenantIdKeyRange));
             if (hasMinMaxRange) {
--- a/phoenix-core/src/main/java/org/apache/phoenix/exception/SQLExceptionCode.java
+++ b/phoenix-core/src/main/java/org/apache/phoenix/exception/SQLExceptionCode.java
@@ -235,7 +235,8 @@ public enum SQLExceptionCode {
             
     CANNOT_CREATE_TENANT_SPECIFIC_TABLE(1030, "42Y89", "Cannot create table for tenant-specific connection"),
     DEFAULT_COLUMN_FAMILY_ONLY_ON_CREATE_TABLE(1034, "42Y93", "Default column family may only be specified when creating a table."),
-    INSUFFICIENT_MULTI_TENANT_COLUMNS(1040, "42Y96", "A MULTI_TENANT table must have two or more PK columns with the first column being NOT NULL and of type VARCHAR or CHAR."),
+    INSUFFICIENT_MULTI_TENANT_COLUMNS(1040, "42Y96", "A MULTI_TENANT table must have two or more PK columns with the first column being NOT NULL."),
+    TENANTID_IS_OF_WRONG_TYPE(1041, "42Y97", "The TenantId could not be converted to correct format for this table."),
     VIEW_WHERE_IS_CONSTANT(1045, "43A02", "WHERE clause in VIEW should not evaluate to a constant."),
     CANNOT_UPDATE_VIEW_COLUMN(1046, "43A03", "Column updated in VIEW may not differ from value specified in WHERE clause."),
     TOO_MANY_INDEXES(1047, "43A04", "Too many indexes have already been created on the physical table."),
--- a/phoenix-core/src/main/java/org/apache/phoenix/execute/BaseQueryPlan.java
+++ b/phoenix-core/src/main/java/org/apache/phoenix/execute/BaseQueryPlan.java
@@ -204,8 +204,18 @@ public abstract class BaseQueryPlan implements QueryPlan {
         } else {
             ScanUtil.setTimeRange(scan, context.getScanTimeRange());
         }
-        
-        ScanUtil.setTenantId(scan, connection.getTenantId() == null ? null : connection.getTenantId().getBytes());
+        byte[] tenantIdBytes;
+        if( table.isMultiTenant() == true ) {
+            tenantIdBytes = connection.getTenantId() == null ? null :
+                    ScanUtil.getTenantIdBytes(
+                            table.getRowKeySchema(),
+                            table.getBucketNum()!=null,
+                            connection.getTenantId());
+        } else {
+            tenantIdBytes = connection.getTenantId() == null ? null : connection.getTenantId().getBytes();
+        }
+
+        ScanUtil.setTenantId(scan, tenantIdBytes);
         String customAnnotations = LogUtil.customAnnotationsToString(connection);
         ScanUtil.setCustomAnnotations(scan, customAnnotations == null ? null : customAnnotations.getBytes());
         // Set local index related scan attributes. 
--- a/phoenix-core/src/main/java/org/apache/phoenix/util/ScanUtil.java
+++ b/phoenix-core/src/main/java/org/apache/phoenix/util/ScanUtil.java
@@ -46,14 +46,16 @@ import org.apache.phoenix.compile.ScanRanges;
 import org.apache.phoenix.compile.StatementContext;
 import org.apache.phoenix.coprocessor.BaseScannerRegionObserver;
 import org.apache.phoenix.coprocessor.MetaDataProtocol;
-import org.apache.phoenix.execute.DescVarLengthFastByteComparisons;
-import org.apache.phoenix.filter.BooleanExpressionFilter;
+import org.apache.phoenix.exception.SQLExceptionCode;
+import org.apache.phoenix.exception.SQLExceptionInfo;
+import org.apache.phoenix.execute.DescVarLengthFastByteComparisons;import org.apache.phoenix.filter.BooleanExpressionFilter;
 import org.apache.phoenix.filter.SkipScanFilter;
 import org.apache.phoenix.query.KeyRange;
 import org.apache.phoenix.query.KeyRange.Bound;
 import org.apache.phoenix.query.QueryConstants;
 import org.apache.phoenix.query.QueryServices;
 import org.apache.phoenix.query.QueryServicesOptions;
+import org.apache.phoenix.schema.IllegalDataException;
 import org.apache.phoenix.schema.PName;
 import org.apache.phoenix.schema.PNameFactory;
 import org.apache.phoenix.schema.RowKeySchema;
@@ -536,7 +538,7 @@ public class ScanUtil {
         while (schema.next(ptr, pos, maxOffset) != null) {
             pos++;
         }
-        Field field = schema.getField(pos-1);
+        Field field = schema.getField(pos - 1);
         if (!field.getDataType().isFixedWidth()) {
             byte[] newLowerRange = new byte[key.length + 1];
             System.arraycopy(key, 0, newLowerRange, 0, key.length);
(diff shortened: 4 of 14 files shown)


phoenix-core/src/main/java/org/apache/phoenix/compile/WhereOptimizer.java
```diff
@@ -114,6 +116,7 @@ public static Expression pushKeyExpressionsToScan(StatementContext context, Filt
     	boolean isMultiTenant = tenantId != null && table.isMultiTenant();
     	if (isMultiTenant) {
     		tenantId = ScanUtil.padTenantIdIfNecessary(schema, isSalted, tenantId);
```
JamesRTaylor: Take a look at ScanUtil.padTenantIdIfNecessary(), as it assumes tenantId is a string:
    public static PName padTenantIdIfNecessary(RowKeySchema schema, boolean isSalted, PName tenantId) {
        int pkPos = isSalted ? 1 : 0;
        String tenantIdStr = tenantId.getString();
        Field field = schema.getField(pkPos);
        PDataType dataType = field.getDataType();
        boolean isFixedWidth = dataType.isFixedWidth();
        Integer maxLength = field.getMaxLength();
        if (isFixedWidth && maxLength != null) {
            if (tenantIdStr.length() < maxLength) {
                tenantIdStr = (String)dataType.pad(tenantIdStr, maxLength);
                return PNameFactory.newName(tenantIdStr);
            }
        }
        return tenantId;
    }

Probably best if your ScanUtil.getTenantIdBytes() function takes into account padding as well and we can get rid of this util function (see comment below on that).

phoenix-core/src/main/java/org/apache/phoenix/util/ScanUtil.java
```diff
@@ -657,6 +660,22 @@ public static PName padTenantIdIfNecessary(RowKeySchema schema, boolean isSalted
...
 
+    public static byte[] getTenantIdBytes(RowKeySchema schema, boolean isSalted, PName tenantId)
+            throws SQLException {
+        int pkPos = isSalted ? 1 : 0;
+        Field field = schema.getField(pkPos);
+        PDataType dataType = field.getDataType();
+        byte[] convertedValue;
+        try {
+            Object value = dataType.toObject(tenantId.getString());
+            convertedValue = dataType.toBytes(value);
```
JamesRTaylor: You can use the following code to pad the convertedValue (this will be a noop for types that aren't padded):

    ImmutableBytesWritable ptr = new ImmutableBytesWritable(convertedValue);
    dataType.pad(ptr, field.getMaxLength(), field.getSortOrder());
    convertedValue = ByteUtil.copyKeyBytesIfNecessary(ptr);

phoenix-core/src/main/java/org/apache/phoenix/util/ScanUtil.java
```diff
@@ -657,6 +660,22 @@ public static PName padTenantIdIfNecessary(RowKeySchema schema, boolean isSalted
...
     }
 
+    public static byte[] getTenantIdBytes(RowKeySchema schema, boolean isSalted, PName tenantId)
+            throws SQLException {
+        int pkPos = isSalted ? 1 : 0;
+        Field field = schema.getField(pkPos);
+        PDataType dataType = field.getDataType();
+        byte[] convertedValue;
+        try {
+            Object value = dataType.toObject(tenantId.getString());
```
JamesRTaylor: For this dataType.toBytes() call, use the one that accepts sortOrder so that it'll still work if the tenantId column is descending:

    convertedValue = dataType.toBytes(value, field.getSortOrder());

phoenix-core/src/test/java/org/apache/phoenix/util/TenantIdByteConversionTest.java
```diff
@@ -0,0 +1,275 @@
...
+import org.junit.runner.RunWith;
+import org.junit.runners.Parameterized;
+import org.junit.runners.Parameterized.Parameters;
+
+import com.google.common.collect.Lists;
+import org.mockito.Mockito;
+
+/*Test the getTenantIdBytes method in ScanUtil*/
+@RunWith(Parameterized.class)
+public class TenantIdByteConversionTest {
```
JamesRTaylor: Nice tests! I'd add a couple more for tenantId as BINARY data type and tenantId declared as DESC

JamesRTaylor:
Thanks so much for the contribution, @JeffreyLyonsD2L. Would you mind helping to review and commit this, @elilevine? I think the main thing to check is that there aren't any places in the code that Jeffrey missed updating.

elilevine:
Thanks for the ping, @JamesRTaylor. Haven't seen this PR until now. I'll review and commit today or tomorrow if everything looks good.

JeffreyLyonsD2L:
Added the changes requested, and removed the redundant padding.

phoenix-core/src/it/java/org/apache/phoenix/end2end/MultiTenantTableIT.java
```diff
@@ -0,0 +1,140 @@
...
+import org.apache.phoenix.exception.SQLExceptionCode;
+import org.apache.phoenix.schema.TableAlreadyExistsException;
+import org.apache.phoenix.util.PhoenixRuntime;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.junit.runners.Parameterized;
+import org.junit.runners.Parameterized.Parameters;
+
+@RunWith(Parameterized.class)
+public class MultiTenantTableIT extends BaseClientManagedTimeIT {
```
elilevine: Excellent tests, Jeffrey. Can you rename this class to something that is more specific to the work you did? Otherwise the name of this class implies that it is a full test suite for everything related to multi-tenant tables, while there are a number of other test classes for multi-tenancy. Maybe something like TenantIdTypeIT?

samarthjain: Tests should extend BaseClientManagedTimeIT when they want to manage timestamps themselves, for ex- tests dealing with snapshot schema/queries. Your test class doesn't seem to be doing that. So make sure to extend BaseHBaseManagedTimeIT instead.

phoenix-core/src/it/java/org/apache/phoenix/end2end/MultiTenantTableIT.java
```diff
@@ -0,0 +1,140 @@
...
+        testCases.add(new Object[] { "DECIMAL", "3.402823466", "3.402823465" });
+        testCases.add(new Object[] { "VARCHAR", "\'NameOfTenant\'", "\'Nemesis\'" });
+        testCases.add(new Object[] { "CHAR(10)", "\'1234567890\'", "\'Nemesis\'" });
+
+        return testCases;
+    }
+
+    @Test
+    public void testMultiTenantTables() throws Exception {
+        Connection conn = regularConnection(getUrl());
```
samarthjain: Include this in a try-with-resources construct to ensure connection is closed.

phoenix-core/src/main/java/org/apache/phoenix/exception/SQLExceptionCode.java
```diff
@@ -236,6 +236,7 @@ public SQLException newException(SQLExceptionInfo info) {
     CANNOT_CREATE_TENANT_SPECIFIC_TABLE(1030, "42Y89", "Cannot create table for tenant-specific connection"),
     DEFAULT_COLUMN_FAMILY_ONLY_ON_CREATE_TABLE(1034, "42Y93", "Default column family may only be specified when creating a table."),
     INSUFFICIENT_MULTI_TENANT_COLUMNS(1040, "42Y96", "A MULTI_TENANT table must have two or more PK columns with the first column being NOT NULL and of type VARCHAR or CHAR."),
```
elilevine: Type information at the end of the error message no longer applies.

phoenix-core/src/it/java/org/apache/phoenix/end2end/MultiTenantTableIT.java
```diff
@@ -0,0 +1,140 @@
...
+
+        String query = "upsert into " + table +
+                " values (" + tenantId + ", 1 , 'valid')";
+
+        conn.createStatement().execute("upsert into " + table +
+                " values (" + tenantId + ", 1 , 'valid')");
+        conn.createStatement().execute("upsert into " + table +
+                " values (" + otherTenantId + ", 2 , 'invalid')");
+
+        Connection conn2 = tenantConnection(getUrl());
```
samarthjain: Same as above - try-with-resources

phoenix-core/src/main/java/org/apache/phoenix/execute/BaseQueryPlan.java
```diff
@@ -204,8 +204,18 @@ public final ResultIterator iterator(final List<? extends SQLCloseable> dependen
         } else {
             ScanUtil.setTimeRange(scan, context.getScanTimeRange());
         }
-        
-        ScanUtil.setTenantId(scan, connection.getTenantId() == null ? null : connection.getTenantId().getBytes());
+        byte[] tenantIdBytes;
```
elilevine: Can you explain why you need the extra if/else here?

phoenix-core/src/it/java/org/apache/phoenix/end2end/MultiTenantTableIT.java
```diff
@@ -0,0 +1,140 @@
...
+
+    @Test
+    public void testMultiTenantTables() throws Exception {
+        Connection conn = regularConnection(getUrl());
+        conn.setAutoCommit(true);
+        conn.createStatement().execute(ddl);
+
+
+        try {
+            conn.createStatement().execute(ddl);
```
samarthjain: Is there a reason why you are calling conn.createStatement().execute(ddl) twice? Also no need to catch TableAlreadyExistsException since the constructor is called once per test method. Also the @After method in BaseHBaseManagedTimeIT takes care of dropping the table after every test method.

elilevine:
Good stuff, @JeffreyLyonsD2L. Added a few questions and comments to the PR.  

There is one more issue that it not tackled here that we need to at least discuss IMHO. Phoenix's metadata tables, such as SYSTEM.CATALOG lead with TENANT_ID VARCHAR column (take a look in QueryConstants.java). This works well in a world where all multi-tenant tables can only have VARCHAR tenant id columns. Your changes introduce the possibility that different multi-tenant tables will have different types of tenant id columns.

What's the right solution here? The right solution is probably changing TENANT_ID column type to BINARY in system tables. But that will make upgrades for existing clusters a pain. Leave them as VARCHAR?

phoenix-core/src/test/java/org/apache/phoenix/util/TenantIdByteConversionTest.java
elilevine: Also, can you test non-VARCHAR tenant ids with tenant-specific secondary indexes and sequences?

phoenix-core/src/main/java/org/apache/phoenix/execute/BaseQueryPlan.java
JeffreyLyonsD2L: This is for a case where you have a tenant-specific connection set up and are accessing a table that is not multi-tenant, but it has a different type than your tenant ids. Without this check, the tenantId fails to convert and would prevent us from reading the table even though it is unrestricted.

phoenix-core/src/it/java/org/apache/phoenix/end2end/MultiTenantTableIT.java
JeffreyLyonsD2L: So I think this is convention I picked up from some other tests, but the goal of that try block is to actually throw the exception and then catch it and carry on. It seemed like the way that other tests were verifying table creation was to try creating it again and verifying that the second create fails because it already exists. I do want to verify the table got created, because prior to my changes a tenantId of anything but CHAR or VARCHAR would fail at creation. If there is a better way of checking if a table exists, I would certainly be open to it though.

phoenix-core/src/main/java/org/apache/phoenix/execute/BaseQueryPlan.java
elilevine: I guess the question is why do we need to grab tenantIdBytes from the table (in the if {} block) now instead of just getting them from Connection.getTenantId().getBytes() like we did before?

JeffreyLyonsD2L:
Thanks for all the feedback @elilevine and @samarthjain! I'm east coast, so I'm just heading out of the office now, but I should be able to get the changes up sometime tomorrow.

The above question kind of goes outside my realm of interaction with Phoenix so I'm not sure if my handle on it is correct. Is the point of the TenantId in a table like System.Catalog to tie specific schema to a tenant? If so I would lean toward leaving as VARCHAR since any TenantId has to be at least convertible to a VARCHAR since it comes down the connection as one, though I agree the your 'correct' solution seems like the most complete way forward.

elilevine:
The TENANT_ID column in system tables is used to quickly filter metadata belonging to a tenant. Agree we could probably get away with leaving it VARCHAR because that is how tenant ids are passed to connections anyway.

Now that got me thinking... if we extending that logic further, why then do we need to implement support for other data types for tenant id columns if tenant ids are always passed in as strings to connections? Why not keep them all as VARCHAR/CHAR, like they are now?

JamesRTaylor:
This feature would allow the tenant ID to be represented through more types than just a VARCHAR. Some users may have tenants identified by an int or a long instead of a string. The tenant ID would still be passed as the stringified value, but the underlying table could use an int or a long as the column type. As you mentioned, the system tables (SYSTEM.CATALOG and SYSTEM.SEQUENCE) would need to continue to use the string representation. Every type can be converted to a string, so this should be ok (pending testing).

elilevine:
Yeah, I understand that the feature allows TENANT_ID on data tables to be a different type. What I'm not sure about is how useful this feature would be in real life if internally we are treating tenant ids as strings in connections and system tables anyway. Why not just do the same for data tables, as well (which is how it is now implemented)? As you mentioned, every type can be converted to a string and users can do easily do that for data tables. 

If we relax type constrains for data tables but not for metadata tables/connections we are making the implementation of multi-tenancy in phoenix internally inconsistent. Should be all or nothing IMHO.

JamesRTaylor:
The system tables need to be able to store the tenant ID regardless of its type in the data table, so it wouldn't make sense to type the system tables (i.e. the *all* approach isn't viable IMHO). If a table already exists with a different typed leading column than CHAR/VARCHAR, then it would not be possible to map a multi-tenant Phoenix table to it. This just provides a little bit more flexibility. I'll ping the original filer of the JIRA and see if he has further comment on his use case.

phoenix-core/src/main/java/org/apache/phoenix/execute/BaseQueryPlan.java
JeffreyLyonsD2L: In both cases we are getting it from the connection. In the if {} block we are using the table to figure out the RowKeySchema so we know what to convert it into.

The else might be unnecessary, I leaned towards keeping it in because then it mimicked the old behaviour in the case where you were scanning a regular table on a tenant-specific connection.

JeffreyLyonsD2L:
I can probably provide a little more information about our use case once I meet back with my team on Monday, but I can definitely say that James' point about being able to map existing tables with the concept of a tenantId without having to convert all of the data would be a big plus for us.

elilevine:
Everything looks good in terms of code. Excellent work, @JeffreyLyonsD2L. The only thing to figure out now is whether or not we should add the feature at all. I agree that changing the schema of SYSTEM.* tables is off the table. So do we allow the tenant id column in data tables to be of different type than in metadata tables? It does not sit well with me and makes multi-tenancy seem inconsistent IMHO, as I mentioned above, since tenant id is always passed in as string to connections and is stored as such in system tables.

@JeffreyLyonsD2L, maybe you can shine some light on our use-case. You mentioned that not having to do data conversion is a plus. In what way? Are you starting with existing data and don't want to do data conversion?

JeffreyLyonsD2L:
Hey @elilevine, thanks and I appreciate the review, I've been learning a lot :)

In terms of our use case, we have a pre-existing schema that we don't want to modify. For one, it would be a significant refactor to some of our other code in order to handle converting our existing TenantIds to strings. 

The other issue is that we have a lot of traffic that is interacting with the cluster directly through HBase and not through Phoenix. For this reason we like having the TenantIds as fixed length, so that the rowkeys are easy and logical to parse.

It also seems reasonable to me that if the TenantId is naturally a number instead of a name that it could be represented with one, instead of forcing a conversion to string when we store it.

elilevine:
Thanks for the details, @JeffreyLyonsD2L. I have to say I'm still torn on whether or not this Jira makes much sense. What if I told you that your concern #2 is already addressed? Currently tenant id columns can be either VARCHAR or fixed-length CHAR. Does that make it easier for you to use multi-tenancy in Phoenix?

Regarding your points #1 and 3, you are forced to do the conversion once per Connection when you set the TenantId parameter either way. Not sure what kind of conversion you mean beyond that. Normally when you read/write via a tenant-specific connection you only set TenantId on the Connection and don't need to actually write it or read it. Are you working with this data over non-tenant-specific connections?
---
I went though your code again. Everything looks good. The only thing missing now is:
1. Tests that create tenant views with non-string tenant id columns and test data upsertion/retrieval.
2. Same as above but negative. Test what happens when conversion fails.

JeffreyLyonsD2L:
For points #1 and #3, I think I really mean converting the data we already have. We have a lot of data in our system with our current schema and would have to upgrade the entire schema and most of our historical data to fit the new standards, which for us would be a non-trivial process.

We are also writing to the underlying tables through our own library, not through Phoenix, and that system is set up to have TenantId as an UNSIGNED_INT. This code is pretty ingrained in our system, and would require a massive refactor and significant restructure of how our cluster works to switch over to using something like fixed-length CHAR arrays.

As for the tests, I'll get them up as soon as possible.

elilevine:
Makes sense, @JeffreyLyonsD2L. The existing data and integrations with other components interacting with the HBase tables do make the feature worth it IMHO. I'll commit after you add those tests. Thanks!

JeffreyLyonsD2L:
Thanks @elilevine, it has been a very interesting discussion for me to see :)

With regards to the tests you mentioned earlier, I've added them to the TenantIdTypeIT. Although in looking over the review, I saw I missed one of your previous comments regarding testing tenant-specific indexes and sequences, so I'll add some of those this afternoon.

I'm fairly new to committing to open source, so is there anything you need from me organizationally or git-wise once the final code changes are in (something like going through and resolving the conflicts, or rebasing)?

JeffreyLyonsD2L:
Added the last few tests.

JeffreyLyonsD2L:
Alright @elilevine, final revisions to the tests are up. As I mentioned in my earlier post, I'm not quite sure where my work ends, but I'll be available to fix/help with anything needed for the next few days.

elilevine:
Good stuff, @JeffreyLyonsD2L! One last thing: can you update this pull request so it merges with Phoenix master cleanly (there are some conflicts now)? Also, would be great to see a PR for one of the 4.x branches if this patch does not merge cleanly there. You don't need to do it for each 4.x branch, one is fine. I'll take care of the rest. Again, thank you for your contributions.

JeffreyLyonsD2L:
You're very welcome, I've found the process streamlined and would be encouraged to contribute again! I resolved the conflicts on this branch, and created another PR for the 4.x-HBase-0.98 branch (#108 for reference)

I'll be available fore help if there is anything else needed.

elilevine:
This PR is now in master and can be closed.

PR #105, opened 2015-08-06
PHOENIX-1791. Adding ability for Pherf Scenarios to write to a multi-tenant view.

jfernandosf:
Small change to Pherf to enable Pherf to allow Scenario declarations to specify multi-tenant views for the tableName attribute and then load data into the tenant specific view.  This change is  to add the tenantId attribute to the scenario XML. If a view is specified in the tableName and a tenantId attribute is specified Pherf will use a tenant specific connection to write to the view.

In the code we pass this tenantId when we get a connection. If it is not specified null is returned and therefore Pherf gets a standard connection when working with that Scenario. If it is specified then Pherf gets a multi-tenant connection for that scenario and then a multi-tenant view can be specified as value for the tableName attribute for the scenario.


phoenix-pherf/src/it/java/org/apache/phoenix/pherf/DataIngestIT.java
```diff
@@ -144,4 +148,52 @@ public void testRWWorkload() throws Exception {
...
+            @Override public Integer handle(ResultSet resultSet, Statement statement)
+                    throws SQLException {
+                while (resultSet.next()) {
+                    return resultSet.getInt(1);
+                }
+                return null;
+            }
+        });
+        assertNotNull("Could not retrieve count. " + count);
+        assertTrue("Could not query any rows for in " + scenario.getTableName(), count > 0);
```
codymarcel: This assertion is redundant with the one below. The next one is better. I can remove this one on commit.

stoty:
Already merged

PR #110, opened 2015-08-14
PHOENIX-2023 Build tar.gz only on release profile

gliptak:
The release build breaks without the license headers included.

### DIFF
--- a/phoenix-assembly/pom.xml
+++ b/phoenix-assembly/pom.xml
@@ -86,40 +86,6 @@
               </descriptors>
             </configuration>
           </execution>
-          <execution>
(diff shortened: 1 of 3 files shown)


gliptak:
@ndimiduk Would you like to see additional changes? Thanks

ndimiduk:
I haven't done a phoenix release, but this looks like what I had in mind. Thanks for fixing the missing headers while you're in there -- these should be caught by the apache:rat plugin; i'm not sure why they're not.

@JamesRTaylor and @chrajeshbabu have done phoenix releases recently. Does this look good by you guys?

PR #115, opened 2015-09-01
PHOENIX-2182

mujtabachohan:


phoenix-pherf/src/main/java/org/apache/phoenix/pherf/Pherf.java
```diff
@@ -76,6 +78,8 @@
         options.addOption("d", "debug", false, "Put tool in debug mode");
         options.addOption("stats", false,
                 "Update Phoenix Statistics after data is loaded with -l argument");
+		options.addOption("label", true, "Label a run. Result file name will be suffixed with specified label");
```
codymarcel: Is this an override to then dynamically generated label? Also, it would be good if the label was applied as a column value in the result data. That way we can distinguish between runs on analysis.

phoenix-pherf/src/main/java/org/apache/phoenix/pherf/Pherf.java
```diff
@@ -136,6 +142,8 @@ public Pherf(String[] args) throws Exception {
                 command.getOptionValue("writerThreadSize",
                         properties.getProperty("pherf.default.dataloader.threadpool"));
         properties.setProperty("pherf. default.dataloader.threadpool", writerThreadPoolSize);
+        label = command.getOptionValue("label", null);
+        compareResults = command.getOptionValue("compare", null);
```
codymarcel: default to false instead of null

phoenix-pherf/src/main/java/org/apache/phoenix/pherf/Pherf.java
```diff
@@ -179,6 +188,15 @@ public void run() throws Exception {
                 }
                 return;
             }
+
+            // Compare results and exit  
+			if (null != compareResults) {
```
codymarcel: if(!compareResults){}

phoenix-pherf/src/main/java/org/apache/phoenix/pherf/Pherf.java
codymarcel: Use boolean and set to false as default

phoenix-pherf/src/main/java/org/apache/phoenix/pherf/Pherf.java
codymarcel: if(!compareResults){}

phoenix-pherf/src/main/java/org/apache/phoenix/pherf/Pherf.java
```diff
@@ -179,6 +188,15 @@ public void run() throws Exception {
...
+				logger.info("\nStarting to compare results and exiting for " + compareResults);
+				ResultsComparator rc = new ResultsComparator(compareResults);
```
codymarcel: Should this be the label?

phoenix-pherf/src/main/java/org/apache/phoenix/pherf/PherfConstants.java
```diff
@@ -62,6 +62,8 @@
     public static final int MONITOR_FREQUENCY = 5000;
     public static final String MONITOR_FILE_NAME = "STATS_MONITOR";
 
+    public static final String SUMMARY_HTML_FILE_NAME = "summary.html";
```
codymarcel: Can you put this in the properties file?

codymarcel: I am trying to make everything here configurable from properties.

phoenix-pherf/src/main/java/org/apache/phoenix/pherf/result/ResultUtil.java
```diff
@@ -147,9 +147,9 @@ public String getSuffix() {
         if (null == FILE_SUFFIX) {
             Date date = new Date();
             Format formatter = new SimpleDateFormat("YYYY-MM-dd_hh-mm-ss");
-            FILE_SUFFIX = "_" + formatter.format(date);
+            FILE_SUFFIX = formatter.format(date);
```
codymarcel: Should you call setFileSuffix()?

phoenix-pherf/src/main/java/org/apache/phoenix/pherf/util/PhoenixUtil.java
```diff
@@ -274,4 +274,34 @@ public void updatePhoenixStats(String tableName, Scenario scenario) throws Excep
...
+     * @param query
+     * @return
+     * @throws SQLException
+     */
+    public String getExplainPlan(Query query) throws SQLException {
+        Connection conn = null;
+        ResultSet rs = null;
+        PreparedStatement statement = null;
+        StringBuilder buf = new StringBuilder();
+        try {
```
codymarcel: You might want to consider using the JdbcSession() to handle all the boiler plate try/catch/finally stuff on the connection. 

Check out testRWWorkload()
https://github.com/apache/phoenix/blob/master/phoenix-pherf/src/it/java/org/apache/phoenix/pherf/DataIngestIT.java

phoenix-pherf/src/main/java/org/apache/phoenix/pherf/util/ResultsComparator.java
```diff
@@ -0,0 +1,353 @@
...
+import org.apache.commons.csv.CSVFormat;
+import org.apache.commons.csv.CSVParser;
+import org.apache.commons.csv.CSVRecord;
+import org.apache.phoenix.pherf.PherfConstants;
+import org.apache.phoenix.pherf.result.file.ResultFileDetails;
+
+/**
+ * Compare results based on set threshold and render results as Google Charts
+ */
+public class ResultsComparator {
```
codymarcel: Should this implement Comparable? Calling it Comparator is misleading if it's not doing that.

phoenix-pherf/src/main/java/org/apache/phoenix/pherf/util/ResultsComparator.java
```diff
@@ -0,0 +1,353 @@
...
+
+	private String[] labels;
+	private final Map<String, DataNode> datanodes = new TreeMap<String, DataNode>();
+	private final PherfConstants constants = PherfConstants.create();
+	private final String resultDir = constants.getProperty("pherf.default.results.dir");
+	private final double threshold = Double.parseDouble(constants.getProperty("pherf.default.comparison.threshold"));
+
+	public ResultsComparator(String labels) {
```
codymarcel: nit: label should be singular ;)

phoenix-pherf/src/main/java/org/apache/phoenix/pherf/util/ResultsComparator.java
```diff
@@ -0,0 +1,353 @@
...
+	
+	String[] getLabels() {
+		return labels;
+	}
+
+	void setLabels(String[] labels) {
+		this.labels = labels;
+	}
+
+	void setLabelsAsString(String labels) {
```
codymarcel: This is setting as array. Why not just call it setLabels() and let overloading do it's thing?

phoenix-pherf/src/main/java/org/apache/phoenix/pherf/util/ResultsComparator.java
```diff
@@ -0,0 +1,353 @@
...
+    	}
+    	return true;
+    }
+    
+    /**
+     * Render results as Google charts
+     * @throws FileNotFoundException
+     * @throws UnsupportedEncodingException
+     */
+    private void renderAsGoogleChartsHTML() throws FileNotFoundException, UnsupportedEncodingException {
```
codymarcel: Should all this chart stuff go in a util object? It seems unrelated to a Result comparison.

phoenix-pherf/src/main/java/org/apache/phoenix/pherf/util/ResultsComparator.java
```diff
@@ -0,0 +1,353 @@
...
+			e.printStackTrace();
+		}
+	}
+
+	/**
+	 * Reads aggregate file and convert it to DataNode 
+	 * @param label
+	 * @throws Exception
+	 */
+    private void read(String label) throws Exception {
```
codymarcel: Can we use a CSVResultHandler with a ResultFileDetails.CSV_AGGREGATE_PERFORMANCE? It looks like there is a fair amount of redundancy with that and the ResultWriter. It makes sense that you need to wrap the compare logic into this, but result reading and writing should use the existing framework for that.

phoenix-pherf/src/main/java/org/apache/phoenix/pherf/util/ResultsComparator.java
```diff
@@ -0,0 +1,353 @@
...
+    		this.setTenantId(tenantId);
+    		this.setLabel(label);
+    		this.setRowCount(rowCount);
+    	}
+    	
+		String getExplainPlan() {
+			return explainPlan;
+		}
+		String getExplainPlanAsHTML() {
+			return "</br><font face=arial size=1><b>EXPLAIN PLAN </b>"
```
codymarcel: Perhaps we should use an html generation library instead of hard coding all this in strings?

Never tried this one, bit looks like it has god template support.
http://freemarker.org/docs/dgui_quickstart_basics.html

phoenix-pherf/src/main/java/org/apache/phoenix/pherf/workload/MultithreadedDiffer.java
```diff
@@ -51,7 +53,7 @@ private void diffQuery() throws Exception {
         Date startDate = Calendar.getInstance().getTime();
         String newCSV = queryVerifier.exportCSV(query);
         boolean verifyResult = queryVerifier.doDiff(query, newCSV);
-        String explainPlan = queryVerifier.getExplainPlan(query);
```
codymarcel: Could you add the explainPlan stuff as a different Jira/CL? It might get checked in faster if it's detached from this CL. This one might need a little more love.

stoty:
Already merged.

PR #116, opened 2015-09-09
PHOENIX-2238 Support more delimiter when using CsvBulkImport

979969786:
Add a function to support  more delimiter when using CsvBulkImport, such as \001, \t, \n.


gabrielreid:
I think that this is an interesting idea, but did you see my comment on [PHOENIX-2238](https://issues.apache.org/jira/browse/PHOENIX-2238)? There is support built in to the shell which allows you to supply non-printable characters.

I can see how this might be more user-friendly for some, while being confusing for others (who are already familiar with the standard method of providing non-printable characters on the command line). 

Another concern that I have is that providing these kinds of literals that start with backslashes also require some understanding about the rules of escaping when supplying command line parameters, so this kind of makes me think that it might be better to go completely for expecting users to know how to work with the shell properly.

What do you think?

PR #117, opened 2015-09-28
Phoenix 572 Truncate CHAR with max length on upsert

d9liang:
This transaction truncate string to the CHAR max length value on upsert. A test case is added and passed. Please review.


phoenix-core/src/main/java/org/apache/phoenix/compile/UpsertCompiler.java
```diff
@@ -787,15 +789,26 @@ public MutationState execute() throws SQLException {
                                 constantExpression.getDataType(), column.getDataType(), "expression: "
                                         + constantExpression.toString() + " in column " + column);
                         }
-                        if (!column.getDataType().isSizeCompatible(ptr, value, constantExpression.getDataType(),
+                        if ((column.getDataType() != PChar.INSTANCE) && // Char is truncated later
+                                !column.getDataType().isSizeCompatible(ptr, value, constantExpression.getDataType(),
```
JamesRTaylor: Is there a reason we wouldn't want to truncate the value for *all* fixed width types (i.e. column.getDataType().isFixedWidth)? In general, we try not to match on a particular type, but rather on categories of types. That way, when a new type gets added, it behaves a bit better.

phoenix-core/src/test/java/org/apache/phoenix/query/UpsertQueryTest.java
```diff
@@ -0,0 +1,92 @@
...
+import org.junit.AfterClass;
+import org.junit.BeforeClass;
+import org.junit.Test;
+
+/**
+ * Tests for Upsert Statement Compilation
+ * testUpsertTruncChar() String should be truncated to max CHAR size
+ */
+
+public class UpsertQueryTest extends BaseTest {
```
JamesRTaylor: Extend from BaseHBaseTimeManagedIT instead of BaseTest directly.

phoenix-core/src/test/java/org/apache/phoenix/query/UpsertQueryTest.java
```diff
@@ -0,0 +1,92 @@
...
+        assertTrue(rs.next());
+        assertEquals(expectedStr, rs.getString(1));
+        assertEquals(charSize, rs.getInt(2));
+
+        stmt = "DROP TABLE " + testTable;
+        conn.createStatement().execute(stmt);
+        conn.close();
+    }
+
+    @BeforeClass
```
JamesRTaylor: Both the doSetup() and doTeardown() overrides can be removed once you derived from BaseHBaseTimeManagedIT.

JamesRTaylor:
Thanks for the patch, @d9liang  -  I reviewed and left you some minor feedback. Overall, it looks very good. One other thing to check (and add tests for) is other code paths that insert data: namely UPSERT SELECT (on client-side and server-side) and CSV bulk load (which goes through a similar code path as UPSERT VALUES, so you might already be covered there, not sure). This would ensure that we're consistently truncating fixed width types in all cases.

PR #118, opened 2015-10-02
PHOENIX-2256: Correct two bugs in PMetaDataImpl

chiastic-security:
PHOENIX-2256: More entries than necessary could end up being evicted.

This PR also adds a new test to catch one of the bugs.

Previously, one of the bugs (`>` should have been `>=`) was being picked up on Java 8, but not Java 7, because the traversal ordering of one of the JDK collections happened to have changed. The bug was still present in Java 7, but the collection ordering meant that the test didn't pick it up.

The second bug (`if` should have been `while`) is picked up by the new test on both Java 7 and Java 8. The essence of the bug is that if several things are queued for eviction, and something large and old gets added to the eviction queue, the implementation was only considering removing one element from the eviction queue, whereas it should have removed as many as possible from the eviction queue while still leaving enough to clear space for the new element being added to the table.

### DIFF
--- a/phoenix-core/src/main/java/org/apache/phoenix/schema/PMetaDataImpl.java
+++ b/phoenix-core/src/main/java/org/apache/phoenix/schema/PMetaDataImpl.java
@@ -160,7 +160,7 @@ public class PMetaDataImpl implements PMetaData {
                     newCache.put(tableRef.table.getKey(), new PTableRef(tableRef));
                     toRemove.add(tableRef);
                     toRemoveBytes += tableRef.estSize;
-                    if (toRemoveBytes - toRemove.peekLast().estSize > overage) {
+                    while (toRemoveBytes - toRemove.peekLast().estSize >= overage) {
                         PTableRef removedRef = toRemove.removeLast();
                         toRemoveBytes -= removedRef.estSize;
                     }
--- a/phoenix-core/src/test/java/org/apache/phoenix/schema/PMetaDataImplTest.java
+++ b/phoenix-core/src/test/java/org/apache/phoenix/schema/PMetaDataImplTest.java
@@ -104,6 +104,25 @@ public class PMetaDataImplTest {
         assertEquals(2, metaData.size());
         assertNames(metaData, "d","e");
     }
+
+    @Test
+    public void shouldNotEvictMoreEntriesThanNecessary() throws Exception {
+        long maxSize = 5;
+        PMetaData metaData = new PMetaDataImpl(5, maxSize, new TestTimeKeeper());
+        metaData = addToTable(metaData, "a", 1);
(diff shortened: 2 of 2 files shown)


phoenix-core/src/main/java/org/apache/phoenix/schema/PMetaDataImpl.java
```diff
@@ -160,7 +160,7 @@ public PMetaDataCache cloneMinusOverage(long overage) {
                     newCache.put(tableRef.table.getKey(), new PTableRef(tableRef));
                     toRemove.add(tableRef);
                     toRemoveBytes += tableRef.estSize;
-                    if (toRemoveBytes - toRemove.peekLast().estSize > overage) {
+                    while (toRemoveBytes - toRemove.peekLast().estSize >= overage) {
```
JamesRTaylor: One minor nit - we should guard against toRemove being empty.

chiastic-security: I did think about that but I don't think it's possible for it to be empty. If you hit the line for the first time, it's just after you've added something; if you hit it on a subsequent iteration, it's because you decided to remove something that left `toRemoveBytes` strictly positive, which means there must still be something left in the eviction queue.

JamesRTaylor: Probably worth a test or two in PMetaDataImplTest. I haven't worked with this code in a while, but I seem to remember that we keep at least one item in PMetaDataImpl, even if it's larger than the configured max size (which would imply that we should break out of the loop when the size is one). Also, what if the max size is configured to be zero?

chiastic-security: I'll look at those (though not till Monday: I'm in the UK and about to head home).

chiastic-security: I've added a couple more tests.

The `cloneMinusOverage()` method does sometimes need to clear everything out, because it's what happens prior to adding the new thing, so it's correct for it not to break out of the loop when the size is one. I've added a test to confirm this (`shouldAlwaysKeepAtLeastOneEntryEvenIfTooLarge()`).

I've also added a test to check things work out OK when the max size is zero. In this case there should always be exactly one thing in the cache (after the first item has been added). This test also passes.

JamesRTaylor:
+1 to the patch. Thanks for the contribution, @chiastic-security. I've pushed it to master and will push to other branches soon.

chiastic-security:
Thanks. Did this get pushed? The PR is still coming up as open in github.

PR #119, opened 2015-10-02
Fix for performance.py not using JAVA_HOME

bpanneton:


chu11:
Note: This patch is for https://issues.apache.org/jira/browse/PHOENIX-2303

My only comment is should this patch look at hbase-env (similar to queryserver.py)?

bpanneton:
Yea, that would be better. I'll fix it up.

bpanneton:
Patch should now follow how queryserver.py gets the JAVA_HOME dir

chu11:
There's another call to 'java' later on in the performance.py that also needs to be adjusted.

bpanneton:
Fixed the java that I missed.

chu11:
LGTM

bpanneton:
Updated to have sqlline, sqlline-thing and psql in there.

stoty:
Already merged as PHOENIX-2303

PR #120, opened 2015-10-06
PHOENIX-2257 Fix broken integration test for query over derived GROUP…

chiastic-security:
… BY table (testDerivedTableWithGroupBy()), which worked on Java 7 but not on Java 8.

The test was retrieving two rows from a query with no ORDER BY clause, and assuming that they would come back in a specific order. This happened to match the order that Java 7 returned them in, but the order has changed in Java 8.

Test fixed by not assuming any particular order.

### DIFF
--- a/phoenix-core/src/it/java/org/apache/phoenix/end2end/DerivedTableIT.java
+++ b/phoenix-core/src/it/java/org/apache/phoenix/end2end/DerivedTableIT.java
@@ -18,33 +18,7 @@
 
 package org.apache.phoenix.end2end;
 
-import static org.apache.phoenix.util.TestUtil.A_VALUE;
-import static org.apache.phoenix.util.TestUtil.B_VALUE;
-import static org.apache.phoenix.util.TestUtil.C_VALUE;
-import static org.apache.phoenix.util.TestUtil.E_VALUE;
-import static org.apache.phoenix.util.TestUtil.ROW1;
-import static org.apache.phoenix.util.TestUtil.ROW2;
-import static org.apache.phoenix.util.TestUtil.ROW3;
-import static org.apache.phoenix.util.TestUtil.ROW4;
-import static org.apache.phoenix.util.TestUtil.ROW5;
-import static org.apache.phoenix.util.TestUtil.ROW6;
-import static org.apache.phoenix.util.TestUtil.ROW7;
-import static org.apache.phoenix.util.TestUtil.ROW8;
-import static org.apache.phoenix.util.TestUtil.ROW9;
-import static org.apache.phoenix.util.TestUtil.TEST_PROPERTIES;
-import static org.junit.Assert.assertEquals;
-import static org.junit.Assert.assertFalse;
-import static org.junit.Assert.assertTrue;
-
-import java.sql.Array;
-import java.sql.Connection;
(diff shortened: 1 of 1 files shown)


phoenix-core/src/it/java/org/apache/phoenix/end2end/DerivedTableIT.java
```diff
@@ -54,14 +28,17 @@
 import org.junit.runners.Parameterized;
 import org.junit.runners.Parameterized.Parameters;
 
-import com.google.common.collect.Lists;
+import java.sql.*;
+import java.util.*;
```
JamesRTaylor: Please see https://phoenix.apache.org/contributing.html#Code_conventions. You'll want to conform to these - easiest way is to import dev/eclipse_prefs_phoenix.epf and dev/phoenix.importorder into Eclipse. This is basically to ensure we don't get a lot of code differences due to different import ordering and whitespacing.

phoenix-core/src/it/java/org/apache/phoenix/end2end/DerivedTableIT.java
```diff
@@ -313,21 +290,25 @@ public void testDerivedTableWithGroupBy() throws Exception {
             query = "SELECT DISTINCT COLLECTDISTINCT(t.b) FROM (SELECT b_string b, a_string a FROM aTable GROUP BY a_string, b_string) AS t GROUP BY t.a";
             statement = conn.prepareStatement(query);
             rs = statement.executeQuery();
-            assertTrue (rs.next());
+            Set<Array> queryResults = new HashSet<>();
```
JamesRTaylor: Use BaseTest.assertValuesEqualsResultSet() instead which does the same thing as this.

phoenix-core/src/it/java/org/apache/phoenix/end2end/DerivedTableIT.java
```diff
@@ -313,21 +290,25 @@ public void testDerivedTableWithGroupBy() throws Exception {
             query = "SELECT DISTINCT COLLECTDISTINCT(t.b) FROM (SELECT b_string b, a_string a FROM aTable GROUP BY a_string, b_string) AS t GROUP BY t.a";
```
JamesRTaylor: @maryannxue - do you think the underlying order returned is indeterminate? For our tests, we have phoenix.query.force.rowkeyorder set to true, so underlying scans will return in row key order. Just want to make sure @chiastic-security fix isn't masking a bug.

maryannxue: @JamesRTaylor Do we have a deterministic ordering for arrays? If we do, I don't think this change should be made without finding out the cause.

maryannxue: @JamesRTaylor Please ignore my previous (deleted) comment. I just realized that the server group-by has replaced tree map with hash map, which means an optimization I did for client group-by based on the assumption that the input is a sorted group-by is now invalid. I will fix this ASAP. And after the fix, the test case should fine, we should keep it as it is.

stoty:
Already merged.

PR #121, opened 2015-10-16
PHOENIX-2304 NullPointerException when using an index and a char array

navis:
An NPE occurs when Phoenix attempts to retrieve a char array from an index. Retrieving other columns in the table is unaffected.


JamesRTaylor:
Thanks for the pull request, @navis. Would you mind including a unit test as well?

phoenix-core/src/main/java/org/apache/phoenix/parse/ColumnDef.java
```diff
@@ -78,7 +79,7 @@
          }
          
          this.isNull = isNull;
-         if (this.dataType == PChar.INSTANCE) {
+         if (this.dataType == PChar.INSTANCE || this.dataType == PCharArray.INSTANCE) {
```
JamesRTaylor: I'd be curious if BINARY type would require the same fix. A more general check would be:
    if (this.dataType.isFixedWidth()) {

along with generalizing the error messages.

navis:
Made two version of patches (a69735b and a20e123).

PR #129, opened 2015-11-19
PHOENIX-1674 Snapshot isolation transaction support through Tephra

twdsilva:
@JamesRTaylor  @samarthjain  

PR will code review feedback.

PR #130, opened 2015-11-20

twdsilva:
@samarthjain  @JamesRTaylor 
Closed https://github.com/apache/phoenix/pull/129 since its missing some files.
I added the review feedback, please +1 if everything looks OK


twdsilva:
For the last few files that aren't showing up in the diff view , you can view the diffs here
https://patch-diff.githubusercontent.com/raw/apache/phoenix/pull/130.patch

phoenix-core/src/main/java/org/apache/phoenix/exception/SQLExceptionCode.java
```diff
@@ -258,7 +264,18 @@ public SQLException newException(SQLExceptionInfo info) {
...
     DESC_VARBINARY_NOT_SUPPORTED(1056, "43A13", "Descending VARBINARY columns not supported"),
+    
+    DEFAULT_COLUMN_FAMILY_ON_SHARED_TABLE(1057, "43A13", "Default column family not allowed on VIEW or shared INDEX"),
+    ONLY_TABLE_MAY_BE_DECLARED_TRANSACTIONAL(1070, "44A01", "Only tables may be declared as transactional"),
+    TX_MAY_NOT_SWITCH_TO_NON_TX(1071, "44A02", "A transactional table may not be switched to non transactional"),
+	STORE_NULLS_MUST_BE_TRUE_FOR_TRANSACTIONAL(1072, "44A03", "Store nulls must be true when a table is transactional"),
+    CANNOT_START_TRANSACTION_WITH_SCN_SET(1073, "44A04", "Cannot start a transaction on a connection with SCN set"),
+    TX_MAX_VERSIONS_MUST_BE_GREATER_THAN_ONE(1074, "44A05", "A transactional table must define VERSION of greater than one"),
+    CANNOT_SPECIFY_SCN_FOR_TXN_TABLE(1075, "44A06", "Cannot use a connection with SCN set for a transactional table"),
+    NULL_TRANSACTION_CONTEXT(1076, "42901", "No Tranasction Context available"),
```
twdsilva: Will change the sqlState for this and the following exception.

samarthjain:
The check in TableRef.equals() isn't complete. It should be something like this:

if (((table.getName() == null && other.table.getName() != null)
||(table.getName() != null && other.table.getName() == null) 
|| !table.getName().getString().equals(other.table.getName().getString()))) 
return false;

phoenix-core/src/it/java/org/apache/phoenix/end2end/index/IndexIT.java
```diff
@@ -45,19 +45,24 @@
 	private final boolean localIndex;
 	private final String tableDDLOptions;
```
JamesRTaylor: Instead of using System.currentMillis() in the table name (which will create a new physical table for every test), can you create a TestUtil.getTableName(String prefix, boolean localIndex, boolean mutable, boolean transactional, Integer saltBuckets) and creates a sensible table name like T_LOCAL_MUTABLE_UNSALTED.

phoenix-core/src/main/java/org/apache/phoenix/schema/MetaDataClient.java
```diff
@@ -415,24 +425,47 @@ private long getClientTimeStamp() {
...
-            return new MetaDataMutationResult(MutationCode.TABLE_ALREADY_EXISTS,QueryConstants.UNSET_TIMESTAMP,table);
+        
+        boolean defaultTransactional = connection.getQueryServices().getProps().getBoolean(
+							                QueryServices.DEFAULT_TRANSACTIONAL_ATTRIB,
+							                QueryServicesOptions.DEFAULT_TRANSACTIONAL);
+        // start a txn if all table are transactional by default or if we found the table in the cache and it is transactional
+        // TODO if system tables become transactional remove the check 
+        boolean isTransactional = defaultTransactional || (table!=null && table.isTransactional());
+        if (!systemTable && isTransactional && !connection.getMutationState().isTransactionStarted()) {
+        	connection.getMutationState().startTransaction();
```
samarthjain: Question for my understanding - what is the purpose of starting a transaction when updating the metadatacache. Is there a code path which leads to updateCache and  the table involved isn't already part of a transaction?

phoenix-core/src/main/java/org/apache/phoenix/index/PhoenixTransactionalIndexer.java
```diff
@@ -0,0 +1,515 @@
...
+                }
+            }
+            /*
+            Cell cell = r.getColumnLatestCell(emptyColRef.getFamily(), emptyColRef.getQualifier());
+            if (cell != null) {
+                ImmutableBytesWritable ptr = new ImmutableBytesWritable();
+                ptr.set(cell.getValueArray(), cell.getValueOffset(), cell.getValueLength());
+                valueMap.put(emptyColRef, ptr);
+            }
+            */
```
JamesRTaylor: Remove commented out code.

phoenix-core/src/main/java/org/apache/phoenix/index/PhoenixTransactionalIndexer.java
```diff
@@ -0,0 +1,515 @@
...
+
+        @Override
+        public List<? extends IndexedColumnGroup> getIndexColumnHints() {
+            return Collections.emptyList();
+        }
+
+        public void applyMutation() {
+            /*if (mutation instanceof Delete) {
+                valueMap.clear();
+            } else */ {
```
JamesRTaylor: Here too - remove please.

phoenix-core/src/main/java/org/apache/phoenix/schema/MetaDataClient.java
JamesRTaylor: This is the point at which the timestamp is established, that's why the transaction is started here - because we need the transaction ID which corresponds to the timestamp. It's possible you could have non transactional or transactional tables here.

phoenix-core/src/it/java/org/apache/phoenix/end2end/index/IndexIT.java
twdsilva: Will change the check in TableRef.equals to

if ((table.getName() == null && other.table.getName() != null) 
                || (table.getName()!=null && !table.getName().getString().equals(other.table.getName().getString()))) return false;

PR #131, opened 2015-11-21

twdsilva:
PR without tests.

@JamesRTaylor  @samarthjain


JamesRTaylor:
FYI, @twdsilva broke this up into two separate pulls because GitHub wasn't show the diff for all files. The tests are over in https://github.com/apache/phoenix/pull/132. Both this pull and that one will be committed together - this is just to make the review easier.

phoenix-core/src/main/java/org/apache/phoenix/execute/AggregatePlan.java
```diff
@@ -51,6 +51,7 @@
 import org.apache.phoenix.iterate.SequenceResultIterator;
 import org.apache.phoenix.iterate.SpoolingResultIterator;
 import org.apache.phoenix.iterate.UngroupedAggregatingResultIterator;
+import org.apache.phoenix.jdbc.PhoenixStatement.Operation;
```
samarthjain: Is there no other change other than this import in the file? Probably should be reverted then?

phoenix-core/src/main/java/org/apache/phoenix/execute/MutationState.java
```diff
@@ -153,6 +344,15 @@ public void join(MutationState newMutationState) {
         if (this == newMutationState) { // Doesn't make sense
             return;
         }
+        // TODO: what if new and old have txContext as that's really an error
+        // Really it's an error if newMutation txContext is not null
+        if (txContext != null) {
```
samarthjain: Should the condition be caught and an IllegalStateException thrown?

phoenix-core/src/main/java/org/apache/phoenix/execute/MutationState.java
```diff
@@ -301,28 +476,80 @@ public void remove() {
...
+                if (rowTsColInfo.useServerTimestamp()) {
+                    // regenerate the key with this timestamp.
+                    key = getNewRowKeyWithRowTimestamp(key, timestampToUse, table);
+                } else {
+                    if (rowTsColInfo.getTimestamp() != null) {
+                        timestampToUse = rowTsColInfo.getTimestamp();
+                    }
+                }
+            }
+			PRow row = tableRef.getTable().newRow(connection.getKeyValueBuilder(), timestampToUse, key);
```
samarthjain: Very minor: indentation seems a bit off here.

phoenix-core/src/main/java/org/apache/phoenix/execute/DelegateHTableInterface.java
```diff
@@ -0,0 +1,300 @@
...
+import org.apache.hadoop.hbase.client.coprocessor.Batch.Callback;
+import org.apache.hadoop.hbase.filter.CompareFilter.CompareOp;
+import org.apache.hadoop.hbase.ipc.CoprocessorRpcChannel;
+
+import com.google.protobuf.Descriptors.MethodDescriptor;
+import com.google.protobuf.Message;
+import com.google.protobuf.Service;
+import com.google.protobuf.ServiceException;
+
+public class DelegateHTableInterface implements HTableInterface {
```
samarthjain: Should this class be renamed instead to just DelegateHTable?

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/MetaDataEndpointImpl.java
```diff
@@ -1222,7 +1228,7 @@ public void createTable(RpcController controller, CreateTableRequest request,
                 // exists without making an additional query
                 PTable table =
                         loadTable(env, tableKey, cacheKey, clientTimeStamp, HConstants.LATEST_TIMESTAMP);
-                if (table != null) {
+                if (table != null && !isTableDeleted(table)) {
```
JamesRTaylor: Took a closer look at this change and it doesn't look right. The if statement has a check inside of it for the table not being deleted. Can you revert this change and isolate any tests that are failing?

phoenix-core/src/main/java/org/apache/phoenix/execute/AggregatePlan.java
twdsilva: Will revert this file.

phoenix-core/src/main/java/org/apache/phoenix/execute/MutationState.java
twdsilva: @JamesRTaylor  should I add a check if the newMutation  txContext is not null and throw an IllegalStateException?

phoenix-core/src/main/java/org/apache/phoenix/execute/MutationState.java
twdsilva: Will fix this.

phoenix-core/src/main/java/org/apache/phoenix/execute/DelegateHTableInterface.java
twdsilva: Will rename.

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/MetaDataEndpointImpl.java
twdsilva: I reverted this change and ran all the tests and they passed, so I will revert this change.

PR #133, opened 2015-11-23

twdsilva:
@JamesRTaylor @samarthjain

Final PR with all feedback


phoenix-core/src/test/java/org/apache/phoenix/hbase/index/covered/TestLocalTableState.java
```diff
@@ -23,9 +23,10 @@
 import java.util.List;
```
samarthjain: Revert this file on commit.

phoenix-core/src/test/java/org/apache/phoenix/query/ParallelIteratorsSplitTest.java
```diff
@@ -455,4 +455,4 @@ public boolean useRoundRobinIterator() {
         List<KeyRange> keyRanges = parallelIterators.getSplits();
         return keyRanges;
     }
-}
\ No newline at end of file
+}
```
JamesRTaylor: Revert on commit

phoenix-core/src/main/java/org/apache/phoenix/util/ServerUtil.java
```diff
@@ -66,7 +67,9 @@ public static IOException createIOException(String msg, Throwable t) {
         } else if (t instanceof IOException) {
             // If the IOException does not wrap any exception, then bubble it up.
             Throwable cause = t.getCause();
-            if (cause == null || cause instanceof IOException) {
+            if (cause instanceof RetriesExhaustedWithDetailsException)
```
samarthjain: I don't see code changes that are now throwing this exception. Is this needed?

phoenix-core/src/main/java/org/apache/phoenix/util/IndexUtil.java
```diff
@@ -209,13 +212,34 @@ private static boolean isEmptyKeyValue(PTable table, ColumnReference ref) {
...
+            List<Delete> dataMutations, ImmutableBytesWritable ptr, final KeyValueBuilder kvBuilder, PhoenixConnection connection)
+            throws SQLException {
+        try {
+            IndexMaintainer maintainer = index.getIndexMaintainer(table, connection);
+            List<Delete> indexMutations = Lists.newArrayListWithExpectedSize(dataMutations.size());
+            for (final Mutation dataMutation : dataMutations) {
+                long ts = MetaDataUtil.getClientTimeStamp(dataMutation);
+                ptr.set(dataMutation.getRow());
+                Delete delete = maintainer.buildDeleteMutation(kvBuilder, ptr, ts);
+                // TODO: move to TransactionUtil
```
samarthjain: Not sure if the TODO is needed here. Remove on commit.

phoenix-core/src/it/java/org/apache/phoenix/transactions/RecreateTableIT.java
```diff
@@ -0,0 +1,59 @@
...
+
+    @BeforeClass
+    @Shadower(classBeingShadowed = BaseHBaseManagedTimeIT.class)
+    public static void doSetup() throws Exception {
+        Map<String, String> props = Maps.newHashMapWithExpectedSize(1);
+        setUpTestDriver(new ReadOnlyProps(props.entrySet().iterator()));
+    }
+
+    @Test
+    public void testReCreateTxnTableAfterDroppingExistingNonTxnTable() throws SQLException {
```
JamesRTaylor: Any reason this one test can't just be added to TransactionIT instead?

phoenix-core/src/it/java/org/apache/phoenix/transactions/RecreateTableIT.java
```diff
@@ -0,0 +1,59 @@
...
+        Properties props = PropertiesUtil.deepCopy(TEST_PROPERTIES);
+        Connection conn = DriverManager.getConnection(getUrl(), props);
+        conn.setAutoCommit(false);
+        Statement stmt = conn.createStatement();
+        stmt.execute("CREATE TABLE DEMO(k VARCHAR PRIMARY KEY, v1 VARCHAR, v2 VARCHAR)");
+        stmt.execute("DROP TABLE DEMO");
+        stmt.execute("CREATE TABLE DEMO(k VARCHAR PRIMARY KEY, v1 VARCHAR, v2 VARCHAR) TRANSACTIONAL=true");
+        stmt.execute("CREATE INDEX DEMO_idx ON DEMO (v1) INCLUDE(v2)");
```
JamesRTaylor: Seems strange that this test doesn't have any asserts. How about asserting that the table.isTransactional() is true and then that index.isTransactional() is true?

phoenix-core/src/main/java/org/apache/phoenix/schema/TableRef.java
```diff
@@ -111,8 +115,10 @@ public boolean equals(Object obj) {
         if (obj == null) return false;
         if (getClass() != obj.getClass()) return false;
         TableRef other = (TableRef)obj;
+        // a null alias on either side should mean a wildcard and should not fail the equals check
         if ((alias == null && other.alias != null) || (alias != null && !alias.equals(other.alias))) return false;
-        if (!table.getName().getString().equals(other.table.getName().getString())) return false;
+        if ((table.getName() == null && other.table.getName() != null)
```
samarthjain: This still doesn't look correct.

if (((table.getName() == null && other.table.getName() != null)
||(table.getName() != null && other.table.getName() == null) 
|| !table.getName().getString().equals(other.table.getName().getString()))) 
return false;

phoenix-core/src/it/java/org/apache/phoenix/execute/PartialCommitIT.java
```diff
@@ -241,6 +260,12 @@ private void testPartialCommit(List<String> statements, int failureCount, int[]
                 assertArrayEquals(expectedUncommittedStatementIndexes, uncommittedStatementIndexes);
             }
             
+            ResultSet rs1 = sta.executeQuery("select * from A_SUCCESS_TABLE where k='checkThatAllStatementTypesMaintainOrderInConnection' or k like 'z%'");
```
JamesRTaylor: Remove this loop as it's just printing something out?

phoenix-core/src/main/java/org/apache/phoenix/index/PhoenixTransactionalIndexer.java
```diff
@@ -0,0 +1,497 @@
...
+        Iterable<IndexUpdate> deletes = codec.getIndexDeletes(state, indexMetaData);
+        for (IndexUpdate delete : deletes) {
+            if (delete.isValid()) {
+                delete.getUpdate().setAttribute(TxConstants.TX_ROLLBACK_ATTRIBUTE_KEY, attribValue);
+                indexUpdates.add(new Pair<Mutation, byte[]>(delete.getUpdate(),delete.getTableName()));
+            }
+        }
+    }
+
+    boolean generatePuts(
```
JamesRTaylor: Make private if possible

phoenix-core/src/main/java/org/apache/phoenix/index/PhoenixTransactionalIndexer.java
```diff
@@ -0,0 +1,497 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more contributor license agreements. See the NOTICE
```
JamesRTaylor: Check that license matches others - looks like it might have gotten formatted incorrectly.

phoenix-core/src/main/java/org/apache/phoenix/index/PhoenixTransactionalIndexer.java
```diff
@@ -0,0 +1,497 @@
...
+        public byte[] getCurrentRowKey() {
+            return mutation.getRow();
+        }
+
+        @Override
+        public List<? extends IndexedColumnGroup> getIndexColumnHints() {
+            return Collections.emptyList();
+        }
+
+        public void applyMutation() {
```
JamesRTaylor: Can this be private too?

phoenix-core/src/main/java/org/apache/phoenix/util/ServerUtil.java
twdsilva: I had to make this change to get TxWriteFailureIT to work correctly. 
If a write fails it ends up eventually throwing a RetriesExhaustedWithDetailsException.

phoenix-core/src/main/java/org/apache/phoenix/schema/TableRef.java
JamesRTaylor: How about just this?

    if (!Objects.equal(table.getName(),other.table.getName())) return false;

phoenix-core/src/main/java/org/apache/phoenix/schema/PTableRef.java
```diff
@@ -0,0 +1,65 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.phoenix.schema;
+
+public class PTableRef {
+    private final PTable table;
+    private final int estSize;
+	private volatile long lastAccessTime;
```
samarthjain: Is it guaranteed that only one thread will be changing the lastAccessTime? And other threads will just read that value? If yes, then volatile is fine. Otherwise you will have to add more synchronization on it. Probably make it AtomicLong.

phoenix-core/src/it/java/org/apache/phoenix/transactions/RecreateTableIT.java
twdsilva: dropping a table and recreating it as a txn table was failing. I will add the asserts.

phoenix-core/src/main/java/org/apache/phoenix/schema/TableRef.java
twdsilva: I think the issue is table.getName() might be null and we want to compare using  table.getName().getString()

phoenix-core/src/main/java/org/apache/phoenix/schema/TableRef.java
```diff
@@ -111,8 +115,10 @@ public boolean equals(Object obj) {
         if (obj == null) return false;
         if (getClass() != obj.getClass()) return false;
         TableRef other = (TableRef)obj;
+        // a null alias on either side should mean a wildcard and should not fail the equals check
```
samarthjain: I don't think we are really implementing what is stated by this comment. We should remove it. 
Also, the alias needs a similar check the way it is for table.

if (!Objects.equal(alias, other.alias)) return false.

phoenix-core/src/main/java/org/apache/phoenix/schema/TableRef.java
samarthjain: Ah, I see. Then you would probably need to do the check that I recommended. For alias though, Objects.equal() should be fine.

JamesRTaylor: The code I posted will work fine if table.getName() is null, no?

phoenix-core/src/main/java/org/apache/phoenix/schema/MetaDataClient.java
```diff
@@ -1137,7 +1181,7 @@ public MutationState createIndex(CreateIndexStatement statement, byte[][] splits
                     if (hbaseVersion < PhoenixDatabaseMetaData.MUTABLE_SI_VERSION_THRESHOLD) {
                         throw new SQLExceptionInfo.Builder(SQLExceptionCode.NO_MUTABLE_INDEXES).setTableName(indexTableName.getTableName()).build().buildException();
                     }
-                    if (connection.getQueryServices().hasInvalidIndexConfiguration()) {
+                    if (connection.getQueryServices().hasInvalidIndexConfiguration() && !dataTable.isTransactional()) {
```
samarthjain: I think hasInvalidIndexConfiguration() should be renamed to something like isMutableIndexWALCodecInstalled(). The actual check happens on the server side. Renaming this way makes it more explicit that the change has to do with mutable secondary indexes and that if the table is transactional, then we don't care about having the right codec in place on region servers.

phoenix-core/src/main/java/org/apache/phoenix/schema/MetaDataClient.java
```diff
@@ -1268,19 +1312,19 @@ public MutationState createIndex(CreateIndexStatement statement, byte[][] splits
...
-                    long timestamp = scn == null ? HConstants.LATEST_TIMESTAMP : scn;
                     PName tenantId = connection.getTenantId();
                     String tenantIdStr = tenantId == null ? null : connection.getTenantId().getString();
                     PName physicalName = dataTable.getPhysicalName();
                     int nSequenceSaltBuckets = connection.getQueryServices().getSequenceSaltBuckets();
                     SequenceKey key = MetaDataUtil.getViewIndexSequenceKey(tenantIdStr, physicalName, nSequenceSaltBuckets);
-                    // Create at parent timestamp as we know that will be earlier than now
-                    // and earlier than any SCN if one is set.
+                    // if scn is set create at scn-1, so we can see the sequence or else use latest timestamp (so that latest server time is used)
+                    long sequenceTimestamp = scn!=null ? scn-1 : HConstants.LATEST_TIMESTAMP;
```
samarthjain: Are we treating sequences specially? If I remember correctly, if you create a table with connection timestamp as SCN, then the table becomes visible at SCN + 1. Has that changed?

phoenix-core/src/main/java/org/apache/phoenix/schema/TableRef.java
twdsilva: Changed this to 
```java
if (!Objects.equals(alias, other.alias)) return false;
if (!Objects.equals(table.getName(), other.table.getName())) return false;
```

phoenix-core/src/main/java/org/apache/phoenix/schema/PTableRef.java
twdsilva: Thats how it was in master, I just moved PTableRef into its own class and added the resolvedTimeStamp variable.

samarthjain:
+1. This is fantastic work, Thomas! Other than a few minor nits and comments, this looks great!

phoenix-core/src/main/java/org/apache/phoenix/schema/MetaDataClient.java
twdsilva: No, we are just setting the timestamp at which the sequence correctly instead of based on the data table timestamp.
This is required so that don't make unnesscary rpc calls while resolving a table.
Previously we were just creating the sequence at the parent table tiemstamp,  which caused issues

twdsilva:
@samarthjain  @JamesRTaylor 
Thanks for the review, I will make the remaining changes on commit.

PR #137, opened 2015-11-30
PHOENIX-2464 Convert phoenix-spark module to 'JUnitSuite' instead of …

gliptak:
…'FunSuite'

### DIFF
--- a/phoenix-spark/pom.xml
+++ b/phoenix-spark/pom.xml
@@ -534,6 +534,7 @@
           <reportsDirectory>${project.build.directory}/surefire-reports</reportsDirectory>
           <junitxml>.</junitxml>
           <filereports>WDF TestSuite.txt</filereports>
+          <jUnitClasses>org.apache.phoenix.spark.PhoenixSparkIT</jUnitClasses>
         </configuration>
         <executions>
           <execution>
--- a/phoenix-spark/src/it/scala/org/apache/phoenix/spark/PhoenixSparkIT.scala
+++ b/phoenix-spark/src/it/scala/org/apache/phoenix/spark/PhoenixSparkIT.scala
@@ -27,7 +27,9 @@ import org.apache.spark.sql.{Row, SaveMode, execution, SQLContext}
 import org.apache.spark.sql.types._
 import org.apache.spark.{SparkConf, SparkContext}
 import org.joda.time.DateTime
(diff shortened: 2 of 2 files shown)


jmahonin:
Hi @gliptak 

I tried to run the tests with the patch applied, and got a number of failures. I think that 'beforeAll' and 'afterAll' aren't actually getting invoked in with the JUnitSuite for some reason.

I tried adding the '@BeforeClass' and '@AfterClass' annotations, but JUnit warns that the methods should be static. Sadly, I think the only way to make that work is with a companion object (e.g. 'object PhoenixSparkIT' which contains the beforeAll and afterAll definitions).

gliptak:
@jmahonin Could you run the tests locally (my local setup is not quite working ...)? Thanks

gliptak:
Now it runs locally.

jmahonin:
Definitely closer, it does run now, but it takes about 4x longer to run than before. The 'doSetup' and 'doTeardown' used to be run only at the very beginning and at the very end, and now they're being run before every test due to the '@Before' and '@After' annotations.

I wonder if it's possible to do something like the 'ScalaClassRuleTest' in [1] and either specify a '@BeforeAll' or an '@ClassRule' in a companion object and move the 'doSetup' and 'doTeardown' there?

[1] http://randomallsorts.blogspot.ca/2012/11/junit-411-whats-new-rules.html

gliptak:
I measured 2x longer (with the Scalatest throwing warnings/errors during tear down ...). I will try to add the companion object.

gliptak:
Scala "static" methods on the companion object are not recognized as such by the ```@BeforeClass``` and ```@AfterClass``` annotations.

gliptak:
Another possibility is to duplicate ```BaseHBaseManagedTimeIT```/```BaseTest``` functionality in ```Scala```/```FunSuite``` ...

jmahonin:
Thanks @gliptak for looking into this.

I'll spend a bit of time seeing if I can work something out here. I think the main goal is just to try be more compatible with JUnit annotations in inherited classes, but I do like having these run as fast as possible as well.

I don't think it's necessarily a major problem if we do nothing here either - the issue only became apparent after merging in transaction support, a pretty large infrastructure addition, which only took a few lines to work around with the existing tests.

PR #141, opened 2015-12-02
PHOENIX-2481 JSON bulkload tool

ndimiduk:


ndimiduk:
Add a tool for loading json documents (one record per line) and refactor code in common between JSON and CSV loading.

JamesRTaylor:
Awesome, @ndimiduk - I love it. Couple of questions and some nits:
- For JSON, is only single level supported? If not, how is nested JSON handled?
- Looks like lists are converted into ARRAYs which is nice. What happens if the array elements have different data types? I guess you'd just log an error and ignore that row?
- Can you make sure the indenting conforms to the 4 space convention?
- Also, please make sure imports don't use * and that their order is correct (see dev/phoenix.importorder and dev/eclipse_prefs_phoenix.epf).

@gabrielreid - got a few spare cycles for a review?

ndimiduk:
> For JSON, is only single level supported? If not, how is nested JSON handled?

No special handling is provided. A nested json object will be read as Map and presented to the upsert statement as such. Since Phoenix doesn't support maps, an error would result.

> Looks like lists are converted into ARRAYs which is nice. What happens if the array elements have different data types? I guess you'd just log an error and ignore that row?

Indeed. Or fail the job, if `--ignore-errors` is not specified.

> Can you make sure the indenting conforms to the 4 space convention?

Pardon me, @JamesRTaylor. Done.

> Also, please make sure imports don't use * and that their order is correct

Done.

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/FormatToKeyValueMapper.java
```diff
@@ -0,0 +1,254 @@
...
+import com.google.common.annotations.VisibleForTesting;
+import com.google.common.base.Function;
+import com.google.common.base.Joiner;
+import com.google.common.base.Splitter;
+import com.google.common.base.Throwables;
+import com.google.common.collect.ImmutableList;
+import com.google.common.collect.Iterables;
+import com.google.common.collect.Lists;
+
+public abstract class FormatToKeyValueMapper<RECORD> extends Mapper<LongWritable, Text, TableRowkeyPair,
```
joshelser: Some javadocs would be nice :)

ndimiduk:
Okay. Went through all javadocs. Also rebased onto master.

gabrielreid:
Very cool, looks good! I noted a few small nits in the commit, but apart from that it all looks good to me.

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/JsonToKeyValueMapper.java
```diff
@@ -0,0 +1,75 @@
...
+import org.apache.phoenix.util.ColumnInfo;
+import org.apache.phoenix.util.UpsertExecutor;
+import org.apache.phoenix.util.json.JsonUpsertExecutor;
+import org.codehaus.jackson.map.ObjectMapper;
+
+import com.google.common.annotations.VisibleForTesting;
+import com.google.common.base.Preconditions;
+
+/**
+ * MapReduce mapper that converts CSV input lines into KeyValues that can be written to HFiles.
```
JamesRTaylor: CSV -> JSON

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/JsonBulkLoadTool.java
```diff
@@ -0,0 +1,53 @@
...
+import java.util.List;
+
+import org.apache.commons.cli.CommandLine;
+import org.apache.hadoop.conf.Configuration;
+import org.apache.hadoop.mapreduce.Job;
+import org.apache.hadoop.util.ToolRunner;
+import org.apache.phoenix.util.ColumnInfo;
+
+/**
+ * Base tool for running MapReduce-based ingests of data.
```
JamesRTaylor: Tool for running MapReduce-based ingests of JSON data.

JamesRTaylor: Nested JSON data is not handled and lists are converted to ARRAYs.

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/JsonToKeyValueMapper.java
ndimiduk: Done.

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/JsonBulkLoadTool.java
ndimiduk: done.

nagarjunak758:
Would be great if the apache phoenix documentation is also updated, possibly with some examples.

nagarjunak758:
Would be great if the documentation is updated, possibly with few examples.

JamesRTaylor:
Thanks for the feedback, @nagarjunak758. I've added a mention of the JSON bulk loader in our Bulk Loading page here: https://phoenix.apache.org/bulk_dataload.html. Please feel free to send us a patch with additional information/examples. The information on how to do that is here: https://phoenix.apache.org/building_website.html

stoty:
Already merged.

PR #142, opened 2015-12-05
PHOENIX-2492 Expose PhoenixRecordWritable outside of spark

ndimiduk:
This is a simple translation of phoenix-spark's `PhoenixRecordWritable` to java, and has been helpful for my needs. Probably the next step is to rip out the scala version and replumb to use the single implementation everywhere.

### DIFF
--- a/phoenix-core/src/main/java/org/apache/phoenix/mapreduce/PhoenixRecordWritable.java
+++ b/phoenix-core/src/main/java/org/apache/phoenix/mapreduce/PhoenixRecordWritable.java
@@ -1,0 +1,192 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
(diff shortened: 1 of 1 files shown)


JamesRTaylor:
This seems like a good idea. @mravi - would you have a few spare cycles to review?

jmahonin:
This is a good idea. There are actually a few implementations of DBWritable in the codebase:

- PhoenixRecordWritable
- PhoenixPigDBWritable
- PhoenixIndexDBWritable

The code seems very similar across implementations, so we definitely should have one base implementation that works well across the board and can be simply extended, if need be. Having it in 'phoenix-core' is a bonus too.

I'd like to see the phoenix-spark 'PhoenixRecordWritable' renamed to maybe 'PhoenixSparkRecordWritable' and have it trimmed down to implement this new class, but maybe that's better suited to a follow-up PR.

mravi:
+1 for a single implementation to be used across all modules. Off late, I had to create a new one for Sqoop support  :( .

ndimiduk:
Sounds like we have a game-plan to proceed. I'm happy to pursue the work of consolidation, though I'm concerned about touching (and breaking) so many means of interaction. We'll see how testing goes.

jmahonin:
Unless it's already on your radar @ndimiduk , I can take a stab some time this week at updating the scala version to extend the new Java version.

ndimiduk:
Please help yourself @jmahonin. I've been able to move forward with my own project via the class in this PR. I should be getting into some spark stuff later this week, which is the only reason this might come up again on my end.

stoty:
Already merged.

PR #144, opened 2015-12-11
PHOENIX-2484 Disable logback for all but tephra code.

joshelser:


joshelser:
Gabriel's patch is the right way to fix this.

PR #146, opened 2016-01-04
PHOENIX-2364 Fix Reverse timestamp row key

freakyzoidberg:
When inverting the nanosecond part of a Timestamp key the offset should be reset.


JamesRTaylor:
Thanks for the patch, @freakyzoidberg. @samarthjain - would you mind reviewing and committing? It looks good to me. Probably worth making a quick pass through the code for calls to SortOrder.invert() and make sure we're setting the offset as is done in this patch.

PR #147, opened 2016-01-15
PHOENIX-2417 Compress memory used by row key byte[] of guideposts

ankitsinghal:
- Still need to tweak some copying of bytes
- Having only these two test case failing currently 
    ViewIT.testNonSaltedUpdatableViewWithIndex:129->BaseViewIT.testUpdatableViewWithIndex:85->BaseViewIT.testUpdatableViewIndex:158 expected:<6> but was:<2>
  ViewIT.testNonSaltedUpdatableViewWithIndex:129->BaseViewIT.testUpdatableViewWithIndex:85->BaseViewIT.testUpdatableViewIndex:158 expected:<6> but was:<2>


phoenix-core/src/main/java/org/apache/phoenix/compile/ScanRanges.java
```diff
@@ -237,7 +237,9 @@ public void initializeScan(Scan scan) {
         return temp;
     }
     
-    public Scan intersectScan(Scan scan, final byte[] originalStartKey, final byte[] originalStopKey, final int keyOffset, boolean crossesRegionBoundary) {
+    public Scan intersectScan(Scan scan, final ImmutableBytesWritable originalStartKeyPtr, final ImmutableBytesWritable originalStopKeyPtr, final int keyOffset, boolean crossesRegionBoundary) {
+        byte[] originalStartKey= originalStartKeyPtr.get();
+        byte[] originalStopKey= originalStopKeyPtr.get();
```
JamesRTaylor: You can't treat an ImmutableBytesWritable the same as what was a byte[] before because an ImmutableBytesWritable has an offset and a length. By doing it this way, you're assuming that the offset is 0 and the length is byte[].length. Instead, you'd want to change the type and adjust the code as necessary:

    ImmutableBytesWritable originalStartKey= originalStartKeyPtr;
    ImmutableBytesWritable originalStopKey = originalStopKeyPtr;

phoenix-core/src/main/java/org/apache/phoenix/schema/stats/GuidePostsInfo.java
```diff
@@ -53,22 +70,21 @@
...
      */
-    public GuidePostsInfo(long byteCount, List<byte[]> guidePosts, long rowCount) {
-        this.guidePosts = ImmutableList.copyOf(guidePosts);
-        int size = 0;
-        for (byte[] key : guidePosts) {
-            size += key.length;
-        }
-        this.keyByteSize = size;
+    public GuidePostsInfo(long byteCount, ImmutableBytesWritable guidePosts, long rowCount, int maxLength, int guidePostsCount) {
+        this.guidePosts = guidePosts;
```
JamesRTaylor: Best to make a copy of the ImmutableBytesWritable here in case the caller uses the ptr again. Note that this doesn't copy the underlying bytes:

    this.guidePosts = new ImmutableBytesWritable(guidePosts);

phoenix-core/src/main/java/org/apache/phoenix/schema/PTableImpl.java
```diff
@@ -1003,26 +1003,16 @@ public static PTable createFromProto(PTableProtos.PTable table) {
...
-            }
             long guidePostsByteCount = pGuidePosts.getByteCount();
             long rowCount = pGuidePosts.getRowCount();
-            // TODO : Not exposing MIN/MAX key outside to client 
-            GuidePostsInfo info =
-                    new GuidePostsInfo(guidePostsByteCount, value, rowCount);
+            int maxLength = pGuidePosts.getMaxLength();
+            int guidePostsCount = pGuidePosts.getGuidePostsCount();
+            GuidePostsInfo info = new GuidePostsInfo(guidePostsByteCount,
+                    new ImmutableBytesWritable(pGuidePosts.getGuidePosts().toByteArray()), rowCount, maxLength, guidePostsCount);
```
JamesRTaylor: Use new ImmutableBytesWritable(HBaseZeroCopyByteString.zeroCopyGetBytes(pGuidePosts.getGuidePosts())) here (and elsewhere) instead to prevent copying the underlying byte array.

phoenix-core/src/main/java/org/apache/phoenix/iterate/BaseResultIterators.java
```diff
@@ -490,15 +493,37 @@ private static String toString(List<byte[]> gps) {
...
+        PrefixByteDecoder decoder = null;
+        int guideIndex = 0;
+        if (gpsSize > 0) {
+            stream = new ByteArrayInputStream(guidePosts.get(), guidePosts.getOffset(), guidePosts.getLength());
+            input = new DataInputStream(stream);
+            if (gps.getMaxLength() > 0) {
+                decoder = new PrefixByteDecoder(gps.getMaxLength());
+            } else {
+                decoder = new PrefixByteDecoder();
+            }
```
JamesRTaylor: Instead of this if statement everywhere, just do this (and ensure that PrefixByteDecoder does the right thing if the maxLength is zero):

    decoder = new PrefixByteDecoder(gps.getMaxLength());

phoenix-core/src/main/java/org/apache/phoenix/iterate/BaseResultIterators.java
```diff
@@ -490,15 +493,37 @@ private static String toString(List<byte[]> gps) {
...
+            try {
+                currentGuidePost = CodecUtils.decode(decoder, input);
+                while (currentKey.getLength() != 0 && Bytes.compareTo(currentKey.get(), currentGuidePost.get()) <= 0) {
```
JamesRTaylor: You can't treat an ImmutableBytesWritable the same as a byte[]. To compare them, you do this:

    while (currentKey.compareTo(currentGuidePost = CodecUtils.decode(decoder, input)) <= 0) {
        guideIndex++;
    }

JamesRTaylor:
Thanks for the patch, @ankitsinghal. The main remaining issues are around not treating an ImmutableBytesWritable the same as a byte[]. Maybe easiest for now to have BaseResultIterators.getParallelScans() declare currentKey as an ImmutableBytesWritable, but keep

    // Do this as infrequently as possible to prevent a copy of the backing byte array
    byte[] currentKeyBytes = SchemaUtil.copyKeyIfNecessary(currentKey);
    byte[] currentGuidePostBytes = SchemaUtil.copyKeyIfNecessary(currentGuidePost);

    Scan newScan = scanRanges.intersectScan(scan, currentKeyBytes, currentGuidePostBytes, keyOffset, false);
    scans = addNewScan(parallelScans, scans, newScan, currentGuidePostBytes, false, regionLocation);

ankitsinghal:
Thanks @JamesRTaylor  for the review.
I have made the changes you have suggested above except this one.
byte[] currentGuidePostBytes = SchemaUtil.copyKeyIfNecessary(currentGuidePost);

As PrefixByteDecoder updates the previous buffer only whenever maxLength is passed as a part of optimization.

public ImmutableBytesWritable decode(DataInput in) throws IOException {
        int prefixLen = WritableUtils.readVInt(in);
        int suffixLen = WritableUtils.readVInt(in);
        int length = prefixLen + suffixLen;
        byte[] b;
        if (maxLength == -1) { // Allocate new byte array each time
            b = new byte[length];
            System.arraycopy(previous.get(), previous.getOffset(), b, 0, prefixLen);
        } else { // Reuse same buffer each time
            b = previous.get();
        }
        in.readFully(b, prefixLen, suffixLen);
        previous.set(b, 0, length);
        return previous;
    }

so I need to copy bytes even if the length of the ImmutableByteWritable is equal to byte[] contained in it.

What do you think about this?

And also I have fixed the failure test cases as well. there was logical operator problem while incrementing 
guidePosts till the start key.


I have run the complete test suite now and will confirm you once it is completed.

phoenix-protocol/src/main/PTable.proto
```diff
@@ -52,11 +52,12 @@ message PColumn {
 
 message PTableStats {
   required bytes key = 1;
-  repeated bytes values = 2;
+  optional bytes guidePosts = 2;
```
JamesRTaylor: This might be a b/w compat issue. Safest would be to add guidePosts at the end and keep values as-is.

phoenix-core/src/main/java/org/apache/phoenix/iterate/BaseResultIterators.java
```diff
@@ -490,15 +492,35 @@ private static String toString(List<byte[]> gps) {
         }
         List<List<Scan>> parallelScans = Lists.newArrayListWithExpectedSize(stopIndex - regionIndex + 1);
         
-        byte[] currentKey = startKey;
-        int guideIndex = currentKey.length == 0 ? 0 : getIndexContainingInclusive(gps, currentKey);
-        int gpsSize = gps.size();
+        ImmutableBytesWritable currentKey = new ImmutableBytesWritable(startKey, 0, startKey.length);
```
JamesRTaylor: Minor nit: you can just do this instead:

    ImmutableBytesWritable currentKey = new ImmutableBytesWritable(startKey);

phoenix-core/src/main/java/org/apache/phoenix/iterate/BaseResultIterators.java
```diff
@@ -490,15 +492,35 @@ private static String toString(List<byte[]> gps) {
...
+            input = new DataInputStream(stream);
+            decoder = new PrefixByteDecoder(gps.getMaxLength());
+            try {
+                while (currentKey.compareTo(currentGuidePost = CodecUtils.decode(decoder, input)) >= 0
+                        && currentKey.getLength() != 0) {
+                    guideIndex++;
+                }
+            } catch (EOFException e) {}
+        }
+        byte[] currentKeyBytes = currentKey.copyBytes();
```
JamesRTaylor: Always use ByteUtil.copyKeyBytesIfNecessary(currentKey) instead of copyBytes() as it's often not necessary to actually copy the bytes.

phoenix-core/src/main/java/org/apache/phoenix/iterate/BaseResultIterators.java
```diff
@@ -490,15 +492,35 @@ private static String toString(List<byte[]> gps) {
...
+
         // Merge bisect with guideposts for all but the last region
         while (regionIndex <= stopIndex) {
-            byte[] currentGuidePost, endKey, endRegionKey = EMPTY_BYTE_ARRAY;
+            byte[] currentGuidePostBytes = currentGuidePost.copyBytes();
```
JamesRTaylor: And here

phoenix-core/src/main/java/org/apache/phoenix/iterate/BaseResultIterators.java
```diff
@@ -510,14 +532,18 @@ private static String toString(List<byte[]> gps) {
...
-                currentKey = currentGuidePost;
-                guideIndex++;
-            }
-            Scan newScan = scanRanges.intersectScan(scan, currentKey, endKey, keyOffset, true);
+            try {
+                while (guideIndex < gpsSize && (currentGuidePost.compareTo(endKey) <= 0 || endKey.length == 0)) {
+                    Scan newScan = scanRanges.intersectScan(scan, currentKeyBytes, currentGuidePostBytes, keyOffset,
+                            false);
+                    scans = addNewScan(parallelScans, scans, newScan, currentGuidePostBytes, false, regionLocation);
+                    currentKeyBytes = currentGuidePost.copyBytes();
```
JamesRTaylor: And here

phoenix-core/src/main/java/org/apache/phoenix/iterate/BaseResultIterators.java
```diff
@@ -510,14 +532,18 @@ private static String toString(List<byte[]> gps) {
...
+                    currentGuidePost = CodecUtils.decode(decoder, input);
+                    currentGuidePostBytes = currentGuidePost.copyBytes();
```
JamesRTaylor: And here

phoenix-core/src/main/java/org/apache/phoenix/iterate/BaseResultIterators.java
```diff
@@ -461,9 +463,9 @@ private static String toString(List<byte[]> gps) {
         for(Pair<byte[], byte[]> where : context.getWhereConditionColumns()) {
           whereConditions.add(where.getFirst());
         }
-        List<byte[]> gps = getGuidePosts(whereConditions);
+        GuidePostsInfo gps = getGuidePosts(whereConditions);
         if (logger.isDebugEnabled()) {
-            logger.debug("Guideposts: " + toString(gps));
+            logger.debug("Guideposts: " + gps);
         }
```
JamesRTaylor: Minor: I'd get rid of this logger if statement. The output could be pretty huge.

phoenix-core/src/main/java/org/apache/phoenix/schema/stats/GuidePostsInfo.java
```diff
@@ -79,67 +96,66 @@ public long getRowCount() {
...
-        if (guidePosts.isEmpty() || Bytes.compareTo(row, guidePosts.get(guidePosts.size() - 1)) > 0) {
-            List<byte[]> newGuidePosts = Lists.newArrayListWithExpectedSize(this.getGuidePosts().size() + 1);
-            newGuidePosts.addAll(guidePosts);
-            newGuidePosts.add(row);
-            this.guidePosts = ImmutableList.copyOf(newGuidePosts);
-            this.byteCount += byteCount;
-            this.keyByteSize += row.length;
-            this.rowCount+=rowCount;
-            return true;
+    public boolean encodeAndCollectGuidePost(byte[] row, long byteCount, long rowCount) {
```
JamesRTaylor: This feels a little funky here. How about if we have a GuidePostsInfoBuilder or GuidePostsInfoWriter with a writeGuidePost(DataOutput out, byte[] row, long byteCount, long rowCount) method?

phoenix-core/src/main/java/org/apache/phoenix/schema/stats/GuidePostsInfo.java
```diff
@@ -79,67 +96,66 @@ public long getRowCount() {
...
+        return encodeAndCollectGuidePost(row, 0, 0);
     }
 
-    public boolean addGuidePost(byte[] row, long byteCount) {
-        return addGuidePost(row, byteCount, 0);
+    public boolean encodeAndCollectGuidePost(byte[] row, long byteCount){
+        return encodeAndCollectGuidePost(row, byteCount, 0);
     }
 
+    public void close() {
```
JamesRTaylor: This would move to the GuidePostsInfoWriter class or maybe be unnecessary at all as the stream would be created outside of the class.

JamesRTaylor:
Looks good functionally now, @ankitsinghal. Just needs a little bit of cleanup. Nice work!

phoenix-core/src/main/java/org/apache/phoenix/iterate/BaseResultIterators.java
ankitsinghal: But @JamesRTaylor , here the problem is that PrefixByteDecoder reused the same byte[] for subsequent values too.

public ImmutableBytesWritable decode(DataInput in) throws IOException {
int prefixLen = WritableUtils.readVInt(in);
int suffixLen = WritableUtils.readVInt(in);
int length = prefixLen + suffixLen;
byte[] b;
if (maxLength == -1) { // Allocate new byte array each time
b = new byte[length];
System.arraycopy(previous.get(), previous.getOffset(), b, 0, prefixLen);
} else { // Reuse same buffer each time
b = previous.get();
}
in.readFully(b, prefixLen, suffixLen);
previous.set(b, 0, length);
return previous;
}

And, ByteUtil.copyKeyBytesIfNecessary(currentKey) copies bytes only when ImmutableByteWritable length is not equal to contained byte[] length.

So ,we need to copyBytes everytime as the subsequent call to decode will overwrite the same array object.

If you agree on above below comments will follow the same.

phoenix-core/src/main/java/org/apache/phoenix/iterate/BaseResultIterators.java
ankitsinghal: same as above.

phoenix-core/src/main/java/org/apache/phoenix/iterate/BaseResultIterators.java
ankitsinghal: same as above.

phoenix-core/src/main/java/org/apache/phoenix/iterate/BaseResultIterators.java
ankitsinghal: same as above. As all are using the same byte[] object which the decoder is using.

phoenix-core/src/main/java/org/apache/phoenix/iterate/BaseResultIterators.java
JamesRTaylor: Yes, you're right, @ankitsinghal. Never mind about these changes.

phoenix-protocol/src/main/PTable.proto
JamesRTaylor: There's a pretty big backward compatibility issue due to PHOENIX-2143 and this one. The case you'll need to make work is an old pre 4.7.0 client that's running against a new 4.7.0 server. The client will expect the stats to be in the original format. In the following call:

    public void getTable(RpcController controller, GetTableRequest request,
            RpcCallback<MetaDataResponse> done) {

You'll need to pass request.getClientVersion() through doGetTable(), into getTable() and finally into StatisticsUtil.readStatistics(). You should preserve the old code (we can dump it when we do a major release), and use that code path if the stats have not been regenerated yet. You can detect this based on the existence of the GUIDE_POSTS key value (which you'll want to project into the scan for the new code for this b/w compatibility case). If the stats have been regenerated, there'd be two cases: the client is pre 4.7.0 in which case you'd want to use the new code but put the data in the old format, or the client is 4.7.0 or above in which case your existing code is fine.

With PHOENIX-2143, when compaction runs, we'll generate stats in the new format. It's possible that the SYSTEM.STATS table hasn't been updated yet (as this gets triggered when a new 4.7.0 client connects to the server which may not yet have happened). We'd need to issue the previous Delete marker based on the old row key structure to ensure that the stats for the region are deleted. We wouldn't want to issue the query that does the range delete in this case because it might delete rows across multiple regions (ugh). So we'd need to know if the schema upgrade has been done yet when compaction runs. We could detect this by querying the SYSTEM.CATALOG table directly or by using the MetaDataProtocol.getTable() call and pulling over the PTable and then conditionally do the delete the old way versus the new way.

WDYT, @ankitsinghal? A more radical alternative would be to call this release 5.0. Users could still upgrade the server and client as with a minor release, but they'd need to truncate the SYSTEM.STATS table manually before upgrading the server. In that case, I think it'd be acceptable to return an empty guidepost for the protobuf values field (as essentially stats would be disabled for older clients running against the newer server).

JamesRTaylor: I'm leaning toward having this release be 5.0. We can add some code MetaDataRegionObserver.postOpen() that does a checkAndPut on the SYSTEM.CATALOG table for the SYSTEM.STATS row where we conditionally truncate the table (i.e. invoke the code you already wrote).

JamesRTaylor: Let's do something in-the-middle. We can stick with the plan that this is still 4.7.0 release, but we can do the above in MetaDataRegionObserver to ensure that the SYSTEM.STATS table is truncated. Here what needs to be done:
* conditionally truncate SYSTEM.STATS table in MetaDataRegionObserver.postOpen() based on checkAndPut
* keep values field at protobuf position 2 and return an empty PGuidePosts for that field. We'll document that stats are essentially disabled for an old client once you upgrade your server (but nothing will break).

phoenix-core/src/main/java/org/apache/phoenix/schema/stats/GuidePostsInfo.java
ankitsinghal: moved it to GuidePostsInfoWriter but it has no affect as ByteArrayOutputStream doesn't have close implementation. so there is also no need to copy buffer from TrustedByteArrayOutputStream as it is available after close also.

creating stream in writer only as it may be easy to maintain and track streams against guidePosts collected at multiple column family.

Updated the code with the guidePostInfoWriter

phoenix-protocol/src/main/PTable.proto
ankitsinghal: yes it makes sense, as truncate of system.stats is necessary. I'll try this tomorrow.

Yes ,as per your previous comments , I'll be keeping "values" field at position 2, so this will automatically ensure that empty PGuidePosts is returned when client older than 4.7 is used right?
As per below code from version older than 4.7 (PTableImpl.createFromProto() )
GuidePostsInfo info =
                    new GuidePostsInfo(guidePostsByteCount, value, rowCount);//Prior 4.7 version :- empty "value" list.

3) I have created a GuidePostsInfoWriter and you can review the changes in this pull request now.

phoenix-core/src/main/java/org/apache/phoenix/util/CodecUtils.java
```diff
@@ -0,0 +1,104 @@
...
+import java.io.DataOutputStream;
+import java.io.EOFException;
+import java.io.IOException;
+import java.util.List;
+
+import org.apache.hadoop.hbase.io.ImmutableBytesWritable;
+
+import com.google.common.collect.Lists;
+
+public class CodecUtils {
```
JamesRTaylor: Rename to PrefixByteCodec

phoenix-core/src/main/java/org/apache/phoenix/schema/stats/GuidePostsInfoWriter.java
```diff
@@ -0,0 +1,129 @@
...
+import org.apache.phoenix.util.ByteUtil;
+import org.apache.phoenix.util.CodecUtils;
+import org.apache.phoenix.util.PrefixByteEncoder;
+import org.apache.phoenix.util.TrustedByteArrayOutputStream;
+
+/*
+ * Writer to help in writing guidePosts and creating guidePostInfo. This is used when we are collecting stats or reading stats for a table.
+ */
+
+public class GuidePostsInfoWriter {
```
JamesRTaylor: Much better, but let's rename to GuidePostsInfoBuilder

phoenix-core/src/main/java/org/apache/phoenix/schema/stats/GuidePostsInfoWriter.java
```diff
@@ -0,0 +1,129 @@
...
+
+    /**
+     * The guide posts, rowCount and byteCount are accumulated every time a guidePosts depth is
+     * reached while collecting stats.
+     * @param row
+     * @param byteCount
+     * @return
+     * @throws IOException 
+     */
+    public boolean writeGuidePosts( byte[] row, long byteCount, long rowCount) {
```
JamesRTaylor: Rename to addGuidePosts

phoenix-core/src/main/java/org/apache/phoenix/schema/stats/GuidePostsInfoWriter.java
```diff
@@ -0,0 +1,129 @@
...
+                lastRow = row;
+                return true;
+            } catch (IOException e) {
+                return false;
+            }
+        }
+        return false;
+    }
+    
+    public boolean writeGuidePosts(byte[] row){
```
JamesRTaylor: Here too: addGuidePosts

phoenix-core/src/main/java/org/apache/phoenix/schema/stats/GuidePostsInfoWriter.java
```diff
@@ -0,0 +1,129 @@
...
+        return writeGuidePosts(row, 0, 0);
+    }
+
+    public boolean writeGuidePosts(byte[] row, long byteCount){
```
JamesRTaylor: And here: addGuidePosts

phoenix-core/src/main/java/org/apache/phoenix/schema/stats/GuidePostsInfoWriter.java
```diff
@@ -0,0 +1,129 @@
...
+        return writeGuidePosts(row, byteCount, 0);
+    }
+
+    public void close() {
```
JamesRTaylor: Make this private and call at end of build() method.

phoenix-core/src/main/java/org/apache/phoenix/schema/stats/GuidePostsInfoWriter.java
```diff
@@ -0,0 +1,129 @@
...
+    public long getRowCount() {
+        return rowCount;
+    }
+
+    public ImmutableBytesWritable getGuidePosts() {
+        this.guidePosts.set(stream.getBuffer(), 0, stream.size());
+        return guidePosts;
+    }
+    
+    public GuidePostsInfo createGuidePostInfo(){
```
JamesRTaylor: Rename to build() and call close() at end.

phoenix-core/src/main/java/org/apache/phoenix/schema/stats/GuidePostsInfoWriter.java
```diff
@@ -0,0 +1,129 @@
...
+
+    public void close() {
+        CodecUtils.close(stream);
+    }
+
+    public void incrementRowCount() {
+        this.rowCount++;
+    }
+
+    public long getByteCount() {
```
JamesRTaylor: Make the getters private if possible (otherwise leave as is)

JamesRTaylor:
Looking good. One more pass, and I think we'll be good to go. Thanks!

ankitsinghal:
Made the changes in GuidePostInfoBuilder as suggested and attach a wip patch for stats upgrade from server side on ticket. 
Please review and confirm if I'm on the right track.

phoenix-protocol/src/main/PGuidePosts.proto
```diff
@@ -24,7 +24,9 @@ option java_generic_services = true;
 option java_generate_equals_and_hash = true;
 option optimize_for = SPEED;
 message PGuidePosts {
-  repeated bytes guidePosts = 1;
```
JamesRTaylor: Is this protobuf change required? Will it cause b/w compat issues?

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/MetaDataRegionObserver.java
```diff
@@ -96,10 +111,109 @@ public void start(CoprocessorEnvironment env) throws IOException {
...
+                    Thread.sleep(1000);
+                    LOG.info("Stats will be deleted for upgrade 4.7 requirement!!");
+                    metaTable = env.getTable(TableName.valueOf(PhoenixDatabaseMetaData.SYSTEM_CATALOG_NAME));
+                    List<Cell> columnCells = metaTable
+                            .get(new Get(SchemaUtil.getTableKey(null, PhoenixDatabaseMetaData.SYSTEM_SCHEMA_NAME,
+                                    PhoenixDatabaseMetaData.SYSTEM_CATALOG_TABLE)))
+                            .getColumnCells(PhoenixDatabaseMetaData.TABLE_FAMILY_BYTES,
+                                    QueryConstants.EMPTY_COLUMN_BYTES);
+                    if (!columnCells.isEmpty()
+                            && columnCells.get(0).getTimestamp() < MetaDataProtocol.MIN_SYSTEM_TABLE_TIMESTAMP_4_7_0) {
```
JamesRTaylor: This if is not needed.

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/MetaDataRegionObserver.java
```diff
@@ -96,10 +111,109 @@ public void start(CoprocessorEnvironment env) throws IOException {
...
+    }
 
     @Override
     public void postOpen(ObserverContext<RegionCoprocessorEnvironment> e) {
+        final RegionCoprocessorEnvironment env = e.getEnvironment();
+
+        Runnable r = new Runnable() {
+            @Override
+            public void run() {
+                HTableInterface metaTable = null;
```
JamesRTaylor: Move the Thread.sleep(1000) outside here and move this logic into new UpgradeUtil.truncateStats(HTableInterface metaTable, HTableInterface statsTable) method.

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/MetaDataRegionObserver.java
```diff
@@ -69,6 +82,8 @@
     protected ScheduledThreadPoolExecutor executor = new ScheduledThreadPoolExecutor(1);
     private boolean enableRebuildIndex = QueryServicesOptions.DEFAULT_INDEX_FAILURE_HANDLING_REBUILD;
     private long rebuildIndexTimeInterval = QueryServicesOptions.DEFAULT_INDEX_FAILURE_HANDLING_REBUILD_INTERVAL;
+    public static final byte[] UPGRADE_TO_4_7_COLUMN_NAME = Bytes.toBytes("UPGRADE_TO_4_7");
+    public static final String UPGRADE_TO_4_7_COLUMN_VALUE = "TRUNCATED_4_7";
```
JamesRTaylor: Minor nit - no need for a column value. Just using ByteUtil.EMPTY_BYTE_ARRAY.

JamesRTaylor:
Looks good. Couple of minor nits and one b/w compat question. Looks like it needs to be rebased too. Thanks for the contribution, @ankitsinghal  - nice work!

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/MetaDataRegionObserver.java
ankitsinghal: I think this is needed to prevent system.stats being deleted after the client upgrade also as upgrade in connectionQueryServicesImpl will delete all the rows of system.stats from catalog including the dummy column. WDYT?

phoenix-protocol/src/main/PGuidePosts.proto
ankitsinghal: These fields were optional and will not impact as we are not using stats with old client. But anyways I have made the change to add a new field and keep the old one as it is.

phoenix-core/src/main/java/org/apache/phoenix/util/UpgradeUtil.java
```diff
@@ -1215,4 +1218,67 @@ public static void addRowKeyOrderOptimizableCell(List<Mutation> tableMetadata, b
...
     }
+
+    public static boolean truncateStats(HTableInterface metaTable, HTableInterface statsTable)
+            throws IOException, InterruptedException {
+        List<Cell> columnCells = metaTable
+                .get(new Get(SchemaUtil.getTableKey(null, PhoenixDatabaseMetaData.SYSTEM_SCHEMA_NAME,
+                        PhoenixDatabaseMetaData.SYSTEM_CATALOG_TABLE)))
+                .getColumnCells(PhoenixDatabaseMetaData.TABLE_FAMILY_BYTES, QueryConstants.EMPTY_COLUMN_BYTES);
+        if (!columnCells.isEmpty()
+                && columnCells.get(0).getTimestamp() < MetaDataProtocol.MIN_SYSTEM_TABLE_TIMESTAMP_4_7_0) {
```
JamesRTaylor: The if isn't needed because the client-side code that truncates the stats table has been removed. This is the only place we do it. If stats building gets triggered *before* the client-side upgrade code has run (for example, through compaction), then it will build using the new logic with the new schema. It should be fine, b/c the code uses straight HBase APIs, not Phoenix APIs. Since we always send back empty guideposts for the protobuf field that old clients will be looking at, clients will just get empty stats until the client-side upgrade code runs.

phoenix-protocol/src/main/PTable.proto
```diff
@@ -57,6 +57,8 @@ message PTableStats {
   optional int64 keyBytesCount = 4;
   optional int32 guidePostsCount = 5;
   optional PGuidePosts pGuidePosts = 6;
+  optional bytes encodedGuidePosts = 7;
```
JamesRTaylor: I'm a bit confused by this, though. Doesn't the PTable still send multiple PGuidePosts (one per column family)? It's just the PGuidePosts that have changed, no? You already have maxLength on PGuidePosts (which is where it belongs). Why is it needed here?

phoenix-core/src/main/java/org/apache/phoenix/util/UpgradeUtil.java
JamesRTaylor: Actually, I take it back. Your way is better.

ankitsinghal:
Thanks @JamesRTaylor for all the review and help.

PR #148, opened 2016-01-19
bugfix: PHOENIX-2454 Upsert with Double.NaN returns NumberFormatException

chuxi:
the bug is on the lira: https://issues.apache.org/jira/browse/PHOENIX-2454

I met the problem when some value in the DataFrame is NaN(Double.NaN and Float.NaN), and it is really important part as storing DataFrame into Phoenix Table. 

So can the community merge this PR? 

Int, Long, and Short, Byte has no NaN in Scala

It has simple logic inside. I have already did basic test. But maybe someone else could test totally. :)


chuxi:
the test. 

screen shot 2016-01-19 at 11 05 41 am


screen shot 2016-01-19 at 11 08 23 am

PR #153, opened 2016-03-22
PHOENIX-1311 HBase namespaces surfaced in phoenix

ankitsinghal:


JamesRTaylor:
Thanks for spinning up a pull request, @ankitsinghal. One functional question: how are you handling the case of a VIEW having a different schema name than it's base/physical table? Since the rows of a view are in the same physical table as the base table, they cannot be in different namespaces, correct?

Please review, @samarthjain.

ankitsinghal:
Currently ,VIEW can be created in different schema name than it's base table name. are you recommending to use same schema?

I checked MySQL and postgres and they also allow creation of VIEW in different namespace/schema than its physical table. (may be it's because VIEW is logical and can be created on joined tables residing in different namespaces [though phoenix doesn't support join query in VIEW currently])

http://dev.mysql.com/doc/refman/5.7/en/create-view.html
http://www.postgresql.org/docs/9.2/static/sql-createview.html
"If a schema name is given (for example, CREATE VIEW myschema.myview ...) then the view is created in the specified schema. Otherwise it is created in the current schema"

What do you think about VIEW Indexes created with different schema as they all share a single physical table for storage and currently with above changes we are keeping a index table in same hbase namespace as of it's base/physical table.

JamesRTaylor:
I just wonder how that'll work, @ankitsinghal. If it works, I suppose it's ok, but it muddies things a bit as views are all supposed to be contained in the same physical table. Do we change the views physical table name property? Do you have test cases around this?

ankitsinghal:
still also, all views will be pointing to same physical table even if they have different schemas . we do update physical table property during the creation of view as per the parent physical table.

I have updated one test ViewIT#testViewAndTableInDifferentSchemas in latest commit. Is this you are expecting?

phoenix-core/src/it/java/org/apache/phoenix/end2end/CreateSchemaIT.java
```diff
@@ -0,0 +1,67 @@
...
+
+public class CreateSchemaIT extends BaseClientManagedTimeIT {
+
+    @Test
+    public void testCreateSchema() throws Exception {
+        long ts = nextTimestamp();
+        Properties props = new Properties();
+        props.setProperty(PhoenixRuntime.CURRENT_SCN_ATTRIB, Long.toString(ts));
+        Connection conn = DriverManager.getConnection(getUrl(), props);
+        HBaseAdmin admin = driver.getConnectionQueryServices(getUrl(), TestUtil.TEST_PROPERTIES).getAdmin();
```
samarthjain: You can also alternatively directly go through PhoenixConnection to get hold of the admin. Like this:

conn.unwrap(PhoenixConnection.class).getQueryServices().getAdmin(). 

Minor nit: Even though this is test only, it is always better to close connections and HBaseAdmin using a try-with-resources construct.

phoenix-core/src/it/java/org/apache/phoenix/end2end/CreateTableIT.java
```diff
@@ -94,6 +99,8 @@ public void testCreateTable() throws Exception {
                 "                id INTEGER not null primary key desc\n" + 
                 "                ) ";
         conn.createStatement().execute(ddl);
+        HBaseAdmin admin = driver.getConnectionQueryServices(getUrl(), props).getAdmin();
+        assertNotEquals(null, admin.getTableDescriptor(Bytes.toBytes(tableName)));
```
samarthjain: Minor nits: assertNotNull reads better here. Same recommendations as above regarding getting and closing HBaseAdmin.

phoenix-core/src/it/java/org/apache/phoenix/end2end/NamespaceSchemaMappingIT.java
```diff
@@ -0,0 +1,105 @@
...
+import org.apache.phoenix.util.TestUtil;
+import org.junit.Test;
+
+public class NamespaceSchemaMappingIT extends BaseHBaseManagedTimeIT {
+
+    @Test
+    @SuppressWarnings("deprecation")
+    public void testBackWardCompatibility() throws Exception {
+
+        String namespace = "TEST_SCHEMA";
```
samarthjain: Its great that you have added tests to check backward compatibility. Can you add some comments to the test here, though. It is difficult to follow right now exactly what this test is doing here.

phoenix-protocol/src/main/MetaDataService.proto
```diff
@@ -72,6 +79,12 @@ message GetFunctionsRequest {
   optional int32 clientVersion = 5;
 }
 
+message GetSchemaRequest {
+  required string schemaName = 1;
+  required int64 clientTimestamp = 2;
+  optional int32 clientVersion = 3;
```
samarthjain: Should the client version be optional or required? Considering this is a new metadata entity, I would think most of the things here should have "required" unless of course functionally they don't have to be present always. This applies to all the other schema related requests.

phoenix-core/src/test/java/org/apache/phoenix/query/BaseTest.java
```diff
@@ -825,9 +825,27 @@ protected static void ensureTableCreated(String url, String tableName, Long ts)
...
+        if (tableName.contains(".")) {
+            String schema = tableName.substring(0, tableName.indexOf("."));
+            if (!schema.equals("")) {
+                Properties props = new Properties();
+                if (ts != null) {
+                    props.setProperty(PhoenixRuntime.CURRENT_SCN_ATTRIB, Long.toString(ts));
+                }
+                try (Connection conn = DriverManager.getConnection(url, props);) {
+                    conn.createStatement().executeUpdate("CREATE SCHEMA IF NOT EXISTS " + schema);
+                } catch (TableAlreadyExistsException e) {
```
samarthjain: Remove the catch() block here as it doesn't apply.

phoenix-core/src/main/java/org/apache/phoenix/schema/PTable.java
```diff
@@ -341,4 +345,13 @@ public static LinkType fromSerializedValue(byte serializedValue) {
...
     long getUpdateCacheFrequency();
+
+    boolean isNamespaceMapped();
+
+    /**
+     * For a view, return the name of table in Phoenix that physically stores data.
+     * 
+     * @return the name of the Phoenix table storing the data.
+     */
+    PName getPhoenixPhysicalName();
```
samarthjain: Can you tell me more about this new method? What is its purpose? Looking at the comments it looks like exactly same as getPhysicalName()

phoenix-core/src/main/java/org/apache/phoenix/parse/PSchema.java
```diff
@@ -0,0 +1,86 @@
...
+        this.schemaName = null;
+    }
+
+    public PSchema(String schemaName) {
+        this(schemaName, HConstants.LATEST_TIMESTAMP);
+    }
+
+    public PSchema(String schemaName, long timeStamp) {
+        this.schemaName = PNameFactory.newName(SchemaUtil.normalizeIdentifier(schemaName));
+        this.schemaKey = new PTableKey(null, this.schemaName.getString());
```
samarthjain: Thinking about this a little bit more, is it possible for tenant views on the same multi-tenant table to have different schema/namespaces? If yes, how would you differentiate these schemas considering the tenantId is being set as null here. At a minimum, it would be ideal to have some testing around this scenario.

phoenix-core/src/main/java/org/apache/phoenix/schema/MetaDataClient.java
```diff
@@ -544,7 +556,28 @@ private MetaDataMutationResult updateCache(PName tenantId, String schemaName, St
...
+
+    public MetaDataMutationResult updateCache(String schemaName) throws SQLException {
+        return updateCache(schemaName, false);
+    }
+
+    public MetaDataMutationResult updateCache(String schemaName, boolean alwaysHitServer) throws SQLException {
+        long clientTimeStamp = getClientTimeStamp();
+        PSchema schema = null;
+        try {
+            schema = connection.getMetaDataCache().getSchema(new PTableKey(null, schemaName));
```
samarthjain: Is tenantId a factor here? Is it always ok to have tenantId as null.

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/MetaDataEndpointImpl.java
```diff
@@ -1115,6 +1212,24 @@ private PFunction loadFunction(RegionCoprocessorEnvironment env, byte[] key,
...
+        Region region = env.getRegion();
+        Cache<ImmutableBytesPtr, PMetaDataEntity> metaDataCache = GlobalCache.getInstance(this.env).getMetaDataCache();
+        PSchema schema = (PSchema)metaDataCache.getIfPresent(cacheKey);
+        // We always cache the latest version - fault in if not in cache
+        if (schema != null) { return schema; }
+        ArrayList<byte[]> arrayList = new ArrayList<byte[]>(1);
+        arrayList.add(key);
+        List<PSchema> schemas = buildSchemas(arrayList, region, asOfTimeStamp, cacheKey);
+        if (schemas != null) return schemas.get(0);
+        // if not found then check if newer table already exists and add delete marker for timestamp
```
samarthjain: Change comment to mention schema and not table.

phoenix-core/src/main/java/org/apache/phoenix/schema/MetaDataClient.java
```diff
@@ -3301,4 +3356,86 @@ private PTable getParentOfView(PTable view) throws SQLException {
...
+                schemaUpsert.execute();
+                schemaMutations = connection.getMutationState().toMutations(null).next().getSecond();
+                connection.rollback();
+            }
+            MetaDataMutationResult result = connection.getQueryServices().createSchema(schemaMutations,
+                    schema.getSchemaName());
+            MutationCode code = result.getMutationCode();
+            switch (code) {
+            case SCHEMA_ALREADY_EXISTS:
+                if (result.getTable() != null) { // Can happen for transactional table that already exists as HBase
```
samarthjain: result.getSchema() ? Also, please remove the comment.

phoenix-core/src/main/java/org/apache/phoenix/schema/PSchemaKey.java
```diff
@@ -0,0 +1,67 @@
+/*
```
samarthjain: Where is this used? I don't see any mention of PSchemaKey outside this class itself.

phoenix-core/src/main/java/org/apache/phoenix/query/ConnectionQueryServicesImpl.java
```diff
@@ -634,15 +642,19 @@ private PMetaData metaDataMutated(PName tenantId, String tableName, long tableSe
         }
      }
 
-    @Override
-    public PMetaData addColumn(final PName tenantId, final String tableName, final List<PColumn> columns, final long tableTimeStamp, 
-            final long tableSeqNum, final boolean isImmutableRows, final boolean isWalDisabled, final boolean isMultitenant, 
-            final boolean storeNulls, final boolean isTransactional, final long updateCacheFrequency, final long resolvedTime) throws SQLException {
-        return metaDataMutated(tenantId, tableName, tableSeqNum, new Mutator() {
+	@Override
```
samarthjain: Is this just white-space diff here? If yes, please revert the change.

phoenix-core/src/main/java/org/apache/phoenix/query/ConnectionQueryServicesImpl.java
```diff
@@ -958,6 +973,32 @@ private boolean allowOnlineTableSchemaUpdate() {
                 QueryServicesOptions.DEFAULT_ALLOW_ONLINE_TABLE_SCHEMA_UPDATE);
     }
 
+    private NamespaceDescriptor ensureNamespaceCreated(String schemaName) throws SQLException {
+        SQLException sqlE = null;
+        try (HBaseAdmin admin = getAdmin()) {
+            final String quorum = ZKConfig.getZKQuorumServersString(config);
```
samarthjain: Remove this two variables and the following logger statement.

phoenix-core/src/main/java/org/apache/phoenix/query/ConnectionQueryServicesImpl.java
```diff
@@ -958,6 +973,32 @@ private boolean allowOnlineTableSchemaUpdate() {
...
+            final String znode = this.props.get(HConstants.ZOOKEEPER_ZNODE_PARENT);
+            logger.debug("Found quorum: " + quorum + ":" + znode);
+            boolean nameSpaceExists = true;
+            NamespaceDescriptor namespaceDescriptor = null;
+            try {
+                namespaceDescriptor = admin.getNamespaceDescriptor(schemaName);
```
samarthjain: How about you try creating the name space and catch the expected NamespaceExistException?

phoenix-core/src/main/java/org/apache/phoenix/schema/PSchemaKey.java
ankitsinghal: It is not needed anymore as we are storing schema in SYSTEM.CATALOG only. So I have removed in latest commit

phoenix-core/src/main/java/org/apache/phoenix/schema/MetaDataClient.java
ankitsinghal: I'm not sure that whether we should keep schema per tenantId so currently it is kept as global , if you or @JamesRTaylor think, then I can make the necessary changes to support schema per tenantId.

phoenix-core/src/main/java/org/apache/phoenix/query/ConnectionQueryServicesImpl.java
ankitsinghal: No, isNamespaceMapped parameter is added.

ankitsinghal:
Thanks @samarthjain for reviews, I have incorporated the changes as you suggested except one: supporting schema per tenantId  , for which I need your confirmation again.

phoenix-core/src/main/java/org/apache/phoenix/schema/MetaDataClient.java
JamesRTaylor: No, we wouldn't need a schema per tenantId. With namespace support, is all data still stored in the same physical table for all views, even if the views have different schemas? For example, given the following:

    CREATE VIEW s1.a AS SELECT * FROM t WHERE k=1;
    CREATE VIEW s2.b AS SELECT * FROM t WHERE k=2;
    CREATE VIEW s3.c AS SELECT * FROM t WHERE k=3;

Would the following query still return all rows across all three views?

    SELECT * FROM t;

phoenix-core/src/main/java/org/apache/phoenix/schema/PTable.java
ankitsinghal: getPhoenixPhysicalName():- This is needed only in case of views when we want phoenix representation(X.Y) of data table.

getPhysicalName():- this is to get physical/hbase representation(X:Y) of any data table/index.

phoenix-core/src/main/java/org/apache/phoenix/schema/MetaDataClient.java
ankitsinghal: yes, union of all above views will be equal to `select * from t`.
as s1.a , s2.b and s3.c are pointing to same physical table "t" only.

JamesRTaylor: But each view will be in a different namespace? I thought namespace was a physical property.

ankitsinghal: views with schema may not be needed to mapped to any namespace right as view does not have any physical presence, so they can reside in different schema referencing to any table(in any namespace/schema). 
And, It seems tat other database like MySQL and Postgres also support views across schemas like this.

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/MetaDataEndpointImpl.java
```diff
@@ -1562,7 +1678,7 @@ private MetaDataMutationResult doDropTable(byte[] key, byte[] tenantId, byte[] s
             }
 
             if (tableType != PTableType.VIEW) { // Add to list of HTables to delete, unless it's a view
-                tableNamesToDelete.add(table.getName().getBytes());
+                tableNamesToDelete.add(table.getPhysicalName().getBytes());
```
JamesRTaylor: There seems to be a lot of these changes of getName to getPhysicalName which I don't believe is correct. The getName() returns the logical name while getPhysicalName() returns the physical table name. Why was this change made?

phoenix-core/src/main/java/org/apache/phoenix/iterate/ChunkedResultIterator.java
```diff
@@ -74,7 +74,7 @@ public ChunkedResultIteratorFactory(ParallelIteratorFactory
 
         @Override
         public PeekingResultIterator newIterator(StatementContext context, ResultIterator scanner, Scan scan, String tableName) throws SQLException {
-            if (logger.isDebugEnabled()) logger.debug(LogUtil.addCustomAnnotations("ChunkedResultIteratorFactory.newIterator over " + tableRef.getTable().getName().getString() + " with " + scan, ScanUtil.getCustomAnnotations(scan)));
+            if (logger.isDebugEnabled()) logger.debug(LogUtil.addCustomAnnotations("ChunkedResultIteratorFactory.newIterator over " + tableRef.getTable().getPhysicalName().getString() + " with " + scan, ScanUtil.getCustomAnnotations(scan)));
```
JamesRTaylor: Another getName -> getPhysicalName. Why?

phoenix-core/src/main/java/org/apache/phoenix/iterate/ExplainTable.java
```diff
@@ -120,7 +120,7 @@ protected void explain(String prefix, List<String> planSteps) {
         } else {
             explainSkipScan(buf);
         }
-        buf.append("OVER ").append(tableRef.getTable().getPhysicalName().getString());
+        buf.append("OVER ").append(tableRef.getTable().getPhoenixPhysicalName().getString());
```
JamesRTaylor: Now here getPhysicalName -> getPhoenixPhysicalName. Why are we introducing a new name method?

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/MetaDataEndpointImpl.java
ankitsinghal: Yes!! it was not correct earlier but it didn't matter because logical and physical representation was same. but now this change is required to get correct physical name so that we can delete descriptors on the client.

phoenix-core/src/main/java/org/apache/phoenix/iterate/ChunkedResultIterator.java
ankitsinghal: I changed it as there is a reference to scan so I thought, physical names in the logs would be required for better debugging. let me know if you think we should keep phoenix names in **logging** as it is.

phoenix-core/src/main/java/org/apache/phoenix/iterate/ExplainTable.java
ankitsinghal: getPhoenixPhysicalName() is required to get phoenix table name (X.Y) , this is different from getName() as in case of view, it return phoenix name of the data table ,which sometimes required for "Explain" plan or sometimes to get tableRef for data table. I think , we can have different name for this method to avoid confusion.

getPhysicalName gives the hbase representation(X:Y) of table/view depending upon isNamespaceMapped flag.

JamesRTaylor: I'd rather just have one getPhysicalName that returns X:Y if namespace enabled. Can we get away with that?

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/MetaDataEndpointImpl.java
JamesRTaylor: Ah, ok. Thanks for the bug fix :-). Please make sure there's a unit test that goes with it.

phoenix-core/src/main/java/org/apache/phoenix/iterate/ChunkedResultIterator.java
JamesRTaylor: Good question. Not sure. Maybe we can do that in a separate JIRA if we want to change it? I'd probably lean toward using HBase physical name everywhere for explain plan and logging.

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/MetaDataEndpointImpl.java
ankitsinghal: Yeah , LocalIndexIT and ViewIT covers dropTable tests.

phoenix-core/src/main/java/org/apache/phoenix/iterate/ChunkedResultIterator.java
ankitsinghal: https://issues.apache.org/jira/browse/PHOENIX-2807
Created a jira for the same for now.

phoenix-core/src/main/java/org/apache/phoenix/iterate/ExplainTable.java
ankitsinghal: yeah.. is possible. did it in my last commit.

ankitsinghal:
@JamesRTaylor @samarthjain , would you mind reviewing 
-- Upgrade utility to map existing table to namespace
-- changes in bulload tool
-- changes in IndexFailurePolicy
-- let me know if you need more test for coverage.

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/MetaDataEndpointImpl.java
```diff
@@ -3201,4 +3367,155 @@ private MetaDataMutationResult doDropFunction(long clientTimeStamp, List<byte[]>
...
+                long currentTimeStamp = MetaDataUtil.getClientTimeStamp(schemaMutations);
+                builder.setReturnCode(MetaDataProtos.MutationCode.SCHEMA_NOT_FOUND);
+                builder.setMutationTime(currentTimeStamp);
+                done.run(builder.build());
+                return;
+            } finally {
+                region.releaseRowLocks(locks);
+            }
+        } catch (Throwable t) {
+            logger.error("createFunction failed", t);
```
samarthjain: Change the message to:
"Creating the schema " + schemaName + " failed."

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/MetaDataEndpointImpl.java
```diff
@@ -2910,6 +3026,16 @@ private static MetaDataMutationResult checkTableKeyInRegion(byte[] key, Region r
                 EnvironmentEdgeManager.currentTimeMillis(), null);
     }
 
+    private static MetaDataMutationResult checkSchemaKeyInRegion(byte[] key, Region region) {
```
samarthjain: You should reuse checkTableKeyInRegion and rename it to checkKeyInRegion. You would pass MutationCode to include in the result an extra parameter when the key is not in region. While you are it, also please remove the method checkFunctionKeyInRegion.

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/MetaDataEndpointImpl.java
```diff
@@ -3201,4 +3367,155 @@ private MetaDataMutationResult doDropFunction(long clientTimeStamp, List<byte[]>
...
+            if (result != null) {
+                done.run(MetaDataMutationResult.toProto(result));
+                return;
+            }
+            List<RowLock> locks = Lists.newArrayList();
+            long clientTimeStamp = MetaDataUtil.getClientTimeStamp(schemaMutations);
+            try {
+                acquireLock(region, lockKey, locks);
+                // Get as of latest timestamp so we can detect if we have a
+                // newer function that already
```
samarthjain: newer schema. Also, please format the comments.

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/MetaDataEndpointImpl.java
```diff
@@ -3201,4 +3367,155 @@ private MetaDataMutationResult doDropFunction(long clientTimeStamp, List<byte[]>
...
+                        builder.setMutationTime(EnvironmentEdgeManager.currentTimeMillis());
+                        builder.setSchema(PSchema.toProto(schema));
+                        done.run(builder.build());
+                        return;
+                    }
+                }
+                region.mutateRowsWithLocks(schemaMutations, Collections.<byte[]> emptySet(), HConstants.NO_NONCE,
+                        HConstants.NO_NONCE);
+
+                // Invalidate the cache - the next getTable call will add it
```
samarthjain: Please update these comments.

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/MetaDataEndpointImpl.java
```diff
@@ -3201,4 +3367,155 @@ private MetaDataMutationResult doDropFunction(long clientTimeStamp, List<byte[]>
...
+                    if (Bytes.compareTo(kv.getRowArray(), kv.getRowOffset(), kv.getRowLength(), key, 0,
+                            key.length) != 0) {
+                        areTablesExists = true;
+                        break;
+                    }
+                    results.clear();
+                    scanner.next(results);
+                } while (!results.isEmpty());
+            }
+            if (areTablesExists) { return new MetaDataMutationResult(MutationCode.UNALLOWED_SCHEMA_MUTATION, schema,
```
samarthjain: I think it would be better to throw a more specific mutation code here. Something like TABLES_EXIST_ON_SCHEMA and then have proper handling in MetadataClient.dropSchema to throw the right SQLExceptionCode with appropriate message.

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/MetaDataEndpointImpl.java
```diff
@@ -3201,4 +3367,155 @@ private MetaDataMutationResult doDropFunction(long clientTimeStamp, List<byte[]>
...
+                    EnvironmentEdgeManager.currentTimeMillis()); }
+
+            return new MetaDataMutationResult(MutationCode.SCHEMA_ALREADY_EXISTS, schema,
+                    EnvironmentEdgeManager.currentTimeMillis());
+        }
+        return new MetaDataMutationResult(MutationCode.SCHEMA_NOT_FOUND, EnvironmentEdgeManager.currentTimeMillis(),
```
samarthjain: This is already returned over here:
if (schema == null) { return new MetaDataMutationResult(MutationCode.SCHEMA_NOT_FOUND,
                EnvironmentEdgeManager.currentTimeMillis(), null); 
}

May be move the SCHEMA_ALREADY_EXISTS out of the if block or some other way?

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/MetaDataEndpointImpl.java
JamesRTaylor: I think it'd be best if MutationCode returned an int error code as we can go directly from that to a SQLExceptionCode on the client. We can use negative numbers for non exception cases.

JamesRTaylor: These changes can be done in a separate pull request, though IMO.

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/MetaDataEndpointImpl.java
ankitsinghal: I think SCHEMA_ALREADY_EXISTS is needed in if block , because we are checking if the schema timestamp is less than the clientstamp otherwise we are returning MutationCode.SCHEMA_NOT_FOUND

phoenix-core/src/main/java/org/apache/phoenix/schema/MetaDataClient.java
JamesRTaylor: I see - so regardless of the schema used for a view, the physical data will still be stored in the one physical table. The schema in this case for the view does not map to an HBase namespace. Is that correct?

ankitsinghal: Yes @JamesRTaylor  .. Absolutely correct !!

ankitsinghal:
@samarthjain @JamesRTaylor , any update on further review?

phoenix-core/src/main/java/org/apache/phoenix/compile/FromCompiler.java
```diff
@@ -265,12 +332,18 @@ public SingleTableColumnResolver(PhoenixConnection connection, NamedTableNode ta
                if (def.getColumnDefName().getFamilyName() != null) {
                    families.add(new PColumnFamilyImpl(PNameFactory.newName(def.getColumnDefName().getFamilyName()),Collections.<PColumn>emptyList()));
                }
-           }
-           Long scn = connection.getSCN();
-           PTable theTable = new PTableImpl(connection.getTenantId(), table.getName().getSchemaName(), table.getName().getTableName(), scn == null ? HConstants.LATEST_TIMESTAMP : scn, families);
+            }
+            Long scn = connection.getSCN();
+            String schema = table.getName().getSchemaName();
+            if (connection.getSchema() != null) {
```
samarthjain: Not sure if functionally this is the right thing to do here. I would expect Phoenix to thrown an exception when the connection's setting of schema/namespace is different from the the table's namespace/schema. Or is my understanding not correct here? Would be good to add a test case for this too in UseSchemaIT.

phoenix-core/src/test/java/org/apache/phoenix/query/BaseTest.java
```diff
@@ -825,9 +825,25 @@ protected static void ensureTableCreated(String url, String tableName, Long ts)
 
     protected static void ensureTableCreated(String url, String tableName, byte[][] splits, Long ts) throws SQLException {
         String ddl = tableDDLMap.get(tableName);
+        createSchema(url,tableName, ts);
         createTestTable(url, ddl, splits, ts);
     }
 
+    public static void createSchema(String url, String tableName, Long ts) throws SQLException {
+        if (tableName.contains(".")) {
```
samarthjain: Use SchemaUtil.getSchemaNameFromFullName and related methods here to parse the schema name and table name.

phoenix-core/src/main/java/org/apache/phoenix/util/SchemaUtil.java
```diff
@@ -897,4 +941,86 @@ public static boolean hasRowTimestampColumn(PTable table) {
...
+
+    public static boolean isSchemaCheckRequired(PTableType tableType, ReadOnlyProps props) {
+        if (PTableType.TABLE.equals(tableType) && isNamespaceMappingEnabled(tableType, props)) { return true; }
+        return false;
+    }
+
+    public static boolean isNamespaceMappingEnabled(PTableType type, ReadOnlyProps readOnlyProps) {
+        return readOnlyProps.getBoolean(QueryServices.IS_NAMESPACE_MAPPING_ENABLED,
+                QueryServicesOptions.DEFAULT_IS_NAMESPACE_MAPPING_ENABLED)
+                && (type == null || !PTableType.SYSTEM.equals(type)
```
samarthjain: Can the type be null here. If not, I would annotate the argument as @Nonnull and have a Preconditions.checkNotNull(type) check at the beginning of the method.

phoenix-core/src/main/java/org/apache/phoenix/util/SchemaUtil.java
```diff
@@ -897,4 +941,86 @@ public static boolean hasRowTimestampColumn(PTable table) {
...
+    }
+
+    public static PName getPhysicalHBaseTableName(String tableName, boolean isNamespaceMapped, PTableType type) {
+        if (!isNamespaceMapped) { return PNameFactory.newName(tableName); }
+        return PNameFactory
+                .newName(tableName.replace(QueryConstants.NAME_SEPARATOR, QueryConstants.NAMESPACE_SEPARATOR));
+    }
+
+    public static boolean isSchemaCheckRequired(PTableType tableType, ReadOnlyProps props) {
+        if (PTableType.TABLE.equals(tableType) && isNamespaceMappingEnabled(tableType, props)) { return true; }
```
samarthjain: nit: how about 
return PTableType.TABLE.equals(tableType) && isNamespaceMappingEnabled(tableType, props);

phoenix-core/src/main/java/org/apache/phoenix/util/UpgradeUtil.java
```diff
@@ -1279,4 +1304,129 @@ public static boolean truncateStats(HTableInterface metaTable, HTableInterface s
         }
         return false;
     }
+
+    public static void mapTableToNamespace(HBaseAdmin admin, HTableInterface metatable, String srcTableName,
+            String destTableName, ReadOnlyProps props, Long ts, String phoenixTableName, PTableType pTableType)
+                    throws SnapshotCreationException, IllegalArgumentException, IOException, InterruptedException,
+                    SQLException {
+        srcTableName = SchemaUtil.normalizeIdentifier(srcTableName);
+        if (!SchemaUtil.isNamespaceMappingEnabled(
```
samarthjain: Can you not just pass the pTableType argument here?

phoenix-core/src/main/java/org/apache/phoenix/util/UpgradeUtil.java
```diff
@@ -1279,4 +1304,129 @@ public static boolean truncateStats(HTableInterface metaTable, HTableInterface s
...
+                                + " also needs to be enabled along with " + QueryServices.IS_NAMESPACE_MAPPING_ENABLED
+                        : QueryServices.IS_NAMESPACE_MAPPING_ENABLED + " is not enabled"); }
+
+        if (PTableType.TABLE.equals(pTableType) || PTableType.INDEX.equals(pTableType)) {
+            admin.snapshot(srcTableName, srcTableName);
+            admin.cloneSnapshot(srcTableName.getBytes(), destTableName.getBytes());
+            admin.disableTable(srcTableName);
+            admin.deleteTable(srcTableName);
+        }
+        if (phoenixTableName == null) {
```
samarthjain: This looks a bit hacky/unclear to me. When can phoenixTableName be null here? And when is it ok to use the srcTableName. At a minimum some method comments would help.

phoenix-core/src/main/java/org/apache/phoenix/util/UpgradeUtil.java
```diff
@@ -1279,4 +1304,129 @@ public static boolean truncateStats(HTableInterface metaTable, HTableInterface s
...
+                    InterruptedException, SQLException {
+        String destTablename = SchemaUtil
+                .normalizeIdentifier(SchemaUtil.getPhysicalTableName(tableName, props).getNameAsString());
+        mapTableToNamespace(admin, metatable, tableName, destTablename, props, ts, null, PTableType.TABLE);
+    }
+
+    public static void upgradeTable(PhoenixConnection conn, String srcTable) throws SQLException,
+            SnapshotCreationException, IllegalArgumentException, IOException, InterruptedException {
+        ReadOnlyProps readOnlyProps = conn.getQueryServices().getProps();
+        if (conn.getClientInfo(PhoenixRuntime.TENANT_ID_ATTRIB) != null) { throw new SQLException(
```
samarthjain: What if the connection has schema property set? Should we report an error. Looks like currently we are ignoring it.

phoenix-core/src/main/java/org/apache/phoenix/util/UpgradeUtil.java
```diff
@@ -1279,4 +1304,129 @@ public static boolean truncateStats(HTableInterface metaTable, HTableInterface s
...
+                    String srcTableName = index.getPhysicalName().getString();
+                    if(srcTableName.contains(QueryConstants.NAMESPACE_SEPARATOR)){
+                        //this condition occurs in case of multiple views on table
+                        //skip already migrated tables 
+                        continue;
+                    }
+                    String destTableName = null;
+                    String phoenixTableName = index.getName().getString();
+                    boolean updateLink = false;
+                    if (srcTableName.startsWith(MetaDataUtil.LOCAL_INDEX_TABLE_PREFIX)) {
```
samarthjain: Would be good to add a method like isLocalIndex in SchemaUtil

phoenix-core/src/main/java/org/apache/phoenix/util/UpgradeUtil.java
```diff
@@ -1279,4 +1304,129 @@ public static boolean truncateStats(HTableInterface metaTable, HTableInterface s
...
+                    SchemaUtil.getTableNameFromFullName(tableName));
+            if (result.getMutationCode() != MutationCode.TABLE_ALREADY_EXISTS) { throw new TableNotFoundException(
+                    tableName); }
+            table = result.getTable();
+            // check whether table is properly upgraded before upgrading indexes
+            if (table.isNamespaceMapped()) {
+                for (PTable index : table.getIndexes()) {
+                    String srcTableName = index.getPhysicalName().getString();
+                    if(srcTableName.contains(QueryConstants.NAMESPACE_SEPARATOR)){
+                        //this condition occurs in case of multiple views on table
```
samarthjain: Not sure I follow the comment here. Can you please elaborate?

phoenix-core/src/main/java/org/apache/phoenix/util/UpgradeUtil.java
```diff
@@ -1279,4 +1304,129 @@ public static boolean truncateStats(HTableInterface metaTable, HTableInterface s
...
+            conn.createStatement().execute("CREATE SCHEMA IF NOT EXISTS " + schemaName);
+            String newPhysicalTablename = SchemaUtil
+                    .normalizeIdentifier(SchemaUtil.getPhysicalTableName(table.getPhysicalName().getString(), readOnlyProps).getNameAsString());
+
+            // Upgrade the data or main table
+            UpgradeUtil.mapTableToNamespace(admin, metatable, tableName, newPhysicalTablename, readOnlyProps,
+                    PhoenixRuntime.getCurrentScn(readOnlyProps), tableName, table.getType());
+
+            // clear the cache and get new table
+            conn.getQueryServices().clearCache();
```
samarthjain: I think you should be calling clearTableFromCache() here instead.

phoenix-core/src/main/java/org/apache/phoenix/util/UpgradeUtil.java
```diff
@@ -1279,4 +1304,129 @@ public static boolean truncateStats(HTableInterface metaTable, HTableInterface s
...
+                                .getPhysicalTableName(index.getPhysicalName().getString(), readOnlyProps)
+                                .getNameAsString();
+                    }
+                    if (updateLink) {
+                        updateLink(conn, srcTableName, destTableName);
+                    }
+                    UpgradeUtil.mapTableToNamespace(admin, metatable, srcTableName, destTableName, readOnlyProps,
+                            PhoenixRuntime.getCurrentScn(readOnlyProps), phoenixTableName, index.getType());
+                }
+                conn.getQueryServices().clearCache();
```
samarthjain: Similarly here. Call clearTableFromCache.

phoenix-core/src/main/java/org/apache/phoenix/util/UpgradeUtil.java
```diff
@@ -1279,4 +1304,129 @@ public static boolean truncateStats(HTableInterface metaTable, HTableInterface s
...
+
+            // Confirm table is not already upgraded
+            PTable table = PhoenixRuntime.getTable(conn, tableName);
+            if (table.isNamespaceMapped()) { throw new IllegalArgumentException("Table is already upgraded"); }
+            conn.createStatement().execute("CREATE SCHEMA IF NOT EXISTS " + schemaName);
+            String newPhysicalTablename = SchemaUtil
+                    .normalizeIdentifier(SchemaUtil.getPhysicalTableName(table.getPhysicalName().getString(), readOnlyProps).getNameAsString());
+
+            // Upgrade the data or main table
+            UpgradeUtil.mapTableToNamespace(admin, metatable, tableName, newPhysicalTablename, readOnlyProps,
```
samarthjain: minor nit: no need to use UpgradeUtil. here

phoenix-core/src/main/java/org/apache/phoenix/util/UpgradeUtil.java
```diff
@@ -1279,4 +1304,129 @@ public static boolean truncateStats(HTableInterface metaTable, HTableInterface s
         }
         return false;
     }
+
+    public static void mapTableToNamespace(HBaseAdmin admin, HTableInterface metatable, String srcTableName,
```
samarthjain: Is it possible to add any tests around the upgrade code here?

phoenix-core/src/main/java/org/apache/phoenix/util/UpgradeUtil.java
```diff
@@ -1279,4 +1304,129 @@ public static boolean truncateStats(HTableInterface metaTable, HTableInterface s
...
+                SchemaUtil.isSystemTable(srcTableName.getBytes()) ? PTableType.SYSTEM : null,
+                props)) { throw new IllegalArgumentException(SchemaUtil.isSystemTable(srcTableName.getBytes())
+                        ? "For system table " + QueryServices.IS_SYSTEM_TABLE_MAPPED_TO_NAMESPACE
+                                + " also needs to be enabled along with " + QueryServices.IS_NAMESPACE_MAPPING_ENABLED
+                        : QueryServices.IS_NAMESPACE_MAPPING_ENABLED + " is not enabled"); }
+
+        if (PTableType.TABLE.equals(pTableType) || PTableType.INDEX.equals(pTableType)) {
+            admin.snapshot(srcTableName, srcTableName);
+            admin.cloneSnapshot(srcTableName.getBytes(), destTableName.getBytes());
+            admin.disableTable(srcTableName);
```
samarthjain: Does disabling and deleting the srcTable here also delete the snapshot you created above. If not, would it make sense to delete the snapshot? This is also making me think of failure scenarios. What if any of the steps here fails? Do we restore the table using the snapshot?

phoenix-core/src/main/java/org/apache/phoenix/util/SchemaUtil.java
ankitsinghal: Yes, null is allowed for type here.

phoenix-core/src/main/java/org/apache/phoenix/util/UpgradeUtil.java
ankitsinghal: Deleting srcTable will not delete the snapshot. 
And, Yes for restore only ,I thought we should not delete snapshot

phoenix-core/src/main/java/org/apache/phoenix/util/UpgradeUtil.java
ankitsinghal: phoenixTableName and physical table name are different in case of indexes .. so it is required there .. but now in my latest commit , I have removed the null check and pass the phoenixTableName always.

phoenix-core/src/main/java/org/apache/phoenix/util/UpgradeUtil.java
ankitsinghal: yes .. Now, I'm throwing exception if schema is set in connection.

phoenix-core/src/main/java/org/apache/phoenix/util/UpgradeUtil.java
ankitsinghal: Yeah, I'll add a test for upgrade code.

phoenix-core/src/main/java/org/apache/phoenix/compile/FromCompiler.java
ankitsinghal: Actually, when the schema for the table{T} is not present, we try to resolve table with connection schema {<connection_schema>.T} if set in connection.
I have added a test case UseSchemaIT#testMappedView, is it you are expecting?

ankitsinghal:
Thanks @samarthjain  for the review, I have incorporated the review comments in last commit.
I'll write a test for upgrade code soon though I have tested it on local cluster for tables with views and indexes.
please proceed with your further review.

ankitsinghal:
@samarthjain , any more review comments?

phoenix-core/src/main/java/org/apache/phoenix/compile/FromCompiler.java
samarthjain: I meant if the connection has a schema present, and the table's schema is different from the schema property in the connection, IMHO we should probably throw an error like SchemaMismatchException. I could think of a few combinations that we should validate and test:
is namespace mapped                 connection schema
yes                                                    null
yes                                                    different
yes                                                    same
no                                                      null
no                                                      non-null

You could probably add a method in SchemaUtil to do the above checks.

phoenix-core/src/it/java/org/apache/phoenix/end2end/QueryDatabaseMetaDataIT.java
```diff
@@ -190,14 +190,8 @@ public void testSchemaMetadataScan() throws SQLException {
 
         rs = dbmd.getSchemas(null, null);
         assertTrue(rs.next());
-        assertEquals(rs.getString("TABLE_SCHEM"),null);
```
samarthjain: Is there a reason why these lines are removed?

phoenix-core/src/main/java/org/apache/hadoop/hbase/ipc/controller/MetadataRpcController.java
```diff
@@ -36,7 +37,16 @@
 			.add(PhoenixDatabaseMetaData.SYSTEM_CATALOG_NAME)
 			.add(PhoenixDatabaseMetaData.SYSTEM_STATS_NAME)
 			.add(PhoenixDatabaseMetaData.SYSTEM_SEQUENCE_NAME)
-			.add(PhoenixDatabaseMetaData.SYSTEM_FUNCTION_NAME).build();
+			.add(PhoenixDatabaseMetaData.SYSTEM_FUNCTION_NAME)
+            .add(SchemaUtil.getPhysicalTableName(PhoenixDatabaseMetaData.SYSTEM_CATALOG_NAME_BYTES, true)
```
samarthjain: is the namespace mapping enabled always for system tables (which is what I am inferring from true being passed from isNamespaceMapped) ? Looking at QueryServicesOptions.java, looks like it isn't.

phoenix-core/src/main/java/org/apache/phoenix/query/ConnectionQueryServicesImpl.java
```diff
@@ -2387,8 +2452,27 @@ public Void call() throws Exception {
                                     logger.info("Update of SYSTEM.CATALOG complete");
 									clearCache();
                                 }
-                                
+
+                                if (currentServerSideTableTimeStamp < MetaDataProtocol.MIN_SYSTEM_TABLE_TIMESTAMP_4_8_0) {
+                                    // Add these columns one at a time, each with different timestamps so that if folks
```
samarthjain: Please remove this comment as it doesn't make sense here.

phoenix-core/src/main/java/org/apache/phoenix/query/QueryServicesOptions.java
```diff
@@ -133,11 +135,13 @@
...
     // latency and less parallelization.
     public static final long DEFAULT_SCAN_RESULT_CHUNK_SIZE = 2999;
+    public static final boolean DEFAULT_IS_NAMESPACE_MAPPING_ENABLED = false;
+    public static final boolean DEFAULT_IS_SYSTEM_TABLE_MAPPED_TO_NAMESPACE = false;
 
     //
     // Spillable GroupBy - SPGBY prefix
     //
-    // Enable / disable spillable group by
+    // Enable / disablfalsellable group by
```
samarthjain: Please undo.

phoenix-core/src/main/java/org/apache/phoenix/schema/MetaDataClient.java
```diff
@@ -631,7 +664,7 @@ private boolean addIndexesFromPhysicalTable(MetaDataMutationResult result, Long
         if (view.getType() != PTableType.VIEW || view.getViewType() == ViewType.MAPPED) {
             return false;
         }
-        String physicalName = view.getPhysicalName().getString();
+        String physicalName = view.getPhysicalNames().get(0).toString();
```
samarthjain: Please undo this change as we still don't allow creating views over multiple tables.

phoenix-core/src/main/java/org/apache/phoenix/schema/PTableKey.java
```diff
@@ -26,7 +28,8 @@
     public PTableKey(PName tenantId, String name) {
         Preconditions.checkNotNull(name);
         this.tenantId = tenantId;
-        this.name = name;
+        this.name = !name.contains(QueryConstants.NAMESPACE_SEPARATOR) ? name
```
samarthjain: So if the table name was A:B, the key would become A.B. Can you add some test case around this and make sure creating, query, upserting to a table named like this works fine.

phoenix-core/src/main/java/org/apache/phoenix/schema/stats/StatisticsCollectorFactory.java
```diff
@@ -63,6 +64,10 @@ public static StatisticsCollector createStatisticsCollector(
         DISABLE_STATS.add(TableName.valueOf(PhoenixDatabaseMetaData.SYSTEM_FUNCTION_NAME));
         DISABLE_STATS.add(TableName.valueOf(PhoenixDatabaseMetaData.SYSTEM_SEQUENCE_NAME));
         DISABLE_STATS.add(TableName.valueOf(PhoenixDatabaseMetaData.SYSTEM_STATS_NAME));
+        DISABLE_STATS.add(SchemaUtil.getPhysicalTableName(PhoenixDatabaseMetaData.SYSTEM_CATALOG_NAME_BYTES,true));
```
samarthjain: Same as in the other place. Is name space mapping always true?

phoenix-core/src/main/java/org/apache/phoenix/util/PhoenixRuntime.java
```diff
@@ -632,6 +645,16 @@ public static ExecutionCommand parseArgs(String[] args) {
             return execCmd;
         }
 
+        private static String validateTableName(String tableName) {
```
samarthjain: Looking at this method it seems like table names can't have : in them. Does the change in PTableKey.java still make sense then?

phoenix-core/src/it/java/org/apache/phoenix/tx/TxCheckpointIT.java
```diff
@@ -96,6 +97,7 @@ public void testUpsertSelectDoesntSeeUpsertedData() throws Exception {
         Connection conn = DriverManager.getConnection(getUrl(), props);
         conn.setAutoCommit(true);
         conn.createStatement().execute("CREATE SEQUENCE "+seqName);
+        BaseTest.createSchema(getUrl(), fullTableName, null);
```
samarthjain: Is this change needed?

phoenix-core/src/it/java/org/apache/phoenix/end2end/QueryDatabaseMetaDataIT.java
ankitsinghal: As now , we will show only those schemas which are created with "create schema" command and having table as empty string. so null is not expected

phoenix-core/src/main/java/org/apache/hadoop/hbase/ipc/controller/MetadataRpcController.java
ankitsinghal: yeah, namespace mapping for system tables will not always be true as it is controlled with phoenix.connection.mapSystemTablesToNamespace, 
but this is just the list of system tables, so it includes SYSTEM.CATAOG and SYSTEM:CATALOG both to set the priority for RPC accordingly. 
do you think, it could have any impact?

phoenix-core/src/main/java/org/apache/phoenix/schema/MetaDataClient.java
ankitsinghal: Yes, we don't support views over multiple tables that's why I used get(0) only. But I have undo it by modifying the other api's to resolve schema and table-name correctly even if namespace separator is present. like SchemaUtil.getTableNameFromFullName()

phoenix-core/src/main/java/org/apache/phoenix/schema/PTableKey.java
ankitsinghal: We are still maintaining phoenix table name(A.B) in cache. so that's why if by mistake physical table is passed as a key , it should be resolve correctly. I think I just kept this for precaution only. LocalIndexIT/ViewIndexIT covers it right?

phoenix-core/src/main/java/org/apache/phoenix/schema/stats/StatisticsCollectorFactory.java
ankitsinghal: same as above, it just to avoid collecting stats for system table, so list contains both.

phoenix-core/src/main/java/org/apache/phoenix/util/PhoenixRuntime.java
ankitsinghal: no this is to validate during upgrade only, if user is passing tablename with ":" by mistake in thought of hbase tablename.

phoenix-core/src/it/java/org/apache/phoenix/tx/TxCheckpointIT.java
ankitsinghal: no, this is not needed , I have removed it.

phoenix-core/src/main/java/org/apache/phoenix/compile/FromCompiler.java
ankitsinghal: The code in discussion will be used during **creation** of **mapped views** only .. It will never be used anywhere else.
So when user provide table name without schema then connection schema should be used if set and if user is providing tablename with schema then we should ignore connection schema. I think this is how most of the databases work.

for eg:- connection schema is set to 'S' (`phoenix> USE 'S'`)
so , if user create mapped view,
`create view A.T(pk ...)`. //then this will map to A.T table only
`create view T(pk..)` // //then this will map to S.T table

And to map table to default schema (User can unset connection schema back to null by using `USE DEFAULT`)
create view T(pk..) // //then this will map to T table

May be I'm missing something, would you mind giving some examples in terms of sql statement.

JamesRTaylor: Yes, this sounds correct, @ankitsinghal. Thanks for clarifying.

samarthjain: Makes sense. Thanks for the clarification @ankitsinghal

ankitsinghal:
@samarthjain , added upgrade util test case too.. please look at the replies on your review comments and let me know if they are fine.

samarthjain:
@ankitsinghal - thanks for adding the upgrade util test. I have a question/suggestion: we should probably disallow users to execute create and use schema statements if the name space feature isn't enabled. I see that you are currently failing the upgrade if the name space feature isn't enabled which is good. To be consistent we should do the same for create and use schema statements. Another thing to think would be whether we should allow setting schema property on connection if name space feature is disabled. I lean towards not allowing it.

phoenix-core/src/main/java/org/apache/phoenix/util/UpgradeUtil.java
samarthjain: Sounds like we would leaving snapshots dangling around. We should remove snapshots once we have successfully upgraded the table. If the user wants to be cautious he/she can always snapshot their table/s first before running this upgrade. In case of failure, we should try restoring the table/s using the snapshot that Phoenix has taken and attempt to delete  the snapshot after the attempt. We should try and not leavE any residual state from the upgrade, IMHO.

phoenix-core/src/main/java/org/apache/phoenix/schema/PTableKey.java
samarthjain: I think we could potentially mask some bug or have it fail later than it should if we do this. We should remove this check altogether.

ankitsinghal: agreed.. removed in latest commit.

phoenix-core/src/main/java/org/apache/phoenix/util/UpgradeUtil.java
ankitsinghal: ok.. I have added a code to delete the snapshot.

Re-trying from the phoenix snapshot would be risky.. if let's say, upgrade got failed and user keep on using the un-mapped table for a while and then he again go for upgrade, the snapshot taken in last upgrade will be obsolete.

ankitsinghal:
@samarthjain , thanks for the review. 
I'm unsure about disabling schema constructs if isNamespaceMapping is not enabled in config. As this property is client side property and even if it is unset , user can still access the table mapped to namespace created by another client which has this property set. This property just ensure schema mapping to namespace during creation of table.

I think these constructs still make sense independently. but @JamesRTaylor / @samarthjain , if you guys think of disabling it , I'm ok with that too.

samarthjain:
Thanks for the updated request, @ankitsinghal. Almost there! Can you tell me more about what would happen in the following scenario if the name space feature is disabled in config:

CREATE SCHEMA S;
USE SCHEMA S;
CREATE TABLE T;

Will the table T be created in the namespace S? If the feature is disabled, it shouldn't be. Would you mind adding a test for this if it hasn't been already?

samarthjain:
Also, how about if we do a CREATE TABLE SCHEMA.TABLENAME with the feature on. Does it end up creating the namespace named SCHEMA first and then maps the table to that namespace. What about if the feature is off. Would be nice to have tests around this or if you already have them can you point me out which ones are testing these? Thanks!

JamesRTaylor:
There are likely lots of issues if isNamespaceMapping config property changes from true to false, no? Seems like it's meant to enable/disable the feature. Wouldn't existing namespace mapped tables not be found if it was changed from true to false after tables had been created? What would be the purpose of allowing a schema (namespace) to be created if the feature is off? If none, then it's probably best to give an error message if isNamespaceMapping is off and CREATE SCHEMA is used.

ankitsinghal:
@samarthjain 

> Will the table T be created in the namespace S? If the feature is disabled, it shouldn't be. Would you mind adding a test for this if it hasn't been already?

Yes, table T will not be created in namespace S if isNamespaceMapped is disabled. Test is there in CreateTableIT#testCreateTable

> Also, how about if we do a CREATE TABLE SCHEMA.TABLENAME with the feature on. Does it end up creating the namespace named SCHEMA first and then maps the table to that namespace

If features is on and SCHEMA is not already present it will fail with SchemaNotFoundException. User has to create a schema first and then he can create a table if the config is set to true.
but when config is set to false, it will create a table with name SCHEMA.TABLENAME in default namespace without throwing any SchemaNotFoundException.
I have added a test(CreateTableIT#testCreateTableWithoutSchema) in the latest commit for the same.

ankitsinghal:
@JamesRTaylor ,

> There are likely lots of issues if isNamespaceMapping config property changes from true to false, no? 

There will not be any problem if **phoenix.connection.isNamespaceMappingEnabled** is set from true to false unless the system tables are migrated after setting **"phoenix.connection.mapSystemTablesToNamespace"** to true as it will migrate the SYSTEM tables to SYSTEM namespace and it is need to be set at client and server both to have IndexFailure, stats collection to work correctly which directly refers system tables at server. Let me know if name of the properties needs to be change to depict it as server and client side property.

>  Wouldn't existing namespace mapped tables not be found if it was changed from true to false after tables had been created?

Tables mapped will still be accessible. Although old client(<v4.8) will not be able to access tables mapped to namespace

> What would be the purpose of allowing a schema (namespace) to be created if the feature is off?If none, then it's probably best to give an error message if isNamespaceMapping is off and CREATE SCHEMA is used.

Yes, I think we can through exception during CREATE SCHEMA only if the property is not set .
but other constructs (like USE <SCHEMA>) should still be allowed as mapped tables are accessible even isNamespaceMappingEnabled is set to false.

JamesRTaylor:
Sounds good, @ankitsinghal. I think we'll be good to go after these minor changes. Thanks for all your excellent work here.

ankitsinghal:
@JamesRTaylor , I am done with the changes in last commit. so is it good for commit now?

JamesRTaylor:
Yes, LGTM @ankitsinghal. @samarthjain - how about you?

phoenix-core/src/main/java/org/apache/phoenix/schema/IllegalOperationException.java
```diff
@@ -0,0 +1,45 @@
...
+import org.apache.phoenix.exception.SQLExceptionInfo;
+
+/**
+ * 
+ * Exception thrown when any illegal operation is performed.
+ *
+ * 
+ * @since 180
+ */
+public class IllegalOperationException extends RuntimeException {
```
samarthjain: Not sure if you need a new exception for this. Can you not just throw a SQLException instead?

phoenix-core/src/main/java/org/apache/phoenix/exception/SQLExceptionCode.java
```diff
@@ -77,6 +79,7 @@ public SQLException newException(SQLExceptionInfo info) {
     MISSING_MAX_LENGTH(207, "22004", "Max length must be specified for type."),
     NONPOSITIVE_MAX_LENGTH(208, "22006", "Max length must have a positive length for type."),
     DECIMAL_PRECISION_OUT_OF_RANGE(209, "22003", "Decimal precision outside of range. Should be within 1 and " + PDataType.MAX_PRECISION + "."),
+    ILLEGAL_OPERATION(210, "22010", "Illegal Operation."),
```
samarthjain: Change this to something like this:
CREATE_SCHEMA_NOT_ALLOWED(210, "22010", "Cannot create schema because config " + QueryServicesOptions. IS_NAMESPACE_MAPPING_ENABLED "  + " for enabling name space mapping isn't enabled.")

phoenix-core/src/main/java/org/apache/phoenix/query/ConnectionQueryServicesImpl.java
samarthjain: This comment is just a copy paste of the comment in the previous if block. It would make more sense to move this comment to above if (currentServerSideTableTimeStamp < MetaDataProtocol.MIN_SYSTEM_TABLE_TIMESTAMP_4_6_0) {} block as it applies to all the upgrades we will do in future.

samarthjain:
This looks great now @ankitsinghal. I am +1 once you have addressed the last review comments. Thanks a lot for your patience and diligence!

ankitsinghal:
I have done the changes in last commit.
Thanks you so much @samarthjain and @JamesRTaylor for taking time and reviewing it. I'll just get this committed.

jimdowling:
Is anything happening on this branch? We're interested in this feature for multi-tenant access to Phoenix...

ankitsinghal:
This pull request was already merged and feature is available v4.8.0

PR #154, opened 2016-03-22
PHOENIX-2722 support mysql limit,offset clauses

ankitsinghal:


JamesRTaylor:
Would you mind reviewing this, @maryannxue as you'd be the best person to catch any corner cases missed, in particular around joins.

phoenix-core/src/main/java/org/apache/phoenix/compile/SubselectRewriter.java
```diff
@@ -202,14 +204,30 @@ private SelectStatement flatten(SelectStatement select, SelectStatement subselec
...
+                Integer offsetValue = OffsetCompiler.compile(null, select);
+                Integer offsetValueSubselect = OffsetCompiler.compile(null, subselect);
+                if (offsetValue != null && offsetValueSubselect != null) {
+                    offsetRewrite = offsetValue < offsetValueSubselect ? offset : offsetRewrite;
+                } else {
+                    return select;
+                }
+            }
+        }
+
```
maryannxue: Not sure if this would be correct. Suppose we have "select * from (select a from t offset 2 limit 8) offset 3", guess we should return the 5th row instead. If you consider the logic is too complicated to optimize at compiletime, you can simply quit flattening.

phoenix-core/src/main/java/org/apache/phoenix/compile/QueryCompiler.java
```diff
@@ -324,10 +328,13 @@ protected QueryPlan compileJoinQuery(StatementContext context, List<Object> bind
...
             Expression postJoinFilterExpression = joinTable.compilePostFilterExpression(context, table);
             Integer limit = null;
+            Integer offset = null;
             if (!query.isAggregate() && !query.isDistinct() && query.getOrderBy().isEmpty()) {
                 limit = plan.getLimit();
+                offset = plan.getOffset();
             }
-            HashJoinInfo joinInfo = new HashJoinInfo(projectedTable, joinIds, joinExpressions, joinTypes, starJoinVector, tables, fieldPositions, postJoinFilterExpression, limit);
+            HashJoinInfo joinInfo = new HashJoinInfo(projectedTable, joinIds, joinExpressions, joinTypes,
+                    starJoinVector, tables, fieldPositions, postJoinFilterExpression, limit, offset);
```
maryannxue: The "limit" in HashJoinInfo is a little different than the original SQL semantics. It specifies the max number of rows returned by a HashJoinRegionScanner so that it doesn't do too much unnecessary work, and its max rows per region, so in the end the client might get more rows than needed and will do a final limit operation (all done by ScanPlan, don't worry). So, here, I think the limit should be "offset + limit" and there is no need to add an extra "offset" field to HashJoinInfo. Btw, change in serialization/deserialization of HashJoinInfo would cause incompatibility between different versions.

maryannxue:
@ankitsinghal Thank you very much for the pull request! I am impressed by how many details you have taken into account in your commit. It took me quite a while to go through all of the changes. That said, it would be great if you could add more test cases covering most (if not all) of the code changes, e.g. the join case (left outer join w/ offset and w/wo limit), the derived table case, the join with subquery case, etc.

Check JoinCompiler.isFlat(SelectStatement), it returns true only when limit == null, think it should be the same for offset. That function is called when a join contains a subquery, e.g. "select * from (select a, b from t1 offset 3) join (select c, d from t2 limit 10)".

phoenix-core/src/main/java/org/apache/phoenix/compile/SubselectRewriter.java
ankitsinghal: Thanks for pointing out. I thought a little and have updated the logic to 
 if (offsetRewrite != null || (limitRewrite != null & offset != null)) {
            return select;
        } else {
            offsetRewrite = offset;
        }
Let me know if any optimization is possible.

phoenix-core/src/main/java/org/apache/phoenix/compile/QueryCompiler.java
ankitsinghal: updated in the latest commit

ankitsinghal:
@maryannxue Thank you so much for the review! I have incorporated the changes as per your comments and added more test cases related to join, derived table ,group by case and join with subquery as you asked.

phoenix-core/src/it/java/org/apache/phoenix/end2end/DerivedTableIT.java
```diff
@@ -134,7 +134,7 @@ public void testDerivedTableWithWhere() throws Exception {
         Connection conn = DriverManager.getConnection(getUrl(), props);
         try {
             // (where)
-            String query = "SELECT t.eid, t.x + 9 FROM (SELECT entity_id eid, b_string b, a_byte + 1 x FROM aTable WHERE a_byte + 1 < 9) AS t";
+            String query = "SELECT t.eid, t.x + 9 FROM (SELECT entity_id eid, b_string b, a_byte + 1 x FROM aTable WHERE a_byte + 1 < 9 ) AS t";
             PreparedStatement statement = conn.prepareStatement(query);
```
maryannxue: Unnecessary diff

phoenix-core/src/it/java/org/apache/phoenix/end2end/QueryWithLimitIT.java
```diff
@@ -20,6 +20,7 @@
 package org.apache.phoenix.end2end;
 
 import static org.apache.phoenix.util.TestUtil.KEYONLY_NAME;
+import static org.apache.phoenix.util.TestUtil.TEST_PROPERTIES;
```
maryannxue: unnecessary diff

phoenix-core/src/it/java/org/apache/phoenix/end2end/DerivedTableIT.java
```diff
@@ -233,6 +233,34 @@ public void testDerivedTableWithWhere() throws Exception {
...
+
+            // Inner limit < outer query offset
+            query = "SELECT t.eid, t.x + 9 FROM (SELECT entity_id eid, b_string b, a_byte + 1 x FROM aTable LIMIT 1 OFFSET 1 ) AS t WHERE t.b = '"
+                    + C_VALUE + "' OFFSET 2";
+            statement = conn.prepareStatement(query);
+            rs = statement.executeQuery();
+            assertFalse(rs.next());
+
+            // (offset) where
+            query = "SELECT t.eid, t.x + 9 FROM (SELECT entity_id eid, b_string b, a_byte + 1 x FROM aTable WHERE a_byte + 1 < 9 OFFSET 2) AS t";
```
maryannxue: This is (offset where) rather than (offset) where. I think it would make more sense to test "(offset) where" here.

phoenix-core/src/main/java/org/apache/phoenix/execute/ScanPlan.java
```diff
@@ -193,21 +194,32 @@ protected ResultIterator newIterator(ParallelScanGrouper scanGrouper) throws SQL
          * limit is provided, run query serially.
          */
         boolean isOrdered = !orderBy.getOrderByExpressions().isEmpty();
-        boolean isSerial = isSerial(context, statement, tableRef, orderBy, limit, allowPageFilter);
+        boolean isSerial = isSerial(context, statement, tableRef, orderBy, limit, offset, allowPageFilter);
         Integer perScanLimit = !allowPageFilter || isOrdered ? null : limit;
+        if (perScanLimit != null) {
+            perScanLimit += (offset == null ? 0 : offset);
+        }
+		boolean hasOffset = offset != null;
```
maryannxue: Looks like there's <tab> here and a few lines below.

phoenix-core/src/main/java/org/apache/phoenix/execute/ScanPlan.java
```diff
@@ -193,21 +194,32 @@ protected ResultIterator newIterator(ParallelScanGrouper scanGrouper) throws SQL
          * limit is provided, run query serially.
          */
         boolean isOrdered = !orderBy.getOrderByExpressions().isEmpty();
-        boolean isSerial = isSerial(context, statement, tableRef, orderBy, limit, allowPageFilter);
+        boolean isSerial = isSerial(context, statement, tableRef, orderBy, limit, offset, allowPageFilter);
         Integer perScanLimit = !allowPageFilter || isOrdered ? null : limit;
+        if (perScanLimit != null) {
+            perScanLimit += (offset == null ? 0 : offset);
```
maryannxue: Can it use QueryUtil.getOffsetLimit() ?

phoenix-core/src/main/java/org/apache/phoenix/execute/LiteralResultIterationPlan.java
```diff
@@ -85,6 +85,9 @@ public void close() throws SQLException {
 
             @Override
             public Tuple next() throws SQLException {
+                while (!this.closed && (offset != null && count++ < offset) && tupleIterator.hasNext()) {
+                    tupleIterator.next();
+                }
```
maryannxue: This doesn't look right to me. Mind if you correct this and add a correponding test case?

maryannxue:
@ankitsinghal Thank you very much for making the suggested changes! Added a few more suggestions, most of them being minor.

ankitsinghal:
Thanks @maryannxue for suggestions , I have incorporated the review comments.

phoenix-core/src/main/java/org/apache/phoenix/execute/LiteralResultIterationPlan.java
```diff
@@ -85,7 +85,8 @@ public void close() throws SQLException {
 
             @Override
             public Tuple next() throws SQLException {
-                while (!this.closed && (offset != null && count++ < offset) && tupleIterator.hasNext()) {
+                while (!this.closed && (offset != null && count < offset) && tupleIterator.hasNext()) {
+                    count++;
```
maryannxue: What I meant was this "count++" doesn't seem to work with LIMIT(below) together. LIMIT should be number of rows returned starting from the OFFSET row, right?

phoenix-core/src/it/java/org/apache/phoenix/end2end/DerivedTableIT.java
```diff
@@ -241,7 +241,7 @@ public void testDerivedTableWithWhere() throws Exception {
             rs = statement.executeQuery();
             assertFalse(rs.next());
 
-            // (offset) where
+            // (where offset)
```
maryannxue: Guess it doesn't make too much sense here to keep this (where offset) case. You can simply remove it, or otherwise, make it "(where) offset" which is essentially the same as (where offset) but is more meaningful for testing query flattening.

phoenix-core/src/test/java/org/apache/phoenix/execute/CorrelatePlanTest.java
```diff
@@ -172,17 +172,38 @@ public void testCorrelatePlanWithSingleValueOnly() throws SQLException {
                 {2, "2", "2", 20},
                 {5, "5", "5", 100},
         };
-        testCorrelatePlan(LEFT_RELATION, rightRelation, 1, 0, JoinType.Inner, expected);        
+        testCorrelatePlan(LEFT_RELATION, rightRelation, 1, 0, JoinType.Inner, expected);
     }
     
-    private void testCorrelatePlan(Object[][] leftRelation, Object[][] rightRelation, int leftCorrelColumn, int rightCorrelColumn, JoinType type, Object[][] expectedResult) throws SQLException {        
+    @Test
```
maryannxue: Maybe add an independent test file for LiteralResultIterationPlan test, testing OFFSET, LIMIT, and OFFSET + LIMIT.

maryannxue:
@ankitsinghal Some (maybe final) comments added.

phoenix-core/src/main/java/org/apache/phoenix/execute/LiteralResultIterationPlan.java
ankitsinghal: Ah, I was in a notion that I updated limit to limit+offset already.
Thanks for catching it, I have corrected it now.

ankitsinghal:
I have taken care the last review comments in my latest commit.
Let me know if it is good to know. Thanks you so much @maryannxue for review.

ankitsinghal:
@maryannxue , what do you think .. is it good to go now?

maryannxue:
Hi @ankitsinghal, sorry about the delay. I don't get github notification in mailbox... It looks good to me now. Thank you for all the work! @JamesRTaylor What do you think?

phoenix-core/src/main/java/org/apache/phoenix/compile/QueryCompiler.java
```diff
@@ -168,7 +169,11 @@ public QueryPlan compileUnionAll(SelectStatement select) throws SQLException {
             SelectStatement subSelect = unionAllSelects.get(i);
             // Push down order-by and limit into sub-selects.
             if (!select.getOrderBy().isEmpty() || select.getLimit() != null) {
-                subSelect = NODE_FACTORY.select(subSelect, select.getOrderBy(), select.getLimit());
+                if (select.getOffset() == null) {
+                    subSelect = NODE_FACTORY.select(subSelect, select.getOrderBy(), select.getLimit(), null);
+                } else {
+                    subSelect = NODE_FACTORY.select(subSelect, select.getOrderBy(), null, null);
+                }
```
JamesRTaylor: Rather than this if statement, can we do the following to simplify it?

    subSelect = NODE_FACTORY.select(subSelect, select.getOrderBy(), select.getLimit(), select.getOffset())

phoenix-core/src/main/java/org/apache/phoenix/compile/QueryCompiler.java
```diff
@@ -168,7 +169,11 @@ public QueryPlan compileUnionAll(SelectStatement select) throws SQLException {
             SelectStatement subSelect = unionAllSelects.get(i);
             // Push down order-by and limit into sub-selects.
             if (!select.getOrderBy().isEmpty() || select.getLimit() != null) {
```
JamesRTaylor: Does this if statement need to be change to this (and if so, check for other occurrences)?

    if (!select.getOrderBy().isEmpty() || select.getLimit() != null || select.getOffset() != null)

phoenix-core/src/main/java/org/apache/phoenix/compile/SubqueryRewriter.java
```diff
@@ -167,7 +167,7 @@ public ParseNode visitLeave(ExistsParseNode node, List<ParseNode> l) throws SQLE
         JoinConditionExtractor conditionExtractor = new JoinConditionExtractor(subquery, resolver, connection, rhsTableAlias);
         ParseNode where = subquery.getWhere() == null ? null : subquery.getWhere().accept(conditionExtractor);
         if (where == subquery.getWhere()) { // non-correlated EXISTS subquery, add LIMIT 1
-            subquery = NODE_FACTORY.select(subquery, NODE_FACTORY.limit(NODE_FACTORY.literal(1)));
+            subquery = NODE_FACTORY.select(subquery, NODE_FACTORY.limit(NODE_FACTORY.literal(1)), null);
```
JamesRTaylor: Always null here? Just want to confirm that this is correct.

phoenix-core/src/main/java/org/apache/phoenix/compile/SubselectRewriter.java
```diff
@@ -188,6 +190,13 @@ private SelectStatement flatten(SelectStatement select, SelectStatement subselec
             }
         }
         
+        OffsetNode offset = select.getOffset();
+        if (offsetRewrite != null || (limitRewrite != null & offset != null)) {
```
JamesRTaylor: Single & operator here? Shouldn't it be && ?

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/ScanRegionObserver.java
```diff
@@ -218,7 +228,11 @@ protected RegionScanner doPostScannerOpen(final ObserverContext<RegionCoprocesso
         if (j != null) {
             innerScanner = new HashJoinRegionScanner(innerScanner, p, j, tenantId, c.getEnvironment());
         }
-
+        if (scanOffset != null) {
+            innerScanner = getOffsetScanner(c, innerScanner,
+                    new OffsetResultIterator(new RegionScannerResultIterator(innerScanner), scanOffset),
+                    scan.getAttribute(QueryConstants.LAST_SCAN) == null ? false : true);
```
JamesRTaylor: Nit: change this:

    scan.getAttribute(QueryConstants.LAST_SCAN) == null ? false : true

to this:

    scan.getAttribute(QueryConstants.LAST_SCAN) != null

phoenix-core/src/main/java/org/apache/phoenix/iterate/MergeSortTopNResultIterator.java
```diff
@@ -79,9 +86,11 @@ public Tuple peek() throws SQLException {
 
     @Override
     public Tuple next() throws SQLException {
-        if (limit >= 0 && count++ >= limit) {
-            return null;
+        while (count < offset) {
+            if (super.next() == null) { return null; }
+            count++;
         }
+        if (limit >= 0 && count++ >= limit) { return null; }
```
JamesRTaylor: Should this be just {{count}} or {{count++}} ?

phoenix-core/src/main/java/org/apache/phoenix/iterate/OrderedResultIterator.java
```diff
@@ -130,9 +135,9 @@ public OrderedResultIterator(ResultIterator delegate, List<OrderByExpression> or
             SizedUtil.OBJECT_SIZE + estimatedRowSize;
 
         // Make sure we don't overflow Long, though this is really unlikely to happen.
-        assert(limit == null || Long.MAX_VALUE / estimatedEntrySize >= limit);
+        assert (limit == null || Long.MAX_VALUE / estimatedEntrySize >= limit + this.offset);
```
JamesRTaylor: Nit: remove space before assert

phoenix-core/src/main/java/org/apache/phoenix/query/QueryConstants.java
```diff
@@ -138,6 +138,9 @@
     public final static String SYSTEM_SCHEMA_NAME = "SYSTEM";
     public final static byte[] SYSTEM_SCHEMA_NAME_BYTES = Bytes.toBytes(SYSTEM_SCHEMA_NAME);
     public final static String PHOENIX_METADATA = "table";
+    public final static String OFFSET = "offset";
+    public final static byte[] offsetRowKeyBytes = "_OFFSET_".getBytes();
```
JamesRTaylor: Nit: use Bytes.toBytes(OFFSET) instead. Also, did you want it to be "_OFFSET_" for the OFFSET constant?

phoenix-core/src/test/java/org/apache/phoenix/execute/CorrelatePlanTest.java
JamesRTaylor: Also, test for bind parameter for OFFSET, like this:

    SELECT * FROM T OFFSET ?

And check that the ResultSetMetaData and ParameterMetaData are ok.

phoenix-core/src/main/java/org/apache/phoenix/iterate/TableSerialIterators.java
```diff
@@ -0,0 +1,129 @@
...
+
+/**
+ *
+ * Class that parallelizes the scan over a table using the ExecutorService provided.  Each region of the table will be scanned in parallel with
+ * the results accessible through {@link #getIterators()}
+ *
+ * 
+ * @since 0.1
+ */
+public class TableSerialIterators extends BaseResultIterators {
```
JamesRTaylor: We have SerialIterators already. Any reason why we can't use that? I'd hate to introduce a new one if it's not absolutely necessary.

phoenix-core/src/main/java/org/apache/phoenix/compile/QueryCompiler.java
ankitsinghal: Actually , In case of union , we need to apply offset at the final result so we can't apply limit or offset in subselect.
but, if offset is not present , limit can be applied

ankitsinghal: Actually , In case of union , we need to apply offset at the final result so we can't apply limit or offset in subselect.
but, if offset is not present , limit can be applied

phoenix-core/src/main/java/org/apache/phoenix/compile/QueryCompiler.java
ankitsinghal: I think the change is not required as per above comment.

phoenix-core/src/main/java/org/apache/phoenix/iterate/MergeSortTopNResultIterator.java
ankitsinghal: Actually it should count++ only, but there is a bug in offset count .. which I have updated now.

phoenix-core/src/main/java/org/apache/phoenix/iterate/TableSerialIterators.java
ankitsinghal: Yes, I am able re-used it now.. earlier I was facing some issues..

phoenix-core/src/main/java/org/apache/phoenix/iterate/OrderedResultIterator.java
ankitsinghal: I checked and there was no space before assert.

phoenix-core/src/test/java/org/apache/phoenix/execute/CorrelatePlanTest.java
ankitsinghal: Added in a last commit.

ankitsinghal:
Thanks @JamesRTaylor for review.. I have made the changes are per your review comments.

phoenix-core/src/main/java/org/apache/phoenix/iterate/OffsetScanGrouper.java
```diff
@@ -0,0 +1,44 @@
...
+import java.util.List;
+
+import org.apache.hadoop.hbase.client.Scan;
+import org.apache.phoenix.compile.QueryPlan;
+
+/**
+ * Default implementation that creates a scan group if a plan is row key ordered (which requires a merge sort),
+ * or if a scan crosses a region boundary and the table is salted or a local index.   
+ */
+public class OffsetScanGrouper implements ParallelScanGrouper {
```
JamesRTaylor: Why is this class needed as it doesn't appear to do anything.

phoenix-core/src/main/java/org/apache/phoenix/iterate/OrderedResultIterator.java
JamesRTaylor: Based on the above diff, you added a space after "assert". Please remove so our formatting is consistent.

phoenix-core/src/main/java/org/apache/phoenix/query/QueryConstants.java
```diff
@@ -138,6 +138,10 @@
     public final static String SYSTEM_SCHEMA_NAME = "SYSTEM";
     public final static byte[] SYSTEM_SCHEMA_NAME_BYTES = Bytes.toBytes(SYSTEM_SCHEMA_NAME);
     public final static String PHOENIX_METADATA = "table";
+    public final static String SCAN_OFFSET = "OFFSET";
```
JamesRTaylor: Please change this to "_RowOffset" (to make is a little less general) and move to BaseScannerRegionObserver (this is where we keep all attributes we set on the Scan so we hopefully notice if we have a duplicate).

JamesRTaylor:
A few minor items, but otherwise it looks good. Nice work, @ankitsinghal! If you could file a JIRA for the corresponding website/doc changes and submit a patch for that when we're closer to the 4.8 release, that'd be much appreciated!

phoenix-core/src/main/java/org/apache/phoenix/compile/SubselectRewriter.java
JamesRTaylor: Based on this not being caught by one of your new unit tests, I'd say that there's likely a hole in the test coverage. Would you mind adding a unit test that would have caught this?

phoenix-core/src/main/java/org/apache/phoenix/iterate/OffsetScanGrouper.java
ankitsinghal: It is needed to avoid scans in parallel even in serialIterators. should I use a anonymous class while creating SerialIterators to do offset on server?

phoenix-core/src/main/java/org/apache/phoenix/compile/SubselectRewriter.java
ankitsinghal: @JamesRTaylor , will it really affect the boolean operands.

JamesRTaylor: You're right, @ankitsinghal - won't have an impact.

phoenix-core/src/main/java/org/apache/phoenix/iterate/OffsetScanGrouper.java
JamesRTaylor: Scans aren't done in parallel for SerialIterator, they're done serially. Have you tried it without this? I can't figure out what purpose it's serving, but perhaps I'm missing something?

ankitsinghal: It seems for serialIterators also , we use DefaultParallelScanGrouper for splitting scans in groups by condition `(isSalted || table.getIndexType() == IndexType.LOCAL) && ScanUtil.shouldRowsBeInRowKeyOrder(orderBy, context)` and run each group in one thread.

Anyways, I have modified the plan to run offset on server by including this condition and re-use the DefaultParallelScanGrouper

PR #155, opened 2016-03-28
PHOENIX-2743 Hive Storage support

ss77892:


phoenix-hive/pom.xml
```diff
@@ -0,0 +1,224 @@
...
+      <type>test-jar</type>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>org.apache.hadoop</groupId>
+      <artifactId>hadoop-common</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>org.apache.hadoop</groupId>
+      <artifactId>hadoop-annotations</artifactId>
```
joshelser: I don't see this actively referenced in your changes.

phoenix-hive/pom.xml
```diff
@@ -0,0 +1,224 @@
...
+    </dependency>
+    <dependency>
+      <groupId>org.apache.hadoop</groupId>
+      <artifactId>hadoop-mapreduce-client-core</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>org.apache.hadoop</groupId>
+      <artifactId>hadoop-minicluster</artifactId>
```
joshelser: Should this be test scope?

phoenix-hive/pom.xml
```diff
@@ -0,0 +1,224 @@
...
+        <groupId>org.codehaus.mojo</groupId>
+        <artifactId>build-helper-maven-plugin</artifactId>
+      </plugin>
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-failsafe-plugin</artifactId>
+      </plugin>
+      <plugin>
+        <artifactId>maven-dependency-plugin</artifactId>
+        <version>${maven-dependency-plugin.version}</version>
```
joshelser: Likely unnecessary version specification

phoenix-hive/pom.xml
```diff
@@ -0,0 +1,224 @@
...
+        <executions>
```
joshelser: Wonky indentation here. Tabs maybe?

phoenix-hive/pom.xml
```diff
@@ -0,0 +1,224 @@
...
+      </exclusions>
+    </dependency>
+    <dependency>
+      <groupId>org.apache.calcite</groupId>
+      <artifactId>calcite-core</artifactId>
+      <version>${calcite.version}</version>
+    </dependency>
+    <dependency>
+      <groupId>org.apache.calcite</groupId>
+      <artifactId>calcite-avatica</artifactId>
```
joshelser: Is this used?

phoenix-hive/pom.xml
```diff
@@ -0,0 +1,224 @@
...
+      <exclusions>
+        <exclusion>
+          <groupId>org.apache.calcite</groupId>
+          <artifactId>*</artifactId>
+        </exclusion>
+      </exclusions>
+    </dependency>
+    <dependency>
+      <groupId>org.apache.calcite</groupId>
+      <artifactId>calcite-core</artifactId>
```
joshelser: Same here, unused?

phoenix-hive/src/main/java/org/apache/phoenix/hive/PhoenixMetaHook.java
```diff
@@ -0,0 +1,245 @@
...
+                        "null,\n");
+                realRowKeys.append(rowKeyName).append(",");
+            } else {
+                // In case of Column
+                String columnName = columnMappingMap.get(fieldName);
+
+                if (columnName == null) {
+                    // Use field definition.
+                    columnName = fieldName;
+//					throw new MetaException("<<<<<<<<<< " + fieldName + " column mapping not exist
```
joshelser: Remove this.

phoenix-hive/src/main/java/org/apache/phoenix/hive/PhoenixRowKey.java
```diff
@@ -0,0 +1,69 @@
...
+
+public class PhoenixRowKey extends RecordIdentifier {
+
+	private Map<String, Object> rowKeyMap = Maps.newHashMap();
+
+	public PhoenixRowKey() {
+
+	}
+
+	//	public void add(String columnName, Object value) {
```
joshelser: Dead code

phoenix-hive/src/main/java/org/apache/phoenix/hive/constants/PhoenixStorageHandlerConstants.java
```diff
@@ -0,0 +1,101 @@
...
+	public static final String TEZ = "tez";
+	public static final String SPARK = "spark";
+
+	public static final String DATE_TYPE = "date";
+	public static final String TIMESTAMP_TYPE = "timestamp";
+	public static final String BETWEEN_COMPARATOR = "between";
+	public static final String IN_COMPARATOR = "in";
+	public static final List<String> COMMON_COMPARATOR = Lists.newArrayList("=", "<", ">", "<=", ">=");
+
+	// date/timestamp 타입 컬럼의 쿼리 조건 변환
```
joshelser: Remove the non-english characters

phoenix-hive/src/main/java/org/apache/phoenix/hive/objectinspector/PhoenixObjectInspectorFactory.java
```diff
@@ -0,0 +1,150 @@
...
+import org.apache.hadoop.hive.serde2.lazy.objectinspector.LazyObjectInspectorFactory;
+import org.apache.hadoop.hive.serde2.lazy.objectinspector.LazySimpleStructObjectInspector;
+import org.apache.hadoop.hive.serde2.objectinspector.ObjectInspector;
+import org.apache.hadoop.hive.serde2.objectinspector.ObjectInspectorFactory.ObjectInspectorOptions;
+import org.apache.hadoop.hive.serde2.typeinfo.ListTypeInfo;
+import org.apache.hadoop.hive.serde2.typeinfo.PrimitiveTypeInfo;
+import org.apache.hadoop.hive.serde2.typeinfo.StructTypeInfo;
+import org.apache.hadoop.hive.serde2.typeinfo.TypeInfo;
+
+public class PhoenixObjectInspectorFactory {
```
joshelser: I'm wondering why you need this factory (and the accompanying OI implementations for the different types) and you can't reuse the Hive ObjectInspector implementations. Something to do with strongly-typed values from Phoenix (as opposed to HBase/Accumulo's implementations)?

phoenix-hive/src/main/java/org/apache/phoenix/hive/mapreduce/PhoenixResultWritable.java
```diff
@@ -0,0 +1,215 @@
...
+    public PhoenixResultWritable(Configuration config, List<ColumnInfo> columnMetadataList) throws IOException {
+        this(config);
+        this.columnMetadataList = columnMetadataList;
+
+        valueList = Lists.newArrayListWithExpectedSize(columnMetadataList.size());
+    }
+
+    @Override
+    public void write(DataOutput out) throws IOException {
+    }
```
joshelser: No need to implement these?

phoenix-hive/src/main/java/org/apache/phoenix/hive/ql/index/IndexPredicateAnalyzer.java
```diff
@@ -0,0 +1,486 @@
...
+
+    public static IndexPredicateAnalyzer createAnalyzer(boolean equalOnly) {
+        IndexPredicateAnalyzer analyzer = new IndexPredicateAnalyzer();
+        analyzer.addComparisonOp("org.apache.hadoop.hive.ql.udf.generic.GenericUDFOPEqual");
+
+        if (equalOnly) {
+            return analyzer;
+        }
+
+        analyzer.addComparisonOp("org.apache.hadoop.hive.ql.udf.generic.GenericUDFOPEqualOrGreaterThan");
```
joshelser: Is there a reason to use hard-coded strings and not _Class_.getName()?

joshelser:
Some general thoughts (I stopped leaving them inline everytime I saw them). I'm guessing you "inherited" some of these from JeongMin's original work.

* Dbl-check indentations
* Try to remove commented out code
* Some class-level javadoc comments would be *amazing*
* Not a single unit test? :)

Other things that I remember biting me previously:

* Make sure you try to run with Tez as well. Both in the "uber" (local job) mode and a normal tez task. There are.. subtleties between them, sadly (as sadly, I don't remember the specifics anymore).

Other general thoughts:
* The RecordUpdater implementation looks pretty cool. Didn't know they made this available for StorageHandlers.
* Hive has a decent suite for running Hive tests as a part of their build (which includes tests for StorageHandlers) with this qtest/itest modules. You might be able to take some inspiration from these for testing.

Looks good so far. It will be a nice bridge between Phoenix and Hive (as we work towards a common-core of Calcite).

ss77892:
Actually I want to get some integration tests in the similar way as it was done in PHOENIX-331, but there is a clash between versions. This implementation uses Hive 1.2.1, so to run Hive minicluster it's require hadoop version >=  2.6, but hbase artifacts are still using 2.5, so it's hard to get all those miniclusters run together. I still hope to get it resolved. That's why the pom contains dependencies on calcite, hbase-test and others. 
I will add javadocs and clean commented code where it's possible.

phoenix-hive/pom.xml
ss77892: It will be there as soon as I get miniclusters working for ITs.

joshelser:
> it's hard to get all those miniclusters run together. I still hope to get it resolved. That's why the pom contains dependencies on calcite, hbase-test and others. 

If it's just at the test scope, you might be able to circumvent the issue by just overriding the hbase and hadoop versions to compatible versions only at the test scope.

ss77892:
Yep, I know. The problem is that hbase-it is built against 2.5.1 and if I set hadoop version 2.7.1 it will fail to start DFS minicluster for hbase.

JamesRTaylor:
I think it's ok to change Phoenix to specify Hadoop 2.7.1. There's no particular reason it's set to 2.5.1 (other than no one has changed it in a while). See PHOENIX-2761.

phoenix-hive/src/main/java/org/apache/phoenix/hive/ql/index/IndexPredicateAnalyzer.java
ss77892: Actually not, but it was done in the same way as it's done in Hive's IndexPredicateAnalyzer.java

phoenix-hive/src/main/java/org/apache/phoenix/hive/mapreduce/PhoenixResultWritable.java
ss77892: We are using it in our PhoenixRecordWriter and we call write only with statement.

phoenix-hive/src/main/java/org/apache/phoenix/hive/objectinspector/PhoenixObjectInspectorFactory.java
ss77892: Honestly speaking I'm not sure :) That's how it comes from the guys who contribute this code.

joshelser: Haha, ok. I want to say that I think they are superfluous -- Hive already provides implementations for the OI's. {HBase,Accumulo}StorageHandler should already have examples on how they work. It would reduce the amount of code to maintain here which is great.

phoenix-hive/src/main/java/org/apache/phoenix/hive/mapreduce/PhoenixResultWritable.java
joshelser: Maybe add `throws new UnsupportedOperationException()` then?

phoenix-hive/pom.xml
```diff
@@ -117,66 +93,44 @@
...
+      <artifactId>hadoop-hdfs</artifactId>
+      <version>2.7.1</version>
       <scope>test</scope>
     </dependency>
     <dependency>
-      <groupId>org.apache.hbase</groupId>
-      <artifactId>hbase-hadoop2-compat</artifactId>
+      <groupId>org.apache.hadoop</groupId>
+      <artifactId>hadoop-hdfs</artifactId>
+      <version>2.7.1</version>
```
joshelser: Given James' previous comment, I think you can safely change the hadoop version at the top-level pom. Then, you won't have to specify 2.7.1 down here in phoenix-hive.

phoenix-hive/src/main/java/org/apache/phoenix/hive/PhoenixMetaHook.java
```diff
@@ -6,9 +6,9 @@
  * to you under the Apache License, Version 2.0 (the
  * "License"); you may not use this file except in compliance
  * with the License.  You may obtain a copy of the License at
- *
- *     http://www.apache.org/licenses/LICENSE-2.0
- *
+ * <p>
```
joshelser: Adding in the paragraph tags might break the RAT check. Best to revert that.

joshelser:
Recent improvements look good! The new integration test is *very* nice. The ObjectInspector stuff isn't a huge deal to replace, but it's something that can be consolidated later. Left a few minor comments, but I this is good to go after those are fixed.

joshelser:
@ss77892 one more nit-pick, it seems like you removed the version for some of the entries in the pom but then added `${hadoop-two.version}` for others. I would assume that it's not necessary for any of them, but let me know if I'm wrong. I can also just fix this while merging it in as long as @JamesRTaylor thinks this good to go too :)

PR #156, opened 2016-03-29
PHOENIX-2628 Ensure split when iterating through results handled corr…

chrajeshbabu:
The patch fixes issues with splits and merges while scanning local indexes.

### DIFF
--- a/phoenix-core/src/it/java/org/apache/phoenix/end2end/index/LocalIndexIT.java
+++ b/phoenix-core/src/it/java/org/apache/phoenix/end2end/index/LocalIndexIT.java
@@ -727,7 +727,7 @@ public class LocalIndexIT extends BaseHBaseManagedTimeIT {
             assertTrue(rs.next());
             
             HBaseAdmin admin = driver.getConnectionQueryServices(getUrl(), TestUtil.TEST_PROPERTIES).getAdmin();
-            for (int i = 1; i < 5; i++) {
+            for (int i = 1; i < 2; i++) {
                 admin.split(Bytes.toBytes(TestUtil.DEFAULT_DATA_TABLE_NAME), ByteUtil.concat(Bytes.toBytes(strings[3*i])));
                 List<HRegionInfo> regionsOfUserTable =
                         MetaTableAccessor.getTableRegions(getUtility().getZooKeeperWatcher(), admin.getConnection(),
@@ -971,24 +971,6 @@ public class LocalIndexIT extends BaseHBaseManagedTimeIT {
             assertEquals(5, regionsOfIndexTable.size());
             boolean success = latch1.await(WAIT_TIME_SECONDS, TimeUnit.SECONDS);
             assertTrue("Timed out waiting for MockedLocalIndexSplitter.preSplitAfterPONR to complete", success);
-            // Verify the metadata for index is correct.
-            rs = conn1.getMetaData().getTables(null, StringUtil.escapeLike(TestUtil.DEFAULT_SCHEMA_NAME), TestUtil.DEFAULT_INDEX_TABLE_NAME,
-                    new String[] { PTableType.INDEX.toString() });
-            assertTrue(rs.next());
-            assertEquals(TestUtil.DEFAULT_INDEX_TABLE_NAME, rs.getString(3));
-            assertEquals(PIndexState.INACTIVE.toString(), rs.getString("INDEX_STATE"));
-            assertFalse(rs.next());
-            rs = conn1.getMetaData().getTables(null, StringUtil.escapeLike(TestUtil.DEFAULT_SCHEMA_NAME), TestUtil.DEFAULT_INDEX_TABLE_NAME+"_2",
-                new String[] { PTableType.INDEX.toString() });
-            assertTrue(rs.next());
-            assertEquals(TestUtil.DEFAULT_INDEX_TABLE_NAME+"_2", rs.getString(3));
-            assertEquals(PIndexState.INACTIVE.toString(), rs.getString("INDEX_STATE"));
-            assertFalse(rs.next());
-
-            String query = "SELECT t_id,k1,v1 FROM " + TestUtil.DEFAULT_DATA_TABLE_NAME+"2";
-            rs = conn1.createStatement().executeQuery("EXPLAIN " + query);
-            assertEquals("CLIENT PARALLEL " + 1 + "-WAY FULL SCAN OVER " + TestUtil.DEFAULT_DATA_TABLE_NAME+"2",
-                QueryUtil.getExplainPlan(rs));
             latch2.countDown();
        } finally {
             conn1.close();
--- a/phoenix-core/src/it/java/org/apache/phoenix/end2end/index/LocalIndexIT.java
+++ b/phoenix-core/src/it/java/org/apache/phoenix/end2end/index/LocalIndexIT.java
@@ -727,7 +727,7 @@ public class LocalIndexIT extends BaseHBaseManagedTimeIT {
             assertTrue(rs.next());
             
             HBaseAdmin admin = driver.getConnectionQueryServices(getUrl(), TestUtil.TEST_PROPERTIES).getAdmin();
-            for (int i = 1; i < 2; i++) {
+            for (int i = 1; i < 5; i++) {
                 admin.split(Bytes.toBytes(TestUtil.DEFAULT_DATA_TABLE_NAME), ByteUtil.concat(Bytes.toBytes(strings[3*i])));
                 List<HRegionInfo> regionsOfUserTable =
                         MetaTableAccessor.getTableRegions(getUtility().getZooKeeperWatcher(), admin.getConnection(),
--- a/phoenix-core/src/it/java/org/apache/phoenix/end2end/index/MutableIndexIT.java
+++ b/phoenix-core/src/it/java/org/apache/phoenix/end2end/index/MutableIndexIT.java
@@ -629,15 +629,7 @@ public class MutableIndexIT extends BaseHBaseManagedTimeIT {
             ResultSet rs = conn1.createStatement().executeQuery("SELECT * FROM " + tableName);
             assertTrue(rs.next());
             splitDuringScan(conn1, strings, admin, isReverse);
-            props = new Properties();
-            props.setProperty(QueryServices.SCAN_RESULT_CHUNK_SIZE, "10");
-            props.setProperty(QueryServices.SCAN_CACHE_SIZE_ATTRIB, Integer.toString(4));
-            try(Connection conn = DriverManager.getConnection(getUrl(), props)){
-                dropTable(admin, conn);
-                createTableAndLoadData(conn, strings, isReverse);
-                splitDuringScan(conn, strings, admin, isReverse);
-                dropTable(admin, conn);
-            }
+            dropTable(admin, conn1);
        } 
     }
 
@@ -770,18 +762,14 @@ public class MutableIndexIT extends BaseHBaseManagedTimeIT {
         ResultSet rs;
         String query = "SELECT t_id,k1,v1 FROM " + tableName;
         rs = conn1.createStatement().executeQuery(query);
-    
+        String[] tIdColumnValues = new String[26]; 
+        String[] v1ColumnValues = new String[26];
+        int[] k1ColumnValue = new int[26];
         for (int j = 0; j < 5; j++) {
             assertTrue(rs.next());
-            if(isReverse) {
-                assertEquals(strings[j], rs.getString("t_id"));
-                assertEquals(j, rs.getInt("k1"));
-                assertEquals(strings[25-j], rs.getString("V1"));
-            } else {
-                assertEquals(strings[25 - j], rs.getString("t_id"));
-                assertEquals(25 - j, rs.getInt("k1"));
-                assertEquals(strings[j], rs.getString("V1"));
-            }
+            tIdColumnValues[j] = rs.getString("t_id");
+            k1ColumnValue[j] = rs.getInt("k1");
+            v1ColumnValues[j] = rs.getString("V1");
         }
 
         String[] splitKeys = new String[2];
@@ -817,15 +805,17 @@ public class MutableIndexIT extends BaseHBaseManagedTimeIT {
         }
         for (int j = 5; j < 26; j++) {
             assertTrue(rs.next());
-            if(isReverse) {
-                assertEquals(strings[j], rs.getString("t_id"));
-                assertEquals(j, rs.getInt("k1"));
-                assertEquals(strings[25-j], rs.getString("V1"));
-            } else {
-                assertEquals(strings[25 - j], rs.getString("t_id"));
-                assertEquals(25 - j, rs.getInt("k1"));
-                assertEquals(strings[j], rs.getString("V1"));
-            }
+            tIdColumnValues[j] = rs.getString("t_id");
+            k1ColumnValue[j] = rs.getInt("k1");
+            v1ColumnValues[j] = rs.getString("V1");
+        }
+        Arrays.sort(tIdColumnValues);
+        Arrays.sort(v1ColumnValues);
+        Arrays.sort(k1ColumnValue);
+        assertTrue(Arrays.equals(strings, tIdColumnValues));
+        assertTrue(Arrays.equals(strings, v1ColumnValues));
+        for(int i=0;i<26;i++) {
+            assertEquals(i, k1ColumnValue[i]);
         }
         assertFalse(rs.next());
         return regionsOfUserTable;
--- a/phoenix-core/src/it/java/org/apache/phoenix/end2end/index/MutableIndexIT.java
+++ b/phoenix-core/src/it/java/org/apache/phoenix/end2end/index/MutableIndexIT.java
@@ -23,6 +23,7 @@ import static org.junit.Assert.assertFalse;
 import static org.junit.Assert.assertNull;
 import static org.junit.Assert.assertTrue;
 
+import java.io.IOException;
 import java.sql.Connection;
 import java.sql.DriverManager;
 import java.sql.PreparedStatement;
@@ -31,14 +32,25 @@ import java.sql.SQLException;
 import java.sql.Statement;
 import java.util.Arrays;
 import java.util.Collection;
+import java.util.List;
 import java.util.Map;
 import java.util.Properties;
 
+import jline.internal.Log;
+
+import org.apache.hadoop.hbase.HRegionInfo;
+import org.apache.hadoop.hbase.MetaTableAccessor;
+import org.apache.hadoop.hbase.TableName;
+import org.apache.hadoop.hbase.client.HBaseAdmin;
+import org.apache.hadoop.hbase.util.Bytes;
+import org.apache.hadoop.hbase.util.Threads;
 import org.apache.phoenix.end2end.BaseHBaseManagedTimeIT;
 import org.apache.phoenix.end2end.Shadower;
 import org.apache.phoenix.jdbc.PhoenixConnection;
 import org.apache.phoenix.query.QueryServices;
 import org.apache.phoenix.schema.PTableKey;
+import org.apache.phoenix.util.ByteUtil;
+import org.apache.phoenix.util.MetaDataUtil;
 import org.apache.phoenix.util.PropertiesUtil;
 import org.apache.phoenix.util.QueryUtil;
 import org.apache.phoenix.util.ReadOnlyProps;
@@ -86,8 +98,8 @@ public class MutableIndexIT extends BaseHBaseManagedTimeIT {
 	
 	@Parameters(name="localIndex = {0} , transactional = {1}")
     public static Collection<Boolean[]> data() {
-        return Arrays.asList(new Boolean[][] {     
-                 { false, false }, { false, true }, { true, false }, { true, true }
+        return Arrays.asList(new Boolean[][] {
+                { false, false }, { false, true }, { true, false }, { true, true }
            });
     }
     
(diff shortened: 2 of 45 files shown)


phoenix-core/src/it/java/org/apache/phoenix/end2end/index/LocalIndexIT.java
```diff
@@ -727,7 +727,7 @@ public void testLocalIndexScanAfterRegionSplit() throws Exception {
             assertTrue(rs.next());
             
             HBaseAdmin admin = driver.getConnectionQueryServices(getUrl(), TestUtil.TEST_PROPERTIES).getAdmin();
-            for (int i = 1; i < 5; i++) {
+            for (int i = 1; i < 2; i++) {
```
JamesRTaylor: Any reason to go from 5 to 2 here? Maybe get rid of the loop then? Does it decrease the amount of testing being done?

phoenix-core/src/it/java/org/apache/phoenix/end2end/index/MutableIndexIT.java
```diff
@@ -86,8 +98,8 @@ public static void doSetup() throws Exception {
 	
 	@Parameters(name="localIndex = {0} , transactional = {1}")
     public static Collection<Boolean[]> data() {
-        return Arrays.asList(new Boolean[][] {     
-                 { false, false }, { false, true }, { true, false }, { true, true }
```
JamesRTaylor: Whitespace only change?

phoenix-core/src/main/java/org/apache/hadoop/hbase/regionserver/IndexHalfStoreFileReaderGenerator.java
```diff
@@ -157,6 +162,7 @@ public Reader preStoreFileReaderOpen(ObserverContext<RegionCoprocessorEnvironmen
             } catch (ClassNotFoundException e) {
                 throw new IOException(e);
             } catch (SQLException e) {
+                e.printStackTrace();
```
JamesRTaylor: Remove as we're throwing here

phoenix-core/src/main/java/org/apache/phoenix/compile/ScanRanges.java
```diff
@@ -385,9 +387,25 @@ public Scan intersectScan(Scan scan, final byte[] originalStartKey, final byte[]
         if (scanStopKey.length > 0 && Bytes.compareTo(scanStartKey, scanStopKey) >= 0) { 
             return null; 
         }
-        newScan.setAttribute(SCAN_ACTUAL_START_ROW, scanStartKey);
-        newScan.setStartRow(scanStartKey);
-        newScan.setStopRow(scanStopKey);
+        if(ScanUtil.isLocalIndex(scan)) {
```
JamesRTaylor: The code in this if statement can go in the if (keyOffset > 0) if statement instead. We're trying to avoid checks for local indexes.

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/BaseScannerRegionObserver.java
```diff
@@ -165,7 +183,7 @@ public RegionScanner preScannerOpen(final ObserverContext<RegionCoprocessorEnvir
             scan.setTimeRange(timeRange.getMin(), Bytes.toLong(txnScn));
         }
         if (isRegionObserverFor(scan)) {
-            if (! skipRegionBoundaryCheck(scan)) {
+            if (! skipRegionBoundaryCheck(scan) || ScanUtil.isLocalIndex(scan)) {
```
JamesRTaylor: Add comment, please. Something like this:

    // For local indexes, we need to throw if out of region as we'll get duplicates otherwise while in other cases,
    // it may just mean out client-side data on region boundaries is out of date and can safely be ignored.

Note that we'll need to handle this from the MR integration which might be painful.

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/BaseScannerRegionObserver.java
```diff
@@ -337,6 +384,22 @@ public boolean nextRaw(List<Cell> result) throws IOException {
...
+                                    return next;
+                                }
+                                if (arrayFuncRefs != null && arrayFuncRefs.length > 0 && arrayKVRefs.size() > 0) {
+                                    replaceArrayIndexElement(arrayKVRefs, arrayFuncRefs, result);
+                                }
+                                firstCell = result.get(0);
+                            }
+                        }
+                        System.out.println("&&&&-< "+scan);
+                        System.out.println("&&&&-<< "+result);
```
JamesRTaylor: Remove System.out.println calls

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/BaseScannerRegionObserver.java
```diff
@@ -279,6 +301,31 @@ protected RegionScanner getWrappedScanner(final ObserverContext<RegionCoprocesso
             final ImmutableBytesWritable ptr) {
         return new RegionScanner() {
 
+            private boolean hasReferences = checkForReferenceFiles();
+            private HRegionInfo regionInfo = c.getEnvironment().getRegionInfo();
+            private byte[] actualStartKey = getActualStartKey();
+
+            private boolean checkForReferenceFiles(){
```
JamesRTaylor: Please add a code comment about why we need to check for references. Do we always have to do this? Is it expensive?

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/BaseScannerRegionObserver.java
```diff
@@ -279,6 +301,31 @@ protected RegionScanner getWrappedScanner(final ObserverContext<RegionCoprocesso
...
+                        return scan.getStartRow();
+                    } else {
+                        return startKey;
+                    }
+                }
+                return null;
+            }
+
             @Override
             public boolean next(List<Cell> results) throws IOException {
```
JamesRTaylor: Can we refactor the code between these two next(List<Cell>, ...) implementations, as there's now a fair amount of code which is duplicated here?

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/BaseScannerRegionObserver.java
```diff
@@ -337,6 +384,22 @@ public boolean nextRaw(List<Cell> result) throws IOException {
                         arrayElementCell = result.get(arrayElementCellPosition);
                     }
                     if (ScanUtil.isLocalIndex(scan) && !ScanUtil.isAnalyzeTable(scan)) {
```
JamesRTaylor: We'll need some code comments here. Are we skipping until we find the actualStartKey row? Would it be better/possible to do a seek? This seems like it'd be potentially expensive, no?

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/GroupedAggregateRegionObserver.java
```diff
@@ -423,7 +426,14 @@ private RegionScanner scanUnordered(ObserverContext<RegionCoprocessorEnvironment
                         }
                     } while (hasMore && groupByCache.size() < limit);
                 }
-            } finally {
+            }  catch(NotServingRegionException e){
+                if(ScanUtil.isLocalIndex(scan)) {
+                    Exception cause = new StaleRegionBoundaryCacheException(c.getEnvironment().getRegion().getRegionInfo().getTable().getNameAsString());
+                    throw new DoNotRetryIOException(cause.getMessage(), cause);
+                } else {
+                    throw e;
```
JamesRTaylor: Is the throw ok because the client scanner knows how to deal with this in cases other than local indexes? Let's doc that, please.

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/GroupedAggregateRegionObserver.java
```diff
@@ -402,8 +405,8 @@ private RegionScanner scanUnordered(ObserverContext<RegionCoprocessorEnvironment
             }
 
             Region region = c.getEnvironment().getRegion();
-            region.startRegionOperation();
             try {
+                region.startRegionOperation();
```
JamesRTaylor: Doesn't startRegionOperation() take a lock which would prevent a split from occurring and thus a NotServingRegionException from happening? Or is this the case where the split occurred after pre open, but before we start iterating? Let's doc that please.

phoenix-core/src/main/java/org/apache/phoenix/execute/BaseQueryPlan.java
```diff
@@ -196,26 +196,31 @@ public Expression getDynamicFilter() {
...
         
         if (tableRef == TableRef.EMPTY_TABLE_REF) {
-            return newIterator(scanGrouper);
+            return newIterator(scanGrouper, s);
         }
         
         // Set miscellaneous scan attributes. This is the last chance to set them before we
         // clone the scan for each parallelized chunk.
-        Scan scan = context.getScan();
+        Scan scan = s == null ? context.getScan() : s;
```
JamesRTaylor: Is this null check required?

phoenix-core/src/main/java/org/apache/phoenix/execute/ScanPlan.java
```diff
@@ -209,7 +209,11 @@ protected ResultIterator newIterator(ParallelScanGrouper scanGrouper) throws SQL
         if (isOrdered) {
             scanner = new MergeSortTopNResultIterator(iterators, limit, orderBy.getOrderByExpressions());
         } else {
-            if ((isSalted || table.getIndexType() == IndexType.LOCAL) && ScanUtil.shouldRowsBeInRowKeyOrder(orderBy, context)) {
+            if ((isSalted || table.getIndexType() == IndexType.LOCAL)
+                    && ScanUtil.shouldRowsBeInRowKeyOrder(orderBy, context)
+                    || (table.getIndexType() == IndexType.LOCAL && (Bytes.compareTo(
+                        scan.getStartRow(), this.context.getScan().getStartRow()) != 0 || Bytes
+                            .compareTo(scan.getStopRow(), this.context.getScan().getStopRow()) != 0))) {
```
JamesRTaylor: Why is  a merge sort required for this new condition? Is this the case when the original start/stop row has changed? Is this purely for the chunked result iterator case? Let's document this and add a TODO to remove as the chunked result iterator is deprecated (and if you could consistently add this TODO for other parts of code only necessary due to this deprecated chunking mechanism, that'd be much appreciated).

phoenix-core/src/main/java/org/apache/phoenix/iterate/BaseResultIterators.java
```diff
@@ -326,11 +325,12 @@ private static void optimizeProjection(StatementContext context, Scan scan, PTab
         }
     }
     
-    public BaseResultIterators(QueryPlan plan, Integer perScanLimit, ParallelScanGrouper scanGrouper) throws SQLException {
+    public BaseResultIterators(QueryPlan plan, Integer perScanLimit, ParallelScanGrouper scanGrouper, Scan scan) throws SQLException {
         super(plan.getContext(), plan.getTableRef(), plan.getGroupBy(), plan.getOrderBy(), plan.getStatement().getHint(), plan.getLimit());
         this.plan = plan;
         this.scanGrouper = scanGrouper;
         StatementContext context = plan.getContext();
+        this.scan = scan == null ? context.getScan() : scan;
```
JamesRTaylor: When is scan null?

phoenix-core/src/main/java/org/apache/phoenix/iterate/BaseResultIterators.java
```diff
@@ -465,7 +465,14 @@ private static String toString(List<byte[]> gps) {
     }
 
     private List<List<Scan>> getParallelScans() throws SQLException {
-        return getParallelScans(EMPTY_BYTE_ARRAY, EMPTY_BYTE_ARRAY);
+        if (scan == null
+                || (ScanUtil.isLocalIndex(scan)
+                        && Bytes.compareTo(context.getScan().getStartRow(), scan.getStartRow()) == 0 && Bytes
+                        .compareTo(context.getScan().getStopRow(), scan.getStopRow()) == 0)) {
```
JamesRTaylor: Cover with ScanUtil method, as I've seen this code repeated a number of times.

phoenix-core/src/main/java/org/apache/phoenix/iterate/BaseResultIterators.java
```diff
@@ -465,7 +465,14 @@ private static String toString(List<byte[]> gps) {
...
+            return getParallelScans(EMPTY_BYTE_ARRAY, EMPTY_BYTE_ARRAY);
```
JamesRTaylor: This seems strange - why do we call getParallelScans with empty byte arrays in this case? Please document.

phoenix-core/src/main/java/org/apache/phoenix/iterate/BaseResultIterators.java
```diff
@@ -556,35 +564,55 @@ private static String toString(List<byte[]> gps) {
...
-                    if (newScan != null) {
-                        newScan.setAttribute(EXPECTED_UPPER_REGION_KEY, endRegionKey);
-                    } else if (!scans.isEmpty()) {
-                        scans.get(scans.size()-1).setAttribute(EXPECTED_UPPER_REGION_KEY, endRegionKey);
-                    }
-                }
-                scans = addNewScan(parallelScans, scans, newScan, endKey, true, regionLocation);
-                currentKeyBytes = endKey;
+                if (Bytes.compareTo(scan.getStartRow(), context.getScan().getStartRow()) != 0
+                     || Bytes.compareTo(scan.getStopRow(), context.getScan().getStopRow()) != 0) {
```
JamesRTaylor: Cover with ScanUtil function

JamesRTaylor: Also, this code path is strange. What's the entry point? Does it need to be mixed up with the existing code? It feels like we've completely forked the flow here. Why not have it in it's own method called only when this special case is detected?

phoenix-core/src/main/java/org/apache/phoenix/iterate/BaseResultIterators.java
```diff
@@ -556,35 +564,55 @@ private static String toString(List<byte[]> gps) {
...
+                    Scan newScan = ScanUtil.newScan(scan);
+                    if(ScanUtil.isLocalIndex(scan)) {
+                        newScan.setStartRow(regionInfo.getStartKey());
+                        newScan.setAttribute(SCAN_ACTUAL_START_ROW, regionInfo.getStartKey());
```
JamesRTaylor: Is there some symmetry between SCAN_ACTUAL_START_ROW and EXPECTED_UPPER_REGION_KEY? Should they be named similarly?

phoenix-core/src/main/java/org/apache/phoenix/iterate/ChunkedResultIterator.java
```diff
@@ -56,6 +57,7 @@
     private final MutationState mutationState;
     private Scan scan;
     private PeekingResultIterator resultIterator;
+    private QueryPlan plan;
 
     public static class ChunkedResultIteratorFactory implements ParallelIteratorFactory {
```
JamesRTaylor: FYI, this class is deprecated now with newer versions of HBase.

phoenix-core/src/main/java/org/apache/phoenix/iterate/TableResultIterator.java
```diff
@@ -80,13 +94,19 @@
...
+    public TableResultIterator(MutationState mutationState, Scan scan, CombinableMetric scanMetrics, long renewLeaseThreshold, QueryPlan plan, boolean handleSplitRegionBoundaryFailureDuringInitialization) throws SQLException {
         this.scan = scan;
         this.scanMetrics = scanMetrics;
-        PTable table = tableRef.getTable();
+        PTable table = plan.getTableRef().getTable();
         htable = mutationState.getHTable(table);
         this.scanIterator = UNINITIALIZED_SCANNER;
         this.renewLeaseThreshold = renewLeaseThreshold;
+        this.plan = plan;
+        this.handleSplitRegionBoundaryFailureDuringInitialization = handleSplitRegionBoundaryFailureDuringInitialization;
```
JamesRTaylor: When is handleSplitRegionBoundaryFailureDuringInitialization true versus false? Why would we *not* want to handle it?

phoenix-core/src/main/java/org/apache/phoenix/iterate/TableResultIterator.java
```diff
@@ -107,8 +127,37 @@ public synchronized void close() throws SQLException {
...
+            } catch(StaleRegionBoundaryCacheException e1) {
+                if(scan.getAttribute(NON_AGGREGATE_QUERY)!=null) {
+                    Scan newScan = ScanUtil.newScan(scan);
+                    if(lastTuple != null) {
+                        lastTuple.getKey(ptr);
+                        byte[] startRowSuffix = ByteUtil.copyKeyBytesIfNecessary(ptr);
+                        if(ScanUtil.isLocalIndex(newScan)) {
+                            newScan.setAttribute(SCAN_START_ROW_SUFFIX, ByteUtil.nextKey(startRowSuffix));
+                        } else {
+                            newScan.setStartRow(ByteUtil.nextKey(startRowSuffix));
```
JamesRTaylor: I'm not convinced that this nextKey will universally work. Probably better (if we really want to just get the next possible key) to add a 0 byte instead.

phoenix-core/src/main/java/org/apache/phoenix/iterate/TableResultIterator.java
```diff
@@ -107,8 +127,37 @@ public synchronized void close() throws SQLException {
...
+                        }
+                    }
+                    plan.getContext().getConnection().getQueryServices().clearTableRegionCache(htable.getTableName());
+                    this.scanIterator =
+                            plan.iterator(DefaultParallelScanGrouper.getInstance(), newScan);
```
JamesRTaylor: Getting the ParallelScanGrouper here as state instead of assuming the default one is being used may be necessary (or making it not necessary to have this ParallelScanGrouper - it's an edge case for the MR integration - there's likely a better way - perhaps the plan itself could know what to use for the ParallelScanGrouper?)

phoenix-core/src/main/java/org/apache/phoenix/iterate/TableResultIterator.java
```diff
@@ -107,8 +127,37 @@ public synchronized void close() throws SQLException {
...
+            lastTuple = scanIterator.next();
+            if (lastTuple != null) {
+                ImmutableBytesWritable ptr = new ImmutableBytesWritable();
+                lastTuple.getKey(ptr);
+            }
+        } catch (SQLException e) {
+            try {
+                throw ServerUtil.parseServerException(e);
+            } catch(StaleRegionBoundaryCacheException e1) {
+                if(scan.getAttribute(NON_AGGREGATE_QUERY)!=null) {
```
JamesRTaylor: Minor nit: how about adding a ScanUtil.isAggregateQuery() method instead?

phoenix-core/src/main/java/org/apache/phoenix/iterate/TableResultIterator.java
```diff
@@ -107,8 +127,37 @@ public synchronized void close() throws SQLException {
...
+        try {
+            lastTuple = scanIterator.next();
+            if (lastTuple != null) {
+                ImmutableBytesWritable ptr = new ImmutableBytesWritable();
+                lastTuple.getKey(ptr);
+            }
+        } catch (SQLException e) {
+            try {
+                throw ServerUtil.parseServerException(e);
+            } catch(StaleRegionBoundaryCacheException e1) {
```
JamesRTaylor: This needs code comments, please.

phoenix-core/src/main/java/org/apache/phoenix/iterate/TableResultIterator.java
```diff
@@ -121,8 +170,21 @@ public synchronized void initScanner() throws SQLException {
                 this.scanIterator =
                         new ScanningResultIterator(htable.getScanner(scan), scanMetrics);
             } catch (IOException e) {
-                Closeables.closeQuietly(htable);
-                throw ServerUtil.parseServerException(e);
+                if(handleSplitRegionBoundaryFailureDuringInitialization) {
```
JamesRTaylor: Rather than have a handleSplitRegionBoundaryFailureDuringInitialization, would it work to always throw it and let the caller deal with it? When a variable name gets beyond 50 characters, it's a sign that it may be too special of a case. :-)

JamesRTaylor:
Thanks for the patch, @chrajeshbabu. This is a big improvement over your initial approach. It's still complex, though. There's an on going cost to complicating the code to this degree that would be good to minimize. 

One big thing missing is an overview of all the moving parts. Perhaps that could live in BaseResultIterators? For example - this exception is thrown under these conditions and this state is added to the scan to know how to adjust the scan ranges, etc. etc. Especially that hasReferences bit in the coprocessor - what's that all about?

I'd like to understand how ChunkedResultIterator complicates things too. I think it'd be ok to document that splits during aggregate queries aren't handled if your HBase version is less that XXX if it'd significantly simplify this patch (as that code is essentially deprecated). 

Another potential, different approach would be for Phoenix to universally handle the split during scan case (rather than letting the HBase client scanner handle it for non aggregate case and Phoenix handle it for the aggregate case). Would that simplify things?

phoenix-core/src/it/java/org/apache/phoenix/end2end/index/LocalIndexIT.java
chrajeshbabu: While debugging a failure changed it to 2. 5 is ok will keep it as it is.

phoenix-core/src/it/java/org/apache/phoenix/end2end/index/MutableIndexIT.java
chrajeshbabu: Yes James it's space only change.

phoenix-core/src/main/java/org/apache/hadoop/hbase/regionserver/IndexHalfStoreFileReaderGenerator.java
chrajeshbabu: Yes. This can removed.

phoenix-core/src/main/java/org/apache/phoenix/compile/ScanRanges.java
chrajeshbabu: Sure will try to remove the local index scan check.

chrajeshbabu: @JamesRTaylor  we cannot use keyoffset > 0 check always for considering local indexes scan because in special case like if a table has only one region then region start key length and end key length is zero so keyoffset also becoming zero. In that case we need to check local index scan or not to set the attributes properly. Wdyt?

JamesRTaylor: How about moving the setting of these special attributes outside of this method?

phoenix-core/src/main/java/org/apache/phoenix/iterate/TableResultIterator.java
JamesRTaylor: So we're re-running the scan, only if it's an aggregate scan? Does this apply only for an "ordered" group by (which is grouping by the leading part of the pk and doing so in-place)? In the unordered case, each scan is done during the post open and the results are cached on the RS, so a split won't impact the results. There may be a race condition between the pre open (where we check if we're in the region) and the post open (where we start the region operation), but I think that could be solved by doing the in-region check again). In these cases, we already handle this through the exception catching code in BaseResultIterators.

chrajeshbabu: Yes aggregate queries are already handled properly. This code is only for handling splits when we are in the middle of non aggregate queries. If there are splits in the starting of the query then we are throwing out the stale region exception to BaseResultIterators which handles creating the proper parallel scans.
+                if(scan.getAttribute(NON_AGGREGATE_QUERY)!=null) {

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/GroupedAggregateRegionObserver.java
chrajeshbabu: There is a chance of split can happen after preScannerOpen and before postScannerOpen. I think we can throw stale region boundary exception all the cases because even when we throw NSRE we again need to open the scanner which checks the region boundaries and throw the stale region boundary exception any way. Will check it once.

phoenix-core/src/main/java/org/apache/phoenix/iterate/TableResultIterator.java
JamesRTaylor: Ah, I misread that. The case of an ordered aggregate falls into the same category, however, as we're aggregating in-place. I don't think we need any additional code in the unordered aggregate code path in GroupedAggregateRegionObserver (other than the duplicate in-region check I mentioned above).

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/GroupedAggregateRegionObserver.java
chrajeshbabu: region.startRegionOperation() prevent from splitting and also if the region already closed we are throwing NSRE. I have moved into try block only to catch NSRE and throw Stale region boundary exception.

phoenix-core/src/main/java/org/apache/phoenix/compile/ScanRanges.java
chrajeshbabu: I will try moving outside of this method.

phoenix-core/src/main/java/org/apache/phoenix/iterate/TableResultIterator.java
chrajeshbabu: You mean we can check the in region boundary again in post scanner open and if it's out of range throw stale region boundary exception which will be handled by BaseResultIterators?

phoenix-core/src/main/java/org/apache/phoenix/iterate/TableResultIterator.java
chrajeshbabu: handleSplitRegionBoundaryFailureDuringInitialization will be true for scanning intermediate chunks in ChunkedResultIterator. When we create scanner for new chunk we might see stale region boundaries in that case also we need to forcibly recreate the iterator with new boundaries. In normal case if we get StaleRegionBoundary exception while creating scanner BaseResultIterator handles it.

phoenix-core/src/main/java/org/apache/phoenix/iterate/TableResultIterator.java
chrajeshbabu: Will raise improvement action for this James.

phoenix-core/src/main/java/org/apache/phoenix/iterate/TableResultIterator.java
chrajeshbabu: Will move the check to ScanUtil.

phoenix-core/src/main/java/org/apache/phoenix/iterate/TableResultIterator.java
chrajeshbabu: I will add.

phoenix-core/src/main/java/org/apache/phoenix/iterate/TableResultIterator.java
chrajeshbabu: This is required for ChunkedResultIterator if we are going to deprecate or not use it we don't need these changes. Other than that we might need to deal with it in PhoenixRecordReader(currently we skipping range check but for local indexes it's compulsory to handle otherwise we might miss one of the daughter records)

phoenix-core/src/main/java/org/apache/phoenix/iterate/BaseResultIterators.java
chrajeshbabu: First one is actual start key of the scan second one is end key of region. After this patch we don't need EXPECTED_UPPER_REGION_KEY but kept it for compatibility.

phoenix-core/src/main/java/org/apache/phoenix/iterate/TableResultIterator.java
JamesRTaylor: Yes, exactly. Except for the ordered aggregation case, which needs to be handled as the non aggregate is handled (i.e. we'd still throw the stale region boundary exception, but the client would need to react to it as you've done for the non aggregate case. We need to know which exact row key we're on though (which the client won't know). Seems like the more feasible approach would be to include the row key in the exception and deserialize it on the client side.

phoenix-core/src/main/java/org/apache/phoenix/iterate/TableResultIterator.java
JamesRTaylor: See other comment below, but how about we serialize the row key we found which is out-of-region in the StaleRegionBoundaryCacheException instead of mucking with the startRowSuffix?

phoenix-core/src/main/java/org/apache/phoenix/iterate/BaseResultIterators.java
chrajeshbabu: If I move creating parallel scans for this special case I need to duplicate lot of code. That's why I have added special case as part of creating existing parallel scans only.

phoenix-core/src/main/java/org/apache/phoenix/iterate/BaseResultIterators.java
chrajeshbabu: This check detects whether scan boundaries  are equal to context scan boundaries or not. If they are same we are going by getting all parallel scans for the table. Will document it.

phoenix-core/src/main/java/org/apache/phoenix/iterate/BaseResultIterators.java
chrajeshbabu: Sure will move to ScanUtil

phoenix-core/src/main/java/org/apache/phoenix/iterate/BaseResultIterators.java
chrajeshbabu: This null check is not required will remove it.

phoenix-core/src/main/java/org/apache/phoenix/iterate/BaseResultIterators.java
JamesRTaylor: If we don't need EXPECTED_UPPER_REGION_KEY, let's remove it.

phoenix-core/src/main/java/org/apache/phoenix/iterate/BaseResultIterators.java
```diff
@@ -556,35 +564,55 @@ private static String toString(List<byte[]> gps) {
...
+                        if(Bytes.compareTo(scan.getStartRow(), regionInfo.getStartKey())<=0) {
+                            newScan.setAttribute(SCAN_ACTUAL_START_ROW, regionInfo.getStartKey());
+                            newScan.setStartRow(regionInfo.getStartKey());
+                        }
+                        if(scan.getStopRow().length == 0 || (regionInfo.getEndKey().length != 0 && Bytes.compareTo(scan.getStopRow(), regionInfo.getEndKey())>0)) {
+                            newScan.setStopRow(regionInfo.getEndKey());
+                        }
+                    }   
+                    scans = addNewScan(parallelScans, scans, newScan, endKey, true, regionLocation);
+                 } else {
```
JamesRTaylor: Can we remove the if (isLocalIndex) checks below since the above if statement would have been entered instead?

phoenix-core/src/main/java/org/apache/phoenix/iterate/BaseResultIterators.java
JamesRTaylor: Seems like you've forked the bulk of the code completely. The only think you'd duplicate would be the loop and a bit of setup code up top (that could be factored out). What's the entry point for this? Is the problem that the entry point is the same?

phoenix-core/src/main/java/org/apache/phoenix/iterate/TableResultIterator.java
chrajeshbabu: Agree with you James. I think I can raise another issue for this and work on it. Wdyt?

phoenix-core/src/main/java/org/apache/phoenix/iterate/BaseResultIterators.java
chrajeshbabu: getParalleScans is the entry point. Let me move it out.

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/BaseScannerRegionObserver.java
chrajeshbabu: Added the code comments and checking references is not expensive. We need to do this for local indexes case only.

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/BaseScannerRegionObserver.java
chrajeshbabu: We are getting the results after seek only. Will check once again whether we can handle this in local index scanner.

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/GroupedAggregateRegionObserver.java
chrajeshbabu: Currently changed the code to throw stale region exception all when we get NSRE because hbase client may recreate the scanners with wrong region boundaries on NSRE. I have handling this in BaseScannerRegionObserver so need not handle in all observers.

phoenix-core/src/main/java/org/apache/phoenix/iterate/ChunkedResultIterator.java
chrajeshbabu: Currently removed the changes required for ChunkedResultIterator. The changes in the chunked result iterator are just removing code not required.

chrajeshbabu:
@JamesRTaylor  Thanks for the review. Committed the changes handling the review comments.
Refactored the code and added code comments where ever possible. 

Now not handling the special cases required for ChunkedResultItertor. 

bq. Another potential, different approach would be for Phoenix to universally handle the split during scan case (rather than letting the HBase client scanner handle it for non aggregate case and Phoenix handle it for the aggregate case). Would that simplify things?
Now throwing stale region bound exception all the cases when ever we get NSRE then phoenix client can handle the NSRE than hbase client handling with wrong region boundaries.

The ordered aggregate queries issue you are mentioning can be handled as part of other issue.

Please review the latest code.
Thanks.

JamesRTaylor:
+1. Looks great, @chrajeshbabu. Thanks for the excellent work.

chrajeshbabu:
Thanks @JamesRTaylor  for review. Will rebase the patch and commit.

PR #159, opened 2016-04-08
PHOENIX-2535 Create shaded clients (thin + thick)

ss77892:
Shaded client + PHOENIX-2267


phoenix-assembly/pom.xml
```diff
@@ -31,76 +32,30 @@
   <artifactId>phoenix-assembly</artifactId>
   <name>Phoenix Assembly</name>
   <description>Assemble Phoenix artifacts</description>
-  <packaging>pom</packaging>
+  <packaging>jar</packaging>
```
enis: hbase-assembly/pom.xml is still pom here.

phoenix-query-server/pom.xml
```diff
@@ -0,0 +1,116 @@
...
+  <parent>
+    <groupId>org.apache.phoenix</groupId>
+    <artifactId>phoenix</artifactId>
+    <version>4.8.0-HBase-1.1-SNAPSHOT</version>
+  </parent>
+  <artifactId>phoenix-query-server</artifactId>
+  <name>Phoenix Query Server</name>
+  <description>A query server for exposing Phoenix to thin clients</description>
+
+  <licenses>
```
enis: these are not coming from the apache parent pom? We only have this in the server, not in the client module.

phoenix-server-client/pom.xml
```diff
@@ -26,29 +26,126 @@
...
+          <goals>
+            <goal>shade</goal>
+          </goals>
+          <configuration>
+            <createSourcesJar>false</createSourcesJar>
+            <shadedArtifactAttached>false</shadedArtifactAttached>
+            <promoteTransitiveDependencies>true</promoteTransitiveDependencies>
+            <shadeTestJar>false</shadeTestJar>
+            <finalName>phoenix-${project.version}-thin-client</finalName>
+            <!--artifactSet>
```
enis: why commented out? We can remove if not needed anymore because it is enumerated below.

pom.xml
```diff
@@ -27,9 +27,11 @@
     <module>phoenix-flume</module>
     <module>phoenix-pig</module>
     <module>phoenix-server-client</module>
-    <module>phoenix-server</module>
```
enis: I like the phoenix-queryserver and phoenix-queryserver-client as names of the modules. phoenix-server-client is hard to parse without knowing phoenix-server is phoenix query server.

phoenix-assembly/pom.xml
ss77892: Fixed in next commit

phoenix-server-client/pom.xml
ss77892: Fixed in next commit

pom.xml
ss77892: Fixed in next commit

phoenix-client/dependency-reduced-pom.xml
```diff
@@ -0,0 +1,263 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
enis: Is this used anymore? I could not find any references, if so we can get rid of it.

ss77892:
Updated version for better review

phoenix-client/pom.xml
```diff
@@ -0,0 +1,381 @@
...
+  xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+  <modelVersion>4.0.0</modelVersion>
+  <parent>
+    <groupId>org.apache.phoenix</groupId>
+    <artifactId>phoenix</artifactId>
+    <version>4.8.0-HBase-1.2-SNAPSHOT</version>
+  </parent>
+  <artifactId>phoenix-client</artifactId>
+  <name>Phoenix Client</name>
+  <description>Phoenix Client</description>
```
joshelser: Nit: A better description would be awesome

phoenix-server/pom.xml
```diff
@@ -7,52 +30,98 @@
...
-    </license>
-  </licenses>
-
-  <organization>
-    <name>Apache Software Foundation</name>
-    <url>http://www.apache.org</url>
-  </organization>
-
+  <name>Phoenix Server</name>
+  <description>Phoenix Server</description>
```
joshelser: Nit: A better description would be awesome

bin/queryserver.py
```diff
@@ -119,7 +119,7 @@
 
 # The command is run through subprocess so environment variables are automatically inherited
 java_cmd = '%(java)s -cp ' + hbase_config_path + os.pathsep + hadoop_config_path + os.pathsep + \
-    phoenix_utils.phoenix_queryserver_jar + os.pathsep + phoenix_utils.phoenix_client_jar + \
+    phoenix_utils.phoenix_client_jar + \
```
joshelser: Why is the queryserver jar no longer included when launching the queryserver?

phoenix-client/pom.xml
```diff
@@ -0,0 +1,381 @@
...
+      <groupId>org.apache.phoenix</groupId>
+      <artifactId>phoenix-pig</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>org.apache.phoenix</groupId>
+      <artifactId>phoenix-spark</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>org.apache.phoenix</groupId>
+      <artifactId>phoenix-queryserver</artifactId>
```
joshelser: Why does the phoenix-client depend on the phoenix-queryserver? This seems wrong to me -- there shouldn't be anything in here that a client needs to have (if there is, we should fix that).

phoenix-client/pom.xml
```diff
@@ -0,0 +1,381 @@
...
+    </dependency>
+    <dependency>
+      <groupId>org.apache.phoenix</groupId>
+      <artifactId>phoenix-queryserver-client</artifactId>
```
joshelser: Actually, same with this too -- I don't think we want the queryserver-client dependency here. The bundled phoenix-$VERSION-thin-client.jar in the binary dist is already shaded; we're just missing a GAV for downstream clients to refer to it via Maven. I think a better approach would be to make a phoenix-thin-client module (like this phoenix-client module for the thick-client), rather than make thin-client users eat all of the thick-client dependencies too. I do think we have to make this extra module now though, it can be done as a follow-on.

phoenix-assembly/src/build/components/all-common-jars.xml
```diff
@@ -88,12 +93,12 @@
       <directory>${project.basedir}/../phoenix-server/target/</directory>
```
joshelser: Need to change this directory. There is no phoenix-queryserver*.jar included in the lib/ directory in the tarball.

bin/queryserver.py
ss77892: those classes are part of  client jar,  since it uses calcite which is shaded in client jar. Or  we need to bundle that package into queryserver jar.

phoenix-client/pom.xml
ss77892: Well, actually nothing has been changed. Previously phoenix-client.jar was created as part of phoenix-assembly module and it was including query-server as well.Now the final jar created  in phoenix-client module, so the dependency moved there.

joshelser: > Previously phoenix-client.jar was created as part of phoenix-assembly module and it was including query-server as well

Oh, you're right. Why the heck where they in there! IMO, we should fix that now and remove them. It makes no sense to me that we should push those classes (and dependencies) onto the user. We have the queryserver jar specifically for that purpose (and it is shaded already).

bin/queryserver.py
joshelser: Right, realized that later :). Given my opinion on removing the queryserver classes from the client jar, I think this should be switched back to queryserver + client. It makes a more natural usage, IMO.

phoenix-client/pom.xml
ss77892: Ok. I'll try to do that to  see what will be broken :)

joshelser: Thanks! Let me know when you push the next set of changes, and I can pull it down locally to poke at it.

phoenix-queryserver/pom.xml
```diff
@@ -24,6 +26,50 @@
...
+              <goal>shade</goal>
+            </goals>
+            <configuration>
+              <finalName>phoenix-${project.version}-queryserver</finalName>
+              <shadedArtifactAttached>false</shadedArtifactAttached>
+              <promoteTransitiveDependencies>true</promoteTransitiveDependencies>
+              <shadeTestJar>false</shadeTestJar>
+              <artifactSet>
+                <includes>
+                  <include>org.apache.phoenix:phoenix-server</include>
```
joshelser: Shouldn't these be phoenix-queryserver and phoenix-queryserver-client?

ss77892: Oh. Thanks. The default artifact for module is added automatically, so we can just remove that at all. I will check. Will do a fresh look on all poms after all those moving back and forth.

stoty:
Already merged.

PR #168, opened 2016-05-10
PHOENIX-1734 Local index improvements(Rajeshbabu)

chrajeshbabu:
This is the patch for new implementation of local index where we store local index data in the separate column families in the same table than different table. 

This patch needs https://issues.apache.org/jira/browse/HBASE-15600 to be committed to HBase to write the local index updates directly to the region and make it transactional with data table updates. Phoenix uses MiniBatchOperationInProgress#addOperationsFromCP API introduced at HBASE-15600.


JamesRTaylor:
How about a version that doesn't depend on HBASE-15600 so that we can continue to support older versions of HBase (and can do a 4.8 release)? What we've typically done, is once the change makes it to an HBase release, we update the pom on that version but provide a fallback implementation (conditional on the HBase version at runtime). Then we document that writes aren't transactional unless you're using HBase version XXX or above.

phoenix-core/src/it/java/org/apache/phoenix/end2end/SortMergeJoinIT.java
```diff
@@ -187,13 +186,13 @@ public void initTable() throws Exception {
                 "CREATE LOCAL INDEX \"idx_supplier\" ON " + JOIN_SUPPLIER_TABLE_FULL_NAME + " (name)"
                 }, {
                 "SORT-MERGE-JOIN (LEFT) TABLES\n" +
-                "    CLIENT PARALLEL 1-WAY RANGE SCAN OVER " + MetaDataUtil.LOCAL_INDEX_TABLE_PREFIX + JOIN_SUPPLIER_TABLE_DISPLAY_NAME + " [-32768]\n" +
+                "    CLIENT PARALLEL 1-WAY RANGE SCAN OVER " +JOIN_SUPPLIER_TABLE_DISPLAY_NAME + " [1]\n" +
```
JamesRTaylor: We should update the explain plan code to output something else when a local index is used, otherwise it's too difficult to tell. Is the local index being used?

Something like ... OVER LOCAL INDEX <index name> ...

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/UngroupedAggregateRegionObserver.java
```diff
@@ -165,6 +165,9 @@ public RegionScanner preScannerOpen(ObserverContext<RegionCoprocessorEnvironment
             throws IOException {
         s = super.preScannerOpen(e, scan, s);
         if (ScanUtil.isAnalyzeTable(scan)) {
+            if (!ScanUtil.isLocalIndex(scan)) {
+                scan.getFamilyMap().clear();
+            }
```
JamesRTaylor: Can't we project the right column families from the client into the scan? Our UPDATE STATISTICS command specifies whether or not stats for indexes should be updated or not.

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/UngroupedAggregateRegionObserver.java
```diff
@@ -202,7 +207,10 @@ protected RegionScanner doPostScannerOpen(final ObserverContext<RegionCoprocesso
             ScanUtil.setRowKeyOffset(scan, offsetToBe);
         }
         final int offset = offsetToBe;
-
+        if(MetaDataUtil.hasLocalIndexColumnFamily(region.getTableDesc())) {
+        	nonLocalIndexFamilies = MetaDataUtil.getNonLocalIndexColumnFamilies(region.getTableDesc());
+        }
+
```
JamesRTaylor: Why is this required?

phoenix-core/src/main/java/org/apache/phoenix/execute/DelegateHTable.java
```diff
@@ -297,4 +297,28 @@ public boolean checkAndDelete(byte[] row, byte[] family, byte[] qualifier,
 		return delegate.checkAndDelete(row, family, qualifier, compareOp, value, delete);
 	}
 
+    @Override
```
JamesRTaylor: These should call the method on the delegate.

phoenix-core/src/main/java/org/apache/phoenix/index/IndexMaintainer.java
```diff
@@ -861,7 +871,12 @@ public Put buildUpdateMutation(KeyValueBuilder kvBuilder, ValueGetter valueGette
                     put.setDurability(!indexWALDisabled ? Durability.USE_DEFAULT : Durability.SKIP_WAL);
                 }
                 //this is a little bit of extra work for installations that are running <0.94.14, but that should be rare and is a short-term set of wrappers - it shouldn't kill GC
-                put.add(kvBuilder.buildPut(rowKey, ref.getFamilyWritable(), cq, ts, value));
+                if(this.isLocalIndex) {
+                    ColumnReference columnReference = this.coveredColumnsMap.get(ref);
+					put.add(kvBuilder.buildPut(rowKey, columnReference.getFamilyWritable(), cq, ts, value));
```
JamesRTaylor: Tell me more about this. We keep a mapping on the client-side for column families? I thought we could generate the cf of the local index based on the cf of the data table? Is this to save re-generating the column family name all the time?

phoenix-core/src/main/java/org/apache/phoenix/iterate/ExplainTable.java
```diff
@@ -198,8 +198,14 @@ private void appendPKColumnValue(StringBuilder buf, byte[] range, Boolean isNull
             type.coerceBytes(ptr, type, sortOrder, SortOrder.getDefault());
             range = ptr.get();
         }
-        Format formatter = context.getConnection().getFormatter(type);
-        buf.append(type.toStringLiteral(range, formatter));
+        if (changeViewIndexId) {
```
JamesRTaylor: I like this trick. We should always do this.

phoenix-core/src/main/java/org/apache/phoenix/schema/MetaDataClient.java
```diff
@@ -1016,7 +1016,7 @@ private MutationState buildIndexAtTimeStamp(PTable index, NamedTableNode dataTab
         // connection so that our new index table is visible.
         Properties props = new Properties(connection.getClientInfo());
         props.setProperty(PhoenixRuntime.CURRENT_SCN_ATTRIB, Long.toString(connection.getSCN()+1));
-        PhoenixConnection conn = DriverManager.getConnection(connection.getURL(), props).unwrap(PhoenixConnection.class);
+        PhoenixConnection conn = new PhoenixConnection(connection, connection.getQueryServices(), props);
```
JamesRTaylor: Just curious on this change.

phoenix-core/src/main/java/org/apache/phoenix/hbase/index/write/ParallelWriterIndexCommitter.java
```diff
@@ -116,7 +117,10 @@ public void write(Multimap<HTableInterfaceReference, Mutation> toWrite) throws S
             // doing a complete copy over of all the index update for each table.
             final List<Mutation> mutations = kvBuilder.cloneIfNecessary((List<Mutation>)entry.getValue());
             final HTableInterfaceReference tableReference = entry.getKey();
-            final RegionCoprocessorEnvironment env = this.env;
+            if (env !=null && tableReference.getTableName().equals(
+                env.getRegion().getTableDesc().getNameAsString())) {
+                continue;
+            }
```
JamesRTaylor: Can we conditionally *not* filter our local index rows depending on the HBase version?

phoenix-core/src/main/java/org/apache/phoenix/index/PhoenixTransactionalIndexer.java
```diff
@@ -160,6 +163,9 @@ public void preBatchMutate(ObserverContext<RegionCoprocessorEnvironment> c,
 
             // get the index updates for all elements in this batch
             indexUpdates = getIndexUpdates(c.getEnvironment(), indexMetaData, getMutationIterator(miniBatchOp), txRollbackAttribute);
+            
+            IndexUtil.addLocalUpdatesToCpOperations(c, miniBatchOp, indexUpdates,
+                m.getDurability() != Durability.SKIP_WAL);
```
JamesRTaylor: Can we conditionally *not* run this code depending on the HBase version?

phoenix-core/src/it/java/org/apache/phoenix/end2end/SortMergeJoinIT.java
chrajeshbabu: @JamesRTaylor  Yes agree.  Is this fine "OVER LOCAL INDEX OF DATA_TABLE"? or can we include index table name as well?

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/UngroupedAggregateRegionObserver.java
chrajeshbabu: This is not required. Removed it. Just while debugging one of the issue added it.

phoenix-core/src/main/java/org/apache/phoenix/execute/DelegateHTable.java
chrajeshbabu: These methods required for 1.2.2-SNAPSHOT version so currently not required with older versions. will remove it.

phoenix-core/src/main/java/org/apache/phoenix/index/IndexMaintainer.java
chrajeshbabu: Yes James. It's to save regenerating the column family name all the time.

phoenix-core/src/main/java/org/apache/phoenix/schema/MetaDataClient.java
chrajeshbabu: DriverManager.getConnection has more overhead than initializing the PhoenixConnection right that's why changed.

phoenix-core/src/main/java/org/apache/phoenix/hbase/index/write/ParallelWriterIndexCommitter.java
chrajeshbabu: Yes we should do that. Currently made a patch without HBASE-15600 so once it's ready then will add condition check.

phoenix-core/src/main/java/org/apache/phoenix/index/PhoenixTransactionalIndexer.java
chrajeshbabu: Yes

chrajeshbabu:
James as we discussed I have made a patch working with older versions of HBase first and handled review comments here. Will create new pull request with that patch.

Thanks for reviews.

phoenix-core/src/it/java/org/apache/phoenix/end2end/SortMergeJoinIT.java
JamesRTaylor: Yes, I think it'd be good to mention the local index name as there could be more than one. If you include the full name of the local index (<schema name>.<table name> which is what you get from PTable.getName()), then that's unique so you wouldn't need to include the data table name too.

phoenix-core/src/it/java/org/apache/phoenix/end2end/index/MutableIndexIT.java
```diff
@@ -707,16 +702,17 @@ public void testIndexHalfStoreFileReader() throws Exception {
...
-                  if (regions.size() < numRegions) {
-                    break;
-                  }
-                  Threads.sleep(1000);
+                if(!localIndex) {
+                    long waitStartTime = System.currentTimeMillis();
+                    // wait until merge happened
+                    while (System.currentTimeMillis() - waitStartTime < 10000) {
+                      List<HRegionInfo> regions = admin.getTableRegions(indexTable);
+                      System.out.println("Waiting:" + regions.size());
```
JamesRTaylor: Remove System.out.println() or replace with logging.

phoenix-core/src/main/java/org/apache/phoenix/util/IndexUtil.java
```diff
@@ -667,4 +651,28 @@ public static String getIndexColumnExpressionStr(PColumn col) {
         return col.getExpressionStr() == null ? IndexUtil.getCaseSensitiveDataColumnFullName(col.getName().getString())
                 : col.getExpressionStr();
     }
+
+    public static void addLocalUpdatesToCpOperations(ObserverContext<RegionCoprocessorEnvironment> c,
```
JamesRTaylor: Minor nit, but if this is called in only one place, can we move this there and make it private?

phoenix-core/src/main/java/org/apache/phoenix/query/ConnectionQueryServicesImpl.java
```diff
@@ -2455,6 +2409,19 @@ public Void call() throws Exception {
...
                                 if (currentServerSideTableTimeStamp < MetaDataProtocol.MIN_SYSTEM_TABLE_TIMESTAMP_4_8_0) {
+                                    Properties props = PropertiesUtil.deepCopy(metaConnection.getClientInfo());
+                                    props.remove(PhoenixRuntime.CURRENT_SCN_ATTRIB);
+                                    props.remove(PhoenixRuntime.TENANT_ID_ATTRIB);
+                                    PhoenixConnection conn =
+                                            new PhoenixConnection(ConnectionQueryServicesImpl.this,
+                                                    metaConnection.getURL(), props, metaConnection
+                                                            .getMetaDataCache());
+                                    try {
+                                        UpgradeUtil.upgradeLocalIndexes(conn);
```
JamesRTaylor: Should we only disable all local indexes here as a full rebuild would take a lot of time? Then we can let users use your utility to rebuild them?

JamesRTaylor:
Looks very good, @chrajeshbabu! Thanks so much for this. Are you ok making the changes I mentioned?

phoenix-core/src/main/java/org/apache/phoenix/query/ConnectionQueryServicesImpl.java
chrajeshbabu: Agree with you James. Will do it in other patch.

phoenix-core/src/it/java/org/apache/phoenix/end2end/SortMergeJoinIT.java
chrajeshbabu: I will work on this in other patch James. Again I need to change all the test cases.

phoenix-core/src/main/java/org/apache/phoenix/util/UpgradeUtil.java
```diff
@@ -292,6 +294,119 @@ private static void preSplitSequenceTable(PhoenixConnection conn, int nSaltBucke
...
+               
+                if (!coveredColumns.isEmpty()) {
+                    createIndex.append(" INCLUDE(");
+                    for (int i = 0; i < coveredColumns.size(); i++) {
+                        createIndex.append(coveredColumns.get(i));
+                        if (i < coveredColumns.size() - 1) {
+                            createIndex.append(",");
+                        }
+                    }
+                    createIndex.append(")");
```
JamesRTaylor: Can we just add the ASYNC keyword here so that the rebuilds are asynchronous, can be started when the user is ready, and will run through MR?

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/UngroupedAggregateRegionObserver.java
JamesRTaylor: Can we file a JIRA for this? MetaDataClient.updateStatisticsInternal() would just need to project the correct column family here.

PR #170, opened 2016-05-31
[PHOENIX-2955] use xmllint for parsing phoenix name in pom.xml at make_rc.sh 

shijinkui:
fix problem:
"Cannot open file -q at /System/Library/Perl/Extras/5.18/XML/XPath.pm line 53."
use `xmllint` relace xpath for parse phoenix name in pom.xml


JamesRTaylor:
@joshelser - WDYT? If it's fine, would you mind committing?

PR #171, opened 2016-05-31
Phoenix-2405

RCheungIT:
https://issues.apache.org/jira/browse/PHOENIX-2405


phoenix-core/src/main/java/org/apache/phoenix/coprocessor/ScanRegionObserver.java
```diff
@@ -234,6 +234,12 @@ protected RegionScanner doPostScannerOpen(final ObserverContext<RegionCoprocesso
                     scan.getAttribute(QueryConstants.LAST_SCAN) != null);
         }
         final OrderedResultIterator iterator = deserializeFromScan(scan,innerScanner);
+        
+        if(iterator != null){
+            iterator.setMemoryManager(GlobalCache.getTenantCache(c.getEnvironment(), tenantId).getMemoryManager());
+        }
+
+
```
maryannxue: Is it possible to put this MemoryManager into constructor other than using the set method?

phoenix-core/src/main/java/org/apache/phoenix/iterate/DeferredByteBufferSegmentQueue.java
```diff
@@ -0,0 +1,123 @@
...
+import java.util.*;
+
+public abstract class DeferredByteBufferSegmentQueue<T> extends BufferSegmentQueue<T> {
+
+    final MemoryChunk chunk;
+
+    public DeferredByteBufferSegmentQueue(int index, int thresholdBytes,
+                                          boolean hasMaxQueueSize, MemoryManager memoryManager) {
+        super(index, thresholdBytes, hasMaxQueueSize);
+        chunk = memoryManager.allocate(thresholdBytes);
```
maryannxue: "thresholdBytes" might be confusing here. There are actually two occurrences of memory usage here, first one being in-memory priority queue for sorting, once that part, the size of which is rather an estimate (based on the priority queue data structure) than an actual value, has reached the threshold, the priority queue content should be written to a some kind of file OutputStream, which is now DeferredFileOutputStream. The second memory usage is that used by DeferredFileOutputStream itself, since its content will first stay in memory before its own threshold is reached.
Therefore, we might need to allocate twice (it's not real allocate anyway, it's for tracking memory usage actually). But a better way to do this is to make use of SpoolingResultIterator logic to handle the entire second part as mentioned above. They should be exactly the same logic except that SpoolingResultIterator writes and reads Tuples and what you need here is something that writes and reads ResultEntry. So see if you can apply some abstraction here.

phoenix-core/src/main/java/org/apache/phoenix/iterate/DeferredByteBufferSegmentQueue.java
```diff
@@ -0,0 +1,123 @@
...
+        protected void init(long readIndex) {
+            this.isEnd = false;
+            this.readIndex = readIndex;
+            this.next = null;
+
+            try {
+                BufferedInputStream bufferedInputStream = new BufferedInputStream(new FileInputStream(file));
+                bufferedInputStream.skip(readIndex);
+                this.dataInput = new DataInputStream(bufferedInputStream);
+            } catch (IOException e) {
```
maryannxue: Like I said in the above comment, you are not using DeferredFileOutputStream as a deferred stream, you are just using it as an ordinary FileOutputStream. So maybe forget about handling the DeferredFileOutputStream yourself and use the adapted SpoolingResultIterator instead.

phoenix-core/src/main/java/org/apache/phoenix/iterate/DeferredByteBufferSegmentQueue.java
```diff
@@ -0,0 +1,123 @@
...
+    final MemoryChunk chunk;
+
+    public DeferredByteBufferSegmentQueue(int index, int thresholdBytes,
+                                          boolean hasMaxQueueSize, MemoryManager memoryManager) {
+        super(index, thresholdBytes, hasMaxQueueSize);
+        chunk = memoryManager.allocate(thresholdBytes);
+    }
+
+    abstract protected void writeToBuffer(OutputStream outputStream, T e);
+    abstract protected T readFromBuffer(DataInput dataInput);
```
maryannxue: I know they are the existing method names, but they might be even clearer if changed to "writeRecord(ToBuffer)" and "readRecord(FromBuffer)"?

phoenix-core/src/main/java/org/apache/phoenix/iterate/BufferSegmentQueue.java
```diff
@@ -0,0 +1,251 @@
...
+
+    @Override
+    public Iterator<T> iterator() {
+        if (isClosed)
+            return null;
+
+        if (!flushBuffer)
+            return getInMemoryQueue().iterator();
+
+        SegmentQueueFileIterator iterator = createSegmentQueueFileIterator(thisIterator);
```
maryannxue: Just realized this existing implementation could be very inefficient, by creating a new InputStream and skip a few bytes every time this is called. Is it possible to always return the same iterator with the current read state? Guess it will just work after using the SpoolingResultIterator logic.

phoenix-core/src/main/java/org/apache/phoenix/iterate/DeferredByteBufferSegmentQueue.java
```diff
@@ -0,0 +1,123 @@
...
+        if (totalResultSize >= thresholdBytes) {
+            this.file = File.createTempFile(UUID.randomUUID().toString(), null);
+
+            DeferredFileOutputStream spoolTo = new DeferredFileOutputStream(thresholdBytes, file) {
+                @Override
+                protected void thresholdReached() throws IOException {
+                    try {
+                        super.thresholdReached();
+                    } finally {
+                            chunk.close();
```
maryannxue: There's an indent problem here. You might not need these lines after the suggested change, but would you mind double checking other places for code indent?

phoenix-core/src/main/java/org/apache/phoenix/execute/SortMergeJoinPlan.java
```diff
@@ -572,11 +569,11 @@ public MappedByteBufferTupleQueue(int thresholdBytes) {
...
-        protected Comparator<MappedByteBufferSegmentQueue<Tuple>> getSegmentQueueComparator() {
-            return new Comparator<MappedByteBufferSegmentQueue<Tuple>>() {
+        protected Comparator<BufferSegmentQueue<Tuple>> getSegmentQueueComparator() {
+            return new Comparator<BufferSegmentQueue<Tuple>>() {
                 @Override
-                public int compare(MappedByteBufferSegmentQueue<Tuple> q1, 
-                        MappedByteBufferSegmentQueue<Tuple> q2) {
+                public int compare(BufferSegmentQueue<Tuple> q1,
+                                   BufferSegmentQueue<Tuple> q2) {
                     return q1.index() - q2.index();
```
maryannxue: Think we should eventually get rid of MappedByteBufferSegmentQueue and used the DeferredXXX version instead. Maybe that'll also simply the level of abstraction here. Go with the other changes first and see what you can do here.

JamesRTaylor:
Pull looks good, but please squash all commits into one and prefix the commit message with the JIRA number so we tie the two together: PHOENIX-2405
Improve performance and stability of server side sort for ORDER BY

phoenix-core/src/main/java/org/apache/phoenix/iterate/DeferredByteBufferSegmentQueue.java
JamesRTaylor: FYI, SpoolingResultIterator is deprecated, so no need to refactor. Feel free to copy/paste code from there, though, as conceptually it's what you need.

JamesRTaylor:
Once you get this completed, it'd be good to compare perf of old vs new. Maybe Mujtaba can help with that.

PR #175, opened 2016-06-19
PHOENIX-2405 Not for merge, just request a review to check whether I'm on the right way

RCheungIT:
https://issues.apache.org/jira/browse/PHOENIX-2405


phoenix-core/src/main/java/org/apache/phoenix/iterate/SpoolingByteBufferSegmentQueue.java
```diff
@@ -0,0 +1,357 @@
...
+import java.io.*;
+import java.util.AbstractQueue;
+import java.util.Iterator;
+
+import java.util.concurrent.atomic.AtomicInteger;
+
+import static org.apache.phoenix.monitoring.GlobalClientMetrics.*;
+
+
+public abstract class SpoolingByteBufferSegmentQueue<T>  extends AbstractQueue<T> {
```
maryannxue: The logic is very confusing here. My idea was to extend or modify the current SpoolingResultIterator so that it can take a ResultEntry and/or a Tuple as a record. But meanwhile this does not have to do with the XXXQueue here. XXXQueue deals with the priority queue logic and SpoolingXXX deals with the deferred byte buffer logic. Let me know whether you understand how it's supposed to work.

phoenix-core/src/main/java/org/apache/phoenix/iterate/SpoolingByteBufferSegmentQueue.java
```diff
@@ -0,0 +1,357 @@
...
+        public void close() {
+            memoryChunk.close();
+        }
+
+
+        @Override
+        public Iterator<T> iterator() {
+            return new Iterator<T>(){
+                AtomicInteger iteratorOffset = new AtomicInteger(offset.get());
+                private T next = advance(iteratorOffset);
```
maryannxue: One reminder: I said in my previous review that we'd better avoid discarding a old XXXSegmentQueue and starting a new one all the time when we switch between these segment queues. So we should ultimately remove this offset thing.

phoenix-core/src/main/java/org/apache/phoenix/iterate/SpoolingByteBufferSegmentQueue.java
```diff
@@ -0,0 +1,357 @@
...
+
+    protected abstract OnDiskResultQueue<T> createOnDiskResultQueue(File file);
+
+    @Override
+    public boolean offer(T t) {
+        if (closed || flushed){
+            return false;
+        }
+        boolean result = writeRecord(t, spoolTo);
+        if(result){
```
maryannxue: I'm pretty lost at why this "if" block here, means we flush to disk all the time? I'll stop here though, leaving all other doubts, coz this all (XXXQueue classes) is gonna take a rewrite I think.

phoenix-core/src/main/java/org/apache/phoenix/execute/AggregatePlan.java
```diff
@@ -133,7 +134,10 @@ public PeekingResultIterator newIterator(StatementContext context, ResultIterato
             Expression expression = RowKeyExpression.INSTANCE;
             OrderByExpression orderByExpression = new OrderByExpression(expression, false, true);
             int threshold = services.getProps().getInt(QueryServices.SPOOL_THRESHOLD_BYTES_ATTRIB, QueryServicesOptions.DEFAULT_SPOOL_THRESHOLD_BYTES);
-            return new OrderedResultIterator(scanner, Collections.<OrderByExpression>singletonList(orderByExpression), threshold);
+            String spoolDirectory = services.getProps().get(QueryServices.SPOOL_DIRECTORY, QueryServicesOptions.DEFAULT_SPOOL_DIRECTORY);
```
maryannxue: In my previous review I might not have made it very clear that the idea of moving MemoryManager into constructors was more of a question rather than advice. Since I hadn't studied the code so carefully, I was trying to ask you if that would be better, or worse? Looks like these parameters are getting all over the place and my suggestion was not a good one?

phoenix-core/src/main/java/org/apache/phoenix/execute/SortMergeJoinPlan.java
```diff
@@ -45,10 +45,7 @@
 import org.apache.phoenix.exception.SQLExceptionInfo;
 import org.apache.phoenix.execute.TupleProjector.ProjectedValueTuple;
 import org.apache.phoenix.expression.Expression;
-import org.apache.phoenix.iterate.DefaultParallelScanGrouper;
-import org.apache.phoenix.iterate.MappedByteBufferQueue;
-import org.apache.phoenix.iterate.ParallelScanGrouper;
-import org.apache.phoenix.iterate.ResultIterator;
+import org.apache.phoenix.iterate.*;
```
maryannxue: Please do always check your patch before you submit. This is not the right coding style, and such changes should never appear in a patch/pull request.

phoenix-core/src/main/java/org/apache/phoenix/iterate/SpoolingByteBufferSegmentQueue.java
RCheungIT: Hi @maryannxue, I think I don't get your idea here.  If the modify of SpoolingResultIterator is done, where the modified SpoolingResultIterator should be? What I did here, is modify the SpoolingResultIterator as SpoolingByteBufferSegmentQueue,  so the SpoolingByteBufferSegmentQueue deals with the priority queue logic and the deferred byte buffer logic in the same time.

RCheungIT: Hi @maryannxue, it seems that I get your idea now. Do you mean I should not change the sort logic in memory?  What I only need to do is modify the flush logic, flush the sorted result to SpoolingResultIterator. Is that correct? Thanks.

PR #176, opened 2016-06-22
4.7.0-HBase

KTW81:


stoty:
This may or may not have been merged, but it's definitely no longer relevant.

PR #177, opened 2016-06-25
Support DISTINCT for SUM

Boruch28:


docs/phoenix.csv
```diff
@@ -844,9 +844,10 @@ MIN(NAME)
 "
 
 "Functions (Aggregate)","SUM","
-SUM( { numericTerm } )
+SUM( [ DISTINCT ] { numericTerm } )
```
ramkrish86: For now parser changes are fine. But the CALCITE work may lead to some changes here. But for now fine.

docs/phoenix.csv
```diff
@@ -844,9 +844,10 @@ MIN(NAME)
...
 ","
 The sum of all values.
+When DISTINCT is used, it counts only distinct values.
```
ramkrish86: How are we ensuring that this works only on Numeric data types?

phoenix-core/src/main/java/org/apache/phoenix/expression/aggregator/DecimalSumDistinctAggregator.java
```diff
@@ -0,0 +1,105 @@
...
+import java.math.BigDecimal;
+import java.util.HashSet;
+import java.util.Set;
+
+/**
+ * 
+ * Aggregator that sums BigDecimal values distinct
+ *
+ */
+public class DecimalSumDistinctAggregator extends BaseAggregator {
```
ramkrish86: DistinctValueWithCountClientAggregator - why this class was not subclassed?  There are few tricks here to track the result. We may need to do similar way.

phoenix-core/src/main/java/org/apache/phoenix/expression/aggregator/DoubleSumDistinctAggregator.java
```diff
@@ -0,0 +1,100 @@
...
+import org.apache.phoenix.schema.SortOrder;
+import org.apache.phoenix.schema.tuple.Tuple;
+import org.apache.phoenix.schema.types.PDataType;
+import org.apache.phoenix.schema.types.PDouble;
+import org.apache.phoenix.util.SizedUtil;
+
+import java.util.HashSet;
+import java.util.Set;
+
+public class DoubleSumDistinctAggregator extends BaseAggregator {
```
ramkrish86: Same here. DistinctValueWithCountClientAggregator may be the right one to use here.

phoenix-core/src/main/java/org/apache/phoenix/expression/aggregator/NumberSumDistinctAggregator.java
```diff
@@ -0,0 +1,117 @@
...
+import java.util.HashSet;
+import java.util.Set;
+
+/**
+ * 
+ * Aggregator that sums integral number values distinct
+ * 
+ *
+ */
+abstract public class NumberSumDistinctAggregator extends BaseAggregator {
```
ramkrish86: DistinctValueWithCountClientAggregator to be used here. And some modifications there.

phoenix-core/src/main/java/org/apache/phoenix/expression/function/DistinctSumAggregateFunction.java
```diff
@@ -0,0 +1,148 @@
...
+import org.apache.phoenix.schema.types.PUnsignedDouble;
+import org.apache.phoenix.schema.types.PUnsignedFloat;
+
+import java.math.BigDecimal;
+import java.util.List;
+
+/**
+ * Built-in function for SUM(distinct <expression>) aggregate function.
+ */
+@BuiltInFunction(name = DistinctSumAggregateFunction.NAME, nodeClass = DistinctSumParseNode.class, args = {
```
ramkrish86: DistinctValueWithCountServerAggregator - did you see this in the server side?

ramkrish86:
Thanks for the PR @Boruch28 .  Had a look at the PR and added some comments. 
I fear that the data for mid term evaluation ended by June 27 UTC.  The PR submission was late in my opinion and also the engagement with the community or with the mentor was very less. May be if we had an initial patch some time back it would have been easy for us to guide you thro and then the further patches could have already addressed the comments I had specified here. I would welcome you to continue with the patch based on your time and see if we can take it to closure. Thanks once again @Boruch28 .

PR #185, opened 2016-07-19
[PHOENIX-3097] Incompatibilities with HBase 0.98.6

SsnL:
This fixes most of the incompatibilities. However, there is a call to `HBaseAdmin.truncateTable` (not available in 0.98.6) at https://github.com/apache/phoenix/blob/4.x-HBase-0.98/phoenix-core/src/main/java/org/apache/phoenix/util/UpgradeUtil.java#L598. This isn't easy to fix as the logic is convoluted in HBase side..

Fortunately, this only happens when users of view indexes upgrade from an older Phoenix version. We could alert users with HBase 0.98.6 before they upgrade.


JamesRTaylor:
Looks like there are merge conflicts - would you mind rebasing, @SsnL?

JamesRTaylor:
@apurtell - would you have a few spare cycles to take a look, in particular at the replacements for User.runAsLoginUser()? These changes are so that older 0.98 versions still work with Phoenix 4.8.

SsnL:
Rebasing is done.

SsnL:
I'd appreciate any comments :)

JamesRTaylor:
I confirmed offline with @apurtell that the User.runAsLoginUser() changes are fine, so I'm +1 on these changes. Will get this committed shortly unless anyone has objections.

PR #191, opened 2016-08-17
PHOENIX-3189 Perform Kerberos login before ConnectionInfo is constructed

joshelser:
Now that ConnectionInfo has the current User/UGI stored inside, we must
make sure that any automatic Kerberos login occurs before the ConnectionInfo
object is constructed. Otherwise, we will have multiple instances of
ConnectionInfo that differ only by the User, which will leak HBase/ZK
connections in the connectionQueryServicesMap.

### DIFF
--- a/phoenix-core/src/main/java/org/apache/phoenix/jdbc/PhoenixEmbeddedDriver.java
+++ b/phoenix-core/src/main/java/org/apache/phoenix/jdbc/PhoenixEmbeddedDriver.java
@@ -198,6 +198,7 @@ public abstract class PhoenixEmbeddedDriver implements Driver, SQLCloseable {
      */
     public static class ConnectionInfo {
         private static final org.slf4j.Logger logger = LoggerFactory.getLogger(ConnectionInfo.class);
+        private static final Object KERBEROS_LOGIN_LOCK = new Object();
         private static SQLException getMalFormedUrlException(String url) {
             return new SQLExceptionInfo.Builder(SQLExceptionCode.MALFORMED_CONNECTION_URL)
             .setMessage(url).build().buildException();
@@ -350,12 +351,19 @@ public abstract class PhoenixEmbeddedDriver implements Driver, SQLCloseable {
                         // Check if we need to authenticate with kerberos so that we cache the correct ConnectionInfo
                         UserGroupInformation currentUser = UserGroupInformation.getCurrentUser();
                         if (!currentUser.hasKerberosCredentials() || !currentUser.getUserName().equals(principal)) {
-                            final Configuration config = getConfiguration(props, info, principal, keytab);
-                            logger.info("Trying to connect to a secure cluster as {} with keytab {}", config.get(QueryServices.HBASE_CLIENT_PRINCIPAL),
-                                    config.get(QueryServices.HBASE_CLIENT_KEYTAB));
-                            UserGroupInformation.setConfiguration(config);
-                            User.login(config, QueryServices.HBASE_CLIENT_KEYTAB, QueryServices.HBASE_CLIENT_PRINCIPAL, null);
-                            logger.info("Successful login to secure cluster");
+                            synchronized (KERBEROS_LOGIN_LOCK) {
+                                // Double check the current user, might have changed since we checked last. Don't want
+                                // to re-login if it's the same user.
+                                currentUser = UserGroupInformation.getCurrentUser();
+                                if (!currentUser.hasKerberosCredentials() || !currentUser.getUserName().equals(principal)) {
+                                    final Configuration config = getConfiguration(props, info, principal, keytab);
+                                    logger.info("Trying to connect to a secure cluster as {} with keytab {}", config.get(QueryServices.HBASE_CLIENT_PRINCIPAL),
+                                            config.get(QueryServices.HBASE_CLIENT_KEYTAB));
+                                    UserGroupInformation.setConfiguration(config);
+                                    User.login(config, QueryServices.HBASE_CLIENT_KEYTAB, QueryServices.HBASE_CLIENT_PRINCIPAL, null);
+                                    logger.info("Successful login to secure cluster");
+                                }
+                            }
                         } else {
                             // The user already has Kerberos creds, so there isn't anything to change in the ConnectionInfo.
                             logger.info("Already logged in as {}", currentUser);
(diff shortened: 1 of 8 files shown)


phoenix-core/src/main/java/org/apache/phoenix/jdbc/PhoenixEmbeddedDriver.java
```diff
@@ -333,6 +337,55 @@ public ConnectionInfo normalize(ReadOnlyProps props) throws SQLException {
...
+                    if (null != principal) {
+                        config.set(QueryServices.HBASE_CLIENT_PRINCIPAL, principal);
+                    }
+                    if (null != keytab) {
+                        config.set(QueryServices.HBASE_CLIENT_KEYTAB, keytab);
+                    }
+                    try {
+                        // Check if we need to authenticate with kerberos so that we cache the correct ConnectionInfo
+                        UserGroupInformation currentUser = UserGroupInformation.getCurrentUser();
+                        if (!currentUser.hasKerberosCredentials() || !currentUser.getUserName().equals(principal)) {
```
JamesRTaylor: Looks like we only use config inside of this block. Can we create it here only instead?

JamesRTaylor:
Would an alternative be to not rely on the User.equals method, but instead modify ConnectionInfo.equals to do some other kind of equality check on User?

joshelser:
> Would an alternative be to not rely on the User.equals method, but instead modify ConnectionInfo.equals to do some other kind of equality check on User?

I think I initially had this idea too (wondering why User doesn't implement a "normal" value-based equals/hashCode implementation). Devaraj didn't remember exactly why, but he recalled that there was a reason that UGI (and thus User also) did this. I could try to dig into the Hadoop archives to see if I can understand why this was done.

Mostly, I think this approach keeps us in line with what HDFS and HBase does now which has some value.

phoenix-core/src/main/java/org/apache/phoenix/jdbc/PhoenixEmbeddedDriver.java
joshelser: Yes, good catch. Can definitely do this only when necessary.

JamesRTaylor:
Thanks for pursuing this tricky issue, @joshelser. I think what you have here is definitely an improvement and should be pulled in for 4.8.1, but I do think we should look at doing a value-based equality check for User instead. There's a fair amount of overhead in what you're doing and Phoenix does not do connection pooling but relies on being able to quickly/cheaply get a connection. Would you have some cycles to investigate that a bit first?

dbahir:
This solution is not thread safe and will not allow to safely create multiple instances of a driver on different threads in the JVM. 

This area should be protected, https://github.com/joshelser/phoenix/blob/d17a8d855dc4a2c8cff578dd26e14c6c2c13cc3a/phoenix-core/src/main/java/org/apache/phoenix/jdbc/PhoenixEmbeddedDriver.java#L351.

With that said I am not sure that you can support multiple users and support renewals with the way the UGI works.

If in the same JVM a driver is instantiated for User A and then another driver is instantiated for User B the last call to loginUserFromKeytab will set the the user information in the UGI.

loginUserFromKeytabAndReturnUGI can be used which will preserve the original user info in the UGI but I think will not work correctly with renewing.

Do we want the Phoenix driver to allow multiple instances instantiated with a different logged in user for each in the same JVM ?

joshelser:
> This solution is not thread safe and will not allow to safely create multiple instances of a driver on different threads in the JVM. 

Yes, that's why I directed you over here, bud. That wasn't an initial goal of these changes.

> With that said I am not sure that you can support multiple users and support renewals with the way the UGI works.

Right.. you're catching on to what I was pointing out. This is something that you should be managing inside of Storm. We cannot do this effectively inside of Phoenix. We can only put a bandaid on top.

> Do we want the Phoenix driver to allow multiple instances instantiated with a different logged in user for each in the same JVM ?

The only change I think we can do here is to prevent multiple clients from doing what you're suggesting and hope they don't shoot themselves in the foot.

joshelser:
@dbahir I believe b5be8d8 would address your concerns. WDYT?

joshelser:
> Did you get to look at the comment regarding the user login ?

Assuming you mean the below..

> With that said I am not sure that you can support multiple users and support renewals with the way the UGI works.
> If in the same JVM a driver is instantiated for User A and then another driver is instantiated for User B the last call to loginUserFromKeytab will set the the user information in the UGI.

Yes, the best way to approach this is that Phoenix's automatic Kerberos login should _only_ be used when there a single user accessing Phoenix at one time. I was thinking that Phoenix could likely use its own docs page on the interactions with Kerberos. That would be a good place to warn people and instruct them how they need to properly support multiple concurrent users on their own.

> Can you allow another to login with a different principal? Would that cause a security issue?
If we create one driver(One) with user A and then create another driver(Two) with user B the info in the UGI now is that of user B. So there can be a situation where driver One will be using credentials of user B.

My thinking is this (but I should probably test it): says userA uses phoenix, then userB starts using Phoenix. This implies that userA would be logged out and I would hope that subsequent interactions with Phoenix by the "human" userA would fail because its credentials are no longer there (the RPCs themselves would fail). In other words, my thinking is that the RPC authentication would protect us in Phoenix from this being an issue.

joshelser:
> Thanks for pursuing this tricky issue, @joshelser. I think what you have here is definitely an improvement and should be pulled in for 4.8.1

Thanks as always for your careful eye.

> but I do think we should look at doing a value-based equality check for User instead. There's a fair amount of overhead in what you're doing and Phoenix does not do connection pooling but relies on being able to quickly/cheaply get a connection. Would you have some cycles to investigate that a bit first?

To my point in the last comment, I believe it's become apparent that people are misusing this functionality. So, I believe there is some education that needs to happen about how clients of Phoenix who need to support multiple Kerberos users concurrently should interact with the system. I think this documentation is point 1. It flat-out shouldn't be used to concurrently support multiple Kerberos users -- developers must write their own code to do this correctly. It's difficult to handle this for them because of the required doAs() call. Can investigate this further too, maybe I'm overlooking something obvious.

I do need to make good on my promise to dig through Hadoop/UGI git-log and try to piece together why exactly UGI isn't doing value-based checks already. This is point 2.

I'd like to move ahead and push then (rather then let it rot on the vine), but will try to address both of these points this week if that's OK with you.

joshelser:
> My thinking is this (but I should probably test it): says userA uses phoenix, then userB starts using Phoenix. This implies that userA would be logged out and I would hope that subsequent interactions with Phoenix by the "human" userA would fail because its credentials are no longer there (the RPCs themselves would fail). In other words, my thinking is that the RPC authentication would protect us in Phoenix from this being an issue.

No dice. This is still a hole (although not a new hole as I understand it). Will track this separately.

PR #192, opened 2016-08-18
Phoenix Cursors implementation

anirudha:
//todo

### DIFF
--- a/phoenix-core/src/main/antlr3/PhoenixSQL.g
+++ b/phoenix-core/src/main/antlr3/PhoenixSQL.g
@@ -130,6 +130,11 @@ tokens
     USE='use';
     OFFSET ='offset';
     FETCH = 'fetch';
+    DECLARE = 'declare';
+    CURSOR = 'cursor';
+    OPEN = 'open';
+    PRIOR = 'prior';
+    CLOSE = 'close';
     ROW = 'row';
     ROWS = 'rows';
     ONLY = 'only';
@@ -403,6 +408,10 @@ oneStatement returns [BindableStatement ret]
     |   s=create_schema_node
     |   s=create_view_node
     |   s=create_index_node
+    |   s=cursor_open_node
+    |   s=cursor_close_node
+    |   s=cursor_fetch_node
+    |   s=declare_cursor_node
     |   s=drop_table_node
     |   s=drop_index_node
     |   s=alter_index_node
@@ -714,7 +723,26 @@ upsert_column_refs returns [Pair<List<ColumnDef>,List<ColumnName>> ret]
     :  d=dyn_column_name_or_def { if (d.getDataType()!=null) { $ret.getFirst().add(d); } $ret.getSecond().add(d.getColumnDefName()); } 
        (COMMA d=dyn_column_name_or_def { if (d.getDataType()!=null) { $ret.getFirst().add(d); } $ret.getSecond().add(d.getColumnDefName()); } )*
 ;
-	
+
+// Parse a full declare cursor expression structure.
+declare_cursor_node returns [DeclareCursorStatement ret]
+    :    DECLARE c=cursor_name CURSOR FOR s=select_node
+        {ret = factory.declareCursor(c, s); }
+    ;
+
+cursor_open_node returns [OpenStatement ret]
+    :    OPEN c=cursor_name {ret = factory.open(c);}
+    ;
+
+cursor_close_node returns [CloseStatement ret]
+    :    CLOSE c=cursor_name {ret = factory.close(c);}
+    ;
+
+cursor_fetch_node returns [FetchStatement ret]
+    :    FETCH NEXT FROM c=cursor_name {ret = factory.fetch(c,true); }
+    |    FETCH PRIOR FROM c=cursor_name {ret = factory.fetch(c,false); }
+    ;
+
 // Parse a full delete expression structure.
 delete_node returns [DeleteStatement ret]
     :   DELETE (hint=hintClause)? FROM t=from_table_name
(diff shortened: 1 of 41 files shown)


phoenix-core/src/main/antlr3/PhoenixSQL.g
```diff
@@ -726,8 +727,11 @@ upsert_column_refs returns [Pair<List<ColumnDef>,List<ColumnName>> ret]
 
 // Parse a full declare cursor expression structure.
 declare_cursor_node returns [DeclareCursorStatement ret]
-    :    DECLARE c=cursor_name CURSOR FOR s=select_node
-        {ret = factory.declareCursor(c, s); }
+@init{boolean isStatic = false;}
+    :   DECLARE c=cursor_name CURSOR
+        (STATIC {isStatic = true;})?
```
ankitsinghal: here it can be replaced with, (static=STATIC)? and use factory.declareCursor(c,s static!=null)

phoenix-core/src/main/java/org/apache/phoenix/util/CursorUtil.java
```diff
@@ -62,24 +64,44 @@
...
+
+                String columnName = table.getColumns().get(columnIndex).getName().getString();
+                whereBuilder.append(columnName).append(") <= (");
+                String timestampFilter = "TO_DATE('"+DateUtil.getDateFormatter(DateUtil.DEFAULT_DATE_FORMAT).format(new java.sql.Date(System.currentTimeMillis()))+"')";
+                whereBuilder.append(timestampFilter).append(") AND (");
+            }
+
+            List<String> listPKColNames = new ArrayList<String>(Arrays.asList(table.getPKColumns().toString().replaceAll("[\\[ \\]]", "").toUpperCase().split(",")));
+
+            //Process ORDER BY expressions in the internal select statement
```
ankitsinghal: By static cursor you mean that subsequent fetches on the cursor should see the new data upserted after the cursor is open? 
If that so then we don't need to have timestamp filter for it, we can use scan.setTimeRange(0, CURRENT_SCN!=null ?  CURRENT_SCN, compileTimeORCursorOpenTime));

phoenix-core/src/main/java/org/apache/phoenix/util/CursorUtil.java
```diff
@@ -247,11 +269,11 @@ private void updateSelectValues(Tuple currentRow) throws SQLException {
...
                     if(value == null){
-                        whereExpressionNext = whereExpressionNext.replaceFirst(":"+Integer.toString(colBinding),"NULL");
-                        whereExpressionPrior = whereExpressionPrior.replaceFirst(":"+Integer.toString(colBinding),"NULL");
+                        whereExpressionNext = whereExpressionNext.replaceFirst("::"+Integer.toString(colBinding),"NULL");
+                        whereExpressionPrior = whereExpressionPrior.replaceFirst("::"+Integer.toString(colBinding),"NULL");
                     } else{
-                        whereExpressionNext = whereExpressionNext.replaceFirst(":"+Integer.toString(colBinding),formatString(value, type));
-                        whereExpressionPrior = whereExpressionPrior.replaceFirst(":"+Integer.toString(colBinding),formatString(value, type));
+                        whereExpressionNext = whereExpressionNext.replaceFirst("::"+Integer.toString(colBinding),formatString(value, type));
+                        whereExpressionPrior = whereExpressionPrior.replaceFirst("::"+Integer.toString(colBinding),formatString(value, type));
```
ankitsinghal: I don't think , we need to modify queries for the this support.   You may need to remove this code.(And instead of replacing why we can't use prepared statements directly).

ankitsinghal:
@anirudha , thanks for the pull request. Is it possible for you to change the JIRA subject to include PHOENIX-2606 or link somehow, so that we can see the comment logged on JIRA as well.


With current implementation,  every query is modified to include primary key columns and tries to use RVC, which may produce improper results in following cases and also sometimes there is no advantage in using it as they don’t limit scan.

- duplicate values for a column
- order by on non-primary key axis.
- Primary key columns having null value.
- Aggregate queries


@samarthjain can you please confirm as per your observation for what all queries RVC cannot be used?

@anirudha 
To abstract the query complexities and for initial support of Cursors, we should not modify any query but instead we can keep the ResultSet object open for corresponding cursor(with timeout) and start caching rows as we proceed further with next() calls(FETCH NEXT FROM cursor) , the cache will be used for previous() calls(FETCH PRIOR FROM cusror) on resultSet. (cache will also be used for next() calls if we go previous() in the cache).​

Pros/Cons of above approach:-

Pros:-

- Highly abstracted, We don’t need to understand each and every query and develop logic separately for them.
- As per the current implementation, We don’t need to expend ORDER BY(on non-primary key axis) all the time to include primary key column for uniqueness. As this will cause problem at the server because we will have more keys(almost all keys) to sort every time. (RVC will not restrict in this case).
- Cache size can also be limited by the user and if we exhaust the cache , cache can be updated by using re-runing the query with LIMIT+OFFSET only 
- An optimization can be done for flat queries(including INDEX queries) using last and peeked SCAN keys(instead of RVC to handle null and duplicate properly) for updating the cache instead of LIMIT+OFFSET.
- Snapshot/Static queries can be provided by storing the compile time of OPEN CURSOR (and can be used to limit the scan upper bound for timestamp with it).

Cons:-

- As previous() is not supported on the server(Hbase), so cache overhead is there to maintain the results to support previous() at client.
- Re-calculation of results after we reach cache limit or scanner timeout.


@JamesRTaylor , WDYT? any suggestions

JamesRTaylor:
Thanks for the patch, @anirudha. I like your suggestions, @ankitsinghal. The RVC approach will be limited (as Ankit mentioned it will not work in all the cases) and it will be much harder to implement than the approach that Ankit suggested (i.e. just using a ResultSet).

To simplify things, we can initially not support the previous() operation (I think it's optional anyway). Also, cursors will not see newly inserted inserted data, but I think that's ok in an initial implementation too as you can specify in the metadata that it functions this way.

PR #193, opened 2016-08-23
Improvements to Phoenix Web App

AyolaJayamaha:


JamesRTaylor:
Please squash all commits into a single commit. Also, looks like it needs to be rebased.

chrajeshbabu:
@AyolaJayamaha  Currently we are trying to get the webapp related files form target directory in org.apache.phoenix.tracingwebapp.http.Main. Because of this we cannot use the runnable jar independently from the build. Can you also make changes in such a way that both runnable jars and webapp jars reside in build and pick them in the code?
{code}
        URL location = domain.getCodeSource().getLocation();
        String webappDirLocation = location.toString().split("target")[0] +"src/main/webapp";
        Server server = new Server(port);
        WebAppContext root = new WebAppContext();
{code}

AyolaJayamaha:
The squashed pull request is as follows  #202

PR #195, opened 2016-08-24
Zipkin

AyolaJayamaha:
Flipping zipkin chart on vertical axis would be done on the zipkin codebase.
The zipkin codebase does not support date/time in milliseconds.


phoenix-zipkin/pom.xml
```diff
@@ -0,0 +1,102 @@
...
+-->
+
+  <project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+  xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+
+    <parent>
+      <groupId>org.apache.phoenix</groupId>
+      <artifactId>phoenix</artifactId>
+      <version>4.8.0-HBase-1.2-SNAPSHOT</version>
```
mujtabachohan: Why is it tied to a specific Phoenix version?

phoenix-zipkin/pom.xml
```diff
@@ -0,0 +1,102 @@
...
+      </dependency>
+      <dependency>
+        <groupId>io.zipkin.java</groupId>
+        <artifactId>zipkin-autoconfigure-ui</artifactId>
+        <version>1.5.1</version>
+      </dependency>
+      <dependency>
+        <groupId>org.apache.phoenix</groupId>
+        <artifactId>phoenix-core</artifactId>
+        <version>4.8.0-HBase-1.2-SNAPSHOT</version>
```
mujtabachohan: Why is it tied to a specific Phoenix version?

mujtabachohan: and again Phoenix version is specified. This should be removed.

JamesRTaylor:
Please squash all commits into a single commit.

mujtabachohan:
Please merge with phoenix-tracing-webapp and name it phoenix-tracing. Use default properties for zipkin configuration in case they are not configurable.

PR #199, opened 2016-08-26
PHOENIX-2645 Wildcard characters do not match newline characters

kliewkliew:


twdsilva:
@kliewkliew  +1 I will get this committed . Any chance you can pick up PHOENIX-1874 also? None of our tests set the QueryServices.USE_BYTE_BASED_REGEX_ATTRIB  to true currently.

kliewkliew:
Sure, I'll pick up that JIRA. The unit tests evaluate both byte and string based regex (and compares the two results) but the integration tests only evaluate string based regex.

PR #200, opened 2016-08-26
PHOENIX-2641 Implicit wildcard in LIKE predicate search pattern

kliewkliew:


kliewkliew:
Needs more testing

phoenix-core/src/main/java/org/apache/phoenix/compile/ExpressionCompiler.java
```diff
@@ -517,8 +518,13 @@ public Expression visitLeave(LikeParseNode node, List<Expression> children) thro
                       return new ComparisonExpression(Arrays.asList(lhs,rhs), op);
                   }
                 }
-            } else if (index == 0 && pattern.length() == 1) {
-                return IsNullExpression.create(lhs, true, context.getTempPtr());
+            } else {
+                byte[] nullExpressionString = new byte[pattern.length()];
+                byte[] wildcard = {StringUtil.MULTI_CHAR_LIKE};
+                StringUtil.fill(nullExpressionString, 0, pattern.length(), wildcard, 0, 1, false);
+                if (pattern.equals(new String (nullExpressionString))) {
```
twdsilva: Can we just check if pattern equals StringUtil.MULTI_CHAR_LIKE here ?

kliewkliew: I guess that it is unlikely enough that a user would submit a query with `~ LIKE '%%'` instead of `~ LIKE '%'` that it isn't worth the overhead of covering that case. 

I did that in revision 53242fdb31c23879a9ae13ef6f429d585c35b307 with `pattern.equals(Character.toString(StringUtil.MULTI_CHAR_LIKE))` and tested it. That revision works if you would like to commit that.

twdsilva: +1 that makes sense. Thanks for the contribution, I will get this committed.

twdsilva:
@kliewkliew  
Can you please rebase? I cannot apply the patch to master.

kliewkliew:
@twdsilva Changes were made and rebased,

twdsilva:
@kliewkliew 

Thanks for the contribution. I did not include your last commit 523f0 , its fine to handle multiple wild card characters. I have committed this to the 4.x and 4.8 branches.

kliewkliew:
@twdsilva Thanks for committing

PR #201, opened 2016-08-26
PHOENIX-2793 Date-time functions on null dates return no data

kliewkliew:
I cannot reproduce the issue but I added a test case.


kliewkliew:
https://github.com/apache/phoenix/commit/06d37e534d1e07e8206eaefb82dd7bf3f50455b0

PR #206, opened 2016-09-01
PHOENIX-2946 Projected comparison between date and timestamp columns …

kliewkliew:
…always returns true


kliewkliew:
Are some casting operations more expensive than coercion? Should the `isCastableTo` checks be moved to their own `else-if` branches below the `isCoercibleTo` checks?

PR #208, opened 2016-09-09
PHOENIX-3046 `NOT LIKE '%'` unexpectedly returns results

kliewkliew:


phoenix-core/src/main/java/org/apache/phoenix/compile/ExpressionCompiler.java
```diff
@@ -523,7 +523,12 @@ public Expression visitLeave(LikeParseNode node, List<Expression> children) thro
                 byte[] wildcard = {StringUtil.MULTI_CHAR_LIKE};
                 StringUtil.fill(nullExpressionString, 0, pattern.length(), wildcard, 0, 1, false);
                 if (pattern.equals(new String (nullExpressionString))) {
-                    return IsNullExpression.create(lhs, true, context.getTempPtr());
+                    if (node.isNegate()) {
+                        return LiteralExpression.newConstant(false, Determinism.ALWAYS);
```
JamesRTaylor: So WHERE col IS NOT '%' always returns false? Or should it become WHERE col IS NULL?

kliewkliew: The specification is:
```
5) Case:

            a) If M and P are character strings whose lengths are variable
              and if the lengths of both M and P are 0, then

                 M LIKE P

              is true.

            b) The <predicate>

                 M LIKE P

              is true if there exists a partitioning of M into substrings
              such that:

              i) A substring of M is a sequence of 0 or more contiguous
                 <character representation>s of M and each <character repre-
                 sentation> of M is part of exactly one substring.

             ii) If the i-th substring specifier of P is an arbitrary char-
                 acter specifier, the i-th substring of M is any single
                 <character representation>.

            iii) If the i-th substring specifier of P is an arbitrary string
                 specifier, then the i-th substring of M is any sequence of
                 0 or more <character representation>s.

             iv) If the i-th substring specifier of P is neither an arbi-
                 trary character specifier nor an arbitrary string speci-
                 fier, then the i-th substring of M is equal to that sub-
                 string specifier according to the collating sequence of
                 the <like predicate>, without the appending of <space>
                 characters to M, and has the same length as that substring
                 specifier.

              v) The number of substrings of M is equal to the number of
                 substring specifiers of P.

            c) Otherwise,

                 M LIKE P

              is false.
```

Given that `LEN(NULL)` is `NULL`, `WHERE col IS NOT LIKE '%'` fails cases `a`, `b.i`, and `b.iii`; defaulting to case `c` and always returning false.

However, I looked through the docs again and noticed the following:

```
         3) "M NOT LIKE P" is equivalent to "NOT (M LIKE P)".
```

in which case `WHERE col IS *NOT LIKE* '%'` should return the inverse result set of `WHERE col IS *LIKE* '%'` (and should compile to `WHERE col IS NULL`).

I might have misinterpreted something but the specification seems to contradict itself.

kliewkliew: Though it is probably more useful to compile to `WHERE col IS NULL`; I don't see a use case for ` select * from table where false`. I'll change it to compile to `WHERE col IS NULL`

phoenix-core/src/main/java/org/apache/phoenix/compile/ExpressionCompiler.java
```diff
@@ -523,7 +523,12 @@ public Expression visitLeave(LikeParseNode node, List<Expression> children) thro
                 byte[] wildcard = {StringUtil.MULTI_CHAR_LIKE};
                 StringUtil.fill(nullExpressionString, 0, pattern.length(), wildcard, 0, 1, false);
                 if (pattern.equals(new String (nullExpressionString))) {
-                    return IsNullExpression.create(lhs, true, context.getTempPtr());
```
JamesRTaylor: I think the original code, optimizing to COL IS NOT NULL is not correct, because if COL is null, it'll evaluate to false when it should evaluate to null. Instead, I think it should become COL >= CAST(KeyRange.IS_NOT_NULL_RANGE AS VARCHAR) and for the negate case, it'd be COL < CAST(KeyRange.IS_NOT_NULL_RANGE AS VARCHAR). That way in any case, if COL is null, the expression will evaluate to null. If COL is any string then the first expression will evaluate to true and the latter to false.

The code will look something like this:
    // Declare this as a static constant at top of file
    private static final Expression NOT_NULL_STRING = LiteralExpression.newConstant(PVarchar.INSTANCE.toObject(KeyRange.IS_NOT_NULL_RANGE.getLowerRange()));
    List<Expression> compareChildren = Arrays.asList(lhs, NOT_NULL_STRING);
    return new ComparisonExpression.create(compareChildren, node.isNegate() ? CompareOp.LESS : CompareOp.GREATER_OR_EQUAL);

Then make sure we have tests around COL being null too.

kliewkliew: I made the suggested changes and added test coverage to confirm that `NULL like '%'` returns NULL. JDBC `getBoolean` transforms NULL to false but they provide a `wasNull` method that I used to verify the test cases.

JamesRTaylor:
+1 to the patch. Nice work, @kliewkliew. Will get this committed shortly.

PR #209, opened 2016-09-26
add configuration  function.

danieldingzju:
as mentioned in https://phoenix.apache.org/tuning.html.

I could not set the configuration of phoenix, so I'v tired to add the code to solve this problem.

then I can just set the phoenix like this:

image


joshelser:
> I could not set the configuration of phoenix, so I'v tired to add the code to solve this problem.

What were your steps to set these configuration properties before this patch? If you set them in the hbase-site.xml that was included on the classpath, they should have taken effect. If this is not happening, it's a bug.

PR #210, opened 2016-09-27
PHOENIX-2890 Extend IndexTool to allow incremental index rebuilds

ankitsinghal:


phoenix-core/src/it/java/org/apache/phoenix/end2end/IndexToolForPartialBuildIT.java
```diff
@@ -0,0 +1,264 @@
...
+            final String[] cmdArgs = getArgValues(schemaName, dataTableName);
+            int status = indexingTool.run(cmdArgs);
+            assertEquals(0, status);
+
+            // insert two more rows
+            upsertRow(stmt1, 8000);
+            upsertRow(stmt1, 9000);
+            conn.commit();
+
+            rs = stmt1.executeQuery("SELECT LPAD(UPPER(NAME),11,'x')||'_xyz',ID FROM " + fullTableName);
```
chrajeshbabu: This is not needed. We are not using the result for anything.

phoenix-core/src/it/java/org/apache/phoenix/end2end/IndexToolForPartialBuildIT.java
```diff
@@ -0,0 +1,264 @@
...
+            PTable pindexTable = PhoenixRuntime.getTable(conn, SchemaUtil.getTableName(schemaName, indxTable));
+            assertEquals(PIndexState.BUILDING, pindexTable.getIndexState());
+            assertEquals(rs.getLong(1), pindexTable.getTimeStamp());
+
+            String selectSql = String.format("SELECT LPAD(UPPER(NAME),11,'x')||'_xyz',ID FROM %s", fullTableName);
+            rs = conn.createStatement().executeQuery("EXPLAIN " + selectSql);
+            String actualExplainPlan = QueryUtil.getExplainPlan(rs);
+
+            // assert we are pulling from data table.
+            assertTrue(actualExplainPlan.contains(
```
chrajeshbabu: Use assertExplainPlan method to check the plan.

phoenix-core/src/it/java/org/apache/phoenix/end2end/IndexToolForPartialBuildWithNamespaceEnabled.java
```diff
@@ -0,0 +1,71 @@
...
+        serverProps.put(QueryServices.IS_NAMESPACE_MAPPING_ENABLED, "true");
+        Map<String, String> clientProps = Maps.newHashMapWithExpectedSize(1);
+        clientProps.put(QueryServices.IS_NAMESPACE_MAPPING_ENABLED, "true");
+        setUpTestDriver(new ReadOnlyProps(serverProps.entrySet().iterator()), new ReadOnlyProps(clientProps.entrySet().iterator()));
+    }
+    
+    @Parameters(name="localIndex = {0} , isNamespaceEnabled = {1}")
+    public static Collection<Boolean[]> data() {
+        return Arrays.asList(new Boolean[][] {     
+                 { false, true},{ true, false }
```
chrajeshbabu: We can make IndexToolForPartialBuildIT itself to run with namespace enabled and disabled. We don't need extra IT test.

phoenix-core/src/it/java/org/apache/phoenix/end2end/index/IndexMetadataIT.java
```diff
@@ -568,4 +580,52 @@ public void testAsyncCreatedDate() throws Exception {
         assertTrue(d2.after(d1));
         assertFalse(rs.next());
     }
+    
+    @Test
+    public void testAsyncRebuildTimestamp() throws Exception {
+        long l0 = System.currentTimeMillis();
```
chrajeshbabu: can you use meaning full names for variables l0 and l1.

phoenix-core/src/it/java/org/apache/phoenix/end2end/index/MutableIndexFailureIT.java
```diff
@@ -385,6 +383,8 @@ private void updateTable(Connection conn, String tableName) throws SQLException
     }
 
     public static class FailingRegionObserver extends SimpleRegionObserver {
+        public static volatile boolean FAIL_WRITE = false;
```
chrajeshbabu: Sharing these may make the tests flaky if both MutableIndexFailureIT and IndexToolForPartialBuildIT run parallel.

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/MetaDataRegionObserver.java
```diff
@@ -197,10 +203,15 @@ public void run() {
...
         // running
         private final static AtomicInteger inProgress = new AtomicInteger(0);
+        private static final String UPDATE_DISABLED_TS = "UPSERT INTO " + SYSTEM_CATALOG + "(" + TENANT_ID + ","
+                + TABLE_SCHEM + "," + TABLE_NAME + "," + INDEX_DISABLE_TIMESTAMP + ") values (?,?,?,?)";
         RegionCoprocessorEnvironment env;
+        private long rebuildIndexBatchSize = HConstants.LATEST_TIMESTAMP;
 
         public BuildIndexScheduleTask(RegionCoprocessorEnvironment env) {
             this.env = env;
+            this.rebuildIndexBatchSize = env.getConfiguration().getLong(
```
chrajeshbabu: The configuration is confusing it's similar to number of mutations but it's actually a timestamp. It can be something like INDEX_FAILURE_HANDLING_REBUILD_PERIOD?

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/index/IndexTool.java
```diff
@@ -85,7 +102,7 @@
     private static final Option DATA_TABLE_OPTION = new Option("dt", "data-table", true,
             "Data table name (mandatory)");
     private static final Option INDEX_TABLE_OPTION = new Option("it", "index-table", true,
-            "Index table name(mandatory)");
+            "Index table name(not required in case of partial rebuilding)");
```
chrajeshbabu: Better to have an argument for partial rebuilding than go by rebuild if we don't mention index table.

phoenix-core/src/main/java/org/apache/phoenix/schema/MetaDataClient.java
```diff
@@ -305,16 +305,27 @@
...
     private static final String UPDATE_INDEX_STATE_TO_ACTIVE =
             "UPSERT INTO " + SYSTEM_CATALOG_SCHEMA + ".\"" + SYSTEM_CATALOG_TABLE + "\"( " +
                     TENANT_ID + "," +
                     TABLE_SCHEM + "," +
                     TABLE_NAME + "," +
                     INDEX_STATE + "," +
-                    INDEX_DISABLE_TIMESTAMP +
-                    ") VALUES (?, ?, ?, ?, ?)";
+                    INDEX_DISABLE_TIMESTAMP +","+
+                    ASYNC_REBUILD_TIMESTAMP + " " + PLong.INSTANCE.getSqlTypeName() +
```
chrajeshbabu: Here ASYNC_REBUILD_TIMESTAMP is dynamic column. Are there any problems if we use dynamic columns for system tables? why can't you make normal column?

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/index/IndexTool.java
```diff
@@ -167,50 +180,152 @@ private void printHelpAndExit(Options options, int exitCode) {
...
+            }
+            ImmutableBytesWritable indexMetaDataPtr = new ImmutableBytesWritable(ByteUtil.EMPTY_BYTE_ARRAY);
+            IndexMaintainer.serializeAdditional(pdataTable, indexMetaDataPtr, disabledPIndexes, connection.unwrap(PhoenixConnection.class));
+            PhoenixConfigurationUtil.setIndexMaintainers(configuration, indexMetaDataPtr);
+            
+            //Prepare raw scan 
+            Scan scan = IndexManagementUtil.newLocalStateScan(maintainers);
+            scan.setTimeRange(minDisableTimestamp - 1, maxTimestamp);
+            scan.setRaw(true);
+            scan.setCacheBlocks(false);
```
ankitsinghal: As these blocks will not read again , we can just leverage OS cache for disk read rather than copying them in our JVM and then evict which may lead GC a lot. WDYT?
And, currently ExportTSV also doesn't bring blocks in cache by default.

phoenix-core/src/it/java/org/apache/phoenix/end2end/IndexToolForPartialBuildWithNamespaceEnabled.java
ankitsinghal: As discussed with you, it's difficult to do .. As mini cluster can be shared across test and in parallel environment restarting a cluster may fail other tests.

phoenix-core/src/main/java/org/apache/phoenix/schema/MetaDataClient.java
chrajeshbabu: Ok. It's fine. Thanks Ankit

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/index/IndexTool.java
chrajeshbabu: @ankitsinghal  The rebuilding also a kind of scan from other clients so HBase should cope up with the caches and GCs. But setting it to true makes the job run faster. We can check the performance difference and decide on this.

phoenix-core/src/it/java/org/apache/phoenix/end2end/IndexToolForPartialBuildWithNamespaceEnabled.java
chrajeshbabu: Agree with you.

phoenix-core/src/it/java/org/apache/phoenix/end2end/IndexToolForPartialBuildIT.java
```diff
@@ -217,39 +223,47 @@ public void testSecondaryIndex() throws Exception {
 
             assertFalse(rs.next());
 
-            conn.createStatement().execute(String.format("DROP INDEX  %s ON %s", indxTable, fullTableName));
+           // conn.createStatement().execute(String.format("DROP INDEX  %s ON %s", indxTable, fullTableName));
```
chrajeshbabu: We can remove this commented code.

phoenix-core/src/it/java/org/apache/phoenix/end2end/index/IndexMetadataIT.java
```diff
@@ -216,6 +219,15 @@ public void testIndexCreateDrop() throws Exception {
...
             
             assertActiveIndex(conn, INDEX_DATA_SCHEMA, indexDataTable);
+            
+            ddl = "ALTER INDEX " + indexName + " ON " + INDEX_DATA_SCHEMA + QueryConstants.NAME_SEPARATOR + indexDataTable + " REBUILD ASYNC";
+            conn.createStatement().execute(ddl);
+            // Verify the metadata for index is correct.
+            rs = conn.getMetaData().getTables(null, StringUtil.escapeLike(INDEX_DATA_SCHEMA), indexName , new String[] {PTableType.INDEX.toString()});
+            assertTrue(rs.next());
+            assertEquals(indexName , rs.getString(3));
+            assertEquals(PIndexState.BUILDING.toString(), rs.getString("INDEX_STATE"));
```
ankitsinghal: This was taken care in the last commit.

phoenix-core/src/it/java/org/apache/phoenix/end2end/index/MutableIndexFailureIT.java
ankitsinghal: This was taken care in the last commit.

phoenix-core/src/it/java/org/apache/phoenix/end2end/OrderByIT.java
```diff
@@ -500,14 +500,12 @@ public void testOrderByWithExpression() throws Exception {
             stmt.execute();
             conn.commit();
 
-            String query = "SELECT col1+col2, col4, a_string FROM " + tableName + " ORDER BY 1, 2";
```
ankitsinghal: yes @chrajeshbabu , change was rollbacked in the last commit

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/index/IndexTool.java
ankitsinghal: This was taken care in the last commit.

ankitsinghal:
Thank you so much @chrajeshbabu for the review, Raised PHOENIX-3525 for the same.
@JamesRTaylor  ,do you want to take a look or it's ok to commit now?

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/index/IndexTool.java
ankitsinghal: Yes @chrajeshbabu , I did a below test and found the performance of both the scans(with or without cache) are same, file descriptor count during the test is almost same.

// with cache on
- create a new table
- Insert 500k records
- Flush the data
- scan with cache on

// with cache off
- create a new table again
- Insert 500k records
- Flush the data
- scan with cache off

Do you want me to do some other tests to verify the GC impact and all? or is it fine for now? as the cache is also kept off in org.apache.hadoop.hbase.mapreduce.Export by default.

stoty:
Already merged.

PR #211, opened 2016-09-27
PHOENIX-3254 IndexId Sequence is incremented even if index exists alr…

ankitsinghal:
…eady

### DIFF
--- a/phoenix-core/src/main/java/org/apache/phoenix/coprocessor/MetaDataEndpointImpl.java
+++ b/phoenix-core/src/main/java/org/apache/phoenix/coprocessor/MetaDataEndpointImpl.java
@@ -195,6 +195,9 @@ import org.apache.phoenix.schema.PTable.LinkType;
 import org.apache.phoenix.schema.PTable.ViewType;
 import org.apache.phoenix.schema.PTableImpl;
 import org.apache.phoenix.schema.PTableType;
+import org.apache.phoenix.schema.SequenceAllocation;
+import org.apache.phoenix.schema.SequenceAlreadyExistsException;
+import org.apache.phoenix.schema.SequenceKey;
 import org.apache.phoenix.schema.SequenceNotFoundException;
 import org.apache.phoenix.schema.SortOrder;
 import org.apache.phoenix.schema.TableNotFoundException;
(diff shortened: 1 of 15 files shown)


phoenix-core/src/main/java/org/apache/phoenix/coprocessor/MetaDataEndpointImpl.java
```diff
@@ -1499,6 +1502,53 @@ public void createTable(RpcController controller, CreateTableRequest request,
...
+                    PName physicalName = parentTable.getPhysicalName();
+                    int nSequenceSaltBuckets = connection.getQueryServices().getSequenceSaltBuckets();
+                    SequenceKey key = MetaDataUtil.getViewIndexSequenceKey(tenantIdStr, physicalName,
+                            nSequenceSaltBuckets, parentTable.isNamespaceMapped() );
+                        // TODO Review Earlier sequence was created at (SCN-1/LATEST_TIMESTAMP) and incremented at the client max(SCN,dataTable.getTimestamp), but it seems we should
+                        // use always LATEST_TIMESTAMP to avoid seeing wrong sequence values by different connection having SCN
+                        // or not. 
+                    long sequenceTimestamp = HConstants.LATEST_TIMESTAMP;
+                    try {
+                        connection.getQueryServices().createSequence(key.getTenantId(), key.getSchemaName(), key.getSequenceName(),
```
chrajeshbabu: @ankitsinghal  This patch introducing inter table rpc calls but seems it's needed. What do you think of moving create sequence to meta data client and perform sequence increment alone here? So that atleast we can reduce the rpc calls here.

phoenix-core/src/main/java/org/apache/phoenix/query/ConnectionQueryServicesImpl.java
```diff
@@ -1445,6 +1445,9 @@ public MetaDataResponse call(MetaDataService instance) throws IOException {
                     builder.addTableMetadataMutations(mp.toByteString());
                 }
                 builder.setClientVersion(VersionUtil.encodeVersion(PHOENIX_MAJOR_VERSION, PHOENIX_MINOR_VERSION, PHOENIX_PATCH_NUMBER));
+                        if (allocateIndexId) {
```
chrajeshbabu: nit: just format code here.

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/MetaDataEndpointImpl.java
JamesRTaylor: The RPCs are minimal here because of the way sequences are implemented. By default, we cache 100 sequences and dole them out as needed. So there's only a single RPC per 100 invocations of this.

JamesRTaylor:
+1 on the changes (but please fix any formatting issues). Nice work, @ankitsinghal.

stoty:
Already merged.

PR #212, opened 2016-09-29
PHOENIX-3264 Allow TRUE and FALSE to be used as literal constants

lomoree:


phoenix-core/src/main/java/org/apache/phoenix/calcite/CalciteUtils.java
```diff
@@ -1036,4 +1048,40 @@ public Void visitCall(RexCall call) {
...
         }
     }
+
+    public static Object convertLiteral(SqlLiteral literal, PhoenixRelImplementor implementor) {
+        try {
+            final SchemaPlus rootSchema = Frameworks.createRootSchema(true);
+            final FrameworkConfig config = Frameworks.newConfigBuilder()
+                    .parserConfig(SqlParser.Config.DEFAULT)
+                    .defaultSchema(rootSchema).build();
+            Planner planner = Frameworks.getPlanner(config);
```
maryannxue: Since this method will be called once for each SQLLiteral, it might be better make these objects static members of CalciteUtils if possible.

phoenix-core/src/main/java/org/apache/phoenix/calcite/CalciteUtils.java
```diff
@@ -1036,4 +1048,40 @@ public Void visitCall(RexCall call) {
             return null;
         }
     }
+
+    public static Object convertLiteral(SqlLiteral literal, PhoenixRelImplementor implementor) {
```
maryannxue: Other convertXXX methods are to convert RexNode to Expression, so I'm thinking something like "convertSqlLiteral(...)" could be less confusing?

phoenix-core/src/main/java/org/apache/phoenix/calcite/CalciteUtils.java
```diff
@@ -1036,4 +1048,40 @@ public Void visitCall(RexCall call) {
...
+            Planner planner = Frameworks.getPlanner(config);
+
+            SqlParserPos POS = SqlParserPos.ZERO;
+            final SqlNodeList selectList =
+                    new SqlNodeList(
+                            Collections.singletonList(literal),
+                            SqlParserPos.ZERO);
+
+
+            String sql = new SqlSelect(POS, SqlNodeList.EMPTY, selectList, null, null, null, null,
```
maryannxue: Can we skip making a sql string and parsing again, coz you already have a SqlSelect node? We can go directly to planner.validate() and planner.rel(), right?

phoenix-core/src/main/java/org/apache/phoenix/calcite/CalciteUtils.java
```diff
@@ -1036,4 +1048,40 @@ public Void visitCall(RexCall call) {
...
+                    SqlNodeList.EMPTY, null, null, null).toString();
+            SqlNode sqlNode = planner.parse(sql);
+            sqlNode = planner.validate(sqlNode);
+            Project proj = (Project) (planner.rel(sqlNode).rel);
```
maryannxue: Probably better to add an "assert" here to ensure the rel object is of Project class before casting it. Similarly, assert that this Project has only one expression for the following statement.

phoenix-core/src/main/java/org/apache/phoenix/calcite/CalciteUtils.java
```diff
@@ -1036,4 +1048,40 @@ public Void visitCall(RexCall call) {
...
+            RexNode rex = proj.getChildExps().get(0);
+
+            Expression e = CalciteUtils.toExpression(rex, implementor);
+            ImmutableBytesWritable ptr = new ImmutableBytesWritable();
+            e = ExpressionUtil.getConstantExpression(e, ptr);
+            Object ret = e.getDataType().toObject(ptr);
+            if(ret instanceof NlsString){
```
maryannxue: Is this step necessary? I suppose no. But if yes, CalciteUtils.toExpression() must have not done its job correctly.

phoenix-core/src/main/java/org/apache/phoenix/calcite/CalciteUtils.java
```diff
@@ -1036,4 +1048,40 @@ public Void visitCall(RexCall call) {
...
+                ret = ((NlsString) ret).toString();
+            }
+            return ret;
+        }
+        catch (Exception e){
+            throw new RuntimeException("Could not convert literal to its object type: " + e);
```
maryannxue: Normally you don't append the Exception itself to the end of the error message. You message should include the exact SqlLiteral it failed to convert and add the original Exception as inner exception.

phoenix-core/src/main/java/org/apache/calcite/sql/SqlOptionNode.java
```diff
@@ -38,12 +41,10 @@ public SqlOptionNode(SqlParserPos pos, SqlIdentifier key, SqlLiteral literal) {
...
         }
-        final Object v = SqlLiteral.value(literal);
-        if (v instanceof NlsString) {
-            value = ((NlsString) v).toString();
-        } else {
-            value = v;
-        }
+
+        PhoenixRelImplementor
+                implementor = new PhoenixRelImplementorImpl(new RuntimeContextImpl());
```
maryannxue: Looks like we need a "dummy" RuntimeContext for this kind of "dry run" implementor. I'll do this part when I commit it.

phoenix-core/src/main/java/org/apache/phoenix/calcite/CalciteUtils.java
lomoree: I tried desperately to do this, but Calcite has pretty strict enforcement on Sql States. When the node is created its in STATE 1: READY, and it will not allow you to go straight to a validation state. Is there some way of overriding this?

phoenix-core/src/main/java/org/apache/phoenix/calcite/CalciteUtils.java
lomoree: Agreed, will update.

phoenix-core/src/main/java/org/apache/phoenix/calcite/CalciteUtils.java
maryannxue: Ok, I once ran into this as well. Let's just do string then.

phoenix-core/src/main/java/org/apache/phoenix/calcite/CalciteUtils.java
lomoree: I could also abstract this to the SqlOptionNode and pass it in to this method. I had that at one point but ended up opting for this since the code didn't seem like it belonged there. Let me know which you prefer.

phoenix-core/src/main/java/org/apache/phoenix/calcite/CalciteUtils.java
lomoree: Will add both of these assertions, good catch. Thanks

phoenix-core/src/main/java/org/apache/phoenix/calcite/CalciteUtils.java
maryannxue: Maybe let's leave it here in CalciteUtils if that works. They don't necessarily belong anywhere, and CalciteUtils seems to be a class best for hiding all the details.

phoenix-core/src/main/java/org/apache/phoenix/calcite/CalciteUtils.java
lomoree: I'm not sure if it's necessary, but decided to use it per James' recommendation. He might have some insight on why it's needed.

phoenix-core/src/main/java/org/apache/phoenix/calcite/CalciteUtils.java
lomoree: Will amend this, thanks!

phoenix-core/src/main/java/org/apache/phoenix/calcite/CalciteUtils.java
maryannxue: I was talking about the "if" block, the NlsString is a Calcite thing, so once we have converted a RexNode into an Phoenix Expression, there should only be Java String objects. So you should probably just remove it and verify with a test case (maybe already covered by an existing test case).

phoenix-core/src/main/java/org/apache/phoenix/calcite/CalciteUtils.java
maryannxue: Thank you, Eric, for doing a nice job and trying to work things out in the best you could find! But if you run into issues like this next time, you can just ping me quickly. I can probably give you some useful information since I've had quite some experience dealing with Calcite.

phoenix-core/src/main/java/org/apache/phoenix/calcite/CalciteUtils.java
lomoree: Yes I meant that as well. I think it's because we had originally intended to use Calcite utilities to do this. I just ran through a debugger and all inputs parse down to either a Char or a String. It is definitely covered by test cases.

PR #213, opened 2016-09-30
PHOENIX-2827 Support OFFSET in Calcite-Phoenix

lomoree:


phoenix-core/src/main/java/org/apache/phoenix/calcite/rel/PhoenixLimit.java
```diff
@@ -76,20 +76,27 @@ public RelOptCost computeSelfCost(RelOptPlanner planner, RelMetadataQuery mq) {
     
     @Override 
     public double estimateRowCount(RelMetadataQuery mq) {
-        double rows = super.estimateRowCount(mq);        
+        double rows = super.estimateRowCount(mq);
+        if(offset != null) {
+            return Math.max(0, Math.min(RexLiteral.intValue(fetch), rows - RexLiteral.intValue(offset)));
+        }
         return Math.min(RexLiteral.intValue(fetch), rows);
```
maryannxue: I'd write:
int offset = this.offset == null ? 0 : RexLiteral.intValue(this.offset);
return Math.min(RexLiteral.intValue(fetch), rows - offset);

phoenix-core/src/main/java/org/apache/phoenix/calcite/rel/PhoenixLimit.java
```diff
@@ -76,20 +76,27 @@ public RelOptCost computeSelfCost(RelOptPlanner planner, RelMetadataQuery mq) {
...
 
     @Override
     public QueryPlan implement(PhoenixRelImplementor implementor) {
         QueryPlan plan = implementor.visitInput(0, (PhoenixQueryRel) getInput());
         int fetchValue = RexLiteral.intValue(fetch);
+        int offsetValue = 0;
+        if (offset != null){
+            offsetValue = RexLiteral.intValue(offset);
+        }
         if (plan.getLimit() == null) {
```
maryannxue: should be "if (plan.getLimit() == null && plan.getOffset() == null)"

phoenix-core/src/main/java/org/apache/phoenix/calcite/rel/PhoenixLimit.java
```diff
@@ -76,20 +76,27 @@ public RelOptCost computeSelfCost(RelOptPlanner planner, RelMetadataQuery mq) {
...
+        if(offset != null) {
+            return Math.max(0, Math.min(RexLiteral.intValue(fetch), rows - RexLiteral.intValue(offset)));
+        }
         return Math.min(RexLiteral.intValue(fetch), rows);
     }
 
     @Override
     public QueryPlan implement(PhoenixRelImplementor implementor) {
         QueryPlan plan = implementor.visitInput(0, (PhoenixQueryRel) getInput());
         int fetchValue = RexLiteral.intValue(fetch);
```
maryannxue: I believe now that we've enabled offset, "this.fetch" could be also be null, which means it should be handled the same way as this.offset now. Could you please also add test cases to cover situations like "offset != null" but "limit == null"?

phoenix-core/src/main/java/org/apache/phoenix/calcite/rel/PhoenixLimit.java
```diff
@@ -76,20 +76,27 @@ public RelOptCost computeSelfCost(RelOptPlanner planner, RelMetadataQuery mq) {
...
+        }
         if (plan.getLimit() == null) {
-            return plan.limit(fetchValue);
+            return plan.limit(fetchValue, offsetValue);
         }
         
         return new ClientScanPlan(plan.getContext(), plan.getStatement(), 
                 implementor.getTableMapping().getTableRef(), RowProjector.EMPTY_PROJECTOR, 
-                fetchValue, null, null, OrderBy.EMPTY_ORDER_BY, plan);
+                fetchValue, offsetValue, null, OrderBy.EMPTY_ORDER_BY, plan);
```
maryannxue: Since either of them could be null, Boolean objects should be used instead of primitive boolean type.

phoenix-core/src/main/java/org/apache/phoenix/calcite/rules/PhoenixConverterRules.java
```diff
@@ -232,8 +232,7 @@ public RelNode convert(RelNode rel) {
         private static Predicate<LogicalSort> HAS_FETCH = new Predicate<LogicalSort>() {
             @Override
             public boolean apply(LogicalSort input) {
-                return input.offset == null 
-                        && input.fetch != null;
+                return input.fetch != null;
```
maryannxue: should be "input.fetch != null || input.offset != null"

phoenix-core/src/main/java/org/apache/phoenix/execute/ClientScanPlan.java
```diff
@@ -106,12 +106,18 @@ public ExplainPlan getExplainPlan() throws SQLException {
     }
 
     @Override
-    public QueryPlan limit(Integer limit) {
-        if (limit == this.limit || (limit != null && limit.equals(this.limit)))
+    public QueryPlan limit(Integer limit, Integer offset) {
+        if (limit == this.limit || (limit != null && limit.equals(this.limit))) {
```
maryannxue: I can see there's a big confusion here. Basically, offset and limit should be treated the same way. So here, it should be "if both the old limit and the old offset are same as as the new ones specified, we return this; otherwise we create a new object."

lomoree: This does definitely have ripple effects throughout, but I had done this on the basis that SQL only allows an offset if their is an imposed limit (since an offset doesn't make any logical sense without a limit).

phoenix-core/src/main/java/org/apache/phoenix/calcite/rules/PhoenixConverterRules.java
lomoree: With respect to my comment below, if there is a limit, then we apply the rule, otherwise we do not (regardless of offset).
 I may have the wrong idea here, but based on my understanding of offset, all of the test cases pass as expected. Are the test cases I have laid out in CalciteIT what you would expect?

phoenix-core/src/main/java/org/apache/phoenix/execute/ClientScanPlan.java
maryannxue: Is it a SQL-92 standard that OFFSET can only be used together with LIMIT? I know there are some dialects that allow different OFFSET/LIMIT or OFFSET/FETCH grammars, so I'm just confused... I tried with Calcite parser and it does allow OFFSET without LIMIT.

JamesRTaylor: For right or wrong, Phoenix allows an OFFSET to be used without a LIMIT today.

lomoree: Ah, my mistake. I'll get working on a patch that also supports offset without limit. This should incorporate all of the changes noted above. Thanks for clarifying.

PR #214, opened 2016-10-03
PHOENIX-3298 Single column primary key may not be null

lomoree:
Note: This patch will not immediately resolve the testInvalidPrimaryKeyDecl() test case due to another issue, PHOENIX-3345. However, they are independent issues so there's no reason to wait on this.

### DIFF
--- a/phoenix-core/src/test/java/org/apache/phoenix/compile/QueryCompilerTest.java
+++ b/phoenix-core/src/test/java/org/apache/phoenix/compile/QueryCompilerTest.java
@@ -1453,7 +1453,7 @@ public class QueryCompilerTest extends BaseConnectionlessQueryTest {
             }
         }
     }
-    
+
     @Test
     public void testInvalidNullCompositePrimaryKey() throws Exception {
         Connection conn = DriverManager.getConnection(getUrl());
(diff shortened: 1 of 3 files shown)


JamesRTaylor:
+1 to the fix. @chrajeshbabu or @maryannxue - would you mind checking this in?

maryannxue:
Sure. I'll do that.

phoenix-core/src/test/java/org/apache/phoenix/compile/QueryCompilerTest.java
```diff
@@ -1453,7 +1453,7 @@ public void testInvalidPrimaryKeyDecl() throws Exception {
             }
         }
     }
-    
+
```
maryannxue: @lomoree It's nice to avoid these trailing spaces, but most of the time, we need to avoid such changes in our patch, especially for 'calcite' branch at the moment, for it would make future merge to master more difficult.

lomoree: Of course, I will avoid fixing these in the future. Thanks!

PR #215, opened 2016-10-06
PHOENIX-3345 SQLException code's not propagating as expected

lomoree:
-- Primarily impacts unit tests in QueryCompilerTest.java

### DIFF
--- a/phoenix-core/src/main/java/org/apache/calcite/jdbc/PhoenixCalciteFactory.java
+++ b/phoenix-core/src/main/java/org/apache/calcite/jdbc/PhoenixCalciteFactory.java
@@ -44,6 +44,7 @@ import org.apache.calcite.schema.SchemaPlus;
 import org.apache.calcite.tools.Program;
 import org.apache.calcite.tools.Programs;
 import org.apache.calcite.util.Holder;
+import org.apache.phoenix.calcite.CalciteUtils;
 import org.apache.phoenix.calcite.PhoenixSchema;
 import org.apache.phoenix.calcite.rel.PhoenixRel;
 import org.apache.phoenix.exception.SQLExceptionCode;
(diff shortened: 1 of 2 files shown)


phoenix-core/src/main/java/org/apache/calcite/jdbc/PhoenixCalciteFactory.java
```diff
@@ -161,6 +162,29 @@ public Object apply(
             }));
         }
 
+        public CalciteStatement createStatement(String sql, int resultSetType, int resultSetConcurrency, int resultSetHoldability) throws SQLException {
+            try {
+                return super.createStatement(resultSetType, resultSetConcurrency, resultSetHoldability);
+            } catch (SQLException e) {
+                if (e.getCause().getCause() instanceof SQLException) {
+                    throw (SQLException) e.getCause().getCause();
```
maryannxue: Can we can have a utility method that unwraps the Exception? And as we might different number of levels of Exception wrapping, can we just look down one level at a time till we hit an instance of SQLException?

lomoree: I wondered about this myself. While I didn't encounter any cases where exception levels varied, it's a good idea to account for it. Will make these changes, thanks @maryannxue

lomoree: The new changes go for the deepest SQL exception. Can quickly change it if you would like only the next most deep SQL exception.

phoenix-core/src/main/java/org/apache/phoenix/calcite/CalciteUtils.java
```diff
@@ -1120,8 +1120,11 @@ public static SQLException unwrapSqlException(SQLException root){
         Exception e = root;
         while(e.getCause() != null){
             e = (Exception) e.getCause();
+            if(e instanceof RuntimeException && e.getCause() instanceof SQLException) {
```
lomoree: Latest changes attempt to locate the nested SQLException within an RuntimeException. We still throw the first SQLException seen in case there's no wrapping argument exception. And if no matches are found, we throw the root so as to avoid unexpected behavior.

JamesRTaylor: FYI, there's already a Phoenix util that does something similar in ServerUtil.

PR #220, opened 2016-11-03
PHOENIX-3394 Handle SequenceResolving through ConnectionQueryServices interface

lomoree:


phoenix-core/src/main/java/org/apache/phoenix/calcite/PhoenixSchema.java
```diff
@@ -329,24 +333,14 @@ private TableRef fixTableMultiTenancy(TableRef tableRef) throws SQLException {
...
-                    + " and " + PhoenixDatabaseMetaData.TENANT_ID
-                    + (tenantId == null ? " is null" : " = '" + tenantId + "'");
-            ResultSet rs = pc.createStatement().executeQuery(q);
-            if (rs.next()) {
-                return new PhoenixSequence(schemaName, name, pc);
-            }
-        } catch (SQLException e) {
-            throw new RuntimeException(e);
+            SequenceManager manager = new SequenceManager((PhoenixStatement)pc.createStatement());
+            manager.newSequenceReference(pc.getTenantId(), TableName.createNormalized(schemaName, name) , null, SequenceValueParseNode.Op.NEXT_VALUE);
```
maryannxue: @JamesRTaylor Could you also take a look here to make sure that the method is called with right parameters? And is there a chance (and a need) we can avoid creating a new "Statement" each time?

phoenix-core/src/it/java/org/apache/phoenix/calcite/CalciteIT.java
```diff
@@ -1315,6 +1315,17 @@ public void initTable() throws Exception {
                         {6L, "00A323122312312"},
                         {8L, "00A423122312312"}})
                 .close();
+
```
maryannxue: It's nice to add more test case in CalciteIT, but I'm just curious, why aren't existing tests in testSequence() sufficient?

lomoree: Calcite seems to function a little bit differently with respect to CURRENT VALUE, so I thought it would be wise to include a case for it. Plus it's the only case in CalciteIT covering CURRENT VALUE.
See: https://issues.apache.org/jira/browse/PHOENIX-3437

phoenix-core/src/main/java/org/apache/phoenix/calcite/PhoenixSchema.java
lomoree: Good question. From my testing, passing in null doesn't work. A static member is an option if nothing else.

PR #221, opened 2016-11-07
PHOENIX-3427 fix saveToRdd for tenant-specific connections

nico-pappagianis:
@JamesRTaylor @jmahonin 

Here is the fix for writing to tenant-specific objects via RDDs.

I added an optional tenantId argument to saveToPhoenix. Let me know if there's another way we'd like to approach this (setting it in the conf, perhaps).

Thanks again for all of the help!


nico-pappagianis:
@jmahonin I've added a couple tests for DataFrames and refactored the test code a bit. Thanks for looking at this again.

phoenix-spark/src/main/scala/org/apache/phoenix/spark/ProductRDDFunctions.scala
```diff
@@ -16,19 +16,20 @@ package org.apache.phoenix.spark
...
 import org.apache.spark.Logging
 import org.apache.spark.rdd.RDD
+
 import scala.collection.JavaConversions._
 
 class ProductRDDFunctions[A <: Product](data: RDD[A]) extends Logging with Serializable {
 
   def saveToPhoenix(tableName: String, cols: Seq[String],
-                    conf: Configuration = new Configuration, zkUrl: Option[String] = None)
+                    conf: Configuration = new Configuration, zkUrl: Option[String] = None, tenantId: Option[String] = None)
```
nico-pappagianis: Sounds good

nico-pappagianis: Looking at this closer it appears the only way to pass the tenantId to PairRDDFunctions.saveAsNewAPIHadoopFile is through the Configuration parameter. 

I was able to avoid setting TenantId in the partitionConfig by overloading getUpsertColumnMetadataList to take in a Properties object. However PairRDDFunctions is part of spark-core, so I don't know if we want to make a change to spark-core so we can avoid setting the tenantId in the outputConfig. 

Let me know what you think.

JamesRTaylor: I'm not familiar enough with the lifecycle of the Configuration object in the Spark integration to know if this is an issue. If every save operation creates a new Configuration, then this is likely fine (assuming you handle the case where tenantID is not set). WDYT, @jmahonin?

JamesRTaylor: Thanks, @jmahonin. If we have to use Configuration object it's ok as long as it's not shared between other users (which might have a different TenantId or no TenantId). So if you can let us know that there's no sharing going on, that's fine too.

jmahonin: Ok taking a bit of a deeper dive, I think setting it directly on the Configuration object is about the best we can do here. Those objects are created on every save operation, so I don't think there's any possibility of bleeding across multiple users.

In a bit more detail, the limitation is that from within the phoenix-spark module, we don't have direct visibility or control over the Connection object. We essentially just tell Spark to use the `PhoenixOutputFormat` and to use `outConfig` as the Hadoop Configuration:

https://github.com/apache/phoenix/blob/master/phoenix-spark/src/main/scala/org/apache/phoenix/spark/ProductRDDFunctions.scala#L51-L56

Eventually, Spark will invoke `phoenix.mapreduce.PhoenixRecordWriter`, which will create the connection based on the configuration we've given it above:

https://github.com/apache/phoenix/blob/master/phoenix-spark/src/main/scala/org/apache/phoenix/spark/ProductRDDFunctions.scala#L51-L56

Following that call-path, the Connection object is created using a Properties object that's derived from the initial Configuration:

https://github.com/apache/phoenix/blob/master/phoenix-core/src/main/java/org/apache/phoenix/mapreduce/util/ConnectionUtil.java#L85

https://github.com/apache/phoenix/blob/master/phoenix-core/src/main/java/org/apache/phoenix/util/PropertiesUtil.java#L51-L60

In effect, the Configuration is really the *only* interface to configuring the Connection object at all.

JamesRTaylor: Ok, thanks for checking. I'm fine with this patch if you are, @jmahonin. We can potentially get this fix into 4.9.0 as we're getting ready to roll a new RC in the next couple of days.

JamesRTaylor: Yep, that'd be great. Thanks, @jmahonin and nice work, @nico-pappagianis!

nico-pappagianis: Awesome! Thanks for all the help [~jamestaylor] and [~jmahonin], I'll be submitting the latest code that I committed along with the suggestion James made about not using the URL string in the test setup function.

Thanks again!

jmahonin: Whoops, a bit sleep deprived, I pushed this up already. Just ping me when the new commit is up and I'll put in a follow-up.

jmahonin: Or even better, if you can submit a patchfile between your latest and master, that'd save me a few keystrokes. :)

nico-pappagianis: Sure, can you elaborate on the steps a bit. Its my first open source patch commit o.O

JamesRTaylor: Just do a git pull --rebase origin master on your local branch to pick up what Josh committed. Then push your next commit to your repro (either in the pull request or a new one).

Another way would be to just stash the changes for your last commit, get the latest by doing a git pull origin master, then apply your stashed change, and push out to this pull request.

nico-pappagianis:
Thanks James. I did the following:

from my local branch:

git pull --rebase origin master

git push origin multiTenantSaveToPhoenixRDDFix

jmahonin:
Thanks @nico-pappagianis I think I've got it squared away. Your changes are contained in the following commits:

https://github.com/apache/phoenix/commit/3886638b04a3e494be70e1e681ad56cb1c7e33c5
https://github.com/apache/phoenix/commit/3b07c2f1bad96210e9a81079aed1149f5fcb71bc

Please let me know if I missed anything. Note that I pruned some of the whitespace-only changes to unrelated files. The unit tests pass for me though. Thanks and well done!

nico-pappagianis:
Looks like everything's there.
Oh sorry about the white space - I must have hit IntelliJ auto format
shortcut.

Thanks again for all the help!

stoty:
Already merged.

PR #222, opened 2016-11-10
PHOENIX-3355 Register Phoenix built-in functions as Calcite functions

lomoree:
This patch works for nearly all built in functions. Aggregate functions are not yet supported.

Ultimately we may want to rework:
- The use of parse nodes as a factory
- Generic return types to explicit return types

These issues are hand in hand as abstract built in functions (NowFunction, FloorFunction) are where the parse node factory & generic return types are necessary.

Potential resolution from my point of view is 2 parts
- Create a link somewhere between an abstract function and its concrete implementations (maybe require inheritance, create an annotation that links the abstract function to its implementations, or include the logic in the below mentioned factory)
- Create a new factory that replaces the use of parse nodes

### DIFF
--- a/phoenix-core/src/main/java/org/apache/phoenix/parse/ParseNodeFactory.java
+++ b/phoenix-core/src/main/java/org/apache/phoenix/parse/ParseNodeFactory.java
@@ -129,6 +129,7 @@ public class ParseNodeFactory {
         }
         int nArgs = d.args().length;
         BuiltInFunctionInfo value = new BuiltInFunctionInfo(f, d);
+
         SINGLE_SIGNATURE_BUILT_IN_FUNCTION_MAP.put(value.getName(), value);
         do {
             // Add function to function map, throwing if conflicts found
--- a/phoenix-core/src/main/java/org/apache/phoenix/parse/ParseNodeFactory.java
+++ b/phoenix-core/src/main/java/org/apache/phoenix/parse/ParseNodeFactory.java
@@ -185,7 +185,7 @@ public class ParseNodeFactory {
         return info;
     }
 
-    public static Collection<BuiltInFunctionInfo> getAll(){
+    public static Collection<BuiltInFunctionInfo> getSingleEntryFunctionMap(){
         initBuiltInFunctionMap();
         return SINGLE_SIGNATURE_BUILT_IN_FUNCTION_MAP.values();
     }
--- a/phoenix-core/src/main/java/org/apache/phoenix/parse/ParseNodeFactory.java
+++ b/phoenix-core/src/main/java/org/apache/phoenix/parse/ParseNodeFactory.java
@@ -25,6 +25,7 @@ import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.Collection;
 import java.util.concurrent.atomic.AtomicInteger;
 
 import org.apache.hadoop.hbase.filter.CompareFilter.CompareOp;
@@ -73,6 +74,7 @@ public class ParseNodeFactory {
         AvgAggregateFunction.class
         );
     private static final Map<BuiltInFunctionKey, BuiltInFunctionInfo> BUILT_IN_FUNCTION_MAP = Maps.newHashMap();
+    private static final Map<String, BuiltInFunctionInfo> SINGLE_SIGNATURE_BUILT_IN_FUNCTION_MAP = Maps.newHashMap();
     private static final BigDecimal MAX_LONG = BigDecimal.valueOf(Long.MAX_VALUE);
 
 
@@ -127,6 +129,7 @@ public class ParseNodeFactory {
         }
         int nArgs = d.args().length;
         BuiltInFunctionInfo value = new BuiltInFunctionInfo(f, d);
+        SINGLE_SIGNATURE_BUILT_IN_FUNCTION_MAP.put(value.getName(), value);
         do {
             // Add function to function map, throwing if conflicts found
             // Add entry for each possible version of function based on arguments that are not required to be present (i.e. arg with default value)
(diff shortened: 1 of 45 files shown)


phoenix-core/src/main/java/org/apache/phoenix/parse/ParseNodeFactory.java
```diff
@@ -127,6 +129,7 @@ private static void addBuiltInFunction(Class<? extends FunctionExpression> f) th
         }
         int nArgs = d.args().length;
         BuiltInFunctionInfo value = new BuiltInFunctionInfo(f, d);
+        SINGLE_SIGNATURE_BUILT_IN_FUNCTION_MAP.put(value.getName(), value);
```
maryannxue: We'll have to adjust the map initialization logic here according to the new annotation. Avoid "ABSTRACT" for "SINGLE_SIGNATURE_BUILT_IN_FUNCTION_MAP" and "DERIVED" for "BUILT_IN_FUNCTION_MAP".

phoenix-core/src/main/java/org/apache/phoenix/parse/ParseNodeFactory.java
```diff
@@ -181,6 +184,11 @@ public static BuiltInFunctionInfo get(String normalizedName, List<ParseNode> chi
         return info;
     }
 
+    public static Collection<BuiltInFunctionInfo> getAll(){
```
maryannxue: Give it a better name please :)

lomoree:
1. There are still a number of cases that don't pass. Many of them are due to existing features not being supported, and many are due to other issues within Calcite-Phoenix. I have compiled a list of the major common issues though:
- No differentiation between the regex functions (String based and Byte based)
- Invert and MD5 Function don't work (phoenix doesn't supply the correct arguments in annotation)
- Type matching doesn't work on input arguments (date/timestamp)
- Timezone parameter doesn't work properly in ToDateFunction (basic cases work, but custom timezones do not act as expected)
- Array constructors are not supported (another issue entirely)

2. I believe you're referring to those in PhoenixScalarFunction, createBuiltinFunctions() and evaluateReturnType()? In the case of createBuiltinFunctions() it makes sense to be in PhoenixScalarFunction since it create instances of that class. We could move evaluateReturnType() to PhoenixSchema since it has more generic behavior, but I'm not entirely convinced that it should move.

maryannxue:
No differentiation between the regex functions (String based and Byte based)
-- Let's just add support for this one?
Invert and MD5 Function don't work (phoenix doesn't supply the correct arguments in annotation)
-- Can we make it right, or is it more complicated than that? If the latter, we can open another issue.
Type matching doesn't work on input arguments (date/timestamp)
-- It's probably a Calcite issue. Let's try repo it as a Calcite test case. Let me know if you need help working with Calcite unit tests.
Timezone parameter doesn't work properly in ToDateFunction (basic cases work, but custom timezones do not act as expected)
I probably know this one. Not sure if this is the issue though: the date/time/timestamp objects are local time (w/ timezone offset) in Calcite/Avatica, while they are UTC time on the Phoenix side.
Array constructors are not supported (another issue entirely)
-- Yes, another issue. We could open a JIRA listing those test failures blocking on this issue as a way to keep track of them.

As to the code refinement, I think there are 4 (maybe more?) big chunks of code: createBuiltinFunctions(), evaluateReturnType(), overloadArguments() and registerBuiltinFunctions(). We should probably put all these in one place. I prefer in PhoenixSchema, otherwise CalciteUtils could also do. What makes sense to where the code should live, aside from what class of object it instantiates, is also who is the caller or who decides the logic or routines. So it seems to me that PhoenixScalarFunction here is only the "callee" class which is used by PhoenixSchema and CalciteUtils.
There are a couple of things here that eventually forms the complete list of functions that we register in Calcite:
1. We create multiple signatures corresponding to each overload of the same function.
2. We merge all versions of the same signature but with different default parameter counts into one function.
3. We deal with aliases.
Part of the reason why we are doing this is that we are trying to reorganize the existing function dictionary into something that would match the Calcite conventions. The implementation could end up being very different if the existing Phoenix function structure were something else. So I guess it'd be easier to maintain if we put these methods all in one place and maybe we should add more code comment describing what each of these methods do.

lomoree:
1. No differentiation between the regex functions (String based and Byte based)
Yep I'll try some approaches to make this work.

2. Invert and MD5 Function don't work (phoenix doesn't supply the correct arguments in annotation)
Might be complicated. Phoenix uses a generic "term" as the input argument for these functions.
See https://phoenix.apache.org/language/index.html#term
We don't really have a type that encapsulates this kind of behavior as far as I know. 
This type of argument is represented by the following annotation:
@BuiltInFunction(name = InvertFunction.NAME, args = { @Argument() })

3. Timezone parameter doesn't work properly in ToDateFunction (basic cases work, but custom timezones do not act as expected)
I believe that's correct. When I was doing testing all of the results are inexplicably off by an offset that is exactly the difference between UTC and and EST :)

4. Type matching doesn't work on input arguments (date/timestamp)
I explored this after our meeting. This used to be an issue in Calcite but was patched about a year ago with the corresponding test cases added. When debugging these test cases, I never touched the typePrecedence block that's an issue for us. Looks like I'll need to investigate more.

5. Array constructors are not supported (another issue entirely)
Yes, I did make a jira for it, but perhaps we can expand it as you said.
https://issues.apache.org/jira/browse/PHOENIX-3502

6. Code refactoring:
Having them in one spot is starting to make more sense; in that case, I agree that PhoenixSchema would be the best location for everything. What I can do is having them ordered according to how they logically progress, and have a comment block at the top that describes the behavior. 

In my opinion it's appropriate and reasonable to include 1, 3, and 6 in this pull request. The other issues may take longer, and would postpone integrating this component.

maryannxue:
2. We'll leave it then. I think the future solution can be either 1) add as built-in funcs in Calcite; or 2) add type derivation mechanism for UDFs in Calcite.
3. See "getAdjustedTime" in CalciteUtils then. You might need to do something similar somewhere, or this method is doing something wrong in the first place.

Please go ahead with 1, 3 and 6, but feel free to leave out 3 if it doesn't work out easily for you.

PR #223, opened 2016-11-28
PHOENIX-3488 Support COUNT(DISTINCT x) in Phoenix-Calcite Integration

lomoree:


phoenix-core/src/main/java/org/apache/phoenix/calcite/rules/PhoenixConverterRules.java
```diff
@@ -899,9 +899,6 @@ public static boolean isConvertible(Aggregate input) {
         if (input.getGroupSets().size() > 1)
             return false;
         
-        if (input.containsDistinctCall())
```
maryannxue: One minor nit, would you mind add check for other distinct functions? coz we don't support SUM DISTINCT yet.

lomoree: @maryannxue added a commit to detect for any distinct calls that aren't of type COUNT. Thanks!

PR #224, opened 2016-12-03
PHOENIX-3477 Support sequence arithmetic in Calcite-Phoenix

lomoree:
PhoenixConverterRules is meant to route any logical project with a sequence through PhoenixClientProject. Before, sequences were not properly being found. This patch simplifies the logic and adds that support.
-- Check for nested sequences
-- Use the Sequence Manager now available in StatementContext
-- Add relevant test cases to CalciteIT

### DIFF
--- a/phoenix-core/src/it/java/org/apache/phoenix/calcite/CalciteIT.java
+++ b/phoenix-core/src/it/java/org/apache/phoenix/calcite/CalciteIT.java
@@ -1343,6 +1343,28 @@ public class CalciteIT extends BaseCalciteIT {
                         {7L, "0000000005", "T5", "0000000005", "S5"},
                         {8L, "0000000006", "T6", "0000000006", "S6"}})
                 .close();
+
+        start(false, 1000f).sql("select NEXT VALUE FOR my.seq1 + 1, entity_id from aTable where a_string = 'a'")
+                .explainIs("PhoenixToEnumerableConverter\n" +
+                        "  PhoenixClientProject(EXPR$0=[+(NEXT_VALUE('\"MY\".\"SEQ1\"'), 1)], ENTITY_ID=[$1])\n" +
+                        "    PhoenixTableScan(table=[[phoenix, ATABLE]], filter=[=($2, 'a')])\n")
+                .resultIs(1, new Object[][]{
+                        {11L, "00A123122312312"},
+                        {13L, "00A223122312312"},
+                        {15L, "00A323122312312"},
+                        {17L, "00A423122312312"}})
+                .close();
+
+        start(false, 1000f).sql("select (NEXT VALUE FOR my.seq1 * 2) + 1, entity_id from aTable where a_string = 'a'")
+                .explainIs("PhoenixToEnumerableConverter\n" +
+                        "  PhoenixClientProject(EXPR$0=[+(*(NEXT_VALUE('\"MY\".\"SEQ1\"'), 2), 1)], ENTITY_ID=[$1])\n" +
+                        "    PhoenixTableScan(table=[[phoenix, ATABLE]], filter=[=($2, 'a')])\n")
+                .resultIs(1, new Object[][]{
+                        {37L, "00A123122312312"},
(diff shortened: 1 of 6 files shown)


phoenix-core/src/it/java/org/apache/phoenix/calcite/CalciteIT.java
```diff
@@ -1343,6 +1343,28 @@ public void initTable() throws Exception {
                         {7L, "0000000005", "T5", "0000000005", "S5"},
                         {8L, "0000000006", "T6", "0000000006", "S6"}})
                 .close();
+
```
maryannxue: I remember you said you ran into a problem using PhoenixServerProject, but it's PhoenixClientProject here again. Does it oscillate between the two? Could you please try {{start(false, 1f)}} instead?

phoenix-core/src/main/java/org/apache/phoenix/calcite/CalciteUtils.java
```diff
@@ -1115,6 +1099,11 @@ public Void visitCall(RexCall call) {
                         || call.getKind() == SqlKind.NEXT_VALUE)) {
                 sequenceValueCall = call;
             }
+            if (sequenceValueCall == null){
```
maryannxue: Think we are good to remove this class, right?

phoenix-core/src/main/java/org/apache/phoenix/calcite/rel/PhoenixClientProject.java
```diff
@@ -73,13 +70,10 @@ public QueryPlan implement(PhoenixRelImplementor implementor) {
...
         implementor.popContext();
         
-        
-        PhoenixSequence sequence = CalciteUtils.findSequence(this);
-        final SequenceManager seqManager = sequence == null ?
-                null : new SequenceManager(new PhoenixStatement(sequence.pc));
-        implementor.setSequenceManager(seqManager);
-        TupleProjector tupleProjector = project(implementor);
-        if (seqManager != null) {
+
```
maryannxue: This entire wrapping logic should live in PhoenixToEnumerableConverter instead. Wrapping it with an iterator not a query plan would be better, something like:
{code}

        return new DelegateQueryPlan((QueryPlan) plan) {
            @Override
            public ResultIterator iterator() throws SQLException {
                ResultIterator iterator = iterator(DefaultParallelScanGrouper.getInstance());
                if (phoenixImplementor.getSequenceManager().getSequenceCount() > 0) {
                    iterator = new SequenceResultIterator(iterator, phoenixImplementor.getSequenceManager());
                }
                return iterator;
            }
            @Override
            public ExplainPlan getExplainPlan() throws SQLException {
{code}

phoenix-core/src/it/java/org/apache/phoenix/calcite/CalciteIT.java
lomoree: While that was my initial thinking, upon looking deeper I discovered a different source for the issue

The reason a ServerProject was being created was because of visitCall(). It wasn't recursively checking for sequences. Therefore, when an operator like SqlKind.PLUS was present and the sequence was present, the sequence would be pushed down to a child operand and not be seen by visitCall()

PhoenixConverterRules is designed so that a ServerProject does not get created if a sequence is present

 
 ```
private static Predicate<LogicalProject> NO_SEQUENCE = new Predicate<LogicalProject>() {
            @Override
            public boolean apply(LogicalProject input) {
                return !CalciteUtils.hasSequenceValueCall(input);
            }            
        };
```

maryannxue: Thank you for the info, @lomoree! Very helpful! I was wondering if there had been any check of that kind during conversion. Anyway, would you mind removing that NO_SEQUENCE predicate and trying the changes I suggested below?

lomoree:
After discussing options, we kept original design and merged as is. Closing PR.

PR #232, opened 2017-02-15
Pull request for merging the encode columns branch to 4.x-HBase-0.98

samarthjain:


samarthjain:
@jtaylor-sfdc @twdsilva please review

JamesRTaylor:
@enis, @chrajeshbabu, @ankitsinghal - would you guys like to take a look? Would be great to get some feedback.

@samarthjain - maybe you can point out some key parts of the pull request to focus where you'd like feedback?

phoenix-core/src/main/java/org/apache/phoenix/schema/tuple/PositionBasedMultiKeyValueTuple.java
```diff
@@ -0,0 +1,86 @@
...
+
+import static com.google.common.base.Preconditions.checkArgument;
+
+import java.util.List;
+
+import org.apache.hadoop.hbase.Cell;
+import org.apache.hadoop.hbase.io.ImmutableBytesWritable;
+
+/**
+ * Tuple that uses the
```
samarthjain: This is a new tuple implementation that is tied with the EncodedColumnQualifierCellsList. It essentially provides a way of getting hold of cell belonging to a cq/cf by doing a position based look up as opposed to a MultiKeyValueTuple where we have to do a binary search in the List<Cell>.

phoenix-core/src/main/java/org/apache/phoenix/index/IndexMaintainer.java
```diff
@@ -291,39 +324,59 @@ public static void serializeAdditional(PTable table, ImmutableBytesWritable inde
...
+    private QualifierEncodingScheme encodingScheme;
+    private ImmutableStorageScheme immutableStorageScheme;
+    /*
+     * Information for columns of data tables that are being indexed. The first part of the pair is column family name
+     * and second part is the column name. The reason we need to track this state is because for certain storage schemes
+     * like ImmutableStorageScheme#SINGLE_CELL_ARRAY_WITH_OFFSETS, the column for which we need to generate an index
+     * table put/delete is different from the columns that are indexed in the phoenix schema. This information helps us
+     * determine whether or not certain operations like DROP COLUMN should impact the index.
+     */
+    private Set<Pair<String, String>> indexedColumnsInfo;
```
JamesRTaylor: This isn't part of index maintenance so it doesn't really belong here. We should at a minimum file a JIRA to get rid of it or ideally get rid of it - take a look and see where it's used. We can do this in a follow up release too.

phoenix-core/src/main/java/org/apache/phoenix/query/QueryServicesOptions.java
```diff
@@ -263,6 +264,8 @@
     public static final String DEFAULT_CLIENT_CACHE_ENCODING = PTableRefFactory.Encoding.OBJECT.toString();
     public static final boolean DEFAULT_AUTO_UPGRADE_ENABLED = true;
     public static final int DEFAULT_CLIENT_CONNECTION_CACHE_MAX_DURATION = 86400000;
+    public static final int DEFAULT_COLUMN_ENCODED_BYTES = QualifierEncodingScheme.FOUR_BYTE_QUALIFIERS.getSerializedMetadataValue();
```
JamesRTaylor: I think a 2 byte qualifier is a better default. WDYT?

phoenix-core/src/main/java/org/apache/phoenix/schema/PTableImpl.java
```diff
@@ -106,7 +115,8 @@
     private List<PColumnFamily> families;
     private Map<byte[], PColumnFamily> familyByBytes;
     private Map<String, PColumnFamily> familyByString;
-    private ListMultimap<String,PColumn> columnsByName;
+    private ListMultimap<String, PColumn> columnsByName;
+    private Map<KVColumnFamilyQualifier, PColumn> kvColumnsByQualifiers;
```
JamesRTaylor: It's unfortunate that we need another map. Is this used often?

phoenix-core/src/main/java/org/apache/phoenix/compile/FromCompiler.java
```diff
@@ -858,7 +868,7 @@ public ColumnRef resolveColumn(String schemaName, String tableName, String colNa
                 while (iterator.hasNext()) {
                     TableRef tableRef = iterator.next();
                     try {
-                        PColumn column = tableRef.getTable().getColumn(colName);
+                        PColumn column = tableRef.getTable().getPColumnForColumnName(colName);
```
JamesRTaylor: I see why you renamed it. How about getColumnFromName() and getColumnFromQualifier()?

PR #237, opened 2017-03-24
4.9 h base 1.2

1emo:


joshelser:
@1emo what's going on with this PR? What are you trying to achieve? Please open a JIRA issue describing what you want to do.

stoty:
If Josh couldn't figure what this PR is, then we certainly don't need this.

PR #239, opened 2017-04-24
PHOENIX-3744: Support snapshot scanners for MR-based Non-aggregate queries

akshita-malhotra:
- Parallel Scan grouper is extended to differentiate the functionality for getting region boundaries

- Added integration test, compares the snapshot read result with the result from select query by setting CurrentScn value.

- the configuration parameter is the snapshot name key, if set do a snapshot read

- Used an existing PhoenixIndexDBWritable class for the purpose of testing, will add a new one as I will add more tests.

- ExpressionProjector functionality is extended for snapshots as the keyvalue format returned from TableSnapshotScanner is different from ClientScanner and therefore not properly interrupted by Phoenix thereby returning null in case of projected columns.
For the same table, following shows the different format of the keyvalues:

1. ClientScanner:
keyvalues={AAPL/_v:\x00\x00\x00\x01/1493061452132/Put/vlen=7/seqid=0/value=SSDD}

2. TableSnapshotScanner:
keyvalues={AAPL/0:\x00\x00\x00\x00/1493061673859/Put/vlen=1/seqid=4/value=x, 
AAPL/0:\x80\x0B/1493061673859/Put/vlen=4/seqid=4/value=SSDD}

@JamesRTaylor @lhofhansl 

To DO:
Add more integration tests to cover different scenarios such as where clause etc

### DIFF
--- a/phoenix-core/src/main/java/org/apache/phoenix/iterate/BaseResultIterators.java
+++ b/phoenix-core/src/main/java/org/apache/phoenix/iterate/BaseResultIterators.java
@@ -501,6 +501,11 @@ public abstract class BaseResultIterators extends ExplainTable implements Result
             return scans;
     }
 
+    private List<HRegionLocation> getRegionBoundaries(ParallelScanGrouper scanGrouper)
+        throws SQLException{
+        return scanGrouper.getRegionBoundaries(context, physicalTableName);
+    }
+
     private static List<byte[]> toBoundaries(List<HRegionLocation> regionLocations) {
         int nBoundaries = regionLocations.size() - 1;
         List<byte[]> ranges = Lists.newArrayListWithExpectedSize(nBoundaries);
@@ -564,7 +569,9 @@ public abstract class BaseResultIterators extends ExplainTable implements Result
     private List<Scan> addNewScan(List<List<Scan>> parallelScans, List<Scan> scans, Scan scan, byte[] startKey, boolean crossedRegionBoundary, HRegionLocation regionLocation) {
         boolean startNewScan = scanGrouper.shouldStartNewScan(plan, scans, startKey, crossedRegionBoundary);
         if (scan != null) {
-            scan.setAttribute(BaseScannerRegionObserver.SCAN_REGION_SERVER, regionLocation.getServerName().getVersionedBytes());
+            if (regionLocation.getServerName() != null) {
+                scan.setAttribute(BaseScannerRegionObserver.SCAN_REGION_SERVER, regionLocation.getServerName().getVersionedBytes());
+            }
         	scans.add(scan);
         }
         if (startNewScan && !scans.isEmpty()) {
@@ -592,8 +599,7 @@ public abstract class BaseResultIterators extends ExplainTable implements Result
      * @throws SQLException
      */
     private List<List<Scan>> getParallelScans(Scan scan) throws SQLException {
-        List<HRegionLocation> regionLocations = context.getConnection().getQueryServices()
-                .getAllTableRegions(physicalTableName);
+        List<HRegionLocation> regionLocations = getRegionBoundaries(scanGrouper);
         List<byte[]> regionBoundaries = toBoundaries(regionLocations);
         int regionIndex = 0;
         int stopIndex = regionBoundaries.size();
@@ -645,8 +651,7 @@ public abstract class BaseResultIterators extends ExplainTable implements Result
      * @throws SQLException
      */
     private List<List<Scan>> getParallelScans(byte[] startKey, byte[] stopKey) throws SQLException {
-        List<HRegionLocation> regionLocations = context.getConnection().getQueryServices()
-                .getAllTableRegions(physicalTableName);
+        List<HRegionLocation> regionLocations = getRegionBoundaries(scanGrouper);
         List<byte[]> regionBoundaries = toBoundaries(regionLocations);
         ScanRanges scanRanges = context.getScanRanges();
         PTable table = getTable();
--- a/phoenix-core/src/main/java/org/apache/phoenix/iterate/DefaultParallelScanGrouper.java
+++ b/phoenix-core/src/main/java/org/apache/phoenix/iterate/DefaultParallelScanGrouper.java
@@ -17,46 +17,58 @@
  */
 package org.apache.phoenix.iterate;
 
+import java.sql.SQLException;
 import java.util.List;
 
+import com.google.common.base.Preconditions;
+import org.apache.hadoop.hbase.HRegionLocation;
 import org.apache.hadoop.hbase.client.Scan;
 import org.apache.phoenix.compile.QueryPlan;
+import org.apache.phoenix.compile.StatementContext;
 import org.apache.phoenix.schema.PTable;
 import org.apache.phoenix.schema.PTable.IndexType;
 import org.apache.phoenix.schema.SaltingUtil;
+import org.apache.phoenix.schema.TableRef;
 import org.apache.phoenix.util.ScanUtil;
 
 /**
  * Default implementation that creates a scan group if a plan is row key ordered (which requires a merge sort),
- * or if a scan crosses a region boundary and the table is salted or a local index.   
+ * or if a scan crosses a region boundary and the table is salted or a local index.
  */
 public class DefaultParallelScanGrouper implements ParallelScanGrouper {
-	
-	private static final DefaultParallelScanGrouper INSTANCE = new DefaultParallelScanGrouper();
 
-    public static DefaultParallelScanGrouper getInstance() {
-        return INSTANCE;
+  private static DefaultParallelScanGrouper INSTANCE = new DefaultParallelScanGrouper();
+
+
+  public DefaultParallelScanGrouper() {
+  }
+
+  public static DefaultParallelScanGrouper getInstance() {
+    return INSTANCE;
+  }
+
+
+  @Override
+  public boolean shouldStartNewScan(QueryPlan plan, List<Scan> scans, byte[] startKey, boolean crossedRegionBoundary) {
+    PTable table = plan.getTableRef().getTable();
+    boolean startNewScanGroup = false;
+    if (!plan.isRowKeyOrdered()) {
+      startNewScanGroup = true;
+    } else if (crossedRegionBoundary) {
+      if (table.getIndexType() == IndexType.LOCAL) {
+        startNewScanGroup = true;
+      } else if (table.getBucketNum() != null) {
+        startNewScanGroup = scans.isEmpty() ||
+            ScanUtil.crossesPrefixBoundary(startKey,
+                ScanUtil.getPrefix(scans.get(scans.size()-1).getStartRow(), SaltingUtil.NUM_SALTING_BYTES),
+                SaltingUtil.NUM_SALTING_BYTES);
+      }
     }
-    
-    private DefaultParallelScanGrouper() {}
+    return startNewScanGroup;
+  }
 
-	@Override
-	public boolean shouldStartNewScan(QueryPlan plan, List<Scan> scans, byte[] startKey, boolean crossedRegionBoundary) {
-		PTable table = plan.getTableRef().getTable();
-		boolean startNewScanGroup = false;
-        if (!plan.isRowKeyOrdered()) {
-            startNewScanGroup = true;
-        } else if (crossedRegionBoundary) {
-            if (table.getIndexType() == IndexType.LOCAL) {
-                startNewScanGroup = true;
-            } else if (table.getBucketNum() != null) {
-                startNewScanGroup = scans.isEmpty() ||
-                        ScanUtil.crossesPrefixBoundary(startKey,
-                                ScanUtil.getPrefix(scans.get(scans.size()-1).getStartRow(), SaltingUtil.NUM_SALTING_BYTES), 
-                                SaltingUtil.NUM_SALTING_BYTES);
-            }
-        }
-        return startNewScanGroup;
-    }
-
+  @Override
+  public List<HRegionLocation> getRegionBoundaries(StatementContext context, byte[] tableName) throws SQLException{
+    return context.getConnection().getQueryServices().getAllTableRegions(tableName);
+  }
 }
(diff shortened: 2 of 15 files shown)


akshita-malhotra:
- Snapshot scanner for non-aggregate queries.
- Added integration tests (simple select query, conditional and limit)
- Abstracted out ScanRegionObserver code to fetch the processed region scanner without the coprocessor environment
- To make the snapshot work for aggregate queries, due to the complexity of the aggregate region observer  code it is almost impossible to refactor the code without fully understanding the functionality. Will require some guidance if the aggregate query use case is required.
@JamesRTaylor @lhofhansl

JamesRTaylor:
I don't think it's necessary to fully understand the functionality to do the refactoring I've mentioned, @akshita-malhotra. Here's how I'd recommend approaching it:

* create a new interface solely for the purpose of only abstracting RegionCoprocessorEnvironment access called RegionContext. The interface would have at least two methods: getRegion and getConfiguration. We might need more if other methods are called in RegionCoprocessorEnvironment.
* have two implementations of this interface: RegionCoprocessorContext and RegionShapshotContext. The constructor of RegionCoprocessorContext would take a RegionCoprocessorEnvironment as an argument, while the RegionShapshotContext would take a Region and Configuration.
* do an across the board replace of RegionCoprocessorEnvironment with RegionContext. You can likely not do this for secondary index related code (org.apache.phoenix.hbase.index.Indexer and PhoenixTransactionalIndexer). You'll find out here if other methods are called from RegionCoprocessorEnvironment or ObserverContext (which can be dealt with in a variety of ways, for example by throwing an UnsupportedOperationException if need be in the snapshot implementation).
* in the top level coprocessor methods that pass in RegionCoprocessorEnvironment (mostly abstract BaseScannerRegionObserver class), instantiate a RegionCoprocessorContext by passing in the RegionCoprocessorEnvironment. From this point onward, all access will go through the RegionContext interface.

You could do this refactoring completely separate from the PHOENIX-3744 so that you don't mix the two. Then PHOENIX-3744 would have something like a RegionScannerFactory (your RegionObserverUtil) that gives you back a RegionScanner given a RegionContext and you'd create a RegionShapshotContext as the backing implementation in your snapshot reading code.

JamesRTaylor:
Please amend your commit message to be prefixed with PHOENIX-3744 (instead of Phoenix-3744) so that review comments show up as comments on the JIRA, @akshita-malhotra. Also, let's get a patch attached to the JIRA so we can get a test run to make sure there are no regressions. Directions here: https://phoenix.apache.org/contributing.html#Generate_a_patch

akshita-malhotra:
Sure, I will do that. Thanks @JamesRTaylor

phoenix-core/src/main/java/org/apache/phoenix/iterate/BaseResultIterators.java
```diff
@@ -496,6 +496,17 @@ public BaseResultIterators(QueryPlan plan, Integer perScanLimit, Integer offset,
             return scans;
     }
 
+    private List<HRegionLocation> getRegionBoundaries(ParallelScanGrouper scanGrouper)
+        throws SQLException{
+        if (scanGrouper instanceof DefaultParallelScanGrouper) {
```
JamesRTaylor: These casts should not be necessary. Either add an init method to ParallelScanGrouper or call the init from the implementing class in getRegionBoundaries.

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/PhoenixInputFormat.java
```diff
@@ -185,7 +188,10 @@ private QueryPlan getQueryPlan(final JobContext context, final Configuration con
                 scan.setAttribute(BaseScannerRegionObserver.TX_SCN, Bytes.toBytes(Long.valueOf(txnScnValue)));
             }
             // Initialize the query plan so it sets up the parallel scans
-            queryPlan.iterator(MapReduceParallelScanGrouper.getInstance());
+            if(snapshotName != null)
```
JamesRTaylor: Add an init method to the ParallelScanGrouper interface that returns an instance and have the TableMRParallelScanGrouper implementation of that return TableMRParallelScanGrouper.getInstance(). We shouldn't need switching logic here based on snapshotName.

phoenix-core/src/main/java/org/apache/phoenix/monitoring/MetricType.java
```diff
@@ -25,6 +25,7 @@
     QUERY_TIME("Query times"),
     NUM_PARALLEL_SCANS("Number of scans that were executed in parallel"),
     SCAN_BYTES("Number of bytes read by scans"),
+    SNAPSHOT_SCAN_BYTES("Number of bytes read by snapshot scans"),
```
JamesRTaylor: Do we need a different metric here? Or would it make more sense to use SCAN_BYTES for the snapshot case too?

phoenix-core/src/main/java/org/apache/phoenix/util/NonAggregateROUtil.java
```diff
@@ -0,0 +1,355 @@
...
+import java.io.DataInputStream;
+import java.io.IOException;
+import java.sql.SQLException;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Set;
+
+import static org.apache.phoenix.util.EncodedColumnsUtil.getMinMaxQualifiersFromScan;
+
+public class NonAggregateROUtil extends RegionObserverUtil {
```
JamesRTaylor: In general, our Util classes only have static methods. How about a different naming convention for these that a post fixed with "Factory" instead of "Util, like NonAggregateRegionScannerFactory?

phoenix-core/src/test/java/org/apache/phoenix/query/BaseTest.java
```diff
@@ -592,7 +592,7 @@ private static String initMiniCluster(Configuration conf, ReadOnlyProps override
         setUpConfigForMiniCluster(conf, overrideProps);
         utility = new HBaseTestingUtility(conf);
         try {
-            utility.startMiniCluster(NUM_SLAVES_BASE);
+            utility.startMiniCluster(NUM_SLAVES_BASE, true);
```
JamesRTaylor: Just curious on this change. Is it necessary?

JamesRTaylor:
Thanks for the updates, @akshita-malhotra - I made a few comments/questions.

akshita-malhotra:
Thanks @JamesRTaylor . I squashed the commits and changed the prefix of the commit message. I will answer/make appropriate changes and upload the patch onto the jira.

akshita-malhotra:
@JamesRTaylor 
- Changed ParallelScanGrouper classes as per the review
- Changes to BaseTest were to avoid the following error:
"Restore directory cannot be a sub directory of HBase root directory"
Therefore, was sending true to create the root dir. Changed to use a random dir instead to avoid to make these changes
- Refactored the util classes to Factory

Also, uploaded the patch on the jira.

phoenix-core/src/main/java/org/apache/phoenix/iterate/ScanningResultIterator.java
```diff
@@ -38,10 +38,12 @@
 public class ScanningResultIterator implements ResultIterator {
     private final ResultScanner scanner;
     private final CombinableMetric scanMetrics;
+    private final boolean snapshotScan;
```
JamesRTaylor: Looks like the changes to this file aren't necessary. Please revert.

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/util/PhoenixMapReduceUtil.java
```diff
@@ -63,6 +67,66 @@ public static void setInput(final Job job, final Class<? extends DBWritable> inp
...
+     *
+     * @param job
+     * @param inputClass DBWritable class
+     * @param snapshotName The name of a snapshot (of a table) to read from
+     * @param tableName Input table name
+     * @param restoreDir a temporary dir to copy the snapshot files into
+     * @param conditions Condition clause to be added to the WHERE clause. Can be <tt>null</tt> if there are no conditions.
+     * @param fieldNames fields being projected for the SELECT query.
+     */
+    public static void setInput(final Job job, final Class<? extends DBWritable> inputClass, final String snapshotName, String tableName,
```
JamesRTaylor: Why wouldn't we want to take the snapshot here instead of passing in the snapshot name?

phoenix-core/src/main/java/org/apache/phoenix/iterate/DefaultParallelScanGrouper.java
```diff
@@ -17,46 +17,79 @@
...
-                startNewScanGroup = true;
-            } else if (table.getBucketNum() != null) {
-                startNewScanGroup = scans.isEmpty() ||
-                        ScanUtil.crossesPrefixBoundary(startKey,
-                                ScanUtil.getPrefix(scans.get(scans.size()-1).getStartRow(), SaltingUtil.NUM_SALTING_BYTES), 
-                                SaltingUtil.NUM_SALTING_BYTES);
-            }
-        }
-        return startNewScanGroup;
+  private static DefaultParallelScanGrouper INSTANCE = new DefaultParallelScanGrouper();
```
JamesRTaylor: I don't think that DefaultParallelScanGrouper can be a singleton with the state of context and tableName inside of it.

phoenix-core/src/main/java/org/apache/phoenix/iterate/TableMRParallelScanGrouper.java
```diff
@@ -0,0 +1,99 @@
...
+import org.apache.hadoop.hbase.snapshot.SnapshotManifest;
+import org.apache.phoenix.compile.StatementContext;
+import org.apache.phoenix.mapreduce.util.PhoenixConfigurationUtil;
+
+import java.sql.SQLException;
+import java.util.List;
+
+public class TableMRParallelScanGrouper extends MapReduceParallelScanGrouper {
+
+  private static TableMRParallelScanGrouper INSTANCE = null;
```
JamesRTaylor: Same comment here - how can this be a singleton?

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/PhoenixInputFormat.java
```diff
@@ -185,7 +184,7 @@ private QueryPlan getQueryPlan(final JobContext context, final Configuration con
                 scan.setAttribute(BaseScannerRegionObserver.TX_SCN, Bytes.toBytes(Long.valueOf(txnScnValue)));
             }
             // Initialize the query plan so it sets up the parallel scans
-            queryPlan.iterator(MapReduceParallelScanGrouper.getInstance());
+            queryPlan.iterator(TableMRParallelScanGrouper.init(configuration));
```
JamesRTaylor: Can't this just instantiate a TableMRParallelScanGrouper and the init method can be invoked in the constructor so that it doesn't need to be exposed in the interface?

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/util/PhoenixConfigurationUtil.java
```diff
@@ -192,6 +196,18 @@ public static void setOutputTableName(final Configuration configuration, final S
     public static void setUpsertColumnNames(final Configuration configuration,final String[] columns) {
         setValues(configuration, columns, MAPREDUCE_UPSERT_COLUMN_COUNT, MAPREDUCE_UPSERT_COLUMN_VALUE_PREFIX);
     }
+
+    public static void setSnapshotNameKey(final Configuration configuration, final String snapshotName) {
+        Preconditions.checkNotNull(configuration);
+        Preconditions.checkNotNull(snapshotName);
+        configuration.set(SNAPSHOT_NAME_KEY, snapshotName);
```
JamesRTaylor: The idea of having the snapshot name in the configuration is not going to translate well when we want to expose snapshot reads for queries in general as the configuration is a global object.

phoenix-core/src/main/java/org/apache/phoenix/iterate/RegionContext.java
```diff
@@ -0,0 +1,15 @@
...
+
+/**
+ * Maintains coprocessor environment state to
+ * make region observer functionality work for
+ * non-coprocessor environment
+ */
+public interface RegionContext {
+
+  /** @return the Coprocessor environment */
+  RegionCoprocessorEnvironment getEnvironment();
```
JamesRTaylor: If this is the only method, we don't really need this interface. Why not just keep using RegionCoprocessorEnvironment in our APIs and you can just instantiate you snapshot-based one prior to getting the RegionScanner?

JamesRTaylor:
Thanks for the revision, @akshita-malhotra. It's looking very good. I made some comments inline.

akshita-malhotra:
@JamesRTaylor  Thanks a lot for the review. I have made the suggested changes and uploaded the updated patch on the jira. 
Regarding creating snapshot to generalize the use of snapshots for M/R jobs, I was under the impression that we are passing the snapshot name as input after our last discussion with @lhofhansl and Rahul G.
If we are to follow the former approach, I will go ahead and make the changes.

JamesRTaylor:
Patch looks very good, @akshita-malhotra. What's the advantage, @lhofhansl, of forcing users to create the snapshot themselves before starting the job? Wouldn't it be simpler for the snapshot to be created during the setup/initialization of the MR job? In the non MR case, when we want to support running arbitrary queries over snapshot(s), seems like we'd want Phoenix to create them, no? Otherwise, we'd need to provide the user with some means of associating a snapshot with a table name (which might get cumbersome). The alternative is to let Phoenix manage this transparently.

JamesRTaylor:
Had an offline discussion with @lhofhansl and he brought up the use case of having a snapshot at a known point in time (i.e. midnight last night) and wanting to run Phoenix queries over it through MR. That seems like a good reason to pass in the snapshot name rather than build it when the MR starts.

I'll commit this into 4.x and master. Nice work, @akshita-malhotra!

stoty:
Already merged.

PR #242, opened 2017-05-01
PQS metrics -  https://issues.apache.org/jira/browse/PHOENIX-3655 

rahulsIOT:
Phoenix Query Server metrics
https://issues.apache.org/jira/browse/PHOENIX-3655

The following properties is added along with the defaults
1. "phoenix.query.server.metrics" - default is true ( means pqs will collect metrics by default )
2. "phoenix.query.server.metrics.report.interval.ms" - default is 30 sec ( every 30 sec, the global metrics will be output to sink )
3. "phoenix.query.server.metrics.type.of.sink" - default is file  ( other option is slf4j)
4. "phoenix.query.server.filesink.filename" - Filename if using fileSink (default is  /tmp/pqsSinkFile)

if user setup "phoenix.query.server.metrics"  in hbase-site.xml to true, we add the following properties to JDBC connection (QueryServices.COLLECT_REQUEST_LEVEL_METRICS=  true). The starts the metrics collect at request level. 

There are two types of metrics 
1. Global - This metrics is collect every interval and pushed to sink ( File or Slf4j). Apart from this, it is also exposed in JMX via Mbeans.
2. Request - These metrics is collected at statement and connection close level. Once collected, it is pushed to the sink. These metrics are not exposed via JMX as metrics are for specific table requests.  For requests level metrics, we have file and Slf4j ( LOG) as types of sink. So, the metrics will be pushed to a specified file or log file .

For details on Phoenix Metrics - refer - https://phoenix.apache.org/metrics.html


phoenix-queryserver/src/main/java/org/apache/phoenix/queryserver/metrics/PqsMetricsSystem.java
```diff
@@ -0,0 +1,82 @@
...
+    public PqsSink pqsSink;
+
+
+    public PqsMetricsSystem(){
+
+        PqsGlobalMetrics pqsGlobalMetricsToJMX = null;
+        try {
+            pqsSink = getSinkObject();
+            pqsGlobalMetricsToJMX = new PqsGlobalMetrics(pqsSink);
+            globalMetricThread = new Thread(pqsGlobalMetricsToJMX);
```
joshelser: Should set an uncaught exception handler on this thread.

phoenix-queryserver/src/main/java/org/apache/phoenix/queryserver/metrics/PqsMetricsSystem.java
```diff
@@ -0,0 +1,82 @@
...
+     }
+
+      private PqsSink getSinkObject(){
+        PqsSink pqsSink;
+        String typeOfSink = PqsConfiguration.getTypeOfSink();
+        switch(typeOfSink) {
+            case "file":
+                pqsSink = new PqsFileSink();
+                break;
+            case "slf4j": pqsSink = new PqsSlf4jSink() ; break;
```
joshelser: You don't need to duplicate the body of this case statement, just remove it and let it fall into the default case.

phoenix-queryserver/src/main/java/org/apache/phoenix/queryserver/metrics/sink/PqsFileSink.java
```diff
@@ -0,0 +1,66 @@
...
+    private static final Logger LOG = LoggerFactory.getLogger(PqsFileSink.class);
+    private String filename = PqsConfiguration.getFileSinkFilename();
+
+    public PqsFileSink() {
+        try {
+            writer = filename == null ? System.out
+                    : new PrintStream(new FileOutputStream(new File(filename)),
+                    true, "UTF-8");
+        } catch (FileNotFoundException e) {
+            LOG.error("Error creating "+ filename, e);
```
joshelser: If one of these exceptions is thrown, it will cause this object to be unusable.

phoenix-queryserver/src/main/java/org/apache/phoenix/queryserver/metrics/sink/PqsSlf4jSink.java
```diff
@@ -0,0 +1,44 @@
...
+
+    @Override
+    public void close()  {}
+
+    @Override
+    public void writeJson(String json){
+        if (json != null)
+        {
+            if (logger.isInfoEnabled())
+                logger.info(json);
```
joshelser: Should this level be configurable? Are certain metrics worthy of INFO while others should be DEBUG?

phoenix-queryserver/src/main/java/org/apache/phoenix/queryserver/server/PhoenixMetaFactoryImpl.java
```diff
@@ -68,7 +70,10 @@ public Meta create(List<String> args) {
             "0 or 1 argument expected. Received " + Arrays.toString(args.toArray()));
       }
       // TODO: what about -D configs passed in from cli? How do they get pushed down?
-      return new JdbcMeta(url, info);
+      if (PqsConfiguration.isMetricsTurnedOn()) {
```
joshelser: Only return a `PQSMetricsMeta` if metrics collection is enabled.

phoenix-queryserver/src/it/java/org/apache/phoenix/end2end/QueryServerBasicsIT.java
```diff
@@ -57,10 +64,17 @@
   private static QueryServerThread AVATICA_SERVER;
   private static Configuration CONF;
   private static String CONN_STRING;
+  private static FileReader fileReader;
 
   @BeforeClass
   public static void beforeClass() throws Exception {
+    //clear up old file in /tmp directory
+    PrintWriter writer = new PrintWriter(new File(PqsConfiguration.DEFAULT_PHOENIX_PQS_FILE_SINK_FILENAME));
```
joshelser: Create this file in the maven build dir "target"

phoenix-queryserver/src/it/java/org/apache/phoenix/end2end/QueryServerBasicsIT.java
```diff
@@ -161,5 +184,19 @@ public void smokeTest() throws Exception {
         }
       }
     }
+    assertFalse(" metrics file should not be empty", checkIfTempFileIsEmpty());
```
joshelser: Can you add a test which actually verifies that some of the metrics we'd expect to be in this file are there?

This test would silently pass if garbage/errors were somehow written to this file.

phoenix-queryserver/src/it/java/org/apache/phoenix/end2end/QueryServerBasicsIT.java
```diff
@@ -161,5 +184,19 @@ public void smokeTest() throws Exception {
...
+
   }
-}
+
+  private static boolean checkIfTempFileIsEmpty() throws Exception {
+    BufferedReader br = new BufferedReader(fileReader);
```
joshelser: Use try-with-resources?

phoenix-queryserver/src/main/java/org/apache/phoenix/queryserver/metrics/PqsConfiguration.java
```diff
@@ -0,0 +1,67 @@
...
+
+public class PqsConfiguration {
+
+    public static final Configuration configuration= HBaseConfiguration.create();
+    public static final String PHOENIX_QUERY_SERVER_METRICS = "phoenix.query.server.metrics";
+    public static final String PHOENIX_PQS_METRIC_REPORTING_INTERVAL_MS = "phoenix.query.server.metrics.report.interval.ms";
+    public static final String PHOENIX_PQS_TYPE_OF_SINK = "phoenix.query.server.metrics.type.of.sink";
+    public static final String PHOENIX_PQS_FILE_SINK_FILENAME = "phoenix.query.server.filesink.filename";
+
+    public enum typeOfSink {
```
joshelser: nit: `MetricsSinkType`

phoenix-queryserver/src/main/java/org/apache/phoenix/queryserver/metrics/PqsConfiguration.java
```diff
@@ -0,0 +1,67 @@
...
+ * limitations under the License.
+ */
+
+package org.apache.phoenix.queryserver.metrics;
+
+
+import org.apache.hadoop.conf.Configuration;
+import org.apache.hadoop.hbase.HBaseConfiguration;
+
+public class PqsConfiguration {
```
joshelser: Why not in QueryServices/QueryServicesOptions like the rest of the PQS configuration keys/defaults?

phoenix-queryserver/src/main/java/org/apache/phoenix/queryserver/metrics/PqsConfiguration.java
```diff
@@ -0,0 +1,67 @@
...
+
+    public static final Configuration configuration= HBaseConfiguration.create();
```
joshelser: You shouldn't be creating another configuration option. The PhoenixConnection will already have one.

phoenix-queryserver/src/main/java/org/apache/phoenix/queryserver/metrics/PqsGlobalMetrics.java
```diff
@@ -0,0 +1,76 @@
...
+                        }
+                    });
+        }
+
+    }
+
+    @Override
+    public void run() {
+        while (true) {
+            int delay = PqsConfiguration.getReportingInterval();
```
joshelser: Isn't this constant? Get and set this value once.

phoenix-queryserver/src/main/java/org/apache/phoenix/queryserver/metrics/PqsGlobalMetrics.java
```diff
@@ -0,0 +1,76 @@
...
+            //besides writing to JMX, the global metrics is also written to pqsSink
+            //default pqsSink is slf4j ( logger)
+            phoenixGlobalMetricsCollection = PhoenixRuntime.getGlobalPhoenixClientMetrics();
+            this.pqsSink.writeGlobal(phoenixGlobalMetricsCollection);
+            try {
+                Thread.sleep(delay);
+            } catch (InterruptedException e) {
+                LOG.error(" Sleep thread interrupted for metrics collection ");
```
joshelser: Exit the thread if it was interrupted.

phoenix-queryserver/src/main/java/org/apache/phoenix/queryserver/metrics/PqsMetricsSystem.java
```diff
@@ -0,0 +1,82 @@
...
+                    LOG.error(" unable to stop the global metrics thread",ine);
+                }
+            }
+
+        }
+     }
+
+      private PqsSink getSinkObject(){
+        PqsSink pqsSink;
+        String typeOfSink = PqsConfiguration.getTypeOfSink();
```
joshelser: You should normalize the values set by a user `SLF4J` is treated differently than `slf4j`.

phoenix-queryserver/src/main/java/org/apache/phoenix/queryserver/metrics/sink/PqsFileSink.java
```diff
@@ -0,0 +1,66 @@
...
+    }
+
+
+    @Override
+    public void writeJson(String json){
+        if (json != null)
+        {
+            writer.print(json);
+        }
+        writer.println();
```
joshelser: Why print an empty line on every invocation?

phoenix-queryserver/src/main/java/org/apache/phoenix/queryserver/metrics/sink/PqsSlf4jSink.java
rahulsIOT: I think the following makes sense. i.e. if logging level is set to debug , then log  both global and request level metrics. But if logging level is set to info, then we log only request level metrics only. Do you agree ?

phoenix-queryserver/src/main/java/org/apache/phoenix/queryserver/server/PhoenixMetaFactoryImpl.java
rahulsIOT: ok. agreed.

phoenix-queryserver/src/it/java/org/apache/phoenix/end2end/QueryServerBasicsIT.java
rahulsIOT: agreed.

phoenix-queryserver/src/main/java/org/apache/phoenix/queryserver/metrics/PqsConfiguration.java
rahulsIOT: Ok. will refactor the configurations in  QueryServices/QueryServicesOption

phoenix-queryserver/src/main/java/org/apache/phoenix/queryserver/metrics/PqsConfiguration.java
rahulsIOT: removed.

phoenix-queryserver/src/main/java/org/apache/phoenix/queryserver/metrics/PqsGlobalMetrics.java
rahulsIOT: moved it out of run method to set the value once when PqsGlobalMetrics.java is instantiated.

phoenix-queryserver/src/main/java/org/apache/phoenix/queryserver/metrics/PqsGlobalMetrics.java
rahulsIOT: agreed.

rahulsIOT:
Metrics Snippet
**Statement Level -** 
{"Statement-OverAllReadRequestMetrics":{"RESULT_SET_TIME_MS":22,"QUERY_FAILED_COUNTER":0,"QUERY_TIMEOUT_COUNTER":0,"WALL_CLOCK_TIME_MS":52,"CACHE_REFRESH_SPLITS_COUNTER":0,"NUM_PARALLEL_SCANS":1}}

{"Statement-RequestReadMetrics":{"QUERYSERVERBASICSIT1494892379503":{"TASK_QUEUE_WAIT_TIME":0,"SCAN_BYTES":208,"TASK_END_TO_END_TIME":10,"TASK_EXECUTED_COUNTER":1,"TASK_EXECUTION_TIME":10,"TASK_REJECTED_COUNTER":0}}}

**Global Metrics -**

{"global":[{"numberOfSamples":0,"description":"Batch sizes of mutations","currentMetricState":"MUTATION_BATCH_SIZE: 0, Number of samples: 0","totalSum":0,"name":"MUTATION_BATCH_SIZE","value":0},{"numberOfSamples":0,"description":"Size of mutations in bytes","currentMetricState":"MUTATION_BYTES: 0, Number of samples: 0","totalSum":0,"name":"MUTATION_BYTES","value":0},{"numberOfSamples":0,"description":"Time it took to commit mutations","currentMetricState":"MUTATION_COMMIT_TIME: 0, Number of samples: 0","totalSum":0,"name":"MUTATION_COMMIT_TIME","value":0},{"numberOfSamples":7,"description":"Query times","currentMetricState":"QUERY_TIME: 52, Number of samples: 7","totalSum":52,"name":"QUERY_TIME","value":52},{"numberOfSamples":11,"description":"Number of scans that were executed in parallel","currentMetricState":"NUM_PARALLEL_SCANS: 11, Number of samples: 11","totalSum":11,"name":"NUM_PARALLEL_SCANS","value":11},{"numberOfSamples":2,"description":"Number of bytes read by scans","currentMetricState":"SCAN_BYTES: 192, Number of samples: 2","totalSum":192,"name":"SCAN_BYTES","value":192},{"numberOfSamples":0,"description":"Size of spool files created in bytes","currentMetricState":"SPOOL_FILE_SIZE: 0, Number of samples: 0","totalSum":0,"name":"SPOOL_FILE_SIZE","value":0},{"numberOfSamples":0,"description":"Number of bytes allocated by the memory manager","currentMetricState":"MEMORY_CHUNK_BYTES: 0, Number of samples: 0","totalSum":0,"name":"MEMORY_CHUNK_BYTES","value":0},{"numberOfSamples":0,"description":"Number of milliseconds threads needed to wait for memory to be allocated through memory manager","currentMetricState":"MEMORY_WAIT_TIME: 0, Number of samples: 0","totalSum":0,"name":"MEMORY_WAIT_TIME","value":0},{"numberOfSamples":11,"description":"Time in milliseconds tasks had to wait in the queue of the thread pool executor","currentMetricState":"TASK_QUEUE_WAIT_TIME: 5, Number of samples: 11","totalSum":5,"name":"TASK_QUEUE_WAIT_TIME","value":5},{"numberOfSamples":11,"description":"Time in milliseconds spent by tasks from creation to completion","currentMetricState":"TASK_END_TO_END_TIME: 139, Number of samples: 11","totalSum":139,"name":"TASK_END_TO_END_TIME","value":139},{"numberOfSamples":11,"description":"Time in milliseconds tasks took to execute","currentMetricState":"TASK_EXECUTION_TIME: 134, Number of samples: 11","totalSum":134,"name":"TASK_EXECUTION_TIME","value":134},{"numberOfSamples":99,"description":"Counter for number of mutation sql statements","currentMetricState":"MUTATION_SQL_COUNTER: 99, Number of samples: 99","totalSum":99,"name":"MUTATION_SQL_COUNTER","value":99},{"numberOfSamples":7,"description":"Counter for number of sql queries","currentMetricState":"SELECT_SQL_COUNTER: 7, Number of samples: 7","totalSum":7,"name":"SELECT_SQL_COUNTER","value":7},{"numberOfSamples":11,"description":"Counter for number of tasks submitted to the thread pool executor","currentMetricState":"TASK_EXECUTED_COUNTER: 11, Number of samples: 11","totalSum":11,"name":"TASK_EXECUTED_COUNTER","value":11},{"numberOfSamples":0,"description":"Counter for number of tasks that were rejected by the thread pool executor","currentMetricState":"TASK_REJECTED_COUNTER: 0, Number of samples: 0","totalSum":0,"name":"TASK_REJECTED_COUNTER","value":0},{"numberOfSamples":0,"description":"Number of times query timed out","currentMetricState":"QUERY_TIMEOUT_COUNTER: 0, Number of samples: 0","totalSum":0,"name":"QUERY_TIMEOUT_COUNTER","value":0},{"numberOfSamples":0,"description":"Number of times query failed","currentMetricState":"QUERY_FAILED_COUNTER: 0, Number of samples: 0","totalSum":0,"name":"QUERY_FAILED_COUNTER","value":0},{"numberOfSamples":0,"description":"Number of spool files created","currentMetricState":"SPOOL_FILE_COUNTER: 0, Number of samples: 0","totalSum":0,"name":"SPOOL_FILE_COUNTER","value":0},{"numberOfSamples":14,"description":"Number of open phoenix connections","currentMetricState":"OPEN_PHOENIX_CONNECTIONS_COUNTER: 0, Number of samples: 14","totalSum":0,"name":"OPEN_PHOENIX_CONNECTIONS_COUNTER","value":0},{"numberOfSamples":1,"description":"Number of ConnectionQueryServicesImpl instantiated","currentMetricState":"QUERY_SERVICES_COUNTER: 1, Number of samples: 1","totalSum":1,"name":"QUERY_SERVICES_COUNTER","value":1},{"numberOfSamples":1,"description":"Number of HConnections created by phoenix driver","currentMetricState":"HCONNECTIONS_COUNTER: 1, Number of samples: 1","totalSum":1,"name":"HCONNECTIONS_COUNTER","value":1}]}


Global Metrics is also exposed via JMX. ( screen shot attached)
global_metrics

phoenix-core/src/main/java/org/apache/phoenix/query/QueryServicesOptions.java
```diff
@@ -272,6 +278,11 @@
     public static final String DEFAULT_IMMUTABLE_STORAGE_SCHEME = ImmutableStorageScheme.SINGLE_CELL_ARRAY_WITH_OFFSETS.toString();
     public static final String DEFAULT_MULTITENANT_IMMUTABLE_STORAGE_SCHEME = ImmutableStorageScheme.ONE_CELL_PER_COLUMN.toString();
 
+    public static final Integer DEFAULT_PHOENIX_PQS_REPORTING_INTERVAL_MS = 10000; // 10 sec
+    public static final String DEFAULT_PHOENIX_PQS_TYPE_OF_SINK = "file";
+    public static final boolean DEFAULT_PHOENIX_QUERY_SERVER_METRICS = true;
```
joshelser: I think this should be off by default and something have to opt-in to enable.

phoenix-queryserver/src/main/java/org/apache/phoenix/queryserver/metrics/sink/PqsFileSink.java
joshelser: Ping: this still needs to be addressed.

e.g. what happens if /tmp is nonexistent or non-writable?

phoenix-queryserver/src/main/java/org/apache/phoenix/queryserver/metrics/PqsMetricsSystem.java
```diff
@@ -0,0 +1,105 @@
...
+import static org.apache.phoenix.query.QueryServices.PHOENIX_PQS_METRIC_REPORTING_INTERVAL_MS;
+import static org.apache.phoenix.query.QueryServices.PHOENIX_PQS_TYPE_OF_SINK;
+
+
+public class PqsMetricsSystem {
+
+    public final static String statementReadMetrics = "Statement-RequestReadMetrics";
+    public final static String overAllReadRequestMetrics = "Statement-OverAllReadRequestMetrics";
+    public final static String connectionWriteMetricsForMutations = "Connection-WriteMetricsForMutations";
+    public final static String connectionReadMetricsForMutations = "Connection-ReadMetricsForMutations";
```
joshelser: These names leave a bit to be desired for me. Reading them, I don't really know what information they're going to contain. How about instead..

* QueryReadMetrics
* GlobalQueryReadMetrics
* ConnectionMutationWriteMetrics
* ConnectionMutationReadMetrics

I don't have a good understanding of what these metrics contain; it just seemed like you copied the names from the internal Phoenix methods. If we're exposing them to users, they need to be self-explanatory (or have documentation explaining them -- maybe we have that?).

phoenix-queryserver/src/main/java/org/apache/phoenix/queryserver/metrics/PqsMetricsSystem.java
```diff
@@ -0,0 +1,105 @@
...
+            }
+
+        }
+     }
+
+    public static PqsSink getSinkObject(String typeOfSink,String filename){
+        PqsSink pqsSink;
+        switch(typeOfSink.toLowerCase()) {
+            case "file":
+                pqsSink = new PqsFileSink(filename);
```
rahulsIOT: Hadoop metrics using concept of source and sink and ties them together with metric system. If you use sink , you need a source of gauges, meters, counters etc as source. But the only interface Phoenix driver exposes is via static methods in PhoenixRuntime, which provides aggregate values for various metrics. So, it is difficult to use Hadoop metrics or Codahale.

phoenix-queryserver/src/main/java/org/apache/phoenix/queryserver/metrics/sink/PqsFileSink.java
rahulsIOT: Not sure which version you are looking at but there is a finally statement below. 

finally {
            if (writer == null) {
                writer = System.out;
            }

If exception is thrown, then writer will be null and we set writer to be system.out.

phoenix-core/src/main/java/org/apache/phoenix/query/QueryServicesOptions.java
rahulsIOT: made the metrics collection to be off by default.

phoenix-queryserver/src/main/java/org/apache/phoenix/queryserver/metrics/sink/PqsFileSink.java
joshelser: Ah, I didn't look down below.

Follow-on, why do you want this to fail? This could lead to an operator who thinks they're collecting metrics to not actually get them. It seems reasonable to me that, if the user configures metrics with an invalid/unwritable path, they should see an error propagated.

phoenix-queryserver/src/main/java/org/apache/phoenix/queryserver/metrics/PqsMetricsSystem.java
joshelser: > But the only interface Phoenix driver exposes is via static methods in PhoenixRuntime, which provides aggregate values for various metrics. So, it is difficult to use Hadoop metrics or Codahale.

I can understand why you can't use it directly, but I'm not seeing why you couldn't write a layer to convert from the Phoenix internal representation to the metrics system of choice (a shim).

With Phoenix, I don't think we should be in the business of writing our own metrics reporting code. It's debt for the project as well as more "tribal knowledge" that our users have to learn to access these metrics instead of metrics automatically showing up in aggregated stores (such as Ambari Metrics System via Hadoop Metrics2).

Again, sorry I did not speak up about this when you put up a design doc. Curious what others think.

karanmehta93:
We decided not to follow this approach, Hence will be removing the reference to this PR from JIRA.

PR #244, opened 2017-05-05
PHOENIX-3824 Mutable Index partial rebuild adds more than one index r…

vincentpoon:
…ow for updated data row


stoty:
Already merged.

PR #248, opened 2017-05-15
Phoenix 3534

churrodog:
I'll squash these down after a review


JamesRTaylor:
I'll check it out, but in the meantime would you mind amending your commit message from "Phoenix 3534" to "PHOENIX-3534 Support multi region SYSTEM.CATALOG table" - that way comments on the pull will become JIRA comments.

churrodog:
@twdsilva would you mind taking a look as well?

phoenix-core/src/main/java/org/apache/phoenix/schema/PTableImpl.java
```diff
@@ -500,8 +510,20 @@ private void init(PName tenantId, PName schemaName, PName tableName, PTableType
...
         }
-        for (PColumn column : columns) {
-            allColumns[column.getPosition()] = column;
+        // Must do this as with the new method of storing diffs, we just care about ordinal position
+        // relative order and not the true ordinal value itself.
+        List<PColumn> sortedColumns = Lists.newArrayList(columns);
+        Collections.sort(sortedColumns, new Comparator<PColumn>() {
+            @Override
+            public int compare(PColumn o1, PColumn o2) {
+                return Integer.valueOf(o1.getPosition()).compareTo(o2.getPosition());
```
JamesRTaylor: What will prevent overlapping ordinals between a base table and derived views?

phoenix-core/src/it/java/org/apache/phoenix/end2end/AlterTableWithViewsIT.java
```diff
@@ -107,12 +107,14 @@ public void testAddNewColumnsToBaseTableWithViews() throws Exception {
             assertTableDefinition(conn, tableName, PTableType.TABLE, null, 0, 3, QueryConstants.BASE_TABLE_BASE_COLUMN_COUNT, "ID", "COL1", "COL2");
             
             viewConn.createStatement().execute("CREATE VIEW " + viewOfTable + " ( VIEW_COL1 DECIMAL(10,2), VIEW_COL2 VARCHAR ) AS SELECT * FROM " + tableName);
-            assertTableDefinition(conn, viewOfTable, PTableType.VIEW, tableName, 0, 5, 3, "ID", "COL1", "COL2", "VIEW_COL1", "VIEW_COL2");
+            assertTableDefinition(conn, viewOfTable, PTableType.VIEW, tableName, 0, 5, 3, "VIEW_COL1", "VIEW_COL2");
```
JamesRTaylor: Are these test changes due to only storing diffs now? Would be nice to keep the check for "this is what the combined table should look like" in addition to checking that only the diffs are actually stored. WDYT?

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/MetaDataEndpointImpl.java
```diff
@@ -2324,96 +2488,97 @@ else if (pkCount <= COLUMN_NAME_INDEX
...
-                        PSmallint.INSTANCE.getCodec().encodeShort(newKeySeq, keySeqBytes, 0);
-                        viewColumnPut.add(PhoenixDatabaseMetaData.TABLE_FAMILY_BYTES,
-                                PhoenixDatabaseMetaData.KEY_SEQ_BYTES, keySeqBytes);
-                        addMutationsForAddingPkColsToViewIndexes(mutationsForAddingColumnsToViews, clientTimeStamp, view,
-                                deltaNumPkColsSoFar, columnName, viewColumnPut);
-                    }
-                    columnsAddedToView++;
-                    columnsAddedToBaseTable++;
                 }
+//                else {
```
JamesRTaylor: You'll likely get to this later, but let's remove the commented out code when you're ready.

phoenix-core/src/main/java/org/apache/phoenix/schema/PColumnImpl.java
```diff
@@ -159,15 +199,6 @@ public SortOrder getSortOrder() {
     public String toString() {
         return (familyName == null ? "" : familyName.toString() + QueryConstants.NAME_SEPARATOR) + name.toString();
     }
-
```
JamesRTaylor: Should this be removed and if so why?

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/MetaDataEndpointImpl.java
```diff
@@ -381,7 +383,8 @@
             IS_VIEW_REFERENCED_KV,
             COLUMN_DEF_KV,
             IS_ROW_TIMESTAMP_KV,
-            COLUMN_QUALIFIER_KV
+            COLUMN_QUALIFIER_KV,
+            DROPPED_COLUMN_KV
```
JamesRTaylor: I don't think you need this DROPPED_COLUMN_KV since you're not reading this when the column is read (but instead when the linking rows are read). If that's right, let's remove it.

JamesRTaylor:
This is looking really good, @churrodog. For testing purposes, I think it'd be really good if we split the SYSTEM.CATALOG table at points that'll force it to span multiple regions during the running of tests like AlterTableWithViewsIT, TenantSpecificTablesDDLIT, AlterTableIT, QueryDatabaseMetaDataIT, and maybe more.

Any full table scans let in the CREATE or DROP code path?

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/TableViewFinderResult.java
```diff
@@ -0,0 +1,50 @@
...
+
+import com.google.common.collect.Lists;
+import org.apache.hadoop.hbase.client.Result;
+
+import java.util.List;
+
+/**
+ * Certain operations, such as DROP TABLE are not allowed if there a table has child views. This class wraps the
+ * Results of a scanning the Phoenix Metadata for child views for a specific table and stores an additional flag for
+ * whether whether SYSTEM.CATALOG has split across multiple regions.
```
twdsilva: modify this comment

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/MetaDataEndpointImpl.java
```diff
@@ -535,19 +539,115 @@ public void getTable(RpcController controller, GetTableRequest request,
...
+                allColumns.add(pColumn);
+            }
+
+        }
+        // now go up from child to parent all the way to the base table:
+        for (int i = 0; i < listOBytes.size(); i++) {
+            byte[] tableInQuestion = listOBytes.get(i);
+            PTable pTable = this.doGetTable(tableInQuestion, timestamp);
+            if (pTable == null) {
+                throw new TableNotFoundException("fill in with valuable info");
```
twdsilva: fix the exception message

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/ViewFinder.java
```diff
@@ -0,0 +1,241 @@
...
+                resultTuple.getKey(ptr);
+                results.add(result);
+            }
+            return new TableViewFinderResult(results);
+        } finally {
+            scanner.close();
+        }
+    }
+
+    static Graph<TableInfo> findOrphans(Table systemCatalog, long timestamp) throws IOException {
```
twdsilva: is this method used somewhere?

phoenix-core/src/it/java/org/apache/phoenix/end2end/AlterTableWithViewsIT.java
churrodog: Sure we can do that, I think we will need to modify a ton of tests to make this happen.  Its not only done here. 

Another thing is that there are a few tests where the assumptions of how the system works has changed.  A few examples are tests like this: 

ViewIT.testDisallowDropOfColumnOnParentTable()
ViewIT.testDisallowDropOfReferencedColumn()
QueryCompiler.testDuplicatePKColumn() 
(this one because we have to support both the diff based approach and the current approach for our migration step, we remove duplicate columns on read, thus we don't prevent this from happening). 

There are quite a few more, we can maybe discuss if these tests are no longer valid or how we can change them to make them meaningful.

JamesRTaylor:
Looks like there are still conflicts, @churrodog. Would you mind amending your commit message to "PHOENIX-3534 Support multi region SYSTEM.CATALOG table" so that the JIRA is linked to the PR?

stoty:
Already merged.

PR #257, opened 2017-06-07
PHOENIX-3917 Change to calculate query projection size

bijugs:
Change to perform the size estimation of query row projector after all the conditions to add ``column families`` to the query ``scan`` object are done.


bijugs:
Closing the PR since the Jira ticket got resolved.

PR #261, opened 2017-06-15
PHOENIX-3944 ReadOnlyTableException occurs when we map Phoenix view t…

brfrn169:
…o an existing HBase table with Namespace Mapping enabled

### DIFF
--- a/phoenix-core/src/it/java/org/apache/phoenix/end2end/CreateTableIT.java
+++ b/phoenix-core/src/it/java/org/apache/phoenix/end2end/CreateTableIT.java
@@ -49,10 +49,7 @@ import org.apache.phoenix.schema.PTable.QualifierEncodingScheme;
 import org.apache.phoenix.schema.PTableKey;
 import org.apache.phoenix.schema.SchemaNotFoundException;
 import org.apache.phoenix.schema.TableAlreadyExistsException;
-import org.apache.phoenix.util.PhoenixRuntime;
-import org.apache.phoenix.util.PropertiesUtil;
-import org.apache.phoenix.util.SchemaUtil;
-import org.apache.phoenix.util.TestUtil;
+import org.apache.phoenix.util.*;
 import org.junit.Assert;
 import org.junit.Test;
 
@@ -707,5 +704,60 @@ public class CreateTableIT extends BaseClientManagedTimeIT {
           connection.close();
         }
       }
+    }
+
+    @Test
+    public void testCreateTableWithNamespaceMappingEnabled() throws Exception {
+        final String NS = "NS_" + generateUniqueName();
+        final String TBL = "TBL_" + generateUniqueName();
+        final String CF = "CF";
+
+        Properties props = new Properties();
+        props.setProperty(QueryServices.IS_NAMESPACE_MAPPING_ENABLED, Boolean.TRUE.toString());
+
+        try (Connection conn = DriverManager.getConnection(getUrl(), props)) {
+            conn.createStatement().execute("CREATE SCHEMA " + NS);
+
+            // test for a table that is in non-default schema
+            {
+                String table = NS + "." + TBL;
+                conn.createStatement().execute(
+                    "CREATE TABLE " + table + " (PK VARCHAR PRIMARY KEY, " + CF + ".COL VARCHAR)");
+
+                assertTrue(QueryUtil.getExplainPlan(
+                    conn.createStatement().executeQuery("explain select * from " + table))
+                    .contains(NS + ":" + TBL));
+
+                conn.createStatement().execute("DROP TABLE " + table);
+            }
+
+            // test for a table whose name contains a dot (e.g. "AAA.BBB") in default schema
+            {
+                String table = "\"" + NS + "." + TBL + "\"";
+                conn.createStatement().execute(
+                    "CREATE TABLE " + table + " (PK VARCHAR PRIMARY KEY, " + CF + ".COL VARCHAR)");
+
+                assertTrue(QueryUtil.getExplainPlan(
+                    conn.createStatement().executeQuery("explain select * from " + table))
+                    .contains(NS + "." + TBL));
+
+                conn.createStatement().execute("DROP TABLE " + table);
+            }
+
+            // test for a view whose name contains a dot (e.g. "AAA.BBB") in non-default schema
+            {
+                String table = NS + ".\"" + NS + "." + TBL + "\"";
+                conn.createStatement().execute(
+                    "CREATE TABLE " + table + " (PK VARCHAR PRIMARY KEY, " + CF + ".COL VARCHAR)");
+
+                assertTrue(QueryUtil.getExplainPlan(
+                    conn.createStatement().executeQuery("explain select * from " + table))
+                    .contains(NS + ":" + NS + "." + TBL));
+
+                conn.createStatement().execute("DROP TABLE " + table);
+            }
+
+            conn.createStatement().execute("DROP SCHEMA " + NS);
+        }
     }
 }
--- a/phoenix-core/src/it/java/org/apache/phoenix/end2end/ViewIT.java
+++ b/phoenix-core/src/it/java/org/apache/phoenix/end2end/ViewIT.java
@@ -36,6 +36,9 @@ import java.sql.SQLException;
 import java.util.List;
 import java.util.Properties;
 
+import org.apache.hadoop.hbase.HColumnDescriptor;
+import org.apache.hadoop.hbase.HTableDescriptor;
+import org.apache.hadoop.hbase.TableName;
 import org.apache.hadoop.hbase.client.HBaseAdmin;
 import org.apache.phoenix.compile.QueryPlan;
 import org.apache.phoenix.exception.SQLExceptionCode;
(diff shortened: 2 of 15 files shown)


phoenix-core/src/main/java/org/apache/phoenix/compile/FromCompiler.java
```diff
@@ -196,7 +196,8 @@ public static ColumnResolver getResolverForCreation(final CreateTableStatement s
                     if (htable != null) Closeables.closeQuietly(htable);
                 }
                 tableNode = NamedTableNode.create(null, baseTable, statement.getColumnDefs());
-                return new SingleTableColumnResolver(connection, tableNode, e.getTimeStamp(), new HashMap<String, UDFParseNode>(1), false);
+                boolean isNamespaceMapped = SchemaUtil.isNamespaceMappingEnabled(statement.getTableType(), connection.getQueryServices().getProps());
```
brfrn169: Originally the SingleTableColumnResolver constructor argument "isNamespaceMapped" is hard-coded as false, but we should set it from props.

phoenix-core/src/main/java/org/apache/phoenix/schema/MetaDataClient.java
```diff
@@ -2135,7 +2135,7 @@ private PTable createTableInternal(CreateTableStatement statement, byte[][] spli
                 // Upsert physical name for mapped view only if the full physical table name is different than the full table name
                 // Otherwise, we end up with a self-referencing link and then cannot ever drop the view.
                 if (viewType != ViewType.MAPPED
-                        || !physicalNames.get(0).getString().equals(SchemaUtil.getTableName(schemaName, tableName))) {
+                        || !physicalNames.get(0).getString().equals(SchemaUtil.getPhysicalName(schemaName, tableName))) {
```
brfrn169: Originally, for some reason this is comparing between physicalNames (formatted as "<namespace>:<table>") and the result of getTableName() which is formatted as "<schema>.<table>". However, the format of both values being compared should be "<namespace>:<table>".

phoenix-core/src/it/java/org/apache/phoenix/end2end/ViewIT.java
```diff
@@ -710,6 +714,27 @@ public void testCreateViewWithUpdateCacheFrquency() throws Exception {
...
+
+        HBaseAdmin admin = conn.unwrap(PhoenixConnection.class).getQueryServices().getAdmin();
+        admin.createNamespace(NamespaceDescriptor.create(ns).build());
+        HTableDescriptor desc = new HTableDescriptor(TableName.valueOf(ns, table));
+        desc.addFamily(new HColumnDescriptor(fam));
+        admin.createTable(desc);
+
+        conn.createStatement().execute("CREATE SCHEMA " + ns);
+        conn.createStatement().execute("CREATE VIEW " + ns + "." + table + " "
+            + "(PK VARCHAR PRIMARY KEY, " + fam + ".COL VARCHAR)");
```
ankitsinghal: can you update a test to include a case where user has "NS.TBL" in default namespace and he wants to map a view.(\"NS.TBL\")(for backward compatibility).
```
desc = new HTableDescriptor(TableName.valueOf(ns+"."+table));
        desc.addFamily(new HColumnDescriptor(fam));
        admin.createTable(desc);

conn.createStatement().execute("CREATE VIEW \"" + ns + "." + table + "\""
                + "(PK VARCHAR PRIMARY KEY, " + fam + ".COL VARCHAR)");
```

phoenix-core/src/main/java/org/apache/phoenix/compile/FromCompiler.java
ankitsinghal: To handle view mapped \"NS.TBL\"(without schema)

```
-                byte[] fullTableName = SchemaUtil.getPhysicalName(
-                        SchemaUtil.getTableNameAsBytes(baseTable.getSchemaName(), baseTable.getTableName()),
-                        connection.getQueryServices().getProps()).getName();
+                byte[] fullTableName = SchemaUtil
+                        .getPhysicalName(Bytes.toBytes(baseTable.toString()),
+                                connection.getQueryServices().getProps())
+                        .getName();
```

phoenix-core/src/main/java/org/apache/phoenix/schema/MetaDataClient.java
ankitsinghal: can you include isNamespaceMapped as table name with schema can also be in default namespace.
```
-                        || !physicalNames.get(0).getString().equals(SchemaUtil.getTableName(schemaName, tableName))) {
+                        || !physicalNames.get(0).getString().equals(SchemaUtil.getPhysicalTableName(
+                                SchemaUtil.getTableNameAsBytes(schemaName, tableName), isNamespaceMapped).getNameAsString())) {
```

phoenix-core/src/main/java/org/apache/phoenix/util/SchemaUtil.java
```diff
@@ -302,6 +302,10 @@ public static String getTableName(byte[] schemaName, byte[] tableName) {
         return getName(schemaName, tableName);
     }
 
+    public static String getPhysicalName(String schemaName, String tableName) {
+        return schemaName + QueryConstants.NAMESPACE_SEPARATOR + tableName;
+    }
+
     public static String getColumnDisplayName(byte[] cf, byte[] cq) {
```
ankitsinghal: we probably don't need this method after above changes.

phoenix-core/src/main/java/org/apache/phoenix/query/ConnectionQueryServicesImpl.java
```diff
@@ -1420,7 +1420,7 @@ public MetaDataMutationResult createTable(final List<Mutation> tableMetaData, fi
         byte[] tenantIdBytes = rowKeyMetadata[PhoenixDatabaseMetaData.TENANT_ID_INDEX];
         byte[] schemaBytes = rowKeyMetadata[PhoenixDatabaseMetaData.SCHEMA_NAME_INDEX];
         byte[] tableBytes = rowKeyMetadata[PhoenixDatabaseMetaData.TABLE_NAME_INDEX];
-        byte[] tableName = physicalTableName != null ? physicalTableName : SchemaUtil.getTableNameAsBytes(schemaBytes, tableBytes);
+        byte[] tableName = SchemaUtil.getTableNameAsBytes(schemaBytes, tableBytes);
```
brfrn169: I think this is correct because this variable is used as the first argument of ensureTableCreated method and it doesn't expect a physical table name but a phoenix table name.
Also,  this variable is used by the following code but it expects a phoenix table name as well.

```
            ensureViewIndexTableCreated(
                    SchemaUtil.getPhysicalHBaseTableName(tableName, isNamespaceMapped, tableType).getBytes(),
                    tableProps, familiesPlusDefault, MetaDataUtil.isSalted(m, kvBuilder, ptr) ? splits : null,
                            MetaDataUtil.getClientTimeStamp(m), isNamespaceMapped);
```

phoenix-core/src/main/java/org/apache/phoenix/util/SchemaUtil.java
```diff
@@ -1016,7 +1017,8 @@ public static TableName getPhysicalTableName(byte[] fullTableName, boolean isNam
     public static PName getPhysicalHBaseTableName(String tableName, boolean isNamespaceMapped, PTableType type) {
         if (!isNamespaceMapped) { return PNameFactory.newName(tableName); }
         return PNameFactory
-                .newName(tableName.replace(QueryConstants.NAME_SEPARATOR, QueryConstants.NAMESPACE_SEPARATOR));
+                .newName(StringUtils.replaceOnce(
```
brfrn169: If we want to create the view NS."AAA.BBB", the original logic returns NS:AAA:BBB as a physical hbase table name. We need the name NS:AAA.BBB .

phoenix-core/src/main/java/org/apache/phoenix/compile/FromCompiler.java
brfrn169: I think we don't need this change. So I didn't include this change in the new commit. What do you think?

ankitsinghal: I think it is necessary. How about asserting in the test that tables are mapped properly. You can like this.
assert(QueryUtil.getExplainPlan(conn.createStatement().executeQuery("explain select * from \""+ ns + "." + table+"\"")).contains(ns+"."+table));

phoenix-core/src/main/java/org/apache/phoenix/query/ConnectionQueryServicesImpl.java
ankitsinghal: Why we need this change? I think it should be physical table name, we need it to verify meta data when views created on phoenix table (create view v as select * from table) in ensureTableCreated(tableName, tableType, tableProps, families, splits, true, isNamespaceMapped);

phoenix-core/src/main/java/org/apache/phoenix/util/SchemaUtil.java
```diff
@@ -983,14 +983,21 @@ public static boolean hasRowTimestampColumn(PTable table) {
         return SchemaUtil.getTableKey(null, schemaName, MetaDataClient.EMPTY_TABLE);
     }
 
-    public static PName getPhysicalHBaseTableName(PName pName, boolean isNamespaceMapped, PTableType type) {
-        return getPhysicalHBaseTableName(pName.toString(), isNamespaceMapped, type);
+    public static PName getPhysicalHBaseTableName(PName schemaName, PName tableName, boolean isNamespaceMapped) {
```
brfrn169: I added "schemaName" as an argument to this method.
I think we need both schemaName and tableName to handle a table like "AAA.BBB" in default schema correctly, because if we have only the table name "AAA.BBB", we can't distinguish between "AAA.BBB" in the default schema and "BBB" in "AAA" schema.

Also we don't need the "type" argument, so I deleted it.

phoenix-core/src/it/java/org/apache/phoenix/end2end/CreateTableIT.java
```diff
@@ -708,4 +705,52 @@ public void testCreateTableWithUpdateCacheFrequencyAttrib() throws Exception {
         }
       }
     }
+
+    @Test
```
brfrn169: I also added the test for createTable with Namespace Mapping enabled. This test fails against the original logic.

phoenix-core/src/main/java/org/apache/phoenix/compile/FromCompiler.java
```diff
@@ -184,9 +184,9 @@ public static ColumnResolver getResolverForCreation(final CreateTableStatement s
             // A tenant-specific connection may not create a mapped VIEW.
             if (connection.getTenantId() == null && statement.getTableType() == PTableType.VIEW) {
                 ConnectionQueryServices services = connection.getQueryServices();
-                byte[] fullTableName = SchemaUtil.getPhysicalName(
-                        SchemaUtil.getTableNameAsBytes(baseTable.getSchemaName(), baseTable.getTableName()),
-                        connection.getQueryServices().getProps()).getName();
+                boolean isNamespaceMapped = SchemaUtil.isNamespaceMappingEnabled(statement.getTableType(), connection.getQueryServices().getProps());
+                byte[] fullTableName = SchemaUtil.getPhysicalHBaseTableName(
```
brfrn169: As @ankitsinghal mentioned in the previous comment, I have changed this to handle "AAA.BBB"(without schema) correctly using SchemaUtil.getPhysicalHBaseTableName() (which I've further modified in this commit).

phoenix-core/src/main/java/org/apache/phoenix/schema/PTableImpl.java
```diff
@@ -1184,8 +1184,11 @@ public synchronized boolean getIndexMaintainers(ImmutableBytesWritable ptr, Phoe
 
     @Override
     public PName getPhysicalName() {
-        return SchemaUtil.getPhysicalHBaseTableName(physicalNames.isEmpty() ? getName() : physicalNames.get(0),
-                isNamespaceMapped, type);
+        if (physicalNames.isEmpty()) {
```
brfrn169: I changed this logic to handle "AAA.BBB"(without schema) correctly.

phoenix-core/src/main/java/org/apache/phoenix/util/SchemaUtil.java
```diff
@@ -983,14 +983,21 @@ public static boolean hasRowTimestampColumn(PTable table) {
...
+            schemaName == null ? null : schemaName.toString(), tableName.toString(), isNamespaceMapped);
     }
 
-    public static PName getPhysicalHBaseTableName(byte[] tableName, boolean isNamespaceMapped, PTableType type) {
-        return getPhysicalHBaseTableName(Bytes.toString(tableName), isNamespaceMapped, type);
+    public static PName getPhysicalHBaseTableName(byte[] schemaName, byte[] tableName, boolean isNamespaceMapped) {
+        return getPhysicalHBaseTableName(Bytes.toString(schemaName), Bytes.toString(tableName), isNamespaceMapped);
     }
 
+    /**
```
brfrn169: As per this comment, the 4 methods (getPhysicalTableName, getPhysicalName) can return an unexpected value. 
I don't believe this situation has been seen in the wild as-of-yet, however users should be aware of this behaviour.

brfrn169:
Hi @ankitsinghal , thank you for reviewing. I've changed my pull request according to your feedback. Could you please review this pull request again?

ankitsinghal:
Thanks @brfrn169, Now the changes look great. Let me commit this today or tomorrow.

brfrn169:
This patch was committed.

PR #262, opened 2017-06-16
PHOENIX 153 implement TABLESAMPLE clause

aertoria:
table sampling on each Table basis (at the 'FROM' part of the query). Sample size decided by the input sampling rate applies on Primary Key's frequency.

Syntax: 
`select name from person SAMPLE(0.10) where name='ethan'`

++
++Belows are SUPPORTED
++
===BASE CASE====
select * from Person;
select * from PERSON TABLESAMPLE(45);
select * from PERSON TABLESAMPLE (45);

===WHERE CLAUSE====
select * from PERSON where ADDRESS = 'CA' OR name>'tina3';
select * from PERSON TABLESAMPLE (49) where ADDRESS = 'CA' OR name>'tina3';
select * from PERSON TABLESAMPLE (49) where ADDRESS = 'CA' OR name>'tina3' ORDER BY ADDRESS LIMIT 2;

===Wired Table===
select * from LOCAL_ADDRESS TABLESAMPLE (79);
select * from SYSTEM.STATS TABLESAMPLE (41);

===CORNER CASE===
select * from PERSON TABLESAMPLE (0);
select * from PERSON TABLESAMPLE (100.0);
select * from PERSON TABLESAMPLE (145.99);  (out of boundary exception)
select * from PERSON TABLESAMPLE ko; (syntax error exception)

===AGGREGATION===
select count(*) from PERSON TABLESAMPLE (49) LIMIT 2
select count(*) from (select NAME from PERSON TABLESAMPLE (49) limit 20)

===SUB QUERY===
select * from (select /*+NO_INDEX*/ * from PERSON tablesample (10) where Name > 'tina10') where ADDRESS = 'CA'

===JOINS===
select * from PERSON1, PERSON2 tablesample (70) where PERSON1.Name = PERSON2.NAME
select /*+NO_INDEX*/ count(*) from PERSON tableSample (19), US_POPULATION tableSample (28) where PERSON.Name > US_POPULATION.STATE

===QUERY being OPTMIZED===
select * from PERSON tablesample (80) (may go to IDX_ADDRESS_PERSON index table, but table sampling rate carry on)

===INSERT SELECT====
upsert into personbig(ID, ADDRESS) select id, address from personbig tablesample (1);

### DIFF
--- a/phoenix-core/src/it/java/org/apache/phoenix/monitoring/PhoenixMetricsIT.java
+++ b/phoenix-core/src/it/java/org/apache/phoenix/monitoring/PhoenixMetricsIT.java
@@ -29,6 +29,19 @@ import static org.apache.phoenix.monitoring.GlobalClientMetrics.GLOBAL_SELECT_SQ
 import static org.apache.phoenix.monitoring.GlobalClientMetrics.GLOBAL_SPOOL_FILE_COUNTER;
 import static org.apache.phoenix.monitoring.GlobalClientMetrics.GLOBAL_TASK_END_TO_END_TIME;
 import static org.apache.phoenix.monitoring.GlobalClientMetrics.GLOBAL_TASK_EXECUTION_TIME;
+
+import static org.apache.phoenix.monitoring.GlobalClientMetrics.GLOBAL_HBASE_COUNT_RPC_CALLS;
+import static org.apache.phoenix.monitoring.GlobalClientMetrics.GLOBAL_HBASE_COUNT_REMOTE_RPC_CALLS;
+import static org.apache.phoenix.monitoring.GlobalClientMetrics.GLOBAL_HBASE_COUNT_MILLS_BETWEEN_NEXTS;
+import static org.apache.phoenix.monitoring.GlobalClientMetrics.GLOBAL_HBASE_COUNT_NOT_SERVING_REGION_EXCEPTION;
+import static org.apache.phoenix.monitoring.GlobalClientMetrics.GLOBAL_HBASE_COUNT_BYTES_REGION_SERVER_RESULTS;
+import static org.apache.phoenix.monitoring.GlobalClientMetrics.GLOBAL_HBASE_COUNT_BYTES_IN_REMOTE_RESULTS;
+import static org.apache.phoenix.monitoring.GlobalClientMetrics.GLOBAL_HBASE_COUNT_SCANNED_REGIONS;
+import static org.apache.phoenix.monitoring.GlobalClientMetrics.GLOBAL_HBASE_COUNT_RPC_RETRIES;
+import static org.apache.phoenix.monitoring.GlobalClientMetrics.GLOBAL_HBASE_COUNT_REMOTE_RPC_RETRIES;
+import static org.apache.phoenix.monitoring.GlobalClientMetrics.GLOBAL_HBASE_COUNT_ROWS_SCANNED;
+import static org.apache.phoenix.monitoring.GlobalClientMetrics.GLOBAL_HBASE_COUNT_ROWS_FILTERED;
+
 import static org.apache.phoenix.monitoring.MetricType.MEMORY_CHUNK_BYTES;
 import static org.apache.phoenix.monitoring.MetricType.SCAN_BYTES;
 import static org.apache.phoenix.monitoring.MetricType.TASK_EXECUTED_COUNTER;
@@ -136,6 +149,11 @@ public class PhoenixMetricsIT extends BaseUniqueNamesOwnClusterIT {
         assertTrue(GLOBAL_QUERY_TIME.getMetric().getTotalSum() > 0);
         assertTrue(GLOBAL_TASK_END_TO_END_TIME.getMetric().getTotalSum() > 0);
         assertTrue(GLOBAL_TASK_EXECUTION_TIME.getMetric().getTotalSum() > 0);
+
+        assertTrue(GLOBAL_HBASE_COUNT_RPC_CALLS.getMetric().getTotalSum() > 0);
+        assertTrue(GLOBAL_HBASE_COUNT_MILLS_BETWEEN_NEXTS.getMetric().getTotalSum() > 0);
+        assertTrue(GLOBAL_HBASE_COUNT_BYTES_REGION_SERVER_RESULTS.getMetric().getTotalSum() > 0);
+        assertTrue(GLOBAL_HBASE_COUNT_SCANNED_REGIONS.getMetric().getTotalSum() > 0);
     }
 
     @Test
@@ -181,6 +199,11 @@ public class PhoenixMetricsIT extends BaseUniqueNamesOwnClusterIT {
         assertEquals(0, GLOBAL_FAILED_QUERY_COUNTER.getMetric().getTotalSum());
         assertEquals(0, GLOBAL_SPOOL_FILE_COUNTER.getMetric().getTotalSum());
         assertEquals(0, GLOBAL_MUTATION_BATCH_FAILED_COUNT.getMetric().getTotalSum());
+
+        assertTrue(GLOBAL_HBASE_COUNT_RPC_CALLS.getMetric().getTotalSum() > 0);
+        assertTrue(GLOBAL_HBASE_COUNT_MILLS_BETWEEN_NEXTS.getMetric().getTotalSum() > 0);
+        assertTrue(GLOBAL_HBASE_COUNT_BYTES_REGION_SERVER_RESULTS.getMetric().getTotalSum() > 0);
+        assertTrue(GLOBAL_HBASE_COUNT_SCANNED_REGIONS.getMetric().getTotalSum() > 0);
     }
 
     private static void resetGlobalMetrics() {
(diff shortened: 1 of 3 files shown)


aertoria:
@JamesRTaylor @gjacoby126  @samarthjain

phoenix-core/src/it/java/org/apache/phoenix/end2end/QueryWithTableSampleIT.java
```diff
@@ -0,0 +1,261 @@
...
+            assertTrue(rs.next());
+            assertEquals(31, rs.getInt(1));
+        } finally {
+            conn.close();
+        }
+    }
+    
+    final private static void inspect(final ResultSet rs) throws SQLException{
+    	while(rs.next()){
+    		System.out.println(rs.getInt(1)+","+rs.getInt(2));
```
JamesRTaylor: Minor nit: we don't write to system.out for testing (though I understand the value while debugging). Perhaps simplest to just switch to logging instead.

phoenix-core/src/main/java/org/apache/phoenix/compile/QueryCompiler.java
```diff
@@ -539,6 +539,7 @@ protected QueryPlan compileSingleFlatQuery(StatementContext context, SelectState
         if (table.getViewStatement() != null) {
             viewWhere = new SQLParser(table.getViewStatement()).parseQuery().getWhere();
         }
+
```
JamesRTaylor: Please revert changes to this file as there are only whitespace changes.

phoenix-core/src/main/java/org/apache/phoenix/parse/SelectStatement.java
```diff
@@ -267,6 +267,12 @@ public LimitNode getLimit() {
     }
     
     @Override
+    public Double getTableSamplingRate(){
+    	if(fromTable==null || !(fromTable instanceof ConcreteTableNode)) return null;
+    	return ((ConcreteTableNode)fromTable).getTableSamplingRate();
```
JamesRTaylor: Do we need this method? What happens in the case of a join, where there are multiple concrete tables?

phoenix-core/src/it/java/org/apache/phoenix/end2end/QueryWithTableSampleIT.java
aertoria: Correct. I forgot to remove it.

JamesRTaylor:
This looks very good. Couple of minor comments. One addition that I think is worth making would be to include the sampling rate to the explain plan. See ExplainTable and let's figure out the best place to add this.

phoenix-core/src/main/java/org/apache/phoenix/compile/QueryCompiler.java
aertoria: +1

phoenix-core/src/main/java/org/apache/phoenix/parse/SelectStatement.java
aertoria: This is a valid concern and I also have thought about it during developing. Below is my thought process. Please correct me @JamesRTaylor 

So for a join statment, when there are two concrete tables, during the early stage of the plan preparation phase, they will exist inside `select-statement` object as one `JoinTableNode` object. During this time, `getTableSamplingRate` method is not called. Even if in the future some one calls it, it will return null because type safety check inside method.

Later on, during the plan compiling phase, this `select-statement` is going through JoinCompiler, where two new `select-statment` objects will be created with each one has a brand new `fromTable` object. `FromTable` object is populated from `joinTable` object which essentially is the end product when `JoinTable.Compile()` spliting the original `JoinTableNode` into two. For this code please check Line 136 of `JoinCompiler.java` (Link below)

https://github.com/apache/phoenix/blob/ecd287f618470d1e34082c20feb9939cbb0ea65b/phoenix-core/src/main/java/org/apache/phoenix/compile/JoinCompiler.java


So by the time `getTableSamplingRate` method is called, which is during the "getting iterator phase", each select statments will have exactly one `ConcreteTableNode` with its corresponding sampling rate associate with it.

JamesRTaylor: Sounds reasonable. Please add a comment explaining as you've done here.

phoenix-core/src/it/java/org/apache/phoenix/end2end/QueryWithTableSampleIT.java
```diff
@@ -0,0 +1,261 @@
...
+        Properties props = PropertiesUtil.deepCopy(TestUtil.TEST_PROPERTIES);
+        Connection conn = DriverManager.getConnection(getUrl(), props);
+        try {
+            prepareTableWithValues(conn, 100);
+            String query = "SELECT count(*) FROM " + tableName +" as A tablesample (45), "+joinedTableName+" as B tablesample (75) where A.i1=B.i1";
+            System.out.println(query);
+            ResultSet rs = conn.createStatement().executeQuery(query);
+
+            assertTrue(rs.next());
+            assertEquals(31, rs.getInt(1));
```
JamesRTaylor: Same here - this would be good to assert that the explain plan looks as expected once those changes are made.

aertoria:
>  See ExplainTable and let's figure out the best place to add this.

Purposing adding this logic at `BaseResultIterators`.`explain()` method. Which is between **Line1080** and **Line1081**  of `BaseResultIterators.java` (link below)

https://github.com/apache/phoenix/blob/b9bb918610c04e21b27df8d3fe1c42df508a96f0/phoenix-core/src/main/java/org/apache/phoenix/iterate/BaseResultIterators.java


Reason for this location:
Currently, `BaseResultIterators` inherits abstract class `ExplainTable`.  The explaining string for the table part is essentially finished by these two classes, one following another. Now, the information of tablesampling is stored in `QueryPlan.Statement.fromtable` object, it resides in `BaseResultIterators`. Its parent, the abstract class `ExplainTable` will not have this info (unless we want to modify `PTable` interface and get it from `tableref`).  

In addition, table sampling main logic resides in `BaseResultIterators` when it does `getParallelScans()`. Now we are just making it to also explain the plan when it overrides `explain()`. All thing considered I think `BaseResultIterators` should be the best place to put it.  Please let me know your feedback!

After implemented, it looks like this on single select
```
CLIENT 3-CHUNK 30 ROWS 2370 BYTES PARALLEL 1-WAY ROUND ROBIN FULL SCAN OVER PERSON
TABLESAMPING BY 0.19
```

On join select
```
CLIENT 9-CHUNK 30 ROWS 2370 BYTES PARALLEL 1-WAY ROUND ROBIN FULL SCAN OVER PERSON
TABLESAMPING BY 0.65
    PARALLEL INNER-JOIN TABLE 0
        CLIENT 2-CHUNK 1 ROWS 32 BYTES PARALLEL 1-WAY ROUND ROBIN FULL SCAN OVER US_POPULATION
        TABLESAMPING BY 0.9504
    AFTER-JOIN SERVER FILTER BY PERSON.ADDRESS > US_POPULATION.STATE
```

JamesRTaylor:
Your idea for updating the explain plan is good. One minor nit. How about "SAMPLED AT 0.9504" instead of TABLESAMPING?

aertoria:
Sounds good. I will start prepare another patch.

aertoria:
New patch summary:
1, Rebased.
2, Adding in plan explain for table sample, as well as corresponding test cases.
example:
```
CLIENT 3-CHUNK 30 ROWS 2370 BYTES PARALLEL 1-WAY ROUND ROBIN FULL SCAN OVER PERSON
TABLESAMPING BY 19.0
```

Other minor changes:
reverse unnecessary empty new line change on one file
revised comments for one method.
remove unnecessary system print on test cases.

JamesRTaylor:
+1 after these couple minor tweaks. Nice job, @aertoria!

JamesRTaylor:
Ping @aertoria - would you have a few spare cycles to make that last change? Also, please squash all commits into one and amend your commit message to be prefixed with PHOENIX-153 (i.e. include the dash). Otherwise, we the pull request isn't tied to the JIRA.

aertoria:
The last changes (if we are referring to explain plan change etc) has already been done and went with the last commits.

I'm now rebase with the latest phoenix master and squash all commits into one.

Thanks! @JamesRTaylor

aertoria:
Commit 0507d4f change list:
1, explain plan a new way (Thanks for the suggestion)
2, squash previous four commit into one
3, revise all commit message to start with PHOENIX-153+space 


preview on a Single select
`CLIENT 3-CHUNK 30 ROWS 2370 BYTES PARALLEL 1-WAY 0.2-SAMPLED ROUND ROBIN FULL SCAN OVER PERSON`

on a Join select
```
CLIENT 1-CHUNK 0 ROWS 0 BYTES PARALLEL 1-WAY 0.65-SAMPLED ROUND ROBIN FULL SCAN OVER INX_ADDRESS_PERSON
    SERVER FILTER BY FIRST KEY ONLY
    PARALLEL INNER-JOIN TABLE 0
        CLIENT 1-CHUNK 1 ROWS 32 BYTES PARALLEL 1-WAY 0.15-SAMPLED ROUND ROBIN FULL SCAN OVER US_POPULATION
    DYNAMIC SERVER FILTER BY TO_CHAR("INX_ADDRESS_PERSON.0:ADDRESS") IN (US_POPULATION.STATE)
```

aertoria:
Closing this P.R. as it has been merged.

PR #265, opened 2017-06-26
PHOENIX-3598

Wancy:
Add two params "phoenix.queryserver.doAs.enabled" and "phoenix.queryserver.doAs.param" to control whether to get enduser from request parameters and what is the parameter key name.


phoenix-queryserver/src/main/java/org/apache/phoenix/queryserver/server/QueryServer.java
```diff
@@ -274,6 +282,47 @@ public int run(String[] args) throws Exception {
...
+    private final HttpRequestRemoteUserExtractor requestRemoteUserExtractor;
+    private final boolean enableDoAs;
+    private final String doAsParam;
+
+    public PhoenixRemoteUserExtractor(Configuration conf) {
+      this.requestRemoteUserExtractor = new HttpRequestRemoteUserExtractor();
+      this.doAsParam = conf.get(QueryServices.QUERY_SERVER_DOAS_PARAM,
+              QueryServicesOptions.DEFAULT_QUERY_SERVER_DOAS_PARAM);
+      this.paramRemoteUserExtractor = new HttpQueryStringParameterRemoteUserExtractor(doAsParam);
+      this.enableDoAs = conf.getBoolean(QueryServices.QUERY_SERVER_DOAS_ENABLED_ATTRIB,
```
joshelser: Can you move this check of whether or not we enable `doAs` above to selectively call `withRemoteUserExtractor`, please?

phoenix-queryserver/src/main/java/org/apache/phoenix/queryserver/server/QueryServer.java
```diff
@@ -274,6 +282,47 @@ public int run(String[] args) throws Exception {
...
+              QueryServicesOptions.DEFAULT_QUERY_SERVER_DOAS_ENABLED);
+    }
+
+    @Override
+    public String extract(HttpServletRequest request) throws RemoteUserExtractionException {
+      if (request.getParameter(doAsParam) != null && enableDoAs) {
```
joshelser: This can be simplified when we remove the `enableDoAs` logic.

phoenix-queryserver/src/main/java/org/apache/phoenix/queryserver/server/QueryServer.java
```diff
@@ -274,6 +282,47 @@ public int run(String[] args) throws Exception {
...
+        UserGroupInformation ugi = UserGroupInformation.createRemoteUser(request.getRemoteUser());
+        UserGroupInformation proxyUser = UserGroupInformation.createProxyUser(doAsUser, ugi);
+
+        // Check if this user is allowed to be impersonated.
+        // Will throw AuthorizationException if the impersonation as this user is not allowed
+        try {
+          ProxyUsers.authorize(proxyUser, request.getRemoteAddr());
+          return doAsUser;
+        } catch (AuthorizationException e) {
+          throw new RemoteUserExtractionException(e.getMessage());
```
joshelser: Can the exception be passed into the RemoteUserExtractionException instead of just the message? (to preserve the stack trace)

phoenix-queryserver/src/test/java/org/apache/phoenix/queryserver/server/PhoenixRemoteUserExtractorTest.java
```diff
@@ -0,0 +1,102 @@
...
+    PhoenixRemoteUserExtractor extractor = new PhoenixRemoteUserExtractor(conf);
+    try {
+      assertEquals("enduser", extractor.extract(request));
+    } catch (RemoteUserExtractionException e) {
+      LOG.info(e.getMessage());
+    }
+  }
+
+  @Test
+  public void testDoNotUseDoAs() {
```
joshelser: To test this code if you take my above suggestion, you could make a new method in QueryServer which does

```java
Builder setRemoteUserExtractorIfNecessary(Builder b, Configuration conf) {
  if (conf.getBoolean(QueryServices.QUERY_SERVER_DOAS_ENABLED_ATTRIB,
          QueryServicesOptions.DEFAULT_QUERY_SERVER_DOAS_ENABLED)) {
    return builder.withRemoteUserExtractor(new PhoenixRemoteUserExtractor(getConf()));
  }
  return builder;
}
```

This would let you easily mock the Builder and verify that your extractor is configured when the property is set to "true".

phoenix-queryserver/pom.xml
```diff
@@ -147,6 +147,10 @@
       <groupId>commons-logging</groupId>
       <artifactId>commons-logging</artifactId>
     </dependency>
+    <dependency>
+      <groupId>org.mockito</groupId>
+      <artifactId>mockito-all</artifactId>
```
joshelser: Needs a `<scope>test</scope>`

phoenix-queryserver/src/test/java/org/apache/phoenix/queryserver/server/PhoenixRemoteUserExtractorTest.java
Wancy: Hi @joshelser,
Is there a way to check if builder called withRemoteUserExtractor or not? I tried used "equals" but there will always be two new builder object to compare. Also there is no getRemoteUserExtractor method for HttpBuilder.

joshelser: No, there is no getter on the builder to verify it's called. Instead you can use the `Mockito.verify(builder)` method. Something like:

```java
Configuration conf = createTestConfiguration();
Builder b = Mockito.mock(Builder.class);
Mockito.when(b.withRemoteUserExtractor(Mockito.any(PhoenixRemoteUserExtractor.class))).thenReturn(b);
setRemoteUserExtractorIfNecessary(b, conf);
Mockito.verify(b);
```

This should essentially verify that `withRemoteUserExtractor` was invoked by `setRemoteUserExtractorIfNecessary`

Wancy:
Hi @joshelser,
I made some changes according to your comments, please review, thanks.

phoenix-queryserver/src/main/java/org/apache/phoenix/queryserver/server/QueryServer.java
```diff
@@ -273,6 +282,54 @@ public int run(String[] args) throws Exception {
...
+              QueryServicesOptions.DEFAULT_QUERY_SERVER_REMOTEUSEREXTRACTOR_PARAM);
+      this.paramRemoteUserExtractor = new HttpQueryStringParameterRemoteUserExtractor(userExtractParam);
+    }
+
+    @Override
+    public String extract(HttpServletRequest request) throws RemoteUserExtractionException {
+      if (request.getParameter(userExtractParam) != null) {
+        String extractedUser = paramRemoteUserExtractor.extract(request);
+        UserGroupInformation ugi = UserGroupInformation.createRemoteUser(request.getRemoteUser());
+        UserGroupInformation proxyUser = UserGroupInformation.createProxyUser(extractedUser, ugi);
```
joshelser: In re-reading the above, I'm a little worried about the edge-cases. With PQS right now, we have the following cases "supported"

1) Kerberos+SPNEGO as the Kerberos user (elserj@EXAMPLE.COM authenticates to PQS and the PQS credentials are used to query Phoenix as elserj@EXAMPLE.COM)
2) Kerberos auth with HBase but no SPNEGO for PQS clients (legacy support for how things used to work before the SPNEGO auth was built -- PQS user does everything for users)
3) Without Kerberos, all queries run as the PQS user (out of the box).

Avatica supports more than what point 3 above does, but we haven't prioritized wiring that up as most people just want point 1. @Wancy, I had originally thought you were just trying to enable a PQS client with Kerberos credentials to say that they are someone else (extension of point 1 -- Credentials to PQS are for "elserj" but instead of querying Phoenix as "elserj", query as "bob").

Did you also want this to work for cases when Kerberos is not in the mix? I think that would require some additional changes as I don't think this will work as-is.

phoenix-queryserver/src/main/java/org/apache/phoenix/queryserver/server/QueryServer.java
```diff
@@ -228,7 +235,9 @@ public int run(String[] args) throws Exception {
         builder.withSpnego(ugi.getUserName(), additionalAllowedRealms)
             .withAutomaticLogin(keytab)
             .withImpersonation(new PhoenixDoAsCallback(ugi, getConf()));
+
       }
+      setRemoteUserExtractorIfNecessary(builder, getConf());
```
joshelser: With respect to my long-winded comment below, if you're only looking to support Kerberos, we want to move this line into the above if-block.

phoenix-queryserver/src/main/java/org/apache/phoenix/queryserver/server/QueryServer.java
Wancy: Hi @joshelser,
I think I understand your concern of the edge cases. I originally wanna add it just for kerberos cases, but I thought user extract could be extended to simple auth as well in the future, but seems a lot more work needs to be done than I thought. Also like you said most people just want point1.

I think for this jira just add it for the kerberos case is more practical.

phoenix-queryserver/src/main/java/org/apache/phoenix/queryserver/server/QueryServer.java
Wancy: Agree to put it inside the if-block for only kerberos case :)

phoenix-queryserver/src/main/java/org/apache/phoenix/queryserver/server/QueryServer.java
joshelser: Agreed! I think the work you've put in would be nice to support for the non-Kerberos case, but let's not hold up this change for that.

I will try to write up a test case for PQS (mini-hbase, mini-kdc, and PQS) to validate your changes here before I commit.

joshelser:
Woops. I forgot to close this via commit message. If you could close it at your convenience, @Wancy, I'd appreciate it!

Wancy:
Thanks @joshelser !!

PR #268, opened 2017-07-21
PHOENIX-4010 Hash Join cache may not be send to all regionservers whe…

ankitsinghal:
…n we have stale HBase meta cache

### DIFF
--- a/phoenix-core/src/it/java/org/apache/phoenix/end2end/BaseJoinIT.java
+++ b/phoenix-core/src/it/java/org/apache/phoenix/end2end/BaseJoinIT.java
@@ -17,6 +17,7 @@
  */
 package org.apache.phoenix.end2end;
 
+import static org.apache.phoenix.util.TestUtil.TEST_PROPERTIES;
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertTrue;
 
@@ -28,11 +29,12 @@ import java.sql.SQLException;
 import java.sql.Timestamp;
 import java.text.SimpleDateFormat;
 import java.util.Map;
+import java.util.Properties;
 import java.util.regex.Pattern;
 
 import org.apache.hadoop.hbase.HConstants;
 import org.apache.phoenix.cache.ServerCacheClient;
-import org.apache.phoenix.end2end.HashJoinCacheIT.InvalidateHashCacheRandomly;
+import org.apache.phoenix.util.PropertiesUtil;
 import org.apache.phoenix.util.ReadOnlyProps;
 import org.apache.phoenix.util.SchemaUtil;
 import org.apache.phoenix.util.StringUtil;
@@ -42,7 +44,8 @@ import org.junit.BeforeClass;
 import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.Maps;
 
-public abstract class BaseJoinIT extends BaseUniqueNamesOwnClusterIT {
+public abstract class BaseJoinIT extends ParallelStatsDisabledIT {
+	
     protected static final String JOIN_SCHEMA = "Join";
     protected static final String JOIN_ORDER_TABLE = "OrderTable";
     protected static final String JOIN_CUSTOMER_TABLE = "CustomerTable";
@@ -56,16 +59,6 @@ public abstract class BaseJoinIT extends BaseUniqueNamesOwnClusterIT {
     protected static final String JOIN_COITEM_TABLE_FULL_NAME = '"' + JOIN_SCHEMA + "\".\"" + JOIN_COITEM_TABLE + '"';
 
     private static final Map<String,String> tableDDLMap;
-    
-    @BeforeClass
-    public static void doSetup() throws Exception {
-        Map<String, String> serverProps = Maps.newHashMapWithExpectedSize(10);
-        serverProps.put(HConstants.HBASE_CLIENT_RETRIES_NUMBER, "2");
-        serverProps.put(HConstants.HBASE_RPC_TIMEOUT_KEY, "10000");
-        serverProps.put("hbase.client.pause", "5000");
-        Map<String, String> clientProps = Maps.newHashMapWithExpectedSize(1);
-        setUpTestDriver(new ReadOnlyProps(serverProps.entrySet().iterator()), new ReadOnlyProps(clientProps.entrySet().iterator()));
-    }
     
     static {
         ImmutableMap.Builder<String,String> builder = ImmutableMap.builder();
@@ -457,6 +450,12 @@ public abstract class BaseJoinIT extends BaseUniqueNamesOwnClusterIT {
         conn.commit();
     }
 
+	protected Connection getConnection() throws SQLException {
+		Properties props = PropertiesUtil.deepCopy(TEST_PROPERTIES);
+		props.put(ServerCacheClient.HASH_JOIN_SERVER_CACHE_RESEND_PER_SERVER, "true");
+		return DriverManager.getConnection(getUrl(), props);
+	}
+	
     protected void createIndexes(Connection conn, String virtualName, String realName) throws Exception {
         if (indexDDL != null && indexDDL.length > 0) {
             for (String ddl : indexDDL) {
--- a/phoenix-core/src/it/java/org/apache/phoenix/end2end/BaseJoinIT.java
+++ b/phoenix-core/src/it/java/org/apache/phoenix/end2end/BaseJoinIT.java
@@ -30,14 +30,19 @@ import java.text.SimpleDateFormat;
 import java.util.Map;
 import java.util.regex.Pattern;
 
+import org.apache.hadoop.hbase.HConstants;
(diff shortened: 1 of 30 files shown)


phoenix-core/src/it/java/org/apache/phoenix/end2end/HashJoinIT.java
```diff
@@ -3453,4 +3496,27 @@ public void testJoinWithSetMaxRows() throws Exception {
             conn.close();
         }
     }
+    
+    public static class InvalidateHashCacheRandomly extends SimpleRegionObserver {
```
JamesRTaylor: These join tests run on the same cluster as all other tests derived from ParallelStatsDisabledIT. Rather than having a random test failure, let's move this to a dedicated test derived from BaseUniqueNamesOwnClusterIT and have it consistently fail.

phoenix-core/src/main/java/org/apache/phoenix/cache/ServerCacheClient.java
```diff
@@ -394,4 +420,82 @@ public static String idToString(byte[] uuid) {
...
+                    });
+        } catch (Throwable t) {
+            throw new Exception(t);
+        }
+        if (results != null && results.size() == 1) { return results.values().iterator().next().getReturn(); }
+        return false;
+
+    }
+    
+    public static ServerCache getCacheForId(List<ServerCache> caches, Long cacheId) {
```
JamesRTaylor: How about using a Map<ImmutableBytesPtr,ServerCache> instead of a List and switching id to ImmutableBytesPtr? Then you can get rid of this static method.

phoenix-core/src/main/java/org/apache/phoenix/iterate/TableResultIterator.java
```diff
@@ -163,8 +185,38 @@ public Tuple next() throws SQLException {
...
+                                    "Retrying when Hash Join cache is not found on the server ,by sending the cache again");
+                            if (retry <= 0) { throw e1; }
+                            retry--;
+                            try {
+                                Long cacheId = ((HashJoinCacheNotFoundException)e1).getCacheId();
+                                if (!hashCacheClient.addHashCacheToServer(newScan.getStartRow(),
+                                        ServerCacheClient.getCacheForId(caches, cacheId), plan.getTableRef().getTable())) { throw e1; }
+                                
+                                try {
+                                    if(ScanUtil.isClientSideUpsertSelect(newScan)) {
```
JamesRTaylor: Why is this necessary? Can we generalize this is a better way?

phoenix-core/src/main/java/org/apache/phoenix/cache/ServerCacheClient.java
```diff
@@ -303,73 +318,84 @@ public TaskExecutionMetricsHolder getTaskExecutionMetric() {
...
+            List<byte[]> keys = cache.getKeys();
+            ConnectionQueryServices services = connection.getQueryServices();
+            Throwable lastThrowable = null;
+            TableRef cacheUsingTableRef = cacheUsingTableRefMap.get(Bytes.mapKey(cacheId));
+            final PTable cacheUsingTable = cacheUsingTableRef.getTable();
+            byte[] tableName = cacheUsingTableRef.getTable().getPhysicalName().getBytes();
+            iterateOverTable = services.getTable(tableName);
+
+            List<HRegionLocation> locations = services.getAllTableRegions(tableName);
+            Set<HRegionLocation> remainingOnServers = new HashSet<HRegionLocation>(servers);
```
JamesRTaylor: Instead of Set<HRegionLocation> can we change this to Set<ServerName>? It'd make it more clear IMHO

phoenix-core/src/main/java/org/apache/phoenix/cache/ServerCacheClient.java
```diff
@@ -303,73 +318,84 @@ public TaskExecutionMetricsHolder getTaskExecutionMetric() {
...
+             * through the current metadata boundaries and remove the cache once for each server that we originally sent
+             * to.
+             */
+            if (LOG.isDebugEnabled()) {
+                LOG.debug(addCustomAnnotations("Removing Cache " + cacheId + " from servers.", connection));
+            }
+            for (HRegionLocation entry : locations) {
+             // Call once per server
+                if (remainingOnServers.contains(entry)
+                        || (keys != null && ByteUtil.contains(keys,entry.getRegionInfo().getStartKey()))) {
```
JamesRTaylor: Why is this extra check required? Is this when the Hash Cache is sent over to the RS that doesn't have it? If so, can we document that?

phoenix-core/src/main/java/org/apache/phoenix/compile/UpsertCompiler.java
```diff
@@ -804,6 +804,7 @@ public Long getEstimatedBytesToScan() throws SQLException {
             ////////////////////////////////////////////////////////////////////
             // UPSERT SELECT run client-side
             /////////////////////////////////////////////////////////////////////
+            ScanUtil.setClientSideUpsertSelect(queryPlan.getContext().getScan());
```
JamesRTaylor: This would be better to encapsulate in the StatementPlan if we even need it. Why do we need it?

phoenix-core/src/main/java/org/apache/phoenix/iterate/TableResultIterator.java
```diff
@@ -108,6 +126,10 @@ public TableResultIterator(MutationState mutationState, Scan scan, ScanMetricsHo
         this.scanIterator = UNINITIALIZED_SCANNER;
         this.renewLeaseThreshold = renewLeaseThreshold;
         this.scanGrouper = scanGrouper;
+        this.hashCacheClient = new HashCacheClient(plan.getContext().getConnection());
+        this.caches = caches;
+        this.retry=plan.getContext().getConnection().getQueryServices().getConfiguration()
```
JamesRTaylor: Can we use getProps() here instead of going through the Configuration object? We've found the synchronization done by Configuration to be a bottleneck in the past (and not necessary since these properties don't change).

phoenix-core/src/main/java/org/apache/phoenix/compile/UpsertCompiler.java
ankitsinghal: this change is actually a part of PHOENIX-3163, let me see if we can improve.

JamesRTaylor: Let's do those changes in a separate patch, please.

ankitsinghal:
@JamesRTaylor , can you please review, I have removed the dependency on PHOENIX-3163.

ankitsinghal:
Ping @JamesRTaylor

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/BaseScannerRegionObserver.java
```diff
@@ -109,12 +110,16 @@
 
     /** Exposed for testing */
     public static final String SCANNER_OPENED_TRACE_INFO = "Scanner opened on server";
+    public static final String CLIENT_SIDE_UPSERT_SELECT = "_ClientSideUpsertSelect";
+    protected Configuration rawConf;
```
JamesRTaylor: What's the reason for this new rawConf member variable? Where is it used?

phoenix-core/src/it/java/org/apache/phoenix/end2end/BaseJoinIT.java
```diff
@@ -30,14 +30,19 @@
...
 import org.apache.phoenix.util.SchemaUtil;
 import org.apache.phoenix.util.StringUtil;
 import org.junit.Before;
+import org.junit.BeforeClass;
 
 import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.Maps;
 
-public abstract class BaseJoinIT extends ParallelStatsDisabledIT {
+public abstract class BaseJoinIT extends BaseUniqueNamesOwnClusterIT {
```
JamesRTaylor: Can we keep BaseJoinIT derived from ParallelStatsDisabledIT? Otherwise, there's a lot more mini clusters created which increases our end-to-end test time run more and more.

If the reason is because you're adding InvalidateHashCacheRandomly, just add it instead dynamically to the table created in your test. To prevent flapping, you need to poll until you the descriptor matches what you've changed. See PartialIndexRebuilderIT.addWriteFailingCoprocessor() for an example you can generalize and/or copy/paste.

JamesRTaylor:
Thanks for the ping, @ankitsinghal. The code looks good, but if we can fix up the tests with a couple of tweaks, we'll be good to go after rebasing.

ankitsinghal:
Thanks @JamesRTaylor , have taken care the review comments in latest commit. Let me know if it is fine now

JamesRTaylor:
+1

stoty:
Already merged.

PR #270, opened 2017-08-26
 `DEFAULT_TIME_ZONE_ID`  should take value from `TimeZone.getDefault().getID()` instead of constant `GMT`

szq80140:
`DEFAULT_TIME_ZONE_ID`  should take value from `TimeZone.getDefault().getID()` instead of constant `GMT`, so that all `DATE`  and  `TIMESTAMP`  column values are formatted  using timezone configured in JVM.


ieayoio:
I think what you said is quite right. Why did you close it? I solved the time zone problem of thick client in this way, but the time zone of thin client is still very chaotic. I don't know what the problem is

PR #275, opened 2017-10-05
PHOENIX-4237: Add function to calculate Java collation keys

shehzaadn:
Here we implement a generalized solution for calculating Java collation keys by creating Java collators based on a user locale. These collation keys can then be used in an ORDER BY clause to sort strings in a natural-language-appropriate way. We add a new Phoenix function COLLKEY. In general usage for this function will be:

select name from my_table order by COLLKEY(name, 'zh_TW')

We use artifacts from the ICU4J project and recently open-sourced grammaticus project (by Maven dependency). We were forced to include some code from ICU4J because some jars produced by that project aren't published in Maven. We also include code from Salesforce that has been licensed for open-source release but not yet published as artifacts in maven.

There are three commits that split the changes into three logical pieces:

1) f8cb121: Add the external source code described above
2) fdbb5e0: Make changes needed to the Phoenix license due to the above (and fix to what seems to be an existing bug) 
3) 98cfc10: The actual function implementation of COLLKEY - new code that uses the code introduced above and newly introduced dependencies via maven.

Thanks in advance to the Phoenix community for your feedback on this.


phoenix-core/src/main/java/com/force/db/i18n/OracleUpper.java
```diff
@@ -0,0 +1,66 @@
+/*
```
JamesRTaylor: @joshelser - my take, based on this[1], is that it's ok to include source code in an ASF project with a BSD license (as opposed to only having BSD licensed software as an external dependency). WDYT?

[1] http://apache.org/licenses/#code-developed-elsewhere-received-under-a-category-a-license-incorporated-into-apache-projects-distributed-by-apache-and-licensed-to-downstream-users-under-its-original-license

phoenix-core/src/main/java/org/apache/phoenix/expression/function/CollationKeyFunction.java
```diff
@@ -0,0 +1,233 @@
...
+			byte[] collationKeyByteArray = collator.getCollationKey(inputValue).toByteArray();
+
+			if(LOG.isDebugEnabled()) {
+				LOG.debug("Collation key bytes:" + Arrays.toString(collationKeyByteArray));
+			}
+			
+			// byte is signed in Java, but we need unsigned values for comparison
+			// https://www.programcreek.com/java-api-examples/index.php?api=java.text.CollationKey
+			// Byte.toUnsignedInt will convert a byte value between [-128,127] to an int value
+			// between [0,255]
```
JamesRTaylor: This code should not be necessary. Phoenix sorting is based on lexicographical order of the bytes. So -128 is really like 0 and 127 is like 255.

phoenix-core/src/main/java/com/force/db/i18n/OracleUpper.java
joshelser: Yup! You got it right, James. Whether we include the code in binary form or source form, for BSD, we treat them the same (propagate in LICENSE, and copyright/etc in NOTICE). If there's a license header for the file, we would also leave that, IIRC.

phoenix-core/src/test/java/org/apache/phoenix/expression/function/CollationKeyFunctionTest.java
```diff
@@ -0,0 +1,134 @@
...
+
+import com.google.common.collect.Lists;
+
+/**
+ * "Unit" tests for CollationKeyFunction
+ * 
+ * @author snakhoda
+ *
+ */
+public class CollationKeyFunctionTest {
```
JamesRTaylor: We'll need more tests. You really want to test the sort order of a list of strings matches the expected linguistic sort order. These tests don't have a lot of meaning in terms of validating the sort order is correct IMHO.

We'll also want end2end tests that use the new function.

phoenix-core/src/main/java/org/apache/phoenix/expression/function/CollationKeyFunction.java
```diff
@@ -0,0 +1,233 @@
...
+
+import com.ibm.icu.impl.jdkadapter.CollatorICU;
+import com.ibm.icu.util.ULocale;
+
+/**
+ * A Phoenix Function that calculates a collation key for an input string based
+ * on a caller-provided locale and collator strength and decomposition settings.
+ * 
+ * It uses the open-source grammaticus and i18n packages to obtain the collators
+ * it needs.
```
JamesRTaylor: We should include more comments here. In particular, what sort order will we get? Does this mimic some other databases behavior (i.e. Oracle)? Does it deviate from that at all? Does Oracle follow some standard that we could point to?

Also, please make sure to budget time to update our online reference manual: https://phoenix.apache.org/language/functions.html. This lives in phoenix.csv in our SVN repo as described here: https://phoenix.apache.org/building_website.html

JamesRTaylor:
Thanks for the patch, @shehzaadn. This looks like a general enough built-in function to include in Phoenix IMHO. See inline for more specific comments. It'd be much better to include the first two commits as external dependencies. If we don't do that, we'll need to quickly follow up with replacing them with external dependencies (and make sure we don't change those files at all).

phoenix-core/src/main/java/org/apache/phoenix/expression/function/CollationKeyFunction.java
snakhoda-sfdc: I started out with a straight byte array (VARBINARY), but realized the issue is related to 2s complement binary representation.

Here's what I was able to gather:

The collator's collation key produces a byte array that holds values that are supposed to be treated as unsigned. However, since byte is a signed (2s complement) datatype in Java, doing a Java comparison on them with regular operators (which is presumably what Phoenix does) produces incorrect results.

For example, 00000001 represents 1 and 11111111 represents -1 in 2s complement. If we simply compare those bytes using standard operators then 11111111 is less than 00000001. But a collator key byte array that returns these two bytes would like them to be treated as "1" and "255" respectively and would need 11111111 to be greater than 00000001.

Getting this effect out of a byte array basically means taking each byte and widening it (to an int) which results in the lower order bytes representing the unsigned positive number that was intended by the collation key.

See this example usage of toByteArray, which does the same thing: http://icu-project.org/apiref/icu4j59m1/com/ibm/icu/text/CollationKey.html#toByteArray--

BTW, Byte.toUnsignedInt in Java 8 is basically a bitwise AND with 0xFF.

JamesRTaylor: > doing a Java comparison on them with regular operators (which is presumably what Phoenix does) produces incorrect results

That's not what Phoenix does. That's what I tried to explain in my previous comment. In your unit tests, the best way to confirm this is through an end-to-end integration test that uses an ORDER BY COLLKEY(my_varchar_col):
* Create a simple table CREATE TABLE t (k INTEGER PRIMARY KEY, v VARCHAR)
* Populate the table with some multi-byte string values:
    UPSERT INTO t VALUES(1,'ണഫɰɸ')
    UPSERT INTO t VALUES(2,'ɚɦ')
* Execute an ORDER BY on the table: SELECT * FROM t ORDER BY COLLKEY(v)
* Confirm that the rows are ordered as expected by a linguistic sort

The above test is a must-have. We can also have lower level unit tests that don't need to run as an integration test to test the sorting. You'd need to manually create an OrderedResultIterator. See OrderedResultIteratorTest as an example.

shehzaadn-vd: Yes, I'm working on the tests you mentioned and your bulleted list is the sequence I had in mind.

My comment above was trying to explain that without the conversion of the collator key byte array to an array of unsigned integers the ORDER BY does not work as expected. My comment on Phoenix behaviour was speculative, without having looked inside. Introducing conversion to unsigned resulted in the behaviour we need.

JamesRTaylor: Let's start with the tests. That conversion should not be necessary.

snakhoda-sfdc:
@JamesRTaylor Thanks for your comments. I added two further commits:

199c389: This addresses your comment about the byte array comparison. You were right! I must have got confused earlier with what was being displayed on sqlline.py not matching the sort order.  I also added a formatter for PVarBinary because without it you simply get a Java hash code in sqlline.py which is hard to do anything with.

8cc2b5c: This adds the end-to-end tests you mentioned and also changes the unit test to use the hex representation of the byte array to make it easier to read.

phoenix-core/src/main/java/org/apache/phoenix/expression/function/CollationKeyFunction.java
```diff
@@ -0,0 +1,221 @@
...
+	}
+
+	public CollationKeyFunction(List<Expression> children) throws SQLException {
+		super(children);
+	}
+
+	@Override
+	public boolean evaluate(Tuple tuple, ImmutableBytesWritable ptr) {
+		try {
+			String inputValue = getInputValue(tuple, ptr);
```
JamesRTaylor: The evaluate method is called for every row during processing, so we want to have as little code here as possible. You can create a Collator local variable and move all the code that sets it up to an init() method. You'd call the init() method in the CollationKeyFunction(List<Expression> children) constructor and in an overridden readFields method like this (see InstrFunction for an example):

    @Override
        super.readFields(input);
        init();
    }

phoenix-core/src/main/java/org/apache/phoenix/expression/function/CollationKeyFunction.java
```diff
@@ -0,0 +1,221 @@
...
+			if (collatorDecomposition != null) {
+				collator.setDecomposition(collatorDecomposition);
+			}
+
+			if(LOG.isDebugEnabled()) {
+				LOG.debug(String.format("Collator: [strength: %d, decomposition: %d], Special-Upper-Case: %s",
+					collator.getStrength(), collator.getDecomposition(), BooleanUtils.isTrue(useSpecialUpperCaseCollator)));
+			}
+			
+			byte[] collationKeyByteArray = collator.getCollationKey(inputValue).toByteArray();
```
JamesRTaylor: Is collator.getCollationKey(inputValue) thread safe? If not, we'll need to make one more tweak, so please let me know.

phoenix-core/src/test/java/org/apache/phoenix/expression/function/CollationKeyFunctionTest.java
```diff
@@ -96,33 +96,35 @@ private static boolean testExpression(String inputStr, String localeIsoCode, Sor
...
 		Expression collationKeyFunction = new CollationKeyFunction(expressions);
 		ImmutableBytesWritable ptr = new ImmutableBytesWritable();
 		boolean ret = collationKeyFunction.evaluate(null, ptr);
 		if (ret) {
-			PhoenixArray result = (PhoenixArray) collationKeyFunction.getDataType().toObject(ptr,
+			byte[] result = (byte[]) collationKeyFunction.getDataType().toObject(ptr,
 					collationKeyFunction.getSortOrder());
 
+			byte[] expectedCollationKeyByteArray = Hex.decodeHex(expectedCollationKeyBytesHex.toCharArray());
+
```
JamesRTaylor: Why not use assertArrayEquals here instead?

phoenix-core/src/main/java/org/apache/phoenix/jdbc/PhoenixConnection.java
```diff
@@ -336,6 +338,7 @@ public ReadOnlyProps getProps() {
         formatters.put(PUnsignedTimestamp.INSTANCE, timestampFormat);
         formatters.put(PDecimal.INSTANCE,
                 FunctionArgumentType.NUMERIC.getFormatter(numberPattern));
+        formatters.put(PVarbinary.INSTANCE, VarBinaryFormatter.INSTANCE);
```
JamesRTaylor: +1. Nice idea!

phoenix-core/src/main/java/org/apache/phoenix/expression/function/CollationKeyFunction.java
```diff
@@ -0,0 +1,221 @@
...
+		@FunctionParseNode.Argument(allowedTypes = { PBoolean.class }, defaultValue = "false", isConstant = true),
+		// collator strength
+		@FunctionParseNode.Argument(allowedTypes = { PInteger.class }, defaultValue = "null", isConstant = true),
+		// collator decomposition
+		@FunctionParseNode.Argument(allowedTypes = { PInteger.class }, defaultValue = "null", isConstant = true) })
+public class CollationKeyFunction extends ScalarFunction {
+
+	private static final Log LOG = LogFactory.getLog(CollationKeyFunction.class);
+
+	public static final String NAME = "COLLKEY";
```
JamesRTaylor: Is there a convention in other RDBMS for the name of this function? Is it spelled out COLLATION_KEY or abbreviated as you've done? If abbreviated, then IMHO, it'd be better to name the class and unit tests CollKeyFunction, CollKeyFunctionIT, etc. to make it easier to find (i.e. based on the function name). That's our typical convention.

JamesRTaylor:
This is looking very good, @shehzaadn - thanks for the revisions. Couple more comments, but it's getting pretty close IMHO. How is the publishing to maven of the dependent jars looking?

phoenix-core/src/main/java/org/apache/phoenix/expression/function/CollationKeyFunction.java
snakhoda-sfdc: There's no convention as such. Oracle functions are nlssort/ nls_upper, etc.

We can call it COLLATION_KEY here. I'd rather have the name be more descriptive than less. Does that work?

phoenix-core/src/main/java/org/apache/phoenix/expression/function/CollationKeyFunction.java
snakhoda-sfdc: Unfortunately, the collators that we get from ICU4J are not thread-safe. There's a way to make them thread-safe but that would require a change to i18n-util.

snakhoda-sfdc:
@JamesRTaylor thanks for the feedback and support! So we have the i18n-util jar on maven now, but not the icu4j jars. Once the icu4j jars are published to maven, i18n-util will have to change to upgrade its dependency to the new version. I'm hoping that change will be in next week.

Once that happens, I was thinking of creating a new PR that removes the outside code and introduces the external dependency.

phoenix-core/src/main/java/org/apache/phoenix/expression/function/CollationKeyFunction.java
snakhoda-sfdc: @JamesRTaylor  Won't that require that the collator be thread-safe? Or will the CollationKeyFunction not be shared across threads? (Maybe the tweak you were mentioning is for this purpose?)

JamesRTaylor: You can indicate that a function is not thread safe. I'll give you an easy way to do that and let you know what you need to do. In the meantime, if you could do the above, that'd be good.

phoenix-core/src/test/java/org/apache/phoenix/expression/function/CollationKeyFunctionTest.java
snakhoda-sfdc: Good point. Will do.

phoenix-core/src/main/java/org/apache/phoenix/expression/function/CollationKeyFunction.java
```diff
@@ -0,0 +1,230 @@
...
+	}
+
+	private <T> T getLiteralValue(int childIndex, Class<T> type) {
+		Expression expression = getChildren().get(childIndex);
+		if (LOG.isDebugEnabled()) {
+			LOG.debug("child: " + childIndex + ", expression: " + expression);
+		}
+		if (!(expression instanceof LiteralExpression)) {
+			throw new ExpressionEvaluationException(expression);
+		}
```
JamesRTaylor: This if statement isn't necessary. Phoenix already checks this and throws if it's not the case based on your declaration of the argument as being a constant.

phoenix-core/src/main/java/org/apache/phoenix/expression/function/CollationKeyFunction.java
```diff
@@ -0,0 +1,230 @@
...
+	}
+
+	@Override
+	public PDataType getDataType() {
+		return PVarbinary.INSTANCE;
+	}
+
+	@Override
+	public String getName() {
+		return "COLLATION_KEY";
```
JamesRTaylor: Use CollationKeyFunction.NAME here instead of repeating string constant.

phoenix-core/src/main/java/org/apache/phoenix/expression/function/CollationKeyFunction.java
```diff
@@ -0,0 +1,230 @@
...
+		if (collatorDecomposition != null) {
+			collator.setDecomposition(collatorDecomposition);
+		}
+
+		if (LOG.isDebugEnabled()) {
+			LOG.debug(String.format("Collator: [strength: %d, decomposition: %d], Special-Upper-Case: %s",
+					collator.getStrength(), collator.getDecomposition(),
+					BooleanUtils.isTrue(useSpecialUpperCaseCollator)));
+		}
+	}
```
JamesRTaylor: As of PHOENIX-4294 (just make sure you've pulled the latest), you can declare your function as not being thread safe by adding the following method override:

    @Override
    public boolean isThreadSafe() {
        return false;
    }

phoenix-core/src/test/java/org/apache/phoenix/expression/function/CollationKeyFunctionTest.java
```diff
@@ -0,0 +1,143 @@
...
+		assertEquals(ret1, ret2);
+	}
+
+	private static boolean testExpression(String inputStr, String localeIsoCode, SortOrder sortOrder,
+			String expectedCollationKeyBytesHex) throws Exception {
+		LiteralExpression inputStrLiteral, localeIsoCodeLiteral, upperCaseBooleanLiteral, strengthLiteral,
+				decompositionLiteral;
+		inputStrLiteral = LiteralExpression.newConstant(inputStr, PVarchar.INSTANCE, sortOrder);
+		localeIsoCodeLiteral = LiteralExpression.newConstant(localeIsoCode, PVarchar.INSTANCE, sortOrder);
+		upperCaseBooleanLiteral = LiteralExpression.newConstant(Boolean.FALSE, PBoolean.INSTANCE, sortOrder);
```
JamesRTaylor: Please add tests around setting upperCaseBooleanLiteral to true.

phoenix-core/src/test/java/org/apache/phoenix/expression/function/CollationKeyFunctionTest.java
```diff
@@ -0,0 +1,143 @@
...
+		strengthLiteral = LiteralExpression.newConstant(null, PInteger.INSTANCE, sortOrder);
+		decompositionLiteral = LiteralExpression.newConstant(null, PInteger.INSTANCE, sortOrder);
```
JamesRTaylor: Please add tests around setting strength and decomposition.

JamesRTaylor:
Looking very good. Couple minor nits and the testing needs to be rounded out just a bit.

snakhoda-sfdc:
@JamesRTaylor i've addressed the last round of comments in this commit (9d6d4f7). Thanks.

phoenix-core/src/main/java/com/ibm/icu/impl/jdkadapter/NumberFormatICU.java
```diff
@@ -0,0 +1,229 @@
...
+
+    @Override
+    public boolean equals(Object obj) {
+        if (obj instanceof NumberFormatICU) {
+            return ((NumberFormatICU)obj).fIcuNfmt.equals(fIcuNfmt);
+        }
+        return false;
+    }
+
+    //public String format(double number)
```
solzy: delete this unusable lien, keep clean!

snakhoda-sfdc: Thanks for taking a look at this PR, @solzy. This code is external and simply copied over from ICU4J 59.1. The reason it's here at all is that that project doesn't have all its artifacts in maven. I'm hoping to have a new PR in the near future to remove this external code and replace it with maven dependencies. CC: @JamesRTaylor

JamesRTaylor:
+1. Nice work, @shehzaadn!

JamesRTaylor:
Would you mind squashing all the commits into a single commit, @shehzaadn and I'll get this committed?

snakhoda-sfdc:
@JamesRTaylor I'm not sure how to do that within this PR. Looking at https://github.com/blog/2141-squash-your-commits, I believe at the time you merge the PR, github should give you the option to squash all commits into one. Will that suffice?

stoty:
Already merged.

PR #277, opened 2017-10-23
PHOENIX-3757 System mutex table not being created in SYSTEM namespace…

karanmehta93:
… when namespace mapping is enabled
On fresh cluster with system table namespace mapping enabled, SYSMUTEX will be created in SYSTEM namespace directly.
It will be migrated to SYSTEM namespace if that property is enabled
The migrations also require locking in SYSMUTEX table since it disables and enables SYSCAT.
SYSCAT upgrades will acquire lock in appropriate SYSMUTEX table based on the client properties.

### DIFF
--- a/phoenix-core/src/it/java/org/apache/phoenix/end2end/MigrateSystemTablesToSystemNamespaceIT.java
+++ b/phoenix-core/src/it/java/org/apache/phoenix/end2end/MigrateSystemTablesToSystemNamespaceIT.java
@@ -21,6 +21,7 @@ import org.apache.hadoop.hbase.HBaseTestingUtility;
 import org.apache.hadoop.hbase.TableName;
 import org.apache.hadoop.security.UserGroupInformation;
 import org.apache.phoenix.coprocessor.MetaDataProtocol;
+import org.apache.phoenix.exception.UpgradeInProgressException;
 import org.apache.phoenix.jdbc.PhoenixConnection;
 import org.apache.phoenix.jdbc.PhoenixDatabaseMetaData;
 import org.apache.phoenix.query.BaseTest;
@@ -233,7 +234,9 @@ public class MigrateSystemTablesToSystemNamespaceIT extends BaseTest {
             });
             fail("Multiple clients should not be able to migrate simultaneously.");
         } catch (Exception e) {
-            // ignore
+            if(!(e.getCause() instanceof UpgradeInProgressException)) {
+                fail("UpgradeInProgressException expected since the user is trying to migrate when SYSMUTEX is locked.");
+            }
         }
 
         hbaseTables = getHBaseTables();
--- a/phoenix-core/src/it/java/org/apache/phoenix/end2end/MigrateSystemTablesToSystemNamespaceIT.java
+++ b/phoenix-core/src/it/java/org/apache/phoenix/end2end/MigrateSystemTablesToSystemNamespaceIT.java
@@ -1,0 +1,399 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to you under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.phoenix.end2end;
+
+import org.apache.hadoop.conf.Configuration;
+import org.apache.hadoop.hbase.HBaseTestingUtility;
+import org.apache.hadoop.hbase.TableName;
+import org.apache.hadoop.security.UserGroupInformation;
+import org.apache.phoenix.coprocessor.MetaDataProtocol;
+import org.apache.phoenix.jdbc.PhoenixConnection;
+import org.apache.phoenix.jdbc.PhoenixDatabaseMetaData;
+import org.apache.phoenix.query.BaseTest;
+import org.apache.phoenix.query.ConnectionQueryServices;
+import org.apache.phoenix.query.ConnectionQueryServicesImpl;
+import org.apache.phoenix.query.QueryConstants;
+import org.apache.phoenix.query.QueryServices;
+import org.apache.phoenix.schema.PTableType;
+import org.apache.phoenix.util.ReadOnlyProps;
+import org.apache.phoenix.util.SchemaUtil;
+import org.junit.After;
+import org.junit.Before;
+import org.junit.Test;
+import org.junit.experimental.categories.Category;
+
+import java.io.IOException;
+import java.security.PrivilegedExceptionAction;
+import java.sql.Connection;
+import java.sql.DriverManager;
+import java.sql.PreparedStatement;
+import java.sql.ResultSet;
+import java.sql.SQLException;
+import java.sql.Statement;
+import java.util.Arrays;
+import java.util.HashSet;
+import java.util.Map;
+import java.util.Properties;
+import java.util.Set;
+
+import static org.junit.Assert.*;
+
+@Category(NeedsOwnMiniClusterTest.class)
+public class MigrateSystemTablesToSystemNamespaceIT extends BaseTest {
+
+    private static final Set<String> PHOENIX_SYSTEM_TABLES = new HashSet<>(Arrays.asList(
+            "SYSTEM.CATALOG", "SYSTEM.SEQUENCE", "SYSTEM.STATS", "SYSTEM.FUNCTION",
+            "SYSTEM.MUTEX"));
+    private static final Set<String> PHOENIX_NAMESPACE_MAPPED_SYSTEM_TABLES = new HashSet<>(
+            Arrays.asList("SYSTEM:CATALOG", "SYSTEM:SEQUENCE", "SYSTEM:STATS", "SYSTEM:FUNCTION",
+                    "SYSTEM:MUTEX"));
+    private static final String SCHEMA_NAME = "MIGRATETEST";
+    private static final String TABLE_NAME =
+            SCHEMA_NAME + "." + MigrateSystemTablesToSystemNamespaceIT.class.getSimpleName().toUpperCase();
+    private static final int NUM_RECORDS = 5;
+
+    private HBaseTestingUtility testUtil = null;
+    private Set<String> hbaseTables;
+
+    // Create Multiple users since Phoenix caches the connection per user
+    // Migration or upgrade code will run every time for each user.
+    final UserGroupInformation user1 =
+            UserGroupInformation.createUserForTesting("user1", new String[0]);
+    final UserGroupInformation user2 =
+            UserGroupInformation.createUserForTesting("user2", new String[0]);
+    final UserGroupInformation user3 =
+            UserGroupInformation.createUserForTesting("user3", new String[0]);
+    final UserGroupInformation user4 =
+            UserGroupInformation.createUserForTesting("user4", new String[0]);
+
+
+    @Before
+    public final void doSetup() throws Exception {
+        testUtil = new HBaseTestingUtility();
+        Configuration conf = testUtil.getConfiguration();
+        enableNamespacesOnServer(conf);
+        testUtil.startMiniCluster(1);
+    }
+
+    @After
+    public void tearDownMiniCluster() {
+        try {
+            if (testUtil != null) {
+                testUtil.shutdownMiniCluster();
+                testUtil = null;
+            }
+        } catch (Exception e) {
+            // ignore
+        }
+    }
+
+    // Tests that client can create and read tables on a fresh HBase cluster with
+    // system namespace mapping enabled from the start
+    @Test
+    public void freshClientsCreateNamespaceMappedSystemTables() throws IOException, InterruptedException {
+
+        user1.doAs(new PrivilegedExceptionAction<Void>() {
+            @Override
+            public Void run() throws Exception {
+                createConnection(getClientPropertiesWithSystemMappingEnabled());
+                createTable(getClientPropertiesWithSystemMappingEnabled());
+                return null;
+            }
+        });
+
+        hbaseTables = getHBaseTables();
+        assertTrue(hbaseTables.containsAll(PHOENIX_NAMESPACE_MAPPED_SYSTEM_TABLES));
+
+        user1.doAs(new PrivilegedExceptionAction<Void>() {
+            @Override
+            public Void run() throws Exception {
+                createConnection(getClientPropertiesWithSystemMappingEnabled());
+                readTable(getClientPropertiesWithSystemMappingEnabled());
+                return null;
+            }
+        });
+
+    }
+
+    // Tests that NEWER clients can read tables on HBase cluster after system tables are migrated
+    @Test
+    public void migrateSystemTablesInExistingCluster() throws IOException, InterruptedException {
+
+        user1.doAs(new PrivilegedExceptionAction<Void>() {
+            @Override
+            public Void run() throws Exception {
+                createConnection(getClientPropertiesWithSystemMappingDisabled());
+                createTable(getClientPropertiesWithSystemMappingDisabled());
+                return null;
+            }
+        });
+
+        hbaseTables = getHBaseTables();
+        assertTrue(hbaseTables.containsAll(PHOENIX_SYSTEM_TABLES));
+
+        user2.doAs(new PrivilegedExceptionAction<Void>() {
+            @Override
+            public Void run() throws Exception {
+                createConnection(getClientPropertiesWithSystemMappingEnabled());
+                readTable(getClientPropertiesWithSystemMappingEnabled());
+                return null;
+            }
+        });
+
+        hbaseTables = getHBaseTables();
+        assertTrue(hbaseTables.containsAll(PHOENIX_NAMESPACE_MAPPED_SYSTEM_TABLES));
+    }
+
+    // Tests that OLDER clients fail after system tables are migrated
+    // Clients should be restarted with new properties which are consistent on both client and server
+    @Test
+    public void oldClientsAfterSystemTableMigrationShouldFail() throws IOException, InterruptedException {
+
+        user1.doAs(new PrivilegedExceptionAction<Void>() {
+            @Override
+            public Void run() throws Exception {
+                createConnection(getClientPropertiesWithSystemMappingEnabled());
+                return null;
+            }
+        });
+
+        hbaseTables = getHBaseTables();
+        assertTrue(hbaseTables.size() == PHOENIX_NAMESPACE_MAPPED_SYSTEM_TABLES.size());
+        assertTrue(hbaseTables.containsAll(PHOENIX_NAMESPACE_MAPPED_SYSTEM_TABLES));
+
+        try {
+            user2.doAs(new PrivilegedExceptionAction<Void>() {
+                @Override
+                public Void run() throws Exception {
+                    createConnection(getClientPropertiesWithSystemMappingDisabled());
+                    return null;
+                }
+            });
+            fail("Client should not be able to connect to cluster with inconsistent SYSTEM table namespace properties");
+        } catch (Exception e) {
+            //ignore
+        }
+
+        hbaseTables = getHBaseTables();
+        assertTrue(hbaseTables.size() == PHOENIX_NAMESPACE_MAPPED_SYSTEM_TABLES.size());
+        assertTrue(hbaseTables.containsAll(PHOENIX_NAMESPACE_MAPPED_SYSTEM_TABLES));
+    }
+
+    // Tests that only one client can migrate the system table to system namespace
+    // Migrate process acquires lock in SYSMUTEX table
+    @Test
+    public void onlyOneClientCanMigrate() throws IOException, InterruptedException, SQLException {
+
+        user1.doAs(new PrivilegedExceptionAction<Void>() {
+            @Override
+            public Void run() throws Exception {
+                createConnection(getClientPropertiesWithSystemMappingDisabled());
+                return null;
+            }
+        });
+
+        hbaseTables = getHBaseTables();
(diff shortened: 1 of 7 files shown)


karanmehta93:
@twdsilva Please review.
@aertoria

twdsilva:
@ankitsinghal @samarthjain @joshelser 
in case you guys want to take a look as well

phoenix-core/src/it/java/org/apache/phoenix/end2end/MigrateSystemTablesToSystemNamespaceIT.java
```diff
@@ -0,0 +1,399 @@
...
+
+        hbaseTables = getHBaseTables();
+        assertTrue(hbaseTables.size() == PHOENIX_SYSTEM_TABLES.size());
+        assertTrue(hbaseTables.containsAll(PHOENIX_SYSTEM_TABLES));
+
+        try {
+            user3.doAs(new PrivilegedExceptionAction<Void>() {
+                @Override
+                public Void run() throws Exception {
+                    createConnection(getClientPropertiesWithSystemMappingEnabled());
```
twdsilva: Can you verify you get a UpgradeInProgressException?

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/MetaDataProtocol.java
```diff
@@ -95,6 +97,7 @@
     // Key is the SYSTEM.CATALOG timestamp for the version and value is the version string.
     private static final NavigableMap<Long, String> TIMESTAMP_VERSION_MAP = new TreeMap<>();
     static {
+        TIMESTAMP_VERSION_MAP.put(MIN_SYSTEM_TABLE_MIGRATION_TIMESTAMP, MIGRATION_IN_PROGRESS);
```
twdsilva: Is this only used in a test? If so can you change the test to use MIN_SYSTEM_TABLE_TIMESTAMP?

phoenix-core/src/main/java/org/apache/phoenix/query/ConnectionQueryServicesImpl.java
```diff
@@ -2905,14 +2918,18 @@ public void upgradeSystemTables(final String url, final Properties props) throws
                     }
                 } finally {
                     if (acquiredMutexLock) {
-                        releaseUpgradeMutex(mutexRowKey);
+                        try {
+                            releaseUpgradeMutex(mutexRowKey);
+                        } catch (IOException e) {
+                            logger.warn("Release of upgrade mutex failed ", e);
```
twdsilva: Is it ok to continue on if we fail to release the upgrade mutex, or should we just let the exception be thrown?

phoenix-core/src/main/java/org/apache/phoenix/query/ConnectionQueryServicesImpl.java
```diff
@@ -3101,49 +3124,68 @@ void ensureSystemTablesUpgraded(ReadOnlyProps props)
...
                             PhoenixDatabaseMetaData.SYSTEM_CATALOG_NAME, props, null, PTableType.SYSTEM,
                             null);
+                    // Invalidate the client-side metadataCache
                     ConnectionQueryServicesImpl.this.removeTable(null,
                             PhoenixDatabaseMetaData.SYSTEM_CATALOG_NAME, null,
                             MetaDataProtocol.MIN_SYSTEM_TABLE_TIMESTAMP_4_1_0);
                 }
                 tableNames.remove(PhoenixDatabaseMetaData.SYSTEM_CATALOG_HBASE_TABLE_NAME);
             }
-            tableNames.remove(PhoenixDatabaseMetaData.SYSTEM_MUTEX_HBASE_TABLE_NAME);
```
twdsilva: I think you still need to remove SYSTEM.MUTEX from tableNames.

phoenix-core/src/main/java/org/apache/phoenix/query/ConnectionQueryServicesImpl.java
```diff
@@ -3101,49 +3124,68 @@ void ensureSystemTablesUpgraded(ReadOnlyProps props)
...
                         null);
                 ConnectionQueryServicesImpl.this.removeTable(null, table.getNameAsString(), null,
                         MetaDataProtocol.MIN_SYSTEM_TABLE_TIMESTAMP_4_1_0);
             }
-            if (!tableNames.isEmpty()) {
-                clearCache();
-            }
+
+            // Clear the server-side metadataCache when all tables are migrated so that the new PTable can be loaded with NS mapping
+            clearCache();
```
aertoria: What does clearCache() here do? Does it clear PMeta cache at the server side?

phoenix-core/src/main/java/org/apache/phoenix/query/ConnectionQueryServicesImpl.java
```diff
@@ -3101,49 +3124,68 @@ void ensureSystemTablesUpgraded(ReadOnlyProps props)
...
         } finally {
             if (metatable != null) {
                 metatable.close();
             }
+            if(acquiredMutexLock) {
```
aertoria: Why wasn't this  "UpgradeUtil.mapTableToNamespace" synchronized in the existing code?

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/MetaDataProtocol.java
karanmehta93: Nope, this one is not only for test. 

When you acquire a lock on SYSMUTEX, the method requires you to pass in a server side timestamp of SYSCAT table (Since it was meant for that earlier) and it checks if the current timestamp is less than the MIN_SYSTEM_TABLE_TIMESTAMP before upgrading it.
I can't use MIN_SYSTEM_TABLE_TIMESTAMP because then the code for `acquireUpgradeMutex()` wont allow me to acquire the mutex lock.
I use the same `UpgradeInProgressException` to prevent multiple clients migrating the system tables together. I use this value in `UpgradeInProgressException` to change the error message displayed to the user.

phoenix-core/src/main/java/org/apache/phoenix/query/ConnectionQueryServicesImpl.java
karanmehta93: I think it is fine not to throw the exception because we have a TTL on the table. Typically it should not matter to the client as such. Moreover, similar behavior is also defined the code that Upgrades the SYSCAT table.

phoenix-core/src/main/java/org/apache/phoenix/query/ConnectionQueryServicesImpl.java
karanmehta93: if (!tableNames.isEmpty()) {
                clearCache();
            }

I don't really understand why it was done this way, but since we are not removing entries from the `tableNames` variable, this piece of code will always get executed and clear the server side cache.
Also, we clear server side cache anytime a SYSTEM table is disabled, using a coprocessor hook.

So I decided to simplify the logic and clear the cache every time. Its good to do that because the server side cache will go out of sync with the SYSCAT table once the migration happens. 

For example, if the cached Ptable represented SYSTEM.CATALOG (IS_NAMESPACE_MAPPED=false), the new Ptable should represent SYSTEM:CATALOG (IS_NAMESPACE_MAPPED=true). This can only happen if the cache entry for that table is invalidated.

phoenix-core/src/main/java/org/apache/phoenix/query/ConnectionQueryServicesImpl.java
karanmehta93: @joshelser @ankitsinghal Can probably comment on this one.

phoenix-core/src/main/java/org/apache/phoenix/query/ConnectionQueryServicesImpl.java
```diff
@@ -2514,7 +2529,7 @@ private void createSysMutexTable(HBaseAdmin admin) throws IOException, SQLExcept
         return Lists.newArrayList(admin.listTableNames(QueryConstants.SYSTEM_SCHEMA_NAME + "\\..*"));
     }
 
-    private void createOtherSystemTables(PhoenixConnection metaConnection) throws SQLException {
+    private void createOtherSystemTables(PhoenixConnection metaConnection, HBaseAdmin hBaseAdmin) throws SQLException {
```
joshelser: nit: s/hBaseAdmin/hbaseAdmin/ camel-case typically isn't observed on "hbase" ;)

phoenix-core/src/main/java/org/apache/phoenix/query/ConnectionQueryServicesImpl.java
```diff
@@ -2526,8 +2541,14 @@ private void createOtherSystemTables(PhoenixConnection metaConnection) throws SQ
         try {
             metaConnection.createStatement().execute(QueryConstants.CREATE_FUNCTION_METADATA);
         } catch (TableAlreadyExistsException ignore) {}
+        // We catch TableExistsException in createSysMutexTable() and ignore it. Hence we will also ignore IOException here.
+        // SYSTEM.MUTEX table should not be exposed to user. Hence it is directly created and used via HBase API.
+        // Using 'CREATE TABLE' statement will add entries to SYSTEM.CATALOG table, which should not happen.
+        try {
+            createSysMutexTable(hBaseAdmin, ConnectionQueryServicesImpl.this.getProps());
+        } catch (IOException ignore) {}
```
joshelser: What about the case where the table creation failed for a legitimate system reason (something an admin needs to correct)? We should at least have some warning to the user that we failed to created the table, right?

phoenix-core/src/main/java/org/apache/phoenix/query/ConnectionQueryServicesImpl.java
```diff
@@ -3086,12 +3103,18 @@ private void restoreFromSnapshot(String tableName, String snapshotName,
...
+
+        boolean acquiredMutexLock = false;
+        byte[] mutexRowKey = SchemaUtil.getTableKey(null, PhoenixDatabaseMetaData.SYSTEM_CATALOG_SCHEMA,
+                PhoenixDatabaseMetaData.SYSTEM_CATALOG_TABLE);
+
         HTableInterface metatable = null;
         try (HBaseAdmin admin = getAdmin()) {
-            // Namespace-mapping is enabled at this point.
+             // SYSTEM namespace needs to be created via HBase API's because "CREATE SCHEMA" statement tries to write its metadata
+             // in SYSTEM:CATALOG table. Without SYSTEM namespace, SYSTEM:CATALOG table cannot be created.
```
joshelser: nice comments here and elsewhere in this method.

phoenix-core/src/main/java/org/apache/phoenix/query/ConnectionQueryServicesImpl.java
```diff
@@ -3101,49 +3124,72 @@ void ensureSystemTablesUpgraded(ReadOnlyProps props)
...
             if (tableNames.size() > 5) {
                 logger.warn("Expected 5 system tables but found " + tableNames.size() + ":" + tableNames);
             }
+
+            // Try acquiring a lock in SYSMUTEX table before migrating the tables since it involves disabling the table
+            // If we cannot acquire lock, it means some old client is either migrating SYSCAT or trying to upgrade the
+            // schema of SYSCAT table and hence it should not be interrupted
+            acquiredMutexLock = acquireUpgradeMutex(0, mutexRowKey);
+            if(acquiredMutexLock) {
+                logger.debug("Acquired lock in SYSMUTEX table for migrating SYSTEM tables to SYSTEM namespace");
```
joshelser: Log a debug message for the `else` case too? Easier than finding the lack of the above message when debugging..

phoenix-core/src/main/java/org/apache/phoenix/query/ConnectionQueryServicesImpl.java
joshelser: >  Also, we clear server side cache anytime a SYSTEM table is disabled, using a coprocessor hook.

My assumption was that the intent was to drop any cached PTables if we moved any system tables around. Given the above, it sounds like this is (now?) unnecessary.

> So I decided to simplify the logic and clear the cache every time. Its good to do that because the server side cache will go out of sync with the SYSCAT table once the migration happens.

Is that being down in a CP too? If so, would it make sense to do this cache-clearing there (instead of client side)?

phoenix-core/src/main/java/org/apache/phoenix/query/ConnectionQueryServicesImpl.java
joshelser: I have no good explanation for why it was the way it was when I last looked at it. I can only assumed I left it as it was (really, just extracted part of this one method into another). If you think it should also be marked synchronized, go for it.

phoenix-core/src/main/java/org/apache/phoenix/query/ConnectionQueryServicesImpl.java
```diff
@@ -3195,6 +3255,18 @@ public boolean releaseUpgradeMutex(byte[] mutexRowKey) {
         return released;
     }
 
+    private byte[] getSysMutexPhysicalTableNameBytes() throws IOException, SQLException {
+        byte[] sysMutexPhysicalTableNameBytes = null;
+        try(HBaseAdmin admin = getAdmin()) {
+            if(admin.tableExists(PhoenixDatabaseMetaData.SYSTEM_MUTEX_NAME_BYTES)) {
```
joshelser: I assume `PhoenixDatabaseMetaData.SYSTEM_MUTEX_NAME_BYTES` is `byte[]`? Would be better to use the `TableName` variable to avoid deprecated API warning.

phoenix-core/src/main/java/org/apache/phoenix/query/ConnectionQueryServicesImpl.java
```diff
@@ -3101,49 +3124,72 @@ void ensureSystemTablesUpgraded(ReadOnlyProps props)
...
             if (tableNames.size() == 0) { return; }
             // Try to move any remaining tables matching "SYSTEM\..*" into "SYSTEM:"
             if (tableNames.size() > 5) {
                 logger.warn("Expected 5 system tables but found " + tableNames.size() + ":" + tableNames);
             }
+
+            // Try acquiring a lock in SYSMUTEX table before migrating the tables since it involves disabling the table
+            // If we cannot acquire lock, it means some old client is either migrating SYSCAT or trying to upgrade the
+            // schema of SYSCAT table and hence it should not be interrupted
+            acquiredMutexLock = acquireUpgradeMutex(0, mutexRowKey);
```
twdsilva: Change 0 to MIN_SYSTEM_TABLE_MIGRATION_TIMESTAMP

phoenix-core/src/it/java/org/apache/phoenix/end2end/MigrateSystemTablesToSystemNamespaceIT.java
karanmehta93: Added this in the new commit.

phoenix-core/src/main/java/org/apache/phoenix/query/ConnectionQueryServicesImpl.java
karanmehta93: Added this in the new commit.

phoenix-core/src/main/java/org/apache/phoenix/query/ConnectionQueryServicesImpl.java
karanmehta93: > My assumption was that the intent was to drop any cached PTables if we moved any system tables around. Given the above, it sounds like this is (now?) unnecessary.
That's correct. We indeed need to remove it from cache because once the migration is complete, we add a row in `SYSCAT` indicating that `IS_NAMESPACE_MAPPED` to true. Although it seems unnecessary, I don't really want to rely on the coprocessor hook as it may change any time. An explicit `clearCache()` should not harm it, I believe.

> Is that being down in a CP too? If so, would it make sense to do this cache-clearing there (instead of client side)?
 What is CP? I didn't quite get you.

phoenix-core/src/main/java/org/apache/phoenix/query/ConnectionQueryServicesImpl.java
karanmehta93: According to my understanding and discussions with @twdsilva, I believe that multiple clients should not be allowed to carry on this portion of code simultaneously. For example, a client is trying to upgrade the SYSCAT version and another client connects with the property `phoenix.schema.mapSystemTablesToNamespace` to true. The latter client might disable SYSCAT and enable it to SYS:CAT, in such a case the client operations from former client might leave SYSCAT inconsistent.
@joshelser Let me know your thoughts about it.

phoenix-core/src/main/java/org/apache/phoenix/query/ConnectionQueryServicesImpl.java
karanmehta93: Will do.

phoenix-core/src/main/java/org/apache/phoenix/query/ConnectionQueryServicesImpl.java
joshelser: If there isn't any mutual exclusion happening at a higher level, most definitely we should make sure this isn't happening concurrently.

What about clients in separate JVMs though? What's the "prior art" for the rest of the SYSTEM tables?

phoenix-core/src/main/java/org/apache/phoenix/query/ConnectionQueryServicesImpl.java
karanmehta93: Will do.

phoenix-core/src/main/java/org/apache/phoenix/query/ConnectionQueryServicesImpl.java
karanmehta93: The code will never reach to the else part. If the mutex lock fails to get acquired, it will throw `UpgradeInProgressException` and the code will return back, leaving for the client to retry again.

phoenix-core/src/main/java/org/apache/phoenix/query/ConnectionQueryServicesImpl.java
karanmehta93: This SYSMUTEX table is in place to make sure that clients in separate JVM's don't carry out the same task at the same time. I think @twdsilva and @samarthjain jain discovered the bug where multiple clients tried to upgrade SYSCAT at the same time and that is when the SYSMUTEX table was introduced. I feel that a similar race condition can be triggered over here, so a mutual exclusion is required, which is done via SYSMUTEX table. 
Does this answer your question?

phoenix-core/src/main/java/org/apache/phoenix/query/ConnectionQueryServicesImpl.java
karanmehta93: Added it in a new commit.

phoenix-core/src/main/java/org/apache/phoenix/query/ConnectionQueryServicesImpl.java
karanmehta93: Added it in a new commit.

phoenix-core/src/main/java/org/apache/phoenix/query/ConnectionQueryServicesImpl.java
karanmehta93: Thanks!

phoenix-core/src/main/java/org/apache/phoenix/query/ConnectionQueryServicesImpl.java
karanmehta93: Added it in a new commit.

phoenix-core/src/main/java/org/apache/phoenix/query/ConnectionQueryServicesImpl.java
joshelser: Yup, fine by me. It seems to me to be the same as it was previously :) The mutex table update would provide exclusion for both same and different JVM cases.

phoenix-core/src/main/java/org/apache/phoenix/query/ConnectionQueryServicesImpl.java
karanmehta93: Yes, this is a serious case that we should discuss. The correct thing to do is to probably FAIL the connection itself or possibly have a retry logic for creating the table. 

This is because in the `acquireUpgradeMutex()` method we check if either SYSMUTEX exists or SYS:MUTEX table exists. The only possible case where both of those tables can be missing is when a client is trying to migrate the table, which disables the old table and creates the new one. There is a brief period of time when none of these table exists. Hence we throw `UpgradeInProgressException` exception in such a case.

We have no way to determine if the table doesn't exist at well v/s the point that the table is in migration.
Is there any other scenario in which this can affect?

phoenix-core/src/main/java/org/apache/phoenix/query/ConnectionQueryServicesImpl.java
karanmehta93: Earlier, we didn't have any locking during table migration. So its a race condition which can lead to unexpected results.

karanmehta93:
> Looks alright to me, but I don't have a lot of context to the intricacy of the change. 

It would be great if you can apply this patch locally and try out a sqlline based end to end testing.

> Thanks for your hard work here, Karan!

Thank you and Thanks for your inputs!

phoenix-core/src/main/java/org/apache/phoenix/query/ConnectionQueryServicesImpl.java
karanmehta93: @twdsilva @aertoria Any thoughts?

aertoria: I think in those case you can just ignore the exception that you want to tolerate. So at least SQLException should be bubbled up.

twdsilva: No need to catch the IOException, just let it bubble up so that the user knows there is an issue.

karanmehta93:
These commits also resolve PHOENIX-4227.

twdsilva:
+1 LGTM 

Thanks for the excellent work.

PR #283, opened 2017-11-17
PHOENIX-672 Add GRANT and REVOKE commands using HBase AccessController

karanmehta93:
Initial patch
Added support for Grant Revoke SQL statements for HBase ACL Permissions
Refactored the tests, BasePermissionsIT is the base test for Permission management tests and provides helper methods. 
Index tables will be automatically granted permissions as base table when created

Ongoing
Removing console debug messages, adding log lines, making patch cleaner
End to end testing with HBase and sqlline.


karanmehta93:
@twdsilva @aertoria @ankitsinghal @joshelser 
Please review.
@JamesRTaylor @apurtell FYI.

karanmehta93:
@samarthjain Please have a look.

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/PhoenixAccessController.java
```diff
@@ -229,17 +227,12 @@ public void handleRequireAccessOnDependentTable(String request, String userName,
                     + dependentTable);
             return;
         }
-        if (isAutomaticGrantEnabled) {
```
twdsilva: Does it make sense to also remove the isStrictMode option? We should always keep the index and data table permissions in sync.

karanmehta93: What is the isStrictMode option?

twdsilva: In handleRequireAccessOnDependentTable if the isStrictMode boolean is set to false then the automatic grant code path is skipped.

phoenix-core/src/main/java/org/apache/phoenix/schema/MetaDataClient.java
```diff
@@ -4168,4 +4176,197 @@ public MutationState useSchema(UseSchemaStatement useSchemaStatement) throws SQL
...
+        }
+        logger.info(revokePermLog.toString());
+
+        HConnection hConnection = connection.getQueryServices().getAdmin().getConnection();
+
+        try {
+            if (revokeStatement.getSchemaName() != null) {
+                // SYSTEM.CATALOG doesn't have any entry for "default" HBase namespace, hence we will bypass the check
+                if(!revokeStatement.getSchemaName().equals(QueryConstants.HBASE_DEFAULT_SCHEMA_NAME)) {
+                    FromCompiler.getResolverForSchema(revokeStatement.getSchemaName(), connection);
```
twdsilva: Does FromCompiler.getResolverForSchema actually check to see if the schema exists?

phoenix-core/src/main/java/org/apache/phoenix/schema/MetaDataClient.java
```diff
@@ -4168,4 +4176,197 @@ public MutationState useSchema(UseSchemaStatement useSchemaStatement) throws SQL
...
+        revokePermissionsFromTable(hConnection, revokeStatement, tableName);
+
+        for(PTable indexTable : inputTable.getIndexes()) {
+            // Local Indexes don't correspond to new physical table, they are just stored in separate CF of base table.
+            if(indexTable.getIndexType().equals(IndexType.LOCAL)) {
+                continue;
+            }
+            logger.info("Revoking perms from IndexTable: " + indexTable.getName() + " BaseTable: " + inputTable.getName());
+            if (inputTable.isNamespaceMapped() != indexTable.isNamespaceMapped()) {
+                throw new TablesNotInSyncException(inputTable.getTableName().getString(),
```
twdsilva: Can this ever happen (base table is namespace mapped but index table is not)?

phoenix-core/src/main/java/org/apache/phoenix/schema/MetaDataClient.java
```diff
@@ -4168,4 +4176,197 @@ public MutationState useSchema(UseSchemaStatement useSchemaStatement) throws SQL
...
+            }
+            tableName = SchemaUtil.getPhysicalTableName(indexTable.getName().getBytes(), indexTable.isNamespaceMapped());
+            revokePermissionsFromTable(hConnection, revokeStatement, tableName);
+        }
+
+        byte[] viewIndexTableBytes = MetaDataUtil.getViewIndexPhysicalName(inputTable.getPhysicalName().getBytes());
+        tableName = org.apache.hadoop.hbase.TableName.valueOf(viewIndexTableBytes);
+        boolean viewIndexTableExists = connection.getQueryServices().getAdmin().tableExists(tableName);
+        if(!viewIndexTableExists && inputTable.isMultiTenant()) {
+            logger.error("View Index Table not found for MultiTenant Table: " + inputTable.getName());
```
twdsilva: Not sure if this check is needed as we always create the view index table for multi-tenant tables.

phoenix-core/src/test/java/org/apache/phoenix/parse/QueryParserTest.java
```diff
@@ -66,6 +68,46 @@ private void parseQueryThatShouldFail(String sql) throws Exception {
         }
     }
 
+    @Test
```
twdsilva: Do the permissions/user/group have to be in single quotes?

twdsilva:
ping @ankitsinghal 

Karan's patch removes the automatic grant option and always keeps the index and data tables in sync. Can you please take a look as well.

phoenix-core/src/it/java/org/apache/phoenix/end2end/BasePermissionsIT.java
```diff
@@ -0,0 +1,635 @@
...
+        if (testUtil != null) {
+            testUtil.shutdownMiniCluster();
+            testUtil = null;
+        }
+    }
+
+    public static HBaseTestingUtility getUtility(){
+        return testUtil;
+    }
+
```
twdsilva: Instead of directly calling the AccessControlClient can the tests just use the GRANT or REVOKE statements?

phoenix-core/src/main/java/org/apache/phoenix/schema/MetaDataClient.java
ankitsinghal: I don't think it can happen. indexes should go in the same namespace in which data table is.

phoenix-core/src/main/java/org/apache/phoenix/schema/MetaDataClient.java
```diff
@@ -4168,4 +4176,197 @@ public MutationState useSchema(UseSchemaStatement useSchemaStatement) throws SQL
...
+            // Local Indexes don't correspond to new physical table, they are just stored in separate CF of base table.
+            if(indexTable.getIndexType().equals(IndexType.LOCAL)) {
+                continue;
+            }
+            logger.info("Revoking perms from IndexTable: " + indexTable.getName() + " BaseTable: " + inputTable.getName());
+            if (inputTable.isNamespaceMapped() != indexTable.isNamespaceMapped()) {
+                throw new TablesNotInSyncException(inputTable.getTableName().getString(),
+                        indexTable.getTableName().getString(), "Namespace properties");
+            }
+            tableName = SchemaUtil.getPhysicalTableName(indexTable.getName().getBytes(), indexTable.isNamespaceMapped());
```
ankitsinghal: you could use indexTable.getPhysicalName api instead.

phoenix-core/src/main/java/org/apache/phoenix/schema/MetaDataClient.java
```diff
@@ -4168,4 +4176,197 @@ public MutationState useSchema(UseSchemaStatement useSchemaStatement) throws SQL
...
+            revokePermissionsFromTable(hConnection, revokeStatement, tableName);
+        }
+
+    }
+
+    private void revokePermissionsFromTable(HConnection hConnection, RevokeStatement revokeStatement, org.apache.hadoop.hbase.TableName tableName)
+            throws Throwable {
+        AccessControlClient.revoke(hConnection, tableName, revokeStatement.getName(),
+                null, null, Permission.Action.values());
+    }
```
ankitsinghal: Avoid using deprecated classes like HConnection.

phoenix-core/src/main/java/org/apache/phoenix/schema/MetaDataClient.java
```diff
@@ -4168,4 +4176,197 @@ public MutationState useSchema(UseSchemaStatement useSchemaStatement) throws SQL
...
+                throw new TablesNotInSyncException(inputTable.getTableName().getString(),
+                        indexTable.getTableName().getString(), "Namespace properties");
+            }
+            tableName = SchemaUtil.getPhysicalTableName(indexTable.getName().getBytes(), indexTable.isNamespaceMapped());
+            revokePermissionsFromTable(hConnection, revokeStatement, tableName);
+        }
+
+        byte[] viewIndexTableBytes = MetaDataUtil.getViewIndexPhysicalName(inputTable.getPhysicalName().getBytes());
+        tableName = org.apache.hadoop.hbase.TableName.valueOf(viewIndexTableBytes);
+        boolean viewIndexTableExists = connection.getQueryServices().getAdmin().tableExists(tableName);
```
ankitsinghal: Admin is not closed.

phoenix-core/src/main/java/org/apache/phoenix/schema/MetaDataClient.java
```diff
@@ -4168,4 +4176,197 @@ public MutationState useSchema(UseSchemaStatement useSchemaStatement) throws SQL
...
+                continue;
+            }
+            logger.info("Granting " + Arrays.toString(grantStatement.getPermsList()) +
+                    " perms to IndexTable: " + indexTable.getName() + " BaseTable: " + inputTable.getName());
+            if (inputTable.isNamespaceMapped() != indexTable.isNamespaceMapped()) {
+                throw new TablesNotInSyncException(inputTable.getTableName().getString(),
+                        indexTable.getTableName().getString(), "Namespace properties");
+            }
+            tableName = SchemaUtil.getPhysicalTableName(indexTable.getName().getBytes(), indexTable.isNamespaceMapped());
+            grantPermissionsToTable(hConnection, grantStatement, tableName);
```
ankitsinghal: you can have a common method to give you set of physical tables on which revoke and grant is required.

phoenix-core/src/main/java/org/apache/phoenix/schema/TablesNotInSyncException.java
```diff
@@ -0,0 +1,16 @@
...
+
+public class TablesNotInSyncException extends SQLException {
+    private static final long serialVersionUID = 1L;
+    private static SQLExceptionCode code = SQLExceptionCode.TABLES_NOT_IN_SYNC;
+
+    public TablesNotInSyncException(String table1, String table2, String diff) {
+        super(new SQLExceptionInfo.Builder(code).setMessage("Table: " + table1 + " and Table: " + table2 + " differ in " + diff).build().toString(), code.getSQLState(), code.getErrorCode());
+    }
+
+}
```
ankitsinghal: we can add some comments here.

phoenix-core/src/test/java/org/apache/phoenix/parse/QueryParserTest.java
```diff
@@ -66,6 +68,46 @@ private void parseQueryThatShouldFail(String sql) throws Exception {
...
+        String sql3 = "GRANT 'CA' ON SCHEMA some_schema2 TO 'user2'";
+        parseQuery(sql3);
+        String sql4 = "GRANT 'RXW' ON some_table3 TO GROUP 'group3'";
+        parseQuery(sql4);
+        String sql5 = "GRANT 'RXW' ON \"some_schema5\".\"some_table5\" TO GROUP 'group5'";
+        parseQuery(sql5);
+        String sql6 = "GRANT 'RWA' TO 'user6'";
+        parseQuery(sql6);
+        String sql7 = "GRANT 'A' TO GROUP 'group7'";
+        parseQuery(sql7);
```
ankitsinghal: Do we have validation on permission type like fail when permission is other than 'RWXCA' at parsing side?

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/PhoenixAccessController.java
ankitsinghal: I think , we can keep this option. as it will help admin if he doesn't want to see the permissions problems on the dependent table and wants to manage the permission by himself once the table is created(with downtime) and don't want to rely on the automatic grant.

phoenix-core/src/main/java/org/apache/phoenix/schema/MetaDataClient.java
```diff
@@ -4168,4 +4176,197 @@ public MutationState useSchema(UseSchemaStatement useSchemaStatement) throws SQL
...
+            } else if (grantStatement.getTableName() != null) {
+                PTable inputTable = PhoenixRuntime.getTable(connection,
+                        SchemaUtil.normalizeFullTableName(grantStatement.getTableName().toString()));
+                if (!(PTableType.TABLE.equals(inputTable.getType()) || PTableType.SYSTEM.equals(inputTable.getType()))) {
+                    throw new AccessDeniedException("Cannot GRANT permissions on INDEX TABLES or VIEWS");
+                }
+                grantPermissionsToTables(hConnection, grantStatement, inputTable);
+
+            } else {
+                grantPermissionsToUser(hConnection, grantStatement);
```
ankitsinghal: If for some reason grant doesn't succeed for all the tables. so do we have plan to give construct like "SHOW GRANTS" or something to the user to know what all grants are still there for the user or on the table.

phoenix-core/src/it/java/org/apache/phoenix/end2end/TableDDLPermissionsIT.java
```diff
@@ -210,25 +98,20 @@ public Void run() throws Exception {
...
+            verifyAllowed(dropSchema(schemaName), regularUser1);
         } finally {
             revokeAll();
         }
     }
 
     @Test
-    public void testAutomaticGrantDisabled() throws Throwable{
-        testIndexAndView(false);
-    }
```
ankitsinghal: Why is this removed?

phoenix-core/src/it/java/org/apache/phoenix/end2end/BasePermissionsIT.java
karanmehta93: This is the base class for all Permission Tests. Checkout the method on Line 258. The original tests have been just refactored; GrantRevokePermissionsIT contains the required tests for this JIRA.

phoenix-core/src/it/java/org/apache/phoenix/end2end/TableDDLPermissionsIT.java
karanmehta93: I removed the option of AUTOMATIC_GRANT and set it defaulted to true.

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/PhoenixAccessController.java
karanmehta93: I didn't quite clearly understand the difference between automatic grant and strict mode. Could you please elaborate @ankitsinghal?

twdsilva: This patch removes the automatic grant option and now always keep the permissions in sync between the data table and any index tables. You can only execute a GRANT/REVOKE on the data table and not on the index table. So I don't think we need the isStrictMode option any more. Once that is removed the data table and index table permissions for all uses will always be kept in sync.

phoenix-core/src/main/java/org/apache/phoenix/schema/MetaDataClient.java
karanmehta93: Yes, I have been working on it. Will commit it here soon. Current approach is to log an error message for all the tables whose permission assignment has failed. Does that seem fine?

phoenix-core/src/main/java/org/apache/phoenix/schema/MetaDataClient.java
karanmehta93: Yeah that's a good idea. Let me refactor the code.

phoenix-core/src/main/java/org/apache/phoenix/schema/MetaDataClient.java
karanmehta93: Yes, it tries to resolve the schema immediately and throws an exception if Schema is not found. 
`return new SchemaResolver(connection, SchemaUtil.normalizeIdentifier(schema), updateCacheImmediately: true);`

phoenix-core/src/main/java/org/apache/phoenix/schema/MetaDataClient.java
karanmehta93: Sure. Will refactor.

phoenix-core/src/main/java/org/apache/phoenix/schema/MetaDataClient.java
karanmehta93: Sure. Will add.

phoenix-core/src/main/java/org/apache/phoenix/schema/MetaDataClient.java
karanmehta93: Theoretically yes, This is more of a sanity check. I don't know if you have ever encountered cases where the index table failed to get created. Since Phoenix operations are not atomic, we can end up in a partial state. The HBase Perms API doesn't check if table exists or not, so it will put up an entry even if the table doesn't exist. Let me know your thoughts and we can decide.

phoenix-core/src/main/java/org/apache/phoenix/schema/MetaDataClient.java
karanmehta93: I will look into the alternate. Will refactor.

phoenix-core/src/main/java/org/apache/phoenix/schema/TablesNotInSyncException.java
karanmehta93: Sure. Will refactor.

phoenix-core/src/test/java/org/apache/phoenix/parse/QueryParserTest.java
karanmehta93: Yes. permissions/user/group are treated as literals and table name is treated as identifier.

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/PhoenixAccessController.java
ankitsinghal: @karanmehta93 
Strict mode:- It will check permissions for dependent tables as well. For eg, If a user who has all access on data table is creating an index, then we need to ensure that all others users of data table can also access a new index table.
AutomaticGrant:- It will automatically grant required permissions to dependent table users.

@twdsilva , what about the case when a new index is been created?
Purpose of the automatic grant:- let's say there are three users A and B have READ permission on the data-table and user C has RWC permission on data-table. so if user B creates an index, then we need to ensure that user A and C should also be able to read the index and C should be able to write to this Index and can drop the index also. so we will give only the required permission to the users of data-table on the index table. So, Access should propagate like this. 

user | access data table | access on index table | with Automatic grant(access on index table will change like this) | comments
-- | -- | -- | -- | --
A | RAX | no access | RX | RX will be given on index table
B | RX | RWXC | RWXC | no grant will happen
C | RWXAC | no access | RWCX | read ,write and create will be given so that it can read/write to index table and drop as well.

phoenix-core/src/test/java/org/apache/phoenix/parse/QueryParserTest.java
karanmehta93: HBase API logs a message and ignores it if any other character is found apart from 'RWXCA'. If same character is passed again, it will put up the same string in the hbase:acl table.

 An example of this is
`GRANT 'RXXRRRRCAAAA' to 'user11'`
results in
` hbase:acl   column=l:user11, timestamp=1511376020328, value=RXXRRRRCAAAA`                                                                                                 

 A simple check should be to limit the size of string to 5 characters. Does that seem fine?

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/PhoenixAccessController.java
JamesRTaylor: I don't get the idea of the strict mode option. If this is on, then the grants for secondary indexes would not be done, essentially breaking things. We already have enough config options, so I'd be in favor of removing this one.

ankitsinghal: Oh, I see.. here we are just talking about removing a config, not the automatic grant flow in the coprocessor. I'm fine with either way, having these configs with suitable default or removing the configs completely.

phoenix-core/src/main/java/org/apache/phoenix/schema/MetaDataClient.java
karanmehta93: Theoretically yes, However since the operation is not atomic, it may result into weird case where they are differing.

karanmehta93: This is more like sanity check again.

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/PhoenixAccessController.java
karanmehta93: > Oh, I see.. here we are just talking about removing a config, not the automatic grant flow in the coprocessor. I'm fine with either having these configs with suitable default or removing the configs completely

Yes the code flow will still remain and the scenario that you have explained is a valid one. I will remove both the configs from the code path everywhere.

phoenix-core/src/it/java/org/apache/phoenix/end2end/BasePermissionsIT.java
twdsilva: I meant change the existing tests to use the GRANT/REVOKE statements (the method you added in line 258) and remove the old methods that call AccessControlClient.grant directly.

phoenix-core/src/main/java/org/apache/phoenix/schema/MetaDataClient.java
karanmehta93: @ankitsinghal 
I will file a JIRA for SHOW GRANTS and work on it. As of now, the code will log an error message for all the tables whose permission assignment has failed. Does that seem a good way of exception handling?

phoenix-core/src/it/java/org/apache/phoenix/end2end/BasePermissionsIT.java
karanmehta93: As discussed yesterday, I am still keeping the old tests to use HBase API directly and my tests will use Phoenix GRANT/REVOKE. The old tests verify if a connection can be created and used with minimal permissions and it is good to have them.

phoenix-core/src/main/java/org/apache/phoenix/schema/MetaDataClient.java
karanmehta93: We need HBase TableName object, whereas getPhysicalName api returns PName, so it wont help us here.

JamesRTaylor: This doesn't look correct, @karanmehta93. You need to use indexTableName.getPhysicalName().getBytes() here *and* then call SchemaUtil.getPhysicalTableName like this:

    tableName = SchemaUtil.getPhysicalTableName(indexTableName.getPhysicalName().getBytes(), indexTable.isNamespaceMapped());

Otherwise, for something like an index on a view or a local index, you wouldn't get a physical table name.

karanmehta93: The code that I wrote also works because it takes in the original name and uses its namespace mapped property to determine the Physical table name. 
If you see the code for `getPhysicalName()` in `PTableImpl`, it is the `SchemaUtil.getPhysicalHBaseTableName(schemaName, tableName, isNamespaceMapped)`
The simplest and cleaner option in this case is to use `TableName.valueOf(indexTableName.getPhysicalName().getBytes())`. 
I was trying to avoid using the HBase API's directly and use the SchemaUtil methods. What do you suggest?
@JamesRTaylor

karanmehta93:
@twdsilva @ankitsinghal @JamesRTaylor 

Updates:
Added ChangePermsStatement that combines the Grant and Revoke Statement
Added test for Multitenant tables with view indexes
Updated grammar to reject permission strings with length more than 5
General code refactoring based on comments and added new comments

phoenix-core/src/main/java/org/apache/phoenix/schema/MetaDataClient.java
```diff
@@ -4168,4 +4176,124 @@ public MutationState useSchema(UseSchemaStatement useSchemaStatement) throws SQL
...
+
+        if(schemaInconsistency) {
+            for(PTable table : inconsistentTables) {
+                logger.error("Fail to propagate permissions to Index Table: " + table.getName());
+            }
+            throw new TablesNotInSyncException(inputTable.getTableName().getString(),
+                    inconsistentTables.get(0).getTableName().getString(), "Namespace properties");
+        }
+
+        if(inputTable.isMultiTenant()) {
```
twdsilva: non multi-tenant tables could also have views which have indexes, so you need to check if the view index table exists for them and then keep the perms in sync if it does.

PR #284, opened 2017-12-01
PHOENIX-4424 Allow users to create DEFAULT and HBASE Schema (Uppercas…

karanmehta93:
…e Schema Names)

@twdsilva Please review.

### DIFF
--- a/phoenix-core/src/main/java/org/apache/phoenix/parse/CreateSchemaStatement.java
+++ b/phoenix-core/src/main/java/org/apache/phoenix/parse/CreateSchemaStatement.java
@@ -24,7 +24,7 @@ public class CreateSchemaStatement extends MutableStatement {
 	private final boolean ifNotExists;
 	
 	public CreateSchemaStatement(String schemaName,boolean ifNotExists) {
-		this.schemaName = SchemaUtil.normalizeIdentifier(schemaName);
+		this.schemaName = schemaName;
 		this.ifNotExists = ifNotExists;
 	}
 	
--- a/phoenix-core/src/main/java/org/apache/phoenix/parse/CreateSchemaStatement.java
+++ b/phoenix-core/src/main/java/org/apache/phoenix/parse/CreateSchemaStatement.java
@@ -24,7 +24,7 @@ public class CreateSchemaStatement extends MutableStatement {
 	private final boolean ifNotExists;
 	
 	public CreateSchemaStatement(String schemaName,boolean ifNotExists) {
-		this.schemaName = null == schemaName ? SchemaUtil.SCHEMA_FOR_DEFAULT_NAMESPACE : schemaName;
+		this.schemaName = SchemaUtil.normalizeIdentifier(schemaName);
 		this.ifNotExists = ifNotExists;
 	}
 	
--- a/phoenix-core/src/main/java/org/apache/phoenix/schema/MetaDataClient.java
+++ b/phoenix-core/src/main/java/org/apache/phoenix/schema/MetaDataClient.java
@@ -4089,8 +4089,10 @@ public class MetaDataClient {
                             SQLExceptionCode.CREATE_SCHEMA_NOT_ALLOWED).setSchemaName(create.getSchemaName())
                             .build().buildException(); }
             boolean isIfNotExists = create.isIfNotExists();
-            validateSchema(create.getSchemaName());
             PSchema schema = new PSchema(create.getSchemaName());
+            // Use SchemaName from PSchema object to get the normalized SchemaName
+            // See PHOENIX-4424 for details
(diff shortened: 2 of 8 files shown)


phoenix-core/src/main/java/org/apache/phoenix/parse/CreateSchemaStatement.java
```diff
@@ -24,7 +24,7 @@
 	private final boolean ifNotExists;
 	
 	public CreateSchemaStatement(String schemaName,boolean ifNotExists) {
-		this.schemaName = null == schemaName ? SchemaUtil.SCHEMA_FOR_DEFAULT_NAMESPACE : schemaName;
+		this.schemaName = SchemaUtil.normalizeIdentifier(schemaName);
```
karanmehta93: @JamesRTaylor from JIRA comment
> Identifiers are case sensitive when they are double quoted in Phoenix (and SQL in general). Otherwise they are uppercased. We shouldn’t/can’t change that.

This code change will now make it adhere to SQL standards. The `SchemaUtil.normalizeIdentifier` will help handle the case sensitivity. Otherwise with the current code, if a schema name is passed with double quotes, the full name will be taken as it is for creating the schema, which is incorrect. 

Also look for comments in `MetaDataClient` as well.

phoenix-core/src/main/java/org/apache/phoenix/schema/MetaDataClient.java
```diff
@@ -4125,7 +4125,7 @@ public MutationState createSchema(CreateSchemaStatement create) throws SQLExcept
 
     private void validateSchema(String schemaName) throws SQLException {
         if (SchemaUtil.NOT_ALLOWED_SCHEMA_LIST.contains(
-                schemaName.toUpperCase())) { throw new SQLExceptionInfo.Builder(SQLExceptionCode.SCHEMA_NOT_ALLOWED)
```
karanmehta93: This code would automatically uppercase the schema irrespective of considering if it's enclosed in quotes or not. It will use that value to validate the schema. However, will creating, it will use the regular name. Checkout Line 4093. This seems an inconsistent logic to me. 
For the query
`CREATE SCHEMA "ABC"`, this code will create to create a schema with name `"ABC"` with the double quotes. There was no test that was actually testing this and thus it wasn't caught yet. I wrote a test in `QueryParserTest` and caught it.
@JamesRTaylor Please advice.

JamesRTaylor: Ah, I see. Good catch. Your change is the right one.

karanmehta93: Also checkout the test I updated in `CreateSchemaIT` class.

ankitsinghal: @karanmehta93 , I tried case-sensitive schemas with master code, I don't see any problem. can you check if normalization is happening at some other(or not at the right) place?
```
0: jdbc:phoenix:localhost> create schema "ABC";
No rows affected (0.224 seconds)
```
```
0: jdbc:phoenix:localhost> select TABLE_SCHEM from system.catalog;
+--------------+
| TABLE_SCHEM  |
+--------------+
| ABC          |
```
```
hbase(main):002:0> list_namespace
NAMESPACE
ABC
```
```
0: jdbc:phoenix:localhost> create schema "small_case"
. . . . . . . . . . . . .> ;
No rows affected (0.424 seconds)
0: jdbc:phoenix:localhost> select TABLE_SCHEM from system.catalog;
+--------------+
| TABLE_SCHEM  |
+--------------+
| small_case   |

hbase(main):003:0> list_namespace
NAMESPACE
small_case
```

karanmehta93: @ankitsinghal 
You are correct, the schema gets normalized when we try to create a `PSchema` object and we use `PSchema.getSchemaName()` method which returns the correct schema name. Thanks for pointing it out.
However, the issue is that the normalization doesn't happen before checking the validity of the schema. As of now, the schema is uppercased even before checking if its without quotes or with quotes. This leads to issue when you try to create schema with the names DEFAULT or HBASE, which is the focus for this JIRA.
I will update the patch accordingly.

karanmehta93: @ankitsinghal Thanks for your comments. Please checkout v2 of the patch. I have fixed the bug and I updated the test to verify those scenarios. I also made sure that we don't normalize schema names multiple times, else schema names will always be in capital letters.
FYI. @twdsilva

PR #288, opened 2018-01-11
PHOENIX-4523 phoenix.schema.isNamespaceMappingEnabled problem 

karanmehta93:
(TableExistsException should be ignored when trying to create SYSTEM:MUTEX table)

@JamesRTaylor FYI.

### DIFF
--- a/phoenix-core/src/main/java/org/apache/phoenix/query/ConnectionQueryServicesImpl.java
+++ b/phoenix-core/src/main/java/org/apache/phoenix/query/ConnectionQueryServicesImpl.java
@@ -2543,16 +2543,15 @@ public class ConnectionQueryServicesImpl extends DelegateQueryServices implement
         }
     }
 
-    void createSysMutexTable(HBaseAdmin admin, ReadOnlyProps props) throws IOException, SQLException {
+    void createSysMutexTableIfNotExists(HBaseAdmin admin, ReadOnlyProps props) throws IOException, SQLException {
         try {
-            final TableName mutexTableName = SchemaUtil.getPhysicalTableName(
-                PhoenixDatabaseMetaData.SYSTEM_MUTEX_NAME, props);
-            List<TableName> systemTables = getSystemTableNames(admin);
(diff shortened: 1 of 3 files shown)


phoenix-core/src/main/java/org/apache/phoenix/query/ConnectionQueryServicesImpl.java
```diff
@@ -2543,16 +2543,15 @@ public Void call() throws Exception {
...
 
-    void createSysMutexTable(HBaseAdmin admin, ReadOnlyProps props) throws IOException, SQLException {
+    void createSysMutexTableIfNotExists(HBaseAdmin admin, ReadOnlyProps props) throws IOException, SQLException {
         try {
-            final TableName mutexTableName = SchemaUtil.getPhysicalTableName(
-                PhoenixDatabaseMetaData.SYSTEM_MUTEX_NAME, props);
-            List<TableName> systemTables = getSystemTableNames(admin);
-            if (systemTables.contains(mutexTableName) || admin.tableExists( TableName.valueOf(
-                PhoenixDatabaseMetaData.SYSTEM_SCHEMA_NAME,PhoenixDatabaseMetaData.SYSTEM_MUTEX_TABLE_NAME))) {
+            if(admin.tableExists(PhoenixDatabaseMetaData.SYSTEM_MUTEX_NAME) || admin.tableExists(TableName.valueOf(
```
fpompermaier: why not just leave if (admin.tableExists(mutexTableName)) ?

karanmehta93: The first one checks for SYSTEM.MUTEX and the other one is for SYSTEM:MUTEX, if either of them exists, we can return from the method.

karanmehta93: Thanks for the review @fpompermaier. Closing this PR as this patch has been merged.

PR #289, opened 2018-01-14
PHOENIX-4528 PhoenixAccessController checks permissions only at table…

karanmehta93:
… level when creating views

@ankitsinghal @twdsilva Please review.

@ankitsinghal Please suggest new tests that can be added to verify this patch. The test that I added only verifies that create views would succeed. The change that I have made is generic, however it will be good to add tests that cover scenarios that include creation or dropping of index tables.

### DIFF
--- a/phoenix-core/src/it/java/org/apache/phoenix/end2end/ChangePermissionsIT.java
+++ b/phoenix-core/src/it/java/org/apache/phoenix/end2end/ChangePermissionsIT.java
@@ -145,7 +145,7 @@ public class ChangePermissionsIT extends BasePermissionsIT {
             verifyAllowed(createSchema(SCHEMA_NAME), superUser1);
             verifyAllowed(grantPermissions("C", regularUser1, SCHEMA_NAME, true), superUser1);
         } else {
-            verifyAllowed(grantPermissions("C", regularUser1, "\"" + SchemaUtil.SCHEMA_FOR_DEFAULT_NAMESPACE + "\"", true), superUser1);
+            verifyAllowed(grantPermissions("C", regularUser1, surroundWithDoubleQuotes(SchemaUtil.SCHEMA_FOR_DEFAULT_NAMESPACE), true), superUser1);
         }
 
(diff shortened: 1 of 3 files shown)


twdsilva:
@karanmehta93 

LGTM. If you want to add more test coverage for indexes you can parameterize TableDDLPermissionsIT.testIndexAndView which currently grants permissions directly on the table.

phoenix-core/src/it/java/org/apache/phoenix/end2end/ChangePermissionsIT.java
```diff
@@ -267,4 +267,26 @@ public void testMultiTenantTables() throws Exception {
...
+            verifyAllowed(createSchema(SCHEMA_NAME), superUser1);
+            verifyAllowed(createTable(FULL_TABLE_NAME), superUser1);
+            verifyAllowed(grantPermissions("RX", regularUser1, SCHEMA_NAME, true), superUser1);
+        } else {
+            verifyAllowed(createTable(FULL_TABLE_NAME), superUser1);
+            verifyAllowed(grantPermissions("RX", regularUser1, surroundWithDoubleQuotes(SchemaUtil.SCHEMA_FOR_DEFAULT_NAMESPACE), true), superUser1);
+        }
+
+        verifyAllowed(createView(VIEW1_TABLE_NAME, FULL_TABLE_NAME), regularUser1);
+    }
```
ankitsinghal: regularUser should also have read access on FULL_TABLE_NAME to create view right?

karanmehta93: If the user has access on the SCHEMA of FULL_TABLE_NAME, that should be sufficient, since namespace is bigger in scope that per table scope. Hence I merge all these permissions and then use the `hasAccess()` method to determine the final access.

twdsilva: +1 Please file a separate JIRA to add a test for the index codepath.

PR #292, opened 2018-02-02
PHOENIX-4231: Support restriction of remote UDF load sources

ChinmaySKulkarni:
- Added feature to be able to add jars from an HDFS URI.
- Restrict loading of jars to be only from the hbase.dynamic.jars.dir
directory.

Testing done:
- Tested that the user is able to add jars from an HDFS URI reachable on the
network as well as local filesystem.
- Tested that the user is unable to create a function where the jar is
loaded from any directory apart from hbase.dynamic.jars.dir.
- Tested that HDFS URIs without scheme and authority work.

### DIFF
--- a/phoenix-core/src/main/java/org/apache/phoenix/expression/function/UDFExpression.java
+++ b/phoenix-core/src/main/java/org/apache/phoenix/expression/function/UDFExpression.java
@@ -185,6 +185,7 @@ public class UDFExpression extends ScalarFunction {
                 parent = path.toString();
             }
         }
+        // jarPath is not provided, or it is provided and the jar is inside the configured hbase.dynamic.jars.dir
         if (jarPath == null || jarPath.isEmpty() || config.get(DYNAMIC_JARS_DIR_KEY) != null
                 && (parent != null && parent.equals(config.get(DYNAMIC_JARS_DIR_KEY)))) {
             cl = tenantIdSpecificCls.get(tenantId);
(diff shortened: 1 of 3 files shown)


ChinmaySKulkarni:
@apurtell @twdsilva @jtaylor-sfdc please review. Thanks.

JamesRTaylor:
Thanks for the patch, @ChinmaySKulkarni. Best person to review is @chrajeshbabu who originally added support for UDFs.

phoenix-core/src/main/java/org/apache/phoenix/jdbc/PhoenixStatement.java
```diff
@@ -907,10 +909,15 @@ public MutationState execute() throws SQLException {
...
-                            fs.copyFromLocalFile(new Path(f.getAbsolutePath()), new Path(
-                                    dynamicJarsDir + f.getName()));
+                        for (LiteralParseNode jarPathNode : jarPaths) {
+                          String jarPathName = (String) jarPathNode.getValue();
+                          File f = new File(jarPathName);
+                          Path dynamicJarsDirPathWithJar = new Path(dynamicJarsDir + f.getName());
+                          // Copy the jar (can be local or on HDFS) to the hbase.dynamic.jars.dir directory.
+                          // Note that this does not support HDFS URIs without scheme and authority.
+                          Path jarPath = new Path(jarPathName);
+                          FileUtil.copy(jarPath.getFileSystem(conf), jarPath, fs, dynamicJarsDirPathWithJar,
```
apurtell: If the client does not have perms to write to hbase.dynamic.jars.dir (and I expect normally clients will not have write perms to this directory, only admin clients will have it), the copy will fail and throw an IOException. The result may not be user friendly, though. Did you try this? What happens?

apurtell: Actually, doesn't this imply the client should have write perms to hbase.dynamic.jars.dir? We don't want to allow arbitrary clients to write there.

phoenix-core/src/main/java/org/apache/phoenix/jdbc/PhoenixStatement.java
```diff
@@ -907,10 +909,15 @@ public MutationState execute() throws SQLException {
...
+                            false, true, conf);
```
aertoria: @ChinmaySKulkarni 
How is Hadoop.fileUtil.copy different from filesystem.copy

phoenix-core/src/main/java/org/apache/phoenix/jdbc/PhoenixStatement.java
aertoria: @apurtell 
Since @ChinmaySKulkarni will be absent for a while, I will be wrap up this patch for the team.
do you know how is hbase.dynamic.jars.dir used differently than hbase.local.jars.dir

aertoria: This copy happen at phoenixStatment.java which is at client side. Shouldn't the checking be happening at Server side

phoenix-core/src/main/java/org/apache/phoenix/expression/function/UDFExpression.java
```diff
@@ -198,18 +199,26 @@ public static DynamicClassLoader getClassLoader(final PName tenantId, final Stri
...
-                conf.set(DYNAMIC_JARS_DIR_KEY, parent);
-                cl = new DynamicClassLoader(conf, UDFExpression.class.getClassLoader());
-            }
-            // Cache class loader as a weak value, will be GC'ed when no reference left
-            DynamicClassLoader prev = pathSpecificCls.putIfAbsent(jarPath, cl);
-            if (prev != null) {
-                cl = prev;
+            String rawPath = new Path(config.get(DYNAMIC_JARS_DIR_KEY)).toUri().getRawPath();
+            // jarPath is provided as an HDFS URI without scheme and authority, but the jar is inside the configured hbase.dynamic.jars.dir
+            if(rawPath.equals(parent)) {
```
aertoria: This if block seems will never be reachable. See line 188.  Aren't these two condition equivalent?

phoenix-core/src/main/java/org/apache/phoenix/jdbc/PhoenixStatement.java
aertoria: The comment "Note that this does not support HDFS URIs without scheme and authority." I wonder where did you discover that.

If I'm not mistaken FileUtil.copy internally calls fs.copyFromLocalFile without doing file.islocal check.

aertoria:
Sent out another patch for review. Let us close this p.r.

PR #294, opened 2018-03-06
PHOENIX-4643 Implement ARRAY_REMOVE built in function

xjodoin:


phoenix-core/src/main/java/org/apache/phoenix/expression/function/ArrayRemoveFunction.java
```diff
@@ -0,0 +1,102 @@
...
+		if(baseType.isFixedWidth()) {
+			toCompare = baseType.pad(toCompare, array.getMaxLength());
+		}
+
+		int dimensions = array.getDimensions();
+		List<Object> values = new ArrayList<>();
+		for (int i = 0; i < dimensions; i++) {
+			Object element = array.getElement(i);
+			if (element != null && element.equals(toCompare)
+					|| (element.getClass().isArray() && ArrayUtils.isEquals(element, toCompare))) {
```
JamesRTaylor: We don't support arrays as elements of arrays, so I don't think you need this check.

phoenix-core/src/main/java/org/apache/phoenix/expression/function/ArrayRemoveFunction.java
```diff
@@ -0,0 +1,102 @@
...
+				if (getDataType().isFixedWidth()) {
```
JamesRTaylor: When removing an element, you shouldn't replace it with null. I think you want to just not add the element to values in this case.

JamesRTaylor:
Thanks for the pull request, @xjodoin. Looks very good overall - I made a few specific comments on the implementation of the built-in function.

phoenix-core/src/main/java/org/apache/phoenix/expression/function/ArrayRemoveFunction.java
```diff
@@ -0,0 +1,102 @@
...
+import org.apache.phoenix.schema.types.PArrayDataType;
+import org.apache.phoenix.schema.types.PBinaryArray;
+import org.apache.phoenix.schema.types.PDataType;
+import org.apache.phoenix.schema.types.PVarbinary;
+import org.apache.phoenix.schema.types.PVarbinaryArray;
+import org.apache.phoenix.schema.types.PhoenixArray;
+
+@FunctionParseNode.BuiltInFunction(name = ArrayRemoveFunction.NAME, args = {
+		@FunctionParseNode.Argument(allowedTypes = { PBinaryArray.class, PVarbinaryArray.class }),
+		@FunctionParseNode.Argument(allowedTypes = { PVarbinary.class }, defaultValue = "null") })
```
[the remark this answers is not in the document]
xjodoin: When I remove the default value I have 3 failling tests I started my test from ArrayAppendFunctionIT because I assume the feature set should be similar.
testArrayRemoveFunctionDoublesWithNull
testArrayRemoveFunctionVarcharWithNull
testArrayRemoveFunctionCharsWithNull

phoenix-core/src/main/java/org/apache/phoenix/expression/function/ArrayRemoveFunction.java
```diff
@@ -0,0 +1,102 @@
...
+			PDataType baseType, int arrayLength, Integer maxLength, Expression arrayExp) {
+		SortOrder sortOrder = arrayExp.getSortOrder();
+
+		if (ptr.getLength() == 0 || arrayBytes.length == 0) {
+			ptr.set(arrayBytes, offset, length);
+			return true;
+		}
+
+		PhoenixArray array = (PhoenixArray) getDataType().toObject(arrayBytes, offset, length, getDataType(), sortOrder,
+				maxLength, null);
```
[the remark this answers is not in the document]
xjodoin: Fixed replace with binary compareTo

phoenix-core/src/main/java/org/apache/phoenix/expression/function/ArrayRemoveFunction.java
xjodoin: fix

phoenix-core/src/main/java/org/apache/phoenix/expression/function/ArrayRemoveFunction.java
xjodoin: fix

phoenix-core/src/main/java/org/apache/phoenix/expression/function/ArrayRemoveFunction.java
```diff
@@ -0,0 +1,79 @@
...
+import org.apache.phoenix.schema.types.PArrayDataTypeDecoder;
+import org.apache.phoenix.schema.types.PArrayDataTypeEncoder;
+import org.apache.phoenix.schema.types.PBinaryArray;
+import org.apache.phoenix.schema.types.PDataType;
+import org.apache.phoenix.schema.types.PVarbinary;
+import org.apache.phoenix.schema.types.PVarbinaryArray;
+
+@FunctionParseNode.BuiltInFunction(name = ArrayRemoveFunction.NAME, args = {
+		@FunctionParseNode.Argument(allowedTypes = { PBinaryArray.class, PVarbinaryArray.class }),
+		@FunctionParseNode.Argument(allowedTypes = { PVarbinary.class }, defaultValue = "null") })
```
JamesRTaylor: I filed PHOENIX-4644 with a patch so that you can get rid of this defaultValue. You'll also need to add the nodeClassName attribute here as you see for ARRAY_APPEND.

phoenix-core/src/main/java/org/apache/phoenix/expression/function/ArrayRemoveFunction.java
```diff
@@ -0,0 +1,79 @@
...
+			ptr.set(arrayBytes, offset, length);
+			return true;
+		}
+
+		PArrayDataTypeEncoder arrayDataTypeEncoder = new PArrayDataTypeEncoder(baseType, sortOrder);
+
+		for (int arrayIndex = 0; arrayIndex < arrayLength; arrayIndex++) {
+			ImmutableBytesWritable ptr2 = new ImmutableBytesWritable(arrayBytes, offset, length);
+			PArrayDataTypeDecoder.positionAtArrayElement(ptr2, arrayIndex, baseType, maxLength);
+			if (baseType.compareTo(ptr2, sortOrder, ptr, sortOrder, baseType) != 0) {
```
JamesRTaylor: This looks good. Can you make sure you have tests around removing an element from an array where the element type is slightly different than the array element type? For example, removing an int from a long array or removing a long from a decimal array?

JamesRTaylor:
Changes look good. I've committed PHOENIX-4644, so can you make sure everything works without specify a default value for the second argument for ARRAY_REMOVE?

phoenix-core/src/main/java/org/apache/phoenix/expression/function/ArrayRemoveFunction.java
xjodoin: it's works I added some tests to validate it

phoenix-core/src/it/java/org/apache/phoenix/end2end/ArrayRemoveFunctionIT.java
```diff
@@ -0,0 +1,425 @@
...
+        assertFalse(rs.next());
+    }
+    
+    @Test
+    public void testArrayRemoveFunctionDoubleWithInt() throws Exception {
+        Connection conn = DriverManager.getConnection(getUrl());
+        String tableName = initTables(conn);
+
+        ResultSet rs;
+        rs = conn.createStatement().executeQuery("SELECT ARRAY_REMOVE(doubles,10) FROM " + tableName + " WHERE region_name = 'SF Bay Area'");
```
JamesRTaylor: Make sure to test when 10.0 is in the array and you’re removing with a 10 int argument.

phoenix-core/src/it/java/org/apache/phoenix/end2end/ArrayRemoveFunctionIT.java
```diff
@@ -0,0 +1,425 @@
...
+        assertEquals(array, rs.getArray(1));
+        assertFalse(rs.next());
+    }
+
+    @Test
+    public void testArrayRemoveFunctionBigint() throws Exception {
+        Connection conn = DriverManager.getConnection(getUrl());
+        String tableName = initTables(conn);
+        ResultSet rs;
+        rs = conn.createStatement().executeQuery("SELECT ARRAY_REMOVE(bigints,56) FROM " + tableName + " WHERE region_name = 'SF Bay Area'");
```
JamesRTaylor: Same as above - make sure 56L is in array and is removed.

phoenix-core/src/it/java/org/apache/phoenix/end2end/ArrayRemoveFunctionIT.java
```diff
@@ -0,0 +1,425 @@
...
+                "ARRAY['a','bbbb','c','ddd','e','foo']," +
+                "12,"+
+                "12"+
+                ")";
+        PreparedStatement stmt = conn.prepareStatement(dml);
+        stmt.execute();
+        conn.commit();
+        return tableName;
+    }
+
```
JamesRTaylor: Add tests for:
* removing first array element 
* removing last array element
* removing middle array element 
* removing only element in array 
* removing element which repeats (what should happen?)
* removing CHAR(10) from VARCHAR
* removing VARCHAR from CHAR(10)

xjodoin: I added missing tests, and I think it should remove all matching element from the array

phoenix-core/src/it/java/org/apache/phoenix/end2end/ArrayRemoveFunctionIT.java
```diff
@@ -0,0 +1,383 @@
...
+		assertFalse(rs.next());
+	}
+
+	@Test
+	public void testArrayRemoveFunctionVarcharFromCharArray() throws Exception {
+		ResultSet rs = conn.createStatement().executeQuery(
+				"SELECT ARRAY_REMOVE(chars,varchar1) FROM " + tableName + " WHERE region_name = 'SF Bay Area'");
+		assertTrue(rs.next());
+
+		assertEquals(conn.createArrayOf("CHAR", new String[] { "a", "c", "ddd", "e", "c" }), rs.getArray(1));
```
JamesRTaylor: Instead of removing "bbbb" which is the longest element so won't have trailing spaces, try removing "a" or "c" as it's a better test.

JamesRTaylor:
Looks good - one minor nit and then I think we'll be good to go.

JamesRTaylor:
+1. Thanks for the contribution, @xjodoin. I'll commit this to the various branches.

PR #295, opened 2018-03-22
PHOENIX-4579: Add a config to conditionally create Phoenix meta tables on first client connection

ChinmaySKulkarni:
WIP: Added check for system catalog timestamp while doing client-server compatibility checks.
Added checks for the existence of system catalog to avoid unnecessarily trying to create it.
TODO: Add tests. Test Plan in JIRA comments.


ChinmaySKulkarni:
@twdsilva @JamesRTaylor please take a look and refer to my comments on the JIRA. Thanks!

JamesRTaylor:
Please amend your commit message to be in the form: PHOENIX-4579 <summary>
That way, PR comments will automatically appear as JIRA comments.

ChinmaySKulkarni:
@JamesRTaylor Oops, done. Thanks!

phoenix-core/src/main/java/org/apache/phoenix/query/ConnectionQueryServicesImpl.java
```diff
@@ -2405,16 +2413,26 @@ public Void call() throws Exception {
                             openConnection();
                             hConnectionEstablished = true;
                             boolean isDoNotUpgradePropSet = UpgradeUtil.isNoUpgradeSet(props);
+                            boolean doesSystemCatalogAlreadyExist = false;
```
JamesRTaylor: The flow of this code is very confusing. Let's do the compatibility check only once in ensureTableCreated by making the following changes:
- do not call ensureTableCreated from createTable if the table is a system table
- call ensureTableCreated instead here and remove this entire try block
- do all checks required in ensureTableCreated where we already determine if the table exists or not
- in ensureTableCreated, if SYSTEM.CATALOG doesn't exist and !isAutoUpgradeEnabled or isDoNotUpgradePropSet, then throw our standard UpgradeRequiredException immediately without creating system catalog metadata.
- the only call to checkClientServerCompatibility should be in ensureTableCreated (when isMetaTable is true)
- make sure to be defensive in the creation of the SYSTEM namespace and moving of SYSTEM tables as it's possible that multiple clients may be attempting to do that.

I think this improve the maintainability of this code (and fix this issue too).

JamesRTaylor: I'd also recommend committing PHOENIX-4668 and PHOENIX-4575 first as that'll simplify what you need to do here a bit.

ChinmaySKulkarni: @JamesRTaylor thanks for the review. I've submitted a patch for PHOENIX-4668 and will go ahead with the changes you've mentioned for this patch. I'm not sure about the status of PHOENIX-4575 (@mujtabachohan was working on this before his vacation), so don't want to make any changes to it without understanding its scope.
Thanks.

ChinmaySKulkarni: @JamesRTaylor A few follow-up questions from your comment:

- If we were to call _ensureTableCreated_ at this point inside the init method, don't have the values of other arguments such as the families, splits, etc. which we would need for the admin.createTable call. Instead I propose splitting the _ensureTableCreated_ method and having another private method which just checks for the existence of system catalog and does client-server compatibility checks. We can reuse this stub inside _ensureTableCreated_ and make sure we don't do this again when called via _createTable_.
-  If we throw an UpgradeRequiredException in case an upgrade is required, the user will not get the connection object right? This also prevents the user from being able to run "EXECUTE UPGRADE". Instead I guess we can call _setUpgradeRequired_, log an error and return the connection object. This will disallow the user from running anything except "EXECUTE UPGRADE" since we have this snippet in place:
`                        
if (conn.getQueryServices().isUpgradeRequired() && !conn.isRunningUpgrade()
    && stmt.getOperation() != Operation.UPGRADE) {
        throw new UpgradeRequiredException();
}
`

ChinmaySKulkarni: @JamesRTaylor @twdsilva also, do we really want to avoid calling ensureTableCreated from createTable in case it is a system table? Apart from actually ensuring that the table is created or not and the client-server compatibility checks, this method also modifies the table according to a new descriptor (I don't have enough background as to why we do this here). Any advice guys?

JamesRTaylor: Good points. How about if we leave the ensureTableCreated call where it is, but move the logic in that try block of the init method into the ensureTableCreated method so it’s consolidated?

ChinmaySKulkarni: @JamesRTaylor that could be a possible solution, but then we would always go through executeUpdate-> ... -> _MetaDataClient.createTable_  -> _MetaDataClient.createTableInternal_ -> _CQSI.creataTable_ before finally calling ensureTableCreated and a lot of unnecessary steps would be taken till that point. If we were to disallow the user to proceed in creating/upgrading SYSTEM.CATALOG inside ensureTableCreated, then this would lead to a lot of unnecessary intermediary steps. Instead, can we put all the necessary conditions inside that try block in a separate method which is always only called inside init? This method would also call _checkClientServerCompatibility_ and thus would fail fast in case an upgrade is required, avoiding other unnecessary steps like updating the cache, etc.

JamesRTaylor:
Ok, sounds like a good plan.

JamesRTaylor:
Thinking more, I think it’s fine to go through the CREATE TABLE code path and move the logic in the try block of init down into ensureTableCreated. It’s only called once on the first connection made to a cluster from a client. Keeping the code simple and doing a single RPC to get the metadata and another to get the version will outweigh the minor overhead of compiling CREATE TABLE (plus like you mentioned before, we need to do that to build up the args for ensureTableCreated anyway).

ChinmaySKulkarni:
@JamesRTaylor I'm trying to check if we can move the logic in the try block in init to _ensureTableCreated_ and there seem to be a couple of hurdles while doing so. We don't have access to the DO_NOT_UPGRADE property which is declared to be private inside UpgradeUtil.java and we can't access its getter method 'isNoUpgradeSet' from _ensureTableCreated_ because we have properties of type 'ReadOnlyProps', not 'Properties' (I couldn't find any converter methods between these instances). We need this property to decide whether or not to set _setUpgradeRequired_ to true.

Also, we call _ensureSystemTablesMigratedToSystemNamespace_ inside this try block in init and also have a check which basically throws an InconsistentNamespaceMappingProperties exception if phoenix.schema.mapSystemTablesToNamespace is false and SYSTEM:CATALOG still exists. Now if we move all this logic to _ensureTableCreated_, we would have to add extra checks inside this to make sure whether _ensureTableCreated_ is called for SYSTEM.CATALOG or SYSTEM:CATALOG. 

Because of this, I don't think it's possible to move all this to _ensureTableCreated_.

JamesRTaylor:
I think it's ok to make these changes. Not a problem to make DO_NOT_UPGRADE public if need be. Feel free to add new arguments if necessary to ensureTableCreated. I think any special logic around SYSTEM.CATALOG or SYSTEM:CATALOG is ok as long as it's isolated.

ChinmaySKulkarni:
@JamesRTaylor Ok cool. I've made the changes so checks are inside _ensureTableCreated_. Please take a look now. 
One thing that's confusing me is that inside our _init_ method at the moment, if an upgrade is not required, we create all other SYSTEM tables, however we don't map them to the SYSTEM namespace at that point. 
As of now, when the server gets connected to for the first time i.e. no SYSTEM tables exist, we call
_ensureSystemTablesMigratedToSystemNamespace_, but this just creates the HBase SYSTEM namespace and returns without migrating tables since they haven't been created yet. However, later on when we do create other SYSTEM tables, we aren't migrating them to the SYSTEM namespace right away. 

Now, as per the current design, whenever another client with NS-mapping enabled connects to the server, we will migrate these SYSTEM tables to the SYSTEM namespace when _ensureSystemTablesMigratedToSystemNamespace_ is called within _init_ (with my PR, this will be called inside _ensureTableCreated_ when the table is SYSTEM.CATALOG and NS-mapping is enabled, instead of in _init_). My question is, shouldn't we map SYSTEM tables to the SYSTEM namespace immediately after creating them?

ChinmaySKulkarni:
@aertoria fyi, please take a look since @twdsilva is out. Thanks!

ChinmaySKulkarni:
@aertoria ping. Also, @JamesRTaylor please take a look.
I've added an IT to test various cases encountered for system table creation/upgrade.
Also filed [PHOENIX-4684](https://issues.apache.org/jira/browse/PHOENIX-4684).

ankitsinghal:
bq. As of now, when the server gets connected to for the first time i.e. no SYSTEM tables exist, we call
ensureSystemTablesMigratedToSystemNamespace, but this just creates the HBase SYSTEM namespace and returns without migrating tables since they haven't been created yet. However, later on when we do create other SYSTEM tables, we aren't migrating them to the SYSTEM namespace right away.

As per the current implementation, if there is namespace mapping enabled and no SYSTEM table exists then CREATE TABLE should automatically create SYSTEM tables in SYSTEM namespace. There should not be a need of migrating them. It used to work like this, is it not the case now?

bq. Now, as per the current design, whenever a second client with NS-mapping enabled connects to the server, we will migrate these SYSTEM tables to the SYSTEM namespace when ensureSystemTablesMigratedToSystemNamespace is called within init (with my PR, this will be called inside ensureTableCreated when the table is SYSTEM.CATALOG and NS-mapping is enabled, instead of in init). My question is, shouldn't we map SYSTEM tables to the SYSTEM namespace immediately after creating them?
First connection will be useless, if as per the new design, SYSTEM tables are not getting created in the correct namespace.

phoenix-core/src/main/java/org/apache/phoenix/query/ConnectionQueryServicesImpl.java
```diff
@@ -1039,10 +1065,20 @@ private HTableDescriptor ensureTableCreated(byte[] physicalTableName, PTableType
                 }
             }
 
+            if (isMetaTable && SchemaUtil.isNamespaceMappingEnabled(PTableType.SYSTEM,
```
JamesRTaylor: The ensureSystemTablesMigratedToSystemNamespace has race conditions in the case of multiple clients executing it at the same time. We should treat this as an upgrade step and do it while we have the SYSTEM.MUTEX lock (which ensures only a single client will be executing upgrade code).

phoenix-core/src/main/java/org/apache/phoenix/query/ConnectionQueryServicesImpl.java
```diff
@@ -2643,6 +2661,26 @@ public void upgradeSystemTables(final String url, final Properties props) throws
             metaConnection.setRunningUpgrade(true);
             try {
                 metaConnection.createStatement().executeUpdate(QueryConstants.CREATE_TABLE_METADATA);
+
+                // HBase Namespace SYSTEM is created by {@link ensureSystemTablesMigratedToSystemNamespace(ReadOnlyProps)} method
+                // This statement will create its entry in SYSCAT table, so that GRANT/REVOKE commands can work
+                // with SYSTEM Namespace. (See PHOENIX-4227 https://issues.apache.org/jira/browse/PHOENIX-4227)
+                if (SchemaUtil.isNamespaceMappingEnabled(PTableType.SYSTEM,
+                  ConnectionQueryServicesImpl.this.getProps())) {
```
JamesRTaylor: Inside of this block might be a good place to call ensureSystemTablesMigratedToSystemNamespace. We can treat this as an upgrade step in the 4.14 upgrade path.

phoenix-core/src/it/java/org/apache/phoenix/end2end/SystemCatalogCreationOnConnectionIT.java
```diff
@@ -0,0 +1,533 @@
...
+import org.junit.Before;
+import org.junit.Test;
+
+import java.io.IOException;
+import java.sql.Connection;
+import java.sql.SQLException;
+import java.util.*;
+
+import static org.junit.Assert.*;
+
```
JamesRTaylor: You'll need to add a class-level annotation here so that this test runs when mvn verify is run:

    @Category(NeedsOwnMiniClusterTest.class)

phoenix-core/src/it/java/org/apache/phoenix/end2end/SystemCatalogCreationOnConnectionIT.java
```diff
@@ -0,0 +1,533 @@
...
+            if (testUtil != null) {
+                testUtil.shutdownMiniCluster();
+                testUtil = null;
+            }
+        } catch (Exception e) {
+            // ignore
+        }
+    }
+
+
```
JamesRTaylor: I love all these tests!

phoenix-core/src/main/java/org/apache/phoenix/query/ConnectionQueryServicesImpl.java
ChinmaySKulkarni: @JamesRTaylor We already try to acquire a lock in the SYSMUTEX table inside _ensureSystemTablesMigratedToSystemNamespace_ before we start migrating SYSTEM tables to the SYSTEM namespace, so race conditions shouldn't be a problem here, right? In case multiple clients execute this simultaneously, still only 1 would be able to migrate the tables. 

On a side-note, I still think it logically makes sense to move this to inside _upgradeSystemTables_ for consistency with the method definition.

phoenix-core/src/main/java/org/apache/phoenix/query/ConnectionQueryServicesImpl.java
ChinmaySKulkarni: @JamesRTaylor Makes sense. Note that we don't **already** have the SYSMUTEX lock at this point (referring to your earlier comment). This is in fact necessary to avoid an _UpgradeInProgressException_ from _ensureSystemTablesMigratedToSystemNamespace_ since we try to acquire the SYSMUTEX lock inside here as well. So, moving _ensureSystemTablesMigratedToSystemNamespace_ to this place shouldn't lead to any problems.

phoenix-core/src/it/java/org/apache/phoenix/end2end/SystemCatalogCreationOnConnectionIT.java
ChinmaySKulkarni: Thanks!

phoenix-core/src/main/java/org/apache/phoenix/query/ConnectionQueryServicesImpl.java
JamesRTaylor: If autoUpgrade is enabled, we don't want the user to have to run EXECUTE UPGRADE. Move the creation of the namespace to ensureTableCreated if necessary (but keep the ensureSystemTablesMigratedToSystemNamespace logic in upgradeSystemTables). Let's have a test for this too.

I think it'd be better to have the acquireUpgradeMutex only in upgradeSystemTables so we only do this once. We can acquire it before calling ensureSystemTablesMigratedToSystemNamespace.

ChinmaySKulkarni:
@JamesRTaylor addressed review comments on ensureSystemTablesMigratedToSystemNamespace. Basically, now we return an UpgradeRequiredException even in the case that we need to migrate SYSTEM tables to the SYSTEM namespace. Just doing that was not sufficient since we would have to run 'EXECUTE UPGRADE' once for the migration and once for the upgrade (if required). 

So, to avoid this, I am storing the SYSCAT timestamp as part of the UpgradeRequiredException and I have separated the SYSCAT schema upgrade code to another method. Now we only acquire the SYSMUTEX lock once (1. if we need to just perform an upgrade OR 2. if we need to migrate SYSTEM tables to the SYSTEM namespace and/or also perform an upgrade). 
Added a test for this as well. Please let me know what you think about this approach. Thanks!

phoenix-core/src/it/java/org/apache/phoenix/end2end/SystemCatalogCreationOnConnectionIT.java
```diff
@@ -0,0 +1,577 @@
...
+        hbaseTables = getHBaseTables();
+        assertFalse(hbaseTables.contains(PHOENIX_SYSTEM_CATALOG) || hbaseTables.contains(PHOENIX_NAMESPACE_MAPPED_SYSTEM_CATALOG));
+        assertTrue(hbaseTables.size() == 0);
+        assertEquals(1, countUpgradeAttempts);
+
+        // We use the same ConnectionQueryServices instance to run "EXECUTE UPGRADE"
+        Connection conn = driver.getConnectionQueryServices(getJdbcUrl(), new Properties()).connect(getJdbcUrl(), new Properties());
+        try {
+            conn.createStatement().execute(EXECUTE_UPGRADE_COMMAND);
+        } catch (Exception e) {
```
JamesRTaylor: Remove this catch and let the exception be thrown as that'll make the cause easier to diagnose (and the test will fail too which is what you want).

phoenix-core/src/it/java/org/apache/phoenix/end2end/SystemCatalogCreationOnConnectionIT.java
```diff
@@ -0,0 +1,577 @@
...
+        // Since we set an old timestamp on purpose when creating SYSTEM:CATALOG, the second connection attempts to upgrade it
+        assertEquals(1, countUpgradeAttempts);
+        // This connection is unable to actually upgrade SYSTEM:CATALOG due to isDoNotUpgradePropSet
+        assertEquals(0, actualSysCatUpgrades);
+        Connection conn = driver.getConnectionQueryServices(getJdbcUrl(), new Properties()).connect(getJdbcUrl(), new Properties());
+        try {
+            conn.createStatement().execute(EXECUTE_UPGRADE_COMMAND);
+            // Actually upgraded SYSTEM:CATALOG
+            assertEquals(1, actualSysCatUpgrades);
+        } catch (Exception e) {
```
JamesRTaylor: Same here - remove the catch block.

phoenix-core/src/it/java/org/apache/phoenix/end2end/SystemCatalogCreationOnConnectionIT.java
```diff
@@ -0,0 +1,577 @@
...
+        try {
+            conn.createStatement().execute(EXECUTE_UPGRADE_COMMAND);
+        } catch (Exception e) {
+            fail("EXECUTE UPGRADE should not fail");
+        } finally {
+            conn.close();
+        }
+        hbaseTables = getHBaseTables();
+        assertTrue(PHOENIX_SYSTEM_TABLES.containsAll(hbaseTables));
+        // SYSTEM.MUTEX table is not created
```
JamesRTaylor: The SYSTEM.MUTEX table should be created, no? The SYSTEM.MUTEX table isn't a Phoenix table, though, but it should show up in the hbaseTables list, right? We need that to ensure that only a single client is able to run the  EXECUTE UPGRADE command.

JamesRTaylor:
This looks good. Couple of minor things for the tests, and then rebase it and I think we'll be good to go. Nice work, @ChinmaySKulkarni.

phoenix-core/src/it/java/org/apache/phoenix/end2end/SystemCatalogCreationOnConnectionIT.java
ChinmaySKulkarni: @JamesRTaylor In the current (master) implementation of upgradeSystemTables, we only create SYSMUTEX in case we catch a TableAlreadyExistsException [link](https://github.com/apache/phoenix/blob/master/phoenix-core/src/main/java/org/apache/phoenix/query/ConnectionQueryServicesImpl.java#L2683) when creating SYSCAT. Why don't we create SYSMUTEX (but not acquire a lock) in case of a NewerTableAlreadyExistsException or in case SYSCAT is successfully created?

I guess this is because before this JIRA: Since we only intended to use upgradeSystemTables to actually upgrade SYSTEM tables and not to create them for the first time, SYSMUTEX was already created at this point. Is my assumption correct?

With this JIRA, we now also use upgradeSystemTables to create all SYSTEM tables for the first time,  so we should explicitly always try to create SYSMUTEX too.

How about we call createSysMutexTableIfNotExists before trying to create SYSCAT and then only acquire the upgrade mutex in case of a migration and/or upgrade? We won't need this lock before creation of other SYSTEM tables since competing clients will get TableAlreadyExistsException (consistent with current master implementation too).

ChinmaySKulkarni:
@JamesRTaylor In the current (master) implementation of upgradeSystemTables, we only create SYSMUTEX in case we catch a TableAlreadyExistsException [link](https://github.com/apache/phoenix/blob/master/phoenix-core/src/main/java/org/apache/phoenix/query/ConnectionQueryServicesImpl.java#L2683) when creating SYSCAT. Why don't we create SYSMUTEX (but not acquire a lock) in case of a NewerTableAlreadyExistsException or in case SYSCAT is successfully created?

I guess this is because before this JIRA: Since we only intended to use upgradeSystemTables to actually upgrade SYSTEM tables and not to create them for the first time, SYSMUTEX was already created at this point. Is my assumption correct?

With this JIRA, we now also use upgradeSystemTables to create all SYSTEM tables for the first time, so we should explicitly always try to create SYSMUTEX too.

How about we call createSysMutexTableIfNotExists before trying to create SYSCAT and then only acquire the upgrade mutex in case of a migration and/or upgrade? We won't need this lock before creation of other SYSTEM tables since competing clients will get TableAlreadyExistsException (consistent with current master implementation too).

JamesRTaylor:
Thanks for the updates, @ChinmaySKulkarni. Looks like you still need to rebase it. Also, one test that would be really valuable would be to:
* disable auto upgrade
* run execute upgrade
* make some change to the HBase metadata of SYSTEM.CATALOG through  ALTER TABLE or direct HBase API call (for example, set VERSIONS to some value).
* connect a client that create a new ConnectionQueryServices instance to be created(see UpdateCacheAcrossDifferentClientsIT for how to do that)
* ensure that the metadata change to SYSTEM.CATALOG is not undone.
* do the same with auto upgrade enabled (which might cause the change to be reverted which is ok - this is more for making sure we understand this flow in both cases)

This test will ensure that when this hits production, if we have auto upgrade disabled, we can make manual changes to the SYSTEM.CATALOG.

JamesRTaylor:
If possible, we shouldn't modify SYSTEM table metadata after the HBase table has been created even if auto-upgrade is enabled.

ChinmaySKulkarni:
Squashed all commits and rebased. @JamesRTaylor

JamesRTaylor:
+1. Nice work, @ChinmaySKulkarni! I'll get this committed. Might need some help with a patch that applied to 5.x-HBase-2.0 branch.

ChinmaySKulkarni:
@JamesRTaylor the patch to 5.x-HBase-2.0 branch lgtm! Thanks!

PR #298, opened 2018-04-17
PHOENIX-4666 Persistent subquery cache for hash joins

ortutay:


phoenix-core/src/main/java/org/apache/phoenix/cache/ServerCacheClient.java
```diff
@@ -216,22 +234,146 @@ public void close() throws SQLException {
                 }
             }
         }
-        
+    }
+    
+    public ServerCache checkServerCache(final byte[] cacheId, ScanRanges keyRanges, final TableRef cacheUsingTableRef,
```
maryannxue: I am thinking here, can we do this differently?
Instead of making RPC calls of "checkServerCache" for the first and every subsequent queries, we do NOT make any calls, neither "check" or "add" when the persistent-cache hint is available and catches the {{PersistentCacheNotFoundException}} on the first attempt (or later attempts if somehow the cache has been evicted) and then try adding the cache all over again. I think it will be more efficient in general.

phoenix-core/src/main/java/org/apache/phoenix/cache/TenantCacheImpl.java
```diff
@@ -77,57 +153,164 @@ public MemoryManager getMemoryManager() {
...
+
+    private Cache<ImmutableBytesPtr, CacheEntry> buildCache(final int ttl, final boolean isPersistent) {
+        return CacheBuilder.newBuilder()
+            .expireAfterAccess(ttl, TimeUnit.MILLISECONDS)
+            .ticker(getTicker())
+            .removalListener(new RemovalListener<ImmutableBytesPtr, CacheEntry>(){
+                @Override
+                public void onRemoval(RemovalNotification<ImmutableBytesPtr, CacheEntry> notification) {
+                	if (isPersistent || !notification.getValue().getUsePersistentCache()) {
+                        Closeables.closeAllQuietly(Collections.singletonList(notification.getValue()));
```
maryannxue: Looks like the indentation is not right here. Can you perform a check through the entire PR?

phoenix-core/src/main/java/org/apache/phoenix/cache/TenantCacheImpl.java
```diff
@@ -77,57 +153,164 @@ public MemoryManager getMemoryManager() {
...
+            CacheEntry entry = entries[i];
+            if (!entry.isLive()) {
+            	getServerCaches().invalidate(entry.getCacheId());
+            	getPersistentServerCaches().invalidate(entry.getCacheId());
+                available = this.getMemoryManager().getAvailableMemory();
+            }
+        }
+    }
+
+    private CacheEntry maybeGet(ImmutableBytesPtr cacheId) {
```
maryannxue: Think we can just call it "get" or "getIfPresent" and merge {{maybePromote}} into this call.

phoenix-core/src/main/java/org/apache/phoenix/cache/TenantCacheImpl.java
```diff
@@ -77,57 +153,164 @@ public MemoryManager getMemoryManager() {
...
             }
-        }           
+        }
     }
-    
+
     @Override
-    public void removeServerCache(ImmutableBytesPtr cacheId) {
-        getServerCaches().invalidate(cacheId);
+    synchronized public void removeServerCache(ImmutableBytesPtr cacheId) {
```
maryannxue: Is there any reason why we need to "demote" the cache instead of invalidating it here?

phoenix-core/src/main/java/org/apache/phoenix/compile/QueryCompiler.java
```diff
@@ -314,7 +314,8 @@ protected QueryPlan compileJoinQuery(JoinCompiler.Strategy strategy, StatementCo
                     if (i < count - 1) {
                         fieldPositions[i + 1] = fieldPositions[i] + (tables[i] == null ? 0 : (tables[i].getColumns().size() - tables[i].getPKColumns().size()));
                     }
-                    hashPlans[i] = new HashSubPlan(i, subPlans[i], optimized ? null : hashExpressions, joinSpec.isSingleValueOnly(), keyRangeLhsExpression, keyRangeRhsExpression);
+                    boolean usePersistentCache = joinTable.getStatement().getHint().hasHint(Hint.USE_PERSISTENT_CACHE);
```
maryannxue: We can make "usePersistentCache" a member of QueryCompiler and initialize it in the beginning just like "noChildParentJoinOptimization".

phoenix-core/src/main/java/org/apache/phoenix/compile/QueryCompiler.java
```diff
@@ -383,7 +384,7 @@ protected QueryPlan compileJoinQuery(JoinCompiler.Strategy strategy, StatementCo
                         new PTable[]{lhsTable}, new int[]{fieldPosition}, postJoinFilterExpression, QueryUtil.getOffsetLimit(limit, offset));
                 Pair<Expression, Expression> keyRangeExpressions = new Pair<Expression, Expression>(null, null);
                 getKeyExpressionCombinations(keyRangeExpressions, context, joinTable.getStatement(), rhsTableRef, type, joinExpressions, hashExpressions);
-                return HashJoinPlan.create(joinTable.getStatement(), rhsPlan, joinInfo, new HashSubPlan[]{new HashSubPlan(0, lhsPlan, hashExpressions, false, keyRangeExpressions.getFirst(), keyRangeExpressions.getSecond())});
+                return HashJoinPlan.create(joinTable.getStatement(), rhsPlan, joinInfo, new HashSubPlan[]{new HashSubPlan(0, lhsPlan, hashExpressions, false, false, keyRangeExpressions.getFirst(), keyRangeExpressions.getSecond())});
```
maryannxue: This is also hash-join but with left table as build side. I think we should be able to use persistent cache as well. And could you also add another test to cover this point?

phoenix-core/src/main/java/org/apache/phoenix/execute/HashJoinPlan.java
```diff
@@ -488,11 +502,20 @@ public ServerCache execute(HashJoinPlan parent) throws SQLException {
             if (hashExpressions != null) {
                 ResultIterator iterator = plan.iterator();
                 try {
-                    cache =
-                            parent.hashClient.addHashCache(ranges, iterator,
+                    final byte[] cacheId;
+                    if (usePersistentCache) {
```
maryannxue: Like I said in an earlier comment, if using persistent cache, we don't need to call "check" or "add" cache here, instead we catch a specific Exception and addServerCache at this point. Still, we might wanna return "null" if this is persistent cache so later on the cache "close" method won't be called by the HashJoinPlan clean-up routine.

phoenix-core/src/main/java/org/apache/phoenix/memory/GlobalMemoryManager.java
```diff
@@ -159,6 +161,10 @@ protected void finalize() throws Throwable {
         }
 
         private void freeMemory() {
+//        	System.out.println("Free memory! " + size);
```
maryannxue: I think this a mistake, missed from your code clean-up.

phoenix-core/src/main/java/org/apache/phoenix/cache/ServerCacheClient.java
JamesRTaylor: FYI, we already have a mechanism for this. If the server throws a HashJoinCacheNotFoundException, then the client will react by sending the hash cache to the region servers that don't have it.

ortutay: Yea makes sense. For this, I'm thinking of modifying the retry logic in BaseResultIterators to disable persistent caching on retries. I'm thinking I can add a get/set'er pair to the "context" that can be used for this

phoenix-core/src/main/java/org/apache/phoenix/compile/QueryCompiler.java
ortutay: Done

phoenix-core/src/main/java/org/apache/phoenix/memory/GlobalMemoryManager.java
ortutay: Done

phoenix-core/src/main/java/org/apache/phoenix/compile/QueryCompiler.java
ortutay: @maryannxue I've changed the code here to use the `usePersistentCache` flag, but I'm not sure how to force a `HASH_BUILD_LEFT` strategy. The test I've put in the integration test file does not actually use that strategy. Do you know of an easy way to force a left table hash join?

phoenix-core/src/main/java/org/apache/phoenix/cache/TenantCacheImpl.java
ortutay: Fixed it here. FWIW some files mix tabs / spaces

ortutay:
> Another question to consider besides the suggested changes: do we need a SQL command to remove a specific cache entry and/or purge the entire persistent cache?

I think that's a good idea but I would prefer to leave it for a future enhancement PR. There's a lot of improvements that can be made once people start using this and make feature requests.

JamesRTaylor:
Is this ready to go, @ortutay? It looks pretty clean now.

ortutay:
@JamesRTaylor almost ready, we've been testing/deploying this internally and have a few small changes that I'll push out before it's ready to merge

ortutay:
@JamesRTaylor @maryannxue updated for all comments in the PR ; this now uses an exception flow for checking for presence of SQ cache, instead of the previous RPC approach. I also a dded to the integration test to verify that the SQ cache is properly used across queries.

Also, regarding expiring the cache--we have been using a technique internally for this based on nonces. You can add a no-op condition to the query such as "1=1" or "2=2" which changes the persistent cache key, and thus is a way to force cache invalidation. You can do it on a daily basis, eg. 20180731=20180731, and this can also be used to pre-warm caches for upcoming days as well. I'm planning to make a blog post explaining SQ cache + how to rotate nonces once this change has been merged.

ortutay:
@JamesRTaylor @maryannxue is this OK to merge in? I believe all review comments have been addressed, and we've been using this at 23andMe already on custom branch

ortutay:
Updated this to target `master`

phoenix-core/src/main/java/org/apache/phoenix/cache/TenantCacheImpl.java
```diff
@@ -77,57 +152,132 @@ public MemoryManager getMemoryManager() {
...
+                        Closeables.closeAllQuietly(Collections.singletonList(notification.getValue()));
+                	}
+                }
+            })
+            .build();
+    }
     
-    @Override
+    private void evictInactiveEntries(long bytesNeeded) {
+        CacheEntry[] entries = getPersistentServerCaches().asMap().values().toArray(new CacheEntry[]{});
```
joshelser: Is the iteration of `values()` threadsafe in the face of another thread updating the persistentServerCaches?

phoenix-core/src/main/java/org/apache/phoenix/cache/TenantCacheImpl.java
```diff
@@ -77,57 +152,132 @@ public MemoryManager getMemoryManager() {
...
+        Arrays.sort(entries);
```
joshelser: I'm not getting how these `CacheEntries` are being sorted. It's based off of the `hits` member variable, right?

So, the persistentCaches that have the least number of queries run against them we would remove first?

Do we want to/are we able to include the size of the persistentCache in this decision?

phoenix-core/src/main/java/org/apache/phoenix/cache/TenantCacheImpl.java
joshelser: Also, how about filtering out `CacheEntry` which are `!isAlive()`?

phoenix-core/src/main/java/org/apache/phoenix/cache/TenantCacheImpl.java
```diff
@@ -77,57 +152,132 @@ public MemoryManager getMemoryManager() {
...
+                available = this.getMemoryManager().getAvailableMemory();
+            }
+        }
+    }
+
+    private CacheEntry getIfPresent(ImmutableBytesPtr cacheId) {
+        CacheEntry entry = getPersistentServerCaches().getIfPresent(cacheId);
+        if (entry != null) {
+            getServerCaches().put(cacheId, entry);
+        }return getServerCaches().getIfPresent(cacheId);
```
joshelser: nit: formatting

phoenix-core/src/main/java/org/apache/phoenix/cache/TenantCacheImpl.java
```diff
@@ -77,57 +152,132 @@ public MemoryManager getMemoryManager() {
...
+            builder.expireAfterWrite(ttl, TimeUnit.MILLISECONDS);
+        } else {
+            builder.expireAfterAccess(ttl, TimeUnit.MILLISECONDS);
+        }
+        return builder
+            .ticker(getTicker())
+            .removalListener(new RemovalListener<ImmutableBytesPtr, CacheEntry>(){
+                @Override
+                public void onRemoval(RemovalNotification<ImmutableBytesPtr, CacheEntry> notification) {
+                	if (isPersistent || !notification.getValue().getUsePersistentCache()) {
```
joshelser: Can we get some DEBUG/TRACE logging here? History has told me that being able to understand evictions about server-side caches in Phoenix is very useful ;)

phoenix-core/src/main/java/org/apache/phoenix/compile/QueryCompiler.java
```diff
@@ -314,7 +316,8 @@ protected QueryPlan compileJoinQuery(JoinCompiler.Strategy strategy, StatementCo
                     if (i < count - 1) {
                         fieldPositions[i + 1] = fieldPositions[i] + (tables[i] == null ? 0 : (tables[i].getColumns().size() - tables[i].getPKColumns().size()));
                     }
-                    hashPlans[i] = new HashSubPlan(i, subPlans[i], optimized ? null : hashExpressions, joinSpec.isSingleValueOnly(), keyRangeLhsExpression, keyRangeRhsExpression);
+//                    boolean usePersistentCache = joinTable.getStatement().getHint().hasHint(Hint.USE_PERSISTENT_CACHE);
```
joshelser: Remove it

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/HashJoinRegionScanner.java
```diff
@@ -23,6 +23,7 @@
 import java.util.List;
 import java.util.Queue;
 
+import org.apache.commons.codec.binary.Hex;
```
joshelser: Unused, remove it.

phoenix-core/src/main/java/org/apache/phoenix/execute/HashJoinPlan.java
```diff
@@ -494,19 +540,37 @@ public ServerCache execute(HashJoinPlan parent) throws SQLException {
...
+                            try {
+                                cache = parent.hashClient.createServerCache(cacheId, parent.delegate);
+                            } catch (IOException e) {
+                                throw new RuntimeException(e);
+                            }
+                        }
+                    } else {
+                        cacheId = Bytes.toBytes(RANDOM.nextLong());
+                    }
+                    System.out.println("Using cache ID " + Hex.encodeHexString(cacheId) + " for " + queryString);
```
joshelser: This, and others, need to use a logger.

phoenix-core/src/main/java/org/apache/phoenix/join/HashCacheFactory.java
```diff
@@ -25,18 +25,22 @@
...
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
+import java.util.Set;
 
 import net.jcip.annotations.Immutable;
 
+import org.apache.commons.codec.binary.Hex;
```
joshelser: Remove

phoenix-core/src/main/java/org/apache/phoenix/memory/GlobalMemoryManager.java
```diff
@@ -17,6 +17,8 @@
  */
 package org.apache.phoenix.memory;
 
+import java.util.Arrays;
```
joshelser: Remove

phoenix-core/src/it/java/org/apache/phoenix/end2end/join/HashJoinPersistentCacheIT.java
```diff
@@ -0,0 +1,137 @@
...
+
+    @Override
+    protected String getTableName(Connection conn, String virtualName) throws Exception {
+        String realName = super.getTableName(conn, virtualName);
+        TestUtil.addCoprocessor(conn, SchemaUtil.normalizeFullTableName(realName), InvalidateHashCache.class);
+        return realName;
+    }
+                
+    @Test
+    public void testPersistentCache() throws Exception {
```
joshelser: This is a very nice test!

phoenix-core/src/main/java/org/apache/phoenix/cache/TenantCacheImpl.java
```diff
@@ -77,57 +152,132 @@ public MemoryManager getMemoryManager() {
...
+        return entry.closeable;
     }
-    
+
     @Override
-    public Closeable addServerCache(ImmutableBytesPtr cacheId, ImmutableBytesWritable cachePtr, byte[] txState, ServerCacheFactory cacheFactory, boolean useProtoForIndexMaintainer, int clientVersion) throws SQLException {
+    public Closeable addServerCache(ImmutableBytesPtr cacheId, ImmutableBytesWritable cachePtr, byte[] txState, ServerCacheFactory cacheFactory, boolean useProtoForIndexMaintainer, boolean usePersistentCache, int clientVersion) throws SQLException {
         getServerCaches().cleanUp();
-        MemoryChunk chunk = this.getMemoryManager().allocate(cachePtr.getLength() + txState.length);
+        long available = this.getMemoryManager().getAvailableMemory();
```
joshelser: Two scary points (that I hope I just missed via my scanning):

* I don't see a corresponding update to the GlobalMemoryManager. Are we accounting for the size of `persistentServerCaches` somewhere else?
* Do users who execute queries with persistentCaches have the ability to deny the queries of users not running persistentCaches (by saturating the total RS's serverCache). The 30min default timeout to start evicting persistentCaches makes me worry as that's a long time.

joshelser: I am fairly certain that we export serverCache metrics to the HBase RS metrics subsystem. Assuming that we do, can we break out the persistentServerCache usage as well? That would be nice to monitor.

phoenix-core/src/main/java/org/apache/phoenix/cache/TenantCacheImpl.java
```diff
@@ -77,57 +152,132 @@ public MemoryManager getMemoryManager() {
...
+        int size = cachePtr.getLength() + txState.length;
+        if (size > available) {
+            evictInactiveEntries(Math.max(size - available + EVICTION_MARGIN_BYTES, EVICTION_MARGIN_BYTES));
```
twdsilva: Seems like this will always be size - available + EVICTION_MARGIN_BYTES since size>available. So if we need to evict we will evict a minimum of EVICTION_MARGIN_BYTES ?

phoenix-core/src/main/java/org/apache/phoenix/cache/TenantCacheImpl.java
```diff
@@ -77,57 +152,132 @@ public MemoryManager getMemoryManager() {
...
     
-    @Override
+    private void evictInactiveEntries(long bytesNeeded) {
+        CacheEntry[] entries = getPersistentServerCaches().asMap().values().toArray(new CacheEntry[]{});
+        Arrays.sort(entries);
+        long available = this.getMemoryManager().getAvailableMemory();
+        for (int i = 0; i < entries.length && available < bytesNeeded; i++) {
+            CacheEntry entry = entries[i];
+            if (!entry.isLive()) {
+            	getServerCaches().invalidate(entry.getCacheId());
```
twdsilva: Does this try to evict from both persistent cache and the regular cache?

phoenix-core/src/main/java/org/apache/phoenix/cache/TenantCacheImpl.java
```diff
@@ -77,57 +152,132 @@ public MemoryManager getMemoryManager() {
...
-            getServerCaches().put(cacheId, element);
+            CacheEntry entry;
+            synchronized(this) {
+                entry = getIfPresent(cacheId);
+                if (entry == null) {
+                    entry = new CacheEntry(
+                            cacheId, cachePtr, cacheFactory, txState, chunk,
+                            usePersistentCache, useProtoForIndexMaintainer,
+                            clientVersion);
+                    getServerCaches().put(cacheId, entry);
```
twdsilva: Do we need to put the entry in both the persistent cache and the regular cache?

phoenix-core/src/main/java/org/apache/phoenix/iterate/TableResultIterator.java
```diff
@@ -194,13 +194,10 @@ public Tuple next() throws SQLException {
                             if (retry <= 0) {
                                 throw e1;
                             }
+                            Long cacheId = ((HashJoinCacheNotFoundException) e1).getCacheId();
                             retry--;
                             try {
-                                Long cacheId = ((HashJoinCacheNotFoundException) e1).getCacheId();
-
-                                ServerCache cache = caches == null ? null :
```
twdsilva: Do we not need this null check any more?

phoenix-core/src/main/java/org/apache/phoenix/cache/ServerCacheClient.java
```diff
@@ -336,6 +356,7 @@ private void removeServerCache(final ServerCache cache, Set<HRegionLocation> rem
             ConnectionQueryServices services = connection.getQueryServices();
             Throwable lastThrowable = null;
             final PTable cacheUsingTable = cacheUsingTableMap.get(Bytes.mapKey(cacheId));
+            cacheUsingTableMap.keySet();
```
dbwong: Does this do anything?

phoenix-core/src/main/java/org/apache/phoenix/execute/HashJoinPlan.java
```diff
@@ -241,8 +251,35 @@ public TaskExecutionMetricsHolder getTaskExecutionMetric() {
...
-        return iterator;
+
+        if (hasSubPlansWithPersistentCache) {
+            return peekForPersistentCache(iterator, scanGrouper, scan);
+        } else {
+            return iterator;
+        }
+    }
+
+    private ResultIterator peekForPersistentCache(ResultIterator iterator, ParallelScanGrouper scanGrouper, Scan scan) throws SQLException {
```
dbwong: Is there a test for this retry logic with the setDisablePersistenCache that I'm missing?  I don't see error path testing.

phoenix-core/src/main/java/org/apache/phoenix/execute/HashJoinPlan.java
```diff
@@ -494,19 +540,37 @@ public ServerCache execute(HashJoinPlan parent) throws SQLException {
...
+                                cache = parent.hashClient.createServerCache(cacheId, parent.delegate);
+                            } catch (IOException e) {
+                                throw new RuntimeException(e);
+                            }
+                        }
+                    } else {
+                        cacheId = Bytes.toBytes(RANDOM.nextLong());
+                    }
+                    System.out.println("Using cache ID " + Hex.encodeHexString(cacheId) + " for " + queryString);
+                    if (cache == null) {
```
dbwong: If disablePersistentCache == true this addHashCace is still using the cacheId from the query string as opposed to a randomID.  Is this intended?  Will this affect the future interactions with that cache?

phoenix-core/src/main/java/org/apache/phoenix/cache/ServerCacheClient.java
ortutay: No, removed it

phoenix-core/src/main/java/org/apache/phoenix/cache/TenantCacheImpl.java
ortutay: Yes, added some DEBUG output to this file

phoenix-core/src/main/java/org/apache/phoenix/cache/TenantCacheImpl.java
ortutay: Yes, it's currently just based on `hits`, which is not very smart, but very simple to code.

I think it should, at least, account for how long the entry has been in the cache, and perhaps take other things into account, like size of the cache entry. I was going to submit a follow-up PR to improve the ranking.

phoenix-core/src/main/java/org/apache/phoenix/cache/TenantCacheImpl.java
ortutay: Good point, added `synchronized` to this method.

The filtering for `!isAlive()` was below, but I've changed to behavior as a bit. Now, nothing that "isAlive" exists in `getPersistentServerCaches()`, so the filter is not necessary.

phoenix-core/src/main/java/org/apache/phoenix/cache/TenantCacheImpl.java
ortutay: fixed

phoenix-core/src/main/java/org/apache/phoenix/cache/TenantCacheImpl.java
ortutay: > I don't see a corresponding update to the GlobalMemoryManager. Are we accounting for the size of persistentServerCaches somewhere else?

I think this is handled, but I could be wrong. There is a call like this: `MemoryChunk chunk = this.getMemoryManager().allocate(size);`. This allocates the server cache. The server cache can then live in one or two hash maps, either only `getServerCaches()`, or both `getServerCaches()` and `getPersistentServerCaches()`. However, even if it lives in both, it only uses `size` bytes, since the two maps have a pointer to the same memory chunk.

phoenix-core/src/main/java/org/apache/phoenix/cache/TenantCacheImpl.java
ortutay: Good point, in fact the check is right above :). Removed the redundant `max`

phoenix-core/src/main/java/org/apache/phoenix/cache/TenantCacheImpl.java
ortutay: Yes, if `usePersistentCache` is true, it will be added on the following line

phoenix-core/src/main/java/org/apache/phoenix/compile/QueryCompiler.java
ortutay: done

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/HashJoinRegionScanner.java
ortutay: done

phoenix-core/src/main/java/org/apache/phoenix/execute/HashJoinPlan.java
ortutay: yes. the naming for `disablePersistentCache` not great, maybe I should change it. That variable is true on a retry. It retries when the persistent cache was missing on the first pass, so now it is effectively warming the cache that will be used in subsequent queries.

Maybe `isPersistentCacheRetry` would be a better variable name here.

phoenix-core/src/main/java/org/apache/phoenix/iterate/TableResultIterator.java
ortutay: reverted, I think this was accidentally changed during git rebase.

phoenix-core/src/main/java/org/apache/phoenix/join/HashCacheFactory.java
ortutay: done

phoenix-core/src/main/java/org/apache/phoenix/memory/GlobalMemoryManager.java
ortutay: done

phoenix-core/src/main/java/org/apache/phoenix/execute/HashJoinPlan.java
ortutay: It is not explicitly tested, however it is tested from `HashJoinPersistentCacheIT`. Basically, this code path handles the first, optimistic query which assumes the cache already exists, and then does the second query which generates the cache. It only does the second one when needed.

The rethrow for this part:

```                if (delegate.getContext().getDisablePersistentCache(cacheId)) {
                    throw e2;
                }
``` is not tested, I think

phoenix-core/src/main/java/org/apache/phoenix/cache/TenantCacheImpl.java
joshelser: > since the two maps have a pointer to the same memory chunk.

Ah, that's the rub. I assumed that when the persistentCaches was used, there would increased memory consumption over when they were not used. If memory use is constant (or at least, negligible), then I agree with you.

phoenix-core/src/main/java/org/apache/phoenix/cache/TenantCacheImpl.java
joshelser: thanks!

joshelser:
> should I post a 4.x branch patch on the Jira ticket now?

I don't have any more feedback for ya, so, as far as I'm concerned, you can work on a backport patch :)

ortutay:
Sounds good, also squashed the commits just now

ortutay:
One last git ammend to clean up whitespace. The backport patch has cleaned up whitespace as well.

ortutay:
@dbwong @twdsilva  any comments? is this ready to merge?

dbwong:
Looks okay to me.  I spent a bit of time looking if there would be any issues for concurrent attempts to create the same sever side cache but I think everything should be fine along that path.

twdsilva:
LGTM can you click submit patch so we can get a test run?

twdsilva:
If you upload a patch file without any branch name the precommit build uses master by default.

ortutay:
Ok, submitted the master (5.x) patch just now, can folllow-up with the 4.x patch after 5.x looks good / is merged

stoty:
Already merged.

PR #301, opened 2018-05-23
PHOENIX-4728 The upsert select must project tuples

xjodoin:


phoenix-core/src/main/java/org/apache/phoenix/compile/UpsertCompiler.java
```diff
@@ -549,7 +549,7 @@ public MutationPlan compile(UpsertStatement upsert) throws SQLException {
             select = SelectStatement.create(select, hint);
             // Pass scan through if same table in upsert and select so that projection is computed correctly
             // Use optimizer to choose the best plan
-            QueryCompiler compiler = new QueryCompiler(statement, select, selectResolver, targetColumns, parallelIteratorFactoryToBe, new SequenceManager(statement), false, false, null);
+            QueryCompiler compiler = new QueryCompiler(statement, select, selectResolver, targetColumns, parallelIteratorFactoryToBe, new SequenceManager(statement), true, false, null);
```
JamesRTaylor: This seems like too general of a change for the specific issue you're trying to fix for ARRAY_APPEND. I'm also not sure *why* it would impact it. Can't you make changes to ArrayAppendFunction or it's base class to get the desired affect?

Any opinions, @maryannxue. Do you remember when/why we need this projectTuples boolean for QueryCompiler?

xjodoin:
No the problem come from the querycompiler for the upsertselect the behavior is different than a simple select query 

Le 23 mai 2018 16:04:46 HAE, James Taylor <notifications@github.com> a écrit :
>JamesRTaylor commented on this pull request.
>
>
>
>> @@ -549,7 +549,7 @@ public MutationPlan compile(UpsertStatement
>upsert) throws SQLException {
>             select = SelectStatement.create(select, hint);
>// Pass scan through if same table in upsert and select so that
>projection is computed correctly
>             // Use optimizer to choose the best plan
>-            QueryCompiler compiler = new QueryCompiler(statement,
>select, selectResolver, targetColumns, parallelIteratorFactoryToBe, new
>SequenceManager(statement), false, false, null);
>+            QueryCompiler compiler = new QueryCompiler(statement,
>select, selectResolver, targetColumns, parallelIteratorFactoryToBe, new
>SequenceManager(statement), true, false, null);
>
>This seems like too general of a change for the specific issue you're
>trying to fix for ARRAY_APPEND. I'm also not sure *why* it would impact
>it. Can't you make changes to ArrayAppendFunction or it's base class to
>get the desired affect?
>
>Any opinions, @maryannxue. Do you remember when/why we need this
>projectTuples boolean for QueryCompiler?
>
>-- 
>You are receiving this because you authored the thread.
>Reply to this email directly or view it on GitHub:
>https://github.com/apache/phoenix/pull/301#pullrequestreview-122745792

-- 
Envoyé de mon appareil Android avec K-9 Mail. Veuillez excuser ma brièveté.

phoenix-core/src/main/java/org/apache/phoenix/compile/UpsertCompiler.java
maryannxue: "Tuple projection" was first used for join queries so that columns are accessed based on positions instead of names. We later applied this to single-table queries, but for some reason (reason that I can't recall right now), we wanted to avoid tuple projection in UPSERT. If this change won't cause any existing test failure, I think it's just fine.

xjodoin: All the tests passed

PR #304, opened 2018-06-01
PHOENIX-3955: Ensure KEEP_DELETED_CELLS, REPLICATION_SCOPE, and TTL properties stay in sync between the physical data table and index tables

ChinmaySKulkarni:
**[DO NOT MERGE - WIP patch]**

- Disallow specifying TTL, KEEP_DELETED_CELLS and REPLICATION_SCOPE at
column family level when creating table and also while creating indexes
- Sync these properties in create index code path after checking for
sync of existing column families in data table
- Modified alter table and alter global index code path to ensure
syncing of properties
- Added syncing properties of column families of tables and their
indexes as a step in execute upgrade code path

TODO:
- Add local testing that has been done
- Test old client new server and execute upgrade
- Handle indexes on views


ChinmaySKulkarni:
@twdsilva @vincentpoon please review. I will be adding stuff mentioned in "TODO", but just wanted your feedback on this initial patch.

ChinmaySKulkarni:
Ran into [PHOENIX-3167](https://issues.apache.org/jira/browse/PHOENIX-3167) when doing some testing for this JIRA. 

This can be a bigger problem now that we force an upgrade in case column families are out of sync for a table.

Steps to repro:
create table test (id INTEGER not null primary key, name varchar(10)) TTL=1200,KEEP_DELETED_CELLS='false',REPLICATION_SCOPE='1';
alter table test add cf1.random varchar(10);
-> New column family cf1 inherits TTL, KEEP_DELETED_CELLS and REPLICATION_SCOPE properties
alter table test set TTL=100000;
create table if not exists test (id INTEGER not null primary key, name varchar(10)) TTL=1200,KEEP_DELETED_CELLS='false',REPLICATION_SCOPE='1';
-> Modifies HBase metadata for the table so TTL for default cf is changed to 1200, whereas newly added cf1's TTL is still 100000.

With this patch, at this point, any future alter table command or create index will fail with UpgradeRequiredException since column family properties are out of sync.

Even if we run the second create table with 'if not exists', same behavior.
Similar problem is seen with alter table for global index. See [PHOENIX-4743](https://issues.apache.org/jira/browse/PHOENIX-4743)
@twdsilva @vincentpoon

twdsilva:
I think you is better If you add code to UpgradeUtil that runs at the next major upgrade that checks all the tables and ensures these three property values are in sync (see addParentToChildLinks for an example). Then you wouldn't need to validate that the properties are in sync in multiple places.

ChinmaySKulkarni:
Closing this PR and will send out a new one

ChinmaySKulkarni:
Please refer to https://github.com/apache/phoenix/pull/346

PR #306, opened 2018-06-27
PHOENIX-4797  file not found or file exist exception when create glob…

492066199:
when create global index use -snaopshot option in mapreduce task
multi mapper use the same restore dir to work with snapshot.
1,mapper will remove the file  after mapper task as the other mapper will read the file
2,mapper will create the file when begin mapper task as the other mapper will create the file too

the 1 will cause the file not found and the 2 will case the file is exists
I change this to let mappers use different dirs to work with snapshot,after this patch,the  -snapshot option can works well .

### DIFF
--- a/phoenix-core/src/main/java/org/apache/phoenix/iterate/TableSnapshotResultIterator.java
+++ b/phoenix-core/src/main/java/org/apache/phoenix/iterate/TableSnapshotResultIterator.java
@@ -39,6 +39,7 @@ import java.util.ArrayList;
 import java.util.Collections;
 import java.util.Iterator;
 import java.util.List;
+import java.util.UUID;
 
 public class TableSnapshotResultIterator implements ResultIterator {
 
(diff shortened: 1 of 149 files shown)


phoenix-core/src/main/java/org/apache/phoenix/iterate/TableSnapshotResultIterator.java
```diff
@@ -35,10 +35,7 @@
 
 import java.io.IOException;
 import java.sql.SQLException;
-import java.util.ArrayList;
-import java.util.Collections;
-import java.util.Iterator;
-import java.util.List;
+import java.util.*;
```
karanmehta93: Nit: you should not have wildcard imports as this is probably unnecessary diff

492066199:
hi, Thanks for reminding me, I remove the wildcard imports.

phoenix-core/src/main/java/org/apache/phoenix/iterate/TableSnapshotResultIterator.java
```diff
@@ -65,7 +66,8 @@ public TableSnapshotResultIterator(Configuration configuration, Scan scan, ScanM
     this.scan = scan;
     this.scanMetricsHolder = scanMetricsHolder;
     this.scanIterator = UNINITIALIZED_SCANNER;
-    this.restoreDir = new Path(configuration.get(PhoenixConfigurationUtil.RESTORE_DIR_KEY));
+    this.restoreDir = new Path(new Path(configuration.get(PhoenixConfigurationUtil.RESTORE_DIR_KEY)),
```
karanmehta93: `Path` also offers other API at https://hadoop.apache.org/docs/r2.8.2/api/org/apache/hadoop/fs/Path.html
You can specify both of the params as `String`

492066199:
Thank you for your advice，I've changed it to use both ”String“ Path api;

PR #308, opened 2018-07-02
Client-side hash aggregation

geraldss:
Client-side hash aggregation for use with sort-merge join.

Implements https://issues.apache.org/jira/browse/PHOENIX-4751


phoenix-core/src/main/java/org/apache/phoenix/iterate/ClientHashAggregatingResultIterator.java
```diff
@@ -0,0 +1,155 @@
...
+    private static final byte[] UNITIALIZED_KEY_BUFFER = new byte[0];
+    private final ResultIterator resultIterator;
+    private final Aggregators aggregators;
+    private final List<Expression> groupByExpressions;
+    private HashMap<ImmutableBytesWritable, Aggregator[]> hash;
+    private List<ImmutableBytesWritable> keyList;
+    private Iterator<ImmutableBytesWritable> keyIterator;
+
+    public ClientHashAggregatingResultIterator(ResultIterator resultIterator, Aggregators aggregators, List<Expression> groupByExpressions) {
+        if (resultIterator == null) throw new NullPointerException();
```
joshelser: Could consolidate this down into `Objects.requireNonNull(resultIterator)` from java.util.Objects.

phoenix-core/src/main/java/org/apache/phoenix/iterate/ClientHashAggregatingResultIterator.java
```diff
@@ -0,0 +1,155 @@
...
+        if (groupByExpressions == null) throw new NullPointerException();
+        this.resultIterator = resultIterator;
+        this.aggregators = aggregators;
+        this.groupByExpressions = groupByExpressions;
+    }
+
+    @Override
+    public Tuple next() throws SQLException {
+        if (keyIterator == null) {
+            populateHash();
```
joshelser: A comment here would be nice to note the side-effect that `hash` and `keyList` are guaranteed to be non-null (and thus, the lack of the defensive null-checks below is OK).

geraldss:
@joshelser Thanks for the review. I can make these changes now and push them. Or should I wait for any other reviewer?

phoenix-core/src/main/java/org/apache/phoenix/iterate/ClientHashAggregatingResultIterator.java
geraldss: Done.

phoenix-core/src/main/java/org/apache/phoenix/iterate/ClientHashAggregatingResultIterator.java
geraldss: Made the side effects more explicit.

phoenix-core/src/main/java/org/apache/phoenix/iterate/ClientHashAggregatingResultIterator.java
```diff
@@ -0,0 +1,155 @@
...
+            throw new SQLException(e);
+        }
+    }
+
+    // Copied from ClientGroupedAggregatingResultIterator
+    protected Tuple wrapKeyValueAsResult(KeyValue keyValue) {
+        return new MultiKeyValueTuple(Collections.<Cell> singletonList(keyValue));
+    }
+
+    private void populateHash() throws SQLException {
```
solzy: For a large result set of join, this code snippet has a risk lead to OOM on client. Maybe add limitation is  more graceful

geraldss: Hi @solzy , thanks for the review. My understanding of the pre-existing code is that classes like OrderedResultIterator do not handle OOM, and they do not provide a limitation. I don't want to introduce a mechanism that is specific to this feature. Is there any place where the client handles OOM, which can be used as a reference?

solzy: @geraldss Sorry for late reply.  i did not find some handle for OOM on the client, which will be affects stability  when uses as QueryServer. anyway, we can improve these features in the future. Thanks for your response , good job.

JamesRTaylor: @geraldss - memory management is tracked by the GlobalMemoryManager. Operations that potentially use memory allocate (and eventually free) a set of MemoryChunk instances. You can see an example of this in GroupedAggregateRegionObserver (the runtime code for aggregation). If the memory used goes over a threshold (phoenix.query.maxGlobalMemoryPercentage and phoenix.query.maxTenantMemoryPercentage as the allowed percentage of Java heap across all queries that is allowed to be used), then the query will fail. Most typically, this mechanism is used on the server side as we don't typically use a lot of memory on the client-side (as we're mostly doing merge joins). One example where we use this on the client side is for our broadcast join implementation (see HashCacheClient) to track memory held onto for Hash Join caches.

Classes you may want to look at (or perhaps you already have?): OrderedResultIterator and MappedByteBufferSortedQueue. Above a certain configurable threshold (phoenix.query.spoolThresholdBytes defaults to 20MB), we output results into memory mapped files. Have you tried decreasing that threshold?
 
Couple of JIRAs you may want to take a look at: PHOENIX-2405 (unclear if this is still an issue) and PHOENIX-3289. Are you running into issues with too many memory mapped files?

solzy:
*@JamesRTaylor * Thanks for your clarify.


----------------------------------------
   Yun Zhang
   Best regards!


2018-07-17 6:18 GMT+08:00 James Taylor <notifications@github.com>:

phoenix-core/src/main/java/org/apache/phoenix/iterate/ClientHashAggregatingResultIterator.java
geraldss: Hi @JamesRTaylor, thanks for the review and feedback. I'm hoping to get a minimum implementation merged into master, and it can then improve over time. I see that HashCacheClient throws an exception after the memory threshold is reached. Would it suffice if I do the same thing?

The other examples, e.g. MappedByteBufferSortedQueue, involve linear memory.

geraldss: @JamesRTaylor @solzy I have added a memory limitation based on SPOOL_THRESHOLD_BYTES. It throws an exception similar to HashCacheClient. Please review. Thanks.

phoenix-core/src/main/java/org/apache/phoenix/execute/ClientAggregatePlan.java
```diff
@@ -135,17 +142,24 @@ public ResultIterator iterator(ParallelScanGrouper scanGrouper, Scan scan) throw
             aggResultIterator = new ClientUngroupedAggregatingResultIterator(LookAheadResultIterator.wrap(iterator), serverAggregators);
             aggResultIterator = new UngroupedAggregatingResultIterator(LookAheadResultIterator.wrap(aggResultIterator), clientAggregators);
         } else {
-            if (!groupBy.isOrderPreserving()) {
-                int thresholdBytes = context.getConnection().getQueryServices().getProps().getInt(
-                        QueryServices.SPOOL_THRESHOLD_BYTES_ATTRIB, QueryServicesOptions.DEFAULT_SPOOL_THRESHOLD_BYTES);
-                List<Expression> keyExpressions = groupBy.getKeyExpressions();
+            List<Expression> keyExpressions = groupBy.getKeyExpressions();
+            if (groupBy.isOrderPreserving()) {
+                aggResultIterator = new ClientGroupedAggregatingResultIterator(LookAheadResultIterator.wrap(iterator), serverAggregators, keyExpressions);
```
JamesRTaylor: Pass through context here too to ClientGroupedAggregatingResultIterator as you'll need it to get the memory manager.

phoenix-core/src/main/java/org/apache/phoenix/iterate/ClientHashSizeException.java
```diff
@@ -0,0 +1,33 @@
...
+ */
+
+package org.apache.phoenix.iterate;
+
+/**
+ * Thrown by {@link org.apache.phoenix.iterate.ClientHashAggregatingResultIterator } when
+ * hash size exceeds memory threshold.
+ * 
+ */
+public class ClientHashSizeException extends RuntimeException {
```
JamesRTaylor: You won't need this as an InsufficientMemoryException will be thrown if you go above the specified memory limit (based on existing Phoenix config properties) and this will be unwound to become a SQLException with the code SQLExceptionCode.INSUFFICIENT_MEMORY.

phoenix-core/src/main/java/org/apache/phoenix/iterate/ClientHashAggregatingResultIterator.java
```diff
@@ -0,0 +1,173 @@
...
+
+        for (Tuple result = resultIterator.next(); result != null; result = resultIterator.next()) {
+            ImmutableBytesWritable key = new ImmutableBytesWritable(UNITIALIZED_KEY_BUFFER);
+            key = getGroupingKey(result, key);
+            Aggregator[] rowAggregators = hash.get(key);
+            if (rowAggregators == null) {
+                // Abort if we exceed memory threshold/2
+                // We use threshold/2 to leave room for the subsequent sort
+                if (thresholdBytes > 0) {
+                    hashSize += key.getSize() + aggSize;
```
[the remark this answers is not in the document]
geraldss: Done.

phoenix-core/src/main/java/org/apache/phoenix/iterate/ClientHashAggregatingResultIterator.java
```diff
@@ -0,0 +1,173 @@
...
+    // Copied from ClientGroupedAggregatingResultIterator
+    protected Tuple wrapKeyValueAsResult(KeyValue keyValue) {
+        return new MultiKeyValueTuple(Collections.<Cell> singletonList(keyValue));
+    }
+
+    private HashMap<ImmutableBytesWritable, Aggregator[]> populateHash() throws SQLException {
+        hash = new HashMap<ImmutableBytesWritable, Aggregator[]>(HASH_AGG_INIT_SIZE, 0.75f);
+
+        final int aggSize = aggregators.getEstimatedByteSize();
+        int hashSize = 0;
```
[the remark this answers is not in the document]
geraldss: Done.

phoenix-core/src/main/java/org/apache/phoenix/execute/ClientAggregatePlan.java
geraldss: Done.

phoenix-core/src/main/java/org/apache/phoenix/iterate/ClientHashSizeException.java
geraldss: I left this in as an additional check that uses SPOOL_THRESHOLD_BYTES. So both the MemoryManager and SPOOL_THRESHOLD_BYTES are enforced, in that order. I hope this is ok.

geraldss:
@JamesRTaylor - I made the changes and they are ready for review. Thanks.

phoenix-core/src/main/java/org/apache/phoenix/iterate/ClientHashAggregatingResultIterator.java
```diff
@@ -0,0 +1,189 @@
...
+                long hashSize = SizedUtil.sizeOfMap(hash.size() + 1, key.getSize(), aggSize);
+                if (hashSize > memoryChunk.getSize() + CLIENT_HASH_AGG_MEMORY_CHUNK_SIZE) {
+                    // This will throw InsufficientMemoryException if necessary
+                    memoryChunk.resize(hashSize + CLIENT_HASH_AGG_MEMORY_CHUNK_SIZE);
+                }
+
+                // Abort if we exceed memory threshold/2
+                // We use threshold/2 to leave room for the subsequent sort
+                if (thresholdBytes > 0) {
+                    if (hashSize > thresholdBytes/2) {
```
JamesRTaylor: I don't think we need this since we're tracking memory in a different way now. This would put a constraint on each client side aggregation (rather than memory usage as a whole). Do you think that additional constraint is necessary and if so why? FYI, you can also set an absolute byte amount for memory usage (but this is mainly used in tests). If you want to keep this, then let's create a different config parameter in QueryServices and a default in QueryServiesOptions. The spooling threshold is not a memory constraint. It's tells Phoenix when to start spooling to disk during a client sort, so it's not really appropriate to use here.

phoenix-core/src/main/java/org/apache/phoenix/iterate/ClientHashAggregatingResultIterator.java
```diff
@@ -0,0 +1,189 @@
...
+        this.groupByExpressions = groupByExpressions;
+        this.thresholdBytes = thresholdBytes;
+        memoryChunk = context.getConnection().getQueryServices().getMemoryManager().allocate(CLIENT_HASH_AGG_MEMORY_CHUNK_SIZE);
+    }
+
+    @Override
+        public Tuple next() throws SQLException {
+        if (keyIterator == null) {
+            hash = populateHash();
+            keyList = sortKeys();
```
JamesRTaylor: I don't think this sort is required at all. You've already guaranteed that each row has a unique row key, so the subsequent GroupedAggregatingResultIterator should work fine (that expects duplicate rows to be adjacent and since every row is unique, that'll be the case). If there's an ORDER BY for the groups, the Phoenix will insert an ordering result iterator.

phoenix-core/src/main/java/org/apache/phoenix/iterate/ClientHashAggregatingResultIterator.java
```diff
@@ -0,0 +1,189 @@
...
+        Tuple tuple = wrapKeyValueAsResult(KeyValueUtil.newKeyValue(key, SINGLE_COLUMN_FAMILY, SINGLE_COLUMN, AGG_TIMESTAMP, value, 0, value.length));
+        return tuple;
+    }
+
+    @Override
+        public void close() throws SQLException {
+        keyIterator = null;
+        keyList = null;
+        hash = null;
+        memoryChunk.resize(0);
```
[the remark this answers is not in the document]
geraldss: Done.

phoenix-core/src/main/java/org/apache/phoenix/iterate/ClientHashAggregatingResultIterator.java
```diff
@@ -0,0 +1,189 @@
...
+
+        hash = new HashMap<ImmutableBytesWritable, Aggregator[]>(HASH_AGG_INIT_SIZE, 0.75f);
+        final int aggSize = aggregators.getEstimatedByteSize();
+
+        for (Tuple result = resultIterator.next(); result != null; result = resultIterator.next()) {
+            ImmutableBytesWritable key = new ImmutableBytesWritable(UNITIALIZED_KEY_BUFFER);
+            key = getGroupingKey(result, key);
+            Aggregator[] rowAggregators = hash.get(key);
+            if (rowAggregators == null) {
+                long hashSize = SizedUtil.sizeOfMap(hash.size() + 1, key.getSize(), aggSize);
```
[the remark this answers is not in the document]
geraldss: Done.

phoenix-core/src/main/java/org/apache/phoenix/iterate/ClientHashAggregatingResultIterator.java
geraldss: Done, removed.

phoenix-core/src/main/java/org/apache/phoenix/iterate/ClientHashAggregatingResultIterator.java
geraldss: I tried avoiding the sort, but could not make the code work without it. In my use case, if there's an ORDER BY after the groups, and the ORDER BY was by primary key, it was not being applied. If the ORDER BY was not by primary key, it was being applied.

I left this sort in because (1) it introduces no error; (2) the sorting is after the grouping, so more efficient than the current solution, which sorts before the grouping; (3) it avoids the problem above.

geraldss:
@JamesRTaylor - pushed another change per your feedback. Fingers crossed :) Thanks.

phoenix-core/src/main/java/org/apache/phoenix/iterate/ClientHashAggregatingResultIterator.java
JamesRTaylor: See my comment in ClientAggregatePlan.java. I believe you can detect this corner case there and not need this sort. It's a little weird to have a hash aggregation that still does a sort (but I get your point about it being better than doing the sort before).

geraldss: @JamesRTaylor - I tried that as well, as part off trying to avoid this sort. That is, I tried catching this special case in ClientAggregatePlan and wrapping an OrderedAggregatingResultIterator there. It did not work, which is why I relented and implemented this sort this way.

One other detail -- for the special case, this sort is faster than the generic sort (it's less general). Bottom line, I have been unable to make the patch meet correctness without this sort.

phoenix-core/src/main/java/org/apache/phoenix/execute/ClientAggregatePlan.java
```diff
@@ -135,17 +142,24 @@ public ResultIterator iterator(ParallelScanGrouper scanGrouper, Scan scan) throw
...
+                if (useHashAgg) {
+                    aggResultIterator = new ClientHashAggregatingResultIterator(context, iterator, serverAggregators, keyExpressions);
+                } else {
+                    iterator = new OrderedResultIterator(iterator, keyExpressionOrderBy, thresholdBytes, null, null, projector.getEstimatedRowByteSize());
+                    aggResultIterator = new ClientGroupedAggregatingResultIterator(LookAheadResultIterator.wrap(iterator), serverAggregators, keyExpressions);
+                }
             }
-            aggResultIterator = new ClientGroupedAggregatingResultIterator(LookAheadResultIterator.wrap(iterator), serverAggregators, groupBy.getKeyExpressions());
             aggResultIterator = new GroupedAggregatingResultIterator(LookAheadResultIterator.wrap(aggResultIterator), clientAggregators);
         }
```
JamesRTaylor: In the below if statement, you should still insert the OrderedAggregatingResultIterator if a hash aggregation is being done and orderBy != OrderBy.EMPTY_ORDER_BY:

    // Still sort the aggregated rows if we're using a hash aggregation and the order by was optimized out
    // since the rows won't be in GROUP BY key order
    if (orderBy.getOrderByExpressions().isEmpty() && (!useHashAgg || orderBy != OrderBy.EMPTY_ORDER_BY)) {

geraldss: @JamesRTaylor - Done. Please review. Thanks.

geraldss: Hi @JamesRTaylor - please review. The sort is now done only when necessary. Thanks.

phoenix-core/src/it/java/org/apache/phoenix/end2end/ClientHashAggregateIT.java
```diff
@@ -0,0 +1,191 @@
...
+import java.sql.PreparedStatement;
+import java.sql.ResultSet;
+import java.sql.Statement;
+import java.util.Properties;
+
+import org.apache.phoenix.util.PropertiesUtil;
+import org.apache.phoenix.util.QueryUtil;
+import org.junit.Test;
+
+public class ClientHashAggregateIT extends ParallelStatsDisabledIT {
```
JamesRTaylor: Add tests for when sort (forward & reverse) required for hash aggregate.

phoenix-core/src/main/java/org/apache/phoenix/execute/ClientAggregatePlan.java
```diff
@@ -135,17 +142,25 @@ public ResultIterator iterator(ParallelScanGrouper scanGrouper, Scan scan) throw
...
+                int thresholdBytes = context.getConnection().getQueryServices().getProps().getInt
+                    (QueryServices.SPOOL_THRESHOLD_BYTES_ATTRIB, QueryServicesOptions.DEFAULT_SPOOL_THRESHOLD_BYTES);
                 List<OrderByExpression> keyExpressionOrderBy = Lists.newArrayListWithExpectedSize(keyExpressions.size());
                 for (Expression keyExpression : keyExpressions) {
                     keyExpressionOrderBy.add(new OrderByExpression(keyExpression, false, true));
                 }
-                iterator = new OrderedResultIterator(iterator, keyExpressionOrderBy, thresholdBytes, null, null, projector.getEstimatedRowByteSize());
+
+                if (useHashAgg) {
+                    boolean sort = orderBy == OrderBy.FWD_ROW_KEY_ORDER_BY;
```
JamesRTaylor: Should be true if OrderBy.RVS_ROW_KEY_ORDER_BY too.

phoenix-core/src/main/java/org/apache/phoenix/execute/ClientAggregatePlan.java
```diff
@@ -135,17 +142,25 @@ public ResultIterator iterator(ParallelScanGrouper scanGrouper, Scan scan) throw
...
+                    aggResultIterator = new ClientHashAggregatingResultIterator(context, iterator, serverAggregators, keyExpressions, sort);
```
JamesRTaylor: You’ll need to pass through if forward or reverse scan. You might just pass through orderBy instead of the boolean. Better still would be to let the code below insert an Ordering result iterator so you wouldn’t  need the sort logic at all in your new iterator.

phoenix-core/src/main/java/org/apache/phoenix/execute/ClientAggregatePlan.java
```diff
@@ -183,13 +198,15 @@ public ExplainPlan getExplainPlan() throws SQLException {
...
-        if (!groupBy.isEmpty()) {
-            if (!groupBy.isOrderPreserving()) {
-                planSteps.add("CLIENT SORTED BY " + groupBy.getKeyExpressions().toString());
-            }
+        if (groupBy.isEmpty()) {
+            planSteps.add("CLIENT AGGREGATE INTO SINGLE ROW");
+        } else if (groupBy.isOrderPreserving()) {
             planSteps.add("CLIENT AGGREGATE INTO DISTINCT ROWS BY " + groupBy.getExpressions().toString());
+        } else if (useHashAgg) {
+            planSteps.add("CLIENT HASH AGGREGATE INTO DISTINCT ROWS BY " + groupBy.getExpressions().toString());
```
JamesRTaylor: Add  CLIENT SORTED BY line here if sorting required for hash aggregate.

geraldss: Done.

phoenix-core/src/main/java/org/apache/phoenix/execute/ClientAggregatePlan.java
geraldss: The GROUP BY cannot cause REV_ROW_KEY_ORDER_BY, because the GROUP BY cannot specific or produce descending keys. This is a pre-existing assumption + design in SORT_MERGE_JOIN. The SORT_MERGE_JOIN creates its own forward sort, and its Tracker reports a forward sort.

phoenix-core/src/main/java/org/apache/phoenix/execute/ClientAggregatePlan.java
geraldss: Same as for previous comment. The GROUP BY cannot produce a reverse sort.

geraldss:
Hi @JamesRTaylor, I pushed another change and replied with some comments. Please review, thanks.

JamesRTaylor:
Looks like the move of 5.x branch to become master messed up this PR. Can you please start a new pull request against the 4.x-HBase-1.4 branch (this is what was master before), @geraldss? FYI, the last step will be to get a test run with your patch in place to make sure there are no regressions. To do that, attach a .patch file to the JIRA (include the branch name in the name of the patch file) and click on the Submit Patch button. See http://phoenix.apache.org/contributing.html#Generate_a_patch for more info on that.

geraldss:
@JamesRTaylor -- ok, I will submit a new PR. Can you comment on my last push and comments?

geraldss:
@JamesRTaylor - Done. I switched the PR to 4.x-HBase-1.4, attached the .patch file, and issued "Submit Patch".

phoenix-core/src/it/java/org/apache/phoenix/end2end/ClientHashAggregateIT.java
JamesRTaylor: Unless I'm missing something, this needs a few more tests:
* verify CLIENT SORTED BY is present in explain plan when sort required for hash aggregate
* verify it's not in explain plan when sort not required
* verify query results when sort not required

phoenix-core/src/main/java/org/apache/phoenix/execute/ClientAggregatePlan.java
JamesRTaylor: Please add comment about reverse sort and point to JIRA (create one if there's not already one). Might be best to just have the code here such that once reverse sort is supported, the hash aggregate will just work.

geraldss:
@JamesRTaylor - Handled reverse sort; added explanatory comments about sorting; added tests for sorting and non-sorting, including EXPLAIN.

Pushed changes; uploaded new .patch file; resubmitted patch.

JamesRTaylor:
+1. Nice work, @geraldss. @twdsilva - would you (or maybe someone else?) have some spare cycles to check this in assuming the test run comes back clean?

geraldss:
Thanks @JamesRTaylor for the reviews and suggestions.

twdsilva:
@geraldss 
The patch has been committed, can you close the PR ?

geraldss:
@twdsilva - PR closed now.

PR #309, opened 2018-07-11
[Do Not Merge] PHOENIX-3817 Verify Replication using SQL conditions

akshita-malhotra:


akshita-malhotra:
@vincentpoon @karanmehta93 @aaraujo @gjacoby126 @twdsilva   Will appreciate your review feedback! Thanks.

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/VerifyReplicationTool.java
```diff
@@ -0,0 +1,477 @@
...
+                    Configuration conf = context.getConfiguration();
+                    byte[] targetStartRow = null;
+                    byte[] targetStopRow = null;
+                    // find source table split
+                    PhoenixInputSplit sourceInputSplit = (PhoenixInputSplit) context.getInputSplit();
+                    if(key != null) {
+                        targetStartRow = value.getSourceKey().get();
+                    }
+                    if(sourceInputSplit.getLength() != 0) {
+                        targetStopRow = sourceInputSplit.getKeyRange().getUpperRange();
```
akshita-malhotra: I am somewhat doubtful about setting the target start row using upper key range function. @twdsilva Do you see any problem with this?

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/util/PhoenixMapReduceUtil.java
```diff
@@ -157,6 +192,192 @@ public static void setOutput(final Job job, final String tableName,final String
...
+     * @return Input splits for the job
+     * @throws IOException If the region information for the splits cannot be retrieved
+     */
+    public static List<InputSplit> generateSplits(final QueryPlan qplan,
+            final List<KeyRange> splits, Configuration config) throws IOException {
+        Preconditions.checkNotNull(qplan);
+        Preconditions.checkNotNull(splits);
+
+        // Get the RegionSizeCalculator
+        org.apache.hadoop.hbase.client.Connection connection = ConnectionFactory.createConnection(config);
```
karanmehta93: Any particular reason as to why you removed the try with resources block over here?
I added it because of a memory leak that I had found out here: https://issues.apache.org/jira/browse/PHOENIX-4489

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/VerifyReplicationTool.java
```diff
@@ -0,0 +1,477 @@
...
+            }
+
+        }
+
+        private void getTargetResultSet(Configuration conf, byte[] startRow, byte[] stopRow) throws IOException, SQLException {
+            Scan targetScan = targetQueryPlan.getContext().getScan();
+            targetScan.setAttribute(BaseScannerRegionObserver.SKIP_REGION_BOUNDARY_CHECK,
+                    Bytes.toBytes(true));
+            if(startRow != null) {
+                targetScan.setStartRow(startRow);
```
karanmehta93: This API is deprecated. You might want to use another one.

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/VerifyReplicationTool.java
```diff
@@ -0,0 +1,477 @@
...
+            }
+            if(stopRow != null) {
+                targetScan.setStopRow(stopRow);
```
karanmehta93: This API is deprecated. You might want to use another one.

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/VerifyReplicationTool.java
```diff
@@ -0,0 +1,477 @@
...
+                } catch (SQLException e) {
+                    LOG.error("Unable to close peer result set", e);
+                }
+            }
+
+        }
+
+        private void getTargetResultSet(Configuration conf, byte[] startRow, byte[] stopRow) throws IOException, SQLException {
+            Scan targetScan = targetQueryPlan.getContext().getScan();
+            targetScan.setAttribute(BaseScannerRegionObserver.SKIP_REGION_BOUNDARY_CHECK,
```
karanmehta93: You don't need this since we skip this check in MR jobs anyways.

akshita-malhotra:
Thanks for the feedback @karanmehta93 ! I will looking into the changes.

phoenix-core/src/it/java/org/apache/phoenix/mapreduce/VerifyReplicationToolIT.java
```diff
@@ -0,0 +1,323 @@
...
+    }
+
+    private void upsertData() throws SQLException {
+        Connection conn = DriverManager.getConnection(getUrl());
+        PreparedStatement sourceStmt =
+                conn.prepareStatement(String.format(UPSERT_USER, sourceTableName));
+        PreparedStatement targetStmt =
+                conn.prepareStatement(String.format(UPSERT_USER, targetTableName));
+        // create 2 tenants with 10 users each in source table
+        for (int t = 0; t < 2; t++) {
```
karanmehta93: nit: you can simplify this by creating 2 separate loops, just like the one you did on Line 183-185. That will be easier to understand.

phoenix-core/src/it/java/org/apache/phoenix/mapreduce/VerifyReplicationToolIT.java
```diff
@@ -0,0 +1,323 @@
...
+        HMaster master = cluster.getMaster();
+        List<HRegionInfo> regions = admin.getTableRegions(table);
+        int numRegions = regions.size();
+        // split the last region in the table until we have the target number of regions
+        while (numRegions < targetRegions) {
+            HRegionInfo region = regions.get(numRegions - 1);
+            ServerName serverName =
+                    master.getAssignmentManager().getRegionStates()
+                            .getRegionServerOfRegion(region);
+            byte[] splitPoint =
```
karanmehta93: I am not sure if this is working as intended. You want the data to be on different regions, right, so that multiple map tasks can be spawned?
The first region has start key `''` and the split will give it the end key as `\x0` and the same will go on for all the splits. Hence all of your data will be in single region only. When I put a break point in `PhoenixInputFormat#getSplits()`, the query plan just returns 1 split for the tenant.

phoenix-core/src/it/java/org/apache/phoenix/mapreduce/VerifyReplicationToolIT.java
```diff
@@ -0,0 +1,323 @@
...
+        sourceTenants.add("tenant" + tenantNum++);
+        sourceTenants.add("tenant" + tenantNum++);
+        sourceOnlyTenant = sourceTenants.get(0);
+        sourceAndTargetTenant = sourceTenants.get(1);
+        targetOnlyTenant = "tenant" + tenantNum++;
+        upsertData();
+        split(sourceTableName, 4);
+        split(targetTableName, 2);
+        // ensure scans for each table touch multiple region servers
+        ensureRegionsOnDifferentServers(sourceTableName);
```
karanmehta93: The method assumes that you have only two regions, however you have 4 regions for the first table. If I understand correctly, you are trying to split the table evenly between the 20 rows of 2 tenants and then assigning them on 2 region servers. However the implementation is not working in that manner.

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/VerifyReplicationTool.java
karanmehta93: The regionSize or Length variable can be 0 because certain config props may disable the regionSizeCalculator, in which case it will always return 0. I would advise to check the `getUpperRange()` to see if there is end to the range and assign it appropriately.

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/VerifyReplicationTool.java
```diff
@@ -0,0 +1,477 @@
...
+                        break;
+                    }
+                    ImmutableBytesWritable targetKey = null;
+                    if (targetResultSet.getCurrentRow() != null) {
+                        targetKey = new ImmutableBytesWritable();
+                        targetResultSet.getCurrentRow().getKey(targetKey);
+                    }
+
+                    int keyCompare;
+                    if(key == null){
```
karanmehta93: When can key be null?

karanmehta93: nit: Use ternary operator here

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/VerifyReplicationTool.java
```diff
@@ -0,0 +1,477 @@
...
+
+        @Override
+        protected void cleanup(Context context) throws IOException, InterruptedException {
+            super.cleanup(context);
+            try {
+                if (!sourceHasData && targetResultSet == null) {
+                    getTargetResultSet(context.getConfiguration(), null, null);
+                }
+                if(targetResultSet != null) {
+                    while (targetHasNext) {
```
karanmehta93: nit: The while condition can use `targetResultSet.next()`

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/VerifyReplicationTool.java
```diff
@@ -0,0 +1,477 @@
...
+                        ImmutableBytesWritable targetKey = null;
+
+                        if (targetResultSet.getCurrentRow() != null) {
+                            targetKey = new ImmutableBytesWritable();
+                            targetResultSet.getCurrentRow().getKey(targetKey);
```
karanmehta93: We are allocating new byte[] for every row. Can this potentially cause OOME inside a container if the size of region is large? Could you reuse these buffer? @apurtell @aaraujo

phoenix-core/src/it/java/org/apache/phoenix/mapreduce/VerifyReplicationToolIT.java
akshita-malhotra: @karanmehta93 : This is part of the code which @aaraujo originally wrote. I didn't get a chance to dive deeper into it yet, I am still adding more test cases. You should see the functionality change in the next update. Thanks!

twdsilva: Karan is right, you will have all the rows on a single region, and the remaining regions without any rows. 
You can use ByteUtil.concat(tenantId, QueryConstants.SEPARATOR_BYTE_ARRAY, userId) to split on the rowkey of a particular row to ensure that rows will be on multiple regions.

phoenix-core/src/it/java/org/apache/phoenix/mapreduce/VerifyReplicationToolIT.java
twdsilva: See BaseTest.splitSystemCatalog which does something similar, splits the table into multiple regions and moves each of them to their own region server. If you can generalize that method, you can just call it from this test.

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/VerifyReplicationTool.java
twdsilva: @karanmehta93  it always uses RegionSizeCalculator to generate the splits right? I don't see a problem with setting the targetStopRow based on the upper range of the split. Will the key ever be null? If it is, should the targetStartRow be set to the lower range of the split?

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/VerifyReplicationTool.java
twdsilva: ImmutableByteWritable does not allocate a new byte[]. Also I think this is only called if there are more rows on the target than the source for this map task.

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/VerifyReplicationTool.java
```diff
@@ -0,0 +1,477 @@
...
+                    byte[] targetStopRow = null;
+                    // find source table split
+                    PhoenixInputSplit sourceInputSplit = (PhoenixInputSplit) context.getInputSplit();
+                    if(key != null) {
+                        targetStartRow = value.getSourceKey().get();
+                    }
+                    if(sourceInputSplit.getLength() != 0) {
+                        targetStopRow = sourceInputSplit.getKeyRange().getUpperRange();
+                    }
+                    getTargetResultSet(conf, targetStartRow, targetStopRow);
```
twdsilva: Does each map task issue a scan over the target table for with the range of rows withing the map task?

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/VerifyReplicationTool.java
akshita-malhotra: @karanmehta93 one of the test covers the region size = 0 scenario, it seems to be working fine. I will check more. 
@twdsilva The key cannot be null, if the source splits doesn't have results it never enter the map task based on the record reader logic.

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/VerifyReplicationTool.java
akshita-malhotra: @twdsilva yes, for every map task the corresponding target scan is fetched. 
> didn't get the latter part of your statement 'target table for with the range of rows withing the map task'. Can you elaborate?

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/VerifyReplicationTool.java
akshita-malhotra: Key cannot be null. It got left from a different approach I was testing earlier. I will remove it.

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/VerifyReplicationTool.java
akshita-malhotra: @twdsilva That's correct.

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/VerifyReplicationTool.java
akshita-malhotra: Oh yeah, that's right. Thanks for pointing it out.

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/VerifyReplicationTool.java
karanmehta93: > one of the test covers the region size = 0 scenario, it seems to be working fine. I will check more

I think all tests are getting region size as 0.

karanmehta93:
It would be good to add some comments @akshita-malhotra once you are done with the dev part.

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/VerifyReplicationTool.java
karanmehta93: > ImmutableByteWritable does not allocate a new byte[]. Also I think this is only called if there are more rows on the target than the source for this map task.

Yes, I saw that. The line 221 will allocate a new byte[] for every row though. I assume that we might not have a huge deviation between rows, so it should be fine.

akshita-malhotra:
@karanmehta93 Updated the test suite based on your comments.
@twdsilva Reused the split system catalog logic as per your suggestion, seems to be working well.
It would be great if you can take another pass at the PS.
@gjacoby126 @vincentpoon Please take a look whenever you get a chance.

akshita-malhotra:
@aaraujo : It will be great if you can take a look as well. Thank you

phoenix-core/src/it/java/org/apache/phoenix/mapreduce/VerifyReplicationToolIT.java
```diff
@@ -0,0 +1,280 @@
...
+        sourceTenants.add("tenant" + tenantNum++);
+        sourceTenants.add("tenant" + tenantNum++);
+        sourceOnlyTenant = sourceTenants.get(0);
+        sourceAndTargetTenant = sourceTenants.get(1);
+        targetOnlyTenant = "tenant" + tenantNum++;
+        /* upsert data and spilt table into multiple regions
+        with each relevant region based on splitpoints
+        residing on a different RS*/
+        upsertData();
+        splitSource(sourceTableName);
```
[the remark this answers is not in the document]
akshita-malhotra: Thank you @twdsilva.I will add the comment. The PR should be ready soon, will wait today in case there is anymore feedback.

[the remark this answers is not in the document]
twdsilva: @akshita-malhotra 
Can you also please squash the commits and attach a .patch file to the JIRA and click submit patch?

phoenix-core/src/it/java/org/apache/phoenix/mapreduce/VerifyReplicationToolIT.java
```diff
@@ -0,0 +1,284 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.phoenix.mapreduce;
+
+import java.io.IOException;
+import java.sql.*;
```
karanmehta93: nit: no wildcard imports

phoenix-core/src/it/java/org/apache/phoenix/mapreduce/VerifyReplicationToolIT.java
```diff
@@ -0,0 +1,284 @@
...
+        verify(String.format("TENANT_ID = '%s'", sourceOnlyTenant), 0, 0, 10, 10, 0, 0);
+    }
+
+    @Test
+    public void testVerifyRowsDifferent() throws Exception {
+        // change three rows on the source table so they no longer match on the target
+        upsertSelectData(sourceTableName, sourceAndTargetTenant, -1, 3);
+        verify(String.format("TENANT_ID = '%s'", sourceAndTargetTenant), 0, 7, 3, 0, 0, 3);
+    }
+
```
gjacoby126: Could we also have a test that verifies that it does the right thing if there are both source-only and target-only rows? I did a whiteboard exercise to prove to myself that it does, but would be good to have it in code too.

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/util/PhoenixMapReduceUtil.java
```diff
@@ -157,6 +192,216 @@ public static void setOutput(final Job job, final String tableName,final String
...
+    public static ResultIterator initializeIterator(Configuration conf, List<Scan> scans,
+            QueryPlan queryPlan) throws SQLException, IOException {
+        List<PeekingResultIterator> iterators = Lists.newArrayListWithExpectedSize(scans.size());
+        StatementContext ctx = queryPlan.getContext();
+        ReadMetricQueue readMetrics = ctx.getReadMetricsQueue();
+        String tableName = queryPlan.getTableRef().getTable().getPhysicalName().getString();
+        String snapshotName = conf.get(PhoenixConfigurationUtil.SNAPSHOT_NAME_KEY);
+
+        // Clear the table region boundary cache to make sure long running jobs stay up to date
+        byte[] tableNameBytes = queryPlan.getTableRef().getTable().getPhysicalName().getBytes();
```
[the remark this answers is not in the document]
akshita-malhotra: This code is just part of the general refactor that Alex did in his original PR, the code was already there. Only the functions were moved.

akshita-malhotra:
@gjacoby126 Thanks a lot for the review. I will incorporate the changes and update the PR and jira patch.

aaraujo:
Hey @akshita-malhotra, I'll take a look at your next patch!

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/util/PhoenixMapReduceUtil.java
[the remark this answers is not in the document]
twdsilva: I think the code to clear the table region boundary cache was added as part of PHOENIX-2599. Since the MR/spark/pig integrations does not support order by which requires rows from a scan to be in sorted order we can skip the region boundary check in BaseScannerRegionObserver.preScannerOpen and also clear the region boundary cache. 
(see the discussion on PHOENIX-2599).

phoenix-core/src/it/java/org/apache/phoenix/mapreduce/VerifyReplicationToolIT.java
akshita-malhotra: @gjacoby126 Sorry Wanted to clarify the test scenario, so there should be no common rows between the source and target and what input condition should I verify upon?
Or the source table should have both source-only and target-only rows? A little confused, didn't realize it earlier.

gjacoby126: @akshita-malhotra - setup is we have Cluster A and Cluster B:
1. Some rows exist only in cluster A
2. Some rows exist in both A and B
3. Some rows exist only in B. 

Assert that your counters are correct. It's essentially combining two of your existing test cases to make sure that it still works with both cases present.

akshita-malhotra: Thanks for the clarification @gjacoby126. I updated the PR with the test and the nit. 
There is some refactored code that I need to periodically rebase, I will check if that's needed and update the patch on jira.

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/util/PhoenixConfigurationUtil.java
```diff
@@ -371,20 +396,31 @@ public static void setUpsertStatement(final Configuration configuration, String
...
         Preconditions.checkNotNull(configuration);
-        String selectStmt = configuration.get(SELECT_STATEMENT);
+        String selectKey = (isTargetStatement) ? SELECT_TARGET_STATEMENT : SELECT_STATEMENT;
+        String selectStmt = configuration.get(selectKey);
         if(isNotEmpty(selectStmt)) {
             return selectStmt;
         }
-        final String tableName = getInputTableName(configuration);
+        final String tableName;
+        if (isTargetStatement) {
```
karanmehta93: nit: ternary operator

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/util/PhoenixConfigurationUtil.java
```diff
@@ -371,20 +396,31 @@ public static void setUpsertStatement(final Configuration configuration, String
...
+        return getSelectStatement(configuration, false);
+    }
+
+    public static String getSelectStatement(final Configuration configuration,
+            boolean isTargetStatement) throws SQLException {
         Preconditions.checkNotNull(configuration);
-        String selectStmt = configuration.get(SELECT_STATEMENT);
+        String selectKey = (isTargetStatement) ? SELECT_TARGET_STATEMENT : SELECT_STATEMENT;
+        String selectStmt = configuration.get(selectKey);
         if(isNotEmpty(selectStmt)) {
```
karanmehta93: Please add a info LOG line here.

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/util/PhoenixMapReduceUtil.java
```diff
@@ -157,6 +192,216 @@ public static void setOutput(final Job job, final String tableName,final String
...
+     */
+    public static QueryPlan getQueryPlan(final Configuration configuration,
+            boolean isTargetConnection) throws SQLException {
+        Preconditions.checkNotNull(configuration);
+        try {
+            final String txnScnValue = configuration.get(PhoenixConfigurationUtil.TX_SCN_VALUE);
+            final String currentScnValue = configuration.get(PhoenixConfigurationUtil.CURRENT_SCN_VALUE);
+            final String tenantId = configuration.get(PhoenixConfigurationUtil.MAPREDUCE_TENANT_ID);
+            final Properties overridingProps = new Properties();
+            if(txnScnValue==null && currentScnValue!=null) {
```
karanmehta93: nit: spaces (applicable everywhere in general)

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/util/PhoenixMapReduceUtil.java
```diff
@@ -17,22 +17,57 @@
...
+import org.apache.hadoop.hbase.client.Scan;
+import org.apache.hadoop.hbase.mapreduce.RegionSizeCalculator;
+import org.apache.hadoop.hbase.util.Bytes;
+import org.apache.hadoop.mapreduce.InputSplit;
 import org.apache.hadoop.mapreduce.Job;
 import org.apache.hadoop.mapreduce.lib.db.DBWritable;
+import org.apache.phoenix.compile.QueryPlan;
+import org.apache.phoenix.compile.StatementContext;
+import org.apache.phoenix.coprocessor.BaseScannerRegionObserver;
+import org.apache.phoenix.iterate.*;
```
karanmehta93: nit: no wildcard import

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/util/PhoenixMapReduceUtil.java
```diff
@@ -157,6 +192,216 @@ public static void setOutput(final Job job, final String tableName,final String
...
+            final Scan scan = queryPlan.getContext().getScan();
+            // since we can't set a scn on connections with txn set TX_SCN attribute so that the
+            // max time range is set by BaseScannerRegionObserver
+            if (txnScnValue != null) {
+                scan.setAttribute(BaseScannerRegionObserver.TX_SCN, Bytes.toBytes(Long.valueOf(txnScnValue)));
+            }
+
+            // setting the snapshot configuration
+            String snapshotName = configuration.get(PhoenixConfigurationUtil.SNAPSHOT_NAME_KEY);
+            if (snapshotName != null)
```
karanmehta93: nit: braces

phoenix-core/src/test/java/org/apache/phoenix/query/BaseTest.java
```diff
@@ -1780,15 +1780,15 @@ private  static void verifySequence(String tenantID, String sequenceName, String
...
-    protected static void splitRegion(byte[] splitPoint) throws SQLException, IOException, InterruptedException {
+    protected static void splitRegion(TableName tableName, byte[] splitPoint) throws SQLException, IOException, InterruptedException {
         Admin admin =
                 driver.getConnectionQueryServices(getUrl(), TestUtil.TEST_PROPERTIES).getAdmin();
-        admin.split(PhoenixDatabaseMetaData.SYSTEM_CATALOG_HBASE_TABLE_NAME, splitPoint);
+        admin.split(tableName, splitPoint);
         // make sure the split finishes (there's no synchronous splitting before HBase 2.x)
-        admin.disableTable(PhoenixDatabaseMetaData.SYSTEM_CATALOG_HBASE_TABLE_NAME);
-        admin.enableTable(PhoenixDatabaseMetaData.SYSTEM_CATALOG_HBASE_TABLE_NAME);
+        admin.disableTable(tableName);
```
karanmehta93: Why do we need this? 
Regions go offline anyways when they split and `SplitRegionRequest` is coming from `BlockingInterface` in HBase 1.x versions.
For 2.x and above you can use `splitRegionSync` method

twdsilva: So that the split completes, or else we have to keep checking the number of regions.

phoenix-core/src/test/java/org/apache/phoenix/query/BaseTest.java
```diff
@@ -1894,10 +1877,36 @@ protected static void splitSystemCatalog(Map<String, List<String>> tenantToTable
...
+                // we don't allow SYSTEM.CATALOG to split within a schema, so to ensure each table
+                // or view is on a separate region they need to have a unique tenant and schema name
+                assertTrue("Schema names of tables/view must be unique ", schemaNameSet.add(tenantId+"."+schemaName));
+                String tableName = SchemaUtil.getTableNameFromFullName(fullName);
+                splitPoints.add(
+                    SchemaUtil.getTableKey(tenantId, "".equals(schemaName) ? null : schemaName, tableName));
+            }
+        }
+        Collections.sort(splitPoints, Bytes.BYTES_COMPARATOR);
+        splitTable(PhoenixDatabaseMetaData.SYSTEM_CATALOG_HBASE_TABLE_NAME, splitPoints);
```
karanmehta93: Does calling `split()` method explicitly check whether the table is allowed to split on that key or not by consulting the `MetaDataSplitPolicy `

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/VerifyReplicationSourceWritable.java
```diff
@@ -0,0 +1,59 @@
...
+    /** For serialization*/
+    public VerifyReplicationSourceWritable() {
+        super();
+    }
+
+    public ImmutableBytesWritable getSourceKey() {
+        return sourceKey;
+    }
+
+    public Map<String, Object> getSourceResults() {
```
karanmehta93: You don't need locally as a field.

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/VerifyReplicationSourceWritable.java
```diff
@@ -0,0 +1,59 @@
...
+        return sourceResults;
+    }
+
+
+    @Override
+    public void readFields(ResultSet resultSet) throws SQLException {
```
karanmehta93: I think you should add these changes to `PhoenixRecordWritable` directly. Should be a good addition over there.

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/PhoenixInputSplit.java
```diff
@@ -82,23 +82,31 @@ private void init() {
     public void readFields(DataInput input) throws IOException {
         regionLocation = WritableUtils.readString(input);
         regionSize = WritableUtils.readVLong(input);
+        scans = readScans(input);
+        init();
+    }
+
+    List<Scan> readScans(DataInput input) throws IOException {
```
karanmehta93: Please add a comment about how this input is serialized and de-serialized.

phoenix-core/src/test/java/org/apache/phoenix/query/BaseTest.java
twdsilva: This just calls admin.splitTable() which checks the split policy.

phoenix-core/src/test/java/org/apache/phoenix/query/BaseTest.java
karanmehta93: The call to split() method should get blocked until the split is complete, isn't it?

twdsilva: No its async in 1.x

karanmehta93: I am not sure if I am missing something over here, `BlockingService` proto service is being used for `AdminProtos` in HBase 1.3 or 1.4, which is the same being used for `MetaDataProtos` as well.

twdsilva: That ends up calling CompactSplitThread.requestSplit which just enqueues the task.

PR #312, opened 2018-07-23

m2je:
https://issues.apache.org/jira/browse/PHOENIX-3547
Currently the number of indices per Phoenix table is bound to maximum of 65535 (java.lang.short) which may limit some application requiring to have arbitrary unlimited number of indices. 
This change will consider any new table created in Phoenix to support indices ids to be in  -9,223,372,036,854,775,808 to 9,223,372,036,854,775,807 (java.lang.Long) range which is undoubtably big enough to cover this requirement.
Any existing Phoenix Phoenix table will still continue to support only maximum of 65535 of indices and will be able to maintain this limits as enforced before.


m2je:
needs more work

PR #313, opened 2018-07-24
PHOENIX-4799 Write cells using checkAndMutate to prevent conflicting …

twdsilva:
…changes

@vincentpoon @karanmehta93 @ChinmaySKulkarni  Can you please review? 
With this patch, when we add a column we write a cell to SYSTEM.CHILD_LINK with row key (tenantId, schemaName, physicalTableName, columnName) to prevent conflicting concurrent modifications. 
While dropping a table or creating a view we also write a cell to SYSTEM.CHILD_LINK with row key (tenantId, schemaName, physicalTableName). 
This is done in MetadataClient before we make an rpc to create view / drop table / add column.

### DIFF
--- a/phoenix-core/src/it/java/org/apache/phoenix/end2end/BasePermissionsIT.java
+++ b/phoenix-core/src/it/java/org/apache/phoenix/end2end/BasePermissionsIT.java
@@ -16,31 +16,11 @@
  */
 package org.apache.phoenix.end2end;
 
-import com.google.common.base.Joiner;
-import com.google.common.base.Throwables;
-import org.apache.commons.logging.Log;
-import org.apache.commons.logging.LogFactory;
-import org.apache.hadoop.conf.Configuration;
-import org.apache.hadoop.hbase.AuthUtil;
-import org.apache.hadoop.hbase.HBaseTestingUtility;
-import org.apache.hadoop.hbase.HConstants;
-import org.apache.hadoop.hbase.TableName;
-import org.apache.hadoop.hbase.security.AccessDeniedException;
-import org.apache.hadoop.hbase.security.User;
-import org.apache.hadoop.hbase.security.access.AccessControlClient;
-import org.apache.hadoop.hbase.security.access.Permission;
-import org.apache.phoenix.jdbc.PhoenixConnection;
-import org.apache.phoenix.jdbc.PhoenixDatabaseMetaData;
-import org.apache.phoenix.jdbc.PhoenixStatement;
-import org.apache.phoenix.query.BaseTest;
-import org.apache.phoenix.query.QueryConstants;
-import org.apache.phoenix.query.QueryServices;
-import org.apache.phoenix.util.PhoenixRuntime;
-import org.apache.phoenix.util.QueryUtil;
-import org.junit.After;
-import org.junit.BeforeClass;
-import org.junit.runner.RunWith;
-import org.junit.runners.Parameterized;
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertNotNull;
+import static org.junit.Assert.assertTrue;
+import static org.junit.Assert.fail;
 
 import java.io.IOException;
 import java.lang.reflect.UndeclaredThrowableException;
@@ -59,11 +39,31 @@ import java.util.List;
 import java.util.Properties;
 import java.util.Set;
 
-import static org.junit.Assert.assertEquals;
-import static org.junit.Assert.assertFalse;
-import static org.junit.Assert.assertNotNull;
-import static org.junit.Assert.assertTrue;
-import static org.junit.Assert.fail;
+import org.apache.commons.logging.Log;
+import org.apache.commons.logging.LogFactory;
+import org.apache.hadoop.conf.Configuration;
+import org.apache.hadoop.hbase.AuthUtil;
+import org.apache.hadoop.hbase.HBaseTestingUtility;
+import org.apache.hadoop.hbase.HConstants;
+import org.apache.hadoop.hbase.TableName;
+import org.apache.hadoop.hbase.security.AccessDeniedException;
+import org.apache.hadoop.hbase.security.User;
+import org.apache.hadoop.hbase.security.access.AccessControlClient;
+import org.apache.hadoop.hbase.security.access.Permission;
+import org.apache.phoenix.jdbc.PhoenixConnection;
+import org.apache.phoenix.jdbc.PhoenixDatabaseMetaData;
+import org.apache.phoenix.jdbc.PhoenixStatement;
+import org.apache.phoenix.query.BaseTest;
+import org.apache.phoenix.query.QueryConstants;
+import org.apache.phoenix.query.QueryServices;
+import org.apache.phoenix.util.PhoenixRuntime;
+import org.junit.After;
+import org.junit.BeforeClass;
+import org.junit.runner.RunWith;
+import org.junit.runners.Parameterized;
+
+import com.google.common.base.Joiner;
+import com.google.common.base.Throwables;
 
 @RunWith(Parameterized.class)
 public class BasePermissionsIT extends BaseTest {
@@ -73,17 +73,23 @@ public class BasePermissionsIT extends BaseTest {
     static String SUPERUSER;
 
     static HBaseTestingUtility testUtil;
-    static final Set<String> PHOENIX_SYSTEM_TABLES = new HashSet<>(Arrays.asList(
-            "SYSTEM.CATALOG", "SYSTEM.SEQUENCE", "SYSTEM.STATS", "SYSTEM.FUNCTION"));
+    static final Set<String> PHOENIX_SYSTEM_TABLES =
+            new HashSet<>(Arrays.asList("SYSTEM.CATALOG", "SYSTEM.SEQUENCE", "SYSTEM.STATS",
+                "SYSTEM.FUNCTION", "SYSTEM.MUTEX"));
 
-    static final Set<String> PHOENIX_SYSTEM_TABLES_IDENTIFIERS = new HashSet<>(Arrays.asList(
-            "SYSTEM.\"CATALOG\"", "SYSTEM.\"SEQUENCE\"", "SYSTEM.\"STATS\"", "SYSTEM.\"FUNCTION\""));
+    static final Set<String> PHOENIX_SYSTEM_TABLES_IDENTIFIERS =
+            new HashSet<>(Arrays.asList("SYSTEM.\"CATALOG\"", "SYSTEM.\"SEQUENCE\"",
+                "SYSTEM.\"STATS\"", "SYSTEM.\"FUNCTION\"", "SYSTEM.\"MUTEX\""));
 
     static final String SYSTEM_SEQUENCE_IDENTIFIER =
             QueryConstants.SYSTEM_SCHEMA_NAME + "." + "\"" + PhoenixDatabaseMetaData.SYSTEM_SEQUENCE_TABLE+ "\"";
 
+    static final String SYSTEM_MUTEX_IDENTIFIER =
+            QueryConstants.SYSTEM_SCHEMA_NAME + "." + "\""
+                    + PhoenixDatabaseMetaData.SYSTEM_MUTEX_TABLE_NAME + "\"";
+
     static final Set<String> PHOENIX_NAMESPACE_MAPPED_SYSTEM_TABLES = new HashSet<>(Arrays.asList(
-            "SYSTEM:CATALOG", "SYSTEM:SEQUENCE", "SYSTEM:STATS", "SYSTEM:FUNCTION"));
+            "SYSTEM:CATALOG", "SYSTEM:SEQUENCE", "SYSTEM:STATS", "SYSTEM:FUNCTION", "SYSTEM:MUTEX"));
 
     // Create Multiple users so that we can use Hadoop UGI to run tasks as various users
     // Permissions can be granted or revoke by superusers and admins only
--- a/phoenix-core/src/it/java/org/apache/phoenix/end2end/ViewIT.java
+++ b/phoenix-core/src/it/java/org/apache/phoenix/end2end/ViewIT.java
@@ -20,11 +20,13 @@ package org.apache.phoenix.end2end;
 import static com.google.common.collect.Lists.newArrayListWithExpectedSize;
 import static org.apache.phoenix.exception.SQLExceptionCode.CANNOT_MODIFY_VIEW_PK;
 import static org.apache.phoenix.exception.SQLExceptionCode.NOT_NULLABLE_COLUMN_IN_ROW_KEY;
+import static org.apache.phoenix.util.TestUtil.TEST_PROPERTIES;
 import static org.apache.phoenix.util.TestUtil.analyzeTable;
 import static org.apache.phoenix.util.TestUtil.getAllSplits;
 import static org.junit.Assert.assertArrayEquals;
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertNull;
 import static org.junit.Assert.assertTrue;
 import static org.junit.Assert.fail;
 
@@ -43,6 +45,13 @@ import java.util.Collections;
 import java.util.List;
 import java.util.Map;
 import java.util.Properties;
+import java.util.concurrent.Callable;
+import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.Future;
+import java.util.concurrent.ThreadFactory;
+import java.util.concurrent.TimeUnit;
 
 import org.apache.curator.shaded.com.google.common.collect.Lists;
 import org.apache.hadoop.hbase.DoNotRetryIOException;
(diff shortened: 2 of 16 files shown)


karanmehta93:
Just curious, why not use `SYSTEM.MUTEX` table for this? Its a table specially added for use-cases that need time based expiration and distributed locks.

twdsilva:
Thanks for the suggestion, I will update the patch to use SYSTEM.MUTEX

twdsilva:
@karanmehta93  I have updated the PR to use SYSTEM.MUTEX, please review.

phoenix-core/src/it/java/org/apache/phoenix/end2end/ViewIT.java
```diff
@@ -1274,28 +1341,145 @@ public void testChildViewCreationFails() throws Exception {
...
+			} catch (ConcurrentTableMutationException e) {
+			}
+			latch2.countDown();
+
+			Exception e = future.get();
+			assertTrue(e == null);
+			
+			// add a new column to the base table to ensure that the cell used
+			// to prevent concurrent modifications was removed
+			tableDdl = "ALTER TABLE " + fullTableName + " ADD v4 INTEGER";
```
ChinmaySKulkarni: You didn't execute this ddl statement

phoenix-core/src/it/java/org/apache/phoenix/end2end/ViewIT.java
```diff
@@ -1274,28 +1341,145 @@ public void testChildViewCreationFails() throws Exception {
...
+			try {
+				// add the same column to the base table with a different type
+				conn.createStatement().execute(tableDdl);
+				fail("Creating a view while concurrently dropping the base table should fail");
+			} catch (ConcurrentTableMutationException e) {
+			}
+			latch2.countDown();
+
+			Exception e = future.get();
+			assertTrue(e == null);
```
ChinmaySKulkarni: nit: `assertNull`

phoenix-core/src/it/java/org/apache/phoenix/end2end/ViewIT.java
```diff
@@ -123,6 +148,48 @@ public static void doSetup() throws Exception {
...
+		public void preBatchMutate(ObserverContext<RegionCoprocessorEnvironment> c,
+				MiniBatchOperationInProgress<Mutation> miniBatchOp) throws IOException {
+			if (shouldFail(c, miniBatchOp.getOperation(0))) {
+				// throwing anything other than instances of IOException result
+				// in this coprocessor being unloaded
+				// DoNotRetryIOException tells HBase not to retry this mutation
+				// multiple times
+				throw new DoNotRetryIOException();
+			} else if (shouldSlowDown(c, miniBatchOp.getOperation(0))) {
+				// simulate a slow write to SYSTEM.CATALOG
```
ChinmaySKulkarni: Nice!

phoenix-core/src/main/java/org/apache/phoenix/schema/MetaDataClient.java
```diff
@@ -1874,6 +1881,31 @@ private static boolean checkAndValidateRowTimestampCol(ColumnDef colDef, Primary
...
+				? SchemaUtil.getColumnKey(tenantId, schemaName, tableName, columnName, null)
+				: SchemaUtil.getTableKey(tenantId, schemaName, tableName);
+		boolean success = connection.getQueryServices().writeMutexCell(rowKey);
+		if (!success) {
+			throw new ConcurrentTableMutationException(schemaName, tableName);
+		}
+	}
+	
+	/**
+	 * Remove the cell that was written to to the SYSTEM.CHILD_LINK table with
```
ChinmaySKulkarni: Update the comment since you're using SYSMUTEX now

phoenix-core/src/main/java/org/apache/phoenix/schema/MetaDataClient.java
```diff
@@ -3034,6 +3086,10 @@ MutationState dropTable(String schemaName, String tableName, String parentTableN
             return new MutationState(0, 0, connection);
         } finally {
             connection.setAutoCommit(wasAutoCommit);
+            // lock the table to prevent concurrent table modifications
```
ChinmaySKulkarni: Update comment

ChinmaySKulkarni:
A few comments and nits. Do we have to do anything special to handle indexes on views? I guess a view on a view should be fine with this implementation since the physical table name would resolve to the final base table. @twdsilva

twdsilva:
We can't add/drop columns directly to an index on a view. Views with their own child views will write a mutex using the physical table.

@ChinmaySKulkarni and @karanmehta93  thanks for the feedback. I have updated the PR, please take a look.

phoenix-core/src/main/java/org/apache/phoenix/schema/MetaDataClient.java
```diff
@@ -3604,6 +3675,17 @@ public MutationState addColumn(PTable table, List<ColumnDef> origColumnDefs,
...
         } finally {
             connection.setAutoCommit(wasAutoCommit);
+            if (!columns.isEmpty()) {
+                for (PColumn pColumn : columns) {
+                    PName physicalName = table.getPhysicalName();
+                    String physicalSchemaName =
+                            SchemaUtil.getSchemaNameFromFullName(physicalName.getString());
+                    String physicalTableName =
+                            SchemaUtil.getTableNameFromFullName(physicalName.getString());
+                    deleteCell(null, physicalSchemaName, physicalTableName,
```
ankitsinghal: @twdsilva , shouldn't you check whether you have acquired a lock(or inserted the cell in Mutex) before deleting the lock cell, because here you might be deleting the locks acquired by some other threads. And, You may need to do this per column I think.

phoenix-core/src/main/java/org/apache/phoenix/schema/MetaDataClient.java
```diff
@@ -3034,6 +3088,11 @@ MutationState dropTable(String schemaName, String tableName, String parentTableN
             return new MutationState(0, 0, connection);
         } finally {
             connection.setAutoCommit(wasAutoCommit);
+            // acquire a mutex on the table to prevent creating views while concurrently
+            // dropping the base table
+            if (tableType == PTableType.TABLE) {
+                deleteCell(null, schemaName, tableName, null);
+            }
```
ankitsinghal: nit: Please update the comment here that you are releasing a mutex.

twdsilva:
@ankitsinghal  I filed PHOENIX-4765 to not allow metadata changes on a base table that has child views when the request is from an older client. This will also allow us to rollback the upgraded server side jar if required.

phoenix-core/src/main/java/org/apache/phoenix/schema/MetaDataClient.java
twdsilva: Done.

phoenix-core/src/main/java/org/apache/phoenix/schema/MetaDataClient.java
twdsilva: I changed the code to only delete the cells if we were able to successfully do the checkAndPut. 
When we add a column we write a cell per column that we are creating. 
When we drop a base table or create a view, we write a single cell with the rowkey of the physical table (to prevent a view being created while we are dropping the base table).

twdsilva:
@ankitsinghal  Thanks for the review. I have updated the PR to based on the feedback. Can you please take a look?

phoenix-core/src/main/java/org/apache/phoenix/schema/MetaDataClient.java
```diff
@@ -3604,6 +3681,18 @@ public MutationState addColumn(PTable table, List<ColumnDef> origColumnDefs,
             }
         } finally {
             connection.setAutoCommit(wasAutoCommit);
+            if (acquiredMutex && !columns.isEmpty()) {
```
ankitsinghal: where is acquiredMutex set to true?

twdsilva: @ankitsinghal  Sorry, I didn't update the PR with all my latest changes, I fixed this. Can you please review?

phoenix-core/src/it/java/org/apache/phoenix/end2end/BasePermissionsIT.java
```diff
@@ -82,6 +82,10 @@
     static final String SYSTEM_SEQUENCE_IDENTIFIER =
             QueryConstants.SYSTEM_SCHEMA_NAME + "." + "\"" + PhoenixDatabaseMetaData.SYSTEM_SEQUENCE_TABLE+ "\"";
 
+    static final String SYSTEM_MUTEX_IDENTIFIER =
+            QueryConstants.SYSTEM_SCHEMA_NAME + "." + "\""
+                    + PhoenixDatabaseMetaData.SYSTEM_MUTEX_TABLE_NAME + "\"";
+
     static final Set<String> PHOENIX_NAMESPACE_MAPPED_SYSTEM_TABLES = new HashSet<>(Arrays.asList(
```
karanmehta93: Do you need to SYSTEM:MUTEX over here as well since now its visible from Phoenix level as well?

phoenix-core/src/main/java/org/apache/phoenix/query/ConnectionQueryServicesImpl.java
```diff
@@ -2723,6 +2712,9 @@ private void createOtherSystemTables(PhoenixConnection metaConnection, HBaseAdmi
         try {
             metaConnection.createStatement().executeUpdate(getChildLinkDDL());
         } catch (TableAlreadyExistsException e) {}
+        try {
+            metaConnection.createStatement().executeUpdate(getMutexDDL());
+        } catch (TableAlreadyExistsException e) {}
         // Catch the IOException to log the error message and then bubble it up for the client to retry.
         try {
             createSysMutexTableIfNotExists(hbaseAdmin);
```
karanmehta93: Do we still need this part of code?

phoenix-core/src/main/java/org/apache/phoenix/query/ConnectionQueryServicesImpl.java
```diff
@@ -3410,17 +3405,10 @@ void ensureSystemTablesMigratedToSystemNamespace()
             // No tables exist matching "SYSTEM\..*", they are all already in "SYSTEM:.*"
             if (tableNames.size() == 0) { return; }
             // Try to move any remaining tables matching "SYSTEM\..*" into "SYSTEM:"
-            if (tableNames.size() > 5) {
+            if (tableNames.size() > 7) {
                 logger.warn("Expected 5 system tables but found " + tableNames.size() + ":" + tableNames);
```
karanmehta93: nit: change log message appropriately. Extract the constant to a variable

phoenix-core/src/main/java/org/apache/phoenix/schema/MetaDataClient.java
```diff
@@ -1913,6 +1940,21 @@ private PTable createTableInternal(CreateTableStatement statement, byte[][] spli
             boolean isLocalIndex = indexType == IndexType.LOCAL;
             QualifierEncodingScheme encodingScheme = NON_ENCODED_QUALIFIERS;
             ImmutableStorageScheme immutableStorageScheme = ONE_CELL_PER_COLUMN;
+            
+            if (tableType == PTableType.VIEW) {
```
karanmehta93: nit: Would be good to have some logs here for debugging purposes.

phoenix-core/src/main/java/org/apache/phoenix/schema/MetaDataClient.java
```diff
@@ -2957,6 +3010,11 @@ MutationState dropTable(String schemaName, String tableName, String parentTableN
                 Delete linkDelete = new Delete(linkKey, clientTimeStamp);
                 tableMetaData.add(linkDelete);
             }
+            if (tableType == PTableType.TABLE) {
+                // acquire a mutex on the table to prevent creating views while concurrently
+                // dropping the base table
+                acquiredMutex = writeCell(null, schemaName, tableName, null);
```
karanmehta93: Nothing is prevent even if `acquiredMutex` is false here. I think you should throw `ConcurrentTableMutationException` here as well? You might might wanna push down that logic to `writeCell()` method?

phoenix-core/src/it/java/org/apache/phoenix/end2end/BasePermissionsIT.java
```diff
@@ -16,8 +16,29 @@
  */
 package org.apache.phoenix.end2end;
 
-import com.google.common.base.Joiner;
-import com.google.common.base.Throwables;
+import static org.junit.Assert.assertEquals;
```
karanmehta93: nit: Diff generated due to change in order of imports

phoenix-core/src/main/java/org/apache/phoenix/query/QueryConstants.java
```diff
@@ -355,5 +355,19 @@
...
+	            TENANT_ID + " VARCHAR NULL," +
+	            TABLE_SCHEM + " VARCHAR NULL," +
+	            TABLE_NAME + " VARCHAR NOT NULL," +
+	            COLUMN_NAME + " VARCHAR NULL," + // null for table row
+	            COLUMN_FAMILY + " VARCHAR NULL " + // using for CF to uniqueness for columns
+	            "CONSTRAINT " + SYSTEM_TABLE_PK_NAME + " PRIMARY KEY (" + TENANT_ID + ","
+	            + TABLE_SCHEM + "," + TABLE_NAME + "," + COLUMN_NAME + "," + COLUMN_FAMILY + "))\n" +
+	            HConstants.VERSIONS + "=%s,\n" +
+	            HColumnDescriptor.KEEP_DELETED_CELLS + "=%s,\n" +
+	            PhoenixDatabaseMetaData.TRANSACTIONAL + "=" + Boolean.FALSE;
```
ankitsinghal: Why there is a need of creating a Phoenix managed table for mutex? 
And also API in QueryServices.writeMutexCell(byte[] rowKey) and deleteMutexCell(byte[] rowKey) don't enforce the schema of the table will be followed.

phoenix-core/src/main/java/org/apache/phoenix/query/ConnectionQueryServicesImpl.java
```diff
@@ -2723,13 +2712,10 @@ private void createOtherSystemTables(PhoenixConnection metaConnection, HBaseAdmi
...
             metaConnection.createStatement().executeUpdate(getChildLinkDDL());
         } catch (TableAlreadyExistsException e) {}
-        // Catch the IOException to log the error message and then bubble it up for the client to retry.
         try {
-            createSysMutexTableIfNotExists(hbaseAdmin);
-        } catch (IOException exception) {
-            logger.error("Failed to created SYSMUTEX table. Upgrade or migration is not possible without it. Please retry.");
-            throw exception;
-        }
+            metaConnection.createStatement().executeUpdate(getMutexDDL());
```
ankitsinghal: Shouldn't the mutex table available and have acquired a mutex already for the upgrade before you call createOtherSystemTables

phoenix-core/src/main/java/org/apache/phoenix/query/QueryConstants.java
twdsilva: For clusters that use namespace mapping and that map the phoenix system tables to the SYSTEM namespace, we want to be able to use the GRANT/REVOKE statements to grant RW access to the SYSTEM:MUTEX table.

twdsilva: Are you suggesting writeMutexCell/deleteMutexCell should take as arguments (tenantid, schema, tablename, column name, column family) instead of a byte[]? I will make that change.

ankitsinghal: ok make sense, so can you just update QueryServices.writeMutexCell(byte[] rowKey) and deleteMutexCell(byte[] rowKey) to accept all the arguments which form the right primary key for the table for consistency.

phoenix-core/src/main/java/org/apache/phoenix/query/ConnectionQueryServicesImpl.java
ankitsinghal: ok, then we are good.

phoenix-core/src/main/java/org/apache/phoenix/query/QueryConstants.java
ankitsinghal: Yes, probably our comments have crossed.

twdsilva:
Testing issues@phoenix.apache.org.

twdsilva:
@ankitsinghal  @karanmehta93  I updated the PR with the latest feedback, can you please review?

phoenix-core/src/main/java/org/apache/phoenix/schema/MetaDataClient.java
```diff
@@ -2957,6 +3006,16 @@ MutationState dropTable(String schemaName, String tableName, String parentTableN
...
+                if (!acquiredMutex) {
+                    logger.info("Failed to acquire mutex on physical table " + physicalTableName);
```
karanmehta93: nit: You can push mutex related logs lines to the specific methods.

twdsilva: Done, can I get a +1?

PR #314, opened 2018-07-26
PHOENIX-4820

comnetwork:
PHOENIX-4820

### DIFF
--- a/phoenix-core/src/main/java/org/apache/phoenix/compile/QueryCompiler.java
+++ b/phoenix-core/src/main/java/org/apache/phoenix/compile/QueryCompiler.java
@@ -559,8 +559,16 @@ public class QueryCompiler {
         groupBy = groupBy.compile(context, innerPlanTupleProjector);
         context.setResolver(resolver); // recover resolver
         RowProjector projector = ProjectionCompiler.compile(context, select, groupBy, asSubquery ? Collections.<PDatum>emptyList() : targetColumns, where);
-        OrderBy orderBy = OrderByCompiler.compile(context, select, groupBy, limit, offset, projector,
-                groupBy == GroupBy.EMPTY_GROUP_BY ? innerPlanTupleProjector : null, isInRowKeyOrder);
+        OrderBy orderBy = OrderByCompiler.compile(
+                context,
+                select,
+                groupBy,
+                limit,
+                offset,
+                projector,
+                groupBy == GroupBy.EMPTY_GROUP_BY ? innerPlanTupleProjector : null,
+                groupBy == GroupBy.EMPTY_GROUP_BY ? isInRowKeyOrder : true,
+                where);
         context.getAggregationManager().compile(context, groupBy);
         // Final step is to build the query plan
         if (!asSubquery) {
--- a/phoenix-core/src/main/java/org/apache/phoenix/expression/Expression.java
+++ b/phoenix-core/src/main/java/org/apache/phoenix/expression/Expression.java
@@ -88,4 +88,10 @@ public interface Expression extends PDatum, Writable {
      * @return
      */
     boolean requiresFinalEvaluation();
+
+    /**
+     *
+     * @return
+     */
+    boolean isConstantIfChildrenAllConstant();
 }
--- a/phoenix-core/src/main/java/org/apache/phoenix/util/ExpressionUtil.java
+++ b/phoenix-core/src/main/java/org/apache/phoenix/util/ExpressionUtil.java
@@ -10,12 +10,18 @@
 package org.apache.phoenix.util;
 
 import java.sql.SQLException;
+import java.util.Iterator;
 import java.util.List;
 
+import org.apache.hadoop.hbase.filter.CompareFilter.CompareOp;
 import org.apache.hadoop.hbase.io.ImmutableBytesWritable;
+import org.apache.phoenix.expression.AndExpression;
+import org.apache.phoenix.expression.ComparisonExpression;
 import org.apache.phoenix.expression.Determinism;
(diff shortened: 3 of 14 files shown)


phoenix-core/src/main/java/org/apache/phoenix/expression/Expression.java
```diff
@@ -88,4 +88,10 @@
      * @return
      */
     boolean requiresFinalEvaluation();
+
+    /**
+     *
+     * @return
+     */
+    boolean isConstantIfChildrenAllConstant();
```
JamesRTaylor: It's not clear that we need this. We already rollup determinism and isStateless for the expression tree. If determinism == Determinism.PER_STATEMENT or Determinism.ALWAYS and isStateless is true, then we know an expression is a constant. We have a utility for this in ExpressionUtil.isConstant(Expression).

phoenix-core/src/main/java/org/apache/phoenix/util/ExpressionUtil.java
```diff
@@ -68,4 +74,85 @@ public static boolean isPkPositionChanging(TableRef tableRef, List<Expression> p
...
+        if(whereExpression == null) {
+            return false;
+        }
+        IsColumnConstantExpressionVisitor isColumnConstantExpressionVisitor =
+                new IsColumnConstantExpressionVisitor(columnExpression);
+        whereExpression.accept(isColumnConstantExpressionVisitor);
+        return isColumnConstantExpressionVisitor.isConstant();
+    }
+
+    private static class IsColumnConstantExpressionVisitor extends StatelessTraverseNoExpressionVisitor<Void> {
```
JamesRTaylor: I'm hoping we don't need a new visitor here. There are probably other cases to check for besides ComparisonExpression. For example, InListExpression, maybe CoerceExpression, etc.

phoenix-core/src/main/java/org/apache/phoenix/compile/QueryCompiler.java
```diff
@@ -559,8 +559,16 @@ protected QueryPlan compileSingleFlatQuery(StatementContext context, SelectState
...
+        OrderBy orderBy = OrderByCompiler.compile(
+                context,
+                select,
+                groupBy,
+                limit,
+                offset,
+                projector,
+                groupBy == GroupBy.EMPTY_GROUP_BY ? innerPlanTupleProjector : null,
+                groupBy == GroupBy.EMPTY_GROUP_BY ? isInRowKeyOrder : true,
+                where);
```
JamesRTaylor: Would be ideal if the change could be isolated to OrderByCompiler being aware of its inner plan to determine correctly whether the OrderBy can be compiled out. Any ideas, @maryannxue?

phoenix-core/src/main/java/org/apache/phoenix/expression/Expression.java
comnetwork: I think ExpressionUtil.isConstant(Expression) is not suitable for this case, just as the comments of jira said, what we want to check in this issue is if a expression is constant when all children of it are constants, just consider following sql:

 select a.ak3  from
order by a.ak3,a.av1

Obviously , because of rand(), the Determinism of expression a.ak1 is Determinism.PER_INVOCATION, so the determinism is  Determinism.PER_INVOCATION and isStateless is false for expression "CASE WHEN coalesce(a.ak1,1) > coalesce(a.av2,2) THEN coalesce(a.ak1,1) ELSE coalesce(a.av2,2) END", but because the a.ak1 and a.av2 are both constants in where clause of outer query, we can regard  "CASE WHEN coalesce(a.ak1,1) > coalesce(a.av2,2) THEN coalesce(a.ak1,1) ELSE coalesce(a.av2,2) END" as constant in IsConstantVisitor.

phoenix-core/src/main/java/org/apache/phoenix/util/ExpressionUtil.java
comnetwork: I did not find a existing visitor  which can statisfy the requirement , I would check more Expression besides ComparisonExpression.

phoenix-core/src/main/java/org/apache/phoenix/compile/QueryCompiler.java
comnetwork: Yes, it is a nice suggestion, I would make a modification

maryannxue: Eliminating order-by based on the inner plan ordering is definitely the right and ultimate solution, and we should get rid of all “dirty fix” flags here. The compile result, in our case, the QueryPlan should have a definite ordering whether it’s from an order by, or a group-by, or a natural row order, or an order-by from the inner query. However, letting QueryPlan contain that information might not be a good idea. So we should see if we can infer the ordering from a QueryPlan.

comnetwork: @maryannxue , yes, you are right , Eliminating order-by based on the inner plan ordering is ultimate solution.  Openning this JIRA is just to optimize the OrderBy for ClientAggregatePlan when there is a GroupBy, it is only to need  to conside the GroupBy of the outer query, and innerQuery is not required to take into account. It is much simpler than purely optimizing OrderBy for ClientScanPlan, which need to  make a huge modification and refactor with OrderByCompiler and OrderPreservingTracker. I have planned to  purely optimize OrderBy for ClientScanPlan after this JIRA by openning a new JIRA.

gjacoby126:

PR #315, opened 2018-07-27
PHOENIX-3655 Global Phoenix Client Metrics for PQS

karanmehta93:
@joshelser Please review.

### DIFF
--- a/phoenix-core/src/it/java/org/apache/phoenix/monitoring/GlobalPhoenixMetricsTestSink.java
+++ b/phoenix-core/src/it/java/org/apache/phoenix/monitoring/GlobalPhoenixMetricsTestSink.java
@@ -1,0 +1,60 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.phoenix.monitoring;
+
+import org.apache.commons.configuration.SubsetConfiguration;
+import org.apache.commons.logging.Log;
+import org.apache.commons.logging.LogFactory;
+import org.apache.hadoop.metrics2.AbstractMetric;
+import org.apache.hadoop.metrics2.MetricsRecord;
+import org.apache.hadoop.metrics2.MetricsSink;
+import org.apache.phoenix.util.PhoenixRuntime;
+
+import java.util.Map;
+
+import static junit.framework.TestCase.assertTrue;
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.fail;
+
+public class GlobalPhoenixMetricsTestSink implements MetricsSink {
+
+    public static final String PHOENIX_METRICS_RECORD_NAME = "PHOENIX";
+    // PhoenixMetricsIT tests verifies these metrics from this sink in a separate thread
+    // GlobalPhoenixMetricsTestSink is invoked based on time defined in hadoop-metrics2.properties
+    // This lock is to prevent concurrent access to metrics Iterable for these threads
+    static Object lock = new Object();
+    static Iterable<AbstractMetric> metrics;
+
+    @Override
+    public void putMetrics(MetricsRecord metricsRecord) {
+        if (metricsRecord.name().equals(PHOENIX_METRICS_RECORD_NAME)) {
+            synchronized (GlobalPhoenixMetricsTestSink.lock) {
+                GlobalPhoenixMetricsTestSink.metrics = metricsRecord.metrics();
+            }
+        }
+    }
+
+    @Override
+    public void flush() {
+    }
+
+    @Override
+    public void init(SubsetConfiguration subsetConfiguration) {
+    }
+
+}
--- a/phoenix-core/src/it/java/org/apache/phoenix/monitoring/PhoenixMetricsIT.java
+++ b/phoenix-core/src/it/java/org/apache/phoenix/monitoring/PhoenixMetricsIT.java
@@ -40,6 +40,7 @@ import static org.apache.phoenix.monitoring.MetricType.TASK_EXECUTION_TIME;
 import static org.apache.phoenix.util.PhoenixRuntime.TENANT_ID_ATTRIB;
 import static org.apache.phoenix.util.PhoenixRuntime.UPSERT_BATCH_SIZE_ATTRIB;
 import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertThat;
 import static org.junit.Assert.assertTrue;
 import static org.junit.Assert.fail;
 
@@ -50,6 +51,7 @@ import java.sql.ResultSet;
 import java.sql.SQLException;
 import java.sql.Statement;
 import java.util.ArrayList;
+import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
@@ -61,6 +63,9 @@ import java.util.concurrent.Executors;
 import java.util.concurrent.Future;
 import java.util.concurrent.atomic.AtomicInteger;
 
+import org.apache.commons.logging.Log;
+import org.apache.commons.logging.LogFactory;
+import org.apache.hadoop.metrics2.AbstractMetric;
 import org.apache.phoenix.compile.StatementContext;
 import org.apache.phoenix.end2end.BaseUniqueNamesOwnClusterIT;
 import org.apache.phoenix.exception.SQLExceptionCode;
@@ -73,6 +78,7 @@ import org.apache.phoenix.log.LogLevel;
 import org.apache.phoenix.query.QueryServices;
 import org.apache.phoenix.util.PhoenixRuntime;
 import org.apache.phoenix.util.ReadOnlyProps;
+import org.hamcrest.CoreMatchers;
 import org.junit.BeforeClass;
 import org.junit.Test;
 import org.mockito.internal.util.reflection.Whitebox;
@@ -83,7 +89,17 @@ import com.google.common.collect.Lists;
 import com.google.common.collect.Maps;
 import com.google.common.collect.Sets;
 
+/**
+ * Tests that
+ * 1. Phoenix Global metrics are exposed via
+ *   a. PhoenixRuntime b. Hadoop-Metrics2 defined sinks
+ * 2. Phoenix Request level metrics are exposed via
+ *   a. PhoenixRuntime
+ */
 public class PhoenixMetricsIT extends BaseUniqueNamesOwnClusterIT {
+
+    private static final Log LOG = LogFactory.getLog(PhoenixMetricsIT.class);
+    private static final int MAX_RETRIES = 5;
 
     private static final List<MetricType> mutationMetricsToSkip =
             Lists.newArrayList(MetricType.MUTATION_COMMIT_TIME);
@@ -97,6 +113,7 @@ public class PhoenixMetricsIT extends BaseUniqueNamesOwnClusterIT {
     @BeforeClass
     public static void doSetup() throws Exception {
         Map<String, String> props = Maps.newHashMapWithExpectedSize(1);
+        // Phoenix Global client metrics are enabled by default
         // Enable request metric collection at the driver level
         props.put(QueryServices.COLLECT_REQUEST_LEVEL_METRICS, String.valueOf(true));
         // disable renewing leases as this will force spooling to happen.
@@ -104,15 +121,18 @@ public class PhoenixMetricsIT extends BaseUniqueNamesOwnClusterIT {
         setUpTestDriver(new ReadOnlyProps(props.entrySet().iterator()));
         // need the non-test driver for some tests that check number of hconnections, etc.
         DriverManager.registerDriver(PhoenixDriver.INSTANCE);
+
     }
 
     @Test
-    public void testResetGlobalPhoenixMetrics() {
+    public void testResetGlobalPhoenixMetrics() throws Exception {
         resetGlobalMetrics();
         for (GlobalMetric m : PhoenixRuntime.getGlobalPhoenixClientMetrics()) {
+            assertThat(m, CoreMatchers.<GlobalMetric>instanceOf(GlobalMetricImpl.class));
             assertEquals(0, m.getTotalSum());
             assertEquals(0, m.getNumberOfSamples());
         }
+        assertTrue(verifyMetricsFromSink());
     }
 
     @Test
@@ -146,6 +166,8 @@ public class PhoenixMetricsIT extends BaseUniqueNamesOwnClusterIT {
         assertTrue(GLOBAL_HBASE_COUNT_MILLS_BETWEEN_NEXTS.getMetric().getTotalSum() > 0);
         assertTrue(GLOBAL_HBASE_COUNT_BYTES_REGION_SERVER_RESULTS.getMetric().getTotalSum() > 0);
         assertTrue(GLOBAL_HBASE_COUNT_SCANNED_REGIONS.getMetric().getTotalSum() > 0);
+
+        assertTrue(verifyMetricsFromSink());
     }
 
     @Test
@@ -163,6 +185,8 @@ public class PhoenixMetricsIT extends BaseUniqueNamesOwnClusterIT {
         assertEquals(0, GLOBAL_FAILED_QUERY_COUNTER.getMetric().getTotalSum());
         assertEquals(0, GLOBAL_SPOOL_FILE_COUNTER.getMetric().getTotalSum());
         assertEquals(0, GLOBAL_MUTATION_BATCH_FAILED_COUNT.getMetric().getTotalSum());
+
+        assertTrue(verifyMetricsFromSink());
     }
 
     @Test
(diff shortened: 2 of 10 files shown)


karanmehta93:
The patch might be tricky to port to other branches. Will also look into that once the initial one gets in.

phoenix-core/src/it/java/org/apache/phoenix/monitoring/PhoenixMetricsIT.java
```diff
@@ -204,6 +219,39 @@ private static void resetGlobalMetrics() {
...
+                LOG.info("Values from Hadoop Metrics Sink match actual values");
+                return true;
+            }
+            Thread.sleep(1000);
+        }
+        return false;
+    }
+
+    private boolean verifyMetricsFromSinkOnce(Map<String, Long> expectedMetrics) {
+        synchronized (GlobalPhoenixMetricsTestSink.lock) {
```
joshelser: If it were me, I would have used a Queue or something to save off the metrics, but there's nothing wrong with your approach here :)

phoenix-core/src/it/java/org/apache/phoenix/monitoring/PhoenixMetricsIT.java
```diff
@@ -204,6 +219,39 @@ private static void resetGlobalMetrics() {
...
+            for (AbstractMetric metric : GlobalPhoenixMetricsTestSink.metrics) {
+                if (expectedMetrics.containsKey(metric.name())) {
```
joshelser: This check doesn't catch the case where you have `expectedMetrics` but `GlobalPhoenixMetricsTestSink.metrics` is empty. I'd switch this around to iterate over your expectations, ensuring that they all exist.

phoenix-core/src/test/resources/hadoop-metrics2.properties
```diff
@@ -32,10 +32,9 @@
 #    [prefix].[source|sink].[instance].[options]
 # See javadoc of package-info.java for org.apache.hadoop.metrics2 for detail
 
-
-# Don't attempt to start jmx mbeans for all sources.
-#  For right now, all metrics are exported to a phoenix table
-*.source.start_mbeans=false
+hbase.source.start_mbeans=true
+hbase.sink.sink0.class=org.apache.phoenix.monitoring.GlobalPhoenixMetricsTestSink
```
joshelser: Can you leave a comment in the test that this configuration is here? Otherwise, might seem like voodoo

phoenix-core/src/main/java/org/apache/phoenix/monitoring/GlobalClientMetrics.java
```diff
@@ -108,40 +114,88 @@
...
+    private static void registerPhoenixMetricsToRegistry(MetricRegistry metricRegistry) {
+        for (GlobalClientMetrics globalMetric : GlobalClientMetrics.values()) {
+            metricRegistry.register(globalMetric.metricType.columnName(),
+                    new PhoenixGlobalMetricGauge(globalMetric.metric));
+        }
+    }
+
+    private static MetricRegistry createMetricRegistry() {
+        LOG.info("Creating Metric Registry for Phoenix Global Metrics");
+        MetricRegistryInfo registryInfo = new MetricRegistryInfo("PHOENIX", "Phoenix Global Metrics",
```
joshelser: What does "Global" mean in this context? Is "Phoenix Client Metrics" more reasonable? Maybe "Phoenix,sub=CLIENT" down below?

phoenix-core/src/test/resources/hadoop-metrics2.properties
karanmehta93: Sure. It was hard for me as well since things were not working as expected in the first place. Took a little while to realize this file was the reason :)

phoenix-core/src/main/java/org/apache/phoenix/monitoring/GlobalClientMetrics.java
karanmehta93: Global here refers that it is an aggregation across all clients (or all Phoenix Connections). 

> Maybe "Phoenix,sub=CLIENT" down below?

Seems reasonable as well.

phoenix-core/src/it/java/org/apache/phoenix/monitoring/PhoenixMetricsIT.java
karanmehta93: I will also soon add a test that disables these metrics as well.
> I'd switch this around to iterate over your expectations, ensuring that they all exist.

The reason why I didn't do it that way because I didn't wanted to iterate over the it for every expected metric. It doesn't really matter since its a test though. One approach can be to remove each entry from the map and check if the size is 0 at the end. How does that sound?

phoenix-core/src/main/java/org/apache/phoenix/monitoring/GlobalClientMetrics.java
joshelser: > Global here refers that it is an aggregation across all clients (or all Phoenix Connections).

In my mind, if you show me "Phoenix Client Metrics", I would naturally assume that they are for all clients that have metrics enabled (as opposed to breakouts by individual clients). As such, the word "Global" to me is just filler, but maybe that's just my interpretation of it.

phoenix-core/src/test/resources/hadoop-metrics2.properties
joshelser: Haha, perfect :)

phoenix-core/src/it/java/org/apache/phoenix/monitoring/PhoenixMetricsIT.java
joshelser: > One approach can be to remove each entry from the map and check if the size is 0 at the end. How does that sound?

That would work too!

karanmehta93:
Addressed your comments @joshelser, Please review!

phoenix-core/src/it/java/org/apache/phoenix/monitoring/PhoenixMetricsIT.java
```diff
@@ -204,6 +230,45 @@ private static void resetGlobalMetrics() {
...
+            expectedMetrics.put(m.getMetricType().name(), m.getTotalSum());
+        }
+
+        for (int i = 0; i < MAX_RETRIES; i++) {
+            LOG.info("Verifying Global Metrics from Hadoop Sink, Retry: " + (i + 1));
+            if (verifyMetricsFromSinkOnce(expectedMetrics)) {
+                LOG.info("Values from Hadoop Metrics Sink match actual values");
+                return true;
+            }
+            Thread.sleep(1000);
```
joshelser: Change this to:
```
try {
  Thread.sleep(1000);
} catch (InterruptedException e) {
  Thread.currentThread().interrupt();
   return false;
}
```
Then, remove the `throws InterruptedException`.
We should be catching, re-setting the interrupted state and just returning ASAP. Likely, this code never gets invoked in a unit-test context, but good practice.

phoenix-core/src/it/java/org/apache/phoenix/monitoring/PhoenixMetricsIT.java
```diff
@@ -204,6 +230,45 @@ private static void resetGlobalMetrics() {
...
+        }
+        return false;
+    }
+
+    private boolean verifyMetricsFromSinkOnce(Map<String, Long> expectedMetrics) {
+        synchronized (GlobalPhoenixMetricsTestSink.lock) {
+            for (AbstractMetric metric : GlobalPhoenixMetricsTestSink.metrics) {
+                if (expectedMetrics.containsKey(metric.name())) {
```
joshelser: I thought there was an issue with unboxing the value from `expectedMetrics.get(metric.name())`.

I think it would be cleaner to do:
```
Long value = expectedMetrics.get(metric.name());
if (value != null) {
  long expectedValue = value.longValue();
  ...
```

phoenix-core/src/main/java/org/apache/phoenix/monitoring/GlobalMetricRegistriesAdapter.java
```diff
@@ -0,0 +1,167 @@
...
+ * The adapter attaches HBase's MetricRegistry to Hadoop's DefaultMetricsSystem
+ * Doesn't handle dynamic attach/detach of registries
+ */
+public class GlobalMetricRegistriesAdapter {
+
+    private static final Log LOG = LogFactory.getLog(GlobalMetricRegistriesAdapter.class);
+    private static GlobalMetricRegistriesAdapter INSTANCE;
+
+    private GlobalMetricRegistriesAdapter() {
+        DefaultMetricsSystem.initialize("HBase");
```
joshelser: s/HBase/Phoenix/

phoenix-core/src/main/java/org/apache/phoenix/monitoring/GlobalMetricRegistriesAdapter.java
```diff
@@ -0,0 +1,167 @@
...
+        JvmMetrics.initSingleton("HBase", "");
```
joshelser: I think collecting JVM metrics from client applications is heavy-handed. I'd suggest you drop this.

phoenix-core/src/it/java/org/apache/phoenix/monitoring/PhoenixMetricsIT.java
```diff
@@ -204,6 +230,45 @@ private static void resetGlobalMetrics() {
         }
     }
 
+    // Phoenix Client Metrics are transported via Hadoop-metrics2 sink
+    // The test sink is defined at GlobalPhoenixMetricsTestSink
+    // Configuration for Hadoop-metrics2 comes from hadoop-metrics2.properties file located in test/resources
```
joshelser: Nice!

phoenix-core/src/it/java/org/apache/phoenix/monitoring/GlobalPhoenixMetricsTestSink.java
```diff
@@ -0,0 +1,57 @@
...
+import java.util.Map;
+
+import static junit.framework.TestCase.assertTrue;
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.fail;
+
+public class GlobalPhoenixMetricsTestSink implements MetricsSink {
+
+    public static final String PHOENIX_METRICS_RECORD_NAME = "PHOENIX";
+    static Object lock = new Object();
```
joshelser: Can you leave a comment here about the importance of this lock? e.g. that the caller should hold it to read `metrics`.

phoenix-core/src/main/java/org/apache/phoenix/monitoring/GlobalMetricRegistriesAdapter.java
karanmehta93: Actually I added this since this is going to be a part of PQS, where having these metrics would help. Should I cover this with a config parameter then?

karanmehta93:
@joshelser I was trying out this patch with an internal PQS version. All these metrics are pushed to JMX and we monitor those via OpenTSDB tcollectors. I haven't done any perf-test here, but I can create a "rate of" on the metrics to track the increase in values over time. Still working on some more metrics patches internally before I can see this end to end. From @twdsilva and @apurtell experience, seems like these metrics should help monitor the health and usage of PQS.

phoenix-core/src/main/java/org/apache/phoenix/monitoring/GlobalMetricRegistriesAdapter.java
joshelser: Oh, right. I keep getting confused over that.

JVM metrics being exported from PQS makes total sense. +1 to that

I am now wondering: should we be exporting Phoenix metrics out of PQS under "Client" or make some specific "sub=PQS" or similar for it?

joshelser:
> All these metrics are pushed to JMX and we monitor those via OpenTSDB tcollectors.

Cool! I was just looking for a sanity check that you are seeing all of the metrics you expect coming out of PQS and that they are "useful". No need to perform any perf testing for me.

karanmehta93:
@joshelser I think the only question pending in this PR is if we want JVMMetrics or not, which are crucial from PQS POV, otherwise not really for a regular client. We can discuss what approach to finalize and then I can commit this PR. Wdyt?

joshelser:
> I think the only question pending in this PR is if we want JVMMetrics or not, which are crucial from PQS POV, otherwise not really for a regular client.

Agreed. What about just making a new configuration key in QueryServices/QueryServicesOptions for including JVM metrics in client-metrics which defaults to `false`? Then, have PQS always set this to be `true` via code.

Going to pull this down right now and look at it too out of curiosity.

joshelser:
Pulling this into `master`, it looks like there are some commons-configuration issues that will need to be worked out for GlobalPhoenixMetricsTestSink.java

joshelser:
> it looks like there are some commons-configuration issues that will need to be worked out for GlobalPhoenixMetricsTestSink.java

Just needs to be switched to the `org.apache.commons.configuration2` package. Also, GlobalPhoenixMetricsTestSink has a lot of unused imports. Would be good to make a pass over your changes for those.

karanmehta93:
Thanks @joshelser will update the PR soon. We have immediate requirements for HBase-1.x versions here. That is why I mentioned that at the start of PR [itself](https://issues.apache.org/jira/browse/PHOENIX-3655?page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel&focusedCommentId=16559438#comment-16559438) that it might be tricky to port these patches. You can try it out with 4.x-HBase-1.4 branch, which is where my current PR is based on. Let me know if you think anything else is required here.

phoenix-core/src/main/java/org/apache/phoenix/monitoring/GlobalMetricRegistriesAdapter.java
karanmehta93: Lets keep it as client since they represent client in the end, even though it PQS that is instantiating it.

joshelser: How will you differentiate this "client" (PQS) from other clients?

karanmehta93: I think its hard to differentiate, but my concern is why do we need to differentiate. We should get that info from the application JVM itself right, whether its PQS or some app server using Phoenix Client.

joshelser: > why do we need to differentiate

As an operator, I'm going to expect much different characteristics coming from client applications than I am from PQS. Since PQS is a SPOF for (potentially) many users, I would expect a higher level of regularity from PQS than I would a client's app.

> We should get that info from the application JVM itself right

I'm sure there's something you can apply to find a PQS metric compared to other clients, but I am just thinking that it would be good to make this as easy as possible. I think watching PQS metrics would be a common use-case, not the exception.

joshelser:
Just used JVisualVM to look at these metrics via the MBeans directly. Worked just fine.

I think naming/convention when being collected from a distributed environment needs to be thought about more though.

phoenix-core/src/main/java/org/apache/phoenix/monitoring/GlobalMetricRegistriesAdapter.java
joshelser: > I am sorry, I didn't you on this one. Could you elaborate?

You're using openTSDB to collect all of this data and your use case is to "understand how PQS runs". To do this, you will need to know what hosts that PQS is running on to just look at PQS metrics. I was suggesting that having a unique name (or a tag) would make an operator's life much more simple -- it would be clear (and presumably easier to filter on) to find just PQS metrics.

For example, the difference of issuing a filter "host in pqs_server1, pqs_server2, pqs_server3" as opposed to just "tag=PQS", or similar. 

Now that I'm thinking about this, leaving this as {{Phoenix,sub=CLIENT}} may be OK and you can instead just configure the push of metrics data to Hadoop Metrics2 to include a "tag" that indicates this is from PQS. That would make me happy.

karanmehta93: So if I understand this correctly, I will include a new property which takes a csv of tags as input. Later on, I can split by comma and add those tags to `HBaseMetrics2HadoopMetricsAdapter#snapshotAllMetrics()`. @joshelser

phoenix-core/src/main/java/org/apache/phoenix/monitoring/GlobalMetricRegistriesAdapter.java
```diff
@@ -0,0 +1,167 @@
...
+    private static final Log LOG = LogFactory.getLog(GlobalMetricRegistriesAdapter.class);
+    private static GlobalMetricRegistriesAdapter INSTANCE;
+
+    private GlobalMetricRegistriesAdapter() {
+        DefaultMetricsSystem.initialize("Phoenix");
+        JvmMetrics.initSingleton("Phoenix", "");
+    }
+
+    public static GlobalMetricRegistriesAdapter getInstance() {
+        if (INSTANCE == null) {
```
xcangCRM: nit: do you need double checking here?
if (INSTANCE == null) {
     synchronized(this) {
        if(INSTANCE == null) {
              ...
  }  
 }
}

phoenix-core/src/main/java/org/apache/phoenix/monitoring/GlobalMetricRegistriesAdapter.java
```diff
@@ -0,0 +1,167 @@
...
+        public void snapshotAllMetrics(MetricRegistry metricRegistry, MetricsCollector collector) {
+            MetricRegistryInfo info = metricRegistry.getMetricRegistryInfo();
+            MetricsRecordBuilder builder = collector.addRecord(Interns.info(info.getMetricsName(), info.getMetricsDescription()));
+            builder.setContext(info.getMetricsContext());
+            this.snapshotAllMetrics(metricRegistry, builder);
+        }
+
+        public void snapshotAllMetrics(MetricRegistry metricRegistry, MetricsRecordBuilder builder) {
+            Map<String, Metric> metrics = metricRegistry.getMetrics();
+            Iterator i$ = metrics.entrySet().iterator();
```
xcangCRM: why do you need this?  i$

phoenix-core/src/test/resources/hadoop-metrics2.properties
```diff
@@ -32,10 +32,9 @@
 #    [prefix].[source|sink].[instance].[options]
 # See javadoc of package-info.java for org.apache.hadoop.metrics2 for detail
 
-
-# Don't attempt to start jmx mbeans for all sources.
-#  For right now, all metrics are exported to a phoenix table
-*.source.start_mbeans=false
+phoenix.source.start_mbeans=true
+phoenix.sink.sink0.class=org.apache.phoenix.monitoring.GlobalPhoenixMetricsTestSink
```
xcangCRM: I think it usually starts with sink 1 .  (?)  at least ni HBase.

karanmehta93: It is just an identifier and can be anything.

phoenix-core/src/main/java/org/apache/phoenix/monitoring/GlobalMetricRegistriesAdapter.java
karanmehta93: Cool. Thanks!

phoenix-core/src/main/java/org/apache/phoenix/monitoring/GlobalMetricRegistriesAdapter.java
karanmehta93: Yes, ideally should not have used it here. I copied the code from a class file. Will update it.

phoenix-core/src/main/java/org/apache/phoenix/monitoring/GlobalMetricRegistriesAdapter.java
karanmehta93: As of now, this method is called from static block in `GlobalClientMetrics`, which can happen only once, so we dont have any race conditions here. 
I have two options here, add a lock here or move this initialization to static block of this class. What do you suggest?

karanmehta93: I think I am moving away from lazy initialization all together. No need of premature optimization here I believe.

joshelser:
One final request, @karanmehta93  -- can you set `phoenix.client.metrics.tag` in QueryServer.java proactively, please?

You have my +1 after that.

karanmehta93:
@joshelser I don't see an easy way of doing it, since we read the tag from `metricTag = QueryServicesOptions.withDefaults().getClientMetricTag();`, which internally uses `InstanceResolver`, which depends on implementations of `ConfigurationFactory`. Ideally, I need to add META-INF/<file> to provide that factory and make it work. I don't think that we want to enforce that. If there is any other option, let me know.

joshelser:
@karanmehta93 but in PQS, we're controlling the Configuration. Isn't this as simple as changing `QueryServer#main(String[])` to use `HBaseFactoryProvider.getConfigurationFactory().getConfiguration();` instead of `HBaseConfiguration.create()`?

joshelser:
> Am I missing something here?

Nope, it's me :). I didn't realize that the Factory was creating a new `Configuration` every time. The setup of `GlobalClientMetrics` just... doesn't give us any way to do this nicely :(

I'm not sure what it takes to write a custom Factory (I think I did it once in an IT), but would be OK if you defer this.

After you commit this, please update the website with these necessary changes (e.g. at least tell people that PQS exports metrics).

karanmehta93:
> The setup of GlobalClientMetrics just... doesn't give us any way to do this nicely
Yes, agreed on that one.

> I'm not sure what it takes to write a custom Factory (I think I did it once in an IT), but would be OK if you defer this
Checkout `PhoenixMetricsDisabledIT` in this patch. I am doing the same.

> After you commit this, please update the website with these necessary changes (e.g. at least tell people that PQS exports metrics).
Sure will file a new Jira for documentation change as well.

One last question, Is it fine for me to change the `QueryServer#main()` to use `HBaseFactoryProvider.getConfigurationFactory().getConfiguration()` instead of the regular one. We can control the config from other projects, if required. What do you think?

joshelser:
> Is it fine for me to change

I'd just leave this as-is for now. I don't think it's "broken" how it is right now (PQS doesn't try to set anything in the Configuration for the PhoenixDriver to pick up).

karanmehta93:
> I'd just leave this as-is for now. I don't think it's "broken" how it is right now (PQS doesn't try to set anything in the Configuration for the PhoenixDriver to pick up).

It's just that some project using PQS wants to change configuration via code, It wont be possible since `HBaseConfiguration.create()` will only read XML on the class path. Either way, I am fine. Let me know your thoughts.

karanmehta93:
Great! That makes sense. 
Thanks for the review @joshelser . I will be (squashing and) pushing this to 4.x-HBase-1.4 branch and will back port / forward port this patch as well.

PR #316, opened 2018-07-27

m2je:

Currently the number of indices per Phoenix table is bound to maximum of 65535 (java.lang.Short) which is a limitation for applications requiring to have unlimited number of indices.
This change will consider any new table created in Phoenix to support view index ids to be in the range of -9,223,372,036,854,775,808 to 9,223,372,036,854,775,807 (java.lang.Long) which is undoubtably big enough to cover this requirement.
Any existing Phoenix table will still continue to support only maximum of 65535 of indices.
A new boolean column (USE_LONG_VIEW_INDEX BOOLEAN DEFAULT FALSE) is added to SYSTEM.CATALOG to specify each Phoenix table's support for large number of indices.
On each new Phoenix table creation the value for USE_LONG_VIEW_INDEX will be set to `true` while this value would be false for any existing table.

### DIFF
--- a/bin/hbase-site.xml
+++ b/bin/hbase-site.xml
@@ -24,8 +24,4 @@
     <name>hbase.regionserver.wal.codec</name>
     <value>org.apache.hadoop.hbase.regionserver.wal.IndexedWALEditCodec</value>
   </property>
-  <property>
-    <name>phoenix.log.level</name>
-    <value>DEBUG</value>
-  </property>
 </configuration>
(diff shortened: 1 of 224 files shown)


twdsilva:
@m2je  Thanks for the PR. Can you please update the PR so that it uses the correct branch, so that we can see only your changes? This PR currently has 250+ commits.

twdsilva:
@m2je  Thanks for the contribution. Can you please update the PR to use the correct branch. It currently has 250+ commits, so it hard to review your changes.

m2je:
I don't know what happened. Let me open a new pull request

PR #322, opened 2018-08-07
PHOENIX-4834 PhoenixMetricsLog interface methods should not depend on…

karanmehta93:
… specific logger


twdsilva:
+1

joshelser:
Ditto, Shipit!

PR #323, opened 2018-08-08
PHOENIX-4839 Check for memstore scanners if it is null before adding …

mnpoonia:
…it to scanner list

### DIFF
--- a/phoenix-core/src/main/java/org/apache/hadoop/hbase/regionserver/IndexHalfStoreFileReaderGenerator.java
+++ b/phoenix-core/src/main/java/org/apache/hadoop/hbase/regionserver/IndexHalfStoreFileReaderGenerator.java
@@ -395,7 +395,10 @@ public class IndexHalfStoreFileReaderGenerator extends BaseRegionObserver {
                         .getReader().getHFileReader().hasMVCCInfo(), readPt));
             }
         }
-        keyValueScanners.addAll(((HStore) store).memstore.getScanners(readPt));
+        List<KeyValueScanner> memstoreScanners = ((HStore) store).memstore.getScanners(readPt);
+        if(memstoreScanners != null) {
+            keyValueScanners.addAll(memstoreScanners);
+        }
         return keyValueScanners;
     }
 }
--- a/phoenix-core/src/main/java/org/apache/hadoop/hbase/regionserver/IndexHalfStoreFileReaderGenerator.java
+++ b/phoenix-core/src/main/java/org/apache/hadoop/hbase/regionserver/IndexHalfStoreFileReaderGenerator.java
@@ -395,9 +395,11 @@ public class IndexHalfStoreFileReaderGenerator extends BaseRegionObserver {
                         .getReader().getHFileReader().hasMVCCInfo(), readPt));
             }
         }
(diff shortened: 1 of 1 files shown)


phoenix-core/src/main/java/org/apache/hadoop/hbase/regionserver/IndexHalfStoreFileReaderGenerator.java
```diff
@@ -395,7 +395,10 @@ public KeyValueScanner preStoreScannerOpen(final ObserverContext<RegionCoprocess
                         .getReader().getHFileReader().hasMVCCInfo(), readPt));
             }
         }
-        keyValueScanners.addAll(((HStore) store).memstore.getScanners(readPt));
+        List<KeyValueScanner> memstoreScanners = ((HStore) store).memstore.getScanners(readPt);
+        if(memstoreScanners != null) {
```
joshelser: What is the case that this would ever be null?

phoenix-core/src/main/java/org/apache/hadoop/hbase/regionserver/IndexHalfStoreFileReaderGenerator.java
```diff
@@ -395,7 +395,10 @@ public KeyValueScanner preStoreScannerOpen(final ObserverContext<RegionCoprocess
                         .getReader().getHFileReader().hasMVCCInfo(), readPt));
             }
         }
-        keyValueScanners.addAll(((HStore) store).memstore.getScanners(readPt));
+        List<KeyValueScanner> memstoreScanners = ((HStore) store).memstore.getScanners(readPt);
```
joshelser: In general, we don't ever want to be casting the interface (`Store`) to the implementation (`HStore`). Is phoenix fundamentally doing something here that requires it to be tied to some specific HStore functionality?

phoenix-core/src/main/java/org/apache/hadoop/hbase/regionserver/IndexHalfStoreFileReaderGenerator.java
mnpoonia: This is the line where it might return null 
https://github.com/apache/hbase/blob/branch-1.3/hbase-server/src/main/java/org/apache/hadoop/hbase/regionserver/DefaultMemStore.java#L637

phoenix-core/src/main/java/org/apache/hadoop/hbase/regionserver/IndexHalfStoreFileReaderGenerator.java
mnpoonia: Actually i didn't want to touch the other part of code to fix this specific issue.

mnpoonia: @joshelser Looks like we need HStore to access memstore specifically. We can't access the memstore and get its scanners from the store object.

mnpoonia: @joshelser - Thanks for pointing that out(my bad i didn't get that in first look at the code ). Updated PR, please have a look.

PR #325, opened 2018-08-15
PHOENIX-4832 Add Canary Test Tool for Phoenix Query Server

ashuparekh:
A suggested improvement is to add a Canary Test tool to the Phoenix Query Server. It will execute a set of Basic Tests (CRUD) against a PQS end-point and report on the proper functioning and testing results.

@joshelser It would be great if you could review this.

@vincentpoon @karanmehta93


stoty:
Already merged.

PR #328, opened 2018-08-16
PHOENIX-4835 LoggingPhoenixConnection should log metrics upon connect…

karanmehta93:
…ion close

@ankitsinghal I did some refactoring in tests and added new tests for the scenario along with now logging metrics both on close and commit. Hence, a quick look at PR would be helpful.
There is one more small pending problem, metrics won't be logged when auto_commit is set to true. Will file a separate Jira for that.

FYI, Closed original PR since it had incorrect title.


ankitsinghal:
LGTM, @karanmehta93 +1

karanmehta93:
Thanks @ankitsinghal, will commit it tomorrow.

PR #329, opened 2018-08-17
PHOENIX-4798 Update encoded col qualifiers on the base table correctly

twdsilva:
@ChinmaySKulkarni  @karanmehta93 Can you please review?
This doc describes how the client side mutexes are used
https://docs.google.com/spreadsheets/d/1EOkuIdp4_OHg-l-ArdZYnAkMiCreDUIBjGXzdlJHW2w/edit?usp=sharing

### DIFF
--- a/phoenix-core/src/main/java/org/apache/phoenix/coprocessor/MetaDataEndpointImpl.java
+++ b/phoenix-core/src/main/java/org/apache/phoenix/coprocessor/MetaDataEndpointImpl.java
@@ -2208,8 +2208,7 @@ public class MetaDataEndpointImpl extends MetaDataProtocol implements Coprocesso
                 MetaDataResponse response =
                         processRemoteRegionMutations(
                             PhoenixDatabaseMetaData.SYSTEM_CHILD_LINK_NAME_BYTES,
-                            childLinkMutations, fullTableName,
-                            MetaDataProtos.MutationCode.UNABLE_TO_CREATE_CHILD_LINK);
+                            childLinkMutations, MetaDataProtos.MutationCode.UNABLE_TO_CREATE_CHILD_LINK);
                 if (response != null) {
                     done.run(response);
                     return;
@@ -2229,8 +2228,12 @@ public class MetaDataEndpointImpl extends MetaDataProtocol implements Coprocesso
                         response =
                                 processRemoteRegionMutations(
                                     PhoenixDatabaseMetaData.SYSTEM_CATALOG_NAME_BYTES,
-                                    remoteMutations, fullTableName,
-                                    MetaDataProtos.MutationCode.UNABLE_TO_UPDATE_PARENT_TABLE);
+                                    remoteMutations, MetaDataProtos.MutationCode.UNABLE_TO_UPDATE_PARENT_TABLE);
+                        clearParentTableFromCache(clientTimeStamp,
+                            parentTable.getSchemaName() != null
+                                    ? parentTable.getSchemaName().getBytes()
+                                    : ByteUtil.EMPTY_BYTE_ARRAY,
+                            parentTable.getName().getBytes());
                         if (response != null) {
                             done.run(response);
                             return;
@@ -2484,8 +2487,7 @@ public class MetaDataEndpointImpl extends MetaDataProtocol implements Coprocesso
                 MetaDataResponse response =
                         processRemoteRegionMutations(
                             PhoenixDatabaseMetaData.SYSTEM_CHILD_LINK_NAME_BYTES,
-                            childLinkMutations, SchemaUtil.getTableName(schemaName, tableName),
-                            MetaDataProtos.MutationCode.UNABLE_TO_CREATE_CHILD_LINK);
+                            childLinkMutations, MetaDataProtos.MutationCode.UNABLE_TO_CREATE_CHILD_LINK);
                 if (response!=null) {
                     done.run(response);
                     return;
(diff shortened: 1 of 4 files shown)


twdsilva:
@ankitsinghal , could you please review?

phoenix-core/src/main/java/org/apache/phoenix/util/MetaDataUtil.java
```diff
@@ -303,6 +303,54 @@ public static long getSequenceNumber(List<Mutation> tableMetaData) {
...
+                    }
+                }
+            }
+        }
+        if (parentTableRowKey == null) {
+            throw new IllegalStateException(
+                    "Could not find mutation to update the encoded column qualifier of the parent table ");
+        }
+        
+        byte[] rowKeyPrefix = ByteUtil.concat(parentTableRowKey, ByteUtil.EMPTY_BYTE_ARRAY);
```
ankitsinghal: wouldn't be rowKeyPrefix and parentTableRowkey equal here and so why two separate loops? is it a typo error for SEPARATOR_BYTE?

phoenix-core/src/main/java/org/apache/phoenix/schema/MetaDataClient.java
```diff
@@ -3681,7 +3691,12 @@ public MutationState addColumn(PTable table, List<ColumnDef> origColumnDefs,
             }
         } finally {
             connection.setAutoCommit(wasAutoCommit);
-            if (!acquiredColumnMutexSet.isEmpty()) {
+            if (EncodedColumnsUtil.usesEncodedColumnNames(table)) {
+                // release the mutex on the physical table (used to prevent concurrent conflicting
+                // add column changes)
+                deleteCell(null, physicalSchemaName, physicalTableName, null);
```
ankitsinghal: should we check here that we have acquired the mutex?

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/MetaDataEndpointImpl.java
```diff
@@ -3207,17 +3226,37 @@ public MetaDataMutationResult updateMutation(PTable table, byte[][] rowKeyMetaDa
...
+                                 * scheme, we need to modify the CQ counters stored in the view's
+                                 * physical table. So to make sure clients get the latest PTable, we
+                                 * need to invalidate the cache entry.
+                                 */
+                                invalidateList.add(new ImmutableBytesPtr(
+                                        MetaDataUtil.getPhysicalTableRowForView(table)));
+                                byte[] parentSchemaName =
+                                        SchemaUtil.getSchemaNameFromFullName(
+                                            table.getPhysicalName().getString()).getBytes();
+                                byte[] parentTableName =
```
ankitsinghal: PTable.getParentTableName and PTable.getParentSchemaName() can be used here.

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/MetaDataEndpointImpl.java
```diff
@@ -3207,17 +3226,37 @@ public MetaDataMutationResult updateMutation(PTable table, byte[][] rowKeyMetaDa
...
+                                        SchemaUtil
+                                                .getTableNameFromFullName(
+                                                    (table.getPhysicalName().getString()))
+                                                .getBytes();
+                                PTable parentTable =
+                                        doGetTable(null, parentSchemaName, parentTableName,
```
ankitsinghal: doGetTable expects table key to be in the same region, so if a view is created in different schema than the parent schema, will it work in case of mult-region SYSTEM.CATALOG table.

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/MetaDataEndpointImpl.java
```diff
@@ -2231,6 +2231,7 @@ public void createTable(RpcController controller, CreateTableRequest request,
                                     PhoenixDatabaseMetaData.SYSTEM_CATALOG_NAME_BYTES,
                                     remoteMutations, fullTableName,
                                     MetaDataProtos.MutationCode.UNABLE_TO_UPDATE_PARENT_TABLE);
+                        clearParentTableFromCache(clientTimeStamp, table);
```
ankitsinghal: Why explicitly we need to clear the cache, doesn't the server cache holds the current table always?

phoenix-core/src/main/java/org/apache/phoenix/util/MetaDataUtil.java
```diff
@@ -303,6 +303,54 @@ public static long getSequenceNumber(List<Mutation> tableMetaData) {
         return getSequenceNumber(getPutOnlyTableHeaderRow(tableMetaData));
     }
     
+    /**
+     * Returns the sequence number of the parent table if we have a mutation that is updating the
+     * parent table's encoded column qualifier
+     */
+    public static long getParentSequenceNumberForAddColumnMutations(List<Mutation> tableMetaData) {
+        byte[] parentTableRowKey = null;
+        for (Mutation tableMutation : tableMetaData) {
```
karanmehta93: I don't think we really need this.

twdsilva: Done. We don't need this because we acquire a mutex on the client side which will prevent concurrent changes to the encoded col qualifier counter.

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/MetaDataEndpointImpl.java
```diff
@@ -3291,6 +3305,29 @@ public MetaDataMutationResult updateMutation(PTable table, byte[][] rowKeyMetaDa
...
+                                        MetaDataUtil.getPhysicalTableRowForView(table)));
+                                byte[] parentSchemaName =
+                                        table.getParentSchemaName() != null
+                                                ? table.getParentSchemaName().getBytes()
+                                                : ByteUtil.EMPTY_BYTE_ARRAY;
+                                byte[] parentTableName = table.getParentTableName().getBytes();
+                                PTable parentTable =
+                                        doGetTable(null, parentSchemaName, parentTableName,
+                                            clientTimeStamp, null, request.getClientVersion(), true,
+                                            true, null);
```
ankitsinghal: You may need to remove doGetTable call as parentTable is not used anywhere now?

twdsilva:
@ankitsinghal  Thanks for the review, I made the change and pushed to 4.x and master branches.

PR #331, opened 2018-08-20
Add example for Spark 2.0.

danicat:
SQLContext was deprecated on Spark 2. I've fixed the first example to be compliant with the new syntax.


twdsilva:
@danicat  Thanks for the contribution, please file a JIRA for tracking purposes so that this PR can be committed.

PR #335, opened 2018-08-23
PHOENIX-4791 Propagate projected array elements through hash join

geraldss:


phoenix-core/src/it/java/org/apache/phoenix/end2end/ProjectArrayElemAfterHashJoinIT.java
```diff
@@ -0,0 +1,179 @@
...
+                assertEquals(5, java.lang.reflect.Array.getLength(obj));
+            }
+
+            assertFalse(rs.next());
+        } finally {
+            rs.close();
+        }
+    }
+
+    private void dropTable(Connection conn, String table) throws Exception {
```
twdsilva: No need to drop table since you are creating creating table with unique names.

geraldss: Thanks @twdsilva . I submitted the patch.

I left the dropTable() in there if it's harmless. Lmk if it's a problem.

twdsilva: The dropTable() just slows the test down. We prevent too many tables from being created by starting a new minicluster after a certain threshold (see .BaseTest.tearDownMiniClusterIfBeyondThreshold() )

geraldss: Hi @twdsilva @JamesRTaylor, done. Thanks for the review.

PR #338, opened 2018-08-30
PHOENIX-4870 LoggingPhoenixConnection should log metrics when AutoCommit is set to True.

swaroopak:


phoenix-core/src/it/java/org/apache/phoenix/monitoring/PhoenixLoggingMetricsIT.java
```diff
@@ -154,6 +154,35 @@ private void verifyQueryLevelMetricsLogging(String query) throws SQLException {
...
+        //with execute() method
+        loggedConn.createStatement().execute(upsertSelect);
+        // Autocommit is turned on explicitly
+        // Hence mutation metrics are expected during implicit commit
+        assertTrue("Mutation write metrics are not logged for " + tableName2,
+                mutationWriteMetricsMap.size()  > 0);
+        assertTrue("Mutation read metrics for not found for " + tableName1,
+                mutationReadMetricsMap.get(tableName1).size() > 0);
+
+        clearAllTestMetricMaps();
```
karanmehta93: no need to call this method unless you are checking any metrics after this.
This automatically gets called in `@Before`

phoenix-core/src/it/java/org/apache/phoenix/monitoring/PhoenixLoggingMetricsIT.java
```diff
@@ -154,6 +154,35 @@ private void verifyQueryLevelMetricsLogging(String query) throws SQLException {
...
+        //with executeUpdate() method
+        // run SELECT to verify read metrics are logged
+        String query = "SELECT * FROM " + tableName1;
+        verifyQueryLevelMetricsLogging(query);
+        // run UPSERT SELECT to verify mutation metrics are logged
+        String upsertSelect = "UPSERT INTO " + tableName2 + " SELECT * FROM " + tableName1;
+        loggedConn.createStatement().executeUpdate(upsertSelect);
+        // Autocommit is turned on explicitly
+        // Hence mutation metrics are expected during implicit commit
+        assertTrue("Mutation write metrics are not logged for " + tableName2,
```
karanmehta93: You should check specifically for `tableName2` in the map.

phoenix-core/src/main/java/org/apache/phoenix/jdbc/LoggingPhoenixStatement.java
```diff
@@ -59,4 +70,17 @@ public ResultSet getGeneratedKeys() throws SQLException {
         return new LoggingPhoenixResultSet(super.getGeneratedKeys(), phoenixMetricsLog, this.sql);
     }
 
+    private void loggingAutoCommitHelper() throws SQLException {
+        Connection conn = this.getConnection();
+
+        if(stmt.getUpdateOperation().isMutation() && conn.getAutoCommit()){
```
karanmehta93: Isn't `conn.getAutoCommit()` check sufficient?

phoenix-core/src/main/java/org/apache/phoenix/jdbc/LoggingPhoenixStatement.java
```diff
@@ -59,4 +70,17 @@ public ResultSet getGeneratedKeys() throws SQLException {
...
+            phoenixMetricsLog.logWriteMetricsfoForMutationsSinceLastReset(
+                    PhoenixRuntime.getWriteMetricInfoForMutationsSinceLastReset(conn));
+            phoenixMetricsLog.logReadMetricInfoForMutationsSinceLastReset(
+                    PhoenixRuntime.getReadMetricInfoForMutationsSinceLastReset(conn));
+            PhoenixRuntime.resetMetrics(conn);
+        }
+        return;
```
karanmehta93: nit: not needed

phoenix-core/src/main/java/org/apache/phoenix/jdbc/LoggingPhoenixStatement.java
```diff
@@ -59,4 +70,17 @@ public ResultSet getGeneratedKeys() throws SQLException {
         return new LoggingPhoenixResultSet(super.getGeneratedKeys(), phoenixMetricsLog, this.sql);
     }
 
+    private void loggingAutoCommitHelper() throws SQLException {
+        Connection conn = this.getConnection();
+
+        if(stmt.getUpdateOperation().isMutation() && conn.getAutoCommit()){
+            phoenixMetricsLog.logWriteMetricsfoForMutationsSinceLastReset(
```
karanmehta93: Can you reuse this piece of code from somewhere?

phoenix-core/src/main/java/org/apache/phoenix/jdbc/LoggingPhoenixStatement.java
karanmehta93: Follow up: Is `Statement stmt` reference needed?

swaroopak: This helper function gets called from both execute and executeUpdate. We need Statment reference to distinguish in case of execute

phoenix-core/src/main/java/org/apache/phoenix/jdbc/LoggingPhoenixStatement.java
swaroopak: Unlikely, but let me check that.

swaroopak: Looks like we can't.

karanmehta93: Can you check `LoggingPhoenixConnection`? It has similar calls multiple times as well. Might be good to refactor this along with it.

phoenix-core/src/main/java/org/apache/phoenix/jdbc/LoggingPhoenixStatement.java
karanmehta93: What happens if its not a mutation and we try logging the metrics (in case of queries)?

karanmehta93: I see this piece of code
```
                        // If transactional, this will move the read pointer forward
                        if (connection.getAutoCommit()) {
                            connection.commit();
                        }
```
@twdsilva any idea?

swaroopak: Yes, in case of executeQuery, metrics get logged. The differentiating function call happens in PhoenixStatement Class. In case of mutation, it directly executes executeMutation.

twdsilva: Not sure if that is needed any more since we start the transaction in MetaDataClient.updateCache

boolean isTransactional = (table!=null && table.isTransactional());
        if (isTransactional) {
            connection.getMutationState().startTransaction(table.getTransactionProvider());
        }

karanmehta93: > In case of mutation, it directly executes executeMutation

You can also provide a mutation through `execute()` method call.

phoenix-core/src/main/java/org/apache/phoenix/jdbc/LoggingPhoenixPreparedStatement.java
```diff
@@ -17,20 +17,20 @@
...
     private String sql;
+    private Connection conn;
     
-    public LoggingPhoenixPreparedStatement(PreparedStatement stmt, PhoenixMetricsLog phoenixMetricsLog, String sql) {
+    public LoggingPhoenixPreparedStatement(Connection conn, PreparedStatement stmt,
+                                           PhoenixMetricsLog phoenixMetricsLog, String sql) {
         super(stmt);
         this.phoenixMetricsLog = phoenixMetricsLog;
         this.sql = sql;
+        this.conn = conn;
```
karanmehta93: nit: Can you move the init of variables in the same order as function arguments?

phoenix-core/src/main/java/org/apache/phoenix/jdbc/LoggingPhoenixStatement.java
```diff
@@ -62,4 +75,10 @@ public ResultSet getGeneratedKeys() throws SQLException {
         return new LoggingPhoenixResultSet(super.getGeneratedKeys(), phoenixMetricsLog, this.sql);
     }
 
+    private void loggingAutoCommitHelper() throws SQLException {
+        if(conn instanceof LoggingPhoenixConnection && conn.getAutoCommit()){
```
karanmehta93: nit: 
just change the order of conditions
add brackets around `conn instanceof LoggingPhoenixConnection`
 and add spaces before brackets in conditions (applicable everywhere)

phoenix-core/src/main/java/org/apache/phoenix/jdbc/LoggingPhoenixStatement.java
```diff
@@ -25,28 +28,38 @@
...
     private PhoenixMetricsLog phoenixMetricsLog;
     private String sql;
-    
-    public LoggingPhoenixStatement(Statement stmt, PhoenixMetricsLog phoenixMetricsLog) {
+    private Connection conn;
+
+    public LoggingPhoenixStatement(Connection conn, Statement stmt, PhoenixMetricsLog phoenixMetricsLog) {
         super(stmt);
         this.phoenixMetricsLog = phoenixMetricsLog;
+        this.conn = conn;
```
karanmehta93: same comment as above.

phoenix-core/src/it/java/org/apache/phoenix/monitoring/PhoenixLoggingMetricsIT.java
```diff
@@ -164,13 +167,51 @@ public void testPhoenixMetricsLoggedOnClose() throws Exception {
                 mutationReadMetricsMap.size() == 0);
     }
 
+
+    @Test
+    public void testPhoenixMetricsLoggedOnAutoCommitTrue() throws Exception {
```
karanmehta93: nit: just add a description of test above as javadoc comment

PR #339, opened 2018-08-31
PHOENIX-4855 Continue to write base table column metadata when creati…

twdsilva:
…ng a view in order to support rollback

Previously in MetadataEndpointImpl we were removing parent column metadata while creating a child view. We still need to write the parent columns in order to be able to rollback if needed. This PR filters out the parent columns while creating the PTable of a child view. When a view is resolved we also resolve all its parents and add their columns (see MetadataEndpointImpl.addDerivedColumnsFromAncestors). 
I added some more tests for salted tables. I also changed PhoenixDatabaseMetaData.getPrimaryKeys() to load the PTable and return the metadata instead of directly querying the SYSTEM.CATALOG table (which will not work any more since we don't store columns inherited from parents in a view)

@ankitsinghal  @ChinmaySKulkarni  can you please review?

### DIFF
--- a/phoenix-core/src/it/java/org/apache/phoenix/end2end/AlterTableWithViewsIT.java
+++ b/phoenix-core/src/it/java/org/apache/phoenix/end2end/AlterTableWithViewsIT.java
@@ -73,19 +73,22 @@ public class AlterTableWithViewsIT extends SplitSystemCatalogIT {
 
     private final boolean isMultiTenant;
     private final boolean columnEncoded;
+    private final boolean salted;
     private final String TENANT_SPECIFIC_URL1 = getUrl() + ';' + TENANT_ID_ATTRIB + "=" + TENANT1;
     private final String TENANT_SPECIFIC_URL2 = getUrl() + ';' + TENANT_ID_ATTRIB + "=" + TENANT2;
     
-    public AlterTableWithViewsIT(boolean isMultiTenant, boolean columnEncoded) {
-        this.isMultiTenant = isMultiTenant;
+    public AlterTableWithViewsIT(boolean columnEncoded, boolean isMultiTenant, boolean salted) {
         this.columnEncoded = columnEncoded;
+        this.isMultiTenant = isMultiTenant;
+        this.salted = salted;
     }
     
-    @Parameters(name="AlterTableWithViewsIT_multiTenant={0}, columnEncoded={1}") // name is used by failsafe as file name in reports
+    // name is used by failsafe as file name in reports
+    @Parameters(name = "AlterTableWithViewsIT_columnEncoded={0}, multiTenant={1}, salted={2}")
     public static Collection<Boolean[]> data() {
-        return Arrays.asList(new Boolean[][] { 
-                { false, false }, { false, true },
-                { true, false }, { true, true } });
+        return Arrays.asList(new Boolean[][] { { false, false, false }, { false, false, true },
+                { false, true, false }, { false, true, true }, { true, false, false },
+                { true, false, true }, { true, true, false }, { true, true, true } });
     }
     
     // transform PColumn to String
@@ -112,6 +115,11 @@ public class AlterTableWithViewsIT extends SplitSystemCatalogIT {
                 optionsBuilder.append(",");
             optionsBuilder.append("MULTI_TENANT=true");
         }
+        if (salted) {
+            if (optionsBuilder.length()!=0)
+                optionsBuilder.append(",");
+            optionsBuilder.append("SALT_BUCKETS=4");
+        }
         return String.format(format, isMultiTenant ? "TENANT_ID VARCHAR NOT NULL, " : "",
             isMultiTenant ? "TENANT_ID, " : "", optionsBuilder.toString());
     }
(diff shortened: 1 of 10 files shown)


twdsilva:
ping @ChinmaySKulkarni @karanmehta93 any chance I can get this reviewed soon?

phoenix-core/src/main/java/org/apache/phoenix/schema/PTableImpl.java
```diff
@@ -226,10 +228,13 @@ public PTableImpl(PName tenantId, PName schemaName, PName tableName, long timest
             familyByString.put(family.getName().getString(), family);
         }
         this.families = families;
+        if (bucketNum!=null) {
```
ChinmaySKulkarni: Can you add a comment as to why we are leaving out the first column here?

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/MetaDataEndpointImpl.java
```diff
@@ -1879,42 +1932,6 @@ public void createTable(RpcController controller, CreateTableRequest request,
                 dropChildMetadata(schemaName, tableName, tenantIdBytes);
             }
             
-            // Here we are passed the parent's columns to add to a view, PHOENIX-3534 allows for a splittable
```
ChinmaySKulkarni: You can mark PHOENIX-4767 fixed as well, with this PR.

ChinmaySKulkarni:
@twdsilva Had a quick glance over the PR. I had a few questions. So in the case a view is created on top of a salted base table, we explicitly filter out the salting column (first column) when resolving the columns for a view, right? Also, in case any ancestor of a view drops a column, that will be part of the excluded columns list and we filter out those columns as well, when resolving a view? What happens if a view diverges from the base table, the base table drops a column and then the diverged view re-adds this column?

Overall, can you please add more comments to make it easier to understand the steps taken when handling excluded columns? I will go over the PR in detail and let you know some more feedback. Thanks

phoenix-core/src/it/java/org/apache/phoenix/end2end/AlterTableWithViewsIT.java
```diff
@@ -810,59 +827,65 @@ public void testDivergedViewsStayDiverged() throws Exception {
...
+            
             // Add a new regular column and pk column  to the base table
             String alterBaseTable = "ALTER TABLE " + baseTable + " ADD V3 VARCHAR, PK2 VARCHAR PRIMARY KEY";
             conn.createStatement().execute(alterBaseTable);
             
             // Column V3 shouldn't have propagated to the diverged view.
-            String sql = "SELECT V3 FROM " + view1;
             try {
-                viewConn.createStatement().execute(sql);
+                viewConn.createStatement().execute("SELECT V3 FROM " + view1);
```
ChinmaySKulkarni: Assert.fail in case this exception is not caught. Similarly for all other places where we expect an exception.

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/MetaDataEndpointImpl.java
```diff
@@ -797,7 +810,7 @@ private PTable addDerivedColumnsFromAncestors(PTable table, long timestamp,
                 if (hasIndexId) {
                     // add all pk columns of parent tables to indexes
                     // skip salted column as it will be added from the base table columns
-                    startIndex = pTable.getBucketNum() != null ? 1 : 0;
+                    int startIndex = pTable.getBucketNum() != null ? 1 : 0;
                     for (int index=startIndex; index<pTable.getPKColumns().size(); index++) {
                         PColumn column = pTable.getPKColumns().get(index);
```
ChinmaySKulkarni: Please rename `column` to `pkColumn` to make it easier to understand

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/MetaDataEndpointImpl.java
```diff
@@ -718,9 +731,9 @@ private PTable addDerivedColumnsFromAncestors(PTable table, long timestamp,
         List<PColumn> excludedColumns = Lists.newArrayList();
```
ChinmaySKulkarni: These are for primary key columns present in the base table, but dropped in the view, right? Let's rename this to `excludedPkColumns`

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/MetaDataEndpointImpl.java
```diff
@@ -3243,20 +3270,31 @@ public MetaDataMutationResult updateMutation(PTable table, byte[][] rowKeyMetaDa
                                 addingCol = true;
                                 if (pkCount > FAMILY_NAME_INDEX
                                         && rowKeyMetaData[PhoenixDatabaseMetaData.FAMILY_NAME_INDEX].length > 0) {
+                                byte[] familyName = rowKeyMetaData[PhoenixDatabaseMetaData.FAMILY_NAME_INDEX];
+                                byte[] columnName = rowKeyMetaData[PhoenixDatabaseMetaData.COLUMN_NAME_INDEX];
+                                if (table.getExcludedColumns().contains(
+                                    PColumnImpl.createExcludedColumn(newPName(familyName), newPName(columnName), 0l))) {
+                                    // if this column was previously dropped in a view do not allow adding the column back
+                                    return new MetaDataMutationResult(
```
ChinmaySKulkarni: Only disallow adding it back if the data type is different instead?

ChinmaySKulkarni:
A few comments. Please also address the test failures inside `ViewIT.java`. Otherwise, lgtm.

PR #343, opened 2018-09-06
PHOENIX-4892 Unable to start load balancer with queryserver

MrSandmanRUS:
Added phoenix_loadbalancer_jar path to phoenix_utils.py,
added phoenix_loadbalancer_jar to classpath in queryserver.py,
added service for registry in meta-inf

### DIFF
--- a/bin/phoenix_utils.py
+++ b/bin/phoenix_utils.py
@@ -71,6 +71,7 @@ def setPath():
     PHOENIX_CLIENT_JAR_PATTERN = "phoenix-*-client.jar"
     PHOENIX_THIN_CLIENT_JAR_PATTERN = "phoenix-*-thin-client.jar"
     PHOENIX_QUERYSERVER_JAR_PATTERN = "phoenix-*-queryserver.jar"
+    PHOENIX_LOADBALANCER_JAR_PATTERN = "phoenix-load-balancer-*[!t][!e][!s][!t][!s].jar"
     PHOENIX_TRACESERVER_JAR_PATTERN = "phoenix-tracing-webapp-*-runnable.jar"
     PHOENIX_TESTS_JAR_PATTERN = "phoenix-core-*-tests*.jar"
     PHOENIX_PHERF_JAR_PATTERN = "phoenix-pherf-*-minimal*.jar"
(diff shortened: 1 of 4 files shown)


MrSandmanRUS:
@joshelser can you review, please?

bin/phoenix_utils.py
```diff
@@ -71,6 +71,7 @@ def setPath():
     PHOENIX_CLIENT_JAR_PATTERN = "phoenix-*-client.jar"
     PHOENIX_THIN_CLIENT_JAR_PATTERN = "phoenix-*-thin-client.jar"
     PHOENIX_QUERYSERVER_JAR_PATTERN = "phoenix-*-queryserver.jar"
+    PHOENIX_LOADBALANCER_JAR_PATTERN = "phoenix-load-balancer-*[!t][!e][!s][!t][!s].jar"
```
joshelser: I think all of the extra square-brackets are unnecessary here. Wouldn't `phoenix-load-balancer-*[!tests].jar` be sufficient?

bin/phoenix_utils.py
```diff
@@ -162,6 +163,13 @@ def setPath():
     if phoenix_queryserver_jar == "":
         phoenix_queryserver_jar = findFileInPathWithoutRecursion(PHOENIX_QUERYSERVER_JAR_PATTERN, os.path.join(current_dir, ".."))
 
+    global phoenix_loadbalancer_jar
+    phoenix_loadbalancer_jar = find(PHOENIX_LOADBALANCER_JAR_PATTERN, os.path.join(current_dir, "..", "phoenix-loadbalancer", "target", "*"))
+    if phoenix_loadbalancer_jar == "":
+        phoenix_loadbalancer_jar = findFileInPathWithoutRecursion(PHOENIX_LOADBALANCER_JAR_PATTERN, os.path.join(current_dir, "..", "lib"))
```
joshelser: I'm just noticing in my 5.1.0-SNAPSHOT build that the phoenix-load-balancer jar isn't there at all. Does phoenix-assembly need to be updated as well?

bin/phoenix_utils.py
MrSandmanRUS: Unfortunately, it is impossible. Here used fnmatch (https://docs.python.org/2/library/fnmatch.html) and [!tests] means matches any character not in "tests" (just 1 character in the end), so if we will have phoenix-load-balancer-someversion-snapshot.jar we will not find it.

bin/phoenix_utils.py
MrSandmanRUS: @joshelser Load balancer is presented as a separate jar-file and the query-server.jar is uses for connection to load-balancer special java-class - ServiceLoader, which is looking for the right jar-file with implementation of necessary class. For release you just need to rebuild load-balancer.jar with my fix ("added service for registry in meta-inf"). For this you can use pom.xml in the load-balancer package.

joshelser: Yes, I understand how the load-balancer is meant to work (I was involved in the reviews and stipulated it be this way ;)).

I'm trying to point out that the logic you've added here is useless on master because the jar doesn't exist in the lib directory of the tarball created by the phoenix-assembly module. Specifically, the following path is never valid because the JAR is not put there on normal tarball builds of Phoenix
```
findFileInPathWithoutRecursion(PHOENIX_LOADBALANCER_JAR_PATTERN, os.path.join(current_dir, "..", "lib"))
```

MrSandmanRUS: Oh, I understand, yeah, I'll fix it

MrSandmanRUS: @joshelser ready :)

bin/phoenix_utils.py
joshelser: Bummer. Maybe we switch this over to just use normal Python `re` later. Fine for now -- thanks for clarifying!

joshelser:
Ok, this looks good to go. Thanks for the quick changes @MrSandmanRUS.

Want to file another Jira issue to make sure the load balance jars make it into the normal phoenix-assembly tarball (and then, re-add the check inside of lib/)?

MrSandmanRUS:
@joshelser yeah, I can do it, but can you help me, where I can find file with specification of libs inside phoenix-assembly tarball? Is it /phoenix/pom.xml? (main pom)

joshelser:
You need to add the phoenix-loadbalancer module as a dependency to phoenix-assembly/pom.xml. I think `phoenix-assembly/src/build/components/all-common-dependencies.xml` already has the necessary configuration in place to land the dependencies into lib/ for you.

MrSandmanRUS:
Can I do it in current branch?

MrSandmanRUS:
or is it better to create a new issue?

joshelser:
> or is it better to create a new issue?

Create a new issue in a little bit please. I'm already working through applying this change :)

MrSandmanRUS:
ok, thank you!

PR #352, opened 2018-09-18
PHOENIX-4909 Missing the phoenix-loadbalancer module dependency

MrSandmanRUS:
Missing the phoenix-loadbalancer module dependency in phoenix-assembly/pom.xml.

Fixed phoenix-assembly.pom
Fixed phoenix_utils.py

### DIFF
--- a/bin/phoenix_utils.py
+++ b/bin/phoenix_utils.py
@@ -166,6 +166,8 @@ def setPath():
     global phoenix_loadbalancer_jar
     phoenix_loadbalancer_jar = find(PHOENIX_LOADBALANCER_JAR_PATTERN, os.path.join(current_dir, "..", "phoenix-loadbalancer", "target", "*"))
     if phoenix_loadbalancer_jar == "":
+        phoenix_loadbalancer_jar = findFileInPathWithoutRecursion(PHOENIX_LOADBALANCER_JAR_PATTERN, os.path.join(current_dir, "..", "lib"))
+    if phoenix_loadbalancer_jar == "":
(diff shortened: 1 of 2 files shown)


MrSandmanRUS:
@joshelser you welcome =)

MrSandmanRUS:
@joshelser these are the fixes that we discussed in PHOENIX-4892, can you review please?

joshelser:
Thanks, @MrSandmanRUS. Sorry for the delay, Vitaliy :)

Let me try to merge this today.

joshelser:
Looks like your change doesn't actually get the phoenix-load-balancer jar showing up. Let me modify..

MrSandmanRUS:
ok

MrSandmanRUS:
Oh, ok, I understood, add commit for all-common-jars.xml?

MrSandmanRUS:
Um, there might be a problem, because we can not add something like "phoenix-load-balancer-*.jar" (we also have phoenix-load-balancer-*-tests.jar), in Python I used "phoenix-load-balancer-*[!t][!e][!s][!t][!s].jar", but don't know, can we use this construction in maven.

joshelser:
> we can not add something like "phoenix-load-balancer-*.jar" (we also have phoenix-load-balancer-*-tests.jar)

We have the tests jars included in `lib` for every other Maven module, so I think this is OK.

MrSandmanRUS:
Ok, super, so I can add commit with phoenix-load-balancer-\*.jar

PR #355, opened 2018-09-21
support alter modify column grammar

jaanai0:
#PHOENIX-4815 

@joshelser @twdsilva  Please help me reviewing this patch.

### DIFF
--- a/phoenix-core/src/it/java/org/apache/phoenix/end2end/AlterTableIT.java
+++ b/phoenix-core/src/it/java/org/apache/phoenix/end2end/AlterTableIT.java
@@ -17,35 +17,6 @@
  */
 package org.apache.phoenix.end2end;
 
-import static org.apache.phoenix.jdbc.PhoenixDatabaseMetaData.COLUMN_FAMILY;
-import static org.apache.phoenix.jdbc.PhoenixDatabaseMetaData.COLUMN_NAME;
-import static org.apache.phoenix.jdbc.PhoenixDatabaseMetaData.COLUMN_QUALIFIER;
-import static org.apache.phoenix.jdbc.PhoenixDatabaseMetaData.COLUMN_QUALIFIER_COUNTER;
-import static org.apache.phoenix.jdbc.PhoenixDatabaseMetaData.TABLE_NAME;
-import static org.apache.phoenix.jdbc.PhoenixDatabaseMetaData.TABLE_SCHEM;
-import static org.apache.phoenix.jdbc.PhoenixDatabaseMetaData.TABLE_SEQ_NUM;
-import static org.apache.phoenix.query.QueryConstants.DEFAULT_COLUMN_FAMILY;
-import static org.apache.phoenix.query.QueryConstants.ENCODED_CQ_COUNTER_INITIAL_VALUE;
-import static org.apache.phoenix.util.TestUtil.TEST_PROPERTIES;
-import static org.apache.phoenix.util.TestUtil.closeConnection;
-import static org.apache.phoenix.util.TestUtil.closeStatement;
-import static org.junit.Assert.assertEquals;
-import static org.junit.Assert.assertFalse;
-import static org.junit.Assert.assertNotNull;
-import static org.junit.Assert.assertNull;
-import static org.junit.Assert.assertTrue;
-import static org.junit.Assert.fail;
-
-import java.sql.Connection;
-import java.sql.DriverManager;
-import java.sql.PreparedStatement;
-import java.sql.ResultSet;
-import java.sql.SQLException;
-import java.sql.Statement;
-import java.util.Arrays;
-import java.util.Collection;
-import java.util.Properties;
-
 import org.apache.hadoop.hbase.TableName;
 import org.apache.hadoop.hbase.client.Admin;
 import org.apache.hadoop.hbase.client.ColumnFamilyDescriptor;
@@ -62,15 +33,27 @@ import org.apache.phoenix.schema.PTable.EncodedCQCounter;
 import org.apache.phoenix.schema.PTable.QualifierEncodingScheme;
 import org.apache.phoenix.schema.PTableKey;
 import org.apache.phoenix.schema.TableNotFoundException;
-import org.apache.phoenix.util.IndexUtil;
-import org.apache.phoenix.util.PropertiesUtil;
-import org.apache.phoenix.util.SchemaUtil;
-import org.apache.phoenix.util.TestUtil;
+import org.apache.phoenix.schema.types.PDataType;
+import org.apache.phoenix.schema.types.PDecimal;
+import org.apache.phoenix.schema.types.PInteger;
+import org.apache.phoenix.util.*;
 import org.junit.Before;
 import org.junit.Test;
 import org.junit.runner.RunWith;
 import org.junit.runners.Parameterized;
 import org.junit.runners.Parameterized.Parameters;
+
+import java.math.BigDecimal;
+import java.sql.*;
+import java.util.Arrays;
+import java.util.Collection;
+import java.util.Properties;
+
+import static org.apache.phoenix.jdbc.PhoenixDatabaseMetaData.*;
+import static org.apache.phoenix.query.QueryConstants.DEFAULT_COLUMN_FAMILY;
+import static org.apache.phoenix.query.QueryConstants.ENCODED_CQ_COUNTER_INITIAL_VALUE;
+import static org.apache.phoenix.util.TestUtil.*;
+import static org.junit.Assert.*;
 
 /**
  *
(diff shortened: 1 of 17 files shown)


phoenix-core/src/it/java/org/apache/phoenix/end2end/AlterTableIT.java
```diff
@@ -1294,7 +1277,162 @@ public void testAddingColumnsToTablesAndViews() throws Exception {
...
+    public void testModifyColumn() throws Exception {
+        String schemaName = generateUniqueName();
+        String baseTableName =  generateUniqueName();
+        String tableName = schemaName + "." + baseTableName;
+        Properties props = new Properties();
+        props.put(QueryServices.IS_NAMESPACE_MAPPING_ENABLED, Boolean.toString(true));
+        try (Connection conn = DriverManager.getConnection(getUrl(), props)) {
+            conn.createStatement().execute("CREATE SCHEMA " + schemaName);
+            PhoenixConnection phxConn = conn.unwrap(PhoenixConnection.class);
+            conn.createStatement().execute(
```
twdsilva: What happens if you decrease the length of a varchar column or decreasing the scale and/or precision of a decimal column that has existing data?

phoenix-core/src/main/antlr3/PhoenixSQL.g
```diff
@@ -655,9 +656,17 @@ alter_session_node returns [AlterSessionStatement ret]
...
+@init { PTableType tt = null;}
+    :   ALTER (TABLE | v=VIEW) t=from_table_name { tt = v==null ? (QueryConstants.SYSTEM_SCHEMA_NAME.equals(t.getSchemaName()) ? PTableType.SYSTEM : PTableType.TABLE) : PTableType.VIEW;}
+            (
+                DROP COLUMN (IF ex=EXISTS)? c=column_names { ret = factory.dropColumn(factory.namedTable(null, t), tt, c, ex!=null);}
+                |
+                    (
+                        ADD (IF NOT ex=EXISTS)? cds=column_defs (p=fam_properties)?
+                        | SET p=fam_properties
+                     ) { ret = factory.addColumn(factory.namedTable(null, t), tt, cds, ex!=null, p);}
+                | MODIFY (IF NOT ex=EXISTS)? cd=column_def { ret = factory.modifyColumn(factory.namedTable(null, t), tt, cd, ex!=null);}
```
twdsilva: If the column does not exists and we try to modify it just throw an exception.

phoenix-core/src/main/java/org/apache/phoenix/schema/MetaDataClient.java
```diff
@@ -3300,6 +3281,134 @@ private void mutateStringProperty(String tenantId, String schemaName, String tab
         }
     }
 
+    public MutationState modifyColumn(ModifyColumnStatement statement) throws SQLException {
+        PTable table = FromCompiler.getResolver(statement, connection).getTables().get(0).getTable();
+        ColumnDef columnDef = statement.getColumnDef();
+        ColumnName columnName =columnDef.getColumnDefName();
+
+        // we can not modify index table/system table and project table.
+        if (table.getType() != PTableType.TABLE && table.getType() != PTableType.VIEW) {
```
twdsilva: just check if the type is TABLE or VIEW

phoenix-core/src/main/java/org/apache/phoenix/schema/MetaDataClient.java
```diff
@@ -3300,6 +3281,134 @@ private void mutateStringProperty(String tenantId, String schemaName, String tab
...
+        }
+
+        if (oldColumn == null) {
+            throw new ColumnNotFoundException(table.getSchemaName().getString(), table.getTableName().getString(),
+                    columnName.getFamilyName(), columnName.getColumnName());
+        }
+
+        // Comparision of row keys were affected when we changed max length of pk columns to pad more placeholder,
+        // so we can not modify length of the PK column.
+        if (oldColumn.isRowTimestamp() || SchemaUtil.isPKColumn(oldColumn)) {
```
twdsilva: What happens if you are changing a column that is present in the primary key of an index table?

phoenix-core/src/main/java/org/apache/phoenix/schema/DataTypeCastException.java
```diff
@@ -0,0 +1,35 @@
...
+import org.apache.phoenix.exception.SQLExceptionCode;
+import org.apache.phoenix.exception.SQLExceptionInfo;
+
+import java.sql.SQLException;
+
+public class DataTypeCastException extends SQLException {
+    private static final long serialVersionUID = 1L;
+    private static SQLExceptionCode code = SQLExceptionCode.INVALID_DATA_TYPE_CAST;
+
+    public DataTypeCastException(String newColumn, String oldColumn) {
```
twdsilva: Maybe also include the incompatible data types here

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/MetaDataEndpointImpl.java
```diff
@@ -4588,4 +4423,113 @@ private TableName getParentPhysicalTableName(PTable table) {
...
+                            } else {
+                                column = table.getPKColumn(new String(rkmd[PhoenixDatabaseMetaData.COLUMN_NAME_INDEX]));
+                            }
+                        }
+
+                        // After PHOENIX-3534, we don't store parent table column metadata along with the child metadata,
+                        // so we don't need to propagate changes to the child views.
+
+                        // Since the row key is not nullable fixed length, the data type has been translated to variable
+                        // length when index tables were created, So we will not process data type of index tables.
```
twdsilva: If the table has indexes you need to modify the covered columns that have changed, right? In metadata client you will need to prevent modifying any indexed columns as well.

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/MetaDataEndpointImpl.java
```diff
@@ -2853,12 +2687,13 @@ MetaDataMutationResult updateMutation(PTable table, byte[][] rowKeyMetaData,
                     return result;
                 } else {
                     table = buildTable(key, cacheKey, region, HConstants.LATEST_TIMESTAMP, clientVersion, false, false, null);
-                    return new MetaDataMutationResult(MutationCode.TABLE_ALREADY_EXISTS, currentTime, table);
```
twdsilva: Is there a reason the MutationCode returned needs to change?

phoenix-core/src/main/java/org/apache/phoenix/schema/MetaDataClient.java
```diff
@@ -3300,6 +3281,134 @@ private void mutateStringProperty(String tenantId, String schemaName, String tab
...
+            throw new SQLExceptionInfo.Builder(SQLExceptionCode.DISALLOW_MODIFY_TIMESTAMP_OR_PK_COLUMN).build().buildException();
+        }
+
```
twdsilva: Also add a check if the column is from a view that it is not derived (i.e. inherited from the parent table). You should thrown an exception in this case.

phoenix-core/src/main/antlr3/PhoenixSQL.g
jaanai0: Yep, that will be confirmed in MetaDataClient
```java
 if (oldColumn == null) {
            throw new ColumnNotFoundException(table.getSchemaName().getString(), table.getTableName().getString(),
                    columnName.getFamilyName(), columnName.getColumnName());
        }
```

phoenix-core/src/it/java/org/apache/phoenix/end2end/AlterTableIT.java
jaanai0: Sorry, I didn't forget to test these case.  I saw that a row of query results will be compressed a KeyValue which is byte representation of the `KeyValueSchema`，so this patch will not work,  I will initializing a PR again after fixing the problems.

phoenix-core/src/main/java/org/apache/phoenix/schema/MetaDataClient.java
jaanai0: what else type of tables should be modified? I think just TABLE and VIEW are be modified

phoenix-core/src/main/java/org/apache/phoenix/schema/DataTypeCastException.java
jaanai0: OK 👍

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/MetaDataEndpointImpl.java
jaanai0: Yes

phoenix-core/src/main/antlr3/PhoenixSQL.g
twdsilva: I meant that you don't need to add the "IF NOT EXISTS" to the grammar.

phoenix-core/src/main/java/org/apache/phoenix/schema/MetaDataClient.java
twdsilva: I misread this, ignore my previous comment.

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/MetaDataEndpointImpl.java
twdsilva: You can add a new COLUMN_MODIFIED MutationCode, but no need to change the existing MutationCode returned for drop column etc.

phoenix-core/src/main/java/org/apache/phoenix/schema/MetaDataClient.java
jaanai0: There is no effect.

jaanai0:
I had updated this patch and added some test. Please review :) @twdsilva

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/MetaDataEndpointImpl.java
```diff
@@ -4588,4 +4424,195 @@ private TableName getParentPhysicalTableName(PTable table) {
...
+                                ColumnReference coveredColumn = indexMaintainer.getCoveredColumnsOfIndexTable(
+                                                new ColumnReference(column.getFamilyName().getBytes(), column.getColumnQualifierBytes()));
+
+                                // Since the columns of fixed length which in present in primary key of index table will be converted
+                                // to variable length when index tables are created, So we will not process indexed columns.
+                                if (isColumnIndexed) {
+                                    // do nothing
+                                }
+
+                                // Modify length/scala of covered columns
```
twdsilva: typo: scale

phoenix-core/src/it/java/org/apache/phoenix/end2end/AlterTableIT.java
```diff
@@ -1294,7 +1276,272 @@ public void testAddingColumnsToTablesAndViews() throws Exception {
...
+            conn.createStatement().execute("DROP TABLE IF EXISTS " + tableName);
+            conn.createStatement().execute("CREATE TABLE " + tableName +
+                    " (ID CHAR(3) NOT NULL, COL1 VARCHAR(5), COL2 VARCHAR(3), COL3 CHAR(4), COL4 CHAR(4)"
+                    + " CONSTRAINT PKVIEW PRIMARY KEY(ID, COL1)) " + tableDDLOptions);
+            conn.createStatement().execute("CREATE INDEX IDX_COL2 ON "+ tableName + "(COL2) INCLUDE(COL3, COL4)");
+
+            conn.createStatement().executeUpdate("UPSERT INTO " + tableName + " VALUES('123','12345','123','1234','1444')");
+
+            //Increasing char length
+            conn.createStatement().execute("ALTER TABLE " + tableName + " modify COL3 CHAR(6)");
```
twdsilva: Can you also test for increasing the length of an indexed column? Also can you please add a comment saying that this is supported because we convert fixed length columns to variable length while creating the index.

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/MetaDataEndpointImpl.java
```diff
@@ -4588,4 +4424,195 @@ private TableName getParentPhysicalTableName(PTable table) {
...
+
+                                IndexMaintainer indexMaintainer = index.getIndexMaintainer(table, connection);
+                                boolean isColumnIndexed = indexMaintainer.getIndexedColumnInfo().contains(
+                                                new Pair<>(column.getFamilyName().getString(), column.getName().getString()));
+                                ColumnReference coveredColumn = indexMaintainer.getCoveredColumnsOfIndexTable(
+                                                new ColumnReference(column.getFamilyName().getBytes(), column.getColumnQualifierBytes()));
+
+                                // Since the columns of fixed length which in present in primary key of index table will be converted
+                                // to variable length when index tables are created, So we will not process indexed columns.
+                                if (isColumnIndexed) {
```
twdsilva: remove this boolean as its not used, just keep the comment about why we don't need to do anything for indexed columns.

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/MetaDataProtocol.java
```diff
@@ -381,10 +371,10 @@ public static MetaDataMutationResult constructFromProto(MetaDataResponse proto)
               result.functions.add(PFunction.createFromProto(function));
           }
           if (proto.getTablesToDeleteCount() > 0) {
-            result.tableNamesToDelete =
+            result.mutatedTableNames =
                 Lists.newArrayListWithExpectedSize(proto.getTablesToDeleteCount());
```
twdsilva: I think it should be possible to rename this field in the MetaDataResponse proto as well without affecting backward compatiblity (see https://developers.google.com/protocol-buffers/docs/proto3) WDYT?

phoenix-core/src/main/java/org/apache/phoenix/exception/SQLExceptionCode.java
```diff
@@ -384,6 +367,12 @@ public SQLException newException(SQLExceptionInfo info) {
     INVALID_IMMUTABLE_STORAGE_SCHEME_AND_COLUMN_QUALIFIER_BYTES(1137, "XCL37", "If IMMUTABLE_STORAGE_SCHEME property is not set to ONE_CELL_PER_COLUMN COLUMN_ENCODED_BYTES cannot be 0"),
     INVALID_IMMUTABLE_STORAGE_SCHEME_CHANGE(1138, "XCL38", "IMMUTABLE_STORAGE_SCHEME property cannot be changed from/to ONE_CELL_PER_COLUMN "),
     CANNOT_SET_GUIDE_POST_WIDTH(1139, "XCL39", "Guide post width can only be set on base data tables"),
+
+    INVALID_DATA_TYPE_CAST(1141, "XCL41", "Invalid data type cast "),
```
twdsilva: Can you change the message to "cannot modify column because of invalid data type cast"

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/MetaDataProtocol.java
jaanai0: Yep,  it will not affect backward compatibility.

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/MetaDataEndpointImpl.java
jaanai0: Ok

phoenix-core/src/main/java/org/apache/phoenix/exception/SQLExceptionCode.java
jaanai0: Ok, Thanks for your review.

twdsilva:
@jaanai0  From the example that @dbwong gave, I don't think we can support decreasing the length of a column. If the column is present in an index this can cause queries to return different results when the index is queried vs the data table.

phoenix-core/src/it/java/org/apache/phoenix/end2end/AlterTableIT.java
```diff
@@ -1294,7 +1276,272 @@ public void testAddingColumnsToTablesAndViews() throws Exception {
...
+        String schemaName = generateUniqueName();
+        String baseTableName =  generateUniqueName();
+        String tableName = schemaName + "." + baseTableName;
+        Properties props = new Properties();
+        props.put(QueryServices.IS_NAMESPACE_MAPPING_ENABLED, Boolean.toString(true));
+        try (Connection conn = DriverManager.getConnection(getUrl(), props)) {
+            conn.setAutoCommit(true);
+            conn.createStatement().execute("CREATE SCHEMA " + schemaName);
+
+            conn.createStatement().execute("DROP TABLE IF EXISTS " + tableName);
```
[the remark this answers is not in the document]
jaanai0: ok

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/MetaDataEndpointImpl.java
jaanai0: ok

phoenix-core/src/it/java/org/apache/phoenix/end2end/AlterTableIT.java
jaanai0: I got it

jaanai0:
Thanks for your point out.  @dbwong This grammar references ORACLE dialect( https://docs.oracle.com/cd/B28359_01/server.111/b28286/statements_3001.htm#i2103956), only one column can be modified at a time.

The modify grammar will not change old data, which just change schema information. we can get correctly query result on the server side as long as the values of filtering conditions are tampered with on the client side.

For Varchar type,  now we can not trim query results if the length of fields is decreased, so the example that @dbwong gave will be ok, for blew queries that haven't result, I also add a new test case in the code which similar with this scenario.

> SELECT * from table WHERE b=='13'; -- does this return 2 rows? or does it return 0?
SELECT * from tableIdx WHERE b=='13'; -- this will return 0 rows

For Char type, we will forbid to decrease the length of fields. The query results will be trimmed on the client side that can not match actual data. For example:

_CREATE TABLE  tableName  (a VARCHAR(5), b VARCHAR(5), CONSTRAINT PK PRIMARY KEY (a));
CREATE INDEX tableIdx ON  tableName  (b);
UPSERT INTO  tableName Values('b','13555');
ALTER TABLE  tableName  modify b VARCHAR(2);
UPSERT INTO tableName Values('d','13');
SELECT /*+ NO_INDEX*/ b from  tableName  WHERE b='13';
SELECT /*+ NO_INDEX*/ b from " + tableName + " WHERE b='13555';
SELECT \"0:B\" from tableIdx WHERE \"0:B\"='13555';
SELECT \"0:B\" from tableIdx WHERE \"0:B\"='13';_

Output:
13
13(this is a incorrect result)
13555
13

@twdsilva @dbwong  I have updated the patch again.

phoenix-core/src/main/java/org/apache/phoenix/schema/KeyValueSchema.java
```diff
@@ -121,7 +122,15 @@ public boolean isNull(int position, ValueBitSet bitSet) {
                         int nBytes = ptr.getLength();
                         b = ensureSize(b, offset, offset + nBytes);
                         System.arraycopy(ptr.get(), ptr.getOffset(), b, offset, nBytes);
-                        offset += nBytes;
+                        if (field.getMaxLength() != null && field.getMaxLength() > 0 ) {
```
twdsilva: Is this change required to handle existing data of type char after we increase the length? We should only do this padding if the type is PChar right? Is it possible to change the length of any other fixed length data type?

phoenix-core/src/it/java/org/apache/phoenix/end2end/AlterTableIT.java
```diff
@@ -1298,7 +1284,322 @@ public void testAddingColumnsToTablesAndViews() throws Exception {
...
+            {
+                ResultSet rs = conn.createStatement().executeQuery("SELECT /*+ NO_INDEX*/ b from " + tableName + " WHERE b='13'");
+                assertTrue(rs.next());
+                assertEquals("13", rs.getString(1));
+            }
+
+            {
+                ResultSet rs = conn.createStatement().executeQuery("SELECT /*+ NO_INDEX*/ b from " + tableName + " WHERE b='13555'");
+                assertTrue(rs.next());
+                assertEquals("13555", rs.getString(1));
```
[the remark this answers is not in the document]
jaanai0: Ok, I agree.

phoenix-core/src/main/java/org/apache/phoenix/schema/KeyValueSchema.java
jaanai0: @twdsilva  Yes, we should only pad Pchar type, through other data types of fixed width which excepts for PChar have a permanent length of columns. But in this cases that cast type of fixed columns, the results may be incorrect.

jaanai0:
@twdsilva  I have updated the new patch. Please review, thank you very much! :)

jaanai0:
I had forbidden to decrease the length of columns. @twdsilva @dbwong  @ChinmaySKulkarni

dbwong:
I'll try to take another look at this sometime tomorrow.

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/MetaDataEndpointImpl.java
```diff
@@ -4626,4 +4628,189 @@ private TableName getParentPhysicalTableName(PTable table) {
...
+            this.tableKey = tableKey;
+            return this;
+        }
+
+        public TableBuilder setClientVersion(Integer clientVersion) {
+            this.clientVersion = clientVersion;
+            return this;
+        }
+
+        public PTable run() throws Exception {
```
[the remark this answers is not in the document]
twdsilva: I agree, rename to build()

phoenix-core/src/main/java/org/apache/phoenix/expression/ComparisonExpression.java
```diff
@@ -327,6 +328,9 @@ public boolean evaluate(Tuple tuple, ImmutableBytesWritable ptr) {
             rhsLength = StringUtil.getUnpaddedCharLength(rhsBytes, rhsOffset, rhsLength, rhsSortOrder);
         }
         if (lhsDataType == PChar.INSTANCE) {
+            // Due to length of PChar columns may be modified, in order to the values of filters can match original data,
+            // we make rhsLength equal with lhsLength.
```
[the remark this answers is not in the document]
twdsilva: In Comparison.evaluate() when a value is being read from a column whose length is modified, do you get the original value or the modified value (either a substring or with extra padding)? 
Please add a test for this.

Hackeruncle:
@twdsilva @jaanaizhang  The end? 
Please provide an latest patch, or the next version will be fixed.
Thanks.

jaanaizhang:
> @twdsilva @jaanaizhang The end?
> Please provide an latest patch, or the next version will be fixed.
> Thanks.

I will fix this issue in the next major version as soon as possible.

PR #359, opened 2018-09-27
PHOENIX-4764 - Cleanup metadata of child views for a base table that has been dropped

kadirozde:
A given Phoenix table may have one or (child) views and these views may have their own views. So, a given table or view may have a tree of views rooted at this given table or view. When a table or view is deleted, its views are also required to be deleted. Given that there is no transaction mechanism to atomically delete a given table and its view, failures during deleting a table or view will lead to orphaned views. This patch is to make sure that all the child views will be dropped and will not stay around as orphan views after their parent is deleted.

### DIFF
--- a/phoenix-core/src/it/java/org/apache/phoenix/end2end/DropTableWithViewsIT.java
+++ b/phoenix-core/src/it/java/org/apache/phoenix/end2end/DropTableWithViewsIT.java
@@ -1,0 +1,151 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.phoenix.end2end;
+
+import static org.apache.phoenix.util.PhoenixRuntime.TENANT_ID_ATTRIB;
+import static org.junit.Assert.assertTrue;
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.fail;
+
+import java.sql.Connection;
+import java.sql.DriverManager;
+import java.sql.ResultSet;
+import java.util.Arrays;
+import java.util.Collection;
+
+import org.apache.hadoop.hbase.HConstants;
+import org.apache.hadoop.hbase.TableName;
+import org.apache.phoenix.coprocessor.TableViewFinderResult;
+import org.apache.phoenix.coprocessor.ViewFinder;
+import org.apache.phoenix.jdbc.PhoenixDatabaseMetaData;
+
+import org.apache.phoenix.query.QueryServices;
+import org.apache.phoenix.query.QueryServicesOptions;
+import org.apache.phoenix.schema.PTable;
+import org.apache.phoenix.util.SchemaUtil;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.junit.runners.Parameterized;
+import org.junit.runners.Parameterized.Parameters;
+
+@RunWith(Parameterized.class)
+public class DropTableWithViewsIT extends SplitSystemCatalogIT {
+
+    private final boolean isMultiTenant;
+    private final boolean columnEncoded;
+    private final String TENANT_SPECIFIC_URL1 = getUrl() + ';' + TENANT_ID_ATTRIB + "=" + TENANT1;
+    
+    public DropTableWithViewsIT(boolean isMultiTenant, boolean columnEncoded) {
+        this.isMultiTenant = isMultiTenant;
+        this.columnEncoded = columnEncoded;
+    }
+    
+    @Parameters(name="DropTableWithViewsIT_multiTenant={0}, columnEncoded={1}") // name is used by failsafe as file name in reports
+    public static Collection<Boolean[]> data() {
+        return Arrays.asList(new Boolean[][] { 
+                { false, false }, { false, true },
+                { true, false }, { true, true } });
+    }
+
+    private String generateDDL(String format) {
+        return generateDDL("", format);
+    }
+    
+    private String generateDDL(String options, String format) {
+        StringBuilder optionsBuilder = new StringBuilder(options);
+        if (!columnEncoded) {
+            if (optionsBuilder.length() != 0)
+                optionsBuilder.append(",");
+            optionsBuilder.append("COLUMN_ENCODED_BYTES=0");
+        }
+        if (isMultiTenant) {
+            if (optionsBuilder.length() !=0 )
+                optionsBuilder.append(",");
+            optionsBuilder.append("MULTI_TENANT=true");
+        }
+        return String.format(format, isMultiTenant ? "TENANT_ID VARCHAR NOT NULL, " : "",
+            isMultiTenant ? "TENANT_ID, " : "", optionsBuilder.toString());
+    }
+    
+    @Test
+    public void testDropTableWithChildViews() throws Exception {
+        String baseTable = SchemaUtil.getTableName(SCHEMA1, generateUniqueName());
+        try (Connection conn = DriverManager.getConnection(getUrl());
+                Connection viewConn =
+                        isMultiTenant ? DriverManager.getConnection(TENANT_SPECIFIC_URL1) : conn) {
+            String ddlFormat =
+                    "CREATE TABLE IF NOT EXISTS " + baseTable + "  ("
+                            + " %s PK2 VARCHAR NOT NULL, V1 VARCHAR, V2 VARCHAR "
+                            + " CONSTRAINT NAME_PK PRIMARY KEY (%s PK2)" + " ) %s";
+            conn.createStatement().execute(generateDDL(ddlFormat));
+            conn.commit();
+            // Create a view tree (i.e., tree of views) with depth of 2 and fanout factor of 4
+            for (int  i = 0; i < 4; i++) {
+                String childView = SchemaUtil.getTableName(SCHEMA2, generateUniqueName());
+                String childViewDDL = "CREATE VIEW " + childView + " AS SELECT * FROM " + baseTable;
+                viewConn.createStatement().execute(childViewDDL);
+                for (int j = 0; j < 4; j++) {
+                    String grandChildView = SchemaUtil.getTableName(SCHEMA2, generateUniqueName());
+                    String grandChildViewDDL = "CREATE VIEW " + grandChildView + " AS SELECT * FROM " + childView;
+                    viewConn.createStatement().execute(grandChildViewDDL);
+                }
+            }
+            // Drop the base table
+            String dropTable = String.format("DROP TABLE IF EXISTS %s CASCADE", baseTable);
+            conn.createStatement().execute(dropTable);
+
+            // Wait for the tasks for dropping child views to complete. The depth of the view tree is 2, so we expect that
+            // this will be done in two task handling runs, i.e., in tree task handling interval at most in general
+            // by assuming that each non-root level will be processed in one interval. To be on the safe side, we will
+            // wait at most 10 intervals.
+            long halfTimeInterval = config.getLong(QueryServices.TASK_HANDLING_INTERVAL_MS_ATTRIB,
+                    QueryServicesOptions.DEFAULT_TASK_HANDLING_INTERVAL_MS)/2;
+            ResultSet rs = null;
+            boolean timedOut = true;
+            Thread.sleep(3 * halfTimeInterval);
+            for (int i = 3; i < 20; i++) {
+                rs = conn.createStatement().executeQuery("SELECT * " +
+                                " FROM " + PhoenixDatabaseMetaData.SYSTEM_TASK_NAME +
+                                " WHERE " + PhoenixDatabaseMetaData.TASK_TYPE + " = " +
+                                PTable.TaskType.DROP_CHILD_VIEWS.getSerializedValue());
+                Thread.sleep(halfTimeInterval);
+                if (!rs.next()) {
+                    timedOut = false;
+                    break;
+                }
+            }
+            if (timedOut) {
+                fail("Drop child view task execution timed out!");
+            }
+            // Views should be dropped by now
+            TableName linkTable = TableName.valueOf(PhoenixDatabaseMetaData.SYSTEM_CHILD_LINK_NAME_BYTES);
+            TableViewFinderResult childViewsResult = new TableViewFinderResult();
+            ViewFinder.findAllRelatives(getUtility().getConnection().getTable(linkTable),
+                    HConstants.EMPTY_BYTE_ARRAY,
+                    SchemaUtil.getSchemaNameFromFullName(baseTable).getBytes(),
+                    SchemaUtil.getTableNameFromFullName(baseTable).getBytes(),
+                    PTable.LinkType.CHILD_TABLE,
+                    childViewsResult);
+            assertTrue(childViewsResult.getLinks().size() == 0);
+            // There should not be any orphan views
+            rs = conn.createStatement().executeQuery("SELECT * FROM " + PhoenixDatabaseMetaData.SYSTEM_CATALOG_NAME +
+                    " WHERE " + PhoenixDatabaseMetaData.TABLE_SCHEM + " = '" + SCHEMA2 +"'");
+            assertFalse(rs.next());
+        }
+    }
+}
--- a/phoenix-core/src/it/java/org/apache/phoenix/end2end/TenantSpecificTablesDDLIT.java
+++ b/phoenix-core/src/it/java/org/apache/phoenix/end2end/TenantSpecificTablesDDLIT.java
@@ -503,8 +503,10 @@ public class TenantSpecificTablesDDLIT extends BaseTenantSpecificTablesIT {
             assertTableMetaData(rs, PhoenixDatabaseMetaData.SYSTEM_CATALOG_SCHEMA, PhoenixDatabaseMetaData.TYPE_SEQUENCE, PTableType.SYSTEM);
             assertTrue(rs.next());
             assertTableMetaData(rs, SYSTEM_CATALOG_SCHEMA, PhoenixDatabaseMetaData.SYSTEM_STATS_TABLE, PTableType.SYSTEM);
+            assertTrue(rs.next());
+            assertTableMetaData(rs, SYSTEM_CATALOG_SCHEMA, PhoenixDatabaseMetaData.SYSTEM_TASK_TABLE, PTableType.SYSTEM);
             assertFalse(rs.next());
-            
+
             rs = meta.getTables(null, "", StringUtil.escapeLike(tenantTable2), new String[] {TABLE.getValue().getString()});
             assertFalse(rs.next());
 
--- a/phoenix-core/src/main/java/org/apache/phoenix/coprocessor/MetaDataEndpointImpl.java
+++ b/phoenix-core/src/main/java/org/apache/phoenix/coprocessor/MetaDataEndpointImpl.java
@@ -86,6 +86,7 @@ import java.security.PrivilegedExceptionAction;
 import java.sql.ResultSetMetaData;
 import java.sql.SQLException;
 import java.sql.Statement;
+import java.sql.Timestamp;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
@@ -1940,7 +1941,7 @@ public class MetaDataEndpointImpl extends MetaDataProtocol implements RegionCopr
                 table = loadTable(env, tableKey, cacheKey, clientTimeStamp, HConstants.LATEST_TIMESTAMP,
                         clientVersion);
             } catch (ParentTableNotFoundException e) {
-                dropChildMetadata(e.getParentSchemaName(), e.getParentTableName(), e.getParentTenantId());
+                dropChildViews(env, e.getParentTenantId(), e.getParentSchemaName(), e.getParentTableName());
             }
             if (table != null) {
                 if (table.getTimeStamp() < clientTimeStamp) {
@@ -1965,7 +1966,7 @@ public class MetaDataEndpointImpl extends MetaDataProtocol implements RegionCopr
             // check if the table was dropped, but had child views that were have not yet
             // been cleaned up by compaction
             if (!Bytes.toString(schemaName).equals(QueryConstants.SYSTEM_SCHEMA_NAME)) {
-                dropChildMetadata(schemaName, tableName, tenantIdBytes);
+                dropChildViews(env, tenantIdBytes, schemaName, tableName);
             }
             
             byte[] parentTableKey = null;
@@ -2346,19 +2347,31 @@ public class MetaDataEndpointImpl extends MetaDataProtocol implements RegionCopr
         }
     }
 
-    private void dropChildMetadata(byte[] schemaName, byte[] tableName, byte[] tenantIdBytes)
+    public static void dropChildViews(RegionCoprocessorEnvironment env, byte[] tenantIdBytes, byte[] schemaName, byte[] tableName)
             throws IOException, SQLException, ClassNotFoundException {
-        TableViewFinderResult childViewsResult = new TableViewFinderResult();
-        findAllChildViews(tenantIdBytes, schemaName, tableName, childViewsResult);
+        Table hTable =
+                ServerUtil.getHTableForCoprocessorScan(env,
+                        SchemaUtil.getPhysicalTableName(
+                                PhoenixDatabaseMetaData.SYSTEM_CHILD_LINK_NAME_BYTES,
+                                env.getConfiguration()));
+        TableViewFinderResult childViewsResult = ViewFinder.findRelatedViews(hTable, tenantIdBytes, schemaName, tableName,
+                PTable.LinkType.CHILD_TABLE, HConstants.LATEST_TIMESTAMP);
+
         if (childViewsResult.hasLinks()) {
+
             for (TableInfo viewInfo : childViewsResult.getLinks()) {
                 byte[] viewTenantId = viewInfo.getTenantId();
                 byte[] viewSchemaName = viewInfo.getSchemaName();
                 byte[] viewName = viewInfo.getTableName();
+                if (logger.isDebugEnabled()) {
+                    logger.debug("dropChildViews :" + Bytes.toString(schemaName) + "." + Bytes.toString(tableName) +
+                            " -> " + Bytes.toString(viewSchemaName) + "." + Bytes.toString(viewName) +
+                            "with tenant id :" + Bytes.toString(viewTenantId));
+                }
                 Properties props = new Properties();
                 if (viewTenantId != null && viewTenantId.length != 0)
                     props.setProperty(PhoenixRuntime.TENANT_ID_ATTRIB, Bytes.toString(viewTenantId));
-                try (PhoenixConnection connection = QueryUtil.getConnectionOnServer(env.getConfiguration())
+                try (PhoenixConnection connection = QueryUtil.getConnectionOnServer(props, env.getConfiguration())
                         .unwrap(PhoenixConnection.class)) {
                     MetaDataClient client = new MetaDataClient(connection);
                     org.apache.phoenix.parse.TableName viewTableName = org.apache.phoenix.parse.TableName
@@ -2369,7 +2382,6 @@ public class MetaDataEndpointImpl extends MetaDataProtocol implements RegionCopr
             }
         }
     }
-
     private boolean execeededIndexQuota(PTableType tableType, PTable parentTable) {
         return PTableType.INDEX == tableType && parentTable.getIndexes().size() >= maxIndexesPerTable;
     }
@@ -2533,6 +2545,7 @@ public class MetaDataEndpointImpl extends MetaDataProtocol implements RegionCopr
                     }
                     throw new IllegalStateException(msg);
                 }
+
                 // drop rows from catalog on this region
                 mutateRowsWithLocks(region, localMutations, Collections.<byte[]> emptySet(), HConstants.NO_NONCE,
                         HConstants.NO_NONCE);
@@ -2645,7 +2658,7 @@ public class MetaDataEndpointImpl extends MetaDataProtocol implements RegionCopr
                         EnvironmentEdgeManager.currentTimeMillis(), null);
             }
 
-            if (tableType == PTableType.TABLE || tableType == PTableType.SYSTEM) {
+            if (tableType == PTableType.TABLE || tableType == PTableType.VIEW || tableType == PTableType.SYSTEM) {
                 // check to see if the table has any child views
                 try (Table hTable =
                         ServerUtil.getHTableForCoprocessorScan(env,
@@ -2655,10 +2668,19 @@ public class MetaDataEndpointImpl extends MetaDataProtocol implements RegionCopr
                     boolean hasChildViews =
                             ViewFinder.hasChildViews(hTable, tenantId, schemaName, tableName,
                                 clientTimeStamp);
-                    if (hasChildViews && !isCascade) {
-                        // DROP without CASCADE on tables with child views is not permitted
-                        return new MetaDataMutationResult(MutationCode.UNALLOWED_TABLE_MUTATION,
-                                EnvironmentEdgeManager.currentTimeMillis(), null);
+                    if (hasChildViews) {
+                        if (!isCascade) {
+                            // DROP without CASCADE on tables with child views is not permitted
+                            return new MetaDataMutationResult(MutationCode.UNALLOWED_TABLE_MUTATION,
+                                    EnvironmentEdgeManager.currentTimeMillis(), null);
+                        }
+                        try {
+                            PhoenixConnection conn = QueryUtil.getConnectionOnServer(env.getConfiguration()).unwrap(PhoenixConnection.class);
+                            TaskRegionObserver.addTask(conn, PTable.TaskType.DROP_CHILD_VIEWS, Bytes.toString(tenantId),
+                                Bytes.toString(schemaName), Bytes.toString(tableName), this.accessCheckEnabled);
+                        } catch (Throwable t) {
+                            logger.error("Adding a task to drop child views failed!", t);
+                        }
                     }
                 }
             }
--- a/phoenix-core/src/main/java/org/apache/phoenix/coprocessor/TaskRegionObserver.java
+++ b/phoenix-core/src/main/java/org/apache/phoenix/coprocessor/TaskRegionObserver.java
@@ -1,0 +1,295 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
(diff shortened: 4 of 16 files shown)


phoenix-core/src/it/java/org/apache/phoenix/end2end/DropTableWithViewsIT.java
```diff
@@ -0,0 +1,164 @@
...
+            assertTrue(childViewsResultBefore.getLinks().size() > 0);
+
+            // Wait for the tasks for dropping child views to complete. The depth of the view tree is 2, so we should wait
+            // 3 times the task handling interval by assuming that each non-root level will be processed in one interval
+
+            long timeInterval = config.getLong(
+                            QueryServices.TASK_HANDLING_INTERVAL_ATTRIB,
+                            QueryServicesOptions.DEFAULT_TASK_HANDLING_INTERVAL);
+
+            Thread.sleep(3*timeInterval);
```
twdsilva: This might flap in some environments, can you run the task manually here so that the child views get cleaned up?

phoenix-core/src/it/java/org/apache/phoenix/end2end/TenantSpecificTablesDDLIT.java
```diff
@@ -503,7 +503,10 @@ public void testTableMetadataScan() throws Exception {
             assertTableMetaData(rs, PhoenixDatabaseMetaData.SYSTEM_CATALOG_SCHEMA, PhoenixDatabaseMetaData.TYPE_SEQUENCE, PTableType.SYSTEM);
             assertTrue(rs.next());
             assertTableMetaData(rs, SYSTEM_CATALOG_SCHEMA, PhoenixDatabaseMetaData.SYSTEM_STATS_TABLE, PTableType.SYSTEM);
+            assertTrue(rs.next());
+            assertTableMetaData(rs, SYSTEM_CATALOG_SCHEMA, PhoenixDatabaseMetaData.SYSTEM_TASK_TABLE, PTableType.SYSTEM);
             assertFalse(rs.next());
+
```
twdsilva: nit: extra whitespace

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/MetaDataEndpointImpl.java
```diff
@@ -2516,6 +2526,8 @@ public void dropTable(RpcController controller, DropTableRequest request,
                     }
                     throw new IllegalStateException(msg);
                 }
+
```
twdsilva: nit: whitespace

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/TaskRegionObserver.java
```diff
@@ -0,0 +1,240 @@
...
+    private long timeMaxInterval = QueryServicesOptions.DEFAULT_TASK_HANDLING_MAX_INTERVAL;
+    @GuardedBy("TaskRegionObserver.class")
+    // Added for test purposes
+    private long initialDelay = 0;
+
+    @Override
+    public void preClose(final ObserverContext<RegionCoprocessorEnvironment> c,
+            boolean abortRequested) {
+        executor.shutdownNow();
+        GlobalCache.getInstance(c.getEnvironment()).getMetaDataCache().invalidateAll();
```
twdsilva: I don't think we need to do this.

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/TaskRegionObserver.java
```diff
@@ -0,0 +1,240 @@
...
+    public Optional<RegionObserver> getRegionObserver() {
+        return Optional.of(this);
+    }
+
+    @Override
+    public void start(CoprocessorEnvironment env) throws IOException {
+        // sleep a little bit to compensate time clock skew when SYSTEM.CATALOG moves
+        // among region servers because we relies on server time of RS which is hosting
+        // SYSTEM.CATALOG
+        Configuration config = env.getConfiguration();
```
twdsilva: Same here, is this really needed?

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/TaskRegionObserver.java
```diff
@@ -0,0 +1,240 @@
...
+                                rs.getString(3), rs.getString(4));
+                    }
+                    catch (Throwable t) {
+                        LOG.warn("Exception while deleting child view " + rs.getString(3) + "." + rs.getString(4) +
+                                " with tenant id " + (rs.getString(2) == null ? " IS NULL" : rs.getString(2)) +
+                                " and timestamp " + rs.getTimestamp(1).toString(), t);
+                    }
+                }
+            } catch (Throwable t) {
+                LOG.warn("DropChildViewsTask failed!", t);
```
twdsilva: Log this as an error

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/TaskRegionObserver.java
```diff
@@ -0,0 +1,240 @@
...
+                        " WHERE "+ PhoenixDatabaseMetaData.TASK_TYPE + " = " + PTable.TaskType.DROP_CHILD_VIEWS.getSerializedValue();
+
+                connForTask = QueryUtil.getConnectionOnServer(env.getConfiguration()).unwrap(PhoenixConnection.class);
+                PreparedStatement taskStatement = connForTask.prepareStatement(taskQuery);
+                ResultSet rs = taskStatement.executeQuery();
+                while (rs.next()) {
+                    try {
+                        // delete child views only if the parent table is deleted from the system catalog
+                        Properties tenantProps = new Properties();
+                        if (rs.getString(3) != null) {
```
twdsilva: the code is more readable if you assign this to a variable (rs.getString(..))

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/TaskRegionObserver.java
```diff
@@ -0,0 +1,240 @@
...
+                            tenantProps.setProperty(PhoenixRuntime.TENANT_ID_ATTRIB, rs.getString(3));
+                        }
+                        PhoenixConnection pconn = QueryUtil.getConnectionOnServer(tenantProps, env.getConfiguration()).unwrap(PhoenixConnection.class);
+
+                        MetaDataProtocol.MetaDataMutationResult result = new MetaDataClient(pconn).updateCache(pconn.getTenantId(),
+                                rs.getString(3), rs.getString(4), true);
+                        if (result.getMutationCode() != MetaDataProtocol.MutationCode.TABLE_ALREADY_EXISTS) {
+                            MetaDataEndpointImpl.dropChildViews(env, rs.getBytes(2), rs.getBytes(3), rs.getBytes(4));
+                        } else if (System.currentTimeMillis() < timeMaxInterval + rs.getTimestamp(1).getTime()) {
+                            // skip this task as it has not been expired and its parent table has not been dropped yet
```
twdsilva: Maybe add an info log line here for this

phoenix-core/src/main/java/org/apache/phoenix/query/QueryConstants.java
```diff
@@ -357,18 +357,28 @@
...
+	        "CONSTRAINT " + SYSTEM_TABLE_PK_NAME + " PRIMARY KEY (" + TENANT_ID + ","
+	        + TABLE_SCHEM + "," + TABLE_NAME + "," + COLUMN_NAME + "," + COLUMN_FAMILY + "))\n" +
+	        HConstants.VERSIONS + "=%s,\n" +
+	        ColumnFamilyDescriptorBuilder.KEEP_DELETED_CELLS + "=%s,\n" +
+	        PhoenixDatabaseMetaData.TRANSACTIONAL + "=" + Boolean.FALSE;
+
+	public static final String CREATE_TASK_METADATA = "CREATE TABLE " + SYSTEM_CATALOG_SCHEMA + ".\""
+            + SYSTEM_TASK_TABLE + "\"(\n" +
+            // PK columns
+            TASK_TYPE + " UNSIGNED_TINYINT NOT NULL," + TASK_TS + " TIMESTAMP NOT NULL," + TENANT_ID + " VARCHAR NULL," + TABLE_SCHEM + " VARCHAR NULL," +
```
twdsilva: You could use the ROW_TIMESTAMP feature to let phoenix set the timestamp automatically, this would speed up queries if there are lots of rows in this table (http://phoenix.apache.org/rowtimestamp.html)

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/TaskRegionObserver.java
```diff
@@ -0,0 +1,261 @@
...
+
+import org.apache.phoenix.util.PhoenixRuntime;
+import org.apache.phoenix.util.QueryUtil;
+
+
+/**
+ * Coprocessor for task related operations. This coprocessor would only be registered
+ * to SYSTEM.TASK table
+ */
+@SuppressWarnings("deprecation")
```
karanmehta93: Why do we have this? What is originating them?

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/TaskRegionObserver.java
```diff
@@ -0,0 +1,261 @@
...
+            DropChildViewsTask task = new DropChildViewsTask(e.getEnvironment(), timeMaxInterval);
+            executor.scheduleWithFixedDelay(task, initialDelay, timeInterval, TimeUnit.MILLISECONDS);
+        } catch (ClassNotFoundException ex) {
+            LOG.error("ScheduleTask cannot start!", ex);
+        }
+    }
+
+    public static void addTask(PhoenixConnection conn, TaskType taskType, String tenantId, String schemaName, String tableName)
+            throws IOException {
+        try {
```
karanmehta93: @twdsilva Do we now require WRITE access to users for doing this?
Should we run this piece of code as `LoginUser` instead?

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/TaskRegionObserver.java
```diff
@@ -0,0 +1,261 @@
...
+ */
+@SuppressWarnings("deprecation")
+public class TaskRegionObserver implements RegionObserver, RegionCoprocessor {
+    public static final Log LOG = LogFactory.getLog(TaskRegionObserver.class);
+    protected ScheduledThreadPoolExecutor executor = new ScheduledThreadPoolExecutor(TaskType.values().length);
+    private long timeInterval = QueryServicesOptions.DEFAULT_TASK_HANDLING_INTERVAL;
+    private long timeMaxInterval = QueryServicesOptions.DEFAULT_TASK_HANDLING_MAX_INTERVAL;
+    @GuardedBy("TaskRegionObserver.class")
+    // Added for test purposes
+    private long initialDelay = 0;
```
karanmehta93: You can instead have a protected variable that you can modify in tests. Adding a config prop is probably overkill.

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/TaskRegionObserver.java
```diff
@@ -0,0 +1,261 @@
...
+
+
+/**
+ * Coprocessor for task related operations. This coprocessor would only be registered
+ * to SYSTEM.TASK table
+ */
+@SuppressWarnings("deprecation")
+public class TaskRegionObserver implements RegionObserver, RegionCoprocessor {
+    public static final Log LOG = LogFactory.getLog(TaskRegionObserver.class);
+    protected ScheduledThreadPoolExecutor executor = new ScheduledThreadPoolExecutor(TaskType.values().length);
```
karanmehta93: I think that its better to make the number of threads configurable and not based on `TaskType` only. We don't often drop tables, but if the variety and heterogeneity of tasks grow in future, we definitely should be able to get more. IMHO, 4 threads should be good default to start with.

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/TaskRegionObserver.java
```diff
@@ -0,0 +1,261 @@
...
+import org.apache.phoenix.util.PhoenixRuntime;
+import org.apache.phoenix.util.QueryUtil;
+
+
+/**
+ * Coprocessor for task related operations. This coprocessor would only be registered
+ * to SYSTEM.TASK table
+ */
+@SuppressWarnings("deprecation")
+public class TaskRegionObserver implements RegionObserver, RegionCoprocessor {
```
karanmehta93: nit: `TaskRegionObserver` --> `SystemTaskRegionObserver`?

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/TaskRegionObserver.java
```diff
@@ -0,0 +1,261 @@
...
+
+    @Override
+    public Optional<RegionObserver> getRegionObserver() {
+        return Optional.of(this);
+    }
+
+    @Override
+    public void start(CoprocessorEnvironment env) throws IOException {
+        Configuration config = env.getConfiguration();
+        timeInterval =
```
karanmehta93: nit: good to have comment or doc here about all the props that we are making configurable.

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/TaskRegionObserver.java
karanmehta93: I have seen checks around `CoprocessorEnvironment` instance type in various pieces of code. Do you think we need one here?

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/TaskRegionObserver.java
```diff
@@ -0,0 +1,261 @@
...
+            PreparedStatement stmt = conn.prepareStatement("UPSERT INTO " +
+                    PhoenixDatabaseMetaData.SYSTEM_TASK_NAME + " (" +
+                    PhoenixDatabaseMetaData.TASK_TYPE + ", " +
+                    PhoenixDatabaseMetaData.TENANT_ID + ", " +
+                    PhoenixDatabaseMetaData.TABLE_SCHEM + ", " +
+                    PhoenixDatabaseMetaData.TABLE_NAME + ") VALUES(?,?,?,?)");
+            stmt.setByte(1, taskType.getSerializedValue());
+            if (tenantId != null) {
+                stmt.setString(2, tenantId);
+            }else {
```
karanmehta93: nit: spaces before/after brackets.

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/TaskRegionObserver.java
karanmehta93: You can also format the log string with `String.format`.

phoenix-core/src/main/java/org/apache/phoenix/query/ConnectionQueryServicesImpl.java
```diff
@@ -3206,6 +3219,9 @@ public void upgradeSystemTables(final String url, final Properties props) throws
             try {
                 metaConnection.createStatement().executeUpdate(getMutexDDL());
             } catch (NewerTableAlreadyExistsException e) {} catch (TableAlreadyExistsException e) {}
+            try {
```
karanmehta93: nit: Can we refactor and extract out these dupe pieces of code into a method?

phoenix-core/src/main/java/org/apache/phoenix/query/QueryConstants.java
```diff
@@ -357,18 +357,28 @@
...
-	            TABLE_NAME + " VARCHAR NOT NULL," +
-	            COLUMN_NAME + " VARCHAR NULL," + // null for table row
-	            COLUMN_FAMILY + " VARCHAR NULL " + // using for CF to uniqueness for columns
-	            "CONSTRAINT " + SYSTEM_TABLE_PK_NAME + " PRIMARY KEY (" + TENANT_ID + ","
-	            + TABLE_SCHEM + "," + TABLE_NAME + "," + COLUMN_NAME + "," + COLUMN_FAMILY + "))\n" +
-	            HConstants.VERSIONS + "=%s,\n" +
-	            ColumnFamilyDescriptorBuilder.KEEP_DELETED_CELLS + "=%s,\n" +
-	            PhoenixDatabaseMetaData.TRANSACTIONAL + "=" + Boolean.FALSE;
-    
+	public static final String CREATE_MUTEX_METADTA =
```
karanmehta93: nit: unnecessary diff

phoenix-core/src/main/java/org/apache/phoenix/query/QueryConstants.java
karanmehta93: nit: different column names in different lines for better readability.

karanmehta93: `UNSIGNED_TINYINT` for `TASK_TYPE` -- Any particular reason for it? What is max value here?

phoenix-core/src/main/java/org/apache/phoenix/query/QueryConstants.java
```diff
@@ -357,18 +357,28 @@
...
+	        PhoenixDatabaseMetaData.TRANSACTIONAL + "=" + Boolean.FALSE;
+
+	public static final String CREATE_TASK_METADATA = "CREATE TABLE " + SYSTEM_CATALOG_SCHEMA + ".\""
+            + SYSTEM_TASK_TABLE + "\"(\n" +
+            // PK columns
+            TASK_TYPE + " UNSIGNED_TINYINT NOT NULL," + TASK_TS + " TIMESTAMP NOT NULL," + TENANT_ID + " VARCHAR NULL," + TABLE_SCHEM + " VARCHAR NULL," +
+            TABLE_NAME + " VARCHAR NOT NULL,\n" +
+            "CONSTRAINT " + SYSTEM_TABLE_PK_NAME + " PRIMARY KEY (" + TASK_TYPE + "," + TASK_TS + " ROW_TIMESTAMP," + TENANT_ID + "," + TABLE_SCHEM + "," +
+            TABLE_NAME + "))\n" +
+            HConstants.VERSIONS + "=%s,\n" +
```
karanmehta93: What value will be populated here? Can we fix the `MAX_VERSIONS` to 1 to make things simpler?

phoenix-core/src/main/java/org/apache/phoenix/schema/PTable.java
```diff
@@ -192,7 +192,39 @@ public static LinkType fromSerializedValue(byte serializedValue) {
...
+        public byte getSerializedValue() {
+            return this.serializedValue;
+        }
+        public static TaskType getDefault() {
+            return DROP_CHILD_VIEWS;
+        }
+        public static TaskType fromToken(String token) {
+            return TaskType.valueOf(token.trim().toUpperCase());
+        }
+        public static TaskType fromSerializedValue(byte serializedValue) {
```
karanmehta93: Why do we need this serialization logic? @twdsilva @kadirozde 
I see that it has been copied from `LinkType` and other Enum's defined above.

phoenix-core/src/main/java/org/apache/phoenix/util/SchemaUtil.java
```diff
@@ -540,6 +540,12 @@ public static boolean isSequenceTable(byte[] tableName) {
                 || Bytes.compareTo(tableName, SchemaUtil
                         .getPhysicalTableName(PhoenixDatabaseMetaData.SYSTEM_SEQUENCE_NAME_BYTES, true).getName()) == 0;
     }
+
+    public static boolean isTaskTable(byte[] tableName) {
```
karanmehta93: There are also some other places in the code where we do special things for SYSTEM tables, one of the examples is that we disable stats for SYSTEM tables. It would good if you can make that pass to make things consistent across code.

phoenix-core/src/it/java/org/apache/phoenix/end2end/DropTableWithViewsIT.java
```diff
@@ -0,0 +1,169 @@
...
+            if (optionsBuilder.length()!=0)
+                optionsBuilder.append(",");
+            optionsBuilder.append("MULTI_TENANT=true");
+        }
+        return String.format(format, isMultiTenant ? "TENANT_ID VARCHAR NOT NULL, " : "",
+            isMultiTenant ? "TENANT_ID, " : "", optionsBuilder.toString());
+    }
+    
+    @Test
+    public void testDropTableWithChildViews() throws Exception {
```
karanmehta93: The test looks pretty big. Do you want it to split this into multiple tests or move some code into `before`and `after` test methods?

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/TaskRegionObserver.java
```diff
@@ -0,0 +1,261 @@
...
+                            Properties tenantProps = new Properties();
+                            tenantProps.setProperty(PhoenixRuntime.TENANT_ID_ATTRIB, tenantId);
+                            pconn = QueryUtil.getConnectionOnServer(tenantProps, env.getConfiguration()).unwrap(PhoenixConnection.class);
+
+                        }
+                        else {
+                            pconn = QueryUtil.getConnectionOnServer(env.getConfiguration()).unwrap(PhoenixConnection.class);
+                        }
+
+                        MetaDataProtocol.MetaDataMutationResult result = new MetaDataClient(pconn).updateCache(pconn.getTenantId(),
```
dbwong: Why are we making a new MetaDataClient per result?  Can't we make it per pconn?

phoenix-core/src/it/java/org/apache/phoenix/end2end/DropTableWithViewsIT.java
```diff
@@ -0,0 +1,169 @@
...
+    }
+
+    private String generateDDL(String format) {
+        return generateDDL("", format);
+    }
+    
+    private String generateDDL(String options, String format) {
+        StringBuilder optionsBuilder = new StringBuilder(options);
+        if (!columnEncoded) {
+            if (optionsBuilder.length()!=0)
```
BinShi-SecularBird: Add white space before and after !=?

phoenix-core/src/it/java/org/apache/phoenix/end2end/DropTableWithViewsIT.java
```diff
@@ -0,0 +1,169 @@
...
+                optionsBuilder.append(",");
+            optionsBuilder.append("COLUMN_ENCODED_BYTES=0");
+        }
+        if (isMultiTenant) {
+            if (optionsBuilder.length()!=0)
```
BinShi-SecularBird: Add white space before and after !=?

phoenix-core/src/it/java/org/apache/phoenix/end2end/DropTableWithViewsIT.java
BinShi-SecularBird: From line 122 to the end of the test, the design of the test case is based on some assumptions which aren't rigorous and cause non-deterministic behavior. For example, the assertion at line 132 could fail at some extreme case because of resource contention on the test machine. I saw many test cases designed in this way were error prone and caused trouble eventually.

BinShi-SecularBird: Maybe you want to add synchronous implementation for "DROP TABLE IF EXISTS ... CASCADE" even just for test purpose, just like what we've done for "UPDATE STATISTICS"

phoenix-core/src/main/java/org/apache/phoenix/query/ConnectionQueryServicesImpl.java
kadirozde: No sure about any value of extracting this simple try catch into a method.

phoenix-core/src/it/java/org/apache/phoenix/end2end/DropTableWithViewsIT.java
twdsilva: I don't think its possible to split this test, it just creates some views, drops the parent and then verifies that our process eventually cleans up the view metadata.

twdsilva: Instead of waiting we could just run the process that reads the tasks so that we don't have to rely on sleeping.

twdsilva: I don't think we need a synchronous drop table. Once the parent is dropped any child views when they are later resolved will throw a TableNotFoundException as the parent doesn't exist.

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/TaskRegionObserver.java
twdsilva: I don't think its needed.

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/TaskRegionObserver.java
twdsilva: Nice catch, if access checks are enabled the user won't have write access to the SYSTEM.TASK table, so you will need to upsert the row using User.runAsLoginUser (see MetaDataEndpointImpl.mutateRowsWithLocks).

phoenix-core/src/main/java/org/apache/phoenix/query/QueryConstants.java
twdsilva: Its configurable with the default set 1.

phoenix-core/src/it/java/org/apache/phoenix/end2end/DropTableWithViewsIT.java
karanmehta93: Alright, then lets try to refactor it into a couple methods. Will help when reading it.

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/TaskRegionObserver.java
karanmehta93: We can actually simplify by not exposing these properties out. The defaults seem to be good enough.
Especially the `initialDelay` parameter which is only used for test purposes.

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/TaskRegionObserver.java
```diff
@@ -0,0 +1,261 @@
...
+    public void postOpen(ObserverContext<RegionCoprocessorEnvironment> e) {
+        final RegionCoprocessorEnvironment env = e.getEnvironment();
+
+        // turn off verbose deprecation logging
+        Logger deprecationLogger = Logger.getLogger("org.apache.hadoop.conf.Configuration.deprecation");
+        if (deprecationLogger != null) {
+            deprecationLogger.setLevel(Level.WARN);
+        }
+        try {
+            Class.forName(PhoenixDriver.class.getName());
```
karanmehta93: We dont need this thing anymore since Java 1.5+, should be okay to remove.

phoenix-core/src/main/java/org/apache/phoenix/query/ConnectionQueryServicesImpl.java
karanmehta93: Sorry for not being clear, we can refactor the method `upgradeSystemTables` and `createOtherSystemTables` in this one.

phoenix-core/src/main/java/org/apache/phoenix/query/QueryConstants.java
karanmehta93: sure, there is some indentation difference between 364-365, would be good to align them as well.

phoenix-core/src/main/java/org/apache/phoenix/schema/PTable.java
karanmehta93: Looking at the code in IDE, I dont see it being used anywhere.

phoenix-core/src/main/java/org/apache/phoenix/util/SchemaUtil.java
karanmehta93: Checkout `StatisticsUtil` class. This is where we disable it.

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/MetaDataEndpointImpl.java
```diff
@@ -2346,19 +2347,29 @@ public void createTable(RpcController controller, CreateTableRequest request,
...
         if (childViewsResult.hasLinks()) {
+
             for (TableInfo viewInfo : childViewsResult.getLinks()) {
                 byte[] viewTenantId = viewInfo.getTenantId();
                 byte[] viewSchemaName = viewInfo.getSchemaName();
                 byte[] viewName = viewInfo.getTableName();
+                logger.debug("dropChildViews :" + Bytes.toString(schemaName) + "." + Bytes.toString(tableName) +
+                        " -> " + Bytes.toString(viewSchemaName) + "." + Bytes.toString(viewName) +
+                        "with tenant id :" + Bytes.toString(viewTenantId));
```
BinShi-SecularBird: In the code base, even just in this file, I saw it prints debug log only when logger.isDebugEnabled() is true. The only exception is that few places print debug info for rare case or failure case (actually it should be changed to info or error log). You might want to add if (logger.isDebugEnabled()) { ... } too for printing debug log.

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/MetaDataEndpointImpl.java
```diff
@@ -2655,10 +2666,19 @@ private MetaDataMutationResult doDropTable(byte[] key, byte[] tenantId, byte[] s
                     boolean hasChildViews =
                             ViewFinder.hasChildViews(hTable, tenantId, schemaName, tableName,
                                 clientTimeStamp);
-                    if (hasChildViews && !isCascade) {
-                        // DROP without CASCADE on tables with child views is not permitted
-                        return new MetaDataMutationResult(MutationCode.UNALLOWED_TABLE_MUTATION,
-                                EnvironmentEdgeManager.currentTimeMillis(), null);
+                    if (hasChildViews) {
+                        if (!isCascade) {
+                            // DROP without CASCADE on tables with child views is not permitted
```
BinShi-SecularBird: Why do we have this limitation? Shouldn't DROP mean "DROP CASCADE" in the case that the table has child views? This comment is orthogonal to this pull request as it exists in the original code base.

twdsilva: Drop without specifying CASCADE will fail if the table has child views, so that the user knows that   child views exist on the table. If the user wants to drop a table and child views they can just used DROP CASCADE.

phoenix-core/src/main/java/org/apache/phoenix/query/ConnectionQueryServicesImpl.java
kadirozde: @karanmehta93, there is really no duplicate pieces of code here. Note that these methods have different try-catch logic

phoenix-core/src/main/java/org/apache/phoenix/query/QueryConstants.java
kadirozde: Will do

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/TaskRegionObserver.java
kadirozde: @dbwong, I do not think we should be concerned about making a newMetaDataClient per result since the objects of this class have only one instance variable which is a pointer to the PhoenixConnection object that is passed via the constructor. Trying to make one MetaDataClient object per connection is probably as expensive as creating a new object per result. I think it will be over engineering if we do that.

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/TaskRegionObserver.java
kadirozde: Will do

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/TaskRegionObserver.java
kadirozde: Will do

phoenix-core/src/main/java/org/apache/phoenix/schema/PTable.java
kadirozde: It is used in my IDE

phoenix-core/src/main/java/org/apache/phoenix/util/SchemaUtil.java
kadirozde: Will do

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/TaskRegionObserver.java
kadirozde: @karanmehta93, I am new to the Phoenix code and following the styles in the existing code base. MetaDataObserver has these parameters defined for the same purposes. I can imagine that we may need to change these values if we hit some issues and a single default value may not be sufficient. I am not sure how you conclude that initialDelay is/will be only used for only test purposes.

karanmehta93: > I am not sure how you conclude that initialDelay is/will be only used for only test purposes.

This is something that's only applicable for once during the lifetime of a region server, since phoenix lies inside HBase process. I am not sure if this will help. This is not the delay that will be applied before every task, is that correct? Also, I see that initialDelay is commented as "for test purposes" on Line 72 in TaskRegionObserver.java.

kadirozde: @karanmehta93, Ok, I will remove the configuration parameter for the initial delay

phoenix-core/src/it/java/org/apache/phoenix/end2end/DropTableWithViewsIT.java
twdsilva: See MutableIndexFailureIT.testIndexWriteFailure which just runs the BuildIndexScheduleTask

kadirozde: That is not the right way of doing an integration testing. That would be more of a unit test. I should not call an internal function directly and verify that it works for the purpose of integration testing. This is one reason not to that. The second reason is that I wanted to verify that the coprocessor (TaskRegionObserver) is initialized,  it creates an instance of DropChildViewsTask,  and this instance is run at regular intervals in addition to DropChildViewsTask functions correctly. The third reason is that even I create another instance of DropChildViewsTask and call it directly, I would not know which instance processed the drop child tasks in the system task table. A view tree can have more one level of views and the run method of DropChildViewsTask needs to be called once for each level. This also introduces concurrency which is not supported by the current implementation.  Please note that I set the scheduling interval to 1 sec  (in BaseTest.java) for testing purpose. Since the test sleeps only half this interval at a time, this test should be done in 2.5 seconds on average (for a view tree of depth 2). The sleeping logic here can only sleep extra 0.5 sec at most. Of course, if the threads are not scheduled on time, this test can take longer. For that reason, the test waits up to 10 seconds.

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/TaskRegionObserver.java
kadirozde: Done

phoenix-core/src/main/java/org/apache/phoenix/query/QueryServices.java
```diff
@@ -322,6 +322,12 @@
 
 	public static final String SYSTEM_CATALOG_SPLITTABLE = "phoenix.system.catalog.splittable";
 
+    // The parameters defined for handling task stored in table SYSTEM.TASK
+	// The time interval between periodic scans of table SYSTEM.TASK
+    public static final String TASK_HANDLING_INTERVAL_ATTRIB = "phoenix.task.handling.interval";
```
karanmehta93: One small nit: Can we add timeunits in the property name?

kadirozde: Will change it to TASK_HANDLING_INTERVAL_MS_ATTRIB

PR #360, opened 2018-10-02
PHOENIX-4731 Make running transactional unit tests for a given provider optional

JamesRTaylor:
Enables turning transaction tests off for Tephra and/or Omid. Please review, @ohadshacham and/or @twdsilva. This is a precursor to support of Omid - no new pom dependencies have been introduced. This also moves us to Tephra 0.15.0 which is required for the TAL changes.


phoenix-core/src/main/java/org/apache/phoenix/execute/PhoenixTxIndexMutationGenerator.java
```diff
@@ -313,7 +326,18 @@ public int compare(Cell o1, Cell o2) {
     private void generateDeletes(PhoenixIndexMetaData indexMetaData,
             Collection<Pair<Mutation, byte[]>> indexUpdates,
             byte[] attribValue, TxTableState state) throws IOException {
-        Iterable<IndexUpdate> deletes = codec.getIndexDeletes(state, indexMetaData);
+        byte[] regionStartKey = this.regionStartKey;
```
twdsilva: Just for my understanding, why do we now need the start and stop key to generate the index deletes and upserts?

phoenix-core/src/main/java/org/apache/phoenix/index/IndexMaintainer.java
```diff
@@ -184,9 +184,24 @@ public boolean apply(PTable index) {
...
         List<PTable> indexes = dataTable.getIndexes();
-        serialize(dataTable, ptr, indexes, connection);
+        serializeServerMaintainedIndexes(dataTable, ptr, indexes, connection);
     }
 
+    public static void serializeServerMaintainedIndexes(PTable dataTable, ImmutableBytesWritable ptr,
+            List<PTable> indexes, PhoenixConnection connection) {
+        Iterator<PTable> indexesItr = Collections.emptyListIterator();
+        boolean onlyLocalIndexes = dataTable.isImmutableRows() || dataTable.isTransactional();
+        if (onlyLocalIndexes) {
```
twdsilva: Do we do anything special for transactional tables with local indexes?

JamesRTaylor: Long story short, we've disallowed local indexes on Omid for now due to PHOENIX-4760. Once this is fixed, we can turn it back on. Given that index updates are transactional with data updates,  there's not a great need IMHO for transactional local indexes, though. Having said that, there were a few things we needed to do for local indexes:
* Support client-side index maintenance for local indexes
* Handle the initial population (i.e. adding shadow columns) for a local index on the server side

stoty:
Already merged.

PR #363, opened 2018-10-12
PHOENIX-4891: An index should inherit UPDATE_CACHE_FREQUENCY setting from parent table

ChinmaySKulkarni:
**(WIP Patch)**
Changes in this commit:
- When creating a local or global index on a physical table/view, we set the UPDATE_CACHE_FREQUENCY value from the base table/view
- When altering the U_C_F value of a base table/view, we alter the same for all of its indexes
- Since we want this property to be in sync amongst tables and their indexes, I am currently disallowing setting U_C_F when creating an index _AND_ disallowing altering this property for an index

**TODO/Discussion:**
- Handle the case of a client connecting to a server with an old timestamp of SYSCAT i.e. the upgrade code-path. Should we sync the U_C_F property amongst tables/views and their indexes? **OR**, should we instead allow altering U_C_F for indexes and make this the user's responsibility to manually do this as they need?
- Currently, when a view is created, it inherits the U_C_F value of the base table already. Thus, indexes created on the view will also inherit the same value. However, we can later alter the U_C_F of a view (which will be propagated to its indexes). I think this is desired behavior, and a view need not have the same U_C_F value as its base table throughout its lifetime (though the index of a view should be in sync with the view), but open to other suggestions.


ChinmaySKulkarni:
@lhofhansl @twdsilva please take a look. I have a couple of discussion points as well.

phoenix-core/src/it/java/org/apache/phoenix/rpc/UpdateCacheIT.java
```diff
@@ -136,14 +137,55 @@ public void testUpdateCacheForChangingUpdateTable() throws Exception {
...
+
+        // Now alter the UPDATE_CACHE_FREQUENCY value of the base table
+        try (Connection conn = DriverManager.getConnection(getUrl(), props)) {
+            conn.createStatement()
+                    .execute("ALTER TABLE " + fullTableName + " SET UPDATE_CACHE_FREQUENCY=ALWAYS");
+        }
+        // Even the indexes should now have the modified value of UPDATE_CACHE_FREQUENCY
+        // Note that when we query the base table, during query plan generation, we make 2 getTable
+        // requests (to retrieve the base table) for each index of the base table
+        helpTestUpdateCache(fullTableName, new int[] {1, 15}, false);
```
twdsilva: why is the number of rps 15 when we read from the table?

twdsilva:
@ChinmaySKulkarni  I don't think we will ALTER TABLE UPDATE_CACHE_FREQUENCY during SYSTEM.CATALOG upgrade. What will currently happen if we do run this statement during the upgrade?
The index on view behavior that you describe sounds correct to me.

phoenix-core/src/it/java/org/apache/phoenix/rpc/UpdateCacheIT.java
ChinmaySKulkarni: @twdsilva when we query the base table, during query plan generation, when we evaluate the query plan for each index of the base table, we make 2 `getTable` requests (to get the base table itself for each index of the base table). Hence, since we have 2 indexes created on this base table, we will make 2* (number of indexes) + 1 (for the base table itself) = 5 calls. Since `helpTestUpdateCache` does 3 select calls, the value to expect is 15.

It is another issue/discussion/possible improvement to not do the extra `getTable` calls when evaluating the query plan for an index, but this is consistent with the currently observed behavior.

ChinmaySKulkarni:
@twdsilva More than issuing an ALTER TABLE for each index, I was talking about adopting a similar approach to how we had done the property syncing in the upgrade codepath in PHOENIX-3955; i.e. I can force the index table U_C_F value to be the same as its base table via altering corresponding rows cells in SYSCAT.

ChinmaySKulkarni:
@twdsilva I have added changes for the upgrade path and also added a test for the same. Please take a look. Once I get your +1, I will squash and rebase my changes. Thanks!

phoenix-core/src/it/java/org/apache/phoenix/end2end/PropertiesInSyncIT.java
```diff
@@ -402,16 +486,17 @@ private String createBaseTableWithProps(Connection conn) throws SQLException {
      * @param baseTableName Name of the HBase base table on which to create an index
      * @param indexType LOCAL, GLOBAL or if we pass in null as the indexType,
      *                 we create a view and an index on that view for the given base table
+     * @param returnLocalIndex if true, when creating a local index, return the local index name
+     *                         of the base table name instead of the physical base table name
      * @return The physical HBase table corresponding to the index created
```
twdsilva: This doesn't always return the physical HBase table that stores the index being created right? Might be a cleaner if you return a pair of the index name and the index physical table name.

phoenix-core/src/main/java/org/apache/phoenix/util/UpgradeUtil.java
```diff
@@ -151,6 +152,14 @@
             + "(TENANT_ID, TABLE_SCHEM, TABLE_NAME, COLUMN_NAME, COLUMN_FAMILY, BASE_COLUMN_COUNT) "
             + "VALUES (?, ?, ?, ?, ?, ?) ";
 
+    public static final String UPSERT_UPDATE_CACHE_FREQUENCY_FOR_INDEXES =
```
twdsilva: nit: rename to UPSERT_UPDATE_CACHE_FREQUENCY

ChinmaySKulkarni:
@twdsilva Addressed review commits, but a force push closed this PR. Please merge https://github.com/apache/phoenix/pull/364 Thanks

PR #397, opened 2018-10-27
PHOENIX-4997 Phoenix MR on snapshots can produce duplicate rows

karanmehta93:
@akshita-malhotra @gjacoby126 
Could you please review?

### DIFF
--- a/phoenix-core/src/it/java/org/apache/phoenix/end2end/TableSnapshotReadsMapReduceIT.java
+++ b/phoenix-core/src/it/java/org/apache/phoenix/end2end/TableSnapshotReadsMapReduceIT.java
@@ -36,6 +36,7 @@ import java.util.UUID;
 
 import org.apache.hadoop.conf.Configuration;
 import org.apache.hadoop.fs.Path;
+import org.apache.hadoop.hbase.HRegionInfo;
 import org.apache.hadoop.hbase.TableName;
 import org.apache.hadoop.hbase.client.Admin;
 import org.apache.hadoop.hbase.client.SnapshotDescription;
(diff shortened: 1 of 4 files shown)


phoenix-core/src/it/java/org/apache/phoenix/end2end/TableSnapshotReadsMapReduceIT.java
```diff
@@ -66,74 +74,62 @@
...
 
   @BeforeClass
   public static void doSetup() throws Exception {
       Map<String,String> props = Maps.newHashMapWithExpectedSize(1);
       setUpTestDriver(new ReadOnlyProps(props.entrySet().iterator()));
   }
 
-  @Test
-  public void testMapReduceSnapshots() throws Exception {
+  @Before
```
karanmehta93: Refactoring done to move common code to before method.

phoenix-core/src/it/java/org/apache/phoenix/end2end/TableSnapshotReadsMapReduceIT.java
```diff
@@ -191,31 +192,52 @@ private void upsertData(PreparedStatement stmt, String field1, String field2, in
...
+
     // upsert data after snapshot
     PreparedStatement stmt = conn.prepareStatement(String.format(UPSERT, tableName));
-    upsertData(stmt, "DDDD", "SNFB", 0004);
+    upsertData(stmt, "DDDD", "SNFB", 45);
     conn.commit();
   }
 
-    public void deleteSnapshot(String tableName) throws Exception {
+  private void splitTableSync(Admin admin, TableName hbaseTableName,
```
karanmehta93: Didn't use the method provided in BaseTest.java since enabling/disabling the table causes the snapshot manifest file to not contain the offline regions. 
The bug is caused because the snapshot manifest file on a regular production table can contain information about regions that are split or offline, which should ideally be ignored when restoring snapshot.

karanmehta93:
@BinShi-SecularBird

phoenix-core/src/main/java/org/apache/phoenix/iterate/MapReduceParallelScanGrouper.java
```diff
@@ -80,18 +83,39 @@ public boolean shouldStartNewScan(QueryPlan plan, List<Scan> scans,
...
+		for (RegionInfo regionInfo : regionInfos) {
+			hRegionLocations.add(new HRegionLocation(regionInfo, null));
 		}
 
-		return regionLocations;
+		return hRegionLocations;
+	}
+
+	// Exclude offline split parent regions
+	private boolean isValidRegion(RegionInfo hri) {
```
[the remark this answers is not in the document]
karanmehta93: I thought about it, however they don't have any classes in common and the implementations of these methods are slightly different. I agree that this is duplicate code and should not exist this way but I couldn't think of a better way for this.

gjacoby126:
lgtm.

karanmehta93:
Thanks @twdsilva and @gjacoby126 for the review. Will push it soon.

PR #398, opened 2018-10-28
PHOENIX-4998 The estimated instance size of GuidePostsInfo doesn't co…

BinShi-SecularBird:
…unt the last update timestamp array.

### DIFF
--- a/phoenix-core/src/main/java/org/apache/phoenix/schema/stats/GuidePostsInfo.java
+++ b/phoenix-core/src/main/java/org/apache/phoenix/schema/stats/GuidePostsInfo.java
@@ -95,15 +95,17 @@ public class GuidePostsInfo {
         this.guidePostsCount = guidePostsCount;
         this.rowCounts = Longs.toArray(rowCounts);
         this.byteCounts = Longs.toArray(byteCounts);
-        int estimatedSize = SizedUtil.OBJECT_SIZE 
+        this.gpTimestamps = Longs.toArray(updateTimes);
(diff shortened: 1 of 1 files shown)


BinShi-SecularBird:
@karanmehta93

karanmehta93:
It would be good to add a unit test for the size estimator function. We would know that we need to change how we estimate when we end up changing the data structure for storing these.

BinShi-SecularBird:
@karanmehta93, I thought about it, but I eventually gave up for the reasons listed below:
1. If we wrote this unit test, the unit test would look like in two ways
    a. In unit test, check GuidePostsInfo.GetEstimatedSize() is equal to a constant value or a value calculated in the ways mentioned in https://dzone.com/articles/estimating-java-object-sizes or https://github.com/twitter/commons/blob/master/src/java/com/twitter/common/objectsize/ObjectSizeCalculator.java. The former is useless if the developer forgot to change "int estimatedSize = ..." in GuidePostsInfo's constructor, and the latter can only get an approximation of the object size.
    b. In unit test, list and check all the fields in GuidePostInfo are equal to what we have now. It seems overkilled and weird.
2. We need to refactor GuidePostsInfo in the near future. I want to postpone how to calculate its object size and the unit test to that time.

Please let me know if you have better idea.

karanmehta93:
That makes sense. @BinShi-SecularBird 
`Those Java equivalents of sizeof() in C/C++, mentioned on the Web, might be overkilled here`
Can you provide some details on this comment?

PR #399, opened 2018-10-30
PHOENIX-4764 Cleanup metadata of child views for a base table that ha…

kadirozde:
…s been dropped (addendum)

### DIFF
--- a/phoenix-core/src/it/java/org/apache/phoenix/end2end/index/MutableIndexIT.java
+++ b/phoenix-core/src/it/java/org/apache/phoenix/end2end/index/MutableIndexIT.java
@@ -56,6 +56,7 @@ import org.apache.phoenix.query.QueryServices;
 import org.apache.phoenix.schema.PIndexState;
 import org.apache.phoenix.schema.PTableKey;
 import org.apache.phoenix.util.EnvironmentEdgeManager;
+import org.apache.phoenix.util.IndexScrutiny;
 import org.apache.phoenix.util.IndexUtil;
 import org.apache.phoenix.util.PhoenixRuntime;
 import org.apache.phoenix.util.PropertiesUtil;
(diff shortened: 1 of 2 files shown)


phoenix-core/src/it/java/org/apache/phoenix/monitoring/PhoenixMetricsIT.java
```diff
@@ -232,6 +232,7 @@ private boolean verifyMetricsFromSinkOnce(Map<String, Long> expectedMetrics) {
                         long actualValue = metric.value().longValue();
                         if (expectedValue != actualValue) {
                             LOG.warn("Metric from Hadoop Sink: " + metric.name() + " didn't match expected.");
+                            LOG.warn("Expected: " + expectedValue + " Actual: " + actualValue);
```
karanmehta93: You can remove this log line or merge it into previous one, otherwise it will be difficult to comprehend when all these log lines are together.

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/TaskRegionObserver.java
```diff
@@ -71,6 +71,11 @@
     @GuardedBy("TaskRegionObserver.class")
     // initial delay before the first task is handled
     private static final long  initialDelay = 10000; // 10 secs
+    private static boolean disableTaskHandling = false;
```
karanmehta93: You still have the `initialDelay` parameter, you could leverage it by exposing out as you did earlier and set it to `Integer.MAX_VALUE`. That approach is followed for the other rebuilder thread task. 
OR
expose a parameter out that disables the complete framework, not just for test purposes. You could use set that parameter to false in tests, defaults to true.

kadirozde: Will do it

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/TaskRegionObserver.java
```diff
@@ -70,7 +70,11 @@
     private long timeMaxInterval = QueryServicesOptions.DEFAULT_TASK_HANDLING_MAX_INTERVAL_MS;
     @GuardedBy("TaskRegionObserver.class")
     // initial delay before the first task is handled
-    private static final long  initialDelay = 10000; // 10 secs
+    private static long initialDelay = 10000; // 10 secs
+
+    public static void disableTaskHandling() {
```
karanmehta93: We should make this method "package-private".

twdsilva: If its package-private then we won't be able to call it from the BasePhoenixMetricsIT.java right? Generally we usually have a config option to set parameters such as these (for example see QueryServices.INDEX_REBUILD_TASK_INITIAL_DELAY). Then you would put it in the server properties and call setUpTestDriver(ReadOnlyProps serverProps, ReadOnlyProps clientProps)

PR #403, opened 2018-11-10
PHOENIX-5008: CQSI.init should not bubble up RetriableUpgradeException to client in case of an UpgradeRequiredException

ChinmaySKulkarni:


phoenix-core/src/it/java/org/apache/phoenix/end2end/SystemCatalogCreationOnConnectionIT.java
```diff
@@ -428,14 +433,78 @@ public void testMetadataAlterRemainsAutoUpgradeEnabled() throws Exception {
         assertEquals(Integer.parseInt(MODIFIED_MAX_VERSIONS), verifyModificationTableMetadata(driver, PHOENIX_SYSTEM_CATALOG));
     }
 
+    // Test the case when an end-user uses the vanilla PhoenixDriver to create a connection and a
+    // requirement for upgrade is detected. In this case, the user should get a connection on which
+    // they are only able to run "EXECUTE UPGRADE"
+    @Test
+    public void testExecuteUpgradeSameConnWithPhoenixDriver() throws Exception {
+        // Register the vanilla PhoenixDriver
+        DriverManager.registerDriver(PhoenixDriver.INSTANCE);
```
twdsilva: Can we just move this to @BeforeClass so that all the tests run with the real driver?

twdsilva: Spoke with @ChinmaySKulkarni  offline, looks like we need to run the other tests with a test driver to verify execute upgrade is called the expected number of times. 
+1 LGTM

ChinmaySKulkarni: Thanks for the review @twdsilva.

PR #406, opened 2018-11-21
PHOENIX-4765 Add client and server side config property to enable rol…

twdsilva:
…lback of splittable System Catalog if required

@ChinmaySKulkarni  can you please review. This PR adds a config that is used to enable rollback of splittable SYSTEM.CATALOG if required. If the config is true, we continue storing the parent table column metadata while creating a view. We also block any add columns and while creating a view try to lock the parent table to prevent conflicting concurrent changes.

### DIFF
--- a/phoenix-core/src/main/java/org/apache/phoenix/coprocessor/MetaDataEndpointImpl.java
+++ b/phoenix-core/src/main/java/org/apache/phoenix/coprocessor/MetaDataEndpointImpl.java
@@ -530,6 +530,10 @@ public class MetaDataEndpointImpl extends MetaDataProtocol implements Coprocesso
     private int maxIndexesPerTable;
     private boolean isTablesMappingEnabled;
 
+    // this flag denotes that we will continue to write parent table column metadata while creating
+    // a child view and also block metadata changes that were previously propagated to children
+    // before 4.15, so that we can rollback the upgrade to 4.15 if required
+    private boolean allowSystemCatalogRollback;
 
     /**
      * Stores a reference to the coprocessor environment provided by the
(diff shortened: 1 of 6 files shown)


twdsilva:
@karanmehta93  Can you please review?

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/MetaDataEndpointImpl.java
```diff
@@ -2150,11 +2156,21 @@ public void createTable(RpcController controller, CreateTableRequest request,
                 ImmutableBytesPtr parentCacheKey = null;
                 PTable parentTable = null;
                 if (parentTableName != null) {
-                    // we lock the parent table when creating an index on a table or a view
-                    if (tableType == PTableType.INDEX) {
+                    // From 4.15 onwards we only need to lock the parent table :
+                    // 1) when creating an index on a table or a view
+                    // 2) if allowSystemCatalogRollback is true we try to lock the parent table to prevent it
```
ChinmaySKulkarni: @twdsilva don't we need to have a way to figure out if a rollback is in progress and only acquire the lock if it is currently ongoing, rather than be based on the value of *allowSystemCatalogRollback* alone?

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/MetaDataEndpointImpl.java
```diff
@@ -530,6 +530,10 @@ private static PName newPName(byte[] keyBuffer, int keyOffset, int keyLength) {
     private int maxIndexesPerTable;
     private boolean isTablesMappingEnabled;
 
+    // this flag denotes that we will continue to write parent table column metadata while creating
+    // a child view and also block metadata changes that were previously propagated to children
+    // before 4.15, so that we can rollback the upgrade to 4.15 if required
+    private boolean allowSystemCatalogRollback;
```
ChinmaySKulkarni: @twdsilva don't we need to have a way to figure out if a rollback is currently in progress as well? Once the rollback has already happened and the SYSTEM.CATALOG timestamp is < the min_splittable_SYSCAT timestamp, can we then acquire locks only when the SYSCAT timestamp is still smaller AND **allowedSystemCatalogRollback** is true, instead of always acquiring locks based on just the value of this new flag?

ChinmaySKulkarni: Had a conversation offline with Thomas and the underlying point in my comment above is already being addressed. Thanks @twdsilva.

ChinmaySKulkarni:
Once flipped back to "false", setting this flag back to "true" may not cleanly rollback since SYSCAT may have split since then. Let's add a comment about that as well as the SYSTEM.CHILD_LINK metadata that we discussed. Otherwise +1. Thanks @twdsilva

PR #408, opened 2018-11-30
PHOENIX-4763: Changing a base table property value should be reflected in child views (if the property wasn't changed)

ChinmaySKulkarni:
[WIP]
- When a view alters a property, add a tag for the corresponding cell in SYSCAT
- Introduce a bitset to keep track of these properties in PTable
- Add tests
- TODO:
1. Handle the case where we "CREATE VIEW <>" with properties specified
2. Add unit tests for tagging changes
3. Refactoring change for PHOENIX-4951 as part of this patch itself.

### DIFF
--- a/phoenix-core/src/main/java/org/apache/phoenix/coprocessor/MetaDataEndpointImpl.java
+++ b/phoenix-core/src/main/java/org/apache/phoenix/coprocessor/MetaDataEndpointImpl.java
@@ -87,7 +87,6 @@ import java.security.PrivilegedExceptionAction;
 import java.sql.ResultSetMetaData;
 import java.sql.SQLException;
 import java.sql.Statement;
-import java.sql.Timestamp;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
@@ -109,11 +108,13 @@ import org.apache.hadoop.hbase.CellComparatorImpl;
 import org.apache.hadoop.hbase.CellUtil;
 import org.apache.hadoop.hbase.CoprocessorEnvironment;
 import org.apache.hadoop.hbase.DoNotRetryIOException;
+import org.apache.hadoop.hbase.ExtendedCellBuilder;
 import org.apache.hadoop.hbase.HConstants;
 import org.apache.hadoop.hbase.KeyValue;
 import org.apache.hadoop.hbase.KeyValue.Type;
 import org.apache.hadoop.hbase.KeyValueUtil;
 import org.apache.hadoop.hbase.TableName;
+import org.apache.hadoop.hbase.TagUtil;
 import org.apache.hadoop.hbase.client.Delete;
 import org.apache.hadoop.hbase.client.Get;
 import org.apache.hadoop.hbase.client.Mutation;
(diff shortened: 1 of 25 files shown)


ChinmaySKulkarni:
@twdsilva Please review. I will rebase changes once the approach looks good.

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/MetaDataEndpointImpl.java
```diff
@@ -2340,7 +2365,7 @@ public void createTable(RpcController controller, CreateTableRequest request,
                 // 3. Finally write the mutations to create the table
 
                 // From 4.15 the parent->child links are stored in a separate table SYSTEM.CHILD_LINK
-                // TODO remove this after PHOENIX-4763 is implemented
+                // TODO remove this after PHOENIX-4763 is implemented -> Remove below?
```
ChinmaySKulkarni: @twdsilva Is this comment just to keep track of the fact that we need to remove this after all of splittable SYSCAT is implemented or is there anything specific to PHOENIX-4763?

ChinmaySKulkarni:
- Added cell tagging change in 'CREATE VIEW' with properties path
- Added tests
- Added refactoring as per PHOENIX-4951
@twdsilva please review. Thanks!

phoenix-core/src/main/java/org/apache/phoenix/util/MetaDataUtil.java
```diff
@@ -353,6 +403,18 @@ public static int getSaltBuckets(List<Mutation> tableMetaData, KeyValueBuilder b
         }
         return 0;
     }
+
```
twdsilva: remove these two functions if they aren't used

phoenix-core/src/test/java/org/apache/phoenix/util/MetaDataUtilTest.java
```diff
@@ -65,21 +100,67 @@ public void testCompatibility() {
...
-        assertEquals(Bytes.toString(newValue), Bytes.toString(cell.getValueArray(), cell.getValueOffset(), cell.getValueLength()));
+        Cell cell = put.get(TABLE_FAMILY_BYTES, QUALIFIER).get(0);
+        assertEquals(Bytes.toString(ORIGINAL_VALUE),
+                Bytes.toString(cell.getValueArray(), cell.getValueOffset(), cell.getValueLength()));
+        MetaDataUtil.mutatePutValue(put, TABLE_FAMILY_BYTES, QUALIFIER, newValue);
+        cell = put.get(TABLE_FAMILY_BYTES, QUALIFIER).get(0);
+        assertEquals(Bytes.toString(newValue),
+                Bytes.toString(cell.getValueArray(), cell.getValueOffset(), cell.getValueLength()));
+    }
+
```
twdsilva: Nice unit tests.

phoenix-core/src/main/java/org/apache/phoenix/util/MetaDataUtil.java
ChinmaySKulkarni: good catch. Will remove

ChinmaySKulkarni:
Thanks for the review @twdsilva

PR #409, opened 2018-12-04
PHOENIX-5034 Log all critical statements in SYSTEM.LOG table

xcangCRM:
@karanmehta93

### DIFF
--- a/phoenix-client/pom.xml
+++ b/phoenix-client/pom.xml
@@ -57,12 +57,9 @@
       <plugin>
         <groupId>org.apache.maven.plugins</groupId>
         <artifactId>maven-jar-plugin</artifactId>
-        <executions>
-          <execution>
-            <id>default-jar</id>
-            <phase>none</phase>
-          </execution>
-        </executions>
+        <configuration>
+          <finalName>phoenix-${project.version}-client</finalName>
+        </configuration>
       </plugin>
       <plugin>
         <groupId>org.apache.maven.plugins</groupId>
--- a/phoenix-server/pom.xml
+++ b/phoenix-server/pom.xml
@@ -61,12 +61,9 @@
       <plugin>
         <groupId>org.apache.maven.plugins</groupId>
         <artifactId>maven-jar-plugin</artifactId>
-        <executions>
-          <execution>
-            <id>default-jar</id>
-            <phase>none</phase>
-          </execution>
-        </executions>
+        <configuration>
+          <finalName>phoenix-${project.version}-server</finalName>
+        </configuration>
       </plugin>
       <plugin>
         <groupId>org.apache.maven.plugins</groupId>


phoenix-core/src/it/java/org/apache/phoenix/end2end/QueryLoggerIT.java
```diff
@@ -250,7 +250,44 @@ public void testWithLoggingOFF() throws Exception{
         assertFalse(foundQueryLog);
         conn.close();
     }
-    
+
+    @Test
```
karanmehta93: Please add a test for `Alter Table` too.

phoenix-core/src/it/java/org/apache/phoenix/end2end/QueryLoggerIT.java
```diff
@@ -250,7 +250,44 @@ public void testWithLoggingOFF() throws Exception{
...
+        while (rs.next()) {
+            rs.getString(1);
+            rs.getString(2);
+        }
+        String logQuery = "SELECT * FROM " + SYSTEM_CATALOG_SCHEMA + ".\"" + SYSTEM_LOG_TABLE
+            + "\"";
+        int delay = 5000;
+
+        // sleep for sometime to let query log committed
+        Thread.sleep(delay);
```
karanmehta93: Instead of just waiting once and checking, use a retry logic for 10 times and wait 1 sec for each iteration. This will make tests more reliable and possibly faster (if it syncs fast)

phoenix-core/src/main/java/org/apache/phoenix/jdbc/PhoenixPreparedStatement.java
```diff
@@ -170,7 +170,7 @@ boolean execute(boolean batched) throws SQLException {
             .build().buildException();
         }
         if (statement.getOperation().isMutation()) {
-            executeMutation(statement);
+            executeMutation(statement, createQueryLogger(statement,query));
```
karanmehta93: nit: spaces after commas (applicable everywhere)

phoenix-core/src/main/java/org/apache/phoenix/log/QueryLogger.java
```diff
@@ -92,9 +92,17 @@ public boolean isSynced(){
...
-                .nextDouble() > connection.getLogSamplingRate()) { return NO_OP_INSTANCE; }
+    public static QueryLogger getInstance(PhoenixConnection connection, boolean isSystemTable,
+        boolean criticalStatement) {
+        // always log critical statements (DROP,ALTER for now)
+        // do not log anything when loglevel is off.
+        // do not log systemTable statement.
+        // do sampling on other statements based on configured perncentage, 1% by default.
+        if ((connection.getLogLevel() == LogLevel.OFF || isSystemTable
+            || ThreadLocalRandom.current().nextDouble() > connection.getLogSamplingRate()) &&
+            !criticalStatement) {
```
karanmehta93: Does this mean that it will log the statements even if the feature is disabled on server side? Can you add a test for that? Do we really want it that way?

phoenix-core/src/main/java/org/apache/phoenix/log/QueryLogger.java
```diff
@@ -92,9 +92,17 @@ public boolean isSynced(){
         }
     };
 
-    public static QueryLogger getInstance(PhoenixConnection connection, boolean isSystemTable) {
-        if (connection.getLogLevel() == LogLevel.OFF || isSystemTable || ThreadLocalRandom.current()
-                .nextDouble() > connection.getLogSamplingRate()) { return NO_OP_INSTANCE; }
+    public static QueryLogger getInstance(PhoenixConnection connection, boolean isSystemTable,
```
karanmehta93: Should we rename this to `createInstance` so that it doesn't feel like we are returning a singleton?

phoenix-core/src/main/java/org/apache/phoenix/log/QueryLogger.java
```diff
@@ -92,9 +92,17 @@ public boolean isSynced(){
...
+        boolean criticalStatement) {
+        // always log critical statements (DROP,ALTER for now)
+        // do not log anything when loglevel is off.
+        // do not log systemTable statement.
+        // do sampling on other statements based on configured perncentage, 1% by default.
```
karanmehta93: nit: perncentage

phoenix-core/src/it/java/org/apache/phoenix/end2end/QueryLoggerIT.java
xcangCRM: Added

phoenix-core/src/it/java/org/apache/phoenix/end2end/QueryLoggerIT.java
xcangCRM: Changed.

phoenix-core/src/main/java/org/apache/phoenix/jdbc/PhoenixPreparedStatement.java
xcangCRM: Fixed

phoenix-core/src/main/java/org/apache/phoenix/log/QueryLogger.java
xcangCRM: Changed, do not log when loglevel is OFF>

phoenix-core/src/main/java/org/apache/phoenix/log/QueryLogger.java
xcangCRM: Good idea, changed.

phoenix-core/src/main/java/org/apache/phoenix/log/QueryLogger.java
xcangCRM: fixed. Good eye.

PR #410, opened 2018-12-04
PHOENIX-5034 Log all critical statements in SYSTEM.LOG table

xcangCRM:


phoenix-core/src/it/java/org/apache/phoenix/end2end/QueryLoggerIT.java
```diff
@@ -250,7 +250,103 @@ public void testWithLoggingOFF() throws Exception{
...
+        boolean foundQueryLog = false;
+
+        int retryCount = 0, maxRetryCount = 10;
+        int sleepTime = 1000;
+        while (retryCount++ < maxRetryCount) {
+            Thread.sleep(sleepTime);
+            try {
+                rs = conn.createStatement().executeQuery(logQuery);
+                while (rs.next()) {
+                    rs.getString(QUERY);
```
karanmehta93: nit: not needed

phoenix-core/src/it/java/org/apache/phoenix/end2end/QueryLoggerIT.java
```diff
@@ -250,7 +250,103 @@ public void testWithLoggingOFF() throws Exception{
         assertFalse(foundQueryLog);
         conn.close();
     }
-    
+
+    @Test
+    //DROP Table statement should be always logged.
+    public void testDropTableStatement() throws Exception{
```
karanmehta93: Sorry for late comment but you can actually merge both the tests into on.

phoenix-core/src/main/java/org/apache/phoenix/jdbc/PhoenixStatement.java
```diff
@@ -1774,11 +1778,20 @@ public QueryLogger createQueryLogger(CompilableStatement stmt, String sql) throw
...
+            isCriticalStatement(stmt));
+        QueryLoggerUtil.logInitialDetails(queryLogger, connection.getTenantId(), 
                 connection.getQueryServices(), sql, getParameters());
+        if (isCriticalStatement(stmt)) {
+            queryLogger.sync(null, null);
+        }
         return queryLogger;
     }
+
+    private boolean isCriticalStatement(CompilableStatement stmt) {
```
karanmehta93: On a second thought, How about logging every DML that is not a "Update statement"? 
It might be an overkill if we have people creating tons of views though, but we could always drop those as well.

phoenix-core/src/main/java/org/apache/phoenix/log/QueryLogger.java
```diff
@@ -92,9 +92,20 @@ public boolean isSynced(){
...
-    public static QueryLogger getInstance(PhoenixConnection connection, boolean isSystemTable) {
-        if (connection.getLogLevel() == LogLevel.OFF || isSystemTable || ThreadLocalRandom.current()
-                .nextDouble() > connection.getLogSamplingRate()) { return NO_OP_INSTANCE; }
+    public static QueryLogger createInstance(PhoenixConnection connection, boolean isSystemTable,
+        boolean criticalStatement) {
+        // always log critical statements (DROP,ALTER for now)
+        // do not log anything when loglevel is off.
+        // do not log systemTable statement.
+        // do sampling on other statements based on configured percentage, 1% by default.
+        if (connection.getLogLevel() == LogLevel.OFF) {
```
karanmehta93: Good work on separating out these conditions into two different ones. It looks cleaner now.

phoenix-core/src/main/java/org/apache/phoenix/jdbc/PhoenixStatement.java
karanmehta93: Also, as discussed offline, we might not want to log `DropTableStatement` which are actually `DROP VIEW` since we can have lot of those.

phoenix-core/src/it/java/org/apache/phoenix/end2end/QueryLoggerIT.java
karanmehta93: Also can you test a scenario with multiple drop/alter statements issued together. What happens if the RingBuffer is full and we are trying to add to it, does it block and wait OR it drops it (in that case this is not complete)?

PR #417, opened 2018-12-15
PHOENIX-4820

comnetwork:
PHOENIX-4820

### DIFF
--- a/phoenix-core/src/main/java/org/apache/phoenix/compile/OrderByCompiler.java
+++ b/phoenix-core/src/main/java/org/apache/phoenix/compile/OrderByCompiler.java
@@ -88,7 +88,8 @@ public class OrderByCompiler {
                                   Integer offset,
                                   RowProjector rowProjector,
                                   TupleProjector tupleProjector,
-                                  boolean isInRowKeyOrder) throws SQLException {
+                                  boolean isInRowKeyOrder,
+                                  Expression whereExpression) throws SQLException {
         List<OrderByNode> orderByNodes = statement.getOrderBy();
         if (orderByNodes.isEmpty()) {
             return OrderBy.EMPTY_ORDER_BY;
@@ -105,9 +106,22 @@ public class OrderByCompiler {
         } else {
             compiler = new ExpressionCompiler(context, groupBy);
         }
+
+        if(groupBy != GroupBy.EMPTY_GROUP_BY) {
+            //if there is groupBy,the groupBy.expressions are viewed as new rowKey columns,so
+            //tupleProjector and isInRowKeyOrder is cleared
+            tupleProjector = null;
+            isInRowKeyOrder = true;
+        }
         // accumulate columns in ORDER BY
         OrderPreservingTracker tracker = 
-                new OrderPreservingTracker(context, groupBy, Ordering.ORDERED, orderByNodes.size(), tupleProjector);
+                new OrderPreservingTracker(
+                        context,
+                        groupBy,
+                        Ordering.ORDERED,
+                        orderByNodes.size(),
+                        tupleProjector,
+                        whereExpression);
         LinkedHashSet<OrderByExpression> orderByExpressions = Sets.newLinkedHashSetWithExpectedSize(orderByNodes.size());
         for (OrderByNode node : orderByNodes) {
             ParseNode parseNode = node.getNode();
--- a/phoenix-core/src/main/java/org/apache/phoenix/expression/Expression.java
+++ b/phoenix-core/src/main/java/org/apache/phoenix/expression/Expression.java
@@ -88,4 +88,10 @@ public interface Expression extends PDatum, Writable {
      * @return
      */
     boolean requiresFinalEvaluation();
+
+    /**
+     * Determines if expression is constant if all children of it are constants.
+     * @return
+     */
+    boolean isConstantIfChildrenAllConstant();
 }
--- a/phoenix-core/src/main/java/org/apache/phoenix/util/ExpressionUtil.java
+++ b/phoenix-core/src/main/java/org/apache/phoenix/util/ExpressionUtil.java
@@ -10,12 +10,20 @@
 package org.apache.phoenix.util;
 
 import java.sql.SQLException;
+import java.util.Iterator;
 import java.util.List;
 
+import org.apache.hadoop.hbase.filter.CompareFilter.CompareOp;
 import org.apache.hadoop.hbase.io.ImmutableBytesWritable;
+import org.apache.phoenix.expression.AndExpression;
+import org.apache.phoenix.expression.ColumnExpression;
+import org.apache.phoenix.expression.ComparisonExpression;
 import org.apache.phoenix.expression.Determinism;
 import org.apache.phoenix.expression.Expression;
+import org.apache.phoenix.expression.IsNullExpression;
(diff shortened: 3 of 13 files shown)


phoenix-core/src/main/java/org/apache/phoenix/expression/Expression.java
```diff
@@ -88,4 +88,10 @@
      * @return
      */
     boolean requiresFinalEvaluation();
+
```
twdsilva: As James said before its not clear to me why you need this new function, can just use ExpressionUtil.isConstant() ?

Your implementation returns true by default and is false only for AggregateFunction, RandomFunction and SequenceValueExpression for which ExpressionUtil.isConstant() should also return false, right?

phoenix-core/src/main/java/org/apache/phoenix/compile/OrderByCompiler.java
```diff
@@ -88,7 +88,8 @@ public static OrderBy compile(StatementContext context,
                                   Integer offset,
                                   RowProjector rowProjector,
                                   TupleProjector tupleProjector,
-                                  boolean isInRowKeyOrder) throws SQLException {
+                                  boolean isInRowKeyOrder,
+                                  Expression whereExpression) throws SQLException {
```
twdsilva: Can you explain why the whereExpression is needed?

phoenix-core/src/main/java/org/apache/phoenix/expression/Expression.java
comnetwork: ExpressionUtil.isConstant(Expression) is not suitable for OrderPreservingTracker.IsConstantVisitor, because ExpressionUtil.isConstant(Expression) depends on Expression.isStateless() and Expression.getDeterminism(). Expression isStateless() is to check if the expression is depend on the server state, even for RowKeyColumnExpression , isStateless() is false. What we want to check is if a expression is constant when all children of it are constants, just consider following sql:

select a.ak3 from
order by a.ak3,a.av1

Obviously , because of rand(), the Determinism of expression a.ak1 is Determinism.PER_INVOCATION, so for expression "CASE WHEN coalesce(a.ak1,1) > coalesce(a.av2,2) THEN coalesce(a.ak1,1) ELSE coalesce(a.av2,2) END", the determinism is Determinism.PER_INVOCATION and isStateless is false , but because the a.ak1 and a.av2 are both constants in where clause of outer query, we can regard "CASE WHEN coalesce(a.ak1,1) > coalesce(a.av2,2) THEN coalesce(a.ak1,1) ELSE coalesce(a.av2,2) END" as constant in IsConstantVisitor.

BTW. The value returned by ExpressionUtil.isConstant() is complicated and runtime related because of Expression.getDeterminism(). In following BaseCompoundExpression.init method, determinism is the combine of its children.
` private void init(List<Expression> children) {
        this.children = ImmutableList.copyOf(children);
        boolean isStateless = true;
        boolean isNullable = false;
        boolean requiresFinalEvaluation = false;
        this.determinism = Determinism.ALWAYS;
        for (int i = 0; i < children.size(); i++) {
            Expression child = children.get(i);
            isNullable |= child.isNullable();
            isStateless &= child.isStateless();
            this.determinism = this.determinism.combine(child.getDeterminism());
            requiresFinalEvaluation |= child.requiresFinalEvaluation();
        }
        this.isStateless = isStateless;
        this.isNullable = isNullable;
        this.requiresFinalEvaluation = requiresFinalEvaluation;
    }`

phoenix-core/src/main/java/org/apache/phoenix/compile/OrderByCompiler.java
comnetwork: We need the whereExpression because in OrderPreservingTracker.IsConstantVisitor, we need whereExpression to help us to check the columnExpression is constant or not. Just take the same sql as example:  

select a.ak3 from
order by a.ak3,a.av1

We need to check if "CASE WHEN coalesce(a.ak1,1) > coalesce(a.av2,2) THEN coalesce(a.ak1,1) ELSE coalesce(a.av2,2) END,a.av1" is constant in OrderPreservingTracker.hasEqualityConstraints method, so we move forward to check if the a.ak1 and a.av2 are constant in  OrderPreservingTracker.IsConstantVisitor, and we must visit the where clause "where a.ak1 = 0.0 and a.av2 = length(substr('abc',1,1))" to check the a.ak1 and a.av2.

BTW. There is no whereExpression in current code of OrderByCompiler because for PHOENIX-3451 which was completed by James and me before, RowKeyColumnExpression is only supported to check in OrderPreservingTracker.hasEqualityConstraints, and two FIXME tags were left in this method at at time:

 private boolean hasEqualityConstraints(int startPos, int endPos) {
        ScanRanges ranges = context.getScanRanges();
        // If a GROUP BY is being done, then the rows are ordered according to the GROUP BY key,
        // not by the original row key order of the table (see PHOENIX-3451).
        // We check each GROUP BY expression to see if it only references columns that are
        // matched by equality constraints, in which case the expression itself would be constant.
        // FIXME: this only recognizes row key columns that are held constant, not all columns.
        // FIXME: we should optimize out any GROUP BY or ORDER BY expression which is deemed to
        // be held constant based on the WHERE clause.
        if (!groupBy.isEmpty()) {

In this patch, I added whereExpression to try to fix this problem to support non-pk columns also.

twdsilva: Thanks for the explanation.

phoenix-core/src/main/java/org/apache/phoenix/util/ExpressionUtil.java
```diff
@@ -68,4 +76,144 @@ public static boolean isPkPositionChanging(TableRef tableRef, List<Expression> p
...
+        private final Expression columnExpression ;
+        private Expression firstRhsConstantExpression = null;
+        private int rhsConstantCount = 0;
+        private boolean isNullExpressionVisited = false;
+
+        public IsColumnConstantExpressionVisitor(Expression columnExpression) {
+            this.columnExpression = columnExpression;
+        }
+        /**
+         * only conside and,for "where a = 3 or b = 9", neither a or b is constant.
```
twdsilva: nit: only consider

phoenix-core/src/main/java/org/apache/phoenix/util/ExpressionUtil.java
```diff
@@ -68,4 +76,144 @@ public static boolean isPkPositionChanging(TableRef tableRef, List<Expression> p
...
+        @Override
+        public Iterator<Expression> visitEnter(AndExpression andExpression) {
+            if(rhsConstantCount > 1) {
+                return null;
+            }
+            return andExpression.getChildren().iterator();
+        }
+        /**
+         * <pre>
+         * We just conside {@link ComparisonExpression} because:
```
twdsilva: nit: only consider

phoenix-core/src/test/java/org/apache/phoenix/compile/QueryCompilerTest.java
```diff
@@ -2951,6 +2951,129 @@ public void testOrderPreservingGroupBy() throws Exception {
             }
         }
     }
+
```
twdsilva: Nice tests!

stoty:
Already merged.

PR #419, opened 2018-12-20
PHOENIX-4009 Run UPDATE STATISTICS command by using MR integration on…

karanmehta93:
… snapshots


karanmehta93:
_First of all, apologies for loooong PR._ (Most of it is refactoring but still its hard to review)

**Here's the high level idea** 
1. 7 classes were inherited from `StatsCollectorIT`, testing stats collection for different types of table properties. There was a lot of redundancy in the test suite. Also, all the tests were running with namespaces enabled all the time (This is because it is set once for the JVM and we cannot go back without restarting the server). We were controlling the parameterized property for new `PhoenixConnection`, which is disallowed according to documentation.
The code is now refactored to have only 3 classes, 
`NamespaceMappedStatsCollectorIT` --> namespaces enabled, collect stats via snapshots as well as SQL statement
`NonTxStatsCollectorIT` --> mutable/immutable tables, column encoded/non column encoded
`TxStatsCollectorIT` --> mutable/immutable tables, column encoded/non column encoded, TEPHRA/OMID

2. The `StatsCollectorIT` is renamed to `BaseStatsCollectorIT` and tests have been improved to cover certain scenarios. More tests coming along the way.

3. Server side changes:
`DefaultStatisticsCollector` is now an abstract class, RegionServerStatisticsCollector and `MapperStatisticsCollector` are its children. The former is triggered for SQL statements and the latter is used for this Jira (Map Reduce Job). Most of the common code is moved to base class.

4. The snapshot scanner has been improved to collect statistics if the scan is configured accordingly. A `NoOpStatisticsCollector` instance is instantiated if its a regular phoenix MR job on snapshots. 

5. Also have the configuration changes in `PhoenixConfigurationUtil` class.

Finally, `UpdateStatisticsTool` is the tool to launch the MR job.

This is the v1 version for some initial feedback. Please comment wherever its not clear.

**Coming up:** 
More tests covering other scenarios.
Perf testing for sample tables and the results.
Better/useful log lines
General code cleanup for nits

karanmehta93:
@twdsilva @ChinmaySKulkarni @joshelser @chrajeshbabu @dbwong @BinShi-SecularBird

Please review.

ChinmaySKulkarni:
@karanmehta93 can you provide URLs to some design docs/basic stats docs? Will help understand the overall idea much better for someone not so familiar with stats.

karanmehta93:
Design doc: https://salesforce.quip.com/fjMhA2bbBkK4 (is also linked to Jira)
Jira has more details too https://issues.apache.org/jira/browse/PHOENIX-4009

karanmehta93:
@gjacoby126 @vincentpoon If you guys have some time.

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/util/PhoenixConfigurationUtil.java
```diff
@@ -154,7 +154,8 @@
 
     public enum SchemaType {
         TABLE,
-        QUERY;
+        QUERY,
+        UPDATE_STATS
```
twdsilva: I am not very familiar with the TABLE/QUERY enum, whats the difference?

phoenix-core/src/main/java/org/apache/phoenix/parse/UpdateStatisticsStatement.java
```diff
@@ -23,6 +23,7 @@
 
 import java.util.Map;
 
+import org.apache.phoenix.jdbc.PhoenixStatement.Operation;
```
twdsilva: nit: revert these changes

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/util/PhoenixConfigurationUtil.java
karanmehta93: I am also not quite sure about the exact use case. The SchemaType.TABLE enum value is only used by phoenix-hive module. The other option is SchemaType.QUERY which is default option to run regular queries.

karanmehta93:
@twdsilva @dbwong @BinShi-SecularBird Please review.

dbwong:
@karanmehta93  I'll try and take a look tonight.

phoenix-core/src/it/java/org/apache/phoenix/end2end/NamespaceDisabledStatsCollectorIT.java
```diff
@@ -15,24 +15,31 @@
...
 
+import org.apache.phoenix.schema.stats.BaseStatsCollectorIT;
+import org.junit.runners.Parameterized;
+
 import java.util.Arrays;
 import java.util.Collection;
 
-import org.apache.phoenix.schema.stats.StatsCollectorIT;
-import org.junit.runners.Parameterized.Parameters;
+public class NamespaceDisabledStatsCollectorIT extends BaseStatsCollectorIT {
```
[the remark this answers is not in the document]
BinShi-SecularBird: Sorry. It is actually in NamespaceEnabledStatsCollectorIT.java. It is irrelevant to your change, but you might want to  delete the unused import.

phoenix-core/src/it/java/org/apache/phoenix/end2end/NamespaceEnabledStatsCollectorIT.java
```diff
@@ -17,48 +17,48 @@
...
 import org.junit.BeforeClass;
-import org.junit.runners.Parameterized.Parameters;
+import org.junit.runners.Parameterized;
 
-import com.google.common.collect.Maps;
+import java.util.Arrays;
+import java.util.Collection;
+import java.util.Map;
+
+import static org.apache.phoenix.query.QueryServicesOptions.DEFAULT_TASK_HANDLING_INTERVAL_MS;
```
BinShi-SecularBird: unused import, delete it.

phoenix-core/src/it/java/org/apache/phoenix/end2end/NamespaceDisabledStatsCollectorIT.java
```diff
@@ -15,24 +15,31 @@
...
-            boolean userTableNamespaceMapped, boolean columnEncoded) {
-        super(mutable, transactionProvider, userTableNamespaceMapped, columnEncoded);
+    public NamespaceDisabledStatsCollectorIT(boolean userTableNamespaceMapped, boolean collectStatsOnSnapshot) {
+        super(userTableNamespaceMapped, collectStatsOnSnapshot);
     }
 
-    @Parameters(name = "mutable={0},transactionProvider={1},isUserTableNamespaceMapped={2},columnEncoded={3}")
-    public static Collection<Object[]> data() {
+    @Parameterized.Parameters(name = "userTableNamespaceMapped={0},collectStatsOnSnapshot={1}")
+    public static Collection<Object[]> provideData() {
```
BinShi-SecularBird: In the whole Phoenix Code base, I didin't find provideData being used. I might miss something. Do you know how provideData is used?

phoenix-core/src/it/java/org/apache/phoenix/end2end/NamespaceEnabledStatsCollectorIT.java
```diff
@@ -17,48 +17,48 @@
...
 
-    @Parameters(name = "mutable={0},transactionProvider={1},isUserTableNamespaceMapped={2},columnEncoded={3}")
-    public static Collection<Object[]> data() {
-        return TestUtil.filterTxParamData(Arrays.asList(
-            new Object[][] { 
-                { true, "TEPHRA", false, false }, { true, "TEPHRA", false, true }, 
-                { true, "OMID", false, false },
-            }), 1);
+    @Parameterized.Parameters(name = "userTableNamespaceMapped={0},collectStatsOnSnapshot={1}")
+    public static Collection<Object[]> provideData() {
```
BinShi-SecularBird: The same as the above.

phoenix-core/src/it/java/org/apache/phoenix/end2end/NamespaceDisabledStatsCollectorIT.java
```diff
@@ -15,24 +15,31 @@
...
         return Arrays.asList(
-            new Object[][] { { false, null, false, false }, { false, null, true, false } });
+                new Object[][] {
+                        // Collect stats on snapshots using UpdateStatisticsTool
+                        { false, true },
```
BinShi-SecularBird: Why is the second column "collectStatsOnSnapshot" true? In my understanding, this is DisableStats test, so it should be set to false, isn't it?

phoenix-core/src/it/java/org/apache/phoenix/end2end/NonTxStatsCollectorIT.java
```diff
@@ -20,23 +20,29 @@
 import java.util.Arrays;
 import java.util.Collection;
 
-import org.apache.phoenix.schema.stats.StatsCollectorIT;
+import org.apache.phoenix.schema.stats.BaseStatsCollectorIT;
 import org.apache.phoenix.util.TestUtil;
```
BinShi-SecularBird: "import org.apache.phoenix.util.TestUtil;" is unused import. Delete it.

phoenix-core/src/main/java/org/apache/phoenix/iterate/SnapshotScanner.java
```diff
@@ -18,54 +18,83 @@
...
+    this.scan = scan;
 
     RegionCoprocessorEnvironment snapshotEnv = getSnapshotContextEnvironment(conf);
 
-    RegionScannerFactory regionScannerFactory;
-    if (scan.getAttribute(BaseScannerRegionObserver.NON_AGGREGATE_QUERY) != null) {
-      regionScannerFactory = new NonAggregateRegionScannerFactory(snapshotEnv);
+    // Collect statistics during scan if ANALYZE_TABLE attribute is set
+    if (ScanUtil.isAnalyzeTable(scan)) {
+      this.scanner = region.getScanner(scan);
```
BinShi-SecularBird: Assert(scan.getAttribute(BaseScannerRegionObserver.NON_AGGREGATE_QUERY) == null) for updateng statistics

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/util/PhoenixConfigurationUtil.java
BinShi-SecularBird: I checked the code. THE QUERY is generally used for phoenix map reduce jobs. I doubt that we should reuse SchemaType and add UPDATE_STATS here, because now for every place where it checks whether schema type is QUERY you might need to check if it can be applied to UPDATE_STATS too, as UPDATE_STATS is also select statement and MR job. I prefer to add one more new property in configuration to indicate whether this is UPDATE STATS or not instead of adding a new UPDATE_STATS enum type to SchemaType.

phoenix-core/src/main/java/org/apache/phoenix/schema/stats/UpdateStatisticsTool.java
```diff
@@ -0,0 +1,212 @@
...
+
+    private Configuration conf;
+    private String tableName;
+    private String snapshotName;
+    private Path restoreDir;
+    private boolean isForeground;
+
+    @Override
+    public int run(String[] args) throws Exception {
+
```
BinShi-SecularBird: remove the empty line at line 85

phoenix-core/src/main/java/org/apache/phoenix/schema/stats/UpdateStatisticsTool.java
```diff
@@ -0,0 +1,212 @@
...
+ */
+public class UpdateStatisticsTool extends Configured implements Tool {
+
+    private static final Logger LOG = LoggerFactory.getLogger(UpdateStatisticsTool.class);
+
+    private static final Option TABLE_NAME_OPTION = new Option("t", "table", true,
+            "Phoenix Table Name");
+    private static final Option SNAPSHOT_NAME_OPTION = new Option("s", "snapshot", true,
+            "HBase Snapshot Name");
+    private static final Option RESTORE_DIR_OPTION = new Option("d", "restore-dir", true,
```
BinShi-SecularBird: Do we provide default restore directory or root directory?

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/PhoenixRecordReader.java
```diff
@@ -108,6 +109,7 @@ public void initialize(InputSplit split, TaskAttemptContext context) throws IOEx
         final PhoenixInputSplit pSplit = (PhoenixInputSplit)split;
         final List<Scan> scans = pSplit.getScans();
         try {
+            LOG.info("Generating iterators for " + scans.size() + " scans in keyrange: " + pSplit.getKeyRange());
```
BinShi-SecularBird: Shall we remove this log or change it to debug log? The same question for the other two info logs you added below.

phoenix-core/src/main/java/org/apache/phoenix/schema/stats/MapperStatisticsCollector.java
```diff
@@ -0,0 +1,135 @@
...
+public class MapperStatisticsCollector extends DefaultStatisticsCollector {
+
+    private static final Log LOG = LogFactory.getLog(MapperStatisticsCollector.class);
+    private PhoenixConnection connection;
+
+    public MapperStatisticsCollector(PhoenixConnection connection, Configuration conf, Region region, String tableName,
+                                     long clientTimeStamp, byte[] family, byte[] gp_width_bytes,
+                                     byte[] gp_per_region_bytes) {
+        super(conf, region, tableName,
+                clientTimeStamp, family, gp_width_bytes, gp_per_region_bytes);
```
BinShi-SecularBird: Does the above two lines (59, 60) need to be two lines?

phoenix-core/src/main/java/org/apache/phoenix/schema/stats/DefaultStatisticsCollector.java
```diff
@@ -121,92 +110,69 @@
...
+        } catch (SQLException e) {
+            throw new IOException(e);
+        }
+        LOG.info("Initialization complete for " +
+                this.getClass() + " statistics collector for table " + tableName);
+    }
+
+    /**
+     * Determine the GPW for statistics collection for the table.
+     * The order of priority is as follows
```
BinShi-SecularBird: Could you highlight the the order of priority is listed from high to low as follows?

phoenix-core/src/main/java/org/apache/phoenix/schema/stats/DefaultStatisticsCollector.java
```diff
@@ -121,92 +110,69 @@
...
-                        LOG.warn("Failed to close " + htable.getName(), e);
-                    }
-                }
-            }
+            long guidepostWidth = getGuidePostDepthFromSystemCatalog();
             if (guidepostWidth >= 0) {
                 this.guidePostDepth = guidepostWidth;
+                LOG.info("Guide post depth determined from SYSTEM.CATALOG: " + guidePostDepth);
             } else {
                 // Last use global config value
```
BinShi-SecularBird: This comment "// Last use global config value" isn't necessary any more as you have already commented the order of priority on the function.

phoenix-core/src/main/java/org/apache/phoenix/schema/stats/DefaultStatisticsCollector.java
```diff
@@ -220,68 +186,62 @@ public void close() throws IOException {
...
+                    Pair<Long, GuidePostsInfoBuilder> emptyGps = new Pair<Long, GuidePostsInfoBuilder>(0l, new GuidePostsInfoBuilder());
+                    guidePostsInfoWriterMap.put(cfKey, emptyGps);
                 }
             }
-        } catch (IOException e) {
-            logger.error("Failed to update statistics table!", e);
-            throw e;
+        }
+        for (ImmutableBytesPtr fam : fams) {
+            if (delete) {
```
BinShi-SecularBird: from line 228 to line 237, you removed the conditional debug logs and add info log, not very sure this is better and necessary.

BinShi-SecularBird: maybe still use debug log is better.

phoenix-core/src/main/java/org/apache/phoenix/schema/stats/MapperStatisticsCollector.java
```diff
@@ -0,0 +1,135 @@
...
+        this.connection = connection;
+    }
+
+    @Override
+    protected void initStatsWriter() throws IOException, SQLException {
+        this.statsWriter = StatisticsWriter.newWriter(connection, tableName, clientTimeStamp, guidePostDepth);
+    }
+
+    @Override
+    protected long getGuidePostDepthFromSystemCatalog() throws IOException, SQLException {
```
BinShi-SecularBird: MapperStatisticsCollector. getGuidePostDepthFromSystemCatalog() and RegionServerStatisticsCollector.getGuidePostDepthFromSystemCatalog() still have a lot of duplicate code (almost the same), can we avoid duplication in these two functions?

BinShi-SecularBird:
I didn't see my concern "if time since last update is less than certain threshold, the job doesn't need to update the regions' stats again. Suppose one MR job failed but some regions still get updated, the rerun job only needs to update the regions that the first job didn't update." described in the PHOENIX-4009 being addressed in this change, and I still think it should be addressed.

phoenix-core/src/it/java/org/apache/phoenix/end2end/NamespaceDisabledStatsCollectorIT.java
karanmehta93: The disabled here is for namespaces and not stats, hence it is that way.

phoenix-core/src/main/java/org/apache/phoenix/schema/stats/UpdateStatisticsTool.java
karanmehta93: I am not sure if there is an easy way of doing that. Snapshots cannot be restored to `hbase.rootDir`, which is the only configuration parameter I know for getting HDFS path. I will dig up a bit to see if default option is possible here since that will ease its operationalization.

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/PhoenixRecordReader.java
karanmehta93: Record reader is initialized once per mapper job and the map jobs are limited by number of regions. Also these logs will be distributed on several machines and not one. It would be good to have some insight here on the iterators each reader is trying to generate.

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/PhoenixInputFormat.java
```diff
@@ -180,23 +179,39 @@ private QueryPlan getQueryPlan(final JobContext context, final Configuration con
...
+              }
               Preconditions.checkNotNull(selectStatement);
 
               final PhoenixStatement pstmt = statement.unwrap(PhoenixStatement.class);
               // Optimize the query plan so that we potentially use secondary indexes
               final QueryPlan queryPlan = pstmt.optimizeQuery(selectStatement);
               final Scan scan = queryPlan.getContext().getScan();
+
+              if (schemaType == SchemaType.UPDATE_STATS) {
+                  StatisticsUtil.setScanAttributes(scan, null);
```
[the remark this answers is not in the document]
karanmehta93: `scan.setAttribute(RUN_UPDATE_STATS_ASYNC_ATTRIB, FALSE_BYTES);` attribute is applicable fo case when stats are being collected using `UPDATE STATISTICS SQL`. If `false`, the client has to wait till it runs on all the machines. If `true`, the client is given an ACK back instantly and RS continues in the background. Hence it is not applicable here.

phoenix-core/src/main/java/org/apache/phoenix/schema/stats/MapperStatisticsCollector.java
karanmehta93: I agree that it doesn't look clean, do you have any suggestions?

phoenix-core/src/main/java/org/apache/phoenix/schema/stats/DefaultStatisticsCollector.java
karanmehta93: All the log lines were changed to INFO since they will be produced only once per mapper (and a mapper would correspond to a region). These lines would really help us in debugging if something goes wrong. We can also export these as metrics if required.

phoenix-core/src/main/java/org/apache/phoenix/schema/stats/MapperStatisticsCollector.java
karanmehta93: Yes, I agree with you on that point, but it was left as is deliberately. The portions of code are bit complicated and I was not sure if I was covering every case properly. What do you suggest? Shall we refactor it in a future PR?

karanmehta93:
> I didn't see my concern "if time since last update is less than certain threshold, the job doesn't need to update the regions' stats again. Suppose one MR job failed but some regions still get updated, the rerun job only needs to update the regions that the first job didn't update." described in the PHOENIX-4009 being addressed in this change, and I still think it should be addressed.

The first aspect should be fairly simple to add. We can update it as part of this Jira or the PHOENIX-5091. 

The second aspect of re-running only the necessary tasks is bit complicated. The mappers would retry for particular region when they fail (upto limit of max attempts). However I dont really feel the need for that optimization as of now. Let me know what your thoughts are.

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/util/PhoenixConfigurationUtil.java
karanmehta93: I feel that adding a new `SchemaType` is cleaner way to handle this. I agree that everywhere we check for the `SchemaType`, we need to do for this one as well, but we don't have it in many places and it provides an insight to the user about the type of MR job it is. If we add aggregation support in future, this can also be extended similarly. The name `SchemaType` is not really appropriate name here, however I didn't change it since it is public enum. It should be something like `PhoenixMRJobType`.

Adding a config property would not eliminate the checks that we have through the code and I feel that it would make it harder to manipulate in future if new stuff is added. @BinShi-SecularBird Thoughts?

phoenix-core/src/main/java/org/apache/phoenix/schema/stats/MapperStatisticsCollector.java
karanmehta93: Ideally, I want that the instantiation of this class also happens through the `StatisticsCollectorFactory` instead of the direct constructor call in the `SnapshotScanner`. Any ideas on that one?

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/util/PhoenixConfigurationUtil.java
BinShi-SecularBird: My point is that the meanings of QUERY and UPDATE_STATS have overlap -- QUERY indicates Phoenix MR jobs and UPDATE_STAS indicates Phoenix MR jobs plus update statistics, which will bring unclean code to the code base. Thoughts?

phoenix-core/src/it/java/org/apache/phoenix/end2end/NamespaceDisabledStatsCollectorIT.java
BinShi-SecularBird: I see.

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/PhoenixRecordReader.java
BinShi-SecularBird: OK, that's fine then.

phoenix-core/src/main/java/org/apache/phoenix/schema/stats/MapperStatisticsCollector.java
BinShi-SecularBird: I checked MapperStatisticsCollector. getGuidePostDepthFromSystemCatalog() and RegionServerStatisticsCollector.getGuidePostDepthFromSystemCatalog(), it looks like they are exactly the same? Can we just move this function to the base class?

phoenix-core/src/main/java/org/apache/phoenix/schema/stats/UpdateStatisticsTool.java
BinShi-SecularBird: Regarding restore directory, is it a full path or relative path to the root path? Can it be a relative path to ease its operationalization.

phoenix-core/src/main/java/org/apache/phoenix/schema/stats/MapperStatisticsCollector.java
karanmehta93: They look almost same, however there is a subtle difference in how we get `Table` object to access the `SYSTEM.CATALOG` table. This difference arises since mapper has a different way of establishing connection to hbase cluster as compared to region server.

I totally agree with you that more refactoring is better, its just that this is a single unit of code that I didn't want to partition and hence I kept it that way. I understand that it is possible to standardize the access of `Table` api and change it only for different implementations. 

We have a similar issue for `StatisticsWriter` class as well, where we have two separate static methods that instantiate `newWriter`. We could combine them in one and improve there as well, its just something by choice that I haven't done as I felt that we can carry those out in a new Jira.

phoenix-core/src/main/java/org/apache/phoenix/schema/stats/UpdateStatisticsTool.java
karanmehta93: So we cannot use any path relative to `hbase.rootDir` according to the code. However after digging in, I figured out that we can use `fs.defaultFS` config to find the `hdfs.rootDir` and append `/tmp` to it, thus ensuring that we restore it in that directory always. This will help ease operationalization. Does this seem fine?

Relevant code from `RestoreSnapshotHelper` class
```
        } else if (restoreDir.toUri().getPath().startsWith(rootDir.toUri().getPath())) {
            throw new IllegalArgumentException("Restore directory cannot be a sub directory of HBase root directory. RootDir: " + rootDir + ", restoreDir: " + restoreDir);
```

phoenix-core/src/main/java/org/apache/phoenix/schema/stats/MapperStatisticsCollector.java
BinShi-SecularBird: Shall we get the table object in the caller of getGuidePostDepthFromSystemCatalog() and pass the object to the function. This is a small factor, and can we do it in this change?

BinShi-SecularBird: I just think we can avoid a large amount of duplicated code with small effort, so we should do it in this change.

phoenix-core/src/main/java/org/apache/phoenix/schema/stats/UpdateStatisticsTool.java
```diff
@@ -0,0 +1,210 @@
...
+/**
+ * Tool to collect table level statistics on HBase snapshot
+ */
+public class UpdateStatisticsTool extends Configured implements Tool {
+
+    private static final Logger LOG = LoggerFactory.getLogger(UpdateStatisticsTool.class);
+
+    private static final Option TABLE_NAME_OPTION = new Option("t", "table", true,
+            "Phoenix Table Name");
+    private static final Option SNAPSHOT_NAME_OPTION = new Option("s", "snapshot", true,
```
BinShi-SecularBird: For a given table, how do we know which snapshot to take and what's the name of snapshot? If we always use the latest snapshot or always take a new snapshot, can we avoid this SNAPSHOT_NAME_OPTION thus reduce operation cost?

karanmehta93: Will be addressed soon as part of PHOENIX-5091.
The PR will be put up once this gets in.

BinShi-SecularBird:
> > I didn't see my concern "if time since last update is less than certain threshold, the job doesn't need to update the regions' stats again. Suppose one MR job failed but some regions still get updated, the rerun job only needs to update the regions that the first job didn't update." described in the PHOENIX-4009 being addressed in this change, and I still think it should be addressed.
> 
> The first aspect should be fairly simple to add. We can update it as part of this Jira or the PHOENIX-5091.
> 
> The second aspect of re-running only the necessary tasks is bit complicated. The mappers would retry for particular region when they fail (upto limit of max attempts). However I dont really feel the need for that optimization as of now. Let me know what your thoughts are.

I didn't mean "The second aspect of re-running only the necessary tasks". What I mean is -- assume the first MR job scheduled 100 mappers, 90 of them succeeded and 10 them failed, so the first job successfully updated 90 regions but the whole job failed. The second job (the retry job, assume it succeed) still schedule 100 mappers but only 10 mapper should actually update the stats on the 10 regions which failed in the first job and the other 90 mapper should skip stats update and succeed.

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/util/PhoenixConfigurationUtil.java
karanmehta93: > My point is that the meanings of QUERY and UPDATE_STATS have overlap

They have some overlap. All of them are Phoenix MR job types. However Phoenix MR is generic framework.
`QUERY` suggests `Phoenix MR jobs` that allow you to run `arbitrary queries` (without aggregations)
`UPDATE_STATS` suggests `Phoenix MR jobs` that will `collect statistics` (internally it might run a fixed query)

Depending on use case, the appropriate mappers/record readers/other MR components etc will be called.

phoenix-core/src/main/java/org/apache/phoenix/schema/stats/UpdateStatisticsTool.java
karanmehta93: I updated the patch to reflect this new change.

phoenix-core/src/main/java/org/apache/phoenix/schema/stats/MapperStatisticsCollector.java
karanmehta93: Alright, that makes sense. Thanks @BinShi-SecularBird , I have updated the patch accordingly.

BinShi-SecularBird:
> > > I didn't see my concern "if time since last update is less than certain threshold, the job doesn't need to update the regions' stats again. Suppose one MR job failed but some regions still get updated, the rerun job only needs to update the regions that the first job didn't update." described in the PHOENIX-4009 being addressed in this change, and I still think it should be addressed.
> > 
> > 
> > The first aspect should be fairly simple to add. We can update it as part of this Jira or the PHOENIX-5091.
> > The second aspect of re-running only the necessary tasks is bit complicated. The mappers would retry for particular region when they fail (upto limit of max attempts). However I dont really feel the need for that optimization as of now. Let me know what your thoughts are.
> 
> I didn't mean "The second aspect of re-running only the necessary tasks". What I mean is -- assume the first MR job scheduled 100 mappers, 90 of them succeeded and 10 them failed, so the first job successfully updated 90 regions but the whole job failed. The second job (the retry job, assume it succeed) still schedule 100 mappers but only 10 mapper should actually update the stats on the 10 regions which failed in the first job and the other 90 mapper should skip stats update and succeed.

this is a common case - we always have some jobs failing after they finish most of work (updated > 90% regions) but fail due to some reason, the retry job should just do minimal work instead of update stats of the whole table again.

karanmehta93:
> this is a common case - we always have some jobs failing after they finish most of work (updated > 90% regions) but fail due to some reason, the retry job should just do minimal work instead of update stats of the whole table again.

This should not be common case. Common case should be that few mappers can fail due to some reason. We have retries for that. The job should NOT fail as a whole. Even MR framework doesn't persist data between jobs and can get really tricky depending on use cases. Better way is to make the job idempotent so that re-run doesn't affect it. 

Also it is hard in this case since we don't know what changed between retries. If snapshot name changed, that can potentially affect region boundaries. We would need another level of orchestration that persists stats MR job information in some table and we look it up before running the current job. These cases would be difficult to handle. I would prefer that we try to avoid that complexity here.

BinShi-SecularBird:
> > this is a common case - we always have some jobs failing after they finish most of work (updated > 90% regions) but fail due to some reason, the retry job should just do minimal work instead of update stats of the whole table again.
> 
> This should not be common case. Common case should be that few mappers can fail due to some reason. We have retries for that. The job should NOT fail as a whole. Even MR framework doesn't persist data between jobs and can get really tricky depending on use cases. Better way is to make the job idempotent so that re-run doesn't affect it.
> 
> Also it is hard in this case since we don't know what changed between retries. If snapshot name changed, that can potentially affect region boundaries. We would need another level of orchestration that persists stats MR job information in some table and we look it up before running the current job. These cases would be difficult to handle. I would prefer that we try to avoid that complexity here.

There are always some jobs (might be a few in a day) failing, because few mappers in a job continuously failing and even retries can't get over the issue which causes the whole job to fail -- this is the case I'm talking about. In this case, I want the retry job to skip the regions whose stats have already been updated and only do minimal work. In our clusters, if phoenix MR jobs failure rate is low, I'm ok with the current approach.

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/util/PhoenixConfigurationUtil.java
BinShi-SecularBird: If you check the usages of SchemaType in the whole code base, you'll find SchemaType is mainly used by Phoenix-Pig and has its specific meaning in Phoenix-Pig. TABLE and QUERY maps to two PhoenixHBaseLoader scheme "hbase://table/" and "hbase://query", and it also maps to corresponding resource schema in Phoenix-Pig. That's why it seems to a big idea to define new schema type and pollute its usage in Phoenix_Pig. I prefer to add a new configuration or new Parameter in PhoenixMapReduceUtil.setInput to indicate this is UPDATE_STATISTICS MR job.

phoenix-core/src/main/java/org/apache/phoenix/schema/stats/DefaultStatisticsCollector.java
```diff
@@ -121,92 +120,121 @@
...
-            if (guidePostWidthBytes != null) {
-                guidepostWidth = PLong.INSTANCE.getCodec().decodeInt(guidePostWidthBytes, 0, SortOrder.getDefault());
-            }
-            this.guidePostDepth = StatisticsUtil.getGuidePostDepth(guidepostPerRegion, guidepostWidth,
-                env.getRegion().getTableDesc());
+            getGuidePostDepthFromStatement();
+            LOG.info("Guide post depth determined from SQL statement: " + guidePostDepth);
         } else {
+            long guidepostWidth = getGuidePostDepthFromSystemCatalog(getHTableForSystemCatalog());
+            if (guidepostWidth >= 0) {
```
dbwong: nit: > 0, as 0 is logically a bad width

phoenix-core/src/main/java/org/apache/phoenix/schema/stats/MapperStatisticsCollector.java
```diff
@@ -0,0 +1,70 @@
...
+import org.apache.phoenix.util.SchemaUtil;
+
+import java.io.IOException;
+import java.sql.SQLException;
+
+/**
+ * Implementation for DefaultStatisticsCollector when running inside Hadoop MR job mapper
+ * Triggered via UpdateStatisticsTool class
+ */
+public class MapperStatisticsCollector extends DefaultStatisticsCollector {
```
dbwong: As discussed offline I still prefer an approach of injecting a statswriter if possible to this inheritance pattern.

phoenix-core/src/main/java/org/apache/phoenix/schema/stats/PhoenixStatsCollectorWritable.java
```diff
@@ -0,0 +1,50 @@
...
+import java.util.Arrays;
+import java.util.List;
+
+public class PhoenixStatsCollectorWritable implements DBWritable {
+
+    private List<Cell> rawCells;
+
+    @Override
+    public void write(PreparedStatement preparedStatement) throws SQLException {
+
```
dbwong: nit: Comment for do nothing intentionally.

phoenix-pig/src/main/java/org/apache/phoenix/pig/util/PhoenixPigSchemaUtil.java
```diff
@@ -62,7 +62,7 @@ public static ResourceSchema getResourceSchema(final Configuration configuration
                 Preconditions.checkNotNull(sqlQuery, "No Sql Query exists within the configuration");
                 final SqlQueryToColumnInfoFunction function = new SqlQueryToColumnInfoFunction(configuration);
                 columns = function.apply(sqlQuery);
-            } else {
+            } else if (SchemaType.TABLE.equals(schemaType)) {
```
dbwong: nit: Earlier you used == for enum comparison, is there a standard for Phoenix?

karanmehta93:
> There are always some jobs (might be a few in a day) failing, because few mappers in a job continuously failing and even retries can't get over the issue which causes the whole job to fail -- this is the case I'm talking about, and it happens more frequently when some bad thing happen in the cluster. 

I understand your concern and I also agree that it can happen often. As you already pointed out, the simplest way to combat that is to retry the whole job again (or at certain intervals) and hope that it eventually succeeds. If not, we can raise appropriate alerts using monitoring infrastructure.

> In this case, I want the retry job to skip the regions whose stats have already been updated and only do minimal work, so it wouldn't worsen the bad situation in the cluster and we can easily catch up to avoid missing SLA. As the current phase, I'm ok to proceed without this skip check.

I understand the idea. Determining which regions data is missing from SYSTEM.STATS table is not possible (as part of this code) since the snapshot might have changed between the two jobs. 
A better way (in my understanding) of implementing this feature would be wrapper class for this tool which is aware about the job id and other details for the previous job. It can ensure that the job runs on the same snapshot everytime and mappers are only spawned accordingly (or even if mappers are launched, most of them are no-op). At this point, I feel that we should skip it, however feel free to add this as an potential enhancement to PHOENIX-5091.

phoenix-pig/src/main/java/org/apache/phoenix/pig/util/PhoenixPigSchemaUtil.java
karanmehta93: Yes I agree that we should standardize it. There is no pre-defined standard here and both of them work in theory. Any particular prefs you suggest for one over other? Also can you point me to other location (if you remember)?

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/util/PhoenixConfigurationUtil.java
karanmehta93: @BinShi-SecularBird  Although its _primarily_ being used for phoenix-pig module, phoenix-core also uses it to determine the query it needs to run for MR job. I want to _extend_ the meaning of `SchemaType` enum to suggest this as well since that can be extended again if we have new types/use cases in future. Adding a configuration each time doesn't seem good to me.

@dbwong?

phoenix-core/src/main/java/org/apache/phoenix/schema/stats/DefaultStatisticsCollector.java
karanmehta93: Having it 0 disables the stats collection here. So it stats were already present previously, this task will attempt to delete all of them from SYSTEM.STATS table. Not reading up 0 as legal value would mean the fallback to default value (which can potentially cause issue).

When a table is created, the `GUIDE_POSTS_WIDTH` column in `SYSTEM.CATALOG` is null, so it always fall back to default global value (and sometimes generates 1 GPW in the table). Having it to 0 explicitly disables it and hence it is put up here.

I will add a comment here at the top.

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/util/PhoenixConfigurationUtil.java
BinShi-SecularBird: This is my last concern about your change. I object to reuse SchemaType for indicating UPDATE_STATISTICS MR job, because we need to keep logical independence of data structure and data type. I don't want to block your progress on your work. If you insist on making this change, you can ignore my comment and proceed with commit.

karanmehta93: As discussed offline, we will add a new enum named `PhoenixMRJobType` to indicate that it is a stats collection job.

karanmehta93: @BinShi-SecularBird Updated the patch with the `MRJobType` enum.

BinShi-SecularBird:
@karanmehta93, the change looks good to me. Signed off.

phoenix-pig/src/main/java/org/apache/phoenix/pig/util/PhoenixPigSchemaUtil.java
dbwong: I prefer == as you won't accidentally order the equality wrong and risk a NPE.

karanmehta93:
Thanks @BinShi-SecularBird @dbwong and @twdsilva for the review. I will push it to this branch.
The patch cannot be ported to 4.x-HBase-1.3 and 4.x-HBase-1.2 branches since they don't have relevant hbase project dependencies (hbase-metrics and hbase-metrics-api). I will forward port this to 5.x-HBase-2.0 branch as well.

PR #423, opened 2019-01-03
PHOENIX-5059 Use the Datasource v2 api in the spark connector

twdsilva:


twdsilva:
The PR adds ```PhoenixDataSource``` which implements the new [DataSourceV2 API](https://issues.apache.org/jira/browse/SPARK-15689). This will allow further optimizations in the future such as limit and aggregate pushdown, see [SPARK-22386](https://issues.apache.org/jira/browse/SPARK-22386) 

```PhoenixSparkIT``` and ```PhoenixSparkITTenantSpecific``` were converted to use this API. 
```PhoenixRelation``` and related classes that implemented the old API were deprecated.
I also bumped up the spark version to  2.4.0 and set the java source and target version to 1.8 for the maven-compiler-plugin.

@jmahonin can you please review?

**Comment by twdsilva** [comment]
@ChinmaySKulkarni  can you please review ?

**Comment by twdsilva** [comment]
@vincentpoon can you review?

File: phoenix-spark/src/it/scala/org/apache/phoenix/spark/AbstractPhoenixSparkIT.scala [hunk 1]
```diff
@@ -50,7 +51,7 @@ class AbstractPhoenixSparkIT extends FunSuite with Matchers with BeforeAndAfter
   final val TenantId = "theTenant"
 
   var conn: Connection = _
-  var sc: SparkContext = _
+  var spark: SparkSession = _
```
ChinmaySKulkarni: @twdsilva What is the difference between SparkContext and SparkSession?

ChinmaySKulkarni: Is this more of an update to use new APIs?

twdsilva: ```SparkSession``` is the new unified entry point (see https://databricks.com/blog/2016/08/15/how-to-use-sparksession-in-apache-spark-2-0.html)

jmahonin:
@twdsilva Very impressive work, well done!

A few questions:
- Does this break compatibility with previous Spark versions? i.e., Spark 2.3.x wouldn't be binary-compatible with Phoenix 5.1.x (or whatever the next version ends up being)

- It seems like much of the boiler-plate MR code has been replicated yet-again into this new integration. The overall structure makes sense here, but I'm wondering if some of that code shouldn't live in phoenix-spark at all?

twdsilva:
@jmahonin  Thanks for the review. The Datasource V2 API is marked evolving and the API changed from Spark 2.3 to 2.4 and it will change again in Spark 3.0. The DataSource v2 connector in this patch will only work with Spark 2.4. 
I deprecated but did not remove the current connector that uses the V1 API, so users should still be able to use the current connector. 
There is a JIRA to move all the connectors to their own repo (see [PHOENIX-5062](https://issues.apache.org/jira/browse/PHOENIX-5062)) which will allow us to release the connectors independently of phoenix core. I was planning on working on that after this PR goes in. 

For 2) are you referring to ```PhoenixRecordWritable```? That class is deprecated and only used in the old connector. 
Spark has code to convert from/to a ResultSet to/from InternalRow which I used (see ```SparkJdbcUtil```). The setters methods are accessible, but there is a bug in handling ```ByteType```, so I had to copy the code. This will be fixed in [SPARK-26499](https://issues.apache.org/jira/browse/SPARK-26499) after which we can switch to just using Spark's ```JdbcUtils.resultSetToSparkInternalRows```
The setters methods are not accessible, so I had to create a copy of them (see ```makeSetter``` in ```SparkJdbcUtil```).

jmahonin:
LGTM, and well done fixing and getting SPARK-26499 merged.

rejeb:
Hi @twdsilva,

This PR has introduced a beacking change regarding the way the schema is inferred :

- In previous version code uses `ColumnInfo.getDisplayName` to define the name of the column in the DF.
- The new class SparkSchemaUtil the method used is  `ColumnInfo.getColumnName` which returns the columnName as `columnFamilyName.columnName`.

This change breaks jobs reading from tables having a non default column family.

The saprk3 connector have the same behavior since it uses the same code for resolving schema.
I noticed that in tests there is no tests using tables with a column family. May be it should be some tests with this cases ?

What do you think about this ? Is there any reason behind introducing this breaking change ?

Thanks.

PR #424, opened 2019-01-03
PHOENIX-4820

comnetwork:


comnetwork:
This PR is merged @twdsilva  reviews of PR 417 and made a few modifications for BaseExpression ,BaseCompoundExpression and BaseSingleExpression in order to pass the Unit tests and IT tests.

phoenix-core/src/main/java/org/apache/phoenix/expression/BaseExpression.java
```diff
@@ -255,4 +255,15 @@ public boolean requiresFinalEvaluation() {
         return false;
     }
 
+    @Override
+    public boolean isCloneExpression()  {
+       return isCloneExpressionByDeterminism(this);
+    }
+
+    protected static boolean isCloneExpressionByDeterminism(BaseExpression expression) {
```
twdsilva: nit: I don't think this method is need, you can just inline this in ```public boolean isCloneExpression() ```

comnetwork: @twdsilva , thank you for the review, I added BaseExpression.isCloneExpressionByDeterminism method here for the purpose that it can also be invoked by RandomFunction. RandomFunction is derived from BaseCompoundExpression , and if RandomFunction does not override isCloneExpression method, the isCloneExpression method of RandomFunction would use the isCloneExpression method of BaseCompoundExpression  and return false,   the existing Unit Test ArithmeticOperationTest would fail.

gjacoby126:
JIRA resolved, closing PR

PR #416, opened 2018-12-15
PHOENIX-5069 please go to JIRA to see the detailed design document at…

BinShi-SecularBird:
…tached.

PHOENIX-4998 The estimated instance size of GuidePostsInfo doesn't count the last update timestamp array.

PR #425, opened 2019-01-08
PHOENIX-5069 Use asynchronous refresh to provide non-blocking Phoenix Stats Client Cache

BinShi-SecularBird:
PHOENIX-5069 please go to JIRA to see the detailed design document attached.

PHOENIX-4998 The estimated instance size of GuidePostsInfo doesn't count the last update timestamp array.

Signed-off-by: Bin <bshi@salesforce.com>

### DIFF
--- a/phoenix-core/src/main/java/org/apache/phoenix/query/GuidePostsCache.java
+++ b/phoenix-core/src/main/java/org/apache/phoenix/query/GuidePostsCache.java
@@ -22,11 +22,11 @@ import static org.apache.phoenix.query.QueryServicesOptions.DEFAULT_STATS_COLLEC
 import java.io.IOException;
 import java.util.List;
 import java.util.Objects;
-import java.util.concurrent.ExecutionException;
-import java.util.concurrent.TimeUnit;
+import java.util.concurrent.*;
 
 import org.apache.hadoop.conf.Configuration;
 import org.apache.hadoop.hbase.HConstants;
+import org.apache.hadoop.hbase.TableName;
 import org.apache.hadoop.hbase.TableNotFoundException;
 import org.apache.hadoop.hbase.client.Table;
 import org.apache.hadoop.hbase.client.TableDescriptor;
@@ -42,7 +42,6 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import com.google.common.cache.CacheBuilder;
-import com.google.common.cache.CacheLoader;
 import com.google.common.cache.LoadingCache;
 import com.google.common.cache.RemovalCause;
 import com.google.common.cache.RemovalListener;
@@ -61,18 +60,25 @@ public class GuidePostsCache {
 
     public GuidePostsCache(ConnectionQueryServices queryServices, Configuration config) {
         this.queryServices = Objects.requireNonNull(queryServices);
+
         // Number of millis to expire cache values after write
         final long statsUpdateFrequency = config.getLong(
                 QueryServices.STATS_UPDATE_FREQ_MS_ATTRIB,
                 QueryServicesOptions.DEFAULT_STATS_UPDATE_FREQ_MS);
-        // Maximum number of entries (tables) to store in the cache at one time
+
+        // Maximum total weight (size in bytes) of stats entries
         final long maxTableStatsCacheSize = config.getLong(
                 QueryServices.STATS_MAX_CACHE_SIZE,
                 QueryServicesOptions.DEFAULT_STATS_MAX_CACHE_SIZE);
+
 		final boolean isStatsEnabled = config.getBoolean(STATS_COLLECTION_ENABLED, DEFAULT_STATS_COLLECTION_ENABLED);
+
+        PhoenixStatsCacheLoader cacheLoader = new PhoenixStatsCacheLoader(
+                isStatsEnabled ? new StatsLoaderImpl() : new EmptyStatsLoader(), config);
+
         cache = CacheBuilder.newBuilder()
-                // Expire entries a given amount of time after they were written
-                .expireAfterWrite(statsUpdateFrequency, TimeUnit.MILLISECONDS)
+                // Refresh entries a given amount of time after they were written
+                .refreshAfterWrite(statsUpdateFrequency, TimeUnit.MILLISECONDS)
                 // Maximum total weight (size in bytes) of stats entries
                 .maximumWeight(maxTableStatsCacheSize)
                 // Defer actual size to the PTableStats.getEstimatedSize()
@@ -83,19 +89,38 @@ public class GuidePostsCache {
                 })
                 // Log removals at TRACE for debugging
                 .removalListener(new PhoenixStatsCacheRemovalListener())
-                // Automatically load the cache when entries are missing
-                .build(isStatsEnabled ? new StatsLoader() : new EmptyStatsLoader());
+                // Automatically load the cache when entries need to be refreshed
+                .build(cacheLoader);
     }
 
     /**
-     * {@link CacheLoader} implementation for the Phoenix Table Stats cache.
+     * {@link PhoenixStatsLoader} implementation for the Stats Loader.
      */
-    protected class StatsLoader extends CacheLoader<GuidePostsKey, GuidePostsInfo> {
+    protected class StatsLoaderImpl implements PhoenixStatsLoader {
         @Override
-        public GuidePostsInfo load(GuidePostsKey statsKey) throws Exception {
-            Table statsHTable = queryServices.getTable(SchemaUtil.getPhysicalName(
+        public boolean needsLoad() {
+            // For now, whenever it's called, we try to load stats from stats table
+            // no matter it has been updated or not.
+            // Here are the possible optimizations we can do here:
+            // 1. Load stats from the stats table only when the stats get updated on the server side.
+            // 2. Support different refresh cycle for different tables.
+            return true;
+        }
+
+        @Override
+        public GuidePostsInfo loadStats(GuidePostsKey statsKey) throws Exception {
+            return loadStats(statsKey, GuidePostsInfo.NO_GUIDEPOST);
+        }
+
+        @Override
+        public GuidePostsInfo loadStats(GuidePostsKey statsKey, GuidePostsInfo prevGuidepostInfo) throws Exception {
+            assert(prevGuidepostInfo != null);
+
+            TableName tableName = SchemaUtil.getPhysicalName(
                     PhoenixDatabaseMetaData.SYSTEM_STATS_NAME_BYTES,
-                            queryServices.getProps()).getName());
+                    queryServices.getProps());
+            Table statsHTable = queryServices.getTable(tableName.getName());
+
             try {
                 GuidePostsInfo guidePostsInfo = StatisticsUtil.readStatistics(statsHTable, statsKey,
                         HConstants.LATEST_TIMESTAMP);
@@ -103,18 +128,17 @@ public class GuidePostsCache {
                 return guidePostsInfo;
             } catch (TableNotFoundException e) {
                 // On a fresh install, stats might not yet be created, don't warn about this.
-                logger.debug("Unable to locate Phoenix stats table", e);
-                return GuidePostsInfo.NO_GUIDEPOST;
+                logger.debug("Unable to locate Phoenix stats table: " + tableName.toString(), e);
+                return prevGuidepostInfo;
             } catch (IOException e) {
-                logger.warn("Unable to read from stats table", e);
-                // Just cache empty stats. We'll try again after some time anyway.
-                return GuidePostsInfo.NO_GUIDEPOST;
+                logger.warn("Unable to read from stats table: " + tableName.toString(), e);
+                return prevGuidepostInfo;
             } finally {
                 try {
                     statsHTable.close();
                 } catch (IOException e) {
                     // Log, but continue. We have our stats anyway now.
-                    logger.warn("Unable to close stats table", e);
+                    logger.warn("Unable to close stats table: " + tableName.toString(), e);
                 }
             }
         }
@@ -125,20 +149,30 @@ public class GuidePostsCache {
         void traceStatsUpdate(GuidePostsKey key, GuidePostsInfo info) {
             if (logger.isTraceEnabled()) {
                 logger.trace("Updating local TableStats cache (id={}) for {}, size={}bytes",
-                      new Object[] {Objects.hashCode(GuidePostsCache.this), key,
-                      info.getEstimatedSize()});
+                        new Object[] {Objects.hashCode(GuidePostsCache.this), key, info.getEstimatedSize()});
             }
         }
     }
 
     /**
+     * {@link PhoenixStatsLoader} implementation for the Stats Loader.
      * Empty stats loader if stats are disabled
      */
-	protected class EmptyStatsLoader extends CacheLoader<GuidePostsKey, GuidePostsInfo> {
-		@Override
-		public GuidePostsInfo load(GuidePostsKey statsKey) throws Exception {
-			return GuidePostsInfo.NO_GUIDEPOST;
-		}
+	protected class EmptyStatsLoader implements PhoenixStatsLoader {
+        @Override
+        public boolean needsLoad() {
+            return false;
+        }
+
+        @Override
+        public GuidePostsInfo loadStats(GuidePostsKey statsKey) throws Exception {
+            return GuidePostsInfo.NO_GUIDEPOST;
+        }
+
+        @Override
+        public GuidePostsInfo loadStats(GuidePostsKey statsKey, GuidePostsInfo prevGuidepostInfo) throws Exception {
+            return GuidePostsInfo.NO_GUIDEPOST;
+        }
 	}
 
     /**
--- a/phoenix-core/src/main/java/org/apache/phoenix/query/PhoenixStatsLoader.java
+++ b/phoenix-core/src/main/java/org/apache/phoenix/query/PhoenixStatsLoader.java
@@ -1,0 +1,57 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.phoenix.query;
+
+import org.apache.phoenix.schema.stats.GuidePostsInfo;
+import org.apache.phoenix.schema.stats.GuidePostsKey;
+
+/**
+ * The interface for talking to underneath layers to load stats from stats table for a given key
+ */
+public interface PhoenixStatsLoader {
+    /**
+     * Use to check whether this is the time to load stats from stats table.
+     * There are two cases:
+     * a. After a specified duration has passed
+     * b. The stats on server side (e.g. in stats table) has been updated
+     *
+     * @return boolean indicates whether we need to load stats or not
+     */
+    boolean needsLoad();
+
+    /**
+     * Called by client stats cache to load stats from underneath layers
+     *
+     * @param statsKey the stats key used to search the stats on server side (in stats table)
+     * @throws Exception
+     *
+     * @return GuidePostsInfo retrieved from sever side
+     */
+    GuidePostsInfo loadStats(GuidePostsKey statsKey) throws Exception;
+
+    /**
+     * Called by client stats cache to load stats from underneath layers
+     *
+     * @param statsKey the stats key used to search the stats on server side (in stats table)
+     * @param prevGuidepostInfo the existing stats cached on the client side or GuidePostsInfo.NO_GUIDEPOST
+     * @throws Exception
+     *
+     * @return GuidePostsInfo retrieved from sever side
+     */
+    GuidePostsInfo loadStats(GuidePostsKey statsKey, GuidePostsInfo prevGuidepostInfo) throws Exception;
+}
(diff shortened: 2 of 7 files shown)


phoenix-core/src/main/java/org/apache/phoenix/query/GuidePostsCache.java
```diff
@@ -85,39 +99,47 @@ public GuidePostsCache(ConnectionQueryServices queryServices, Configuration conf
...
                 return guidePostsInfo;
             } catch (TableNotFoundException e) {
                 // On a fresh install, stats might not yet be created, don't warn about this.
-                logger.debug("Unable to locate Phoenix stats table", e);
-                return GuidePostsInfo.NO_GUIDEPOST;
+                logger.debug("Unable to locate Phoenix stats table: " + tableName.toString(), e);
+                return prevGuidepostInfo;
             } catch (IOException e) {
-                logger.warn("Unable to read from stats table", e);
+                logger.warn("Unable to read from stats table: " + tableName.toString(), e);
```
BinShi-SecularBird: Comment from @karanmehta93 (copied from prev PR https://github.com/apache/phoenix/pull/416): nit: remove this comment

phoenix-core/src/test/java/org/apache/phoenix/query/PhoenixStatsCacheLoaderTest.java
```diff
@@ -0,0 +1,123 @@
...
+                // Log removals at TRACE for debugging
+                .removalListener(new GuidePostsCache.PhoenixStatsCacheRemovalListener())
+                // Automatically load the cache when entries are missing
+                .build(new PhoenixStatsCacheLoader(new TestStatsLoaderImpl(), executor));
+
+        GuidePostsKey guidePostsKey = new GuidePostsKey(new byte[4], new byte[4]);
+        GuidePostsInfo guidePostsInfo = getStats(cache, guidePostsKey);
+        assertTrue(guidePostsInfo.getMaxLength() == 1);
+        sleep(3000);
+        guidePostsInfo = getStats(cache, guidePostsKey);
```
BinShi-SecularBird: Comment from @twdsilva (copied from prev PR #416): Instead of relying on sleep which tends to cause tests to fail on the build machines, can you use a latch or some other mechanism to determine that the guideposts have been refreshed?

BinShi-SecularBird:
Comment from @twdsilva (copied from prev PR #416): Looks pretty good, I had a comment about the test.

BinShi-SecularBird:
@twdsilva @karanmehta93, please use this new PR to review my change for PHOENIX-5069. The old PR https://github.com/apache/phoenix/pull/416 has been abandoned during resolving the conflicts exiting in my original PR.

phoenix-core/src/main/java/org/apache/phoenix/query/PhoenixStatsLoader.java
```diff
@@ -0,0 +1,47 @@
...
+     * There are two cases:
+     * a. After a specified duration has passed
+     * b. The stats on server side (e.g. in stats table) has been updated
+     *
+     * @return boolean indicates whether we need to load stats or not
+     */
+    boolean needsLoad();
+
+    /**
+     * Called by ring buffer event handler to write RingBufferEvent
```
karanmehta93: Is this relevant?

phoenix-core/src/main/java/org/apache/phoenix/query/GuidePostsCache.java
```diff
@@ -59,22 +59,36 @@
...
 
     public GuidePostsCache(ConnectionQueryServices queryServices, Configuration config) {
         this.queryServices = Objects.requireNonNull(queryServices);
+
+        // The size of the thread pool used for refreshing cached table stats
+        final int statsCacheThreadPoolSize = config.getInt(
+                QueryServices.STATS_CACHE_THREAD_POOL_SIZE,
+                QueryServicesOptions.DEFAULT_STATS_CACHE_THREAD_POOL_SIZE);
+
+        executor = Executors.newFixedThreadPool(statsCacheThreadPoolSize);
```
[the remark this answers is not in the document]
karanmehta93: On a second thought, do we know why do we build the cache even when stats are disabled? Can we skip it or would it result in NPE's all around?

phoenix-core/src/main/java/org/apache/phoenix/query/GuidePostsCache.java
BinShi-SecularBird: Why? We need this warning to know some bad thing happened which caused that the stats can't be loaded unexpectedly? This log is printed one time every 15 minutes for each entry only when bad thing is happening, so we won't have too many such logs.

phoenix-core/src/main/java/org/apache/phoenix/query/PhoenixStatsLoader.java
BinShi-SecularBird: Yes. It should be called in PhoenixStatsCacheLoader.reload() as your another comment pointed out, but the change was lost when I pushed my second change after I resolved the conflicts somehow.

phoenix-core/src/main/java/org/apache/phoenix/query/GuidePostsCache.java
karanmehta93: I was actually referring to the comment on Line 135 here, which is not valid anymore.

phoenix-core/src/main/java/org/apache/phoenix/query/PhoenixStatsLoader.java
karanmehta93: Sorry for the confusion here.
I was referring to this comment `Called by ring buffer event handler to write RingBufferEvent`. We are not using any ring buffers for this, right?

BinShi-SecularBird: Good catch. I copied a comment from another class to reuse the format but didn't clean it up thoroughly.

phoenix-core/src/main/java/org/apache/phoenix/query/GuidePostsCache.java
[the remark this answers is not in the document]
BinShi-SecularBird: OK, I changed to initialize thread pool only if stats are enabled.

To the question "do we know why do we build the cache even when stats are disabled?", my thought is that this is fastest and simplest way to fix it, although it isn't the most efficient way.      Fixing in this way still provides the uniform interface to the callers of the GuidePostCache so the calls don't need to something like "if stats are enabled then call this; otherwise call that". One of efficient and formal ways to fix it is that we define the interface of GuidePostsCache, two classes implement the interface - one is GuidePostsEmptyCache and another is current GuidePostsCache. We create object of GuidePostsEmptyCache when stats are disabled.

phoenix-core/src/main/java/org/apache/phoenix/query/GuidePostsCache.java
```diff
@@ -85,39 +106,46 @@ public GuidePostsCache(ConnectionQueryServices queryServices, Configuration conf
...
     /**
-     * {@link CacheLoader} implementation for the Phoenix Table Stats cache.
+     * {@link PhoenixStatsLoader} implementation for the Stats Loader.
      */
-    protected class StatsLoader extends CacheLoader<GuidePostsKey, GuidePostsInfo> {
+    protected class StatsLoaderImpl implements PhoenixStatsLoader {
         @Override
-        public GuidePostsInfo load(GuidePostsKey statsKey) throws Exception {
+        public boolean needsLoad() {
+            // Whenever it's called, we try to load stats from stats table
```
karanmehta93: nit: Put a TODO here describing how we would use in future.

phoenix-core/src/test/java/org/apache/phoenix/query/PhoenixStatsCacheLoaderTest.java
karanmehta93: You can use this instead of actual time to make tests faster and reliable. 
https://stackoverflow.com/questions/29990788/guava-ticker-cache-expire

phoenix-core/src/main/java/org/apache/phoenix/query/GuidePostsCache.java
BinShi-SecularBird: Removed the comment.

karanmehta93:
Please upload a patch file to the Jira for Hadoop QA to run. All commits need to be squashed into 1 before doing it. See: https://phoenix.apache.org/contributing.html

phoenix-core/src/main/java/org/apache/phoenix/query/GuidePostsCache.java
BinShi-SecularBird: added

BinShi-SecularBird:
> Please upload a patch file to the Jira for Hadoop QA to run. All commits need to be squashed into 1 before doing it. See: https://phoenix.apache.org/contributing.html

done.

karanmehta93:
Also, make sure that the commit message follows the format `PHOENIX-XXXX <Jira Title>` in future. This doesn't have to be for the commits that you push in the PR, it is just for the patch file that is uploaded to Jira. Thanks!

phoenix-core/src/test/java/org/apache/phoenix/query/PhoenixStatsCacheLoaderTest.java
BinShi-SecularBird: Fixed by using CountDownLatch to guarantee deterministic test behavior.

phoenix-core/src/main/java/org/apache/phoenix/schema/stats/GuidePostsInfo.java
```diff
@@ -95,15 +95,17 @@ public GuidePostsInfo(List<Long> byteCounts, ImmutableBytesWritable guidePosts,
...
+        // Those Java equivalents of sizeof() in C/C++, mentioned on the Web, might be overkilled here.
+        int estimatedSize = SizedUtil.OBJECT_SIZE
                 + SizedUtil.IMMUTABLE_BYTES_WRITABLE_SIZE + guidePosts.getLength() // guidePosts
                 + SizedUtil.INT_SIZE // maxLength
                 + SizedUtil.INT_SIZE // guidePostsCount
                 + SizedUtil.ARRAY_SIZE + this.rowCounts.length * SizedUtil.LONG_SIZE // rowCounts
                 + SizedUtil.ARRAY_SIZE + this.byteCounts.length * SizedUtil.LONG_SIZE // byteCounts
+                + SizedUtil.ARRAY_SIZE + this.gpTimestamps.length * SizedUtil.LONG_SIZE // gpTimestamps
                 + SizedUtil.INT_SIZE; // estimatedSize
         this.estimatedSize = estimatedSize;
```
dbwong: nit: remove unused intermediate

phoenix-core/src/main/java/org/apache/phoenix/query/GuidePostsCache.java
```diff
@@ -43,13 +43,13 @@
...
 
 import com.google.common.cache.CacheBuilder;
-import com.google.common.cache.CacheLoader;
 import com.google.common.cache.LoadingCache;
 import com.google.common.cache.RemovalCause;
 import com.google.common.cache.RemovalListener;
 import com.google.common.cache.RemovalNotification;
 import com.google.common.cache.Weigher;
 
+
```
dbwong: The GuidePostsCache was mostly rewritten here.  Is it possible to add a unit test for this?

phoenix-core/src/main/java/org/apache/phoenix/query/GuidePostsCache.java
```diff
@@ -59,22 +59,43 @@
...
+        PhoenixStatsCacheLoader cacheLoader;
+
+        if (isStatsEnabled) {
+            // The size of the thread pool used for refreshing cached table stats
+            final int statsCacheThreadPoolSize = config.getInt(
+                    QueryServices.STATS_CACHE_THREAD_POOL_SIZE,
+                    QueryServicesOptions.DEFAULT_STATS_CACHE_THREAD_POOL_SIZE);
+
+            executor = Executors.newFixedThreadPool(statsCacheThreadPoolSize);
```
dbwong: Why was the executorService placed at this level?  Why do we need a private variable we don't use it outside of this one place.  Did we consider pushing it to a place where we can be more assured of it being constructed only once and then providing an accessor that the CacheLoader can get at?  I'm assuming we only want one instance of the GuidePostsCache but the constructor is public and the executor service isn't static and protected.

phoenix-core/src/main/java/org/apache/phoenix/query/GuidePostsCache.java
```diff
@@ -85,39 +106,54 @@ public GuidePostsCache(ConnectionQueryServices queryServices, Configuration conf
...
+     * {@link PhoenixStatsLoader} implementation for the Stats Loader.
      */
-    protected class StatsLoader extends CacheLoader<GuidePostsKey, GuidePostsInfo> {
+    protected class StatsLoaderImpl implements PhoenixStatsLoader {
         @Override
-        public GuidePostsInfo load(GuidePostsKey statsKey) throws Exception {
+        public boolean needsLoad() {
+            // Whenever it's called, we try to load stats from stats table
+            // no matter it has been updated or not.
+            // TODO:
```
dbwong: Can you link a JIRA for this todo?

phoenix-core/src/main/java/org/apache/phoenix/query/GuidePostsCache.java
```diff
@@ -128,20 +164,30 @@ public GuidePostsInfo load(GuidePostsKey statsKey) throws Exception {
...
     /**
+     * {@link PhoenixStatsLoader} implementation for the Stats Loader.
      * Empty stats loader if stats are disabled
      */
-	protected class EmptyStatsLoader extends CacheLoader<GuidePostsKey, GuidePostsInfo> {
-		@Override
-		public GuidePostsInfo load(GuidePostsKey statsKey) throws Exception {
-			return GuidePostsInfo.NO_GUIDEPOST;
-		}
+	protected class EmptyStatsLoader implements PhoenixStatsLoader {
```
dbwong: I dislike these do nothing classes.  Is there too many places to do a stats enabled if else check around the loaders?

phoenix-core/src/main/java/org/apache/phoenix/query/GuidePostsCache.java
```diff
@@ -85,39 +106,54 @@ public GuidePostsCache(ConnectionQueryServices queryServices, Configuration conf
...
+            return true;
+        }
+
+        @Override
+        public GuidePostsInfo loadStats(GuidePostsKey statsKey) throws Exception {
+            return loadStats(statsKey, GuidePostsInfo.NO_GUIDEPOST);
+        }
+
+        @Override
+        public GuidePostsInfo loadStats(GuidePostsKey statsKey, GuidePostsInfo prevGuidepostInfo) throws Exception {
```
dbwong: May want to consider a preconditions.notNull(prevGuidepostInfo)

phoenix-core/src/main/java/org/apache/phoenix/query/GuidePostsCache.java
```diff
@@ -85,39 +106,54 @@ public GuidePostsCache(ConnectionQueryServices queryServices, Configuration conf
...
+                // Automatically load the cache when entries need to be refreshed
+                .build(cacheLoader);
     }
 
     /**
-     * {@link CacheLoader} implementation for the Phoenix Table Stats cache.
+     * {@link PhoenixStatsLoader} implementation for the Stats Loader.
      */
-    protected class StatsLoader extends CacheLoader<GuidePostsKey, GuidePostsInfo> {
+    protected class StatsLoaderImpl implements PhoenixStatsLoader {
```
dbwong: Consider refactoring this into its own class and adding a unit test.

phoenix-core/src/test/java/org/apache/phoenix/query/PhoenixStatsCacheLoaderTest.java
```diff
@@ -0,0 +1,153 @@
...
+            GuidePostsKey guidePostsKey = new GuidePostsKey(new byte[4], new byte[4]);
+            GuidePostsInfo guidePostsInfo = getStats(cache, guidePostsKey);
+            assertTrue(guidePostsInfo.getMaxLength() == 1);
+
+            // Note: With Guava cache, automatic refreshes are performed when the first stale request for an entry occurs.
+
+            // After we sleep here for any time which is larger than the refresh cycle, the refresh of cache entry will be
+            // triggered for its first time by the call of getStats(). This is deterministic behavior, and it won't cause
+            // randomized test failures.
+            sleep(1500);
```
dbwong: nit: Can this be lowered for quicker tests?

phoenix-core/src/test/java/org/apache/phoenix/query/PhoenixStatsCacheLoaderTest.java
```diff
@@ -0,0 +1,153 @@
...
+            guidePostsInfo = getStats(cache, guidePostsKey);
+            // Refresh has been triggered for its first time, but still could get the old value
+            assertTrue(guidePostsInfo.getMaxLength() >= 1);
+            firstTimeRefreshedSignal.await();
+
+            sleep(1500);
+            guidePostsInfo = getStats(cache, guidePostsKey);
+            // Now the second time refresh has been triggered by the above getStats() call, the first time Refresh has completed
+            // and the cache entry has been updated for sure.
+            assertTrue(guidePostsInfo.getMaxLength() >= 2);
```
dbwong: nit: Dislike this assert as there is nothing special about length, can we just have 2 mocks rather than constructing these byte arrays in this increasing length fashion?

phoenix-core/src/test/java/org/apache/phoenix/query/PhoenixStatsCacheLoaderTest.java
karanmehta93: Yes (like 100 ms) and then have a polling loop?

phoenix-core/src/main/java/org/apache/phoenix/query/GuidePostsCache.java
BinShi-SecularBird: In the current code base, The GuidePostsCache doesn't have unit test except the inner class PhoenixStatsCacheRemovealListner. My change DOES NOT rewrite GuidePostsCache that much. As you can see, my main changes in GuidePostsCache are: a. initialize ExecutorService object and initialize Guava cache object with refresh mechanism; b. trivial changes in the inner class StatsLoaderImpl. I didn't change the interface of GuidePostsCaches including get/put/invalidate* APIs. My main changes are in PhoenixStatsCacheLoader class which I added unit test. Adding unit test for GuidePostsCache shouldn't be addressed by this change.

phoenix-core/src/main/java/org/apache/phoenix/query/GuidePostsCache.java
BinShi-SecularBird: You're right. I'll move it to PhoenixStatsCacheLoader class. Previously, I wanted to trigger automatic refresh by GuidePostsCache instead of using its built in refresh mechanism in which automatic refreshes are performed when the first stale request for an entry occurs. Since I'm doing it in that way, It's better to move it to PhoenixStatsCacheLoader for better encapsulation and isolation.

phoenix-core/src/main/java/org/apache/phoenix/query/GuidePostsCache.java
BinShi-SecularBird: I'll move this TODO because here I just want to point out the possible optimizations we can do here which haven't been formalized as the solutions we must do in the future.

phoenix-core/src/main/java/org/apache/phoenix/query/GuidePostsCache.java
BinShi-SecularBird: added assert(prevGuidepostInfo != null);

phoenix-core/src/test/java/org/apache/phoenix/query/PhoenixStatsCacheLoaderTest.java
BinShi-SecularBird: Good suggestion. Changed to 100ms and 150ms accordingly.

phoenix-core/src/test/java/org/apache/phoenix/query/PhoenixStatsCacheLoaderTest.java
BinShi-SecularBird: Here I just want to assert that the cached entry, GuidePostsInfo object, has been changed. This is the simplest way, and defining Mock class, for which most likely we need to define the interface of GuidePostsInfo and child Mock class, which seems to overkill to me. I prefer to keep the current approach. Let me know if you have further concern.

phoenix-core/src/test/java/org/apache/phoenix/query/PhoenixStatsCacheLoaderTest.java
```diff
@@ -0,0 +1,153 @@
...
+/**
+ * Test class around the PhoenixStatsCacheLoader.
+ */
+public class PhoenixStatsCacheLoaderTest {
+    /**
+     * {@link PhoenixStatsLoader} test implementation for the Stats Loader.
+     */
+    protected class TestStatsLoaderImpl implements PhoenixStatsLoader {
+        private int maxLength = 1;
+        private final CountDownLatch firstTimeRefreshedSignal;
```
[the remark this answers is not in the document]
BinShi-SecularBird: No, it can't. With Guava cache, automatic refreshes are performed when the first stale request for an entry occurs, and in the test, the getStats() must be called in expected times.

phoenix-core/src/main/java/org/apache/phoenix/query/GuidePostsCache.java
BinShi-SecularBird: I only made very trivial changes in this inner class StatsLoaderImpl. Adding unit test for StatsLoaderImpl shouldn't be addressed by this change.

phoenix-core/src/main/java/org/apache/phoenix/query/GuidePostsCache.java
karanmehta93: nit: Consider using Preconditions check instead. It is cleaner and usually not good idea to use asserts in production code.

phoenix-core/src/main/java/org/apache/phoenix/query/PhoenixStatsLoader.java
```diff
@@ -0,0 +1,57 @@
...
+
+    /**
+     * Called by client stats cache to load stats from underneath layers
+     *
+     * @param statsKey the stats key used to search the stats on server side (in stats table)
+     * @throws Exception
+     *
+     * @return GuidePostsInfo retrieved from sever side
+     */
+    GuidePostsInfo loadStats(GuidePostsKey statsKey) throws Exception;
```
karanmehta93: Do we really want two separate methods for this?

BinShi-SecularBird: Yes. Any strong reason for not to do so?

gjacoby126:
JIRA committed, closing

PR #429, opened 2019-01-16
Date filter pushdown

twdsilva:
…ark connector

@c


twdsilva:
@ChinmaySKulkarni  Can you please review?

ChinmaySKulkarni:
lgtm @twdsilva. Will this work with just the default date format or any `SimpleDateFormat`?

twdsilva:
This should work with any DateFormat, we are currently converting the java.sql.Date to a String (in the default date format) and then back to a date using TO_DATE in the where clause. 
In the future we should generate a phoenix ```Expression``` from the spark ```Filter``` .

PR #431, opened 2019-01-20
PHOENIX-5105

comnetwork:
PHOENIX-5105

### DIFF
--- a/phoenix-core/src/it/java/org/apache/phoenix/end2end/SortMergeJoinMoreIT.java
+++ b/phoenix-core/src/it/java/org/apache/phoenix/end2end/SortMergeJoinMoreIT.java
@@ -742,4 +742,112 @@ public class SortMergeJoinMoreIT extends ParallelStatsDisabledIT {
             assertFalse(rs.next());
         }
     }
+
+    @Test
+    public void testSortMergeJoinPushFilterThroughSortBug5105() throws Exception {
+        Connection conn = null;
+        try {
+            Properties props = PropertiesUtil.deepCopy(TEST_PROPERTIES);
+            conn = DriverManager.getConnection(getUrl(), props);
+
+            String tableName1=generateUniqueName();
+            String tableName2=generateUniqueName();
+
+            conn.createStatement().execute("DROP TABLE if exists "+tableName1);
+
+            String sql="CREATE TABLE IF NOT EXISTS "+tableName1+" ( "+
+                    "AID INTEGER PRIMARY KEY,"+
+                    "AGE INTEGER"+
+                    ")";
+            conn.createStatement().execute(sql);
+
+            conn.createStatement().execute("UPSERT INTO "+tableName1+"(AID,AGE) VALUES (1,11)");
+            conn.createStatement().execute("UPSERT INTO "+tableName1+"(AID,AGE) VALUES (2,22)");
+            conn.createStatement().execute("UPSERT INTO "+tableName1+"(AID,AGE) VALUES (3,33)");
+            conn.commit();
+
+            conn.createStatement().execute("DROP TABLE if exists "+tableName2);
+            sql="CREATE TABLE IF NOT EXISTS "+tableName2+" ( "+
+                    "BID INTEGER PRIMARY KEY,"+
+                    "CODE INTEGER"+
+                    ")";
+            conn.createStatement().execute(sql);
+
+            conn.createStatement().execute("UPSERT INTO "+tableName2+"(BID,CODE) VALUES (1,66)");
+            conn.createStatement().execute("UPSERT INTO "+tableName2+"(BID,CODE) VALUES (2,55)");
+            conn.createStatement().execute("UPSERT INTO "+tableName2+"(BID,CODE) VALUES (3,44)");
+            conn.commit();
+
+            //test for simple scan
+
+            sql="select /*+ USE_SORT_MERGE_JOIN */ a.aid,b.code from (select aid,age from "+tableName1+" where age >=11 and age<=33) a inner join "+
+                "(select bid,code from "+tableName2+" order by code limit 2) b on a.aid=b.bid where b.code > 50";
+            ResultSet rs=conn.prepareStatement(sql).executeQuery();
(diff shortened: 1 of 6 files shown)


phoenix-core/src/main/java/org/apache/phoenix/compile/JoinCompiler.java
```diff
@@ -761,13 +761,15 @@ public Expression compilePostFilterExpression(StatementContext context) throws S
...
+            if (isSubselect()) {
+                return SubselectRewriter.applyOrderByAndPostFilters(
                         SubselectRewriter.applyPostFilters(subselect, preFilters, tableNode.getAlias()),
                         orderBy,
                         tableNode.getAlias(),
-                        tableNode);
-
+                        postFilters);
+            }
+            //for table, postFilters is empty , because it can safely pushed down as preFilters.
```
twdsilva: If there are ```postFilters``` will they always be pushed down into the sub query?

phoenix-core/src/it/java/org/apache/phoenix/end2end/SortMergeJoinMoreIT.java
```diff
@@ -742,4 +742,112 @@ private static void verifyQueryPlanAndResultForBug4508(
...
+    public void testSortMergeJoinPushFilterThroughSortBug5105() throws Exception {
+        Connection conn = null;
+        try {
+            Properties props = PropertiesUtil.deepCopy(TEST_PROPERTIES);
+            conn = DriverManager.getConnection(getUrl(), props);
+
+            String tableName1=generateUniqueName();
+            String tableName2=generateUniqueName();
+
+            conn.createStatement().execute("DROP TABLE if exists "+tableName1);
```
twdsilva: nit: the drop isn't necessary

phoenix-core/src/it/java/org/apache/phoenix/end2end/SortMergeJoinMoreIT.java
```diff
@@ -742,4 +742,112 @@ private static void verifyQueryPlanAndResultForBug4508(
...
+                "where t1.codesum >=40 and t2.regionsum >= 90";
+
+            rs=conn.prepareStatement(sql).executeQuery();
+            assertTrue(rs.next());
+            assertTrue(rs.getInt(1) == 3);
+            assertTrue(rs.getInt(2) == 44);
+            assertTrue(rs.getInt(3) == 99);
+            assertTrue(!rs.next());
+
+            //test for if SubselectRewriter.isOuterOrderByNodesPrefixOfInner had take effect
```
twdsilva: The inner query on alias t1 orders by ```aid, codesum``` while the outer select order by ```aid desc```, will 
```SubselectRewriter.isOuterOrderByNodesPrefixOfInner```  still remove the outer order by?

File: phoenix-core/src/main/java/org/apache/phoenix/compile/JoinCompiler.java (hunk 1 above)
Comment (comnetwork) [reply phoenix-core/src/main/java/org/apache/phoenix/compile/JoinCompiler.java]: Yes, the postFilters which are pushed down to JoinCompiler.Table could always be pushed down into the subquery, I would omit the useless code to make it more clear.
By contrast, the postFilters of JoinCompiler.JoinTable should be filtered after the join is completed.

File: phoenix-core/src/it/java/org/apache/phoenix/end2end/SortMergeJoinMoreIT.java (hunk 2 above)
Comment (comnetwork) [reply phoenix-core/src/it/java/org/apache/phoenix/end2end/SortMergeJoinMoreIT.java]: Yes, I would omit it.

File: phoenix-core/src/it/java/org/apache/phoenix/end2end/SortMergeJoinMoreIT.java (hunk 3 above)
Comment (comnetwork) [reply phoenix-core/src/it/java/org/apache/phoenix/end2end/SortMergeJoinMoreIT.java]: No, the SubselectRewriter.isOuterOrderByNodesPrefixOfInner would not remove the outer order by, because what SubselectRewriter.isOuterOrderByNodesPrefixOfInner to check is the equality of OrderByNode, which consides the parseNode/nullsLast/orderAscending all together.

In fact, the RHS of the SortMergeJoin :
(select a.aid,sum(c.region) regionsum from "+tableName1+" a inner join "+tableName3+" c on a.aid=c.cid where c.region>=77 and c.region<=99 group by a.aid order by a.aid desc,regionsum desc limit 2)
is a test case, it would be rewritten as :
SELECT $1.AID,$1.REGIONSUM FROM (SELECT A.AID, SUM(C.REGION) REGIONSUM FROM MERGE1 A  Inner JOIN MERGE3 C  ON (A.AID = C.CID) WHERE (C.REGION >= 77 AND C.REGION <= 99) GROUP BY A.AID ORDER BY A.AID DESC,REGIONSUM DESC LIMIT 2) $1
 WHERE $1.REGIONSUM >= 90 ORDER BY $1.AID

the outer order by is not removed, the detailed test is in QueryCompilerTest.testSortMergeJoinPushFilterThroughSortBug5105

File: phoenix-core/src/it/java/org/apache/phoenix/end2end/SortMergeJoinMoreIT.java [hunk 4]
```diff
@@ -742,4 +742,112 @@ private static void verifyQueryPlanAndResultForBug4508(
...
+                    "AGE INTEGER"+
+                    ")";
+            conn.createStatement().execute(sql);
+
+            conn.createStatement().execute("UPSERT INTO "+tableName1+"(AID,AGE) VALUES (1,11)");
+            conn.createStatement().execute("UPSERT INTO "+tableName1+"(AID,AGE) VALUES (2,22)");
+            conn.createStatement().execute("UPSERT INTO "+tableName1+"(AID,AGE) VALUES (3,33)");
+            conn.commit();
+
+            conn.createStatement().execute("DROP TABLE if exists "+tableName2);
```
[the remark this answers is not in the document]
comnetwork: Yes, I would omit it.

gjacoby126:
JIRA resolved, closing PR.

PR #432, opened 2019-01-23
PHOENIX-5105

comnetwork:
This patch is mainly for:
1. following the suggestions of @twdsilva at #431 
2. omitted more useless code to make it more clear that the `postFilters` which are pushed down to `JoinCompiler.Table` could always be pushed down to subquery by wrapping as outer `whereParseNodes`.
By contrast, the `postFilters` of `JoinCompiler.JoinTable` should be filtered separately after the join is completed.


gjacoby126:
JIRA already resolved, closing PR.

PR #433, opened 2019-01-28
PHOENIX-4009 Run UPDATE STATISTICS command by using MR integration on…

karanmehta93:
… snapshots (Addendum)

### DIFF
--- a/phoenix-core/src/it/java/org/apache/phoenix/end2end/SpillableGroupByIT.java
+++ b/phoenix-core/src/it/java/org/apache/phoenix/end2end/SpillableGroupByIT.java
@@ -76,7 +76,7 @@ public class SpillableGroupByIT extends BaseOwnClusterIT {
         
         // Set guidepost width, but disable stats
         props.put(QueryServices.STATS_GUIDEPOST_WIDTH_BYTES_ATTRIB, Long.toString(20));
-        props.put(QueryServices.STATS_ENABLED_ATTRIB, Boolean.toString(false));
+        props.put(QueryServices.STATS_COLLECTION_ENABLED, Boolean.toString(false));
         props.put(QueryServices.EXPLAIN_CHUNK_COUNT_ATTRIB, Boolean.TRUE.toString());
         props.put(QueryServices.EXPLAIN_ROW_COUNT_ATTRIB, Boolean.TRUE.toString());
         // Must update config before starting server
--- a/phoenix-core/src/it/java/org/apache/phoenix/schema/stats/NoOpStatsCollectorIT.java
+++ b/phoenix-core/src/it/java/org/apache/phoenix/schema/stats/NoOpStatsCollectorIT.java
@@ -17,17 +17,18 @@
  */
 package org.apache.phoenix.schema.stats;
 
-import com.google.common.base.Throwables;
-import com.google.common.collect.Iterables;
+import com.google.common.collect.Maps;
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
-import org.apache.hadoop.hbase.security.AccessDeniedException;
 import org.apache.phoenix.end2end.ParallelStatsDisabledIT;
+import org.apache.phoenix.query.QueryServices;
 import org.apache.phoenix.util.PropertiesUtil;
+import org.apache.phoenix.util.ReadOnlyProps;
 import org.apache.phoenix.util.SchemaUtil;
 import org.apache.phoenix.util.TestUtil;
 import org.junit.Assert;
 import org.junit.Before;
+import org.junit.BeforeClass;
 import org.junit.Test;
 
 import java.sql.Array;
@@ -37,12 +38,14 @@ import java.sql.PreparedStatement;
 import java.sql.ResultSet;
 import java.sql.SQLException;
 import java.sql.Statement;
+import java.util.Map;
 import java.util.Properties;
 
 import static org.apache.phoenix.util.TestUtil.TEST_PROPERTIES;
 
 /**
  * Tests the behavior of stats collection code when stats are disabled on server side
+ * explicitly using QueryServices#STATS_COLLECTION_ENABLED property
  */
 public class NoOpStatsCollectorIT extends ParallelStatsDisabledIT {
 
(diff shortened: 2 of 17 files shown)


phoenix-core/src/main/java/org/apache/phoenix/exception/SQLExceptionCode.java
```diff
@@ -464,6 +464,8 @@ public SQLException newException(SQLExceptionInfo info) {
     INSUFFICIENT_MEMORY(999, "50M01", "Unable to allocate enough memory."),
     HASH_JOIN_CACHE_NOT_FOUND(900, "HJ01", "Hash Join cache not found"),
 
+    STATS_COLLECTION_DISABLED_ON_SERVER(1401, "STS01", "Stats collection attempted but is disabled on server"),
```
dbwong: In general the error codes appear to be grouped by number/purpose.  Do we have system level error code or stats codes?  Do we want to reserve a prefix range for this 14XX?  Consider moving to a different location for sorting reasons.

phoenix-core/src/it/java/org/apache/phoenix/schema/stats/NoOpStatsCollectorIT.java
```diff
@@ -0,0 +1,138 @@
...
+import java.sql.SQLException;
+import java.sql.Statement;
+import java.util.Properties;
+
+import static org.apache.phoenix.util.TestUtil.TEST_PROPERTIES;
+
+/**
+ * Tests the behavior of stats collection code when stats are disabled on server side
+ */
+public class NoOpStatsCollectorIT extends ParallelStatsDisabledIT {
```
dbwong: This class tests a bunch of negative cases.  Do we have good coverage on the positive cases such that we are confident the removal of STATS_ENABLED_ATTRIB does not break anything?

karanmehta93: The property was deprecated more than 2 releases ago. As of now I have removed its total usage. 
I don't think that it would break anything.

phoenix-core/src/main/java/org/apache/phoenix/exception/SQLExceptionCode.java
karanmehta93: > In general the error codes appear to be grouped by number/purpose. Do we have system level error code or stats codes?

Yes thats correct, there are some conventions. I am not sure if these are followed here. Phoenix has a bunch of its own exception types not really relevant to SQL.

> Do we want to reserve a prefix range for this 14XX? 

I am not sure if we can do that and if yes, whats the way to do that. 

> Consider moving to a different location for sorting reasons

Will do

phoenix-core/src/it/java/org/apache/phoenix/end2end/SpillableGroupByIT.java
```diff
@@ -76,7 +76,7 @@ public static void doSetup() throws Exception {
         
         // Set guidepost width, but disable stats
         props.put(QueryServices.STATS_GUIDEPOST_WIDTH_BYTES_ATTRIB, Long.toString(20));
-        props.put(QueryServices.STATS_ENABLED_ATTRIB, Boolean.toString(false));
+        props.put(QueryServices.STATS_COLLECTION_ENABLED, Boolean.toString(false));
```
BinShi-SecularBird: Is it better to also set STATS_ENABLED_ATTRIB to be false here, although it is deprecated.

phoenix-core/src/main/java/org/apache/phoenix/exception/SQLExceptionCode.java
dbwong: @twdsilva Are you aware of any conventions we are trying to hold for SQL exceptions?

BinShi-SecularBird: I'm very confusion about the naming convention we used for the first two fields, eg., "1401" and "STS01". The last 4 sql exceptions defined are totally out of orders which mightn't be good.

phoenix-core/src/main/java/org/apache/phoenix/query/GuidePostsCache.java
```diff
@@ -70,8 +69,7 @@ public GuidePostsCache(ConnectionQueryServices queryServices, Configuration conf
         final long maxTableStatsCacheSize = config.getLong(
                 QueryServices.STATS_MAX_CACHE_SIZE,
                 QueryServicesOptions.DEFAULT_STATS_MAX_CACHE_SIZE);
-		final boolean isStatsEnabled = config.getBoolean(STATS_COLLECTION_ENABLED, DEFAULT_STATS_COLLECTION_ENABLED)
-				&& config.getBoolean(STATS_ENABLED_ATTRIB, true);
+		final boolean isStatsEnabled = config.getBoolean(STATS_COLLECTION_ENABLED, DEFAULT_STATS_COLLECTION_ENABLED);
```
BinShi-SecularBird: alignment problem.

phoenix-core/src/main/java/org/apache/phoenix/schema/stats/StatsCollectionDisabledOnServerException.java
```diff
@@ -0,0 +1,18 @@
+package org.apache.phoenix.schema.stats;
+
+import org.apache.phoenix.exception.SQLExceptionCode;
+import org.apache.phoenix.exception.SQLExceptionInfo;
+
+import java.sql.SQLException;
+
+public class StatsCollectionDisabledOnServerException extends SQLException {
+
```
BinShi-SecularBird: remove empty line here?

phoenix-core/src/main/java/org/apache/phoenix/schema/stats/StatsCollectionDisabledOnServerException.java
```diff
@@ -0,0 +1,18 @@
...
+    private static final long serialVersionUID = 1L;
+    private static SQLExceptionCode code = SQLExceptionCode.STATS_COLLECTION_DISABLED_ON_SERVER;
+
+    public StatsCollectionDisabledOnServerException() {
```
BinShi-SecularBird: It's better to pass table name and call setTableNae in SQLExceptionInfo.Builder.

karanmehta93: The exception is wrapped around by `DoNotRetryIOException` from HBase which prints the region name automatically (which contains table name), hence I didn't add it here.

PR #434, opened 2019-01-30
PHOENIX-5018 Index mutations created by UPSERT SELECT will have wrong…

kadirozde:
… timestamps

### DIFF
--- a/phoenix-core/src/it/java/org/apache/phoenix/end2end/IndexBuildTimestampIT.java
+++ b/phoenix-core/src/it/java/org/apache/phoenix/end2end/IndexBuildTimestampIT.java
@@ -1,0 +1,248 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.phoenix.end2end;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertNotNull;
+import static org.junit.Assert.assertTrue;
+
+import java.sql.Connection;
+import java.sql.DriverManager;
+import java.sql.ResultSet;
+import java.sql.SQLException;
+import java.util.Collection;
+import java.util.List;
+import java.util.Properties;
+
+import org.apache.hadoop.hbase.client.ResultScanner;
+import org.apache.hadoop.hbase.client.Scan;
+import org.apache.hadoop.hbase.client.Table;
+import org.apache.phoenix.jdbc.PhoenixConnection;
+import org.apache.phoenix.jdbc.PhoenixResultSet;
+import org.apache.phoenix.query.QueryServices;
+import org.apache.phoenix.schema.PTable;
+import org.apache.phoenix.util.EnvironmentEdge;
+import org.apache.phoenix.util.EnvironmentEdgeManager;
+import org.apache.phoenix.util.PhoenixRuntime;
+import org.apache.phoenix.util.QueryUtil;
+import org.junit.BeforeClass;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.junit.runners.Parameterized;
+import org.junit.runners.Parameterized.Parameters;
+import com.google.common.collect.Lists;
+
+@RunWith(Parameterized.class)
+public class IndexBuildTimestampIT extends BaseUniqueNamesOwnClusterIT {
+    private final boolean localIndex;
+    private final boolean async;
+    private final boolean view;
+    private final String tableDDLOptions;
+
+    public IndexBuildTimestampIT(boolean mutable, boolean localIndex,
+                            boolean async, boolean view) {
+        this.localIndex = localIndex;
+        this.async = async;
+        this.view = view;
+        StringBuilder optionBuilder = new StringBuilder();
+        if (!mutable) {
+            optionBuilder.append(" IMMUTABLE_ROWS=true ");
+        }
+        optionBuilder.append(" SPLIT ON(1,2)");
+        this.tableDDLOptions = optionBuilder.toString();
+    }
+
+    @BeforeClass
+    public static void setup() throws Exception {
+        IndexToolIT.setup();
+    }
+
+    @Parameters(
+            name = "mutable={0},localIndex={1},async={2},view={3}")
+    public static Collection<Object[]> data() {
+        List<Object[]> list = Lists.newArrayListWithExpectedSize(8);
+        boolean[] Booleans = new boolean[]{false, true};
+        for (boolean mutable : Booleans) {
+            for (boolean localIndex : Booleans) {
+                for (boolean async : Booleans) {
+                    for (boolean view : Booleans) {
+                        list.add(new Object[]{mutable, localIndex, async, view});
+                    }
+                }
+            }
+        }
+        return list;
+    }
+
+    public static void assertExplainPlan(Connection conn, boolean localIndex, String selectSql,
+                                         String dataTableFullName, String indexTableFullName) throws SQLException {
+        ResultSet rs = conn.createStatement().executeQuery("EXPLAIN " + selectSql);
+        String actualExplainPlan = QueryUtil.getExplainPlan(rs);
+
+        IndexToolIT.assertExplainPlan(localIndex, actualExplainPlan, dataTableFullName, indexTableFullName);
+    }
+
+    private class MyClock extends EnvironmentEdge {
+        long initialTime;
+        long delta;
+
+        public MyClock(long delta) {
+            initialTime = System.currentTimeMillis() + delta;
+            this.delta = delta;
+        }
+
+        @Override
+        public long currentTime() {
+            return System.currentTimeMillis() + delta;
+        }
+
+        public long initialTime() {
+            return initialTime;
+        }
+    }
+
+    private void populateTable(String tableName, MyClock clock1, MyClock clock2) throws Exception {
+        Connection conn = DriverManager.getConnection(getUrl());
+        conn.createStatement().execute("create table " + tableName +
+                " (id varchar(10) not null primary key, val varchar(10), ts timestamp)" + tableDDLOptions);
+
+        EnvironmentEdgeManager.injectEdge(clock1);
+        conn.createStatement().execute("upsert into " + tableName + " values ('aaa', 'abc', current_date())");
+        conn.commit();
+
+        EnvironmentEdgeManager.injectEdge(clock2);
+        conn.createStatement().execute("upsert into " + tableName + " values ('bbb', 'bcd', current_date())");
+        conn.commit();
+        conn.close();
+
+        Properties props = new Properties();
+        props.setProperty("CurrentSCN", Long.toString(clock1.initialTime()));
+        conn = DriverManager.getConnection(getUrl(), props);
+        ResultSet rs = conn.createStatement().executeQuery("select * from " + tableName);
+        assertFalse(rs.next());
+        conn.close();
+
+        props.setProperty("CurrentSCN", Long.toString(clock2.initialTime()));
+        conn = DriverManager.getConnection(getUrl(), props);
+        rs = conn.createStatement().executeQuery("select * from " + tableName);
+
+        assertTrue(rs.next());
+        assertEquals("aaa", rs.getString(1));
+        assertEquals("abc", rs.getString(2));
+        assertNotNull(rs.getDate(3));
+
+        assertFalse(rs.next());
+        conn.close();
+
+        props.setProperty("CurrentSCN", Long.toString(clock2.currentTime()));
+        conn = DriverManager.getConnection(getUrl(), props);
+        rs = conn.createStatement().executeQuery("select * from " + tableName);
+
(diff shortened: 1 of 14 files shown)


phoenix-core/src/it/java/org/apache/phoenix/end2end/IndexBuildTimestampIT.java
```diff
@@ -0,0 +1,299 @@
...
+        if (!mutable) {
+            optionBuilder.append(" IMMUTABLE_ROWS=true ");
+        }
+        optionBuilder.append(" SPLIT ON(1,2)");
+        this.tableDDLOptions = optionBuilder.toString();
+    }
+
+    @BeforeClass
+    public static void setup() throws Exception {
+        Map<String, String> serverProps = Maps.newHashMapWithExpectedSize(2);
```
gokceni: nit: serverProps seems to have 3 elements. Should not this be 3 instead of 2?

phoenix-core/src/it/java/org/apache/phoenix/end2end/IndexBuildTimestampIT.java
```diff
@@ -0,0 +1,299 @@
...
+            for (boolean localIndex : Booleans) {
+                for (boolean async : Booleans) {
+                    list.add(new Object[]{mutable, localIndex, async});
+                }
+            }
+        }
+        return list;
+    }
+
+    public static void assertExplainPlan(Connection conn, boolean localIndex, String selectSql,
```
gokceni: I see that IndexToolIT has the same method with this name. Can we have a utility class and copy this one there and use from both places?

phoenix-core/src/it/java/org/apache/phoenix/end2end/IndexBuildTimestampIT.java
```diff
@@ -0,0 +1,299 @@
...
+        return args.toArray(new String[0]);
+    }
+
+    public static void runIndexTool(String schemaName,
+                                    String dataTableName, String indexTableName) throws Exception {
+        runIndexTool(schemaName, dataTableName, indexTableName, new String[0]);
+    }
+
+    public static void runIndexTool(String schemaName,
+                                    String dataTableName, String indexTableName, String... additionalArgs) throws Exception {
```
gokceni: Ditto for runIndexTool and getArgs existing in IndexToolIT.
Also, do we need the tenantID parameter test as well? I have a pending PR where we add tenantId as cmd line parameter.

phoenix-core/src/main/java/org/apache/phoenix/compile/ServerBuildIndexCompiler.java
```diff
@@ -0,0 +1,127 @@
...
+            TableRef tableRef = plan.getTableRef();
+            Scan scan = plan.getContext().getScan();
+            ImmutableBytesWritable ptr = new ImmutableBytesWritable();
+            final PTable dataTable = tableRef.getTable();
+            IndexMaintainer.serialize(dataTable, ptr, Collections.singletonList(index), plan.getContext().getConnection());
+            // Set attribute on scan that UngroupedAggregateRegionObserver will switch on.
+            // We'll detect that this attribute was set the server-side and write the index
+            // rows per region as a result. The value of the attribute will be our persisted
+            // index maintainers.
+            // Define the LOCAL_INDEX_BUILD as a new static in BaseScannerRegionObserver
```
gokceni: nit: Could you add to the comment what is the attribute on scan to switch on UngroupedAggregateRegionObserver? 
Also, the comment mentions LOCAL_INDEX_BUILD but in code LOCAL_INDEX_BUILD_PROTO is used. Is there a mismatch?

phoenix-core/src/it/java/org/apache/phoenix/end2end/IndexBuildTimestampIT.java
kadirozde: I will use IndexToolIT.setup

phoenix-core/src/it/java/org/apache/phoenix/end2end/IndexBuildTimestampIT.java
kadirozde: I will use the method of IndexToolIT

phoenix-core/src/it/java/org/apache/phoenix/end2end/IndexBuildTimestampIT.java
kadirozde: I will leverage IndexToolIT methods. This test does not need to use tenantID.

phoenix-core/src/main/java/org/apache/phoenix/compile/ServerBuildIndexCompiler.java
kadirozde: I will update the comments

phoenix-core/src/it/java/org/apache/phoenix/end2end/IndexBuildTimestampIT.java
kadirozde: Done

phoenix-core/src/it/java/org/apache/phoenix/end2end/IndexBuildTimestampIT.java
kadirozde: Done

phoenix-core/src/main/java/org/apache/phoenix/compile/ServerBuildIndexCompiler.java
kadirozde: Done

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/index/PhoenixServerBuildIndexMapper.java
```diff
@@ -0,0 +1,75 @@
...
+import org.apache.phoenix.query.QueryServicesOptions;
+import org.apache.phoenix.util.ColumnInfo;
+import org.apache.phoenix.util.PhoenixRuntime;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * Mapper that does not do much as regions servers actually build the index from the data table regions directly
+ */
+public class PhoenixServerBuildIndexMapper extends
```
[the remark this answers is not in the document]
kadirozde: They are still used for building transactional global indexes as this patch does not fix the timestamp issue for them.

phoenix-core/src/it/java/org/apache/phoenix/end2end/IndexBuildTimestampIT.java
```diff
@@ -0,0 +1,236 @@
...
+
+        // Verify the index timestamps via Phoenix
+        String selectSql = String.format("SELECT * FROM %s WHERE val = 'abc'", dataTableName);
+        conn = DriverManager.getConnection(getUrl());
+        // assert we are pulling from index table
+        assertExplainPlan(conn, localIndex, selectSql, dataTableName, indexName);
+
+        ResultSet rs = conn.createStatement().executeQuery(selectSql);
+        assertTrue (rs.next());
+        assertTrue(rs.unwrap(PhoenixResultSet.class).getCurrentRow().getValue(0).getTimestamp() >= clock1.initialTime());
```
gjacoby126: @kadirozde - maybe I'm misunderstanding the test, but shouldn't this be checking against an _upper_ bound to make sure that the index was populated with the initial timestamp of 'abc' (and in the second check 'bcd')?

phoenix-core/src/main/java/org/apache/phoenix/compile/ServerBuildIndexCompiler.java
```diff
@@ -0,0 +1,133 @@
...
+                }
+            }
+
+            if (dataTable.isTransactional()) {
+                scan.setAttribute(BaseScannerRegionObserver.TX_STATE, connection.getMutationState().encodeTransaction());
+            }
+
+            // Go through MutationPlan abstraction so that we can create local indexes
+            // with a connectionless connection (which makes testing easier).
+            return new BaseMutationPlan(plan.getContext(), PhoenixStatement.Operation.UPSERT) {
```
gjacoby126: nit: I'm trying to get rid of non-trivial anonymous classes in Phoenix, so let's not add more :-)

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/index/IndexTool.java
```diff
@@ -443,17 +448,44 @@ private Job configureJobForAysncIndex(String schemaName, String indexTable, Stri
...
+            PhoenixConfigurationUtil.setDisableIndexes(configuration, indexTable);
+
+            fs = outputPath.getFileSystem(configuration);
+            fs.delete(outputPath, true);
+
+            final String jobName = String.format(INDEX_JOB_NAME_TEMPLATE, schemaName, dataTable, indexTable);
+            final Job job = Job.getInstance(configuration, jobName);
+            job.setJarByClass(IndexTool.class);
+            job.setMapOutputKeyClass(ImmutableBytesWritable.class);
+            FileOutputFormat.setOutputPath(job, outputPath);
```
gjacoby126: Can dispense with the output path, I think

phoenix-core/src/it/java/org/apache/phoenix/end2end/IndexBuildTimestampIT.java
```diff
@@ -0,0 +1,236 @@
...
+        assertNotNull(rs.getDate(3));
+
+        assertTrue(rs.next());
+        assertEquals("bbb", rs.getString(1));
+        assertEquals("bcd", rs.getString(2));
+        assertNotNull(rs.getDate(3));
+        assertFalse(rs.next());
+        conn.close();
+    }
+
```
gjacoby126: Would be good to have a unit test for IndexTool / JobFactory that for different input parameters asserts that the correct mapper class is configured. (Don't have to actually run the job, just configure it.)

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/util/PhoenixConfigurationUtil.java
```diff
@@ -122,6 +122,10 @@
 
     public static final String SCRUTINY_INDEX_TABLE_NAME = "phoenix.mr.scrutiny.index.table.name";
 
+    public static final String INDEX_TOOL_DATA_TABLE_NAME = "phoenix.mr.index_tool.data.table.name";
```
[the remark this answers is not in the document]
kadirozde: Ok

phoenix-core/src/it/java/org/apache/phoenix/end2end/IndexBuildTimestampIT.java
kadirozde: Thx for noticing it. I will correct it.

phoenix-core/src/main/java/org/apache/phoenix/compile/ServerBuildIndexCompiler.java
kadirozde: I will convert it to an inner class

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/index/IndexTool.java
kadirozde: When I remove the code for setting the output path, there is always one other test failing. There are total of 120 tests. I tried restructuring the IndexTool code so that the output path is set only when bulk loading is enabled but did not help. So, I will leave the code as it is.

phoenix-core/src/it/java/org/apache/phoenix/end2end/TableDDLPermissionsIT.java
```diff
@@ -211,23 +211,14 @@ public Void run() throws Exception {
 
             // we should be able to read the data from another index as well to which we have not given any access to
             // this user
-            verifyAllowed(createIndex(indexName2, phoenixTableName), unprivilegedUser);
```
gjacoby126: @kadirozde - I'm curious what the purpose of these changes are. Have DDL permissions changed? If not, and your changes somehow break the existing test, is there a way to modify the test logic to keep the same test coverage?

kadirozde: Currently global indexes are built by using UPSERT SELECT, i..e, by reading mutations from the data table and writing the corresponding mutations to the index table. The user does not need to have the write privilege on the data table in order to build an index table. However, in the new approach, an index table is built by replaying the existing mutations back on the data table again (without actually applying them to the data table) for the purpose of reconstructing the index mutations. Such replay attempts are allowed only when the user has the write privilege. Therefore, I had to delete the scenarios that are not applicable anymore.

phoenix-core/src/main/java/org/apache/phoenix/schema/MetaDataClient.java
```diff
@@ -1741,6 +1743,13 @@ public MutationState createIndex(CreateIndexStatement statement, byte[][] splits
         if (connection.getSCN() != null) {
             return buildIndexAtTimeStamp(table, statement.getTable());
         }
+
+        String dataTableFullName = SchemaUtil.getTableName(
+                tableRef.getTable().getSchemaName().getString(),
+                tableRef.getTable().getTableName().getString());
```
gjacoby126: @kadirozde - curious why the refresh is needed here? The tableRef only seems to be used to get the base table name to give to the ServerBuildIndexDDLCompiler

kadirozde: I needed this for the initial version of the compiler where the index table was searched in the index list of the PTable for the data table. MetadataClient constructs the PTable object for the data table before the index table is created and thus the object does have a reference to the newly created index. The complier has changed and this search was eliminated. So there is no need to refresh the PTable object anymore. I will remove the refresh code.

gjacoby126: Thanks for explaining, @kadirozde . Once the refresh code gets removed, I think this is ready to go so I'll +1 and commit.

kadirozde: @gjacoby126, I have removed the refresh code. Thank you for discovering this issue, helping me understand the code, and reviewing the patch!

PR #439, opened 2019-02-11
PHOENIX-2265 Disallow creation of view over HBase table if PK not spe…

yanxinyi:
We currently allow a Phoenix view to be defined over an HBase table without specifying a primary key. 
Fixed bug by adding pk check at CreateTableCompiler.compile. If no PK specified, it will throw a PK missing exception error.

### DIFF
--- a/phoenix-core/src/main/java/org/apache/phoenix/expression/RowValueConstructorExpression.java
+++ b/phoenix-core/src/main/java/org/apache/phoenix/expression/RowValueConstructorExpression.java
@@ -28,6 +28,7 @@ import java.io.DataInput;
 import java.io.DataOutput;
 import java.io.IOException;
 import java.util.Arrays;
+import java.util.BitSet;
 import java.util.List;
 
 import org.apache.hadoop.hbase.io.ImmutableBytesWritable;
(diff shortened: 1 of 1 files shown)


phoenix-core/src/main/java/org/apache/phoenix/compile/CreateTableCompiler.java
```diff
@@ -128,6 +128,24 @@ public MutationPlan compile(CreateTableStatement create) throws SQLException {
             viewTypeToBe = parentToBe.getViewType() == ViewType.MAPPED ? ViewType.MAPPED : ViewType.UPDATABLE;
             if (whereNode == null) {
                 viewStatementToBe = parentToBe.getViewStatement();
+
```
dbwong: This code is in a whereNode==null condition.  What if my view had a simple where clause in it would we allow the construction?

yanxinyi: didn't look at the Phoenix supported gramma carefully, I moved code outside wherenode check ifelse block.

yanxinyi:
@twdsilva @karanmehta93 @ChinmaySKulkarni 
pls take look when u guys have time, thanks!

ChinmaySKulkarni:
A couple of nits for the tests. Overall lgtm. Thanks @yanxinyi

yanxinyi:
combined multi-commits into one.

yanxinyi:
followed @gjacoby126 suggestion and moved the test to ViewIT file.

PR #441, opened 2019-02-15
PHOENIX-5089 Add tenantId parameter to IndexScrunityTool

gokceni:


gokceni:
@gjacoby126 @kadirozde

phoenix-core/src/it/java/org/apache/phoenix/end2end/IndexScrutinyToolIT.java
```diff
@@ -551,11 +558,132 @@ public void testMaxOutputRows() throws Exception {
...
+            PhoenixRuntime.getTable(conn, MULTI_TENANT_DATA_TABLE_NAME);
+            doesTableExist = true;
+        } catch (TableNotFoundException e) {
+            doesTableExist = false;
+        }
+
+        // We already run this test, return.
+        if (doesTableExist) {
+            return;
+        }
```
kadirozde: This test does not use the test parameters and does not follow the overall design within IndexScrutinyToolIT. I suggest either refactoring it to fit in here or implementing it a separate test class.

phoenix-core/src/test/java/org/apache/phoenix/mapreduce/index/IndexScrutinyTableOutputTest.java
```diff
@@ -43,7 +43,7 @@ public void testConstructMetadataParamQuery() {
     @Test
     public void testGetSqlQueryAllInvalidRows() throws SQLException {
         SourceTargetColumnNames columnNames =
-                new SourceTargetColumnNames.DataSourceColNames(pDataTable, pIndexTable);
+                new SourceTargetColumnNames.DataSourceColNames(pDataTable, pIndexTable, false);
```
kadirozde: Having a new constructor that takes the extra boolean parameter would eliminate changing the existing code.

gokceni: ok

phoenix-core/src/it/java/org/apache/phoenix/end2end/IndexScrutinyToolIT.java
gokceni: will refactor

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/util/IndexColumnNames.java
```diff
@@ -61,6 +68,9 @@ public IndexColumnNames(final PTable pdataTable, final PTable pindexTable) {
         if (pindexTable.getViewIndexId() != null) {
             offset++;
         }
+        if (forTenant && pindexTable.isMultiTenant()) {
```
gjacoby126: @gokceni - Would the forTenant make a difference here? What if it were a multi-tenant global view index? The _IDX table would still be <ViewIndexId, TenantId, Other indexed columns>, but presumably forTenant would be false.

gokceni: Sure. I did not want to break it for the existing code. That is why I added forTenant. Will remove

gjacoby126: Should probably also have a test for a global view index on a multi-tenant table to make sure that combination works, if we don't already have one.

phoenix-core/src/it/java/org/apache/phoenix/end2end/IndexScrutinyToolIT.java
```diff
@@ -81,681 +86,844 @@
...
-        int status = scrutiny.run(cmdArgs);
-        assertEquals(0, status);
-        for (Job job : scrutiny.getJobs()) {
-            assertTrue(job.waitForCompletion(true));
-        }
-        return scrutiny.getJobs();
-    }
+                ///////
+                // Global view on multi-tenant table should work too
+                connGlobal.createStatement().execute(
```
gjacoby126: @gokceni  Would be good to divide this big testTenantId up into several smaller tests, since it's going through several distinct use cases of "create / populate table", "run scrutiny", then "verify scrutiny". This would make it more similar to the other non-tenant test suite and give developers more clarity about what's failing if a test fails.

Other than that, I think this is just about ready to go.

phoenix-core/src/it/java/org/apache/phoenix/end2end/IndexScrutinyToolIT.java
```diff
@@ -81,681 +86,850 @@
...
+
+            String idxStmtGlobal = String.format(createIndexStr, indexNameGlobal, globalViewName);
+            connGlobal.createStatement().execute(idxStmtGlobal);
+            connGlobal.createStatement()
+                    .execute(String.format(upsertQueryStr, globalViewName, "global", 5, "x"));
+            connGlobal.commit();
+            String[] argValues =
+                    getArgValues("", globalViewName, indexNameGlobal, 10L,
+                            SourceTable.INDEX_TABLE_SOURCE, false, null, null, null,
+                            En
```
gjacoby126: @gokceni - could you please file a followup JIRA to allow for output to table when using a tenant id connection.

gjacoby126:
+1, thanks @gokceni

kadirozde:
+1, thanks for the updates @gokceni

phoenix-core/src/it/java/org/apache/phoenix/end2end/IndexScrutinyToolIT.java
gokceni: Done https://issues.apache.org/jira/browse/PHOENIX-5168

PR #442, opened 2019-02-20
PHOENIX-5122: PHOENIX-4322 breaks client backward compatibility

jpisaac:


jpisaac:
@twdsilva @dbwong Please review.

jpisaac:
@ChinmaySKulkarni

phoenix-core/src/main/java/org/apache/phoenix/expression/RowValueConstructorExpression.java
```diff
@@ -46,14 +47,30 @@
...
     private int partialEvalIndex = -1;
     private int estimatedByteSize;
+    
+    // The boolean field that indicated the object is a literal constant, 
+    // has been repurposed to a bitset and now holds additional information. 
+    // This is to facilitate b/w compat to 4.13 clients.
+    // @see <a href="https://issues.apache.org/jira/browse/PHOENIX-5122">PHOENIX-5122</a> 
+    private BitSet extraFields;
+    
+    private enum ExtraFieldPosition {
```
dbwong: I dislike the ordinal Enum approach as it makes this enum order dependant and for example adding a field to the beginning here could cause backwards compatibility problems as well.  I'd prefer a enum constant with values.

jpisaac: I think adding enum constant with values does not help with backward compatibility - since we cannot alter the bit positions anyway. The only benefit of enum with values is for placement in the definition. 
Not sure if there is any value in that, since placement should match with ordinals ideally.

jpisaac: I relied on the current unit tests to ascertain that it is functionally the same for > 4.14 client.

jpisaac: @dbwong @ChinmaySKulkarni 
Let me know if we need more testing than -

org.apache.phoenix.end2end.InListIT
org.apache.phoenix.end2end.SortOrderIT
org.apache.phoenix.end2end.RowValueConstructorIT

jpisaac:
@dbwong @ChinmaySKulkarni

dbwong:
As we discussed about larger work for the backward compatability framework.  Did we want to handle this in the opensource world?  A JIRA, or something?   Other than that looks good to me. +1

jpisaac:
@dbwong Yes, I agree we need a b/w compatibility test setup, Let's explore if we can use what @mujtabachohan had worked on?

PR #443, opened 2019-02-21
PHOENIX-5136

hnguyen08:
`NULL` values for rows that are inserted using `UPSERT ON DUPLICATE KEY UPDATE` will be stored in HBase as empty byte arrays.  `NULL` values represented by empty byte arrays are not handled well in the `AndOrOperator.evaluate()`.  Any child expression of the `AndOrOperator` that evaluates to `NULL` represented by empty byte array will get marked in `partialEvalState`, thus having no effect when the `AndOrOperator` is re-evaluated against a different cell from the same row.  This causes problems because the child expression *should* be re-evaluated since it can still affect the final result.

The fix ensures that `partialEvalState` is not marked if the child expression evaluates to a NULL value represented by empty array.

Let me know if this fix is acceptable or if there is a better approach to this problem!

### DIFF
--- a/phoenix-core/src/it/java/org/apache/phoenix/end2end/OnDuplicateKeyIT.java
+++ b/phoenix-core/src/it/java/org/apache/phoenix/end2end/OnDuplicateKeyIT.java
@@ -580,6 +580,40 @@ public class OnDuplicateKeyIT extends ParallelStatsDisabledIT {
         }
     }
 
+    @Test
+    public void testRowsCreatedViaUpsertOnDuplicateKeyShouldNotBeReturnedInQueryIfNotMatched() throws Exception {
+        Properties props = PropertiesUtil.deepCopy(TEST_PROPERTIES);
+        Connection conn = DriverManager.getConnection(getUrl(), props);
+        String tableName = generateUniqueName();
+        String ddl = " create table " + tableName + "(pk varchar primary key, counter1 bigint, counter2 smallint)";
+        conn.createStatement().execute(ddl);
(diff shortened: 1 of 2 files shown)


phoenix-core/src/it/java/org/apache/phoenix/end2end/OnDuplicateKeyIT.java
```diff
@@ -580,6 +580,40 @@ public void testDuplicateUpdateWithSaltedTable() throws Exception {
...
+        String dml2 = "UPSERT INTO " + tableName + " VALUES('b',1,2) ON DUPLICATE KEY UPDATE " +
+                "counter1 = CASE WHEN (counter1 IS NULL) THEN 1 ELSE counter1 END, " +
+                "counter2 = CASE WHEN (counter1 IS NULL) THEN 2 ELSE counter2 END";
+        conn.createStatement().execute(dml2);
+        conn.commit();
+
+        // Using this statement causes the test to pass
+        //ResultSet rs = conn.createStatement().executeQuery("SELECT * FROM " + tableName + " WHERE counter2 = 2 AND counter1 = 1");
+        // This statement should be equivalent to the one above, but it selects both rows.
+        ResultSet rs = conn.createStatement().executeQuery("SELECT * FROM " + tableName + " WHERE counter2 = 2 AND (counter1 = 1 OR counter1 = 1)");
```
m2je: Use explicit column name: "Select pk, counter1, counter2 "

hnguyen08:
Closing. Re-implemented as #446

PR #444, opened 2019-02-21
PHOENIX-5148

comnetwork:


twdsilva:
@comnetwork  Thanks for the detailed explanation of your approach on the JIRA. I am reviewing the patch.

phoenix-core/src/main/java/org/apache/phoenix/compile/QueryPlan.java
```diff
@@ -92,4 +94,17 @@
...
+     * Get the actual OrderBys of this queryPlan, which may be different from {@link #getOrderBy()},
+     * because {@link #getOrderBy()} is only the compiled result of {@link SelectStatement}.
+     * The return type is List because we can get multiple OrderBys for the query result of {@link SortMergeJoinPlan},
+     * eg. for the sql:
+     * SELECT  * FROM T1 JOIN T2 ON T1.a = T2.a and T1.b = T2.b
+     * The result of the sort-merge-join is sorted on (T1.a, T1.b) and (T2.a, T2.b) at the same time.
+     * </pre>
+     * @return
+     */
+    public List<OrderBy> getActualOutputOrderBys() ;
```
twdsilva: There would only be multiple OrderBys in the case of a join, right?

twdsilva: Also,  maybe we can come up with a better API name like ``getPhysicalOrderBy`` or ``getFinalOrderBy``, WDYT?

twdsilva:
Overall looks pretty good. @dbwong or @BinShi-SecularBird would be good if you guys review this as well.

phoenix-core/src/main/java/org/apache/phoenix/compile/GroupByCompiler.java
```diff
@@ -63,9 +63,10 @@
         private final boolean isOrderPreserving;
```
twdsilva: Do you need this boolean any more? Seems like we always should use ```orderPreservingTrackInfos``` to determine if its order preseving right?

phoenix-core/src/main/java/org/apache/phoenix/compile/OrderPreservingTracker.java
```diff
@@ -9,29 +9,37 @@
...
 import org.apache.phoenix.expression.visitor.StatelessTraverseNoExpressionVisitor;
+import org.apache.phoenix.jdbc.PhoenixConnection;
 import org.apache.phoenix.schema.PTable;
-import org.apache.phoenix.schema.SortOrder;
+import org.apache.phoenix.schema.TableRef;
 import org.apache.phoenix.util.ExpressionUtil;
 
+import com.google.common.collect.ImmutableList;
 import com.google.common.collect.Iterators;
 import com.google.common.collect.Lists;
```
twdsilva: Can you explain the changes you had to make to this class, its a little hard to follow all the changes?

phoenix-core/src/main/java/org/apache/phoenix/execute/AggregatePlan.java
```diff
@@ -327,4 +333,15 @@ public boolean useRoundRobinIterator() throws SQLException {
         return visitor.visit(this);
     }
 
+    private static OrderBy convertActualOutputOrderBy(OrderBy orderBy, GroupBy groupBy, StatementContext statementContext) {
```
twdsilva: Does every plan have to implement this function, with different logic?

phoenix-core/src/main/java/org/apache/phoenix/compile/QueryPlan.java
comnetwork: Yes,now there would only be multiple OrderBys for SortMergeJoin.

Yes,I think the name `getFinalOrderBy` is OK , I named it as 'getActualOutputOrderBys' because
the equivalent name in Spark SQL's SparkPlan is `outputOrdering`, FYI.

phoenix-core/src/main/java/org/apache/phoenix/compile/GroupByCompiler.java
comnetwork: Yes , I think the member variable `isOrderPreserving` is required because it saved the result of `OrderPreservingTracker.isOrderPreserving()`.  `GroupBy.isOrderPreserving()` is used by many classes under compile and execute package.
FYI, the `GroupBy` class is a immutable class , when we compile the `group by clause` at difference time(we compile it twice in `QueryCompiler.compileSingleFlatQuery`) , a new GroupBy instance  is created.

phoenix-core/src/main/java/org/apache/phoenix/compile/OrderPreservingTracker.java
comnetwork: 1. I added a getInputOrderBys method to determine the input OrderBys by the combination of innerQueryPlan's output OrderBys ， GroupBy of current QueryPlan or the rowKeyColumns of current table.

2. Because the innerQueryPlan may have multiple output OrderBys(for `SortMergeJoinPlan`), so I extracted many stateful memeber variables in `OrderPreservingTracker`(such as `orderPreservingTrackInfos`, `isOrderPreserving` and `isReverse` etc.) to a new inner class `TrackOrderByContext`, `TrackOrderByContext` is used to track if a single Input OrderBy matchs
the target OrderByExpressions in `OrderPreservingTracker.isOrderPreserving()`, once we found a `TrackOrderByContext`  satisfied `isOrderPreserving()`,
`OrderPreservingTracker.selectedTrackOrderByContext` is set to this `TrackOrderByContext`, then we can use this `TrackOrderByContext` to implement the `OrderPreservingTracker.getOrderPreservingTrackInfos` and `OrderPreservingTracker. isReverse` etc.

phoenix-core/src/main/java/org/apache/phoenix/execute/AggregatePlan.java
comnetwork: Not every plan have to  implement this function, eg. HashJoinPlan and SortMergeJoinPlan.
In my opinion, this method is not suitable to declared as abstract method in QueryPlan and subclass to override it , because this method is to invoke in the ctor of concrete  subclass.

phoenix-core/src/main/java/org/apache/phoenix/compile/QueryPlan.java
twdsilva: ``getOutputOrdering`` also works.

phoenix-core/src/it/java/org/apache/phoenix/end2end/SortMergeJoinMoreIT.java
```diff
@@ -423,7 +424,7 @@ public void testBug2894() throws Exception {
                         "         ) L\n" +
                         "     ON L.BUCKET = E.BUCKET AND L.TIMESTAMP = E.TIMESTAMP\n" +
                         " ) C\n" +
-                        " GROUP BY C.BUCKET, C.TIMESTAMP";
+                        " GROUP BY C.BUCKET, C.TIMESTAMP ORDER BY C.BUCKET, C.TIMESTAMP";
```
twdsilva: Why do we need to add the order by here?

phoenix-core/src/it/java/org/apache/phoenix/end2end/join/SubqueryUsingSortMergeJoinIT.java
```diff
@@ -516,8 +516,8 @@ public void testAnyAllComparisonSubquery() throws Exception {
             assertEquals(rs.getString(2), "T6");
 
             assertFalse(rs.next());
-            
-            query = "SELECT /*+ USE_SORT_MERGE_JOIN*/ \"order_id\", name FROM " + tableName4 + " o JOIN " + tableName1 + " i ON o.\"item_id\" = i.\"item_id\" WHERE quantity != ANY(SELECT quantity FROM " + tableName4 + " q WHERE o.\"item_id\" = q.\"item_id\" GROUP BY quantity)";
+            //add order by to make the query result stable
+            query = "SELECT /*+ USE_SORT_MERGE_JOIN*/ \"order_id\", name FROM " + tableName4 + " o JOIN " + tableName1 + " i ON o.\"item_id\" = i.\"item_id\" WHERE quantity != ANY(SELECT quantity FROM " + tableName4 + " q WHERE o.\"item_id\" = q.\"item_id\" GROUP BY quantity) order by \"order_id\"";
```
twdsilva: what happens if we don't add the order by?

phoenix-core/src/it/java/org/apache/phoenix/end2end/SortMergeJoinMoreIT.java
comnetwork: I added the order by here mainly for:
1.Make the query result stable,because the original sql is only have the group by. The group by is not `orderPreserving` without this patch and must add a additional `SORTED BY [E.TIMESTAMP, E.BUCKET]`, but after this patch, the group by is  `orderPreserving`, so the query result is not same as without this patch.
2. To test if the GroupBy and OrderBy could both be optimized, from the explain we can see the 
`GROUP BY C.BUCKET, C.TIMESTAMP` is  `orderPreserving` and `ORDER BY C.BUCKET, C.TIMESTAMP` is compiled out.

If I do not add the order by, the explain assertion and query result assertion is same as the order by is added. 

Another curious problem  is that  why the order of query result is reversed after this patch:
the query result without this patch is :
                assertEquals("5SEC", rs.getString(1));
                assertEquals(1462993520000000000L, rs.getLong(2));
                assertTrue(rs.next());
                assertEquals("5SEC", rs.getString(1));
                assertEquals(1462993515000000000L, rs.getLong(2));
                assertTrue(rs.next());
                assertEquals("5SEC", rs.getString(1));
                assertEquals(1462993510000000000L, rs.getLong(2));
                assertTrue(rs.next());
                assertEquals("5SEC", rs.getString(1));
                assertEquals(1462993505000000000L, rs.getLong(2));
                assertTrue(rs.next());
                assertEquals("5SEC", rs.getString(1));
                assertEquals(1462993490000000000L, rs.getLong(2));
                assertTrue(rs.next());
                assertEquals("5SEC", rs.getString(1));
                assertEquals(1462993485000000000L, rs.getLong(2));
                assertTrue(rs.next());
                assertEquals("5SEC", rs.getString(1));
                assertEquals(1462993480000000000L, rs.getLong(2));
                assertTrue(rs.next());
                assertEquals("5SEC", rs.getString(1));
                assertEquals(1462993475000000000L, rs.getLong(2));
                assertTrue(rs.next());
                assertEquals("5SEC", rs.getString(1));
                assertEquals(1462993470000000000L, rs.getLong(2));
                assertTrue(rs.next());
                assertEquals("5SEC", rs.getString(1));
                assertEquals(1462993430000000000L, rs.getLong(2));
the query result after this patch is :
               assertEquals("5SEC", rs.getString(1));
                assertEquals(1462993430000000000L, rs.getLong(2));
                assertTrue(rs.next());
                assertEquals("5SEC", rs.getString(1));
                assertEquals(1462993470000000000L, rs.getLong(2));
                assertTrue(rs.next());
                assertEquals("5SEC", rs.getString(1));
                assertEquals(1462993475000000000L, rs.getLong(2));
                assertTrue(rs.next());
                assertEquals("5SEC", rs.getString(1));
                assertEquals(1462993480000000000L, rs.getLong(2));
                assertTrue(rs.next());
                assertEquals("5SEC", rs.getString(1));
                assertEquals(1462993485000000000L, rs.getLong(2));
                assertTrue(rs.next());
                assertEquals("5SEC", rs.getString(1));
                assertEquals(1462993490000000000L, rs.getLong(2));
                assertTrue(rs.next());
                assertEquals("5SEC", rs.getString(1));
                assertEquals(1462993505000000000L, rs.getLong(2));
                assertTrue(rs.next());
                assertEquals("5SEC", rs.getString(1));
                assertEquals(1462993510000000000L, rs.getLong(2));
                assertTrue(rs.next());
                assertEquals("5SEC", rs.getString(1));
                assertEquals(1462993515000000000L, rs.getLong(2));
                assertTrue(rs.next());
                assertEquals("5SEC", rs.getString(1));
                assertEquals(1462993520000000000L, rs.getLong(2));

phoenix-core/src/it/java/org/apache/phoenix/end2end/join/SubqueryUsingSortMergeJoinIT.java
comnetwork: If we don't add the order by , the order of the query result is reverse after this patch, just as:

phoenix-core/src/main/java/org/apache/phoenix/compile/QueryPlan.java
comnetwork: @twdsilva , thank you for your opinion, in order to in correspondence with the existing name `getOrderBy()`  in `QueryPlan` class, can we named it as `getOutputOrderBy` ?

twdsilva: Sure that works

phoenix-core/src/it/java/org/apache/phoenix/end2end/SortMergeJoinMoreIT.java
twdsilva: Ok, I see now the results are the same even if you don't add the order by and the order by gets optimized out.

phoenix-core/src/it/java/org/apache/phoenix/end2end/join/SubqueryUsingSortMergeJoinIT.java
twdsilva: IMO its fine to not have the ```ORDER BY``` and change the test to verify that the order of results is reversed as previously the ```CLIENT SORTED BY [O.ITEM_ID]``` was not needed.

comnetwork: @twdsilva , thank you for your opinion, the reason why the query result with this patch is not same as before this patch is different from previous test, if there are more than two tuples which O.ITEM_ID are '0000000006', the query result may not be accurately reversed. Following your suggestion, I would add explain assertions to verify the unnecessary sort is avoid.

comnetwork:
@twdsilva , I made a new PR #468  for this JIRA, please help me have a review again, thank you very much.

gjacoby126:
JIRA resolved, closing PR

PR #445, opened 2019-02-22
PHOENIX-2787 support IF EXISTS for ALTER TABLE SET options

yanxinyi:


yanxinyi:
@vincentpoon plz review this pr when u have time, thanks!

phoenix-core/src/it/java/org/apache/phoenix/end2end/AlterTableIT.java
```diff
@@ -1352,7 +1352,34 @@ private void assertSequenceNumber(String schemaName, String tableName, long expe
             assertFalse(rs.next());
         }
     }
-    
+
+    @Test
+    public void testAlterTableIfExistsThenOption() throws Exception {
+        Properties props = PropertiesUtil.deepCopy(TEST_PROPERTIES);
+        Connection conn = DriverManager.getConnection(getUrl(), props);
```
vincentpoon: Use try-with-resources for the connection

phoenix-core/src/it/java/org/apache/phoenix/end2end/AlterTableIT.java
```diff
@@ -1352,7 +1352,34 @@ private void assertSequenceNumber(String schemaName, String tableName, long expe
...
+        conn.setAutoCommit(false);
+        String ddl = "CREATE TABLE FOO (A VARCHAR  PRIMARY KEY, B VARCHAR)";
+        createTestTable(getUrl(), ddl);
+
+        // testing ALTER TABLE ADD WITH IF EXISTS SET OPTION
+        try {
+            conn.createStatement().execute("ALTER TABLE FOO ADD C VARCHAR IF EXISTS" +
+                    "  SET DISABLE_WAL = true");
+        } catch (SQLException e){
+            assertEquals(SQLExceptionCode.MISSING_TOKEN.getErrorCode(), e.getErrorCode());
```
vincentpoon: why does this case throw an exception?

phoenix-core/src/it/java/org/apache/phoenix/end2end/AlterTableIT.java
yanxinyi: will do

phoenix-core/src/it/java/org/apache/phoenix/end2end/AlterTableIT.java
yanxinyi: @vincentpoon Thank you so much for reviewing my pr.
Currently, Phoenix implementation doesn't support the `set` keyword with options at the end.
For instance, `ALTER TABLE NAME ADD COLNAME TYPE` + `OPTIONS` is supported, but `SET OPTIONS` is not. This is an open question. Do you want me to support this new grammar? If we want to support this as well, I can create a new jira and support this.

This jira supports `ALTER TABLE NAME **IF EXISTS**` + `SET OPTIONS`. Correct me if I miss anything here, thanks! :)

yanxinyi:
created a new pr here.
https://github.com/apache/phoenix/pull/462

PR #446, opened 2019-02-23
Phoenix 5136

mspielberg:
Fixes incorrect treatment of NULL as a stop value in AND expression evaluation. `NULL AND FALSE` should evaluate to `FALSE`, not `NULL`, so NULL is not a valid stop value. `NULL AND TRUE` is NULL however, so we need to retain memory of whether a past child evaluated to NULL, even though it is not a stop value.

Also optimizes BooleanExpressionFilter to immediately drop the current row and move on to the next row if the expression evaluates to NULL.

Supercedes #443.

### DIFF
--- a/phoenix-core/src/main/java/org/apache/phoenix/expression/AndOrExpression.java
+++ b/phoenix-core/src/main/java/org/apache/phoenix/expression/AndOrExpression.java
@@ -24,7 +24,7 @@ import org.apache.hadoop.hbase.io.ImmutableBytesWritable;
 import org.apache.phoenix.schema.types.PBoolean;
 import org.apache.phoenix.schema.types.PDataType;
 import org.apache.phoenix.schema.tuple.Tuple;
-
+import org.apache.phoenix.util.ByteUtil;
 
 /**
  * 
@@ -36,6 +36,8 @@ import org.apache.phoenix.schema.tuple.Tuple;
 public abstract class AndOrExpression extends BaseCompoundExpression {
     // Remember evaluation of child expression for partial evaluation
     private BitSet partialEvalState;
+    // true if we have seen NULL as the value of some child expression
+    private boolean seenNull = false;
    
     public AndOrExpression() {
     }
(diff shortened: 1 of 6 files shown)


phoenix-core/src/main/java/org/apache/phoenix/expression/AndOrExpression.java
```diff
@@ -36,6 +36,8 @@
 public abstract class AndOrExpression extends BaseCompoundExpression {
     // Remember evaluation of child expression for partial evaluation
     private BitSet partialEvalState;
+    // true if we have seen NULL as the value of some child expression
```
swaroopak: Redundant comment. Please remove.

phoenix-core/src/test/java/org/apache/phoenix/expression/OrExpressionTest.java
```diff
@@ -0,0 +1,293 @@
...
+    private OrExpression createOr(Expression lhs, Expression rhs) {
+        return new OrExpression(Arrays.asList(lhs, rhs));
+    }
+
+    private OrExpression createOr(Boolean x, Boolean y) {
+        return createOr(LiteralExpression.newConstant(x), LiteralExpression.newConstant(y));
+    }
+
+    private void testImmediateSingle(Boolean expected, Boolean lhs, Boolean rhs) {
+        OrExpression and = createOr(lhs, rhs);
```
twdsilva: nit: or

phoenix-core/src/main/java/org/apache/phoenix/expression/AndOrExpression.java
twdsilva: IMO this comment makes it easier to understand the code.

mspielberg:
Merged upstream in https://github.com/apache/phoenix/commit/f4579a52a950a880368a2ec9f8cecbd3c67cbae9 and https://github.com/apache/phoenix/commit/43310a6dc69433d3df772e17d48fe4d8821d40ac.

PR #447, opened 2019-02-26
PHOENIX-4918 Apache Phoenix website Grammar page is running on an old…

yanxinyi:
… version


swaroopak:
@yanxinyi I merged this on all the active branches about 3 weeks ago. I see that the commit has caught some error https://github.com/apache/phoenix/blob/4.x-HBase-1.3/docs/phoenix.csv which was not there earlier https://github.com/apache/phoenix/blob/4.x-HBase-1.2/docs/phoenix.csv.  Could you please fix it? It should be something simple of removing an unwanted quote etc. 
Thank you.

PR #448, opened 2019-02-26
use PreparedStatement

bd2019us:
Hello,
Is it a good way to use "PreparedStatement" instead of "createStatement" in order to improve performance?  "PreparedStatement" is used to execute dynamic or parameterized SQL queries. It is helpful if we are executing a particular SQL query multiple times ( such as in a loop in this case) since "PreparedStatement" can be precompiled and the query is created only once.

### DIFF
--- a/phoenix-core/src/it/java/org/apache/phoenix/end2end/DeleteIT.java
+++ b/phoenix-core/src/it/java/org/apache/phoenix/end2end/DeleteIT.java
@@ -791,9 +791,12 @@ public class DeleteIT extends ParallelStatsDisabledIT {
         props.setProperty(QueryServices.ENABLE_SERVER_SIDE_MUTATIONS, allowServerSideMutations);
         try (Connection conn = DriverManager.getConnection(getUrl(), props)) {
             conn.createStatement().execute(ddl);
-            Statement stmt = conn.createStatement();
+            //Statement stmt = conn.createStatement();
+			PreparedStatement preStmt = conn.prepareStatement("UPSERT INTO " + tableName + " (pk1, v1) VALUES (" + ? + ",'value')");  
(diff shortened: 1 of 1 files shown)


phoenix-core/src/it/java/org/apache/phoenix/end2end/DeleteIT.java
```diff
@@ -791,9 +791,12 @@ private void testDeleteCount(boolean autoCommit, Integer limit) throws Exception
         props.setProperty(QueryServices.ENABLE_SERVER_SIDE_MUTATIONS, allowServerSideMutations);
         try (Connection conn = DriverManager.getConnection(getUrl(), props)) {
             conn.createStatement().execute(ddl);
-            Statement stmt = conn.createStatement();
+            //Statement stmt = conn.createStatement();
```
swaroopak: nit: Remove the commented code.

phoenix-core/src/it/java/org/apache/phoenix/end2end/DeleteIT.java
```diff
@@ -791,9 +791,12 @@ private void testDeleteCount(boolean autoCommit, Integer limit) throws Exception
...
+			PreparedStatement preStmt = conn.prepareStatement("UPSERT INTO " + tableName + " (pk1, v1) VALUES (" + ? + ",'value')");
```
swaroopak: nit: I would extract the SQL string for readability.

lx0704:
Hi @swaroopak,  sorry for the late reply.   I just submit a patch to [JIRA](https://issues.apache.org/jira/browse/PHOENIX-5412) for Hadoop QA.  Please check it.

PR #455, opened 2019-03-06
PHOENIX-4900 Modify MAX_MUTATION_SIZE_EXCEEDED and MAX_MUTATION_SIZE_…

yanxinyi:
…BYTES_EXCEEDED exception message to recommend turning autocommit off for deletes

### DIFF
--- a/phoenix-core/src/it/java/org/apache/phoenix/end2end/MutationStateIT.java
+++ b/phoenix-core/src/it/java/org/apache/phoenix/end2end/MutationStateIT.java
@@ -59,7 +59,51 @@ public class MutationStateIT extends ParallelStatsDisabledIT {
     }
 
     @Test
-    public void testMaxMutationSize() throws Exception {
+    public void testDeleteMaxMutationSize() throws SQLException {
+        String tableName = generateUniqueName();
+        int NUMBER_OF_ROWS = 20;
+        String ddl = "CREATE TABLE " + tableName + " (V BIGINT PRIMARY KEY, K BIGINT)";
+        PhoenixConnection conn = (PhoenixConnection) DriverManager.getConnection(getUrl());
+        conn.createStatement().execute(ddl);
+
+        for(int i = 0; i < NUMBER_OF_ROWS; i++) {
+            conn.createStatement().execute(
+                    "UPSERT INTO " + tableName + " VALUES (" + i + ", "+ i + ")");
+            conn.commit();
+        }
+
+        Properties props = new Properties();
+        props.setProperty(QueryServices.MAX_MUTATION_SIZE_ATTRIB,
+                String.valueOf(NUMBER_OF_ROWS / 2));
+        PhoenixConnection connection =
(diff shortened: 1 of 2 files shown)


phoenix-core/src/main/java/org/apache/phoenix/exception/SQLExceptionCode.java
```diff
@@ -466,8 +466,11 @@ public SQLException newException(SQLExceptionInfo info) {
         "because this client already has the maximum number" +
         " of connections to the target cluster."),
     
-    MAX_MUTATION_SIZE_EXCEEDED(729, "LIM01", "MutationState size is bigger than maximum allowed number of rows"),
-    MAX_MUTATION_SIZE_BYTES_EXCEEDED(730, "LIM02", "MutationState size is bigger than maximum allowed number of bytes"), 
+    MAX_MUTATION_SIZE_EXCEEDED(729, "LIM01", "MutationState size is bigger" +
+            " than maximum allowed number of rows, please turning autocommit on and try it again."),
```
twdsilva: For upserts it would be slower to turn autocommit on since we would call commit for every row. 
I think it would be better if this message was .. try upserting rows in smaller batches or using autocommit on for deletes.

phoenix-core/src/it/java/org/apache/phoenix/end2end/MutationStateIT.java
```diff
@@ -59,7 +59,54 @@ private void upsertRows(PhoenixConnection conn, String fullTableName) throws SQL
...
+                    SQLExceptionCode.MAX_MUTATION_SIZE_EXCEEDED.getMessage()));
+        }
+
+        props = new Properties();
+        props.setProperty(QueryServices.MAX_MUTATION_SIZE_BYTES_ATTRIB, "10");
+        connection = (PhoenixConnection) DriverManager.getConnection(getUrl(), props);
+        connection.setAutoCommit(false);
+
+        try {
+            for(int i = 0; i < NUMBER_OF_ROWS; i++) {
```
twdsilva: Instead of running the delete in a for loop, just use a single statement "DELETE FROM tableName", this should also throw an exception.

phoenix-core/src/it/java/org/apache/phoenix/end2end/MutationStateIT.java
```diff
@@ -59,7 +59,54 @@ private void upsertRows(PhoenixConnection conn, String fullTableName) throws SQL
...
+                connection.createStatement().execute(
+                        "DELETE FROM " + tableName + " WHERE K = " + i );
+            }
+        } catch (SQLException e) {
+            assertTrue(e.getMessage().contains(
+                    SQLExceptionCode.MAX_MUTATION_SIZE_EXCEEDED.getMessage()));
+        }
+
+        props = new Properties();
```
twdsilva: remove this block as it duplicated

yanxinyi:
@twdsilva updated error msg

twdsilva:
+1 , thanks @yanxinyi

PR #457, opened 2019-03-07
 PHOENIX-5190 Implement TaskRegionObserver for Index rebuild

gokceni:


gokceni:
@gjacoby126 @kadirozde @abhishek-chouhan

gjacoby126:
@gokceni - what JIRA does this correspond to?

gokceni:
@gjacoby126 PHOENIX-5190 is the JIRA.

phoenix-core/src/it/java/org/apache/phoenix/end2end/IndexRebuildTaskIT.java
```diff
@@ -0,0 +1,200 @@
...
+                    PTable.TaskStatus.CREATED.toString(), data, null, null, true);
+
+            // Run IndexRebuildTask
+            TaskRegionObserver.IndexRebuildTask task =
+                    new TaskRegionObserver.IndexRebuildTask(
+                            TaskRegionEnvironment, QueryServicesOptions.DEFAULT_TASK_HANDLING_MAX_INTERVAL_MS);
+
+            task.run();
+
+            String viewIndexTableName = "_IDX_" + baseTable;
```
gjacoby126: nit: should use the appropriate SchemaUtil method to construct the view index table name.

phoenix-core/src/it/java/org/apache/phoenix/end2end/IndexRebuildTaskIT.java
```diff
@@ -0,0 +1,200 @@
...
+            assertTrue(count == 1);
+
+
+            // Remove index contents and try again
+            Admin admin = queryServices.getAdmin();
+            TableName tableName = TableName.valueOf(viewIndexTableName);
+            admin.disableTable(tableName);
+            admin.truncateTable(tableName, false);
+
+            data = "{GusWorkId:abc}";
```
gjacoby126: nit: consider using a constant

phoenix-core/src/it/java/org/apache/phoenix/end2end/IndexRebuildTaskIT.java
```diff
@@ -0,0 +1,200 @@
...
+                    isMultiTenant? TENANT1 : null, null, viewName, indexName,
+                    PTable.TaskStatus.CREATED.toString(), data, null, null, true);
+            task.run();
+
+            Table systemHTable= queryServices.getTable(Bytes.toBytes("SYSTEM."+PhoenixDatabaseMetaData.SYSTEM_TASK_TABLE));
+            count = getUtility().countRows(systemHTable);
+            assertTrue( count == numOfRuns);
+
+            // Check task status and other column values.
+            ResultSet rs = conn.createStatement().executeQuery("SELECT * " +
```
gjacoby126: Seems like this would be useful to have as a utility function somewhere. In general it's good to break up long functions into helper functions with descriptive names when there's a discrete unit of work.

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/TaskRegionObserver.java
```diff
@@ -157,18 +191,58 @@ public static void addTask(PhoenixConnection conn, TaskType taskType, String ten
...
-            if (tenantId != null) {
-                stmt.setString(2, tenantId);
+            stmt = setValuesToAddTaskPS(stmt, taskType, tenantId, schemaName, tableName);
+        } catch (SQLException e) {
+            throw new IOException(e);
+        }
+        mutateSystemTaskTable(conn, stmt, accessCheckEnabled);
+    }
+
+    public static void addTask(PhoenixConnection conn, TaskType taskType, String tenantId, String schemaName,
```
gjacoby126: Seems like there should be a separation of concerns here -- persisting a Task to a Phoenix table and reading it back is distinct from scheduling a task. I'd suggest breaking some of these Task classes into their own files and either fully encapsulate their persistence logic in the Tasks (the Data Access Object pattern) or have standalone persistence classes that handle the persistence (the Repository pattern). 

Either way would improve the readability and reduce the coupling between the scheduling logic and the persistence logic. 

Also agree with @kadirozde 's ideas to use reflection and JSON parsing to reduce the amount of boilerplate necessary to hook in each new task type for this framework.

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/TaskRegionObserver.java
```diff
@@ -298,4 +493,132 @@ public void run() {
...
+     */
+    public static class IndexRebuildTask extends SelfHealingTask {
+
+        public IndexRebuildTask(RegionCoprocessorEnvironment env, long timeMaxInterval) {
+            super.init(env, timeMaxInterval);
+        }
+
+        public static String[] getArgValues(boolean directApi, boolean useSnapshot, String schemaName,
+                String dataTable, String indxTable, String tenantId) {
+            final List<String> args = Lists.newArrayList();
```
gjacoby126: This is leaking IndexTool internals into this task. Would be better to provide an abstraction so you can call the IndexTool directly.

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/TaskRegionObserver.java
```diff
@@ -298,4 +493,132 @@ public void run() {
...
+                ResultSet rs = queryTaskTable(connForTask, TaskType.INDEX_REBUILD, excludeStatues);
+                while (rs.next()) {
+                    try {
+                        TaskRecord taskRecord = parseResult(rs);
+                        tenantId = taskRecord.getTenantId();
+                        schemaName= taskRecord.getSchemaName();
+                        tableName = taskRecord.getTableName();
+                        indexName = taskRecord.getIndexName();
+                        final String[] cmdArgs =
+                                getArgValues(true, false, schemaName, tableName, indexName, tenantId);
```
gjacoby126: As I mentioned above, it would be better to have an API in the IndexTool that could be called directly in an object-oriented fashion. The IndexTool should only have a CLI so that it can be called from a real shell command line, but here we don't need such things.

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/TaskRegionObserver.java
```diff
@@ -298,4 +493,132 @@ public void run() {
...
+                                 tableName, indexName,  PTable.TaskStatus.STARTED.toString(), taskRecord.data, null, null,true);
+
+                        String data = taskRecord.data;
+                        if (Strings.isNullOrEmpty(taskRecord.data)) {
+                            data = "{}";
+                        }
+                        JsonParser jsonParser = new JsonParser();
+                        JsonObject jsonObject = jsonParser.parse(data).getAsJsonObject();
+                        if (jsonObject.has("DisableBefore")) {
+                            String disableBefore = jsonObject.get("DisableBefore").toString();
```
gjacoby126: This seems like logic that should be in the IndexTool itself.

phoenix-core/src/main/java/org/apache/phoenix/jdbc/PhoenixDatabaseMetaData.java
```diff
@@ -219,6 +219,12 @@
     public static final byte[] TASK_TYPE_BYTES = Bytes.toBytes(TASK_TYPE);
     public static final String TASK_TS = "TASK_TS";
     public static final byte[] TASK_TS_BYTES = Bytes.toBytes(TASK_TS);
+    public static final String TASK_STATUS = "TASK_STATUS";
+    public static final String TASK_END_TS = "TASK_END_TS";
+    public static final String PRIORITY = "PRIORITY";
+    public static final String DATA = "DATA";
```
gjacoby126: Maybe make the constant named "TASK_DATA_COLUMN" or something along those lines to indicate what it's for?

gjacoby126: (and same with PRIORITY)

phoenix-core/src/main/java/org/apache/phoenix/query/ConnectionQueryServicesImpl.java
```diff
@@ -3521,6 +3524,31 @@ public void upgradeSystemTables(final String url, final Properties props) throws
...
                 }
             }
+
+            try {
+                metaConnection.createStatement().executeUpdate(getTaskDDL());
+            } catch (NewerTableAlreadyExistsException e) {
+
+            } catch (TableAlreadyExistsException e) {
+                long currentServerSideTableTimeStamp = e.getTable().getTimeStamp();
+                if (currentServerSideTableTimeStamp <= MetaDataProtocol.MIN_SYSTEM_TABLE_TIMESTAMP_5_1_0) {
```
gjacoby126: Keep in mind this change will also likely go into 4.15, so the upgrade will need to happen for either 4.15 or 5.1

phoenix-core/src/main/java/org/apache/phoenix/jdbc/PhoenixDatabaseMetaData.java
gokceni: Will rename as TASK_DATA. The others don't have _COLUMN at the end

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/TaskRegionObserver.java
gokceni: @gjacoby126 This is not a new method, it was already in this file. I moved it to this class. I am not seeing the patterns for other System tables. I will add a Tasks class tho.

gokceni:
@kadirozde (and @gjacoby126)  for this comment:
"Have one thread that goes through the system task table and maps the task type to its class, creates an instance of it with task attributes and finally calls the run method of the instance."

If we do it like this, don't we tie both tasks to a single thread?
From the discussion we had, I thought we said each task having their own thread. Right now, there is 2. If we keep adding different task types, they will all run on 1 thread. Also, we might want to run them on different intervals (some tasks might not be that frequent, some might). What do you think?

kadirozde:
@gokceni, regarding having one main thread responsible form scanning the system table allows us to separate the system table management concerns from the individual task handing ones. Initially having one thread will be fine as we will not have many tasks. I think, when we need more threads and scheduling task based on their priorities, this would be handled within the system task framework without changing task handling classes if the concerns are separated.

phoenix-core/src/main/java/org/apache/phoenix/schema/task/Task.java
```diff
@@ -0,0 +1,314 @@
...
+                    PhoenixDatabaseMetaData.TABLE_SCHEM + ", " +
+                    PhoenixDatabaseMetaData.TABLE_NAME + ", " +
+                    PhoenixDatabaseMetaData.TASK_STATUS + ", " +
+                    PhoenixDatabaseMetaData.TASK_PRIORITY + ", " +
+                    PhoenixDatabaseMetaData.TASK_END_TS + ", " +
+                    PhoenixDatabaseMetaData.TASK_DATA +
+                    " ) VALUES(?,?,?,?,?,?,?,?)");
+            stmt = setValuesToAddTaskPS(stmt, taskType, tenantId, schemaName, tableName);
+            if (taskStatus != null) {
+                stmt.setString(5, taskStatus);
```
gjacoby126: Any reason this logic shouldn't be in setValuesToAddTaskPS? (You could have an overload that took just the original 4 params that calls the full 8 param function)

gokceni: sure, will do

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/TaskRegionObserver.java
```diff
@@ -214,85 +165,77 @@ public DropChildViewsTask(RegionCoprocessorEnvironment env, long timeMaxInterval
...
+
+                        Object obj = concreteClass.newInstance();
+                        Method runMethod = concreteClass.getDeclaredMethod("run",
+                                Task.TaskRecord.class);
+                        Method initMethod = concreteClass.getSuperclass().getDeclaredMethod("init",
+                                RegionCoprocessorEnvironment.class, Long.class);
+                        initMethod.invoke(obj, env, timeMaxInterval);
+
+                        // Change task status to STARTED
+                        Task.addTask(connForTask, taskRecord.getTaskType(), taskRecord.getTenantId(), taskRecord.getSchemaName(),
```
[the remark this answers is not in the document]
gokceni: yes, we exclude Completed and Failed states and get anything else

gjacoby126:
Closing because this was pushed via git command line

PR #458, opened 2019-03-07
PHOENIX-5181 support Math sin/cos/tan functions

yanxinyi:


ChinmaySKulkarni:
@yanxinyi overall looks good. I think the offline discussions we had regarding precision and rounding errors are the main points of concern. Can you add tests that reflect such cases?

yanxinyi:
Offline synced with @ChinmaySKulkarni about the precision issue, and Phoenix PDecimal class already takes care of it. Other similar Math functions, such as power, implemented the same way as I do, but 
 @twdsilva can we have your input here, thanks.

karanmehta93:
Can you add some IT's that test the grammar as well? @yanxinyi

yanxinyi:
@karanmehta93 added IT tests

yanxinyi:
Offline discussed with @karanmehta93, used `@Parameters` feature for better code quality in unit test functions as he suggested.

gjacoby126:
@karanmehta93 , were there any other changes you wanted to the trig function feature?

phoenix-core/src/it/java/org/apache/phoenix/end2end/MathTrigFunctionEnd2EndIT.java
```diff
@@ -0,0 +1,94 @@
...
+        PreparedStatement stmt = null;
+        tableName = generateUniqueName();
+
+        try {
+            conn = DriverManager.getConnection(getUrl());
+            String ddl;
+            ddl =
+                    "CREATE TABLE " + tableName + " (k VARCHAR NOT NULL PRIMARY KEY, doub DOUBLE)";
+            conn.createStatement().execute(ddl);
+            conn.commit();
```
karanmehta93: Commit is not needed after ddl statements.

phoenix-core/src/test/java/org/apache/phoenix/expression/MathTrigFunctionTest.java
```diff
@@ -0,0 +1,179 @@
...
+
+    private void test(Number value, PNumericType dataType, double expectedResult,
+                             String testedFunction)
+            throws SQLException {
+        LiteralExpression literal = LiteralExpression.newConstant(value, dataType, SortOrder.ASC);
+        boolean ret1 = testExpression(literal, expectedResult, testedFunction);
+
+        literal = LiteralExpression.newConstant(value, dataType, SortOrder.DESC);
+        boolean ret2 = testExpression(literal, expectedResult, testedFunction);
+        assertEquals(ret1, ret2);
```
karanmehta93: Why is this required? If both `ret1` and `ret2` return false, this would still pass. Is that intended?

karanmehta93: As discussed offline, @yanxinyi will fix this in another Jira.

PR #460, opened 2019-03-08
PHOENIX-5184: HBase and Phoenix connection leaks in Indexing code path, OrphanViewTool and PhoenixConfigurationUtil

ChinmaySKulkarni:


phoenix-core/src/main/java/org/apache/phoenix/mapreduce/index/IndexScrutinyMapper.java
```diff
@@ -151,6 +151,16 @@ protected void setup(final Context context) throws IOException, InterruptedExcep
             LOG.info("Target table base query: " + targetTableQuery);
             md5 = MessageDigest.getInstance("MD5");
         } catch (SQLException | NoSuchAlgorithmException e) {
+            if (this.connection != null) {
```
vincentpoon: try closing the statement first?

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/index/IndexScrutinyMapper.java
```diff
@@ -151,6 +151,16 @@ protected void setup(final Context context) throws IOException, InterruptedExcep
...
+                try {
+                    this.connection.close();
```
vincentpoon: in the rare case that this throws, `outputConn` will not get closed.  Looks like we need to fix it in `cleanup()` also

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/index/IndexTool.java
```diff
@@ -682,9 +694,15 @@ public int run(String[] args) throws Exception {
                 if (htable != null) {
                     htable.close();
                 }
-                if(regionLocator != null) {
+                if (hConn != null) {
+                    hConn.close();
```
vincentpoon: same here, if any of these closes throw, the ones afterwards won't get closed

vincentpoon: Might want to take a look at Guava's Closer to see if that can help with these cases.  Register all the closeables, then call closer.close() once

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/PhoenixRecordWriter.java
```diff
@@ -52,10 +52,20 @@ public PhoenixRecordWriter(final Configuration configuration) throws SQLExceptio
...
-        this.conn = ConnectionUtil.getOutputConnectionWithoutTheseProps(configuration, propsToIgnore);
-        this.batchSize = PhoenixConfigurationUtil.getBatchSize(configuration);
-        final String upsertQuery = PhoenixConfigurationUtil.getUpsertStatement(configuration);
-        this.statement = this.conn.prepareStatement(upsertQuery);
+        Connection connection = null;
+        try {
+            connection = ConnectionUtil.getOutputConnectionWithoutTheseProps(configuration, propsToIgnore);
+            this.batchSize = PhoenixConfigurationUtil.getBatchSize(configuration);
+            final String upsertQuery = PhoenixConfigurationUtil.getUpsertStatement(configuration);
+            this.statement = connection.prepareStatement(upsertQuery);
```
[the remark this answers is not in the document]
ChinmaySKulkarni: No exception can be thrown after instantiating the statement object, so that shouldn't be necessary. Besides, we close both the statement and the connection in the `close` method

ChinmaySKulkarni:
Thanks for the review @vincentpoon. I have incorporated your review comments. @twdsilva @gjacoby126 @kadirozde please provide feedback as well. Thanks.

ChinmaySKulkarni:
Thanks for the reviews @vincentpoon @kadirozde @twdsilva!

PR #462, opened 2019-03-11
PHOENIX-2787 support IF EXISTS for ALTER TABLE SET options

yanxinyi:


phoenix-core/src/main/java/org/apache/phoenix/jdbc/PhoenixStatement.java
```diff
@@ -430,6 +430,10 @@ public Integer call() throws SQLException {
                                         return executeMutation(stmt, false);
                                     }
                                 }
+                                if (stmt instanceof ExecutableAddColumnStatement &&
+                                        ((ExecutableAddColumnStatement) stmt).ifSetOptionExists()) {
+                                    return 0;
```
vincentpoon: Why do we need this in this catching of exception?  We shouldn't need this here

phoenix-core/src/main/java/org/apache/phoenix/schema/MetaDataClient.java
```diff
@@ -3514,13 +3515,15 @@ private void mutateStringProperty(String tenantId, String schemaName, String tab
...
+                statement.ifNotExists(), false, statement.getTable(), statement.getTableType(),
+                statement.ifSetOptionExists());
     }
 
     public MutationState addColumn(PTable table, List<ColumnDef> origColumnDefs,
             ListMultimap<String, Pair<String, Object>> stmtProperties, boolean ifNotExists,
-            boolean removeTableProps, NamedTableNode namedTableNode, PTableType tableType)
-                    throws SQLException {
+            boolean removeTableProps, NamedTableNode namedTableNode, PTableType tableType,
+                                   boolean ifSetOptionExist) throws SQLException {
```
vincentpoon: looks like you created the `ifSetOptionsExist` boolean and passed it all the way down to here, but never use it?

The overall structure feels wrong.  This shouldn't be something specific to SetOption.  It should apply to any "ALTER TABLE" command, including ADD/DROP column.  I think "ALTER TABLE IF EXISTS DROP COLUMN IF EXISTS" should be a valid command.  This is what PostgreSQL does:
https://www.postgresql.org/docs/9.2/sql-altertable.html

PR #463, opened 2019-03-13
Phoenix stats Initial Commit

dbwong:
Add the utility functions from my local work that @BinShi-SecularBird needed access to.

### DIFF
--- a/phoenix-core/src/main/java/org/apache/phoenix/compile/ScanRanges.java
+++ b/phoenix-core/src/main/java/org/apache/phoenix/compile/ScanRanges.java
@@ -694,4 +694,127 @@ public class ScanRanges {
         return rowTimestampRange;
     }
 
+    /**
+     * Produces the list of KeyRanges representing the fully qualified row key by calling into ScanUtil.setKey
+     * repeatedly for every combination of KeyRanges in the ranges field. The bounds will be set according to the
+     * properties of the setKey method.
+     * 
+     * @return list of KeyRanges representing the fully qualified rowkey, coalesced
+     */
+    public List<KeyRange> getRowKeyRanges() {
+        // If its scanRanges are everything or nothing, then we short circuit and leave
+        // as schema is not filled in
+        if (isEverything()) { return Lists.newArrayList(KeyRange.EVERYTHING_RANGE); }
+        if (isDegenerate()) { return Lists.newArrayList(KeyRange.EMPTY_RANGE); }
+
+        List<KeyRange> queryRowKeyRanges = Lists.newArrayListWithExpectedSize(this.ranges.size());
+
+
+        // If scanRanges.ranges has no information then should be in the scanRanges.scanRange
+        if (ranges.size() == 0) {
+            queryRowKeyRanges.add(this.getScanRange());
+        } else { // We have a composite key need the row key from the combination
+            // make a copy of ranges as we may add items to fully qualify our rowkey
+            List<List<KeyRange>> newRanges = new ArrayList<>(this.getRanges());
+
+            int[] slotSpans = this.getSlotSpans();
+
+            // If the ranges here do not qualify all the keys then those keys are unbound
+            if (newRanges.size() < schema.getMaxFields()) {
+                int originalSize = newRanges.size();
+                for (int i = 0; i < schema.getMaxFields() - originalSize; i++) {
+                    newRanges.add(Lists.newArrayList(KeyRange.EVERYTHING_RANGE));
+                }
+                slotSpans = new int[schema.getMaxFields()];
+                System.arraycopy(this.getSlotSpans(), 0, slotSpans, 0, this.getSlotSpans().length);
+            }
+
+            // Product to bound our counting loop for safety
+            int rangesPermutationsCount = 1;
+            for (int i = 0; i < newRanges.size(); i++) {
+                rangesPermutationsCount *= newRanges.get(i).size();
+            }
+
+            // Have to construct the intersection
+            List<KeyRange> expandedRanges = Lists.newArrayListWithCapacity(rangesPermutationsCount);
+            int[] positions = new int[slotSpans.length];
+
+            int maxKeyLength = SchemaUtil.getMaxKeyLength(schema, newRanges);
+            byte[] keyBuffer = new byte[maxKeyLength];
+
+            // Counting Loop
+            int count = 0;
+            while (count < rangesPermutationsCount) {
+                byte[] lowerBound;
+                byte[] upperBound;
+
+                // Note ScanUtil.setKey internally handles the upper/lower exclusive from a Scan
+                // point of view. It would be good to break it out in the future for rowkey
+                // construction vs ScanKey construction To handle differences between
+                // hbase 2 and hbase 1 scan boundaries
+                int result = ScanUtil.setKey(schema, newRanges, slotSpans, positions, KeyRange.Bound.LOWER, keyBuffer,
+                        0, 0, slotSpans.length);
+
+                if (result < 0) {
+                    // unbound
+                    lowerBound = KeyRange.UNBOUND;
+                } else {
+                    lowerBound = Arrays.copyOf(keyBuffer, result);
+                }
+
+                result = ScanUtil.setKey(schema, newRanges, slotSpans, positions, KeyRange.Bound.UPPER, keyBuffer, 0, 0,
+                        slotSpans.length);
+
+                if (result < 0) {
+                    // unbound
+                    upperBound = KeyRange.UNBOUND;
+                } else {
+                    upperBound = Arrays.copyOf(keyBuffer, result);
+                }
+
+
+                /*
+                 * This is already considered inside of ScanUtil.setKey we may want to refactor to pull these out.
+                 * range/single    boundary       bound      increment
+                 *  range          inclusive      lower         no
+                 *  range          inclusive      upper         yes, at the end if occurs at any slots.
+                 *  range          exclusive      lower         yes
+                 *  range          exclusive      upper         no
+                 *  single         inclusive      lower         no
+                 *  single         inclusive      upper         yes, at the end if it is the last slots.
+                 */
+
+                boolean lowerInclusive = true;
+                // Don't send a null range, send an empty range.
+                if (lowerBound.length == 0 && upperBound.length == 0) {
+                    lowerInclusive = false;
+                }
+
+                KeyRange keyRange = KeyRange.getKeyRange(lowerBound, lowerInclusive, upperBound, false);
+                expandedRanges.add(keyRange);
+
+                // update position
+                // This loops through all settings of each of the primary keys by counting from
+                // the trailing edge based on the number of settings of that key.
+                int i;
+                for (i = positions.length - 1; i >= 0; i--) {
+                    if (positions[i] < newRanges.get(i).size() - 1) {
+                        positions[i]++;
+                        break;
+                    } else {
+                        positions[i] = 0;
+                    }
+                }
+
+                if (i < 0) {
+                    break;
+                }
+                count++;
+            }
+            queryRowKeyRanges.addAll(expandedRanges);
+        }
+
+        return queryRowKeyRanges;
+    }
+
 }
--- a/phoenix-core/src/main/java/org/apache/phoenix/iterate/BaseResultIterators.java
+++ b/phoenix-core/src/main/java/org/apache/phoenix/iterate/BaseResultIterators.java
@@ -138,7 +138,7 @@ import com.google.common.collect.Lists;
  * Class that parallelizes the scan over a table using the ExecutorService provided.  Each region of the table will be scanned in parallel with
  * the results accessible through {@link #getIterators()}
  *
- * 
+ *
  * @since 0.1
  */
 public abstract class BaseResultIterators extends ExplainTable implements ResultIterators {
@@ -162,20 +162,13 @@ public abstract class BaseResultIterators extends ExplainTable implements Result
     private final boolean useStatsForParallelization;
     protected Map<ImmutableBytesPtr,ServerCache> caches;
     private final QueryPlan dataPlan;
-    
-    static final Function<HRegionLocation, KeyRange> TO_KEY_RANGE = new Function<HRegionLocation, KeyRange>() {
-        @Override
-        public KeyRange apply(HRegionLocation region) {
-            return KeyRange.getKeyRange(region.getRegionInfo().getStartKey(), region.getRegionInfo().getEndKey());
-        }
-    };
 
     private PTable getTable() {
         return plan.getTableRef().getTable();
     }
-    
+
     abstract protected boolean isSerial();
-    
+
     protected boolean useStats() {
         /*
          * Don't use guide posts:
@@ -188,7 +181,7 @@ public abstract class BaseResultIterators extends ExplainTable implements Result
         }
         return !isSerial();
     }
-    
+
     private static void initializeScan(QueryPlan plan, Integer perScanLimit, Integer offset, Scan scan) throws SQLException {
         StatementContext context = plan.getContext();
         TableRef tableRef = plan.getTableRef();
@@ -232,7 +225,7 @@ public abstract class BaseResultIterators extends ExplainTable implements Result
                     optimizeProjection = true;
                     if (projector.projectEveryRow()) {
                         if (table.getViewType() == ViewType.MAPPED) {
-                            // Since we don't have the empty key value in MAPPED tables, 
+                            // Since we don't have the empty key value in MAPPED tables,
                             // we must project all CFs in HRS. However, only the
                             // selected column values are returned back to client.
                             context.getWhereConditionColumns().clear();
@@ -258,7 +251,7 @@ public abstract class BaseResultIterators extends ExplainTable implements Result
             if (perScanLimit != null) {
                 ScanUtil.andFilterAtEnd(scan, new PageFilter(perScanLimit));
             }
-            
+
             if(offset!=null){
                 ScanUtil.addOffsetAttribute(scan, offset);
             }
@@ -268,7 +261,7 @@ public abstract class BaseResultIterators extends ExplainTable implements Result
                 cols < plan.getTableRef().getTable().getRowKeySchema().getFieldCount() &&
                 plan.getGroupBy().isOrderPreserving() &&
                 (context.getAggregationManager().isEmpty() || plan.getGroupBy().isUngroupedAggregate())) {
-                
+
                 ScanUtil.andFilterAtEnd(scan,
                         new DistinctPrefixFilter(plan.getTableRef().getTable().getRowKeySchema(),
                                 cols));
@@ -290,7 +283,7 @@ public abstract class BaseResultIterators extends ExplainTable implements Result
             }
         }
     }
-    
+
     private static void setQualifierRanges(boolean keyOnlyFilter, PTable table, Scan scan,
             StatementContext context) throws SQLException {
         if (EncodedColumnsUtil.useEncodedQualifierListOptimization(table, scan)) {
@@ -328,7 +321,7 @@ public abstract class BaseResultIterators extends ExplainTable implements Result
                                 EncodedColumnsUtil.getEmptyKeyValueInfo(table);
                         qualifierSet.add(emptyKeyValueInfo.getFirst());
                     }
-                    // In case of a keyOnlyFilter, we only need to project the 
+                    // In case of a keyOnlyFilter, we only need to project the
                     // empty key value column
                     if (!keyOnlyFilter) {
                         Pair<Integer, Integer> qualifierRangeForFamily =
@@ -366,11 +359,11 @@ public abstract class BaseResultIterators extends ExplainTable implements Result
             }
         }
     }
-    
+
     private static void optimizeProjection(StatementContext context, Scan scan, PTable table, FilterableStatement statement) {
         Map<byte[], NavigableSet<byte[]>> familyMap = scan.getFamilyMap();
         // columnsTracker contain cf -> qualifiers which should get returned.
-        Map<ImmutableBytesPtr, NavigableSet<ImmutableBytesPtr>> columnsTracker = 
+        Map<ImmutableBytesPtr, NavigableSet<ImmutableBytesPtr>> columnsTracker =
                 new TreeMap<ImmutableBytesPtr, NavigableSet<ImmutableBytesPtr>>();
         Set<byte[]> conditionOnlyCfs = new TreeSet<byte[]>(Bytes.BYTES_COMPARATOR);
         int referencedCfCount = familyMap.size();
@@ -452,7 +445,7 @@ public abstract class BaseResultIterators extends ExplainTable implements Result
                     // cols is null means the whole CF will get scanned.
                     if (cols != null) {
                         if (whereCol.getSecond() == null) {
-                            scan.addFamily(family);                            
+                            scan.addFamily(family);
                         } else {
                             scan.addColumn(family, whereCol.getSecond());
                         }
@@ -477,13 +470,13 @@ public abstract class BaseResultIterators extends ExplainTable implements Result
             // in the scan in this case. We still want the other optimization that causes
             // the ExplicitColumnTracker not to be used, though.
             if (!statement.isAggregate() && filteredColumnNotInProjection) {
-                ScanUtil.andFilterAtEnd(scan, 
+                ScanUtil.andFilterAtEnd(scan,
                     trackedColumnsBitset != null ? new EncodedQualifiersColumnProjectionFilter(SchemaUtil.getEmptyColumnFamily(table), trackedColumnsBitset, conditionOnlyCfs, table.getEncodingScheme()) : new ColumnProjectionFilter(SchemaUtil.getEmptyColumnFamily(table),
                         columnsTracker, conditionOnlyCfs, EncodedColumnsUtil.usesEncodedColumnNames(table.getEncodingScheme())));
             }
         }
     }
-    
+
     public BaseResultIterators(QueryPlan plan, Integer perScanLimit, Integer offset, ParallelScanGrouper scanGrouper, Scan scan, Map<ImmutableBytesPtr,ServerCache> caches, QueryPlan dataPlan) throws SQLException {
         super(plan.getContext(), plan.getTableRef(), plan.getGroupBy(), plan.getOrderBy(),
                 plan.getStatement().getHint(), QueryUtil.getOffsetLimit(plan.getLimit(), plan.getOffset()), offset);
@@ -505,7 +498,7 @@ public abstract class BaseResultIterators extends ExplainTable implements Result
         }
         // Used to tie all the scans together during logging
         scanId = new UUID(ThreadLocalRandom.current().nextLong(), ThreadLocalRandom.current().nextLong()).toString();
-        
+
         initializeScan(plan, perScanLimit, offset, scan);
         this.useStatsForParallelization = PhoenixConfigurationUtil.getStatsForParallelizationProp(context.getConnection(), table);
         this.scans = getParallelScans();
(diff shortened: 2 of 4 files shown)


BinShi-SecularBird:
Could you unify coding style first?

phoenix-core/src/main/java/org/apache/phoenix/compile/ScanRanges.java
```diff
@@ -694,4 +694,138 @@ public TimeRange getRowTimestampRange() {
         return rowTimestampRange;
     }
 
+    public List<KeyRange> getRowKeyRanges(){
+
```
BinShi-SecularBird: Do you want to keep this blank line? I don't see we have such blank line at the beginning of function body in other functions.

BinShi-SecularBird:
This change doesn't contain getRegionRowKeyRanges()? Are you going to add it in the next commit?

phoenix-core/src/main/java/org/apache/phoenix/compile/ScanRanges.java
```diff
@@ -694,4 +694,138 @@ public TimeRange getRowTimestampRange() {
         return rowTimestampRange;
     }
 
+    public List<KeyRange> getRowKeyRanges(){
```
BinShi-SecularBird: Could you add comments (function spec?) for this function?

phoenix-core/src/main/java/org/apache/phoenix/iterate/BaseResultIterators.java
```diff
@@ -550,6 +550,76 @@ public BaseResultIterators(QueryPlan plan, Integer perScanLimit, Integer offset,
...
+        List<HRegionLocation> regionLocations = getRegionBoundaries(scanGrouper);  //Load the region information
+
+        List<KeyRange> regionKeyRanges = Lists.newArrayListWithExpectedSize(regionLocations.size());
+
+        //Map each HRegionLocation to a KeyRange - no Java 8
+        for(HRegionLocation regionLocation : regionLocations){
+            HRegionInfo regionInfo = regionLocation.getRegionInfo();
+
+            //Region is upperInclusive false by definition
+            KeyRange range = KeyRange.getKeyRange(regionInfo.getStartKey(), true, regionInfo.getEndKey(), false);
```
BinShi-SecularBird: Should be

{
	byte[] lowerRange = isFirstRegion ? KeyRange.UNBOUND : regionLocations.get(0).getLowerRange; // Even they are both new byte[0], they are different.
    boolean lowerInclusive = ! isFirstRegion;
	byte[] upperRange = isLastRegion ? KeyRange.UNBOUND : regionLocations.get(regionLocations.size()-1).getUpperRange; // Even they are both new byte[0], they are different.
    boolean upperExclusive = false;
}

phoenix-core/src/main/java/org/apache/phoenix/iterate/BaseResultIterators.java
```diff
@@ -550,6 +550,76 @@ public BaseResultIterators(QueryPlan plan, Integer perScanLimit, Integer offset,
         }
         return ranges;
     }
+
+    private List<KeyRange> getRegionRowKeyRanges() throws SQLException {
+        List<HRegionLocation> regionLocations = getRegionBoundaries(scanGrouper);  //Load the region information
+
+        List<KeyRange> regionKeyRanges = Lists.newArrayListWithExpectedSize(regionLocations.size());
```
BinShi-SecularBird: I think we should use Lists.newArrayListWithCapacity(regionLocations.size()), because we know the exact size here. newArrayListWithExpectedSize actually use "given expectsize + 10% more" which we don't need.

phoenix-core/src/main/java/org/apache/phoenix/iterate/BaseResultIterators.java
```diff
@@ -550,6 +550,76 @@ public BaseResultIterators(QueryPlan plan, Integer perScanLimit, Integer offset,
...
+        ScanRanges scanRanges = context.getScanRanges();
+
+        RowKeySchema schema = scanRanges.getSchema();
+
+        PTable table = getTable();
+        boolean isLocalIndex = table.getIndexType() == IndexType.LOCAL;
+
+        //Use the dataplan to build the queryRowKeyRanges
+        if (isLocalIndex) {
+            // should always have a data plan when a local index is being used.
```
BinShi-SecularBird: The original comment says "// TODO: when implementing PHOENIX-4585, we should change this to an assert". Shall we do it now or open a Jira to track it?

phoenix-core/src/main/java/org/apache/phoenix/iterate/BaseResultIterators.java
```diff
@@ -550,6 +550,76 @@ public BaseResultIterators(QueryPlan plan, Integer perScanLimit, Integer offset,
...
+            if (dataPlan != null && dataPlan.getTableRef().getTable().getType()
+                    != PTableType.INDEX) { // Sanity check
```
BinShi-SecularBird: It is ugly to have a new line here. What do you think?

phoenix-core/src/main/java/org/apache/phoenix/iterate/BaseResultIterators.java
```diff
@@ -550,6 +550,76 @@ public BaseResultIterators(QueryPlan plan, Integer perScanLimit, Integer offset,
...
+                scanRanges =
+                        computePrefixScanRanges(dataPlan.getContext().getScanRanges(),
+                                computeColumnsInCommon());
```
BinShi-SecularBird: I prefer the following coding style which seems to be used by phoenix code base generally.

 <type >scanRanges = computePrefixScanRanges(
              dataPlan.getContext().getScanRanges(), computeColumnsInCommon());

Did you follow the following instructions on https://phoenix.apache.org/develop.html to set phoenix coding style.
Get Settings and Preferences Correct
Import Settings from eclipse profile
File -> Settings -> Editor -> Code Style -> Java
Set From… -> Import… -> Eclipse XML Profile -> {repository}/dev/PhoenixCodeTemplate.xml

BinShi-SecularBird:
Not finish the code review yet. Could you handle these feedback first?

phoenix-core/src/main/java/org/apache/phoenix/iterate/BaseResultIterators.java
dbwong: I agree that we can specifically call out the boundaries I dislike this specific code pasted.  I have added the refactor to specify unbound directly rather than implicitly.

phoenix-core/src/main/java/org/apache/phoenix/iterate/BaseResultIterators.java
dbwong: So I looked into this few weeks ago.  That JIRA, 4585, has been submitted but when I changed this to an assert I still found cases where we the dataplan was not filled in.  So I have left the original code and did not spend time investigating.

phoenix-core/src/main/java/org/apache/phoenix/iterate/BaseResultIterators.java
```diff
@@ -550,6 +550,76 @@ public BaseResultIterators(QueryPlan plan, Integer perScanLimit, Integer offset,
...
         return ranges;
     }
+
+    private List<KeyRange> getRegionRowKeyRanges() throws SQLException {
+        List<HRegionLocation> regionLocations = getRegionBoundaries(scanGrouper);  //Load the region information
+
+        List<KeyRange> regionKeyRanges = Lists.newArrayListWithExpectedSize(regionLocations.size());
+
+        //Map each HRegionLocation to a KeyRange - no Java 8
+        for(HRegionLocation regionLocation : regionLocations){
```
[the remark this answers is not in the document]
dbwong: Ok

phoenix-core/src/main/java/org/apache/phoenix/iterate/BaseResultIterators.java
dbwong: I agree but then the line is over 100 characters which previously on phoenix commits have complained.  I have reverted for now we will see later if there are complaints from the format bot.

phoenix-core/src/main/java/org/apache/phoenix/iterate/BaseResultIterators.java
dbwong: I thought this was already on but appeared this was a new workspace.  I have applied in general.

phoenix-core/src/main/java/org/apache/phoenix/iterate/BaseResultIterators.java
```diff
@@ -550,6 +550,76 @@ public BaseResultIterators(QueryPlan plan, Integer perScanLimit, Integer offset,
...
+                    != PTableType.INDEX) { // Sanity check
+                scanRanges =
+                        computePrefixScanRanges(dataPlan.getContext().getScanRanges(),
+                                computeColumnsInCommon());
+            }
+        }
+
+        List<KeyRange> queryRowKeyRanges = scanRanges.getRowKeyRanges();
+
+        if(isLocalIndex){
```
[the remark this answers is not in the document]
dbwong: Done

phoenix-core/src/main/java/org/apache/phoenix/iterate/BaseResultIterators.java
```diff
@@ -550,6 +550,76 @@ public BaseResultIterators(QueryPlan plan, Integer perScanLimit, Integer offset,
...
+            List<KeyRange> newQueryRowKeyRanges = Lists.newArrayListWithExpectedSize(queryRowKeyRanges.size() * regionKeyRanges.size());
+
+            for(KeyRange regionKeyRange : regionKeyRanges) {
```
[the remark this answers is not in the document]
dbwong: Done

phoenix-core/src/main/java/org/apache/phoenix/compile/ScanRanges.java
dbwong: I have added function level comments. Some basic testing was already included in the unit tests.  As this code snippet does nothing, I can not include functional tests.

phoenix-core/src/main/java/org/apache/phoenix/compile/ScanRanges.java
dbwong: Ok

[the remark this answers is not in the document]
dbwong: Not sure if this is referring to the same line.

phoenix-core/src/main/java/org/apache/phoenix/compile/ScanRanges.java
```diff
@@ -694,4 +694,127 @@ public TimeRange getRowTimestampRange() {
...
+                            break;
+                        } else {
+                            position[i] = 0;
+                        }
+                    }
+
+                    if (i == position.length) {
+                        break;
+                    }
+                    rangesPermutations++;
```
BinShi-SecularBird: For the above logic, I really suspect (didn't look at code in details) it should be as follows if we want the returned range list is sorted in ascending order. Could you double check? I might be wrong. (Shall we use Bitmap operations to replace the for loop?)

int count = 0;
while (count < rangesPermutations) {
    ...
	
	for ( ; i < position.length; i++) {
	    int x = position.length - 1 - i;
		if (position[x] < ranges.get(x).size - 1) {
			position[x]++;
			break;
		} else {
			position[x] = 0;
		}
	}

	if (i == position.length) {
		break;
	}

	count++;
}

phoenix-core/src/main/java/org/apache/phoenix/compile/ScanRanges.java
```diff
@@ -694,4 +694,127 @@ public TimeRange getRowTimestampRange() {
...
+
+                int[] slotSpans = this.getSlotSpans();
+
+                // If the ranges here do not qualify all the keys then those keys are unbound
+                if (ranges.size() < schema.getMaxFields()) {
+                    int originalSize = ranges.size();
+                    for (int i = 0; i < schema.getMaxFields() - originalSize; i++) {
+                        ranges.add(Lists.newArrayList(KeyRange.EVERYTHING_RANGE));
+                    }
+                    slotSpans = new int[schema.getMaxFields()];
```
BinShi-SecularBird: here the slotSpans[0...this.getSlotSpans().length-1] is wiped off, right? Should it be:

slotSpans = new int[schema.getMaxFields()];
System.arraycopy(this.getSlotSpans(), 0, slotSpans, 0, this.getSlotSpans().length);

phoenix-core/src/main/java/org/apache/phoenix/compile/ScanRanges.java
```diff
@@ -694,4 +694,127 @@ public TimeRange getRowTimestampRange() {
...
+                        upperInclusive = false;
+                    }
+
+                    KeyRange keyRange = KeyRange.getKeyRange(lowerBound, lowerInclusive, upperBound, upperInclusive);
+                    expandedRanges.add(keyRange);
+
+                    // update position
+                    int i;
+                    for (i = 0; i < position.length; i++) {
+
```
BinShi-SecularBird: remove the blank line here?

BinShi-SecularBird:
Review isn't finished yet. Could you handle these feedback first? I still need some time to check whether we have equivalent refactor for local index.

phoenix-core/src/main/java/org/apache/phoenix/compile/ScanRanges.java
```diff
@@ -694,4 +694,127 @@ public TimeRange getRowTimestampRange() {
...
+                if (ranges.size() < schema.getMaxFields()) {
+                    int originalSize = ranges.size();
+                    for (int i = 0; i < schema.getMaxFields() - originalSize; i++) {
+                        ranges.add(Lists.newArrayList(KeyRange.EVERYTHING_RANGE));
+                    }
+                    slotSpans = new int[schema.getMaxFields()];
+                }
+
+                // Product to bound our counting loop for safety
+                int rangesPermutations = 1;
```
BinShi-SecularBird: How about change name to countOfRangesPermutations? rangesPermutations sounds like the permutations of ranges instead of a counter.

phoenix-core/src/main/java/org/apache/phoenix/compile/ScanRanges.java
```diff
@@ -694,4 +694,127 @@ public TimeRange getRowTimestampRange() {
...
+        if (this.isPointLookup()) {
+            queryRowKeyRanges.addAll(this.getRanges().get(0));
+        } else {
+            // If scanRanges.ranges has no information then should be in the scanRanges.scanRange
+            if (this.getRanges().size() == 0) {
+                queryRowKeyRanges.add(this.getScanRange());
+            } else { // We have a composite key need the row key from the combination
+                // make a copy of ranges as we may add items to fully qualify our rowkey
+                List<List<KeyRange>> ranges = new ArrayList<>();
+                ranges.addAll(this.getRanges());
```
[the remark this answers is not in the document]
dbwong: Ok

phoenix-core/src/main/java/org/apache/phoenix/compile/ScanRanges.java
```diff
@@ -694,4 +694,127 @@ public TimeRange getRowTimestampRange() {
...
+                }
+
+                // Product to bound our counting loop for safety
+                int rangesPermutations = 1;
+                for (int i = 0; i < ranges.size(); i++) {
+                    rangesPermutations *= ranges.get(i).size();
+                }
+
+                // Have to construct the intersection
+                List<KeyRange> expandedRanges = Lists.newArrayListWithExpectedSize(rangesPermutations);
```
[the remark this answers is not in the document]
dbwong: Ok

phoenix-core/src/main/java/org/apache/phoenix/compile/ScanRanges.java
```diff
@@ -694,4 +694,127 @@ public TimeRange getRowTimestampRange() {
...
+                int[] position = new int[slotSpans.length];
```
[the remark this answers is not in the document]
dbwong: Ok

phoenix-core/src/main/java/org/apache/phoenix/compile/ScanRanges.java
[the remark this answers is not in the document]
dbwong: Good catch!, refactoring error.

phoenix-core/src/main/java/org/apache/phoenix/compile/ScanRanges.java
```diff
@@ -694,4 +694,127 @@ public TimeRange getRowTimestampRange() {
...
+
+                    result = ScanUtil.setKey(schema, ranges, slotSpans, position, KeyRange.Bound.UPPER, keyBuffer, 0, 0,
+                            slotSpans.length);
+
+                    if (result < 0) {
+                        // unbound
+                        upperBound = KeyRange.UNBOUND;
+                    } else {
+                        upperBound = Arrays.copyOf(keyBuffer, result);
+                    }
```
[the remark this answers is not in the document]
dbwong: I personally dislike using ternary operator outside of final initializations.

phoenix-core/src/main/java/org/apache/phoenix/compile/ScanRanges.java
dbwong: Good catch this would have been an issue for non-fully qualified RVCs.

phoenix-core/src/main/java/org/apache/phoenix/compile/ScanRanges.java
dbwong: Ok

phoenix-core/src/main/java/org/apache/phoenix/compile/ScanRanges.java
dbwong: Sure changed to rangePermutationCount

phoenix-core/src/main/java/org/apache/phoenix/compile/ScanRanges.java
dbwong: You are correct that doing this from the trailing edge will provide the order constant easier as opposed to resorting.  I'm not sure what you intend by bitmap operations as we are not a strict bitmap.  Can you elaborate?

phoenix-core/src/main/java/org/apache/phoenix/iterate/BaseResultIterators.java
```diff
@@ -550,15 +550,100 @@ public BaseResultIterators(QueryPlan plan, Integer perScanLimit, Integer offset,
...
+                    .newArrayListWithExpectedSize(queryRowKeyRanges.size() * regionKeyRanges.size());
+
+            for (KeyRange regionKeyRange : regionKeyRanges) {
+                for (KeyRange queryRowKeyRange : queryRowKeyRanges) {
+                    KeyRange newQueryRowKeyRange = queryRowKeyRange.prependRange(regionKeyRange.getLowerRange(), 0,
+                            regionKeyRange.getLowerRange().length);
+                    newQueryRowKeyRanges.add(newQueryRowKeyRange);
+                }
+            }
+            queryRowKeyRanges = newQueryRowKeyRanges;
```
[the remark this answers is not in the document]
dbwong: Will do as part 2 as discussed.

phoenix-core/src/main/java/org/apache/phoenix/compile/ScanRanges.java
```diff
@@ -694,4 +694,136 @@ public TimeRange getRowTimestampRange() {
...
+                    /*
+                     * This is already considered inside of ScanUtil.setKey we may want to refactor to pull these out.
+                     * range/single    boundary       bound      increment
+                     *  range          inclusive      lower         no
+                     *  range          inclusive      upper         yes, at the end if occurs at any slots.
+                     *  range          exclusive      lower         yes
+                     *  range          exclusive      upper         no
+                     *  single         inclusive      lower         no
+                     *  single         inclusive      upper         yes, at the end if it is the last slots.
+                     */
```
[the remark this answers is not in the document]
dbwong: Sure

dbwong:
@twdsilva Stats related PR.  This is going into a Stats integration branch that @karanmehta93 made for us.

phoenix-core/src/main/java/org/apache/phoenix/iterate/BaseResultIterators.java
```diff
@@ -550,15 +550,48 @@ public BaseResultIterators(QueryPlan plan, Integer perScanLimit, Integer offset,
...
+            }
+
+            // Region is upperInclusive false by definition
+            byte[] endKey = regionInfo.getEndKey();
+            // The last region has an unbound upper key.
+            if (i == regionLocations.size() - 1) {
+                endKey = KeyRange.UNBOUND;
+            }
+
+            KeyRange range = KeyRange.getKeyRange(startKey, lowerInclusive, regionInfo.getEndKey(), false);
```
karanmehta93: `regionInfo.getEndKey()` or `endKey`?

karanmehta93: Looks like it should not really matter since `KeyRange.UNBOUND` and `HConstants.EMPTY_BYTE_ARRAY` are the same. Should we just refactor these static constants? Some of this code exists from 2014. You can also remove line 56 if you want.

phoenix-core/src/main/java/org/apache/phoenix/iterate/BaseResultIterators.java
```diff
@@ -550,15 +550,48 @@ public BaseResultIterators(QueryPlan plan, Integer perScanLimit, Integer offset,
         }
         return ranges;
     }
-    
+
+    private List<KeyRange> getRegionRowKeyRanges() throws SQLException {
```
karanmehta93: I might be missing some context here but I see this piece of code in this file which can re-used and handles the required corner cases.
```
    static final Function<HRegionLocation, KeyRange> TO_KEY_RANGE = new Function<HRegionLocation, KeyRange>() {
        @Override
        public KeyRange apply(HRegionLocation region) {
            return KeyRange.getKeyRange(region.getRegionInfo().getStartKey(), region.getRegionInfo().getEndKey());
        }
    };
```

karanmehta93: Can you add some comments here as to how this method will help?

phoenix-core/src/main/java/org/apache/phoenix/compile/ScanRanges.java
```diff
@@ -694,4 +694,136 @@ public TimeRange getRowTimestampRange() {
...
+        // as schema is not filled in
+        if (isEverything()) { return Lists.newArrayList(KeyRange.EVERYTHING_RANGE); }
+        if (isDegenerate()) { return Lists.newArrayList(KeyRange.EMPTY_RANGE); }
+
+        List<KeyRange> queryRowKeyRanges = Lists.newArrayListWithExpectedSize(this.ranges.size());
+
+        // Point lookups are stored in the first range as a whole and already a rowkey
+        // see ScanRanges.getPointLookupKeyIterator
+        if (this.isPointLookup()) {
+            queryRowKeyRanges.addAll(this.getRanges().get(0));
```
karanmehta93: nit: use `ranges` directly.

phoenix-core/src/main/java/org/apache/phoenix/compile/ScanRanges.java
```diff
@@ -694,4 +694,136 @@ public TimeRange getRowTimestampRange() {
         return rowTimestampRange;
     }
 
+    /**
+     * Produces the list of KeyRanges representing the fully qualified row key by calling into ScanUtil.setKey
+     * repeatedly for every combination of KeyRanges in the ranges field. The bounds will be set according to the
+     * properties of the setKey method.
+     * 
+     * @return list of KeyRanges representing the fully qualified rowkey, coalesced
```
BinShi-SecularBird: @dbwong, is the upper bound of each generated key range is exclusive? If it isn't, can we call nextKey for the upper bound when it is inclusive?

phoenix-core/src/main/java/org/apache/phoenix/compile/ScanRanges.java
```diff
@@ -694,4 +694,136 @@ public TimeRange getRowTimestampRange() {
...
+                // Have to construct the intersection
+                List<KeyRange> expandedRanges = Lists.newArrayListWithCapacity(rangesPermutationsCount);
+                int[] positions = new int[slotSpans.length];
+
+                int maxKeyLength = SchemaUtil.getMaxKeyLength(schema, newRanges);
+                byte[] keyBuffer = new byte[maxKeyLength];
+
+                // Counting Loop
+                int count = 0;
+                while (count < rangesPermutationsCount) {
```
karanmehta93: Do we really want to generate everything and store in memory for the client? How about using iterator approach instead?

phoenix-core/src/main/java/org/apache/phoenix/compile/ScanRanges.java
```diff
@@ -694,4 +694,136 @@ public TimeRange getRowTimestampRange() {
...
+        if (isDegenerate()) { return Lists.newArrayList(KeyRange.EMPTY_RANGE); }
+
+        List<KeyRange> queryRowKeyRanges = Lists.newArrayListWithExpectedSize(this.ranges.size());
+
+        // Point lookups are stored in the first range as a whole and already a rowkey
+        // see ScanRanges.getPointLookupKeyIterator
+        if (this.isPointLookup()) {
+            queryRowKeyRanges.addAll(this.getRanges().get(0));
+        } else {
+            // If scanRanges.ranges has no information then should be in the scanRanges.scanRange
```
karanmehta93: Can you elaborate a bit here? What is the difference between `ScanRanges` and `KeyRanges`?

phoenix-core/src/main/java/org/apache/phoenix/compile/ScanRanges.java
karanmehta93: Also do we also want to refactor the `SkipScanFilter` class too since it relies on the same logic?

phoenix-core/src/main/java/org/apache/phoenix/compile/ScanRanges.java
```diff
@@ -694,4 +694,136 @@ public TimeRange getRowTimestampRange() {
...
+                        lowerInclusive = false;
+                        upperInclusive = false;
+                    }
+
+                    KeyRange keyRange = KeyRange.getKeyRange(lowerBound, lowerInclusive, upperBound, upperInclusive);
+                    expandedRanges.add(keyRange);
+
+                    // update position
+                    int i;
+                    for (i = positions.length - 1; i >= 0; i--) {
```
karanmehta93: Can you explain the usage of this loop?

phoenix-core/src/test/java/org/apache/phoenix/compile/ScanRangesIntersectTest.java
```diff
@@ -58,12 +64,344 @@ private static void assertIntersect(ScanRanges ranges, String lowerRange, String
...
+            return SortOrder.getDefault();
+        }
+    };
+
+    // Does not handle some edge conditions like empty string
+    private String handleScanNextKey(String key) {
+        char lastChar = key.charAt(key.length() - 1);
+        return key.substring(0, key.length() - 1) + String.valueOf((char)(lastChar + 1));
+    }
+
```
karanmehta93: Can you add tests for `EVERYTHING_RANGE` and `EMPTY_RANGE` as well?

dbwong: Done.

phoenix-core/src/main/java/org/apache/phoenix/compile/ScanRanges.java
dbwong: Added additional comments but this is a counting loop from the trailing edge to the leading edge for all possible settings of the Keys based on the ranges.  That is for PK (key1,key2) and Where clause key1 in {a,b,c} and key2 in {x,y,z}, then we need to generate keys ax,ay,az,bx,by,bz,cx,cy,cz.  See the setKey method for more details.

phoenix-core/src/main/java/org/apache/phoenix/compile/ScanRanges.java
dbwong: This was removed as part of the handling of Bin's request for exclusive upper.  With this there is no reason to handle pointGets differently.

phoenix-core/src/main/java/org/apache/phoenix/compile/ScanRanges.java
dbwong: Sure.

phoenix-core/src/main/java/org/apache/phoenix/compile/ScanRanges.java
dbwong: Changed

phoenix-core/src/main/java/org/apache/phoenix/iterate/BaseResultIterators.java
dbwong: Good catch actually.  endKey is correct a typo.

phoenix-core/src/main/java/org/apache/phoenix/iterate/BaseResultIterators.java
dbwong: Added comments and removed the TO_KEY_RANGE dead code.

phoenix-core/src/main/java/org/apache/phoenix/compile/ScanRanges.java
dbwong: I think using in memory is fine for a first approach.  I discussed this briefly with @BinShi-SecularBird and we felt that even large number of ranges should be okay for a first pass.  Some estimates with a 100 byte key, and 100 keys, with 100 settings, is ~1 mb. Is this a current concern for the memory footprint?  Second with an iterator approach the next steps of intersecting with region boundaries and with guidepost tree structure may be less efficient so we are trading memory for execution time.

dbwong: Can you be more specific on what you think should be refactored in SkipScanFilter?

twdsilva:
@dbwong Can you please rebase and squash the commits? I will get this committed.

PR #471, opened 2019-03-19
PHOENIX-5176 KeyRange.compareUpperRange(KeyRang 1, KeyRang 2) returns wrong result when two key ranges have the same upper bound values but one is inclusive and another is exclusive

BinShi-SecularBird:
#### What does this PR do?
Which problem(s) does it fix and how? What does this PR add?

#### Where should the reviewer start?
Explain what should be set-up first in order to get the review started. Which page has to be viewed etc.

#### How should this be manually tested?
- [ ] Write down all
- [ ] the steps a person
- [ ] needs to take to test it.

##### Screenshots (if appropriate)
If this PR changes something that concerns UI changes, please post before and after screenshots here.

##### Other documents
If you have any documents that have anything to do with this PR.

#### Who should be notified?
- [ ] Write down the name/department this branch needs
- [ ] to be communicated to
- [ ] Check the box(es) to indicate that it has been
- [ ] communicated (add the label)

#### What should happen on deployment? (check all that apply)
- [x] The usual steps.
- [ ] There are database changes, which should be run.
- [ ] There are files that should be manually uploaded.

#### Questions:
- Does this require a blog post?
- Does this require a knowledge base update?
- Does support need training for this?
- Does this has to be communicated to partners?

### DIFF
--- a/phoenix-core/src/main/java/org/apache/phoenix/query/KeyRange.java
+++ b/phoenix-core/src/main/java/org/apache/phoenix/query/KeyRange.java
@@ -519,7 +519,7 @@ public class KeyRange implements Writable {
         return Lists.transform(keys, POINT);
     }
 
-    private static int compareUpperRange(KeyRange rowKeyRange1,KeyRange rowKeyRange2) {
+    public static int compareUpperRange(KeyRange rowKeyRange1,KeyRange rowKeyRange2) {
         int result = Boolean.compare(rowKeyRange1.upperUnbound(), rowKeyRange2.upperUnbound());
         if (result != 0) {
             return result;
@@ -528,7 +528,7 @@ public class KeyRange implements Writable {
         if (result != 0) {
             return result;
         }
-        return Boolean.compare(rowKeyRange2.isUpperInclusive(), rowKeyRange1.isUpperInclusive());
+        return Boolean.compare(rowKeyRange1.isUpperInclusive(), rowKeyRange2.isUpperInclusive());
     }
 
     public static List<KeyRange> intersect(List<KeyRange> rowKeyRanges1, List<KeyRange> rowKeyRanges2) {
--- a/phoenix-core/src/test/java/org/apache/phoenix/query/KeyRangeMoreTest.java
+++ b/phoenix-core/src/test/java/org/apache/phoenix/query/KeyRangeMoreTest.java
@@ -23,6 +23,7 @@ import java.util.Arrays;
 import java.util.Collections;
 import java.util.List;
 
+import com.google.common.collect.Lists;
 import org.apache.hadoop.hbase.util.Bytes;
 import org.apache.phoenix.schema.types.PInteger;
 import org.junit.Test;
@@ -126,8 +127,7 @@ public class KeyRangeMoreTest extends TestCase {
             assertResult(result, maxStart,minEnd);
             Collections.shuffle(rowKeyRanges1);
             Collections.shuffle(rowKeyRanges2);
-
-        };
+        }
     }
 
     private void assertResult(List<KeyRange> result,int start,int end) {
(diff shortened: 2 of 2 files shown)


BinShi-SecularBird:
@dbwong @karanmehta93 @yanxinyi

phoenix-core/src/main/java/org/apache/phoenix/query/KeyRange.java
```diff
@@ -519,7 +520,8 @@ public KeyRange union(KeyRange other) {
         return Lists.transform(keys, POINT);
     }
 
-    private static int compareUpperRange(KeyRange rowKeyRange1,KeyRange rowKeyRange2) {
+    @VisibleForTesting
```
BinShi-SecularBird: I'll remove @VisibleForTesting. This needs to be public, because it is used by changes that I'm working on.

phoenix-core/src/test/java/org/apache/phoenix/query/KeyRangeMoreTest.java
```diff
@@ -192,64 +193,78 @@ public void testListIntersectForBoundary() throws Exception {
...
+                Arrays.asList(true, true));
         listIntersectAndAssert(rowKeyRanges1, rowKeyRanges2, expected);
     }
 
+    @Test
+    public void testKeyRangeCompareUpperRange() throws Exception {
+        List<KeyRange> rowKeyRanges1 = CreateKeyRangeListWithFixedLowerRange(
+                Arrays.asList(Integer.MAX_VALUE, Integer.MAX_VALUE, 10000, 1001, 1000, 1000, 1000, 1000),
+                Arrays.asList(false, false, true, true, true, true, false, true));
+        List<KeyRange> rowKeyRanges2 = CreateKeyRangeListWithFixedLowerRange(
```
priyankporwal: Add some keys with strings in key ranges?

phoenix-core/src/test/java/org/apache/phoenix/query/KeyRangeMoreTest.java
```diff
@@ -192,64 +193,78 @@ public void testListIntersectForBoundary() throws Exception {
...
+                Arrays.asList(Integer.MAX_VALUE, 10000, Integer.MAX_VALUE, 1000, 1001, 1000, 1000, 1000),
+                Arrays.asList(false, false, false, true, true, true, false, false));
```
dbwong: For completion we are missing {0-1000 false} compared to {0-1000 true} as second since the original problem was order issue may want to add.

phoenix-core/src/test/java/org/apache/phoenix/query/KeyRangeMoreTest.java
```diff
@@ -192,64 +193,78 @@ public void testListIntersectForBoundary() throws Exception {
...
+                Arrays.asList(3, 5, 6, 7),
+                Arrays.asList(true, true, true, true));
+        expected = CreateKeyRangeList(
+                Arrays.asList(3, 5),
+                Arrays.asList(true, true));
         listIntersectAndAssert(rowKeyRanges1, rowKeyRanges2, expected);
     }
 
+    @Test
+    public void testKeyRangeCompareUpperRange() throws Exception {
```
dbwong: In general I dislike this type of many tests in a single junit test as the failures are non-descriptive and the cases under test are not readable.

phoenix-core/src/test/java/org/apache/phoenix/query/KeyRangeMoreTest.java
```diff
@@ -192,64 +193,78 @@ public void testListIntersectForBoundary() throws Exception {
...
+            byte[] startKey = keys.get(2*i).equals(Integer.MIN_VALUE) ? KeyRange.UNBOUND : PInteger.INSTANCE.toBytes(keys.get(2*i));
+            byte[] endKey = keys.get(2*i + 1).equals(Integer.MAX_VALUE) ? KeyRange.UNBOUND : PInteger.INSTANCE.toBytes(keys.get(2*i + 1));
+            keyRangeList.add(PInteger.INSTANCE.getKeyRange(startKey, boundaryConditions.get(2*i), endKey, boundaryConditions.get(2*i+1)));
+        }
+
+        return keyRangeList;
+    }
+
     private static void listIntersectAndAssert(List<KeyRange> rowKeyRanges1,List<KeyRange> rowKeyRanges2,List<KeyRange> expected) {
         for(int i=0;i<200;i++) {
```
dbwong: I see you didn't change this test case but as case as it is non-deterministic and the number of permutations are O(n!) which grows large even on small subsets beyond the scope of a normal rng period.

phoenix-core/src/test/java/org/apache/phoenix/query/KeyRangeMoreTest.java
```diff
@@ -192,64 +193,78 @@ public void testListIntersectForBoundary() throws Exception {
...
+        assertEquals(rowKeyRanges1.size(), rowKeyRanges2.size());
+        assertEquals(rowKeyRanges1.size(), expectedResults.size());
+
+        for (int i = 0; i < expectedResults.size(); i++) {
+            int compareResult = KeyRange.compareUpperRange(rowKeyRanges1.get(i), rowKeyRanges2.get(i));
+            assertEquals(expectedResults.get(i).intValue(), compareResult);
+        }
+    }
+
+    private static List<KeyRange> CreateKeyRangeListWithFixedLowerRange(List<Integer> keys, List<Boolean> boundaryConditions) {
```
karanmehta93: nit: function names start with lowercase letters.

phoenix-core/src/test/java/org/apache/phoenix/query/KeyRangeMoreTest.java
BinShi-SecularBird: I prefer to group them. It's efficient for test code author/reviewer/debugger.

phoenix-core/src/test/java/org/apache/phoenix/query/KeyRangeMoreTest.java
BinShi-SecularBird: good catch. Thanks. Resolved

phoenix-core/src/test/java/org/apache/phoenix/query/KeyRangeMoreTest.java
BinShi-SecularBird: The purpose of the original test author just want to randomly shuffle the ranges in list for certain times and make sure the function can correct sort the given tow ranges and calculate the intersection. I don't like this test code because once there is failure we don't know what's the case which caused the failure, but I think improving this is out of scope of my change due to the following two reasons:
1. Using the ranges in ascending order is efficient for testing my fix to the original problem.
2. Reshuffle the ranges should be the unit test of KeyRange.coalesce(...) instead of KeyRange.intersect().

phoenix-core/src/test/java/org/apache/phoenix/query/KeyRangeMoreTest.java
BinShi-SecularBird: It is not necessary. It doesn't test more after the strings being covered to bytes.

phoenix-core/src/test/java/org/apache/phoenix/query/KeyRangeMoreTest.java
BinShi-SecularBird: Make sense. Added the test case.

phoenix-core/src/test/java/org/apache/phoenix/query/KeyRangeMoreTest.java
dbwong: This is okay with me.

phoenix-core/src/test/java/org/apache/phoenix/query/KeyRangeMoreTest.java
dbwong: While the code is smaller it is more difficult to understand what cases are under test.  I'm okay as we are following pattern in the file.  For new code though I'd prefer we follow junit philosophy.  See https://junit.org/junit4/faq.html#atests_12

dbwong:
@twdsilva stats related JIRA, I'm mostly fine with this changes please take a look.

phoenix-core/src/test/java/org/apache/phoenix/query/KeyRangeMoreTest.java
```diff
@@ -23,6 +23,7 @@
 import java.util.Collections;
 import java.util.List;
 
+import org.apache.curator.shaded.com.google.common.collect.Lists;
```
swaroopak: Do we really need this import? Can we use JDK standard Lists?

BinShi-SecularBird: org.apache.curator.shaded.com.google.common.collect.Lists is used by old test code in this file and has nothing to do with my change. I searched in phoenix-core, there are test files using it. If we want to unify the coding style/rules and replace all the places using org.apache.curator.shaded.com.google.common.collect.Lists with JDK standard Lists, we should open an JIRA and make the change in whole code base, but it is orthogonal to and shouldn't be addressed in my change.

BinShi-SecularBird:
@twdsilva @dbwong Do you have any more comments?

twdsilva:
+1 LGTM

twdsilva:
I agree with @dbwong 's comment about using guava.

phoenix-core/src/test/java/org/apache/phoenix/query/KeyRangeMoreTest.java
BinShi-SecularBird: I'll use "import com.google.common.collect.Lists;" instead of shaded jar

PR #473, opened 2019-03-25
PHOENIX-5138 - ViewIndexId sequences created after PHOENIX-5132 shoul…

gjacoby126:
…dn't collide with ones created before it

### DIFF
--- a/phoenix-core/src/it/java/org/apache/phoenix/end2end/UpgradeIT.java
+++ b/phoenix-core/src/it/java/org/apache/phoenix/end2end/UpgradeIT.java
@@ -21,6 +21,7 @@ import static com.google.common.base.Preconditions.checkNotNull;
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertFalse;
 import static org.junit.Assert.assertNotEquals;
+import static org.junit.Assert.assertNull;
 import static org.junit.Assert.assertTrue;
 import static org.junit.Assert.fail;
 
@@ -30,6 +31,7 @@ import java.sql.DriverManager;
 import java.sql.PreparedStatement;
 import java.sql.ResultSet;
 import java.sql.SQLException;
+import java.util.List;
 import java.util.Properties;
 import java.util.Set;
 import java.util.concurrent.Callable;
@@ -38,6 +40,7 @@ import java.util.concurrent.FutureTask;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicInteger;
 
+import com.google.common.collect.Lists;
 import org.apache.curator.shaded.com.google.common.collect.Sets;
 import org.apache.hadoop.hbase.TableName;
 import org.apache.hadoop.hbase.client.Admin;
@@ -60,6 +63,9 @@ import org.apache.phoenix.schema.PNameFactory;
 import org.apache.phoenix.schema.PTable;
 import org.apache.phoenix.schema.PTable.LinkType;
 import org.apache.phoenix.schema.PTableType;
+import org.apache.phoenix.schema.SequenceAllocation;
+import org.apache.phoenix.schema.SequenceKey;
+import org.apache.phoenix.util.EnvironmentEdgeManager;
 import org.apache.phoenix.util.MetaDataUtil;
 import org.apache.phoenix.util.PhoenixRuntime;
 import org.apache.phoenix.util.PropertiesUtil;
@@ -178,25 +184,10 @@ public class UpgradeIT extends ParallelStatsDisabledIT {
             }
             PName tenantId = phxConn.getTenantId();
             PName physicalName = PNameFactory.newName(hbaseTableName);
-            String oldSchemaName = MetaDataUtil.getViewIndexSequenceSchemaName(PNameFactory.newName(phoenixFullTableName),
-                    false);
             String newSchemaName = MetaDataUtil.getViewIndexSequenceSchemaName(physicalName, true);
             String newSequenceName = MetaDataUtil.getViewIndexSequenceName(physicalName, tenantId, true);
-            ResultSet rs = phxConn.createStatement()
-                    .executeQuery("SELECT " + PhoenixDatabaseMetaData.CURRENT_VALUE + "  FROM "
-                            + PhoenixDatabaseMetaData.SYSTEM_SEQUENCE + " WHERE " + PhoenixDatabaseMetaData.TENANT_ID
-                            + " IS NULL AND " + PhoenixDatabaseMetaData.SEQUENCE_SCHEMA + " = '" + newSchemaName
-                            + "' AND " + PhoenixDatabaseMetaData.SEQUENCE_NAME + "='" + newSequenceName + "'");
-            assertTrue(rs.next());
-            assertEquals("-9223372036854775805", rs.getString(1));
-            rs = phxConn.createStatement().executeQuery("SELECT " + PhoenixDatabaseMetaData.SEQUENCE_SCHEMA + ","
-                    + PhoenixDatabaseMetaData.SEQUENCE_SCHEMA + "," + PhoenixDatabaseMetaData.CURRENT_VALUE + "  FROM "
-                    + PhoenixDatabaseMetaData.SYSTEM_SEQUENCE + " WHERE " + PhoenixDatabaseMetaData.TENANT_ID
-                    + " IS NULL AND " + PhoenixDatabaseMetaData.SEQUENCE_SCHEMA + " = '" + oldSchemaName + "'");
-            assertFalse(rs.next());
-            phxConn.close();
+            verifySequenceValue(null, newSequenceName, newSchemaName, -9223372036854775805L);
             admin.close();
-   
         }
     }
 
@@ -507,12 +498,20 @@ public class UpgradeIT extends ParallelStatsDisabledIT {
         return DriverManager.getConnection(getUrl(), props);
     }
 
-    private Connection getConnection(boolean tenantSpecific, String tenantId) throws SQLException {
+    private Connection getConnection(boolean tenantSpecific, String tenantId, boolean isNamespaceMappingEnabled)
+        throws SQLException {
         if (tenantSpecific) {
             checkNotNull(tenantId);
             return createTenantConnection(tenantId);
         }
-        return DriverManager.getConnection(getUrl());
+        Properties props = new Properties();
+        if (isNamespaceMappingEnabled){
+            props.setProperty(QueryServices.IS_NAMESPACE_MAPPING_ENABLED, "true");
+        }
+        return DriverManager.getConnection(getUrl(), props);
+    }
+    private Connection getConnection(boolean tenantSpecific, String tenantId) throws SQLException {
+        return getConnection(tenantSpecific, tenantId, false);
     }
     
     @Test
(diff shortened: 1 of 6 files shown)


phoenix-core/src/it/java/org/apache/phoenix/end2end/UpgradeIT.java
```diff
@@ -588,4 +603,85 @@ public void testMoveParentChildLinks() throws Exception {
...
+        List<SequenceAllocation> afterUpgradeAllocations = Lists.newArrayList();
+        SequenceKey sequenceUpgrade = MetaDataUtil.getViewIndexSequenceKey(null, viewIndexTable, 0, isNamespaceMappingEnabled);
+        afterUpgradeAllocations.add(new SequenceAllocation(sequenceUpgrade, 1));
+        long[] afterUpgradeValues = new long[1];
+        SQLException[] afterUpgradeExceptions = new SQLException[1];
+        cqs.incrementSequences(afterUpgradeAllocations, EnvironmentEdgeManager.currentTimeMillis(), afterUpgradeValues, afterUpgradeExceptions);
+
+        assertNull(afterUpgradeExceptions[0]);
+        if (isNamespaceMappingEnabled){
+            //since one sequence (the global one) will be reused as the "new" sequence,
```
gokceni: Why isNamespaceMappingEnabled and cache related? It is not super obvious why it is incremented 1 in the namespace case.

phoenix-core/src/main/java/org/apache/phoenix/util/UpgradeUtil.java
```diff
@@ -2310,6 +2316,98 @@ private static void mapChildViewsToNamespace(String connUrl, Properties props, L
...
+            for (SequenceKey sequenceKey : existingSequenceKeys){
+                long[] currentValueArray = new long[1];
+                SQLException[] sqlExceptions = new SQLException[1];
+                cqs.incrementSequences(
+                    Lists.newArrayList(new SequenceAllocation(sequenceKey, 1L)),
+                    EnvironmentEdgeManager.currentTimeMillis(),
+                    currentValueArray, new SQLException[1]);
+
+                if (sqlExceptions[0] != null) {
+                    continue;
```
gokceni: Do you want to log this exception? What happens in this case?

phoenix-core/src/main/java/org/apache/phoenix/util/UpgradeUtil.java
```diff
@@ -2310,6 +2316,98 @@ private static void mapChildViewsToNamespace(String connUrl, Properties props, L
...
+    public static void mergeViewIndexIdSequences(ConnectionQueryServices cqs, PhoenixConnection metaConnection)
+        throws SQLException{
+         /* before PHOENIX-5132, there was a per-tenant sequence to generate view index ids,
+           which could cause problems if global and tenant-owned view indexes were mixed for the
+           same physical base table. Now there's just one sequence for all view indexes of the same
+           physical table, but we have to check to see if there are any legacy sequences, and
+           merge them into a single sequence equal to max + 1 of the largest legacy sequence
+           to avoid collisons.
+         */
+        Map<String, List<SequenceKey>> sequenceTableMap = new HashMap<>();
```
twdsilva: maybe add a comment saying this is storing a mapping between physical table to list of view index sequences

phoenix-core/src/main/java/org/apache/phoenix/util/UpgradeUtil.java
```diff
@@ -2310,6 +2316,98 @@ private static void mapChildViewsToNamespace(String connUrl, Properties props, L
...
+            long maxViewIndexId = Long.MIN_VALUE;
+            PName name = PNameFactory.newName(baseTableName);
+            boolean hasNamespaceMapping =
+                SchemaUtil.isNamespaceMappingEnabled(null, cqs.getConfiguration()) ||
+                    cqs.getProps().getBoolean(QueryServices.IS_NAMESPACE_MAPPING_ENABLED, false);
+            List<SequenceKey> existingSequenceKeys = sequenceTableMap.get(baseTableName);
+            for (SequenceKey sequenceKey : existingSequenceKeys){
+                long[] currentValueArray = new long[1];
+                SQLException[] sqlExceptions = new SQLException[1];
+                cqs.incrementSequences(
```
twdsilva: Nice idea, incrementing the sequence by 1 to get the current max.

phoenix-core/src/main/java/org/apache/phoenix/util/UpgradeUtil.java
```diff
@@ -2310,6 +2316,98 @@ private static void mapChildViewsToNamespace(String connUrl, Properties props, L
...
+                if (currentValueArray[0] > maxViewIndexId){
+                    maxViewIndexId = currentValueArray[0];
+                }
+                currentSequenceValues.put(sequenceKey, currentValueArray[0]);
+            }
+            try {
+                //In one case (namespaced-mapped base table, global view index), the new sequence
+                //is the same as the old sequence, so rather than create it we just increment it
+                //to the right value.
+                SequenceKey newSequenceKey = new SequenceKey(null, MetaDataUtil.getViewIndexSequenceSchemaName(name, hasNamespaceMapping),
```
twdsilva: I think there is an extreme edge case if some client is trying to create a view index at the same time the upgrade is happening, so I suggest we add an extra buffer of 100 or something to the maxViewIndexId.

phoenix-core/src/main/java/org/apache/phoenix/util/UpgradeUtil.java
twdsilva: In case you can't increment this particular sequence then you wouldn't include it in your maxViewIndexId calculation. I wonder if if we should fail the upgrade with an exception in this case?

phoenix-core/src/it/java/org/apache/phoenix/end2end/UpgradeIT.java
```diff
@@ -588,4 +602,85 @@ public void testMoveParentChildLinks() throws Exception {
...
+        long val2 = 100;
+        long val3 = 1000;
+        allocations.add(new SequenceAllocation(sequenceOne, val1));
+        allocations.add(new SequenceAllocation(sequenceGlobal, val2));
+        allocations.add(new SequenceAllocation(sequenceTwo, val3));
+
+
+        long[] incrementedValues = new long[3];
+        SQLException[] exceptions = new SQLException[3];
+        cqs.incrementSequences(allocations, EnvironmentEdgeManager.currentTimeMillis(), incrementedValues,
```
twdsilva: Maybe add a comment saying simulate creating many view indexes.

phoenix-core/src/it/java/org/apache/phoenix/end2end/UpgradeIT.java
gjacoby126: For some historical reason, the sequence names were very different for namespace-mapped base tables, and in one case it coincides with the new naming convention. So in that case the "new" sequence is an old sequence that's been incremented an extra time, hence the extra 1

phoenix-core/src/main/java/org/apache/phoenix/util/UpgradeUtil.java
gjacoby126: @twdsilva -- yeah, I'd thought of that and was torn. On the one hand, skipping it like I'm doing here can lead to collisions in edge cases. On the other hand, I didn't want to permanently crash an upgrade if someone had exhausted a single sequence. Happy to hear suggestions. :-)

Maybe only continue for certain kinds of exceptions? If so, which ones?

gjacoby126:
@twdsilva - Think I covered all of your comments (and fixed a test failure that was assuming the old sequence logic). Could you please take another look when you get a minute to make sure I got everything?

phoenix-core/src/main/java/org/apache/phoenix/util/UpgradeUtil.java
twdsilva: Can you log an error saying that we weren't able to increment the sequence and that it has to be handled manually?

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/MetaDataEndpointImpl.java
```diff
@@ -2502,6 +2483,32 @@ public void createTable(RpcController controller, CreateTableRequest request,
...
+            nSequenceSaltBuckets, parentTable.isNamespaceMapped() );
+        // Earlier sequence was created at (SCN-1/LATEST_TIMESTAMP) and incremented at the client max(SCN,dataTable.getTimestamp), but it seems we should
+        // use always LATEST_TIMESTAMP to avoid seeing wrong sequence values by different connection having SCN
+        // or not.
+        long sequenceTimestamp = HConstants.LATEST_TIMESTAMP;
+        try {
+            connection.getQueryServices().createSequence(key.getTenantId(), key.getSchemaName(), key.getSequenceName(),
+                Long.MIN_VALUE, 1, 1, Long.MIN_VALUE, Long.MAX_VALUE, false, sequenceTimestamp);
+        } catch (SequenceAlreadyExistsException e) {
+        }
```
swaroopak: It would be nice to have a log line added here.

phoenix-core/src/main/java/org/apache/phoenix/util/UpgradeUtil.java
```diff
@@ -2243,15 +2249,14 @@ private static void updateIndexesSequenceIfPresent(PhoenixConnection connection,
...
-        String upsert = "UPSERT INTO " + PhoenixDatabaseMetaData.SYSTEM_SEQUENCE + " SELECT  REGEXP_SPLIT("
-                + PhoenixDatabaseMetaData.SEQUENCE_NAME + ",'_')[3] ,\'" + newSchemaName + "\',\'" + newSequenceName
+        String upsert = "UPSERT INTO " + PhoenixDatabaseMetaData.SYSTEM_SEQUENCE + " SELECT NULL,\'" + newSchemaName +
+            "\',\'" + newSequenceName
                 + "\'," + START_WITH + "," + CURRENT_VALUE + "," + INCREMENT_BY + "," + CACHE_SIZE + "," + MIN_VALUE
                 + "," + MAX_VALUE + "," + CYCLE_FLAG + "," + LIMIT_REACHED_FLAG + " FROM "
                 + PhoenixDatabaseMetaData.SYSTEM_SEQUENCE + " WHERE " + PhoenixDatabaseMetaData.TENANT_ID
                 + " IS NULL AND " + PhoenixDatabaseMetaData.SEQUENCE_SCHEMA + " = '" + oldSchemaName + "'";
         connection.createStatement().executeUpdate(upsert);
         // delete old sequence
```
swaroopak: nit: unrelated comment
one of the reasons, I don't prefer commenting inline. :D

phoenix-core/src/it/java/org/apache/phoenix/end2end/index/ViewIndexIT.java
```diff
@@ -153,15 +153,10 @@ public void testDeleteViewIndexSequences() throws Exception {
         conn2.createStatement().executeQuery("SELECT * FROM " + fullTableName).next();
         String sequenceName = getViewIndexSequenceName(PNameFactory.newName(fullTableName), null, isNamespaceMapped);
         String sequenceSchemaName = getViewIndexSequenceSchemaName(PNameFactory.newName(fullTableName), isNamespaceMapped);
-        String seqName = getViewIndexSequenceName(PNameFactory.newName(fullTableName), null, !isNamespaceMapped);
-        String seqSchemaName = getViewIndexSequenceSchemaName(PNameFactory.newName(fullTableName), !isNamespaceMapped);
         verifySequenceValue(null, sequenceName, sequenceSchemaName, -9223372036854775807L);
         verifySequenceValue(null, sequenceName, sequenceSchemaName, -9223372036854775807L);
         conn1.createStatement().execute("CREATE INDEX " + indexName + "_2 ON " + fullViewName + " (v1)");
         verifySequenceValue(null, sequenceName, sequenceSchemaName, -9223372036854775806L);
```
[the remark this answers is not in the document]
gjacoby126: The test is verifying that a CREATE INDEX statement on a view will increment the sequence. And yes, it would be more clear with a constant, because they're offsets of Long.MIN_VALUE.

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/MetaDataEndpointImpl.java
gjacoby126: The logic here is essentially "CREATE SEQUENCE Foo IF NOT EXISTS", so the fact that it already exists isn't really relevant, which is why we eat the exception. I'm going to add a comment saying so.

phoenix-core/src/it/java/org/apache/phoenix/end2end/UpgradeIT.java
```diff
@@ -588,4 +587,87 @@ public void testMoveParentChildLinks() throws Exception {
...
     }
 
+    @Test
+    public void testMergeViewIndexSequences() throws Exception {
+        testMergeViewIndexSequencesHelper(false);
+    }
+
+    @Test
+    public void testMergeViewIndexSequencesWithNamespaces() throws Exception {
+        testMergeViewIndexSequencesHelper(true);
```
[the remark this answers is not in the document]
gjacoby126: This is really the only case where you need parameterization in this test suite, so just having the two tests with a common helper is easier (and quicker, since you don't get an extra test run from the other tests with a parameter it doesn't need.)

phoenix-core/src/it/java/org/apache/phoenix/end2end/UpgradeIT.java
```diff
@@ -38,6 +40,7 @@
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicInteger;
 
+import com.google.common.collect.Lists;
```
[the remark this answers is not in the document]
gjacoby126: I don't believe we have a policy as a community against using Guava.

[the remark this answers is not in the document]
gjacoby126: (As a member of the community, @swaroopak , you are of course free to propose such a policy, but that should be a discussion for the larger community to debate, rather than an issue for a single PR.)

gjacoby126:
Thanks @twdsilva, @swaroopak and @gokceni

phoenix-core/src/it/java/org/apache/phoenix/end2end/UpgradeIT.java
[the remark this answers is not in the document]
swaroopak: My bad, I didn't realize it was coming from Guava. :)

PR #476, opened 2019-04-05
PHOENIX-5228 use slf4j for logging in phoenix project

yanxinyi:


phoenix-core/src/it/java/org/apache/hadoop/hbase/regionserver/wal/WALRecoveryRegionPostOpenIT.java
```diff
@@ -75,14 +73,16 @@
...
+import org.slf4j.LoggerFactory;
 
 import com.google.common.collect.Maps;
 import com.google.common.collect.Multimap;
 
 @Category(NeedsOwnMiniClusterTest.class)
 public class WALRecoveryRegionPostOpenIT extends BaseTest {
 
-    private static final Log LOG = LogFactory.getLog(WALRecoveryRegionPostOpenIT.class);
+    private static final Logger logger = LoggerFactory.getLogger(WALRecoveryRegionPostOpenIT.class);
```
swaroopak: I understand that using logger is tempting. But, if we do not change the variable name here, it will reduce the change in the files. 

Also, variables with such access modifiers are capitalized, for the consistency, do you want to follow the same?

yanxinyi: I didn't plan to change var name in the first place until I got Mihir's comment. 
See https://issues.apache.org/jira/browse/PHOENIX-5228
`We should change variable names for log object also, from LOG to logger , it would be simpler and common across the phoenix project. `

BinShi-SecularBird:
@yanxinyi , could you give me some context for switching to slf4j logging?

PR #477, opened 2019-04-05
PHOENIX-5230 Fix ChangePermissionsIT and TableDDLPermissionIT on master

twdsilva:
@karanmehta93  @swaroopak  Please review. I moved all the tests from ChangePermissionsIT, SystemTablePermissionsIT and TableDDLPermissionIT into BasePermissionsIT and created two subclasses one each for namespaces enabled/disabled. I only create a single minicluster for each class.


phoenix-core/src/it/java/org/apache/phoenix/end2end/BasePermissionsIT.java
```diff
@@ -16,62 +16,66 @@
...
-import java.util.Collections;
-import java.util.HashSet;
-import java.util.List;
-import java.util.Properties;
-import java.util.Set;
-
+import com.google.common.base.Joiner;
+import com.google.common.base.Throwables;
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
```
swaroopak: Given the recent jira for logging, let's use org.slf4j.Logger
and org.slf4j.LoggerFactory.

phoenix-core/src/it/java/org/apache/phoenix/end2end/BasePermissionsIT.java
```diff
@@ -142,37 +148,45 @@ void startNewMiniCluster(Configuration overrideConf) throws Exception{
...
-        regularUser1 = User.createUserForTesting(configuration, "regularUser1", new String[0]);
-        regularUser2 = User.createUserForTesting(configuration, "regularUser2", new String[0]);
-        regularUser3 = User.createUserForTesting(configuration, "regularUser3", new String[0]);
-        regularUser4 = User.createUserForTesting(configuration, "regularUser4", new String[0]);
-
-        groupUser = User.createUserForTesting(testUtil.getConfiguration(), "groupUser", new String[] {GROUP_SYSTEM_ACCESS});
-
-        unprivilegedUser = User.createUserForTesting(configuration, "unprivilegedUser", new String[0]);
+    @Before
+    public void setUp() {
```
swaroopak: I would move this right after the constructor for better readability.

phoenix-core/src/it/java/org/apache/phoenix/end2end/BasePermissionsIT.java
```diff
@@ -16,62 +16,66 @@
...
 
-@RunWith(Parameterized.class)
+@Category(NeedsOwnMiniClusterTest.class)
+@FixMethodOrder(MethodSorters.NAME_ASCENDING)
 public class BasePermissionsIT extends BaseTest {
 
     private static final Log LOG = LogFactory.getLog(BasePermissionsIT.class);
 
-    static String SUPERUSER;
+    static String SUPERUSER = System.getProperty("user.name");
```
swaroopak: nit: SUPER_USER

phoenix-core/src/it/java/org/apache/phoenix/end2end/BasePermissionsIT.java
swaroopak: doSetup is called even before the setup happens, is that correct? in that case ignore the comment.

phoenix-core/src/it/java/org/apache/phoenix/end2end/BasePermissionsIT.java
twdsilva: This will be fixed by @yanxinyi  in PHOENIX-5228

phoenix-core/src/it/java/org/apache/phoenix/end2end/BasePermissionsIT.java
twdsilva: doSetup is called once per class, while setup is called before each test executes. I'll rename these.

twdsilva:
@swaroopak  I uploaded a new PR based on your feedback, please review.

phoenix-core/src/it/java/org/apache/phoenix/end2end/BasePermissionsIT.java
```diff
@@ -16,62 +16,66 @@
...
-@RunWith(Parameterized.class)
-public class BasePermissionsIT extends BaseTest {
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertNotNull;
+import static org.junit.Assert.assertTrue;
+import static org.junit.Assert.fail;
+
+@Category(NeedsOwnMiniClusterTest.class)
+@FixMethodOrder(MethodSorters.NAME_ASCENDING)
```
ChinmaySKulkarni: Why do we need to fix this order of execution for tests?

ChinmaySKulkarni:
One question on the `FixMethodOrder` change. Otherwise, lgtm.

twdsilva:
The test aTestRXPermsReqdForPhoenixConn in BasePermissionIT needs to be the very first test that is run

ChinmaySKulkarni:
@twdsilva is it possible to separate out the part that needs to be run first in `aTestRXPermsReqdForPhoenixConn` into a private method called from your `BeforeClass` setup to prevent this ordering dependency?

twdsilva:
aTestRXPermsReqdForPhoenixConn tests functionality that assumes it is the very first time a client connects to cluster. I don't think I can move this to BeforeClass.

```
// NS is disabled, CQSI tries creating SYSCAT, Two cases here
// 1. First client ever --> Gets ADE, runs client server compatibility check again and gets TableNotFoundException since SYSCAT doesn't exist
// 2. Any other client --> Gets ADE, runs client server compatibility check again and gets AccessDeniedException since it doesn't have EXEC perms
            verifyDenied(getConnectionAction(), TableNotFoundException.class, regularUser1);
```

ChinmaySKulkarni:
@twdsilva I see. +1

PR #478, opened 2019-04-08
PHOENIX-5228 use slf4j for logging in phoenix project

yanxinyi:


yanxinyi:
@karanmehta93 @twdsilva 

As ppl commented on JIRA, used slf4j and renamed variable name to `LOGGER`

PR #480, opened 2019-04-09
PHOENIX-5138 - ViewIndexId sequences created after PHOENIX-5132 shoul…

gjacoby126:
…dn't collide with ones created before it


gjacoby126:
4.x-HBase-1.4 version of PHOENIX-5138, which has already been approved. Merging.

PR #482, opened 2019-04-11
PHOENIX-4925 Use Segment tree to organize Guide Post Info

BinShi-SecularBird:
@dbwong please review with this early draft. Haven't sorted out local index and cross region boundary condition when generating scan. Still adding unit test and integration test and trying to pass the tests. Please focus on main focus on main logic, data structure, main algorithms and interfaces first.


karanmehta93:
@BinShi-SecularBird can you fix the conflicts?

phoenix-core/src/main/java/org/apache/phoenix/iterate/SnapshotScanner.java
```diff
@@ -111,7 +111,11 @@ public SnapshotScanner(Configuration conf, FileSystem fs, Path rootDir,
   public Result next() throws IOException {
     values.clear();
     scanner.nextRaw(values);
-    statisticsCollector.collectStatistics(values);
+    // Needs to call collectStatistics even when values is empty.
+    // In this case, we pass hasMore with value 'false' so a new
+    // guide post will be created for the remaining cached data.
+    statisticsCollector.collectStatistics(values, ! values.isEmpty());
```
karanmehta93: Suggestion: another way of handling the last GPW is with the scanner close method as well or when trying to commit the GPW. With that approach, you won't need to explicitly add this boolean here.

phoenix-core/src/main/java/org/apache/phoenix/schema/stats/DefaultStatisticsCollector.java
```diff
@@ -276,11 +273,14 @@ private void writeStatistics(final Region region, boolean delete, List<Mutation>
...
                 if (!guidePostsInfoWriterMap.containsKey(cfKey)) {
-                    Pair<Long, GuidePostsInfoBuilder> emptyGps = new Pair<Long, GuidePostsInfoBuilder>(0l, new GuidePostsInfoBuilder());
+                    Pair<Pair<Long, Long>, GuidePostChunkBuilder> emptyGps =
+                            new Pair<Pair<Long, Long>, GuidePostChunkBuilder>(new Pair(0l, 0l), new GuidePostChunkBuilder());
                     guidePostsInfoWriterMap.put(cfKey, emptyGps);
                 }
             }
         }
+
+        Set<ImmutableBytesPtr> fams = guidePostsInfoWriterMap.keySet();
```
karanmehta93: This line has been moved from no 259 to 283, which means that `guidePostsInfoWriterMap` will potentially now have 1 new entry as compared to previous behavior. What is this needed for?

phoenix-core/src/main/java/org/apache/phoenix/schema/stats/GuidePostsInfoBuilder.java
```diff
@@ -17,124 +17,112 @@
...
     /**
      * Track a new guide post
      * @param row number of rows in the guidepost
      * @param byteCount number of bytes in the guidepost
      * @param updateTimestamp time at which guidepost was created/updated.
-     * @throws IOException
      */
-    public boolean trackGuidePost(ImmutableBytesWritable row, long byteCount, long rowCount,
-            long updateTimestamp) {
+    public boolean trackGuidePost(ImmutableBytesWritable row, long byteCount, long rowCount, long updateTimestamp) {
```
dbwong: This isn't really following the builder pattern.  consider rename or handling through exceptions or changing the caller.

phoenix-core/src/main/java/org/apache/phoenix/schema/stats/GuidePostChunk.java
```diff
@@ -0,0 +1,224 @@
...
+ * into the SYSTEM_STATS table in the future.
+ */
+public final class GuidePostChunk {
+    public static final int INVALID_GUIDEPOST_CHUNK_INDEX = -1;
+    public static final Logger logger = LoggerFactory.getLogger(GuidePostChunk.class);
+
+    /**
+     * The index of the guide post chunk in the chunk array.
+     */
+    private final int guidePostChunkIndex;
```
dbwong: This seems to be leaking the top level classes lower into this class.  The accessor is never used.  Can you briefly explain why this index is needed to be tracked in multiple levels?  InnerPointLookupResult in GuidePostsInfo even has 2 separate indexes tracked.

phoenix-core/src/main/java/org/apache/phoenix/schema/stats/DefaultStatisticsCollector.java
```diff
@@ -20,11 +20,9 @@
 import java.io.IOException;
 import java.sql.Connection;
 import java.sql.SQLException;
-import java.util.ArrayList;
-import java.util.List;
-import java.util.Map;
-import java.util.Set;
+import java.util.*;
```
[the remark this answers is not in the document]
dbwong: This occurs in several files.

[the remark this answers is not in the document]
BinShi-SecularBird: This was added automatically by IDE. I'll replace it.

phoenix-core/src/main/java/org/apache/phoenix/schema/stats/DefaultStatisticsCollector.java
BinShi-SecularBird: The behavior is the same, after moving the line from 259 to 283. The purpose of movement is to have more robust code which is independent of the concrete implementation of map.

The original logic is:
STEP 1. Get the fams key set;
STEP 2. Discover new entries which are added to the map at lines 280 ~ 281
STEP 3. The new entries added at step 2 actually will be reflected in fams key set too. For such entries, the stats deletions (and no stats insertions) will be generated.

I don't like this code -- getting fams key set first, then adding new entries to the map and the entries automatically reflected in fams key set -- this is non-determistic and depends on the concrete implementation of map, so I moved the line from 259 to 283 to make the code more solid.

phoenix-core/src/main/java/org/apache/phoenix/schema/stats/DefaultStatisticsCollector.java
```diff
@@ -67,11 +64,13 @@
 public class DefaultStatisticsCollector implements StatisticsCollector {
 
     private static final Log LOG = LogFactory.getLog(DefaultStatisticsCollector.class);
-    
-    final Map<ImmutableBytesPtr, Pair<Long, GuidePostsInfoBuilder>> guidePostsInfoWriterMap = Maps.newHashMap();
+
+    // Pair<Byte count, Row count>
```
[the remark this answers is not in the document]
BinShi-SecularBird: You're right. I should use GuidePostEstimation instead of Pair<Long, Long>. Good Suggestion!

phoenix-core/src/main/java/org/apache/phoenix/schema/stats/DefaultStatisticsCollector.java
```diff
@@ -67,11 +65,15 @@
...
+    final Map<ImmutableBytesPtr, Pair<Pair<Long, Long>, GuidePostChunkBuilder>> guidePostsInfoWriterMap = Maps.newHashMap();
+    final Set<ImmutableBytesPtr> writableGudiePosts = Sets.newHashSet();
```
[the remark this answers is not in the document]
BinShi-SecularBird: Fixed.

BinShi-SecularBird:
@dbwong , could you please review BaseResultIterators.getParallelScans()? All the required changes should be there now. BaseResultIterators.getRowKeyRanges() contains the logic of generating query key ranges for local index.

phoenix-core/src/main/java/org/apache/phoenix/iterate/BaseResultIterators.java
```diff
@@ -885,292 +867,215 @@ private static boolean clipKeyRangeBytes(RowKeySchema schema, int fieldIndex, in
...
+
+                byte[] endKey;
+                if ((j < keyRangeCount - 1) || (i == countOfRegionsToScan - 1)) {
+                    endKey = keyRange.getUpperRange();
+                } else {
+                    endKey = regionInfo.getEndKey();
                 }
-                regionIndex++;
+
+                regionScans = addNewScan(parallelScans, regionScans, newScan, endKey, crossesRegionBoundary, regionLocation);
```
dbwong: Lets discuss the naming and usage a bit of this as in ParallelScanGrouper this is "startKey".  This is used for the crossesPrefixBoundary check and indicated in the interface as a startKey and compared to the last scans startRow Prefix looking for for example the salt byte changing or region changing.  By using end key is there any cases we might miss?

phoenix-core/src/main/java/org/apache/phoenix/compile/ScanRanges.java
```diff
@@ -374,9 +375,19 @@ public Scan intersectScan(Scan scan, final byte[] originalStartKey, final byte[]
...
+
+        ImmutableBytesWritable a1 = new ImmutableBytesWritable(scanStartKey);
+        ImmutableBytesWritable b1 = new ImmutableBytesWritable(scanStopKey);
+        byte[] a1StartKey = a1.copyBytes();
+        byte[] b1StartKey = b1.copyBytes();
+        newScan.setAttribute(SCAN_ACTUAL_START_ROW, a1StartKey);
+        newScan.setStartRow(a1StartKey);
+        newScan.setStopRow(b1StartKey);
+        /*
         newScan.setAttribute(SCAN_ACTUAL_START_ROW, scanStartKey);
```
yanxinyi: why do we need to keep the old code here? 
From my point of view, `setStartRow` and `setStopRow` APIs are deprecated, and it will remove soon.

phoenix-core/src/main/java/org/apache/phoenix/iterate/BaseResultIterators.java
```diff
@@ -885,292 +867,234 @@ private static boolean clipKeyRangeBytes(RowKeySchema schema, int fieldIndex, in
...
-     * occurring for a large ingest with stats never having been calculated for the
-     * new region.
+     * merge sorted, depending on the query. Also computes an estimated bytes scanned,
+     * rows scanned, and last update time of statistics.
+     *
      * @return list of parallel scans to run for a given query.
      * @throws SQLException
      */
     private List<List<Scan>> getParallelScans(byte[] startKey, byte[] stopKey) throws SQLException {
+        ScanRanges scanRanges = context.getScanRanges();
```
yanxinyi: why we change the order here?

phoenix-core/src/main/java/org/apache/phoenix/schema/stats/GuidePostChunk.java
```diff
@@ -0,0 +1,227 @@
...
+ * where gp_x is one of guide post collected on the server side. The last guide post chunk is a DUMMY chunk
+ * which contains one DUMMY guide post.
+ *
+ * Eventually the guide post chunk should be built and compressed on the server side when collecting stats.
+ * It is now the minimal data unit for compression/decompression and will be the minimal data unit stored
+ * into the SYSTEM_STATS table in the future.
+ */
+public final class GuidePostChunk {
+    public static final int INVALID_GUIDEPOST_CHUNK_INDEX = -1;
+    public static final Logger logger = LoggerFactory.getLogger(GuidePostChunk.class);
```
yanxinyi: Let's use `LOGGER` instead of `logger` here.

phoenix-core/src/main/java/org/apache/phoenix/schema/stats/StatisticsUtil.java
```diff
@@ -47,13 +46,17 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 import com.google.common.collect.Sets;
 
 /**
  * Simple utility class for managing multiple key parts of the statistic
  */
 public class StatisticsUtil {
+    private static final Logger logger = LoggerFactory.getLogger(StatisticsUtil.class);
```
yanxinyi: please change var name to `LOGGER`

phoenix-core/src/main/java/org/apache/phoenix/schema/stats/StatisticsWriter.java
```diff
@@ -31,6 +31,8 @@
 import java.util.ArrayList;
 import java.util.List;
 
+import org.apache.commons.logging.Log;
```
yanxinyi: let's use `org.slf4j` here instead of `org.apache.commons` to make consistent.
import org.slf4j.Logger;
import org.slf4j.LoggerFactory;

phoenix-core/src/main/java/org/apache/phoenix/schema/stats/StatisticsWriter.java
```diff
@@ -74,6 +76,7 @@
  * Wrapper to access the statistics table SYSTEM.STATS using the HTable.
  */
 public class StatisticsWriter implements Closeable {
+    private static final Log LOG = LogFactory.getLog(StatisticsWriter.class);
```
yanxinyi: Same as above, use the same variable name as `LOGGER`

phoenix-core/src/main/java/org/apache/phoenix/iterate/SnapshotScanner.java
BinShi-SecularBird: Adopted the suggestion

phoenix-core/src/main/java/org/apache/phoenix/query/QueryServices.java
```diff
@@ -185,20 +185,23 @@
...
-    
     public static final String RUN_UPDATE_STATS_ASYNC = "phoenix.update.stats.command.async";
     public static final String STATS_SERVER_POOL_SIZE = "phoenix.stats.pool.size";
     public static final String COMMIT_STATS_ASYNC = "phoenix.stats.commit.async";
     // Maximum size in bytes taken up by cached table stats in the client
     public static final String STATS_MAX_CACHE_SIZE = "phoenix.stats.cache.maxSize";
     // The size of the thread pool used for refreshing cached table stats in stats client cache
     public static final String STATS_CACHE_THREAD_POOL_SIZE = "phoenix.stats.cache.threadPoolSize";
+    // The targeted size of the guide post chunk measured in the number of guide posts.
+    public static final String STATS_TARGETED_CHUNK_SIZE = "phoenix.stats.targeted.chunk.size";
```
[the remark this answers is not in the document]
BinShi-SecularBird: This is a cluster configuration applied to all queries and all tables. Isn't it     the right place to add a cluster configuration?

phoenix-core/src/main/java/org/apache/phoenix/schema/stats/GuidePostChunk.java
```diff
@@ -0,0 +1,224 @@
...
+import java.io.EOFException;
+import java.util.ArrayList;
+import java.util.List;
+
+import org.apache.phoenix.schema.stats.GuidePostChunkBuilder;
+
+/**
+ * A guide post chunk is comprised of a group of guide posts, and it has one of key ranges below:
+ *     (UNBOUND, gp_i0], (gp_i0, gp_i1], (gp_i1, gp_i2], ..., (gp_in, gp_n], (gp_n, UNBOUND)
+ * where gp_x is one of guide post collected on the server side. The last guide post chunk is a DUMMY chunk
```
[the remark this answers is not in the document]
BinShi-SecularBird: Changed name to ending chunk.

phoenix-core/src/it/java/org/apache/phoenix/end2end/ExplainPlanWithStatsEnabledIT.java
```diff
@@ -717,14 +717,164 @@ public void testSelectQueriesWithStatsForParallelizationOn() throws Exception {
         testSelectQueriesWithFilters(true);
     }
 
+    /*
```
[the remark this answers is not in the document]
BinShi-SecularBird: cleanup is done.

phoenix-core/src/it/java/org/apache/phoenix/schema/stats/BaseStatsCollectorIT.java
```diff
@@ -734,20 +734,18 @@ public void testRowCountWhenNumKVsExceedCompactionScannerThreshold() throws Exce
...
             for (String familyName : familyNames) {
                 GuidePostsInfo gps =
-                        StatisticsUtil.readStatistics(statsHTable,
+                        StatisticsUtil.readStatistics(new GuidePostsInfoBuilder(), statsHTable,
                                 new GuidePostsKey(Bytes.toBytes(tableName), Bytes.toBytes(familyName)),
                                 HConstants.LATEST_TIMESTAMP);
-                assertTrue(emptyGuidePostExpected ? gps.isEmptyGuidePost() : !gps.isEmptyGuidePost());
-                assertTrue(gps.getByteCounts()[0] >= guidePostWidth);
-                assertTrue(gps.getGuidePostTimestamps()[0] > 0);
+                assertTrue(gps.getTotalEstimation().getByteCount() >= expectedEstimatedSize);
```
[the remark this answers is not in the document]
BinShi-SecularBird: There are a few of such E2E test cases, e.g., testSomeUpdateEmptyStats in BaseStatsCollectorIT.java

phoenix-core/src/main/java/org/apache/phoenix/schema/stats/GuidePostsInfo.java
```diff
@@ -17,138 +17,771 @@
...
-                + SizedUtil.ARRAY_SIZE + this.gpTimestamps.length * SizedUtil.LONG_SIZE // gpTimestamps
-                + SizedUtil.INT_SIZE; // estimatedSize
-        this.estimatedSize = estimatedSize;
-    }
-    
-    public ImmutableBytesWritable getGuidePosts() {
-        return guidePosts;
+        this.comparator = ScanUtil.getComparator(true, SortOrder.ASC);
+
+        if (this.guidePostChunks != null && this.guidePostChunks.size() > 0) {
```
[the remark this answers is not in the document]
BinShi-SecularBird: Resolved as suggested.

phoenix-core/src/main/java/org/apache/phoenix/schema/stats/GuidePostsInfo.java
```diff
@@ -17,138 +17,771 @@
...
+            int byteCount = 0;
+            for (GuidePostChunk chunk : this.guidePostChunks) {
+                byteCount += chunk.getEstimatedSize();
+            }
+            this.estimatedSize = byteCount;
+
+            Construct(this.guidePostChunks, 0, n - 1, 0);
+        } else {
+            this.treeSize = 0;
+            this.nodes = null;
```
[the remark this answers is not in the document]
BinShi-SecularBird: I don't need to check whether nodes is null and there is no nested check for null object, so use Optional class is overkilled here.

phoenix-core/src/main/java/org/apache/phoenix/schema/stats/GuidePostsInfoBuilder.java
BinShi-SecularBird: This intend not to use builder pattern here. No need to use builder pattern.

phoenix-core/src/main/java/org/apache/phoenix/schema/stats/GuidePostChunk.java
BinShi-SecularBird: It's used in DecodedGuidePostCache to track the guide post chunks in the cache.

phoenix-core/src/main/java/org/apache/phoenix/schema/stats/GuidePostChunk.java
BinShi-SecularBird: Resolved

phoenix-core/src/main/java/org/apache/phoenix/schema/stats/StatisticsUtil.java
BinShi-SecularBird: Resolved

phoenix-core/src/main/java/org/apache/phoenix/compile/ScanRanges.java
BinShi-SecularBird: I forgot to delete the old code. Using setStartRow and setStopRow are original code which aren't supposed to change in this PR.

BinShi-SecularBird: Resolved.

phoenix-core/src/main/java/org/apache/phoenix/iterate/BaseResultIterators.java
BinShi-SecularBird: The whole function was actually rewritten.

phoenix-core/src/main/java/org/apache/phoenix/schema/stats/StatisticsWriter.java
BinShi-SecularBird: Resolved

phoenix-core/src/main/java/org/apache/phoenix/schema/stats/StatisticsWriter.java
BinShi-SecularBird: Resolved

phoenix-core/src/it/java/org/apache/phoenix/schema/stats/BaseStatsCollectorIT.java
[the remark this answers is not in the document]
BinShi-SecularBird: Changed to assertEquals

phoenix-core/src/main/java/org/apache/phoenix/iterate/BaseResultIterators.java
```diff
@@ -885,292 +867,215 @@ private static boolean clipKeyRangeBytes(RowKeySchema schema, int fieldIndex, in
...
-                     * currentKey back to what it was at the beginning of the loop.
-                     */
-                    currentKeyBytes = initialKeyBytes;
+            // Make a scan for every range in regionScanKeyRanges
+            for (int j = 0; j < keyRangeCount; j++) {
+                KeyRange keyRange = regionScanKeyRanges.get(j);
+                boolean crossesRegionBoundary = false;
+
+                if (j == keyRangeCount - 1) {
+                    crossesRegionBoundary = true;
```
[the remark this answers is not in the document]
BinShi-SecularBird: I tried to keep the same semantics as the old code. See line 1102 "Scan newScan = scanRanges.intersectScan(scan, currentKeyBytes, endKey, keyOffset, true);" in  BaseResultIterators.getParallelScans() in original code.

phoenix-core/src/main/java/org/apache/phoenix/iterate/BaseResultIterators.java
```diff
@@ -885,292 +867,215 @@ private static boolean clipKeyRangeBytes(RowKeySchema schema, int fieldIndex, in
...
-                    if (!gpsInThisRegion && delayAddingEst) {
-                        updateEstimates(gps, guideIndex-1, estimates);
-                        gpsInThisRegion = true;
-                        delayAddingEst = false;
-                    }
-                } else if (!gpsInThisRegion) {
-                    delayAddingEst = false;
+                            regionInfo.getEndKey(), newScan.getStartRow(), newScan.getStopRow());
+                } else {
+                    logger.warn("Didn't generate scan for " + keyRange.toString());
```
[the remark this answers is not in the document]
BinShi-SecularBird: Good catch! Fixed.

phoenix-core/src/main/java/org/apache/phoenix/iterate/BaseResultIterators.java
BinShi-SecularBird: Let double check this. This is exactly one of the places where we need to pay attention to. Hopefully I'll add test cases to cover the scenario.

phoenix-core/src/main/java/org/apache/phoenix/query/QueryServices.java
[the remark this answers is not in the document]
dbwong: Mmm what about the case where the chunk size here is > the table guide post size for example will you combine chunks?

phoenix-core/src/main/java/org/apache/phoenix/compile/ScanRanges.java
```diff
@@ -342,8 +343,10 @@ public Scan intersectScan(Scan scan, final byte[] originalStartKey, final byte[]
             // TODO: it seems that our SkipScanFilter or HBase runs into problems if we don't
             // have an enclosing range when we do a point lookup.
             if (isPointLookup) {
-                scanStartKey = ScanUtil.getMinKey(schema, newSkipScanFilter.getSlots(), slotSpan);
-                scanStopKey = ScanUtil.getMaxKey(schema, newSkipScanFilter.getSlots(), slotSpan);
+                scanStartKey = ScanUtil.getMinKey(
+                        newSkipScanFilter.getSchema(), newSkipScanFilter.getSlots(), newSkipScanFilter.getSlotSpan());
```
yanxinyi: Why we made this change? Whats' the diff btw 'RowKeySchema' from the `ScanRangers` and `SkipScanFilter.getSchema()`?

phoenix-core/src/main/java/org/apache/phoenix/query/DefaultGuidePostsCacheFactory.java
```diff
@@ -29,11 +28,15 @@ public PhoenixStatsLoader getPhoenixStatsLoader(ConnectionQueryServices querySer
         Preconditions.checkNotNull(config);
 
         final boolean isStatsEnabled = config.getBoolean(STATS_COLLECTION_ENABLED, DEFAULT_STATS_COLLECTION_ENABLED);
-        if (queryServices == null || !isStatsEnabled) {
+        if (queryServices == null || ! isStatsEnabled) {
```
yanxinyi: extra space?

phoenix-core/src/main/java/org/apache/phoenix/query/QueryServices.java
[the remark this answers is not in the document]
yanxinyi: if we are doing it on the table level, we have to add an extra column to store this chunk size. Adding more and more columns on the syscat is not a good idea, what's the best practice here @dbwong ?

phoenix-core/src/main/java/org/apache/phoenix/schema/stats/GuidePostsInfo.java
```diff
@@ -17,138 +17,771 @@
...
-     *            Prefix byte encoded guidePosts
-     * @param rowCounts
-     *            The rowCounts of each guidePost traversed
-     * @param maxLength
-     *            Maximum length of a guidePost collected
-     * @param guidePostsCount
-     *            Number of guidePosts
-     * @param gpTimestamps
-     *            Times at which guidePosts were updated/created
+     * Construct the Variant Segment Tree (https://salesforce.quip.com/taWiALFmhquO)
```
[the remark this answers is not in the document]
yanxinyi: do you mean Jira link?

phoenix-core/src/it/java/org/apache/phoenix/end2end/ExplainPlanWithMultipleRowsPerGuidePostIT.java
```diff
@@ -0,0 +1,303 @@
...
+            assertEquals((Long) (estimatedRows * estimatedBytesPerRow), info.getEstimatedBytes());
+            assertTrue(info.getEstimateInfoTs() > 0);
+
+            // Query whose end key is before any data. Query Key Range (UNBOUND, 98].
+            sql = "SELECT a FROM " + tableName + " WHERE K <= 98";
+            rs = conn.createStatement().executeQuery(sql);
+            assertFalse(rs.next());
+            // Hit the first guide post.
+            info = getByteRowEstimates(conn, sql, binds);
+            estimatedRows = 10;
```
yanxinyi: We didn't insert rows before 100, why we are getting 10 here?

phoenix-core/src/main/java/org/apache/phoenix/compile/ScanRanges.java
BinShi-SecularBird: When the PK column type is varchar, the column is translated to multiple char[1] columns in SkipScanFilter's schema, and the SkipScanFilter's schema is supposed to use here, otherwise it will cause the test failures in SkipScanFilterIntersectTest.java.

phoenix-core/src/main/java/org/apache/phoenix/query/QueryServices.java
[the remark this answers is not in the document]
BinShi-SecularBird: Chunk size is max(1, count of guide posts).

phoenix-core/src/main/java/org/apache/phoenix/query/DefaultGuidePostsCacheFactory.java
BinShi-SecularBird: I always add an extra space between ! and the variable for better readness. I saw the other places use this coding style too.

phoenix-core/src/main/java/org/apache/phoenix/schema/stats/GuidePostsInfo.java
```diff
@@ -17,138 +17,913 @@
...
+    public int getGuidePostChunkIndex() {
+        return guidePostChunkIndex;
+    }
+
+    public List<byte[]> getGuidePosts() {
+        return guidePosts;
+    }
+
+    @Override
+    public boolean equals(Object o) {
```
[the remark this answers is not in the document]
BinShi-SecularBird: It seems that it's not necessary. I can see my equals function is called for comparison.

phoenix-core/src/it/java/org/apache/phoenix/end2end/ExplainPlanWithMultipleRowsPerGuidePostIT.java
BinShi-SecularBird: Because the query will hit the first guide post which has key 100. This guide post contains 10 estimated rows.

phoenix-core/src/it/java/org/apache/phoenix/end2end/ExplainPlanWithMultipleRowsPerGuidePostIT.java
```diff
@@ -0,0 +1,303 @@
...
+        //                (UNBOUND, 219]                           (219, 339]                              (339, 459]                      \
+        //               /               \                      /               \                        /             \                    \
+        //              /                 \                    /                 \                      /               \                    \
+        //        (UNBOUND, 179]           \               (219, 299]             \                (339, 419]            \              (459, UNBOUND)
+        //         /           \            \            /            \            \             /            \           \             /             \
+        // (UNBOUND, 139]   (139 179]   (179, 219]   (219, 259]   (259, 299]   (299, 339]   (339, 379]   (379, 419]   (419, 459]   (459, 499]   (499, UNBOUND)
+        List<Object> binds = Lists.newArrayList();
+        try (Connection conn = DriverManager.getConnection(getUrl())) {
+            int[] result;
+            Long estimatedBytesPerRow = 100L;
```
yanxinyi: Maybe add explain here? 
such as 20 * 3 fixed bytes from hbase.
Col 1:  4 (ROW) + 1 (CF) + 4 (CQ) + 1 (value) = 10
Col 2:  4 (ROW) + 1 (CF) + 2 (CQ) + 8 (value) = 15
Col 2:  4 (ROW) + 1 (CF) + 2 (CQ) + 8 (value) = 15

phoenix-core/src/main/java/org/apache/phoenix/schema/stats/GuidePostsInfo.java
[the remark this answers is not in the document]
yanxinyi: https://stackoverflow.com/questions/30026097/equals-without-hashcode

ppl believe without hashcode function might result in weird bugs.

phoenix-core/src/main/java/org/apache/phoenix/schema/stats/GuidePostsInfo.java
```diff
@@ -17,138 +17,771 @@
...
-                + SizedUtil.ARRAY_SIZE + this.rowCounts.length * SizedUtil.LONG_SIZE // rowCounts
-                + SizedUtil.ARRAY_SIZE + this.byteCounts.length * SizedUtil.LONG_SIZE // byteCounts
-                + SizedUtil.ARRAY_SIZE + this.gpTimestamps.length * SizedUtil.LONG_SIZE // gpTimestamps
-                + SizedUtil.INT_SIZE; // estimatedSize
-        this.estimatedSize = estimatedSize;
-    }
-    
-    public ImmutableBytesWritable getGuidePosts() {
-        return guidePosts;
+        this.comparator = ScanUtil.getComparator(true, SortOrder.ASC);
```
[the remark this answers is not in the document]
BinShi-SecularBird: Resolved

phoenix-core/src/main/java/org/apache/phoenix/schema/stats/GuidePostsInfo.java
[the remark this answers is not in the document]
BinShi-SecularBird: I added hashCode function, although the bug exists when they are used for hash containers and these objects shouldn't be used for hash containers.

phoenix-core/src/main/java/org/apache/phoenix/schema/stats/GuidePostEstimation.java
```diff
@@ -0,0 +1,89 @@
+package org.apache.phoenix.schema.stats;
```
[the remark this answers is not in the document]
BinShi-SecularBird: Resolved

phoenix-core/src/main/java/org/apache/phoenix/schema/stats/GuidePostEstimation.java
```diff
@@ -0,0 +1,89 @@
...
+    }
+
+    /**
+     * Merge the two guide post estimation objects into one which contains the sum of rows,
+     * the sum of bytes and the least update time stamp of the two objects.
+     * @param left
+     * @param right
+     * @return the new guide post estimation object which contains the "Sum" info.
+     */
+    public static GuidePostEstimation merge(GuidePostEstimation left, GuidePostEstimation right) {
```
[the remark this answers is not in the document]
BinShi-SecularBird: resolved

phoenix-core/src/main/java/org/apache/phoenix/schema/stats/GuidePostEstimation.java
```diff
@@ -0,0 +1,89 @@
+package org.apache.phoenix.schema.stats;
+
+import org.apache.phoenix.util.SizedUtil;
+
+public final class GuidePostEstimation {
```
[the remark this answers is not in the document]
BinShi-SecularBird: resolved

phoenix-core/src/main/java/org/apache/phoenix/schema/stats/GuidePostChunkBuilder.java
```diff
@@ -0,0 +1,170 @@
...
+ * The guide post boundaries:
+ *     gp_0, gp_1, ..., gp_i0, ..., gp_i1, ..., gp_i2, ..., gp_in, ..., gp_n
+ * The guide post chunk boundaries:
+ *     gp_i0, gp_i1, ..., gp_in, ..., gp_n
+ * The key space split by the guide post chunks:
+ *     (UNBOUND, gp_i0](gp_i0, gp_i1](gp_i1, gp_i2]...(gp_in, gp_n](gp_n, UNBOUND)
+ * The last guide post chunk is a DUMMY chunk which contains one DUMMY guide post
+ * KeyRang.UNBOUND with 0 estimated rows/bytes and Long.MAX as last update timestamp.
+ */
+public final class GuidePostChunkBuilder {
```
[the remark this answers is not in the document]
BinShi-SecularBird: resolved

PR #483, opened 2019-04-15
PHEONIX-5243: PhoenixResultSet#next() closes the result set if scanner returns null

swaroopak:
@twdsilva @ChinmaySKulkarni Could you please review this?


ChinmaySKulkarni:
lgtm

PR #484, opened 2019-04-16
PHOENIX-5239: Send persistent subquery caches to all regionservers

JohnNPhillips:
Context on ticket: https://issues.apache.org/jira/browse/PHOENIX-5239

### DIFF
--- a/phoenix-core/src/main/java/org/apache/phoenix/cache/ServerCacheClient.java
+++ b/phoenix-core/src/main/java/org/apache/phoenix/cache/ServerCacheClient.java
@@ -248,18 +248,19 @@ public class ServerCacheClient {
             boolean storeCacheOnClient) throws SQLException {
         final byte[] cacheId = ServerCacheClient.generateId();
         return addServerCache(keyRanges, cacheId, cachePtr, txState, cacheFactory,
-                cacheUsingTable, false, storeCacheOnClient);
+                cacheUsingTable, false, false, storeCacheOnClient);
     }
 
     public ServerCache addServerCache(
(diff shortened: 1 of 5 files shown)


phoenix-core/src/main/java/org/apache/phoenix/cache/ServerCacheClient.java
```diff
@@ -310,7 +311,7 @@ public TaskExecutionMetricsHolder getTaskExecutionMetric() {
                         }
                     }));
                 } else {
-                    if (LOG.isDebugEnabled()) {LOG.debug(addCustomAnnotations("NOT adding cache entry to be sent for " + entry + " since one already exists for that entry", connection));}
```
JohnNPhillips: This generates a log line for every single region (not just unique regionservers) that aren't sent a cache. For large tables, this makes `debug` level logs _very_ noisy, even for testing. However, if phoenix tries to avoid making backwards incompatible changes to debug level logs, I can remove this from the PR

phoenix-core/src/main/java/org/apache/phoenix/cache/ServerCacheClient.java
```diff
@@ -349,7 +350,7 @@ public TaskExecutionMetricsHolder getTaskExecutionMetric() {
                 }
             }
         }
-        if (LOG.isDebugEnabled()) {LOG.debug(addCustomAnnotations("Cache " + cacheId + " successfully added to servers.", connection));}
```
JohnNPhillips: `cacheId` is a byte array which as-is generates useless logs like `Cache [B@dcf3e99 successfully added to servers`

PR #489, opened 2019-04-20
PHOENIX-5252 Add job priority option to UpdateStatisticsTool

yanxinyi:


karanmehta93:
LGTM overall, @yanxinyi Can you provide a brief description on the Jira as to how this option works internally so that the users understand exactly how to use it?

phoenix-core/src/main/java/org/apache/phoenix/schema/stats/UpdateStatisticsTool.java
```diff
@@ -164,12 +168,36 @@ void parseArgs(String[] args) {
...
+             return JobPriority.DEFAULT;
+         }
+
+         switch (jobPriorityOption) {
+             case "0" : return JobPriority.VERY_HIGH;
+             case "1" : return JobPriority.HIGH;
+             case "2" : return JobPriority.NORMAL;
+             case "3" : return JobPriority.LOW;
+             case "4" : return JobPriority.VERY_LOW;
+             case "5" : return JobPriority.DEFAULT;
```
ChinmaySKulkarni: What is the `JobPriority.DEFAULT` level? Perhaps this should be the default case?

yanxinyi: https://issues.apache.org/jira/browse/MAPREDUCE-5870

Yes, if priority is not indicated by the user, it sets to default.

yanxinyi: maybe I should change to normal or default instead UNDEFINED_PRIORITY

PR #490, opened 2019-04-22
PHOENIX-5251: Avoid taking explicit lock by using AtomicReference in …

swaroopak:
…PhoenixAccessController class


swaroopak:
Thank you @gjacoby126 for the review :)

gjacoby126:
@swaroopak - could you please check to see if your patch works with master? If so, I'll commit to master and the 4.x branches. If not could you please create a forward-port patch?

PR #491, opened 2019-04-24
PHOENIX-5168 IndexScrutinyTool to output to Table when that option is…

gokceni:
… given

### DIFF
--- a/phoenix-core/src/it/java/org/apache/phoenix/end2end/IndexScrutinyToolIT.java
+++ b/phoenix-core/src/it/java/org/apache/phoenix/end2end/IndexScrutinyToolIT.java
@@ -10,6 +10,7 @@
  */
 package org.apache.phoenix.end2end;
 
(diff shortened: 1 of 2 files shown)


gokceni:
@gjacoby126 @kadirozde @swaroopak

phoenix-core/src/it/java/org/apache/phoenix/end2end/IndexScrutinyToolIT.java
```diff
@@ -172,6 +173,14 @@
         protected long getCounterValue(Counters counters, Enum<PhoenixScrutinyJobCounters> counter) {
             return counters.findCounter(counter).getValue();
         }
+
+        protected int countRows(Connection conn, String tableFullName) throws SQLException {
+            ResultSet count = conn.createStatement().executeQuery("select count(*) from " + tableFullName);
```
swaroopak: Curious: what happens if the count is way too large?

gokceni: This is integration test. It will not be large.

PR #498, opened 2019-05-02
PHOENIX-5258 Add support for parsing CSV header as columns

prashantvithani:


phoenix-core/src/it/java/org/apache/phoenix/end2end/CsvBulkLoadToolIT.java
```diff
@@ -530,4 +530,91 @@ public void testIgnoreCsvHeader() throws Exception {
...
+                    "--zookeeper", zkQuorum});
+            assertEquals(0, exitCode);
+
+            try (ResultSet rs = stmt.executeQuery("SELECT COUNT(1) FROM S.TABLE14")) {
+                assertTrue(rs.next());
+                assertEquals(3, rs.getInt(1));
+                assertFalse(rs.next());
+            }
+
+            try {
```
swaroopak: Is this failure scenario similar to previous? just that we are expecting an exception here?

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/CsvBulkLoadTool.java
```diff
@@ -96,6 +118,86 @@ protected void configureOptions(CommandLine cmdLine, List<ColumnInfo> importColu
...
+
+    /**
+     * Fetch CSV header (first line) from given HDFS path
+     * @param conf Configured Options
+     * @param path HDFS Path to single input file
+     * @return The header line (first line) from the input file
+     * @throws IOException Exception thrown by FileSystem IO
+     */
+    private String fetchCsvHeader(Configuration conf, String path) throws IOException {
+        FileSystem fs = FileSystem.get(URI.create(path), conf);
```
swaroopak: Can we use try-with-resources here?

phoenix-core/src/it/java/org/apache/phoenix/end2end/CsvBulkLoadToolIT.java
prashantvithani: The previous scenario is when the `--header` argument is not passed. The total number of columns is 4, and the input files contain lesser than that. So, the scenario expects the upsert executor to exit abnormally.

In here, the `--header` is passed but the two input files have different headers, which is not supported and expecting an exception.

PR #499, opened 2019-05-06
Phoenix-5272 ALTER INDEX REBUILD ALL 

gokceni:


phoenix-core/src/it/java/org/apache/phoenix/end2end/index/IndexMetadataIT.java
```diff
@@ -610,7 +613,84 @@ public void testAsyncRebuildTimestamp() throws Exception {
...
+        val = getIndexValue(conn, indexName, 2);
+        assertEquals("val1", val);
+
+        // Add extra row to Index
+        upsertStmt = conn.prepareStatement("UPSERT INTO " + indexName + " VALUES(?, ?)");
+        upsertStmt.setString(1, "key3");
+        upsertStmt.setString(2, "val3");
+        upsertStmt.execute();
+
+        //conn.createStatement().execute("ALTER TABLE "+ testTable +" ADD V41 INTEGER");
```
gjacoby126: nit: Can remove commented out line.

phoenix-core/src/it/java/org/apache/phoenix/end2end/index/IndexMetadataIT.java
```diff
@@ -610,7 +613,84 @@ public void testAsyncRebuildTimestamp() throws Exception {
...
+        conn.createStatement().execute("ALTER INDEX "+indexName+" ON " + testTable +" REBUILD ALL ASYNC");
+
+        Table systemHTable= conn.unwrap(PhoenixConnection.class).getQueryServices()
```
gjacoby126: Rather than counting rows in the task table (which could be brittle, especially if we want to run tests in parallel) could we check for the actual task?

phoenix-core/src/main/java/org/apache/phoenix/parse/AlterIndexStatement.java
```diff
@@ -65,6 +67,10 @@ public boolean isAsync() {
         return async;
     }
 
+    public boolean isAll() {
```
gjacoby126: Name could be a little confusing. Maybe "isRebuildAll", with a similar rename for the variable?

phoenix-core/src/main/java/org/apache/phoenix/schema/MetaDataClient.java
```diff
@@ -4362,18 +4370,40 @@ public MutationState alterIndex(AlterIndexStatement statement) throws SQLExcepti
...
-                            tableUpsert.setString(2, schemaName);
-                            tableUpsert.setString(3, indexName);
-                            tableUpsert.setLong(4, result.getTable().getTimeStamp());
-                            tableUpsert.execute();
-                            connection.commit();
-                        } finally {
-                            if (tableUpsert != null) {
-                                tableUpsert.close();
+                        if (isAll) {
+                            List<Task.TaskRecord> tasks = Task.queryTaskTable(connection, schemaName, tableName, PTable.TaskType.INDEX_REBUILD,
```
gjacoby126: @vincentpoon, @mihir6692 , @twdsilva , we're trying to move manually triggered index rebuilds (ALTER INDEX REBUILD) to use System.Tasks rather than the index disable timestamp + PhoenixMRJobSubmitter. Is anything else you know of depending on the syscat index timestamp being set when doing a manual rebuild?

phoenix-core/src/main/java/org/apache/phoenix/schema/MetaDataClient.java
```diff
@@ -4362,18 +4370,40 @@ public MutationState alterIndex(AlterIndexStatement statement) throws SQLExcepti
...
+                                            dataTableName, PTable.TaskStatus.CREATED.toString(),
+                                            jsonObject.toString(), null, ts, null, true);
+                                    connection.commit();
+                                } catch (IOException e) {
+                                    throw new SQLException("Exception happened while adding a System.Task" + e.toString());
+                                }
+                            }
+                        } else {
+                            try {
+                                tableUpsert = connection.prepareStatement(UPDATE_INDEX_REBUILD_ASYNC_STATE);
```
gjacoby126: Would be good to be able to eliminate this else block and just push everything through Task, just with the rebuild all flag set or not set. However, we need to verify that this doesn't cause unexpected side effects or regressions. Started a thread above about that.

phoenix-core/src/main/antlr3/PhoenixSQL.g
```diff
@@ -606,8 +606,8 @@ drop_index_node returns [DropIndexStatement ret]
 // Parse a alter index statement
 alter_index_node returns [AlterIndexStatement ret]
     : ALTER INDEX (IF ex=EXISTS)? i=index_name ON t=from_table_name
-      ((s=(USABLE | UNUSABLE | REBUILD | DISABLE | ACTIVE)) (async=ASYNC)? ((SET?)p=fam_properties)?)
-      {ret = factory.alterIndex(factory.namedTable(null, TableName.create(t.getSchemaName(), i.getName())), t.getTableName(), ex!=null, PIndexState.valueOf(SchemaUtil.normalizeIdentifier(s.getText())), async!=null, p); }
+      ((s=(USABLE | UNUSABLE | REBUILD | DISABLE | ACTIVE)) (all=ALL)? (async=ASYNC)? ((SET?)p=fam_properties)?)
```
swaroopak: Grammar seems misleading. We are expecting ALL only after REBUILD keyword. Shouldn't it be a part of REBUILD or condition, as we don't support ALL on any other keyword like USABLE | DISABLE etc?

phoenix-core/src/test/java/org/apache/phoenix/util/TestUtil.java
```diff
@@ -934,7 +934,30 @@ public IndexStateCheck(PIndexState indexState, Long indexDisableTimestamp, Boole
...
+            ResultSet rs = conn.createStatement().executeQuery(query);
+            PIndexState actualIndexState = null;
+            if (rs.next()) {
+                actualIndexState = PIndexState.fromSerializedValue(rs.getString(1));
+                boolean matchesExpected = (actualIndexState == expectedIndexState);
+                if (matchesExpected) {
+                    return;
+                }
+            }
+        } while (++nTries < maxTries);
```
swaroopak: Do you wanna utilize exponential backoff? Maybe worth looking at hbase's retryConfig utility.

phoenix-core/src/it/java/org/apache/phoenix/end2end/index/IndexMetadataIT.java
```diff
@@ -610,7 +613,84 @@ public void testAsyncRebuildTimestamp() throws Exception {
...
-    
+
+    @Test
+    public void testAsyncRebuildAll() throws Exception {
+        Properties props = PropertiesUtil.deepCopy(TEST_PROPERTIES);
+        Connection conn = DriverManager.getConnection(getUrl(), props);
+        conn.setAutoCommit(true);
+        String testTable = generateUniqueName();
+
+
```
swaroopak: nit: extra line

phoenix-core/src/it/java/org/apache/phoenix/end2end/index/IndexMetadataIT.java
```diff
@@ -610,7 +613,84 @@ public void testAsyncRebuildTimestamp() throws Exception {
                 "order by table_name" );
         assertFalse(rs.next());
     }
-    
+
+    @Test
+    public void testAsyncRebuildAll() throws Exception {
+        Properties props = PropertiesUtil.deepCopy(TEST_PROPERTIES);
+        Connection conn = DriverManager.getConnection(getUrl(), props);
```
swaroopak: I would suggest using try-with-resources for connection and prepareStatement objects.

phoenix-core/src/it/java/org/apache/phoenix/end2end/index/IndexMetadataIT.java
```diff
@@ -610,7 +613,84 @@ public void testAsyncRebuildTimestamp() throws Exception {
...
+        String indexName = "R_ASYNCIND_" + generateUniqueName();
+        ddl = "CREATE INDEX " + indexName + " ON " + testTable  + " (k, v4)";
+        stmt.execute(ddl);
+
+        // Check that index value is same as table
+        String val = getIndexValue(conn, indexName, 2);
+        assertEquals("val1", val);
+
+        // Update index value, check that index value is still not updated
+        conn.createStatement().execute("ALTER INDEX "+indexName+" ON " + testTable +" DISABLE");
```
priyankporwal: Nit: inconsistent space between addends

phoenix-core/src/main/java/org/apache/phoenix/parse/AlterIndexStatement.java
gokceni: I can do that. But the reason I picked this name is; this class maps directly to Sql statement. The name of the keyword is ALL

phoenix-core/src/main/java/org/apache/phoenix/schema/MetaDataClient.java
gokceni: I thought about that. It will be a huge change with unknown consequences. I think it is a bit risky to move everything without scoping. We already have a Jira for that (PHOENIX-5212). I don't want to completely change how we build indexes with the scope of this jira.

phoenix-core/src/main/antlr3/PhoenixSQL.g
gokceni: ALL is similar to ASYNC and follows the same pattern.

phoenix-core/src/test/java/org/apache/phoenix/util/TestUtil.java
gokceni: I am not sure if exponential backoff is necessary here. It doesn't take a lot of time to wait till indexes are build in the tests.

swaroopak: I see there are 60 maxRetries. This might end up using resources unnecessarily. It's better to utilize an existing retry framework rather than reinventing the wheel. RetryConfig gives an option to specify the policy of the retry mechanism. What do you suggest @gjacoby126?

phoenix-core/src/main/antlr3/PhoenixSQL.g
swaroopak: I am not sure if they should be handled in the same way? @karanmehta93 thoughts?

phoenix-core/src/main/java/org/apache/phoenix/parse/AlterIndexStatement.java
twdsilva: I agree isAll is confusing, please rename it to isRebuildAll. Currently it only applies to rebuild.

phoenix-core/src/main/java/org/apache/phoenix/schema/MetaDataClient.java
twdsilva: Are you asking if the ASYNC_REBUILD_TIMESTAMP is used for anything else? I don't think so, @vincentpoon  might know for sure.

phoenix-core/src/test/java/org/apache/phoenix/util/TestUtil.java
```diff
@@ -934,7 +936,54 @@ public IndexStateCheck(PIndexState indexState, Long indexDisableTimestamp, Boole
...
+                boolean matchesExpected = (actualIndexState == expectedIndexState);
+                if (matchesExpected) {
+                    return;
+                }
+            }
+            retrier.sleepUntilNextRetry();
+        }
+        fail("Ran out of time waiting for index state to become " + expectedIndexState);
+
+        //        do {
```
twdsilva: remove commented code

phoenix-core/src/main/java/org/apache/phoenix/jdbc/PhoenixStatement.java
```diff
@@ -1248,8 +1248,8 @@ public MutationState execute() throws SQLException {
 
     private static class ExecutableAlterIndexStatement extends AlterIndexStatement implements CompilableStatement {
 
-        public ExecutableAlterIndexStatement(NamedTableNode indexTableNode, String dataTableName, boolean ifExists, PIndexState state, boolean async, ListMultimap<String,Pair<String,Object>> props) {
-            super(indexTableNode, dataTableName, ifExists, state, async, props);
+        public ExecutableAlterIndexStatement(NamedTableNode indexTableNode, String dataTableName, boolean ifExists, PIndexState state, boolean isAll, boolean async, ListMultimap<String,Pair<String,Object>> props) {
+            super(indexTableNode, dataTableName, ifExists, state, isAll, async, props);
```
priyankporwal: Let's keep the param name in-sync between the jdbc classes and the parser classes - isAll here and isRebuildAll below. Let's pick one. I'd suggest 'rebuildAll' (to keep it consistent everywhere and inline with async without [is]).

phoenix-core/src/main/java/org/apache/phoenix/parse/AlterIndexStatement.java
```diff
@@ -28,19 +28,21 @@
...
     private final PIndexState indexState;
     private boolean async;
+    private boolean isRebuildAll;
     private ListMultimap<String,Pair<String,Object>> props;
     private static final PTableType tableType=PTableType.INDEX;
 
-    public AlterIndexStatement(NamedTableNode indexTableNode, String dataTableName, boolean ifExists, PIndexState indexState, boolean async) {
-        this(indexTableNode,dataTableName,ifExists,indexState,async,null);
+    public AlterIndexStatement(NamedTableNode indexTableNode, String dataTableName, boolean ifExists, PIndexState indexState, boolean isRebuildAll, boolean async) {
+        this(indexTableNode,dataTableName,ifExists,indexState, isRebuildAll, async,null);
```
priyankporwal: Note: earlier comment about consistent naming of this param

phoenix-core/src/main/java/org/apache/phoenix/parse/AlterIndexStatement.java
```diff
@@ -65,6 +67,10 @@ public boolean isAsync() {
         return async;
     }
 
+    public boolean IsRebuildAll() {
```
priyankporwal: Nit: Function names are better started with verbs and Rebuild fits fine. Suggest rename to RebuildAll() .. and make the member variable rebuildAll

phoenix-core/src/main/java/org/apache/phoenix/schema/MetaDataClient.java
priyankporwal: Good idea @gjacoby126! But, I'd agree with @gokceni to not do it here. Perhaps another Jira or another PR.

phoenix-core/src/main/java/org/apache/phoenix/schema/MetaDataClient.java
```diff
@@ -4362,18 +4365,40 @@ public MutationState alterIndex(AlterIndexStatement statement) throws SQLExcepti
...
-                            tableUpsert.setString(2, schemaName);
-                            tableUpsert.setString(3, indexName);
-                            tableUpsert.setLong(4, result.getTable().getTimeStamp());
-                            tableUpsert.execute();
-                            connection.commit();
-                        } finally {
-                            if (tableUpsert != null) {
-                                tableUpsert.close();
+                        if (isRebuildAll) {
+                            List<Task.TaskRecord> tasks = Task.queryTaskTable(connection, schemaName, tableName, PTable.TaskType.INDEX_REBUILD,
```
priyankporwal: @gokceni Is tableName == data-table-name or index-table-name? If the former, how do we handle concurrent rebuilding of multiple indexes? I noticed that the check on line 4371 prevents multiple tasks being added to the Tasks table.

phoenix-core/src/main/java/org/apache/phoenix/jdbc/PhoenixStatement.java
gokceni: This is auto generated. I don't think I can get rid of 'is' part

gokceni: Also, I prefer isRebuildAll because this is just returning the boolean parameter. However, rebuildAll suggests a function name where you actually rebuild the index. It is harder to read the code that way.

phoenix-core/src/main/java/org/apache/phoenix/parse/AlterIndexStatement.java
gokceni: RebuildAll is misleading. It seems like it will do the actual rebuild. I will go with isRebuildAll

phoenix-core/src/main/java/org/apache/phoenix/schema/MetaDataClient.java
gokceni: It seems like it is data table name. This is the ALTER INDEX statement though. We are changing 1 index and rebuilding it. The task table check makes sure that we don't already have this in the task table (index can be built from the gus case, we don't want to kick one from gus and run Alter with rebuild too) The task table record will get deleted in 10 days (TTL).

phoenix-core/src/it/java/org/apache/phoenix/end2end/index/IndexMetadataIT.java
```diff
@@ -610,7 +615,86 @@ public void testAsyncRebuildTimestamp() throws Exception {
...
+            upsertStmt.execute();
+
+            conn.createStatement().execute(
+                    "ALTER INDEX " + indexName + " ON " + testTable + " REBUILD ALL ASYNC");
+
+            String
+                    queryTaskTable =
+                    "SELECT * FROM " + SchemaUtil.getTableName(SYSTEM_CATALOG_SCHEMA,
+                            PhoenixDatabaseMetaData.SYSTEM_TASK_TABLE);
+            ResultSet rs = conn.createStatement().executeQuery(queryTaskTable);
```
gjacoby126: Is the task table guaranteed to be empty except from this function's task? If not, might need to truncate or tweak the query to search for your particular task.

phoenix-core/src/main/java/org/apache/phoenix/schema/MetaDataClient.java
gjacoby126: @priyankporwal @gokceni - Agree that it can be another JIRA, such as PHOENIX-5212.

phoenix-core/src/test/java/org/apache/phoenix/util/TestUtil.java
```diff
@@ -934,7 +936,35 @@ public IndexStateCheck(PIndexState indexState, Long indexDisableTimestamp, Boole
     		this.success = success;
     	}
     }
-    
+
+    public static void waitForIndexState(Connection conn, String fullIndexName, PIndexState expectedIndexState) throws InterruptedException, SQLException {
+        int maxTries = 60;
+        RetryCounter retrier = new RetryCounter(maxTries,
```
gjacoby126: @gokceni - Unfortunately, the RetryCounter is InterfaceAudience.Private in HBase, which means that outside projects such as Phoenix shouldn't depend on it.

gokceni: @swaroopak asked me to change the wait code to depend on it. Will role back.

phoenix-core/src/it/java/org/apache/phoenix/end2end/index/IndexMetadataIT.java
gokceni: It is expected to be empty  (the other tests don't use TASK table)

gjacoby126: Miniclusters can be shared between test suites unless you specifically request otherwise, which this test suite doesn't seem to do (it's not subclassing one of the OwnCluster suites). Since other test suites can put tasks in the table, it would be good to not assume that your row is alone.

phoenix-core/src/test/java/org/apache/phoenix/util/TestUtil.java
gjacoby126: @swaroopak - as we talked about offline, RetryCounter has InterfaceAudience.Private, which is HBase's way of telling us that external projects shouldn't depend on it because it's an internal implementation detail that can go away or change interface at any time. 

Would be nice to have something similar in Phoenix though, so perhaps we should fork it (as part of a different JIRA).

phoenix-core/src/it/java/org/apache/phoenix/end2end/index/IndexMetadataIT.java
gokceni: done

phoenix-core/src/main/java/org/apache/phoenix/schema/MetaDataClient.java
priyankporwal: @gokceni My question was about multiple indexes on the same data-table. Since you say that the tableName is data-table, does this check prevent running ALTER-INDEX REBUILD ALL on two or more of indexes on the same data-table?

gokceni: queryTaskTable queries Task table with given parameters. The parameters include both index name and table name so, it checks if we have a task for building given index on the given table. If there is another task that builds another index on the same table, tasks.size will be equal to 0. I hope this answers your question

priyankporwal: Yes. Thank you @gokceni!

priyankporwal:
+1

PR #508, opened 2019-05-16
PHOENIX-5283: Add CASCADE INDEX ALL in the SQL Grammar of ALTER TABLE…

swaroopak:
… ADD

### DIFF
--- a/phoenix-core/src/main/antlr3/PhoenixSQL.g
+++ b/phoenix-core/src/main/antlr3/PhoenixSQL.g
@@ -656,8 +656,8 @@ alter_session_node returns [AlterSessionStatement ret]
 // Parse an alter table statement.
 alter_table_node returns [AlterTableStatement ret]
     :   ALTER (TABLE | v=VIEW) t=from_table_name
-        ( (DROP COLUMN (IF ex=EXISTS)? c=column_names) | (ADD (IF NOT ex=EXISTS)? (d=column_defs) (p=fam_properties)?) | (SET (p=fam_properties)) )
-        { PTableType tt = v==null ? (QueryConstants.SYSTEM_SCHEMA_NAME.equals(t.getSchemaName()) ? PTableType.SYSTEM : PTableType.TABLE) : PTableType.VIEW; ret = ( c == null ? factory.addColumn(factory.namedTable(null,t), tt, d, ex!=null, p) : factory.dropColumn(factory.namedTable(null,t), tt, c, ex!=null) ); }
+        ( (DROP COLUMN (IF ex=EXISTS)? c=column_names) | (ADD (cas=CASCADE INDEX (list=indexes | all=ALL))? (IF NOT ex=EXISTS)? (d=column_defs) (p=fam_properties)?) | (SET (p=fam_properties)) )
+        { PTableType tt = v==null ? (QueryConstants.SYSTEM_SCHEMA_NAME.equals(t.getSchemaName()) ? PTableType.SYSTEM : PTableType.TABLE) : PTableType.VIEW; ret = ( c == null ? factory.addColumn(factory.namedTable(null,t), tt, d, ex!=null, p, cas!=null, (all == null ? list : null)) : factory.dropColumn(factory.namedTable(null,t), tt, c, ex!=null) ); }
     ;
 
 update_statistics_node returns [UpdateStatisticsStatement ret]
@@ -682,6 +682,11 @@ properties returns [Map<String,Object> ret]
 column_defs returns [List<ColumnDef> ret]
 @init{ret = new ArrayList<ColumnDef>(); }
     :  v = column_def {$ret.add(v);}  (COMMA v = column_def {$ret.add(v);} )*
+;
+
+indexes returns [List<NamedNode> ret]
+@init{ret = new ArrayList<NamedNode>(); }
+    :  v = index_name {$ret.add(v);}  (COMMA v = index_name {$ret.add(v);} )*
 ;
 
 column_def returns [ColumnDef ret]
--- a/phoenix-core/src/main/java/org/apache/phoenix/exception/SQLExceptionCode.java
+++ b/phoenix-core/src/main/java/org/apache/phoenix/exception/SQLExceptionCode.java
@@ -475,13 +475,15 @@ public enum SQLExceptionCode {
     INSUFFICIENT_MEMORY(999, "50M01", "Unable to allocate enough memory."),
     HASH_JOIN_CACHE_NOT_FOUND(900, "HJ01", "Hash Join cache not found"),
 
-    CANNOT_UPSERT_WITH_SCN_FOR_ROW_TIMSTAMP_COLUMN(901,"43M12",
+    CANNOT_UPSERT_WITH_SCN_FOR_ROW_TIMSTAMP_COLUMN(901, "43M12",
             "Cannot use a connection with SCN set to upsert data for " +
                     "table with ROW_TIMESTAMP column."),
-    CANNOT_UPSERT_WITH_SCN_FOR_MUTABLE_TABLE_WITH_INDEXES(903,"43M14",
+    CANNOT_UPSERT_WITH_SCN_FOR_MUTABLE_TABLE_WITH_INDEXES(903, "43M14",
             "Cannot use a connection with SCN set to " +
-                    "upsert data for a mutable table with indexes.");
+                    "upsert data for a mutable table with indexes."),
 
+    UNIMPLEMENTED_CASCADE_FEATURE(904, "43M15",
+            "This feature is WIP. Please refer to PHOENIX-5261");
 
     private final int errorCode;
     private final String sqlState;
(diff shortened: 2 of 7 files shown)


swaroopak:
@gokceni @priyankporwal @kadirozde Please review.

phoenix-core/src/main/antlr3/PhoenixSQL.g
```diff
@@ -656,8 +656,8 @@ alter_session_node returns [AlterSessionStatement ret]
 // Parse an alter table statement.
 alter_table_node returns [AlterTableStatement ret]
     :   ALTER (TABLE | v=VIEW) t=from_table_name
-        ( (DROP COLUMN (IF ex=EXISTS)? c=column_names) | (ADD (IF NOT ex=EXISTS)? (d=column_defs) (p=fam_properties)?) | (SET (p=fam_properties)) )
-        { PTableType tt = v==null ? (QueryConstants.SYSTEM_SCHEMA_NAME.equals(t.getSchemaName()) ? PTableType.SYSTEM : PTableType.TABLE) : PTableType.VIEW; ret = ( c == null ? factory.addColumn(factory.namedTable(null,t), tt, d, ex!=null, p) : factory.dropColumn(factory.namedTable(null,t), tt, c, ex!=null) ); }
+        ( (DROP COLUMN (IF ex=EXISTS)? c=column_names) | (ADD (cas=CASCADE INDEX (list=indexes | all=ALL))? (IF NOT ex=EXISTS)? (d=column_defs) (p=fam_properties)?) | (SET (p=fam_properties)) )
```
priyankporwal: Please confirm if this grammar works when there is an index named 'ALL'. If there a way to prevent such index/table names?

phoenix-core/src/main/java/org/apache/phoenix/jdbc/PhoenixStatement.java
```diff
@@ -1420,8 +1420,8 @@ public Long getEstimateInfoTimestamp() throws SQLException {
 
     private static class ExecutableAddColumnStatement extends AddColumnStatement implements CompilableStatement {
 
-        ExecutableAddColumnStatement(NamedTableNode table, PTableType tableType, List<ColumnDef> columnDefs, boolean ifNotExists, ListMultimap<String,Pair<String,Object>> props) {
-            super(table, tableType, columnDefs, ifNotExists, props);
+        ExecutableAddColumnStatement(NamedTableNode table, PTableType tableType, List<ColumnDef> columnDefs, boolean ifNotExists, ListMultimap<String,Pair<String,Object>> props, boolean cascade, List<NamedNode> indexes) {
+            super(table, tableType, columnDefs, ifNotExists, props, cascade, indexes);
```
priyankporwal: Did you miss 'ALL' here or does it mean that indexes is populated somewhere with names of all existing indexes?

phoenix-core/src/main/java/org/apache/phoenix/schema/MetaDataClient.java
```diff
@@ -3524,14 +3525,20 @@ private void mutateStringProperty(String tenantId, String schemaName, String tab
...
+        return addColumn(table, statement.getColumnDefs(), statement.getProps(), statement.ifNotExists(), false, statement.getTable(), statement.getTableType(), statement.isCascade(), statement.getIndexes());
     }
 
     public MutationState addColumn(PTable table, List<ColumnDef> origColumnDefs,
             ListMultimap<String, Pair<String, Object>> stmtProperties, boolean ifNotExists,
-            boolean removeTableProps, NamedTableNode namedTableNode, PTableType tableType)
+            boolean removeTableProps, NamedTableNode namedTableNode, PTableType tableType, boolean cascade, List<NamedNode> indexes)
                     throws SQLException {
         connection.rollback();
+        if(cascade || indexes == null || indexes.size()>0) {
```
priyankporwal: Hmmm, when cascade is false and indexes is null, this shouldn't throw the exception on line 3540 - perhaps a no-op [i.e. treat as indexes.size() == 0] or maybe throw an exception saying something like 'CASCADE without indexes is needless'.
Also, when cascade is false, but indexes has something, there should perhaps be another exception, such as 'invalid argument/statement'. Unsure, how that would be - but assume someone bypassed grammar checks.

phoenix-core/src/test/java/org/apache/phoenix/query/AlterTableTest.java
```diff
@@ -0,0 +1,71 @@
...
+                "SELECT * FROM TEST_TABLE" +
+                " WHERE key_column = 1");
+
+    }
+
+    // Exception for WIP for Table
+    @Test
+    public void testAlterTableAddCascade() throws SQLException {
+        String query = "ALTER TABLE TEST_TABLE ADD CASCADE INDEX ALL new_column VARCHAR";
+        exception.expectMessage("This feature is WIP. Please refer to PHOENIX-5261");
```
priyankporwal: This entire string is repeated between prod code and test. Likewise for line 62 below. Is this the convention in existing tests? Possibility to add const strings in prod code and re-use in test?

phoenix-core/src/main/antlr3/PhoenixSQL.g
swaroopak: Good point! ADD is a reserved keyword in Phoenix Grammar which already handles not allowing to create table/index with names ADD.

phoenix-core/src/main/java/org/apache/phoenix/schema/MetaDataClient.java
swaroopak: Sounds good. Also, during my IT, I found a missed condition here. I ll update the patch.

phoenix-core/src/main/java/org/apache/phoenix/jdbc/PhoenixStatement.java
swaroopak: if all is null --> code will use indexes list
if all is not null --> code will get all the indexes on that table. 
will be implemented in the upcoming PR.

phoenix-core/src/main/java/org/apache/phoenix/schema/MetaDataClient.java
swaroopak: Invalid grammar/statement related exceptions will be handled by parser logic. I ll add more UT on this.

phoenix-core/src/main/java/org/apache/phoenix/exception/SQLExceptionCode.java
```diff
@@ -480,8 +480,11 @@ public SQLException newException(SQLExceptionInfo info) {
                     "table with ROW_TIMESTAMP column."),
     CANNOT_UPSERT_WITH_SCN_FOR_MUTABLE_TABLE_WITH_INDEXES(903,"43M14",
             "Cannot use a connection with SCN set to " +
-                    "upsert data for a mutable table with indexes.");
+                    "upsert data for a mutable table with indexes."),
 
+    UNIMPLEMENTED_CASCADE_FEATURE(904,"43M15","This feature is WIP. Please refer to PHOENIX-5261"),
```
priyankporwal: Nit: Please leave spaces between different args like other usages above in this file.

phoenix-core/src/main/java/org/apache/phoenix/exception/SQLExceptionCode.java
```diff
@@ -480,8 +480,11 @@ public SQLException newException(SQLExceptionInfo info) {
...
+
+    NOT_SUPPORTED_CASCADE_FEATURE(905,"43M16","CASCADE INDEX is not supported to alter view");
```
priyankporwal: QQ: What does 43M15/M16 mean?

phoenix-core/src/main/java/org/apache/phoenix/schema/MetaDataClient.java
```diff
@@ -3524,14 +3525,26 @@ private void mutateStringProperty(String tenantId, String schemaName, String tab
...
+        return addColumn(table, statement.getColumnDefs(), statement.getProps(), statement.ifNotExists(), false, statement.getTable(), statement.getTableType(), statement.isCascade(), statement.getIndexes());
     }
 
     public MutationState addColumn(PTable table, List<ColumnDef> origColumnDefs,
             ListMultimap<String, Pair<String, Object>> stmtProperties, boolean ifNotExists,
-            boolean removeTableProps, NamedTableNode namedTableNode, PTableType tableType)
+            boolean removeTableProps, NamedTableNode namedTableNode, PTableType tableType, boolean cascade, List<NamedNode> indexes)
                     throws SQLException {
         connection.rollback();
+        if (cascade && (indexes == null || indexes.size()>0)) {
```
priyankporwal: The condition (indexes==null || indexes.size()>0) seems very un-intuitive. It requires folks to understand that indexes==null is the case with ALL and the latter is when there is explicit list of indexes; which is not very clear just reading this code. 
I'd just remove this condition entirely for now. It is not relevant for the throwing NOT_SUPPORTED_CASCADE_FEATURE exception for views. And indexes should be properly handled when the feature is implemented.

swaroopak: You are right! On a second thought, I agree with you. Ll keep the condition simple for the base PR.

phoenix-core/src/main/java/org/apache/phoenix/exception/SQLExceptionCode.java
swaroopak: My only understanding is -- it is used as code for the exception, similar to standard SQL exception codes. I made sure it gets a unique code.

PR #510, opened 2019-05-24
PHOENIX-5295: Local Index data not replicating for older HBase versio…

hnguyen08:
…ns (<= HBase 1.2)


phoenix-core/src/it/java/org/apache/phoenix/end2end/index/MutableIndexReplicationIT.java
```diff
@@ -42,11 +43,7 @@
 import org.apache.hadoop.hbase.HConstants;
 import org.apache.hadoop.hbase.HTableDescriptor;
 import org.apache.hadoop.hbase.TableName;
-import org.apache.hadoop.hbase.client.HBaseAdmin;
-import org.apache.hadoop.hbase.client.HTable;
-import org.apache.hadoop.hbase.client.Result;
-import org.apache.hadoop.hbase.client.ResultScanner;
-import org.apache.hadoop.hbase.client.Scan;
+import org.apache.hadoop.hbase.client.*;
```
willshen: I think we want to keep the import expanded.

hnguyen08:
The changes allow for the tests to be run in parallel (so that the table names for each test don't conflict)

phoenix-core/src/it/java/org/apache/phoenix/end2end/index/MutableIndexReplicationIT.java
hnguyen08: Reverted back to expanded import

gjacoby126:
@hnguyen08 - while your changes for the table names are good changes, since you're only making them to one branch they'll make it harder to maintain the code going forward. Would probably be best to save them for a different JIRA that goes everywhere.

hnguyen08:
Ok that makes sense.  I can probably leave the existing test to use the constant table names, and generate new table names in the new test.

hnguyen08:
I removed all the changes I had made to the existing test `testReplicationWithMutableIndexes` to make it easier to maintain.

phoenix-core/src/it/java/org/apache/phoenix/end2end/index/MutableIndexReplicationIT.java
```diff
@@ -274,8 +274,133 @@ private boolean ensureAnyRows(HTable remoteTable) throws IOException {
...
+        // Generating new table names so they don't conflict with previous or concurrent test runs.
+        String dataTableName = generateUniqueName();
+        String dataTableFullName = SchemaUtil.getTableName(SCHEMA_NAME, dataTableName);
+        String indexTableName = generateUniqueIndexName();
+        String indexTableFullName = SchemaUtil.getTableName(SCHEMA_NAME, indexTableName);
+
+        Connection conn = getConnection();
+
+        //create the primary and index tables
+        conn.createStatement().execute(
```
swaroopak: nit: Please include this in try-catch or try-with-resources.

phoenix-core/src/it/java/org/apache/phoenix/end2end/index/MutableIndexReplicationIT.java
```diff
@@ -274,8 +274,133 @@ private boolean ensureAnyRows(HTable remoteTable) throws IOException {
...
         Properties props = PropertiesUtil.deepCopy(TEST_PROPERTIES);
         return DriverManager.getConnection(URL, props);
     }
+
+    @Test
+    public void testReplicationWithMutableLocalIndexes() throws Exception {
+        // Generating new table names so they don't conflict with previous or concurrent test runs.
+        String dataTableName = generateUniqueName();
+        String dataTableFullName = SchemaUtil.getTableName(SCHEMA_NAME, dataTableName);
+        String indexTableName = generateUniqueIndexName();
```
twdsilva: @hnguyen08 Is there a reason you can't just use generateUniqueName() for the indexTableName?

PR #511, opened 2019-05-29
PHOENIX-4871: ParameterMetadata for join query plans is empty

BoxMiles:


gjacoby126:
@twdsilva @ChinmaySKulkarni - Mind taking a look? Patch looks good to me but you guys know the read path better than I do.

mspielberg:
The tests already added are a minimized version of the misbehaving query in the JIRA, just with the names changed to align with the existing schema in the test suites. I can add a comment to those two test cases pointing them back to the JIRA. I can also add an integration test to the phoenix-query-server repo for end-to-end, but that would have to be a different PR in the other repository.

gjacoby126:
@BoxMiles , could you please rebase against master? Looks like there's a merge conflict that prevents this patch from being committed.

PR #512, opened 2019-06-04
PHOENIX-5316 Use callable instead of runnable so that Pherf exception…

twdsilva:
…s cause tests to fail

@ChinmaySKulkarni  can you please review?


twdsilva:
@ChinmaySKulkarni , updated the PR please take a look?

PR #514, opened 2019-06-05
PHOENIX-5261: Implement ALTER TABLE/VIEW ADD COLUMN CASCADE

swaroopak:
This PR will allow an option to cascade the column to the corresponding indexes if it is added to the view or base table. (Below example is applicable for the table also)

Let's suppose I have a base table A and a global view on it as A_GLOBAL_VIEW and 3 Global (currently, focusing on global indexes) indexes on the view A_GLOBAL_VIEW_INDEX_1, A_GLOBAL_VIEW_INDEX_2, and A_GLOBAL_VIEW_INDEX_3

The following syntax will add the new_column to the global view and also to all the 3 indexes on the view as an included column:
 
ALTER VIEW A_GLOBAL_VIEW ADD new_column VARCHAR CASCADE INDEX ALL

Let's say I want to add the column only to A_GLOBAL_VIEW_INDEX_2 and A_GLOBAL_VIEW_INDEX_3, the syntax will look like below:

ALTER VIEW A_GLOBAL_VIEW ADD new_column VARCHAR CASCADE INDEX A_GLOBAL_VIEW_INDEX_2, A_GLOBAL_VIEW_INDEX_3

### DIFF
--- a/phoenix-core/src/it/java/org/apache/phoenix/end2end/AlterAddCascadeIndexIT.java
+++ b/phoenix-core/src/it/java/org/apache/phoenix/end2end/AlterAddCascadeIndexIT.java
@@ -1,0 +1,272 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.phoenix.end2end;
+
+import org.apache.phoenix.schema.PTableImpl;
+import org.junit.After;
+import org.junit.Assert;
+import org.junit.Before;
+import org.junit.Rule;
+import org.junit.Test;
+import org.junit.rules.ExpectedException;
+
+import org.junit.runner.RunWith;
+import org.junit.runners.Parameterized;
+import org.junit.runners.Parameterized.Parameters;
+
+import java.sql.Connection;
+import java.sql.DriverManager;
+import java.sql.PreparedStatement;
+import java.sql.ResultSet;
+import java.sql.SQLException;
+import java.util.Arrays;
+import java.util.Collection;
+import java.util.Properties;
+
+@RunWith(Parameterized.class)
+public class AlterAddCascadeIndexIT extends ParallelStatsDisabledIT {
+
+    @Rule
+    public ExpectedException exception = ExpectedException.none();
+    private static Connection conn;
+    private Properties prop;
+    private boolean isViewIndex;
+    private boolean isLocalIndex;
+    private String databaseObjectName;
+    private String indexesName;
+    private String planOutput;
+    private final String tableDDLOptions;
+
+
+    public AlterAddCascadeIndexIT(boolean isViewIndex, boolean isLocalIndex, boolean mutable, boolean columnEncoded) {
+        this.isViewIndex = isViewIndex;
+        this.isLocalIndex = isLocalIndex;
+        StringBuilder optionBuilder = new StringBuilder();
+        if (!columnEncoded) {
+            optionBuilder.append("COLUMN_ENCODED_BYTES=0");
+        }
+        if (!mutable && !columnEncoded) {
+            optionBuilder.append(",IMMUTABLE_ROWS=true");
+        } else if (!mutable && columnEncoded) {
+            optionBuilder.append("IMMUTABLE_ROWS=true");
+        }
+        if (!columnEncoded && !mutable) {
+            optionBuilder.append(",IMMUTABLE_STORAGE_SCHEME="+ PTableImpl.ImmutableStorageScheme.ONE_CELL_PER_COLUMN);
+        }
+        this.tableDDLOptions = optionBuilder.toString();
+    }
+
+    @Parameters(name="AlterAddCascadeIndexIT_isViewIndex={0},isLocalIndex={1},mutable={2},columnEncoded={3}")
+    public static Collection<Object[]> data() {
+        return Arrays.asList(new Object[][] {
+            { true, true, true, true},
+            { true, true, true, false},
+            { true, true, false, true},
+            { true, true, false, false},
+            { true, false, true, true},
+            { true, false, true, false},
+            { true, false, false, true},
+            { true, false, false, false},
+            { false, true, true, true},
+            { false, true, true, false},
+            { false, true, false, true},
+            { false, true, false, false},
+            { false, false, true, true},
+            { false, false, true, false},
+            { false, false, false, true},
+            { false, false, false, false}
+        });
+    }
+
+    @Before
+    public void setup() throws SQLException {
+        prop = new Properties();
+        conn = DriverManager.getConnection(getUrl(), prop);
+        conn.setAutoCommit(true);
+        conn.createStatement().execute("CREATE TABLE IF NOT EXISTS us_population (\n" +
+                "      state CHAR(2) NOT NULL,\n" +
+                "      city VARCHAR NOT NULL,\n" +
+                "      population BIGINT,\n" +
+                "      CONSTRAINT my_pk PRIMARY KEY (state, city)) " + tableDDLOptions);
+
+        loadData();
+
+        if(isViewIndex) {
+            conn.createStatement().execute("CREATE VIEW IF NOT EXISTS us_population_gv" +
+                    "(city_area INTEGER, avg_fam_size INTEGER) AS " +
+                    "SELECT * FROM us_population WHERE state = 'CA'");
+
+            conn.createStatement().execute("CREATE "+(isLocalIndex ? "LOCAL " : "") +"INDEX IF NOT EXISTS us_population_gv_gi ON " +
+                    "us_population_gv (city_area) INCLUDE (population)");
+            conn.createStatement().execute("CREATE "+(isLocalIndex ? "LOCAL " : "")+"INDEX IF NOT EXISTS us_population_gv_gi_2 ON " +
+                    "us_population_gv (avg_fam_size) INCLUDE (population)");
+            databaseObjectName = "us_population_gv";
+            indexesName = "us_population_gv_gi, us_population_gv_gi_2";
+            planOutput = isLocalIndex? "US_POPULATION" : "_IDX_US_POPULATION";
+
+        } else {
+            conn.createStatement().execute("CREATE "+ (isLocalIndex ? "LOCAL " : "") + "INDEX IF NOT EXISTS us_population_gi ON " +
+                    "us_population (population)");
+            conn.createStatement().execute("CREATE "+ (isLocalIndex ? "LOCAL " : "") + "INDEX IF NOT EXISTS us_population_gi_2 ON " +
+                    "us_population (state, population)");
+            databaseObjectName = "us_population";
+            indexesName = "us_population_gi, us_population_gi_2";
+            planOutput = isLocalIndex? "US_POPULATION" : "us_population_gi";
+        }
+    }
+
+    private void loadData() throws SQLException {
+        PreparedStatement ps = conn.prepareStatement("UPSERT INTO us_population VALUES('NY','New York',8143197)");
+        ps.executeUpdate();
+        ps = conn.prepareStatement("UPSERT INTO us_population VALUES('CA','Los Angeles',3844829)");
+        ps.executeUpdate();
+        ps = conn.prepareStatement("UPSERT INTO us_population VALUES('IL','Chicago',2842518)");
+        ps.executeUpdate();
+        ps = conn.prepareStatement("UPSERT INTO us_population VALUES('TX','Houston',2016582)");
+        ps.executeUpdate();
+        ps = conn.prepareStatement("UPSERT INTO us_population VALUES('PA','Philadelphia',1463281)");
+        ps.executeUpdate();
+        ps = conn.prepareStatement("UPSERT INTO us_population VALUES('AZ','Phoenix',1461575)");
+        ps.executeUpdate();
+        ps = conn.prepareStatement("UPSERT INTO us_population VALUES('TX','San Antonio',1256509)");
+        ps.executeUpdate();
+        ps = conn.prepareStatement("UPSERT INTO us_population VALUES('CA','San Diego',1255540)");
+        ps.executeUpdate();
+        ps = conn.prepareStatement("UPSERT INTO us_population VALUES('TX','Dallas',1213825)");
+        ps.executeUpdate();
+        ps = conn.prepareStatement("UPSERT INTO us_population VALUES('CA','San Jose',912332)");
+        ps.executeUpdate();
+    }
+
+    // Test with ALTER TABLE CASCADE INDEX ALL
+    @Test
+    public void testAlterDBOAddCascadeIndexAll() throws SQLException {
+        String query = "ALTER " +(isViewIndex? "VIEW " : "TABLE ") + databaseObjectName +" ADD new_column VARCHAR CASCADE INDEX ALL";
+        conn.createStatement().execute(query);
+        query = "EXPLAIN SELECT new_column FROM "+databaseObjectName;
+        ResultSet rs = conn.prepareStatement(query).executeQuery();
+        rs.next();
+        //confirm it is using global index to access the column
+        Assert.assertTrue(rs.getString(1).contains(planOutput.toUpperCase()));
+    }
+
+    // Test with ALTER VIEW CASCADE INDEX ALL
+    @Test
+    public void testAlterDBOAddCascadeIndexAllUpsert() throws SQLException {
+        String query = "ALTER " +(isViewIndex? "VIEW " : "TABLE ") + databaseObjectName +" ADD new_column_3 VARCHAR CASCADE INDEX ALL";
+        conn.createStatement().execute(query);
+        PreparedStatement ps;
+        if(isViewIndex) {
+            ps = conn.prepareStatement("UPSERT INTO us_population_gv(state,city,population,city_area,avg_fam_size,new_column_3) " +
+                    "VALUES('CA','Santa Barbara',912332,1300,4,'test_column')");
+        } else {
+            ps = conn.prepareStatement("UPSERT INTO us_population(state,city,population,new_column_3) " +
+                    "VALUES('CA','Santa Barbara',912332,'test_column')");
+        }
+        ps.executeUpdate();
+        query = "EXPLAIN SELECT new_column_3 FROM "+databaseObjectName+" where new_column_3 = 'test_column'";
+        ResultSet rs = conn.prepareStatement(query).executeQuery();
+        rs.next();
+        //confirm it is using global index to access the column
+        Assert.assertTrue(rs.getString(1).contains(planOutput.toUpperCase()));
+        query = "SELECT new_column_3 FROM "+databaseObjectName+" where new_column_3 = 'test_column'";
+        rs = conn.prepareStatement(query).executeQuery();
+        rs.next();
+        Assert.assertEquals(rs.getString(1),"test_column");
+    }
+
+    // Test with CASCADE INDEX <index_name>
+    @Test
+    public void testAlterDBOAddCascadeIndex() throws SQLException {
+        String query = "ALTER " + (isViewIndex? "VIEW " : "TABLE ")
+                + databaseObjectName + " ADD new_column_1 FLOAT CASCADE INDEX " + indexesName.split(",")[0];
+        conn.createStatement().execute(query);
+        query = "EXPLAIN SELECT new_column_1 FROM " + databaseObjectName;
+        ResultSet rs = conn.prepareStatement(query).executeQuery();
+        rs.next();
+        //confirm it is using global index to access the column
+        System.out.println(rs.getString(1));
+        Assert.assertTrue(rs.getString(1).contains(planOutput.toUpperCase()));
+    }
+
+    // Test with CASCADE INDEX <index_name>
+    @Test
+    public void testAlterDBOAddCascadeIndexTwoCols() throws SQLException {
+        String query = "ALTER " + (isViewIndex ? "VIEW " : "TABLE ") + databaseObjectName
+                + " ADD new_column_1 FLOAT, new_column_2 BIGINT CASCADE INDEX " + indexesName.split(",")[0];
+        conn.createStatement().execute(query);
+        query = "EXPLAIN SELECT new_column_1, new_column_2 FROM " + databaseObjectName;
+        ResultSet rs = conn.prepareStatement(query).executeQuery();
+        rs.next();
+        //confirm it is using global index to access the column
+        Assert.assertTrue(rs.getString(1).contains(planOutput.toUpperCase()));
+    }
+
+    // Test with CASCADE INDEX <index_name>, <index_name>
+    @Test
+    public void testAlterDBOAddCascadeIndexes() throws SQLException {
+        String query = "ALTER " + (isViewIndex ? "VIEW " : "TABLE ")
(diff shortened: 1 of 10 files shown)


swaroopak:
@gokceni @priyankporwal @kadirozde

phoenix-core/src/it/java/org/apache/phoenix/end2end/ParameterizedAlterTableCascadeIT.java
```diff
@@ -0,0 +1,217 @@
...
+    String databaseObjectName;
+    String indexesName;
+    String planOutput;
+
+    public ParameterizedAlterTableCascadeIT(boolean isViewIndex) {
+		this.isViewIndex = isViewIndex;
+	}
+
+	@Parameterized.Parameters(name = "ParameterizedAlterTableCascadeIT_isViewIndex={0}")
+	public static Collection parametersList() {
```
gokceni: For completeness, could you add tests for Immutable tables, and local indexes as well? 
Also another case for a multi tenant table, view index and a tenant index would be good.

phoenix-core/src/it/java/org/apache/phoenix/end2end/ParameterizedAlterTableCascadeIT.java
```diff
@@ -0,0 +1,217 @@
...
+        rs.next();
+        Assert.assertEquals(rs.getString(1),"test_column");
+    }
+
+
+
+    // Test with CASCADE INDEX <index_name>
+    @Test
+    public void testAlterDBOAddCascadeIndex() throws SQLException {
+        String query = "ALTER "+databaseObject+" "+databaseObjectName + " ADD new_column_1 FLOAT CASCADE INDEX "+indexesName.split(",")[0];
```
gokceni: Can you add multiple columns? If yes, can they all cascade? If yes, let's add a test for that.

phoenix-core/src/it/java/org/apache/phoenix/end2end/ParameterizedAlterTableCascadeIT.java
```diff
@@ -0,0 +1,217 @@
...
+
+@RunWith(Parameterized.class)
+public class ParameterizedAlterTableCascadeIT extends ParallelStatsDisabledIT {
+
+    @Rule
+    public ExpectedException exception = ExpectedException.none();
+    private static Connection conn;
+    private Properties prop;
+    boolean isViewIndex;
+    String databaseObject;
```
gokceni: Let's make these private too.

phoenix-core/src/main/java/org/apache/phoenix/schema/MetaDataClient.java
```diff
@@ -261,6 +262,7 @@
 import com.google.common.collect.Maps;
 import com.google.common.collect.Sets;
 import com.google.common.primitives.Ints;
+import sun.reflect.generics.reflectiveObjects.NotImplementedException;
```
gokceni: nit: this import doesn't seem to be used

phoenix-core/src/main/java/org/apache/phoenix/schema/MetaDataClient.java
```diff
@@ -3693,6 +3706,25 @@ public MutationState addColumn(PTable table, List<ColumnDef> origColumnDefs,
                                 .setTableName(tableName).build().buildException();
                             }
                             PColumn column = newColumn(position++, colDef, PrimaryKeyConstraint.EMPTY, table.getDefaultFamilyName() == null ? null : table.getDefaultFamilyName().getString(), true, columnQualifierBytes, willBeImmutableRows);
+                            HashMap<PTable, PColumn> indexColumn = null;
+                            if (cascade && !colDef.isPK() && numIndexes>0) {
```
gokceni: Perhaps just do and outer if on cascade to make it more readable and maintainable. 
if (cascade) {
     if (colDef.isPK) {
        throw Ex
    }
    for (PTable :) {
   }
}

phoenix-core/src/main/java/org/apache/phoenix/schema/MetaDataClient.java
```diff
@@ -3528,14 +3530,25 @@ private void mutateStringProperty(String tenantId, String schemaName, String tab
...
+                indexes.size() : table.getIndexes().size());
+        // if cascade keyword is passed and indexes are provided either implicitly or explicitly
+        if (cascade && (indexes == null || indexes.size()>0)) {
+            if (indexes == null) {
+                indexesPTable = table.getIndexes();
+            } else {
+                indexesPTable = table.getIndexes(indexes);
+            }
+        }
+        int numIndexes = indexesPTable == null ? 0 : indexesPTable.size();
```
gokceni: nit: this doesn't seem necessary

phoenix-core/src/main/antlr3/PhoenixSQL.g
```diff
@@ -656,8 +656,8 @@ alter_session_node returns [AlterSessionStatement ret]
 // Parse an alter table statement.
 alter_table_node returns [AlterTableStatement ret]
     :   ALTER (TABLE | v=VIEW) t=from_table_name
-        ( (DROP COLUMN (IF ex=EXISTS)? c=column_names) | (ADD (IF NOT ex=EXISTS)? (d=column_defs) (p=fam_properties)?) | (SET (p=fam_properties)) )
-        { PTableType tt = v==null ? (QueryConstants.SYSTEM_SCHEMA_NAME.equals(t.getSchemaName()) ? PTableType.SYSTEM : PTableType.TABLE) : PTableType.VIEW; ret = ( c == null ? factory.addColumn(factory.namedTable(null,t), tt, d, ex!=null, p) : factory.dropColumn(factory.namedTable(null,t), tt, c, ex!=null) ); }
+        ( (DROP COLUMN (IF ex=EXISTS)? c=column_names) | (ADD (IF NOT ex=EXISTS)? (d=column_defs) (p=fam_properties)?) (cas=CASCADE INDEX (list=indexes | all=ALL))? | (SET (p=fam_properties)) )
+        { PTableType tt = v==null ? (QueryConstants.SYSTEM_SCHEMA_NAME.equals(t.getSchemaName()) ? PTableType.SYSTEM : PTableType.TABLE) : PTableType.VIEW; ret = ( c == null ? factory.addColumn(factory.namedTable(null,t), tt, d, ex!=null, p, cas!=null, (all == null ? list : null)) : factory.dropColumn(factory.namedTable(null,t), tt, c, ex!=null) ); }
```
priyankporwal: (all == null ? list : null) ... Don't this need to be (all == null ? null : list) instead?

priyankporwal: nvm ... i read it reverse :)

phoenix-core/src/it/java/org/apache/phoenix/end2end/ParameterizedAlterTableCascade2IT.java
```diff
@@ -0,0 +1,4 @@
+package org.apache.phoenix.end2end;
+
+public class ParameterizedAlterTableCascade2IT {
+}
```
swaroopak: Please ignore this file. mistakely got pushed.

phoenix-core/src/main/java/org/apache/phoenix/schema/MetaDataClient.java
swaroopak: done.

phoenix-core/src/it/java/org/apache/phoenix/end2end/ParameterizedAlterTableCascadeIT.java
swaroopak: done.

phoenix-core/src/main/java/org/apache/phoenix/schema/MetaDataClient.java
swaroopak: I forget why I had numIndexes -- doesnt make sense. Thank you for correction.

phoenix-core/src/main/java/org/apache/phoenix/schema/MetaDataClient.java
swaroopak: done.

phoenix-core/src/main/java/org/apache/phoenix/parse/AddColumnStatement.java
```diff
@@ -29,12 +29,16 @@
     private final List<ColumnDef> columnDefs;
     private final boolean ifNotExists;
     private final ListMultimap<String,Pair<String,Object>> props;
-    
-    protected AddColumnStatement(NamedTableNode table, PTableType tableType, List<ColumnDef> columnDefs, boolean ifNotExists, ListMultimap<String,Pair<String,Object>> props) {
+    private final boolean cascade;
+    private final List<NamedNode> indexes;
+
```
swaroopak: nit: additional line

phoenix-core/src/it/java/org/apache/phoenix/end2end/ParameterizedAlterTableCascadeIT.java
```diff
@@ -0,0 +1,307 @@
...
+    private boolean isViewIndex;
+    private boolean isLocalIndex;
+    private String databaseObjectName;
+    private String indexesName;
+    private String planOutput;
+    private final String tableDDLOptions;
+
+
+    public ParameterizedAlterTableCascadeIT(boolean isViewIndex, boolean isLocalIndex, boolean mutable, boolean columnEncoded, String transactionProvider) {
+		this.isViewIndex = isViewIndex;
```
priyankporwal: nit: indentation

phoenix-core/src/it/java/org/apache/phoenix/end2end/ParameterizedAlterTableCascadeIT.java
```diff
@@ -0,0 +1,307 @@
...
+            if (!columnEncoded) {
+                optionBuilder.append(",IMMUTABLE_STORAGE_SCHEME="+ PTableImpl.ImmutableStorageScheme.ONE_CELL_PER_COLUMN);
+            }
+        }
+        this.tableDDLOptions = optionBuilder.toString();
+	}
+
+	@Parameters(name="ParameterizedAlterTableCascadeIT_isViewIndex={0},isLocalIndex={1},mutable={2},columnEncoded={3},transactionProvider={4}")
+	public static Collection<Object[]> data() {
+        return Arrays.asList(new Object[][] {
```
priyankporwal: Is there no better way to generate these 48 combinations? Seems to be error-prone to typos. Also, every additional param means bloating of these combinations.

phoenix-core/src/main/java/org/apache/phoenix/schema/DelegateTable.java
```diff
@@ -142,6 +143,14 @@ public Integer getBucketNum() {
         return delegate.getIndexes();
     }
 
+    @Override
+    public List<PTable> getIndexes(List<NamedNode> indexes) { return delegate.getIndexes(indexes); }
+
+    @Override
+    public List<PTable> getIndexes(boolean viewOnly) {
```
priyankporwal: Nit: Consistent indentation of the two functions. First one is all of one line, second one (though less wide) is broken down into multiple lines.

phoenix-core/src/main/java/org/apache/phoenix/schema/MetaDataClient.java
```diff
@@ -3528,14 +3529,24 @@ private void mutateStringProperty(String tenantId, String schemaName, String tab
...
+        return addColumn(table, statement.getColumnDefs(), statement.getProps(), statement.ifNotExists(), false, statement.getTable(), statement.getTableType(), statement.isCascade(), statement.getIndexes());
     }
 
     public MutationState addColumn(PTable table, List<ColumnDef> origColumnDefs,
             ListMultimap<String, Pair<String, Object>> stmtProperties, boolean ifNotExists,
-            boolean removeTableProps, NamedTableNode namedTableNode, PTableType tableType)
+            boolean removeTableProps, NamedTableNode namedTableNode, PTableType tableType, boolean cascade, List<NamedNode> indexes)
                     throws SQLException {
         connection.rollback();
+        List<PTable> indexesPTable = Lists.newArrayListWithExpectedSize(indexes != null ?
```
priyankporwal: Do we really need RHS of this statement? Looks like it is overwritten in the if (cascade...) block below.

phoenix-core/src/main/java/org/apache/phoenix/schema/MetaDataClient.java
```diff
@@ -3707,6 +3738,13 @@ public MutationState addColumn(PTable table, List<ColumnDef> origColumnDefs,
                             }
                             colFamiliesForPColumnsToBeAdded.add(column.getFamilyName() == null ? null : column.getFamilyName().getString());
                             addColumnMutation(schemaName, tableName, column, colUpsert, null, pkName, keySeq, table.getBucketNum() != null);
+                            // add new columns for given indexes one by one
+                            if (cascade) {
+                                for (PTable index: indexesPTable) {
+                                    logger.info("Adding column to "+index.getTableName().toString());
```
priyankporwal: Please add the column name here to make this log-line really useful.

phoenix-core/src/main/java/org/apache/phoenix/schema/PTableImpl.java
```diff
@@ -1445,6 +1446,35 @@ public Integer getBucketNum() {
...
+
+    @Override
+    public List<PTable> getIndexes(List<NamedNode> indexes) {
+        List<PTable> indexesPTable = Lists.newArrayListWithExpectedSize(indexes.size());
+        List<String> indexesParam = Lists.newArrayListWithExpectedSize(indexes.size());
+        for (NamedNode index : indexes) {
+            indexesParam.add(index.getName());
+        }
+        for (PTable index : this.indexes) {
+            if(indexesParam.contains(index.getTableName().getString())) {
```
priyankporwal: What if 'indexes' contains names that are not present?

phoenix-core/src/main/java/org/apache/phoenix/schema/PTableImpl.java
```diff
@@ -1445,6 +1446,35 @@ public Integer getBucketNum() {
         return indexes;
     }
 
+    @Override
+    public List<PTable> getIndexes(boolean viewOnly) {
```
priyankporwal: I think a comment block explaining the need for 'viewOnly' here would be useful. What does the function return when it's true vs false.

phoenix-core/src/main/java/org/apache/phoenix/schema/MetaDataClient.java
swaroopak: The initialization is important as the List is statically used to get the size to initialize the hashmap later.

phoenix-core/src/main/java/org/apache/phoenix/schema/MetaDataClient.java
```diff
@@ -3693,6 +3704,26 @@ public MutationState addColumn(PTable table, List<ColumnDef> origColumnDefs,
...
+                                            .build()
+                                            .buildException();
+                                }
+                                indexColumn = new HashMap(indexesPTable.size());
+                                for (PTable index : indexesPTable) {
+                                    int iPos = index.getColumns().size();
+                                    PDataType indexColDataType = IndexUtil.getIndexColumnDataType(colDef.isNull(), colDef.getDataType());
+                                    ColumnName indexColName = ColumnName.caseSensitiveColumnName(IndexUtil.getIndexColumnName(familyName, colDef.getColumnDefName().getColumnName()));
+                                    ColumnDef indexColDef = FACTORY.columnDef(indexColName, indexColDataType.getSqlTypeName(), colDef.isNull(), colDef.getMaxLength(), colDef.getScale(), false, colDef.getSortOrder(), colDef.getExpression(), colDef.isRowTimestamp());
+                                    PColumn iColumn = newColumn(iPos, indexColDef, null, "", false, null, willBeImmutableRows);
```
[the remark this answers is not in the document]
swaroopak: Good catch! Thank you for noticing.

phoenix-core/src/main/java/org/apache/phoenix/schema/MetaDataClient.java
swaroopak: This was intended more for test debugging. Detailed logging to follow.

phoenix-core/src/main/java/org/apache/phoenix/schema/PTableImpl.java
swaroopak: Yes, I agree with you. For the consistency with rest of the code, as this is an overridden method, the comments for all overridden methods are provided in the base class, in this case, PTable.java.

phoenix-core/src/it/java/org/apache/phoenix/end2end/ParameterizedAlterTableCascadeIT.java
swaroopak: I don't know more than using a parameterized framework. Do you have other suggestions?

priyankporwal: Seems like a combinatorics problem ... If each of x places, the i'th place has Ni possible values, you need to generate N1.N2.N3...Nx combinations. No existing java package/library for these things? Also, I noticed that you had many more yesterday - we don't need them anymore?

swaroopak: Not that I am aware of, but I 'll find it if there is something like this, it will certainly benefit such a long list of parameters. 

Yes, it was overkill to fit transactions scenario also in the same test suite, plus my IDE would run out of memory in running more than 128 tests in a single program.

gjacoby126:
@swaroopak , just curious -- what's the current status of this PR / JIRA? Are we waiting on a different JIRA, on additional development work, or just reviewers?

phoenix-core/src/main/java/org/apache/phoenix/schema/PTableImpl.java
gjacoby126: @swaroopak - Perhaps we should use List<String> since NamedNode is specifically part of the parse path.

swaroopak:
@gjacoby126 Have been waiting on my attention to complete it. I would take it up shortly. Thank you for your patience :)

phoenix-core/src/main/java/org/apache/phoenix/schema/PTableImpl.java
```diff
@@ -1445,6 +1446,35 @@ public Integer getBucketNum() {
...
+        List<PTable> indexes = new ArrayList<>();
+        indexes.addAll(this.indexes);
+        if (viewOnly) {
+            for (PTable index: this.indexes) {
+                if(index.getTableName().toString().contains("#")) {
+                    indexes.remove(index);
```
swaroopak, line 1456: to fix : iterating and removing from the same collection

phoenix-core/src/it/java/org/apache/phoenix/end2end/AlterAddCascadeIndexIT.java
```diff
@@ -0,0 +1,272 @@
...
+    private Properties prop;
+    private boolean isViewIndex;
+    private boolean isLocalIndex;
+    private String databaseObjectName;
+    private String indexesName;
+    private String planOutput;
+    private final String tableDDLOptions;
+
+
+    public AlterAddCascadeIndexIT(boolean isViewIndex, boolean isLocalIndex, boolean mutable, boolean columnEncoded) {
```
swaroopak, line 57: to add: test with column family

swaroopak: to add: test if the adding or removing column is atomic

phoenix-core/src/it/java/org/apache/phoenix/end2end/AlterAddCascadeIndexIT.java
```diff
@@ -0,0 +1,272 @@
...
+                    "(city_area INTEGER, avg_fam_size INTEGER) AS " +
+                    "SELECT * FROM us_population WHERE state = 'CA'");
+
+            conn.createStatement().execute("CREATE "+(isLocalIndex ? "LOCAL " : "") +"INDEX IF NOT EXISTS us_population_gv_gi ON " +
+                    "us_population_gv (city_area) INCLUDE (population)");
+            conn.createStatement().execute("CREATE "+(isLocalIndex ? "LOCAL " : "")+"INDEX IF NOT EXISTS us_population_gv_gi_2 ON " +
+                    "us_population_gv (avg_fam_size) INCLUDE (population)");
+            databaseObjectName = "us_population_gv";
+            indexesName = "us_population_gv_gi, us_population_gv_gi_2";
+            planOutput = isLocalIndex? "US_POPULATION" : "_IDX_US_POPULATION";
```
gokceni, line 121: This is not exactly plan output. I recommend either using full plan output or just rename the variable to be actualUsedTable

phoenix-core/src/it/java/org/apache/phoenix/end2end/AlterAddCascadeIndexIT.java
```diff
@@ -0,0 +1,272 @@
...
+    // Test with ALTER TABLE CASCADE INDEX ALL
+    @Test
+    public void testAlterDBOAddCascadeIndexAll() throws SQLException {
+        String query = "ALTER " +(isViewIndex? "VIEW " : "TABLE ") + databaseObjectName +" ADD new_column VARCHAR CASCADE INDEX ALL";
+        conn.createStatement().execute(query);
+        query = "EXPLAIN SELECT new_column FROM "+databaseObjectName;
+        ResultSet rs = conn.prepareStatement(query).executeQuery();
+        rs.next();
+        //confirm it is using global index to access the column
+        Assert.assertTrue(rs.getString(1).contains(planOutput.toUpperCase()));
```
gokceni, line 166: In fact, it would be easier if you have separate variables like globalIndexPlanOutput. It is easier to see what you are comparing against that way. If you add another test and just have planOutput comparison without a comment on top, it would be difficult to understand what you are asserting for.

phoenix-core/src/it/java/org/apache/phoenix/end2end/AlterAddCascadeIndexIT.java
swaroopak: to add: test for drop column

phoenix-core/src/it/java/org/apache/phoenix/end2end/AlterAddCascadeIndexIT.java
```diff
@@ -0,0 +1,272 @@
...
+        exception.expectMessage("Syntax error");
+        conn.createStatement().execute(query);
+    }
+
+    // Exception for invalid grammar
+    @Test
+    public void testAlterDBOInvalidGrammarII() throws SQLException {
+        String query = "ALTER " + (isViewIndex ? "VIEW " : "TABLE ") + databaseObjectName
+                + " ADD new_column VARCHAR CASCADE " + indexesName.split(",")[0];
+        exception.expectMessage("Syntax error");
```
gokceni, line 247: Recommend to make this a constant string

phoenix-core/src/main/java/org/apache/phoenix/schema/MetaDataClient.java
```diff
@@ -3528,14 +3529,24 @@ private void mutateStringProperty(String tenantId, String schemaName, String tab
...
-            boolean removeTableProps, NamedTableNode namedTableNode, PTableType tableType)
+            boolean removeTableProps, NamedTableNode namedTableNode, PTableType tableType, boolean cascade, List<NamedNode> indexes)
                     throws SQLException {
         connection.rollback();
+        List<PTable> indexesPTable = Lists.newArrayListWithExpectedSize(indexes != null ?
+                indexes.size() : table.getIndexes().size());
+        // if cascade keyword is passed and indexes are provided either implicitly or explicitly
+        if (cascade && (indexes == null || indexes.size()>0)) {
+            if (indexes == null) {
+                indexesPTable = table.getIndexes(table.getType().equals(PTableType.VIEW));
```
gokceni, line 3545: This means we are making an assumption here that if indexes is null this must be a VIEW. What if there is no index? Can we instead check if table is a view directly?

phoenix-core/src/it/java/org/apache/phoenix/end2end/AlterAddCascadeIndexIT.java
swaroopak: to add: testing around exception with local indexing scenario

swaroopak:
Please review newly rebased PR: https://github.com/apache/phoenix/pull/864

PR #515, opened 2019-06-06
PHOENIX-5318 Slots passed to SkipScan filter is incorrect for desc pr…

twdsilva:
…imary keys that are prefixes of each other

@dbwong  Can you please review? FYI @swaroopak

### DIFF
--- a/phoenix-core/src/it/java/org/apache/phoenix/end2end/SkipScanQueryIT.java
+++ b/phoenix-core/src/it/java/org/apache/phoenix/end2end/SkipScanQueryIT.java
@@ -29,13 +29,20 @@ import java.sql.DriverManager;
 import java.sql.PreparedStatement;
 import java.sql.ResultSet;
 import java.sql.SQLException;
+import java.sql.Statement;
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.List;
 import java.util.Properties;
 
 import org.apache.hadoop.hbase.client.HBaseAdmin;
+import org.apache.hadoop.hbase.client.Scan;
+import org.apache.hadoop.hbase.filter.Filter;
+import org.apache.hadoop.hbase.filter.FilterList;
 import org.apache.hadoop.hbase.util.Bytes;
+import org.apache.phoenix.compile.QueryPlan;
+import org.apache.phoenix.filter.SkipScanFilter;
+import org.apache.phoenix.jdbc.PhoenixStatement;
 import org.apache.phoenix.util.TestUtil;
 import org.apache.phoenix.util.SchemaUtil;
 import org.apache.phoenix.util.PropertiesUtil;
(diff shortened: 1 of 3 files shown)


swaroopak:
Interesting find, @twdsilva !

priyankporwal:
Awesome! Thanks for finding the bug and having a fix so quickly @twdsilva!

phoenix-core/src/main/java/org/apache/phoenix/compile/ScanRanges.java
```diff
@@ -615,9 +616,9 @@ private static TimeRange getRowTimestampColumnRange(List<List<KeyRange>> ranges,
                 if (ranges != null && ranges.size() > rowTimestampColPos) {
                     List<KeyRange> rowTimestampColRange = ranges.get(rowTimestampColPos);
                     List<KeyRange> sortedRange = new ArrayList<>(rowTimestampColRange);
-                    Collections.sort(sortedRange, KeyRange.COMPARATOR);
-                    //ranges.set(rowTimestampColPos, sortedRange); //TODO: do I really need to do this?
                     Field f = schema.getField(rowTimestampColPos);
+                    Collections.sort(sortedRange, f.getSortOrder() == SortOrder.ASC ? KeyRange.COMPARATOR : KeyRange.DESC_COMPARATOR);
+                    //ranges.set(rowTimestampColPos, sortedRange); //TODO: do I really need to do this?
```
willshen: Do you think it makes sense to remove this commented out code now?

phoenix-core/src/it/java/org/apache/phoenix/end2end/SkipScanQueryIT.java
```diff
@@ -584,4 +585,52 @@ public void testSkipScanJoinOptimization() throws Exception {
...
+
+
+            // test values in the skip scan filter which are prefixes of another value, eg 1,12 and 2,23
+            String sql = "select COL1, COL2 from " + tableName + " where COL1='1' OR COL1='2' OR COL1='3' OR COL1='4' " +
+                    "OR COL1='5' OR COL1='6' OR COL1='8' OR COL1='17' OR COL1='12' OR COL1='23'";
+
+
+            ResultSet rs = stmt.executeQuery(sql);
+            assertTrue(rs.next());
+            QueryPlan plan = stmt.unwrap(PhoenixStatement.class).getQueryPlan();
```
dbwong: nit: We don't seem to use the plan for anything remove?

phoenix-core/src/main/java/org/apache/phoenix/query/KeyRange.java
```diff
@@ -108,6 +109,34 @@ public KeyRange apply(byte[] input) {
         }
     };
```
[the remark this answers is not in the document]
twdsilva: @dbwong  ```KeyRange.COMPARATOR.reversed()``` will not give the correct behavior. 

For eg for the query Where COL1='1' OR COL1='12' OR COL1='2'
the key range before sorting is

```
0 = {KeyRange@8864} "\xCE" -> "1"
1 = {KeyRange@8865} "\xCE\xCD" -> "12" 
2 = {KeyRange@8866} "\xCD" -> "2"
```

Sorting using  ```KeyRange.COMPARATOR``` gives the following which isn't correct.
```
0 = {KeyRange@8879} "\xCD" -> "2"
1 = {KeyRange@8880} "\xCE" -> "1"
2 = {KeyRange@8881} "\xCE\xCD" -> "12"
```
Using ```KeyRange.COMPARATOR.reverse()``` just reverses the following order which isn't correct.

We need "12" to sort before "1" as thats how its stored on disk. Using ```DescVarLengthFastByteComparisons``` gives  the correct ordering:
```
0 = {KeyRange@8899} "\xCD" -> "2"
1 = {KeyRange@8900} "\xCE\xCD" -> "12
2 = {KeyRange@8901} "\xCE" -> "1"
```

phoenix-core/src/main/java/org/apache/phoenix/compile/ScanRanges.java
twdsilva: Its not needed will remove it.

phoenix-core/src/main/java/org/apache/phoenix/query/KeyRange.java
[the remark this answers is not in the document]
dbwong: Ok I looked in more detail and I agree.  While this may not be the ideal fix I'm still a bit worried though on if we have a region boundary there could still be issues as we may not intersect properly as the data on disk will contain the FF.  I don't think this should hurt though maybe we should add this note to https://issues.apache.org/jira/browse/PHOENIX-4306 as something we want to test in the future.

PR #517, opened 2019-06-11
PHOENIX-5211 Consistent Immutable Global Indexes for Non-Transactiona…

gokceni:
…l Tables

### DIFF
--- a/phoenix-core/src/it/java/org/apache/phoenix/end2end/IndexToolForPartialBuildIT.java
+++ b/phoenix-core/src/it/java/org/apache/phoenix/end2end/IndexToolForPartialBuildIT.java
@@ -17,13 +17,10 @@
  */
 package org.apache.phoenix.end2end;
 
-import static org.apache.phoenix.jdbc.PhoenixDatabaseMetaData.SYSTEM_CATALOG_SCHEMA;
-import static org.apache.phoenix.jdbc.PhoenixDatabaseMetaData.SYSTEM_CATALOG_TABLE;
 import static org.apache.phoenix.util.TestUtil.TEST_PROPERTIES;
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertFalse;
 import static org.junit.Assert.assertTrue;
-import static org.junit.Assert.fail;
 
 import java.sql.Connection;
 import java.sql.DriverManager;
@@ -34,21 +31,18 @@ import java.sql.Statement;
 import java.util.List;
 import java.util.Map;
 import java.util.Properties;
-import java.util.Set;
 
 import org.apache.hadoop.conf.Configuration;
-import org.apache.hadoop.hbase.DoNotRetryIOException;
-import org.apache.hadoop.hbase.HBaseIOException;
 import org.apache.hadoop.hbase.HConstants;
-import org.apache.hadoop.hbase.client.Mutation;
-import org.apache.hadoop.hbase.coprocessor.ObserverContext;
-import org.apache.hadoop.hbase.coprocessor.RegionCoprocessorEnvironment;
-import org.apache.hadoop.hbase.coprocessor.SimpleRegionObserver;
-import org.apache.hadoop.hbase.regionserver.MiniBatchOperationInProgress;
-import org.apache.hadoop.hbase.util.Bytes;
-import org.apache.phoenix.jdbc.PhoenixDatabaseMetaData;
+import org.apache.hadoop.hbase.TableName;
+import org.apache.hadoop.hbase.client.Admin;
+import org.apache.hadoop.hbase.client.Delete;
+import org.apache.hadoop.hbase.client.Result;
+import org.apache.hadoop.hbase.client.ResultScanner;
+import org.apache.hadoop.hbase.client.Scan;
+import org.apache.hadoop.hbase.client.Table;
+import org.apache.phoenix.jdbc.PhoenixConnection;
 import org.apache.phoenix.mapreduce.index.IndexTool;
-import org.apache.phoenix.query.QueryConstants;
 import org.apache.phoenix.query.QueryServices;
 import org.apache.phoenix.query.QueryServicesOptions;
 import org.apache.phoenix.schema.PIndexState;
@@ -60,6 +54,7 @@ import org.apache.phoenix.util.QueryUtil;
 import org.apache.phoenix.util.ReadOnlyProps;
 import org.apache.phoenix.util.SchemaUtil;
 import org.apache.phoenix.util.StringUtil;
+import org.apache.phoenix.util.TestUtil;
 import org.junit.BeforeClass;
 import org.junit.Test;
 
@@ -83,7 +78,6 @@ public class IndexToolForPartialBuildIT extends BaseOwnClusterIT {
     public static Map<String, String> getServerProperties() {
         Map<String, String> serverProps = Maps.newHashMapWithExpectedSize(10);
         serverProps.put(QueryServices.EXTRA_JDBC_ARGUMENTS_ATTRIB, QueryServicesOptions.DEFAULT_EXTRA_JDBC_ARGUMENTS);
-        serverProps.put("hbase.coprocessor.region.classes", FailingRegionObserver.class.getName());
         serverProps.put(" yarn.scheduler.capacity.maximum-am-resource-percent", "1.0");
         serverProps.put(HConstants.HBASE_CLIENT_RETRIES_NUMBER, "2");
         serverProps.put(HConstants.HBASE_RPC_TIMEOUT_KEY, "10000");
@@ -105,14 +99,14 @@ public class IndexToolForPartialBuildIT extends BaseOwnClusterIT {
         String schemaName = generateUniqueName();
         String dataTableName = generateUniqueName();
         String fullTableName = SchemaUtil.getTableName(schemaName, dataTableName);
-        final String indxTable = String.format("%s_%s", dataTableName, FailingRegionObserver.INDEX_NAME);
+        final String indxTable = String.format("%s_IDX", dataTableName);
         Properties props = PropertiesUtil.deepCopy(TEST_PROPERTIES);
         props.setProperty(QueryServices.TRANSACTIONS_ENABLED, Boolean.TRUE.toString());
         props.setProperty(QueryServices.EXPLAIN_ROW_COUNT_ATTRIB, Boolean.FALSE.toString());
         props.setProperty(QueryServices.IS_NAMESPACE_MAPPING_ENABLED, Boolean.toString(isNamespaceEnabled));
         final Connection conn = DriverManager.getConnection(getUrl(), props);
         Statement stmt = conn.createStatement();
-        try {
+        try (Admin admin = conn.unwrap(PhoenixConnection.class).getQueryServices().getAdmin();){
             if (isNamespaceEnabled) {
                 conn.createStatement().execute("CREATE SCHEMA IF NOT EXISTS " + schemaName);
             }
@@ -121,25 +115,36 @@ public class IndexToolForPartialBuildIT extends BaseOwnClusterIT {
                             fullTableName, tableDDLOptions));
             String upsertQuery = String.format("UPSERT INTO %s VALUES(?, ?, ?)", fullTableName);
             PreparedStatement stmt1 = conn.prepareStatement(upsertQuery);
-            FailingRegionObserver.FAIL_WRITE = false;
             // insert two rows
             upsertRow(stmt1, 1000);
             upsertRow(stmt1, 2000);
 
             conn.commit();
             stmt.execute(String.format("CREATE INDEX %s ON %s  (LPAD(UPPER(NAME),11,'x')||'_xyz') ", indxTable, fullTableName));
-            FailingRegionObserver.FAIL_WRITE = true;
             upsertRow(stmt1, 3000);
             upsertRow(stmt1, 4000);
             upsertRow(stmt1, 5000);
-            try {
-                conn.commit();
-                fail();
-            } catch (SQLException e) {} catch (Exception e) {}
+            conn.commit();
+
+            // delete these indexes
+            PTable pindexTable = PhoenixRuntime.getTable(conn, SchemaUtil.getTableName(schemaName, indxTable));
+            try (Table hTable = admin.getConnection().
+                    getTable(TableName.valueOf(pindexTable.getPhysicalName().toString()));) {
+                Scan scan = new Scan();
+                ResultScanner scanner = hTable.getScanner(scan);
+                int cnt = 0;
+                for (Result res = scanner.next(); res != null; res = scanner.next()) {
+                    cnt++;
+                    if (cnt > 2) {
+                        hTable.delete(new Delete(res.getRow()));
+                    }
+                }
+            }
+            TestUtil.doMajorCompaction(conn, pindexTable.getPhysicalName().toString());
+
             conn.createStatement()
                     .execute(String.format("ALTER INDEX %s on %s REBUILD ASYNC", indxTable, fullTableName));
             
-            FailingRegionObserver.FAIL_WRITE = false;
             ResultSet rs = conn.getMetaData().getTables(null, StringUtil.escapeLike(schemaName), indxTable,
                     new String[] { PTableType.INDEX.toString() });
             assertTrue(rs.next());
@@ -150,21 +155,6 @@ public class IndexToolForPartialBuildIT extends BaseOwnClusterIT {
             upsertRow(stmt1, 6000);
             upsertRow(stmt1, 7000);
             conn.commit();
-            
-			rs = conn.createStatement()
-					.executeQuery(String.format("SELECT " + PhoenixDatabaseMetaData.ASYNC_REBUILD_TIMESTAMP + ","
-							+ PhoenixDatabaseMetaData.INDEX_DISABLE_TIMESTAMP + " FROM "
-							+"\""+ SYSTEM_CATALOG_SCHEMA + "\"." + SYSTEM_CATALOG_TABLE + " ("
-							+ PhoenixDatabaseMetaData.ASYNC_REBUILD_TIMESTAMP + " bigint) where "
-							+ PhoenixDatabaseMetaData.TABLE_SCHEM + "='" + schemaName + "' and "
-							+ PhoenixDatabaseMetaData.TABLE_NAME + "='" + indxTable + "'"));
-			rs.next();
-            PTable pindexTable = PhoenixRuntime.getTable(conn, SchemaUtil.getTableName(schemaName, indxTable));
-            assertEquals(PIndexState.BUILDING, pindexTable.getIndexState());
-            assertEquals(rs.getLong(1), pindexTable.getTimeStamp());
-
-            //assert disabled timestamp
-            assertEquals(0, rs.getLong(2));
 
             String selectSql = String.format("SELECT LPAD(UPPER(NAME),11,'x')||'_xyz',ID FROM %s", fullTableName);
             rs = conn.createStatement().executeQuery("EXPLAIN " + selectSql);
@@ -173,7 +163,7 @@ public class IndexToolForPartialBuildIT extends BaseOwnClusterIT {
             // assert we are pulling from data table.
 			assertExplainPlan(actualExplainPlan, schemaName, dataTableName, null, isNamespaceEnabled);
 
-            rs = stmt1.executeQuery(selectSql);
+			rs = stmt1.executeQuery(selectSql);
             for (int i = 1; i <= 7; i++) {
                 assertTrue(rs.next());
                 assertEquals("xxUNAME" + i*1000 + "_xyz", rs.getString(1));
@@ -258,26 +248,6 @@ public class IndexToolForPartialBuildIT extends BaseOwnClusterIT {
         stmt.setString(2, "uname" + String.valueOf(i));
         stmt.setInt(3, 95050 + i);
         stmt.executeUpdate();
-    }
-    
-
-    public static class FailingRegionObserver extends SimpleRegionObserver {
-        public static volatile boolean FAIL_WRITE = false;
-        public static final String INDEX_NAME = "IDX";
-        @Override
-        public void preBatchMutate(ObserverContext<RegionCoprocessorEnvironment> c, MiniBatchOperationInProgress<Mutation> miniBatchOp) throws HBaseIOException {
-            if (c.getEnvironment().getRegionInfo().getTable().getNameAsString().contains(INDEX_NAME) && FAIL_WRITE) {
-                throw new DoNotRetryIOException();
-            }
-            Mutation operation = miniBatchOp.getOperation(0);
-            Set<byte[]> keySet = operation.getFamilyCellMap().keySet();
-            for(byte[] family: keySet) {
-                if(Bytes.toString(family).startsWith(QueryConstants.LOCAL_INDEX_COLUMN_FAMILY_PREFIX) && FAIL_WRITE) {
-                    throw new DoNotRetryIOException();
-                }
-            }
-        }
-
     }
     
 }
--- a/phoenix-core/src/it/java/org/apache/phoenix/end2end/index/ImmutableIndexIT.java
+++ b/phoenix-core/src/it/java/org/apache/phoenix/end2end/index/ImmutableIndexIT.java
@@ -17,15 +17,22 @@
  */
 package org.apache.phoenix.end2end.index;
 
+import static org.apache.phoenix.end2end.IndexToolIT.assertExplainPlan;
+import static org.apache.phoenix.jdbc.PhoenixDatabaseMetaData.IMMUTABLE_STORAGE_SCHEME;
+import static org.apache.phoenix.schema.PTable.ImmutableStorageScheme.SINGLE_CELL_ARRAY_WITH_OFFSETS;
 import static org.apache.phoenix.util.TestUtil.TEST_PROPERTIES;
+import static org.apache.phoenix.util.TestUtil.getRowCount;
 import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertFalse;
 import static org.junit.Assert.assertTrue;
 
+import java.io.IOException;
 import java.sql.Connection;
 import java.sql.DriverManager;
 import java.sql.ResultSet;
 import java.sql.SQLException;
 import java.sql.Statement;
+import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
 import java.util.Iterator;
@@ -40,23 +47,37 @@ import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicInteger;
 
 import org.apache.hadoop.hbase.Cell;
+import org.apache.hadoop.hbase.DoNotRetryIOException;
+import org.apache.hadoop.hbase.TableName;
+import org.apache.hadoop.hbase.client.Admin;
 import org.apache.hadoop.hbase.client.Durability;
+import org.apache.hadoop.hbase.client.HTable;
+import org.apache.hadoop.hbase.client.Mutation;
 import org.apache.hadoop.hbase.client.Put;
+import org.apache.hadoop.hbase.client.Result;
+import org.apache.hadoop.hbase.client.ResultScanner;
+import org.apache.hadoop.hbase.client.Scan;
+import org.apache.hadoop.hbase.coprocessor.ObserverContext;
 import org.apache.hadoop.hbase.coprocessor.RegionCoprocessorEnvironment;
 import org.apache.hadoop.hbase.coprocessor.SimpleRegionObserver;
+import org.apache.hadoop.hbase.regionserver.MiniBatchOperationInProgress;
 import org.apache.hadoop.hbase.util.Bytes;
 import org.apache.hadoop.hbase.util.Pair;
 import org.apache.phoenix.end2end.BaseUniqueNamesOwnClusterIT;
+import org.apache.phoenix.hbase.index.IndexRegionObserver;
 import org.apache.phoenix.jdbc.PhoenixConnection;
 import org.apache.phoenix.query.BaseTest;
 import org.apache.phoenix.query.QueryServices;
 import org.apache.phoenix.schema.PIndexState;
+import org.apache.phoenix.schema.PTable;
 import org.apache.phoenix.schema.PTableImpl;
 import org.apache.phoenix.transaction.PhoenixTransactionProvider;
 import org.apache.phoenix.transaction.PhoenixTransactionProvider.Feature;
 import org.apache.phoenix.transaction.TransactionFactory;
+import org.apache.phoenix.util.EncodedColumnsUtil;
 import org.apache.phoenix.util.PhoenixRuntime;
 import org.apache.phoenix.util.PropertiesUtil;
+import org.apache.phoenix.util.QueryUtil;
 import org.apache.phoenix.util.ReadOnlyProps;
 import org.apache.phoenix.util.SchemaUtil;
 import org.apache.phoenix.util.TestUtil;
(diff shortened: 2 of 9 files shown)


phoenix-core/src/main/java/org/apache/phoenix/execute/MutationState.java
```diff
@@ -931,220 +944,323 @@ private void send(Iterator<TableRef> tableRefIterator) throws SQLException {
...
+            // TODO maybe we can be smarter about the table name to string here?
+            Span child = Tracing.child(span, "Writing mutation batch for table: " + Bytes.toString(htableName));
+
+            int retryCount = 0;
+            boolean shouldRetry = false;
+            long numMutations = 0;
+            long mutationSizeBytes = 0;
+            long mutationCommitTime = 0;
+            long numFailedMutations = 0;
+            ;
```
swaroopak: nit: unnecessary ;

phoenix-core/src/main/java/org/apache/phoenix/execute/MutationState.java
```diff
@@ -931,220 +944,323 @@ private void send(Iterator<TableRef> tableRefIterator) throws SQLException {
...
+                    sendMutations(clonedVerifiedMutations.entrySet().iterator(), span,
+                            indexMetaDataPtr);
+                } catch (SQLException ex) {
+                    LOGGER.warn("Ignoring exception that happened during setting index verified value to TRUE", ex);
+                }
+            }
+        }
+    }
+
+    private void sendMutations(Iterator<Entry<TableInfo, List<Mutation>>> mutationsIterator, Span span, ImmutableBytesWritable indexMetaDataPtr)
```
swaroopak: nice refactoring!

phoenix-core/src/main/java/org/apache/phoenix/util/IndexUtil.java
```diff
@@ -294,13 +295,32 @@ private static boolean isEmptyKeyValue(PTable table, ColumnReference ref) {
                             .getLength()) == 0);
     }
 
+
+    public static boolean isGlobalIndexCheckerEnabled(PhoenixConnection connection, PName index) {
+        TableDescriptor desc = null;
+        try {
+            desc = connection.getQueryServices().getTableDescriptor(index.getBytes());
+        } catch (SQLException ex) {
+            //logger.warn("Error during getTableDescriptor ", ex);
```
swaroopak: nit: commented code.

phoenix-core/src/main/java/org/apache/phoenix/execute/MutationState.java
```diff
@@ -931,220 +944,323 @@ private void send(Iterator<TableRef> tableRefIterator) throws SQLException {
...
+                        }
+                    }
+                }
+
+                verifiedMutationsMap.put(pair.getKey(), indexMutations);
+                mapIter.remove();
+            }
+
+        }
+        return verifiedMutationsMap;
```
swaroopak: If I understood this function correctly, I find the variable name little misleading. verifiedMutationMap --> mutations in the map are verified, which doesn't seem to be the case. Rather, the mutations are from the first phase where verified column is set to false.  Please correct me if I am wrong.

phoenix-core/src/main/java/org/apache/phoenix/execute/MutationState.java
```diff
@@ -931,220 +944,323 @@ private void send(Iterator<TableRef> tableRefIterator) throws SQLException {
...
+    private Map<TableInfo, List<Mutation>> setVerifiedColumnForIndexes(List<TableInfo> indexTables, List<Mutation> dataMutations,
+            Map<TableInfo, List<Mutation>> deletedIndexMutations, byte[] verifiedBytes)
+            throws SQLException {
+        Map<TableInfo, List<Mutation>> updatedMap = new LinkedHashMap<>();
+        updatedMap.putAll(deletedIndexMutations);
+        for (TableInfo tableInfo : indexTables) {
+            PTable indexTable = null;
+            PTable dataTable = tableInfo.origTableRef.getTable();
+
+            List<PTable> indexes = dataTable.getIndexes();
```
swaroopak: This function will return all the indexes, including local indexes. Is that the expectation?

phoenix-core/src/main/java/org/apache/phoenix/execute/MutationState.java
```diff
@@ -931,220 +944,323 @@ private void send(Iterator<TableRef> tableRefIterator) throws SQLException {
...
+            for (PTable index : indexes) {
+                if (index.getName().equals(tableInfo.hTableName)) {
```
swaroopak: Although private member hTableName is accessible in the Outer class, better to use the public method to access it.

phoenix-core/src/it/java/org/apache/phoenix/end2end/index/ImmutableIndexIT.java
```diff
@@ -17,10 +17,15 @@
  */
 package org.apache.phoenix.end2end.index;
 
+import static org.apache.phoenix.end2end.IndexToolIT.assertExplainPlan;
+import static org.apache.phoenix.schema.types.PDataType.FALSE_BYTES;
+import static org.apache.phoenix.schema.types.PDataType.TRUE_BYTES;
```
kadirozde: Instead of using TRUE_BYTES and FALSE_BYTES, please use IndexRegionObserver.UNVERIFIED_BYTES and IndexRegionObserver.VERIFIED_BYTES. I have changed this recently.

gokceni: Will do

phoenix-core/src/it/java/org/apache/phoenix/end2end/index/ImmutableIndexIT.java
```diff
@@ -259,7 +275,122 @@ private void assertIndexMutations(Connection conn) throws SQLException {
                 (transactionProvider != null && 
                  transactionProvider.isUnsupported(Feature.MAINTAIN_LOCAL_INDEX_ON_SERVER)), iterator.hasNext());
     }
-    
+
+    private void createTableAndIndexForConsistentIndex(Connection conn, String tableName, String indexName, int numOfRowsToInsert)
+            throws Exception {
+        String ddl = "CREATE TABLE " + TABLE_NAME + TestUtil.TEST_TABLE_SCHEMA + tableDDLOptions;
+        INDEX_DDL =
```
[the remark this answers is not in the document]
gokceni: It is actually used as part of a coprocessor test class in this file. But the tests that use that coprocessor is disabled. So I think it is ok to make them local

phoenix-core/src/it/java/org/apache/phoenix/end2end/index/ImmutableIndexIT.java
```diff
@@ -259,7 +275,122 @@ private void assertIndexMutations(Connection conn) throws SQLException {
...
+                "CREATE " + " INDEX IF NOT EXISTS " + SchemaUtil.getTableNameFromFullName(indexName)
+                        + " ON " + tableName + " (long_pk, varchar_pk)"
```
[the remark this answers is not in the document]
gokceni: Ditto as above. There are existing tests and a test coprocessor that uses them.

phoenix-core/src/it/java/org/apache/phoenix/end2end/index/ImmutableIndexIT.java
```diff
@@ -259,7 +275,122 @@ private void assertIndexMutations(Connection conn) throws SQLException {
...
+                        + " INCLUDE (long_col1, long_col2) ";
+
+        Statement stmt = conn.createStatement();
+        stmt.execute(ddl);
+        conn.createStatement().execute(INDEX_DDL);
+        upsertRows(conn, tableName, numOfRowsToInsert);
+        conn.commit();
+
+        Thread.sleep(15000);
```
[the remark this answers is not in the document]
gokceni: ok

phoenix-core/src/main/java/org/apache/phoenix/execute/MutationState.java
```diff
@@ -24,14 +24,19 @@
 import static org.apache.phoenix.monitoring.GlobalClientMetrics.GLOBAL_MUTATION_COMMIT_TIME;
 import static org.apache.phoenix.query.QueryServices.WILDCARD_QUERY_DYNAMIC_COLS_ATTRIB;
 import static org.apache.phoenix.query.QueryServicesOptions.DEFAULT_WILDCARD_QUERY_DYNAMIC_COLS_ATTRIB;
+import static org.apache.phoenix.schema.types.PDataType.FALSE_BYTES;
+import static org.apache.phoenix.schema.types.PDataType.TRUE_BYTES;
```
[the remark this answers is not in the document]
gokceni: There is a MutationStateIT class, it might count as unit/it tests. 
My changes are mostly refactors. There was a huge send() method, I refactored and moved some part of send code into sendMutations function. This is the place where client creates and sends mutations for index and data table.

phoenix-core/src/main/java/org/apache/phoenix/execute/MutationState.java
gokceni: I see why it is confusing, I will do some more refactoring here

phoenix-core/src/main/java/org/apache/phoenix/execute/MutationState.java
gokceni: That is fine. It will not match the GlobalIndexChecker index name

phoenix-core/src/main/java/org/apache/phoenix/execute/MutationState.java
gokceni: ack

phoenix-core/src/main/java/org/apache/phoenix/util/IndexUtil.java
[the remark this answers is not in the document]
gokceni: ok

phoenix-core/src/it/java/org/apache/phoenix/end2end/index/ImmutableIndexIT.java
```diff
@@ -259,7 +276,136 @@ private void assertIndexMutations(Connection conn) throws SQLException {
...
+                    numRows);
+
+            ResultSet rs;
+            rs = conn.createStatement().executeQuery("SELECT /*+ NO_INDEX */ COUNT(*) FROM " + TABLE_NAME);
+            assertTrue(rs.next());
+            assertEquals(numRows, rs.getInt(1));
+            rs = conn.createStatement().executeQuery("SELECT COUNT(*) FROM " + fullIndexName);
+            assertTrue(rs.next());
+            assertEquals(numRows, rs.getInt(1));
+            verifyRowsForEmptyColValue(conn, fullIndexName, IndexRegionObserver.VERIFIED_BYTES);
```
kadirozde: The return value of verifyRowsForEmptyColValue() is not checked here and in the other places. Is it intentional? If so, what is the purpose of calling verifyRowsForEmptyColValue().

phoenix-core/src/it/java/org/apache/phoenix/end2end/index/ImmutableIndexIT.java
```diff
@@ -259,7 +276,136 @@ private void assertIndexMutations(Connection conn) throws SQLException {
...
+            assertEquals(numRows - 1, rs.getInt(1));
+            verifyRowsForEmptyColValue(conn, fullIndexName, IndexRegionObserver.UNVERIFIED_BYTES);
+
+            // Now read from index and see that we don't get any data
+            rs =
+                    conn.createStatement().executeQuery(
+                            "EXPLAIN SELECT long_pk, varchar_pk, long_col1 FROM " + TABLE_NAME + " WHERE varchar_pk='varchar2' AND long_pk=2");
+            String actualExplainPlan = QueryUtil.getExplainPlan(rs);
+            assertExplainPlan(false, actualExplainPlan, fullTableName, fullIndexName);
+            assertFalse(rs.next());
```
kadirozde: The test has failed to delete the rows WHERE varchar_pk='varchar2'. What is the purpose of this select with WHERE varchar_pk='varchar2' AND long_pk=2? and why to expect assertFalse(rs.next())?

phoenix-core/src/it/java/org/apache/phoenix/end2end/index/ImmutableIndexIT.java
gokceni: good catch!

phoenix-core/src/it/java/org/apache/phoenix/end2end/index/ImmutableIndexIT.java
gokceni: fixing it. The tests makes sure we use index for that select. rs.next should actually test the select and see it is not returning data.

phoenix-core/src/it/java/org/apache/phoenix/end2end/index/ImmutableIndexIT.java
```diff
@@ -259,7 +279,150 @@ private void assertIndexMutations(Connection conn) throws SQLException {
...
+            }
+            assertEquals(true, isDeleteFailed);
+            TestUtil.removeCoprocessor(conn, fullTableName, DeleteFailingRegionObserver.class);
+            assertEquals(numRows - 1, getRawRowCount(conn.unwrap(PhoenixConnection.class).getQueryServices()
+                    .getTable(Bytes.toBytes(fullIndexName)), false));
+            assertEquals(true, verifyRowsForEmptyColValue(conn, fullIndexName, IndexRegionObserver.UNVERIFIED_BYTES));
+
+            // Now delete via hbase, read from unverified index and see that we don't get any data
+            admin.disableTable(TableName.valueOf(fullTableName));
+            admin.truncateTable(TableName.valueOf(fullTableName), true);
```
kadirozde: Should not the table be enabled after this?

phoenix-core/src/it/java/org/apache/phoenix/end2end/index/ImmutableIndexIT.java
```diff
@@ -259,7 +279,150 @@ private void assertIndexMutations(Connection conn) throws SQLException {
...
+            try {
+                conn.createStatement().execute(dml);
+            } catch (Exception ex) {
+                isDeleteFailed = true;
+            }
+            assertEquals(true, isDeleteFailed);
+            TestUtil.removeCoprocessor(conn, fullTableName, DeleteFailingRegionObserver.class);
+            assertEquals(numRows - 1, getRawRowCount(conn.unwrap(PhoenixConnection.class).getQueryServices()
+                    .getTable(Bytes.toBytes(fullIndexName)), false));
+            assertEquals(true, verifyRowsForEmptyColValue(conn, fullIndexName, IndexRegionObserver.UNVERIFIED_BYTES));
```
kadirozde: Do we expect that all the rows are unverified here? It seems only one of the rows should be unverified.

gokceni: We have 1 row here. 1 of them is deleted on line 359

phoenix-core/src/it/java/org/apache/phoenix/end2end/index/ImmutableIndexIT.java
gokceni: If I explicitly Enable, it throws TableNotDisabled exception. I think truncate enables it

kadirozde: Got it

phoenix-core/src/it/java/org/apache/phoenix/end2end/index/ImmutableIndexIT.java
kadirozde: I missed that you set numRows.

phoenix-core/src/main/java/org/apache/phoenix/execute/MutationState.java
```diff
@@ -896,21 +906,31 @@ private void send(Iterator<TableRef> tableRefIterator) throws SQLException {
...
+                        mutationsIterator =
+                        addRowMutations(tableRef, multiRowMutationState, mutationTimestamp,
+                                serverTimestamp, false, sendAll);
                 // build map from physical table to mutation list
                 boolean isDataTable = true;
                 while (mutationsIterator.hasNext()) {
                     Pair<PName, List<Mutation>> pair = mutationsIterator.next();
                     PName hTableName = pair.getFirst();
                     List<Mutation> mutationList = pair.getSecond();
+                    if (clonedDataMutations.size() == 0) {
```
kadirozde: clonedDataMutations are not used anymore. We can remove it.

phoenix-core/src/main/java/org/apache/phoenix/execute/MutationState.java
```diff
@@ -931,220 +951,298 @@ private void send(Iterator<TableRef> tableRefIterator) throws SQLException {
...
-                            // to be done transactionally since the index is only made active after all writes have
-                            // occurred successfully.
-                            hTable = phoenixTransactionContext.getTransactionalTableWriter(connection, table, hTable, tableInfo.isDataTable() && table.getType() == PTableType.INDEX);
+
+            Map<TableInfo, List<Mutation>> unverifiedIndexMutations = new LinkedHashMap<>();
+            Map<TableInfo, List<Mutation>> verifiedOrDeletedIndexMutations = new LinkedHashMap<>();
+            filterIndexCheckerMutations(physicalTableMutationMap, unverifiedIndexMutations,
+                    verifiedOrDeletedIndexMutations);
+
+            // Phase 1: Send verified indexes with VERIFIED=false
```
kadirozde: I suggest rephrasing the comment, for example:
// Phase 1: Send index mutations with the empty column value = "unverified"

phoenix-core/src/main/java/org/apache/phoenix/execute/MutationState.java
```diff
@@ -931,220 +951,298 @@ private void send(Iterator<TableRef> tableRefIterator) throws SQLException {
...
+            // addRowMutations generates the mutations with VERIFIED=false
+            sendMutations(unverifiedIndexMutations.entrySet().iterator(), span, indexMetaDataPtr);
+
+            // Phase 2: Send data table and other indexes
+            Iterator<Entry<TableInfo, List<Mutation>>>
+                    mutationsIterator =
+                    physicalTableMutationMap.entrySet().iterator();
+            sendMutations(mutationsIterator, span, indexMetaDataPtr);
+
+            // Phase 3: Send verfied indexes with VERIFIED = true
```
kadirozde: I suggest rephrasing the comment, for example:
// Phase 3: Send put index mutations with the empty column value = "verified" and/or delete index mutations

phoenix-core/src/main/java/org/apache/phoenix/execute/MutationState.java
```diff
@@ -931,220 +951,298 @@ private void send(Iterator<TableRef> tableRefIterator) throws SQLException {
...
-                            // occurred successfully.
-                            hTable = phoenixTransactionContext.getTransactionalTableWriter(connection, table, hTable, tableInfo.isDataTable() && table.getType() == PTableType.INDEX);
+
+            Map<TableInfo, List<Mutation>> unverifiedIndexMutations = new LinkedHashMap<>();
+            Map<TableInfo, List<Mutation>> verifiedOrDeletedIndexMutations = new LinkedHashMap<>();
+            filterIndexCheckerMutations(physicalTableMutationMap, unverifiedIndexMutations,
+                    verifiedOrDeletedIndexMutations);
+
+            // Phase 1: Send verified indexes with VERIFIED=false
+            // addRowMutations generates the mutations with VERIFIED=false
```
priyankporwal: Is the comment with "addRowMutations" relevant here? If it is, I see that getting called from send() - so is it modifying the unverifiedIndexMutations iterator?

phoenix-core/src/main/java/org/apache/phoenix/execute/MutationState.java
```diff
@@ -931,220 +951,298 @@ private void send(Iterator<TableRef> tableRefIterator) throws SQLException {
...
+            sendMutations(unverifiedIndexMutations.entrySet().iterator(), span, indexMetaDataPtr);
+
+            // Phase 2: Send data table and other indexes
+            Iterator<Entry<TableInfo, List<Mutation>>>
+                    mutationsIterator =
+                    physicalTableMutationMap.entrySet().iterator();
+            sendMutations(mutationsIterator, span, indexMetaDataPtr);
```
priyankporwal: Nit: For readability, we can simply make this sendMutations(physicalTableMutationMap.entrySet().iterator(), span, indexMetaDataPtr) similar to Phase1 & Phase3 send calls.

phoenix-core/src/main/java/org/apache/phoenix/execute/MutationState.java
```diff
@@ -931,220 +951,298 @@ private void send(Iterator<TableRef> tableRefIterator) throws SQLException {
...
+
+            // Phase 3: Send verfied indexes with VERIFIED = true
+            try {
+                sendMutations(verifiedOrDeletedIndexMutations.entrySet().iterator(), span, indexMetaDataPtr);
+            } catch (SQLException ex) {
+                LOGGER.warn(
```
priyankporwal: Please add a TODO to emit metric here. Ignoring the exception may be fine, but it would be good to track how often this happens since it has implications on read-repairs.

phoenix-core/src/main/java/org/apache/phoenix/execute/MutationState.java
```diff
@@ -931,220 +951,298 @@ private void send(Iterator<TableRef> tableRefIterator) throws SQLException {
...
+                        "Ignoring exception that happened during setting index verified value to verified=TRUE",
```
priyankporwal: Does this log-line need table/index names or row-key?

phoenix-core/src/main/java/org/apache/phoenix/execute/MutationState.java
```diff
@@ -931,220 +951,298 @@ private void send(Iterator<TableRef> tableRefIterator) throws SQLException {
...
+                    if (m instanceof Delete) {
+                        Put put = new Put(m.getRow());
+                        put.addColumn(emptyCF, emptyCQ, m.getTimestamp(), IndexRegionObserver.UNVERIFIED_BYTES);
+                        // The Delete gets marked as unverified in Phase 1 and gets deleted on Phase 3.
+                        addToMap(unverifiedIndexMutations, tableInfo, put);
+                        addToMap(verifiedOrDeletedIndexMutations, tableInfo, m);
+                    } else {
+                        // These mutations already have Unverified_Bytes set.
+                        addToMap(unverifiedIndexMutations, tableInfo, m);
+                        // We will need to have verified byte set version for the Phase 3.
```
priyankporwal: Nit: "We will need .." sounds like that is being done somewhere else. Perhaps rephrase as "set verified bytes for phase3"

phoenix-core/src/main/java/org/apache/phoenix/execute/MutationState.java
gokceni: good catch

phoenix-core/src/main/java/org/apache/phoenix/execute/MutationState.java
```diff
@@ -931,220 +951,298 @@ private void send(Iterator<TableRef> tableRefIterator) throws SQLException {
...
+            Entry<TableInfo, List<Mutation>> pair = mapIter.next();
+            TableInfo tableInfo = pair.getKey();
+            if (IndexUtil.isGlobalIndexCheckerEnabled(connection, tableInfo.getHTableName())) {
+                PTable table = tableInfo.getOrigTableRef().getTable();
+                byte[] emptyCF = SchemaUtil.getEmptyColumnFamily(table);
+                byte[] emptyCQ = EncodedColumnsUtil.getEmptyKeyValueInfo(table).getFirst();
+
+                List<Mutation> mutations = pair.getValue();
+
+                if (mutations.size() == 0) {
```
priyankporwal: Remove this size check? Won't the for loop be sufficient?

phoenix-core/src/main/java/org/apache/phoenix/execute/MutationState.java
gokceni: done

phoenix-core/src/main/java/org/apache/phoenix/execute/MutationState.java
gokceni: done

phoenix-core/src/test/java/org/apache/phoenix/util/TestUtil.java
```diff
@@ -957,7 +955,8 @@ public static void waitForIndexState(Connection conn, String fullIndexName, PInd
                 }
             }
         } while (++nTries < maxTries);
-        fail("Ran out of time waiting for index state to become " + expectedIndexState);
+        fail("Ran out of time waiting for index state to become " + expectedIndexState + " last seen actual state is " +
+                (actualIndexState == null ? "" : actualIndexState.toString()));
```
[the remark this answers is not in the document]
gokceni: ok

phoenix-core/src/main/java/org/apache/phoenix/execute/MutationState.java
gokceni: ok

phoenix-core/src/main/java/org/apache/phoenix/execute/MutationState.java
gokceni: done

phoenix-core/src/main/java/org/apache/phoenix/execute/MutationState.java
gokceni: ok

phoenix-core/src/main/java/org/apache/phoenix/query/ConnectionQueryServicesImpl.java
```diff
@@ -911,7 +911,8 @@ private void addCoprocessors(byte[] tableName, TableDescriptorBuilder builder, P
                     QueryServicesOptions.DEFAULT_INDEX_REGION_OBSERVER_ENABLED);
 
             if (tableType == PTableType.INDEX && !isTransactional) {
-                if (globalIndexerEnabled && !newDesc.hasCoprocessor(GlobalIndexChecker.class.getName())) {
+                if (globalIndexerEnabled && !newDesc.hasCoprocessor(GlobalIndexChecker.class.getName()) &&
+                        !isLocalIndexTable(builder.build().getColumnFamilyNames())) {
```
gjacoby126: nit: can use newDesc for the TableDescriptor rather than using builder.build() to reconstruct it.

phoenix-core/src/main/java/org/apache/phoenix/util/IndexUtil.java
```diff
@@ -294,13 +295,29 @@ private static boolean isEmptyKeyValue(PTable table, ColumnReference ref) {
                             .getLength()) == 0);
     }
 
+
+    public static boolean isGlobalIndexCheckerEnabled(PhoenixConnection connection, PName index)
+            throws SQLException {
+        TableDescriptor desc = desc = connection.getQueryServices().getTableDescriptor(index.getBytes());
```
gjacoby126: This is called from generateIndexData which will be high-traffic. What are the perf implications of grabbing the table descriptor very frequently, and do we need caching?

Looks like this creates an HTable each time which will in turn make a remote call to the Master each time
@gokceni @kadirozde

phoenix-core/src/test/java/org/apache/phoenix/util/TestUtil.java
```diff
@@ -877,8 +875,12 @@ public static void dumpTable(Table table) throws IOException {
     }
 
     public static int getRawRowCount(Table table) throws IOException {
+        return getRawRowCount(table, true);
+    }
+
+    public static int getRawRowCount(Table table, boolean isRaw) throws IOException {
```
gjacoby126: nit: Seems a little odd to have a function named getRawRowCount that doesn't do a raw scan if you pass in false. (As opposed to having a getRowCount funciton with an isRaw parameter.)

phoenix-core/src/main/java/org/apache/phoenix/util/IndexUtil.java
gokceni: adding to a static map.

gokceni:
@gjacoby126 for your logging comment, it will complicate the code to get index name and table name at that part of the code. I can print the hashmap though

phoenix-core/src/main/java/org/apache/phoenix/util/IndexUtil.java
gokceni: Actually I am not sure about caching this. If upgrade tool adds the coprocessor, map will go stale.

phoenix-core/src/main/java/org/apache/phoenix/util/IndexUtil.java
```diff
@@ -137,7 +141,12 @@
 public class IndexUtil {
     public static final String INDEX_COLUMN_NAME_SEP = ":";
     public static final byte[] INDEX_COLUMN_NAME_SEP_BYTES = Bytes.toBytes(INDEX_COLUMN_NAME_SEP);
-    
+
+    private static Cache<String, Boolean> IndexNameGlobalIndexCheckerEnabledMap = CacheBuilder.newBuilder()
```
gokceni: @swaroopak Could you empty this cache when you do upgrade?

priyankporwal: @gokceni How? The upgrade tool would run on another machine/process. And this cache is on client-side, right?

gokceni: hmm, even if it is not possible, the cache entries expire every 30 mins.

phoenix-core/src/main/java/org/apache/phoenix/schema/PTableImpl.java
```diff
@@ -1539,6 +1539,7 @@ public IndexType getIndexType() {
     public static PTable createFromProto(PTableProtos.PTable table) {
         if (table==null)
             return null;
+
```
kadirozde: PTableImpl.java should be removed from this patch as it is included by mistake

phoenix-core/src/main/java/org/apache/phoenix/execute/MutationState.java
```diff
@@ -931,220 +947,300 @@ private void send(Iterator<TableRef> tableRefIterator) throws SQLException {
...
+                        continue;
+                    }
+                    if (m instanceof Delete) {
+                        Put put = new Put(m.getRow());
+                        put.addColumn(emptyCF, emptyCQ, IndexRegionObserver.getMaxTimestamp(m),
+                                IndexRegionObserver.UNVERIFIED_BYTES);
+                        // The Delete gets marked as unverified in Phase 1 and gets deleted on Phase 3.
+                        addToMap(unverifiedIndexMutations, tableInfo, put);
+                        addToMap(verifiedOrDeletedIndexMutations, tableInfo, m);
+                    } else {
```
gjacoby126: @gokceni - This will fail with a ClassCastException if it encounters an Append, a third kind of Mutation that acts somewhat like a Put, but reads the existing value and increments / concatenates to it. 

Since Appends also have an addColumn method, you probably just need another else if for them and similar logic to the Put.

gjacoby126: Oh, and Increment.

phoenix-core/src/main/java/org/apache/phoenix/util/IndexUtil.java
```diff
@@ -294,11 +305,37 @@ private static boolean isEmptyKeyValue(PTable table, ColumnReference ref) {
...
+        try {
+            TableDescriptor desc = connection.getQueryServices().getTableDescriptor(index.getBytes());
+
+            if (desc != null) {
+                if (desc.hasCoprocessor(GlobalIndexChecker.class.getName())) {
+                    result = true;
+                }
+            }
+            IndexNameGlobalIndexCheckerEnabledMap.put(indexName, result);
+        } catch (TableNotFoundException ex) {
```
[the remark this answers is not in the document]
gokceni: The thing is I don't have PTable, I have just the name. It is in that TableRef object. There is a ptable but it is the base table

phoenix-core/src/main/java/org/apache/phoenix/execute/MutationState.java
gokceni: According to @kadirozde , "Increment and append should not be supported by Immutable tables by definition
We need to support them for mutable tables". I guess he will have a jira for that.

phoenix-core/src/it/java/org/apache/phoenix/end2end/IndexToolForPartialBuildIT.java
```diff
@@ -121,25 +115,37 @@ public void testSecondaryIndex() throws Exception {
...
+            Scan scan = new Scan();
+            ResultScanner scanner = hTable.getScanner(scan);
+            int cnt=0;
+            for (Result res = scanner.next(); res != null; res = scanner.next()) {
+                cnt++;
+                if (cnt > 2) {
+                    hTable.delete(new Delete(res.getRow()));
+                }
+            }
+            hTable.close();
```
priyankporwal: Please create hTable with try (i.e. try with resources)

PR #518, opened 2019-06-12
PHOENIX-4285 Add PHERF.LOG_PER_NROWS constraint to PHERF.properties file

yanxinyi:


yanxinyi:
Added a getProperties function in order to have functional tests.

phoenix-pherf/src/test/java/org/apache/phoenix/pherf/PherfTest.java
```diff
@@ -40,4 +44,23 @@ public void testUnknownOption() {
         exit.expectSystemExitWithStatus(1);
         Pherf.main(args);
     }
+
```
dbwong: I don't see a test for bad values of this passed in, non-integer negative etc.

yanxinyi: sad path added

phoenix-pherf/src/test/java/org/apache/phoenix/pherf/PherfTest.java
```diff
@@ -40,4 +44,48 @@ public void testUnknownOption() {
...
+    @Test
+    public void testDefaultLogPerNRowsArgument() throws Exception {
+        String[] args = {"-listFiles"};
+        Pherf pherf = new Pherf(args);
+        Properties prop = pherf.getProperties();
+        assertEquals(Long.valueOf(PherfConstants.LOG_PER_NROWS),
```
gjacoby126: Having a constant called LOG_PER_NROWS and another called LOG_PER_N_ROWS in the same class is pretty confusing, especially when one refers to a key and the other to a value. :-)

phoenix-pherf/src/test/java/org/apache/phoenix/pherf/ResourceTest.java
```diff
@@ -51,6 +51,7 @@ public void testResourceListPropertyDirectory() throws Exception {
         PherfConstants constants = PherfConstants.create();
         assertNotNull(constants.getProperty("pherf.default.dataloader.threadpool"));
         assertNotNull(constants.getProperty("pherf.default.results.dir"));
+        assertNotNull(constants.getProperty("pherf.default.log_per_nrows"));
```
gjacoby126: Should use constant for the string

phoenix-pherf/src/test/java/org/apache/phoenix/pherf/PherfTest.java
yanxinyi: Will change `LOG_PER_N_ROWS` to `LOG_PER_NROWS_NAME`

phoenix-pherf/src/test/java/org/apache/phoenix/pherf/ResourceTest.java
yanxinyi: used constant `PherfConstants.LOG_PER_NROWS_NAME` instead of `string`

yanxinyi:
Thanks for the reviewing @dbwong @gjacoby126 :)

yanxinyi:
do you have any suggestions here @gjacoby126

PR #520, opened 2019-06-18
PHOENIX-5333: A tool to upgrade existing tables/indexes to use self-c…

swaroopak:
…onsistent global indexes design

### DIFF
--- a/phoenix-core/src/it/java/org/apache/phoenix/end2end/ParameterizedIndexUpgradeToolIT.java
+++ b/phoenix-core/src/it/java/org/apache/phoenix/end2end/ParameterizedIndexUpgradeToolIT.java
@@ -1,0 +1,328 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.phoenix.end2end;
+
+import com.google.common.collect.Maps;
+import org.apache.commons.cli.CommandLine;
+import org.apache.hadoop.hbase.TableName;
+import org.apache.hadoop.hbase.client.Admin;
+import org.apache.phoenix.hbase.index.IndexRegionObserver;
+import org.apache.phoenix.hbase.index.Indexer;
+import org.apache.phoenix.index.GlobalIndexChecker;
+import org.apache.phoenix.jdbc.PhoenixConnection;
+import org.apache.phoenix.mapreduce.index.IndexTool;
+import org.apache.phoenix.mapreduce.index.IndexUpgradeTool;
+import org.apache.phoenix.query.BaseTest;
+import org.apache.phoenix.query.ConnectionQueryServices;
+import org.apache.phoenix.query.QueryServices;
+import org.apache.phoenix.util.ReadOnlyProps;
+import org.junit.After;
+import org.junit.Assert;
+import org.junit.Before;
+import org.junit.Test;
+import org.junit.experimental.categories.Category;
+import org.junit.runner.RunWith;
+import org.junit.runners.Parameterized.Parameters;
+import org.junit.runners.Parameterized;
+import org.mockito.Mockito;
+
+import java.io.BufferedWriter;
+import java.io.File;
+import java.io.FileWriter;
+import java.io.IOException;
+import java.sql.Connection;
+import java.sql.DriverManager;
+import java.sql.SQLException;
+import java.util.Arrays;
+import java.util.Collection;
+import java.util.Map;
+import java.util.Properties;
+import java.util.UUID;
+
+import static org.apache.phoenix.mapreduce.index.IndexUpgradeTool.ROLLBACK_OP;
+import static org.apache.phoenix.mapreduce.index.IndexUpgradeTool.UPGRADE_OP;
+
+@RunWith(Parameterized.class)
+@Category(NeedsOwnMiniClusterTest.class)
+public class ParameterizedIndexUpgradeToolIT extends BaseTest {
+    //Please do not remove/uncomment commented items in the list until PHOENIX-5385 is fixed
+    private static final String [] INDEXES_LIST = {"TEST.INDEX1", "TEST.INDEX2", "TEST1.INDEX3",
+            "TEST1.INDEX2","TEST1.INDEX1","TEST.INDEX3"/*, "_IDX_TEST.MOCK1", "_IDX_TEST1.MOCK2"*/};
+    private static final String [] INDEXES_LIST_NAMESPACE = {"TEST:INDEX1", "TEST:INDEX2", "TEST1:INDEX3",
+            "TEST1:INDEX2","TEST1:INDEX1","TEST:INDEX3"/*, "TEST:_IDX_MOCK1", "TEST1:_IDX_MOCK2"*/};
+    private static final String [] TABLE_LIST = {"TEST.MOCK1","TEST1.MOCK2","TEST.MOCK3"};
+    private static final String [] TABLE_LIST_NAMESPACE = {"TEST:MOCK1","TEST1:MOCK2","TEST:MOCK3"};
+
+    private static final String INPUT_LIST = "TEST.MOCK1,TEST1.MOCK2,TEST.MOCK3";
+    private static final String INPUT_FILE = "/tmp/input_file_index_upgrade.csv";
+
+    private static Map<String, String> serverProps = Maps.newHashMapWithExpectedSize(1),
+            clientProps = Maps.newHashMapWithExpectedSize(1);
+
+    private final boolean mutable;
+    private final boolean upgrade;
+    private final boolean isNamespaceEnabled;
+
+    private StringBuilder optionsBuilder;
+    private String tableDDLOptions;
+    private Connection conn;
+    private Admin admin;
+    private IndexUpgradeTool iut;
+
+    @Before
+    public void setup () throws Exception {
+        optionsBuilder = new StringBuilder();
+
+        setClusterProperties();
+
+        setUpTestDriver(new ReadOnlyProps(serverProps.entrySet().iterator()),
+                new ReadOnlyProps(clientProps.entrySet().iterator()));
+
+        conn = DriverManager.getConnection(getUrl(), new Properties());
+        conn.setAutoCommit(true);
+        ConnectionQueryServices queryServices = conn.unwrap(PhoenixConnection.class)
+                .getQueryServices();
+        admin = queryServices.getAdmin();
+        iut = new IndexUpgradeTool(upgrade ? UPGRADE_OP : ROLLBACK_OP, INPUT_LIST,
+                null, "/tmp/index_upgrade_" + UUID.randomUUID().toString(),true,
+                Mockito.mock(IndexTool.class));
+        iut.setConf(getUtility().getConfiguration());
+        iut.setTest(true);
+        if (!mutable) {
+            optionsBuilder.append(" IMMUTABLE_ROWS=true");
+        }
+        tableDDLOptions = optionsBuilder.toString();
+        prepareSetup();
+    }
+
+    private void setClusterProperties() {
+        // we need to destroy the cluster if it was initiated using property as true
+        if (Boolean.toString(upgrade).equals(clientProps
+                .get(QueryServices.INDEX_REGION_OBSERVER_ENABLED_ATTRIB))
+                || Boolean.toString(!isNamespaceEnabled).equals(serverProps
+                .get(QueryServices.IS_NAMESPACE_MAPPING_ENABLED))) {
+            tearDownMiniClusterAsync(1);
+        }
+        //setting up properties for namespace
+        clientProps.put(QueryServices.IS_NAMESPACE_MAPPING_ENABLED,
+                Boolean.toString(isNamespaceEnabled));
+        clientProps.put(QueryServices.IS_SYSTEM_TABLE_MAPPED_TO_NAMESPACE,
+                Boolean.toString(isNamespaceEnabled));
+        serverProps.putAll(clientProps);
+        //To mimic the upgrade/rollback scenario, so that table creation uses old/new design
+        clientProps.put(QueryServices.INDEX_REGION_OBSERVER_ENABLED_ATTRIB,
+                Boolean.toString(!upgrade));
+    }
+
+    private void prepareSetup() throws SQLException {
+        //inputList is "TEST.MOCK1,TEST1.MOCK2,TEST.MOCK3";
+        if (isNamespaceEnabled) {
+            conn.createStatement().execute("CREATE SCHEMA TEST");
+            conn.createStatement().execute("CREATE SCHEMA TEST1");
+        }
+        conn.createStatement().execute("CREATE TABLE TEST.MOCK1 (id bigint NOT NULL "
+                + "PRIMARY KEY, a.name varchar, sal bigint, address varchar)"+tableDDLOptions);
+        conn.createStatement().execute("CREATE TABLE TEST1.MOCK2 (id bigint NOT NULL "
+                + "PRIMARY KEY, name varchar, city varchar, phone bigint)"+tableDDLOptions);
+        conn.createStatement().execute("CREATE TABLE TEST.MOCK3 (id bigint NOT NULL "
+                + "PRIMARY KEY, name varchar, age bigint)"+tableDDLOptions);
+        /*
+        Please do not remove/uncomment commented code until PHOENIX-5385 is fixed
+        //views
+        conn.createStatement().execute("CREATE VIEW TEST.MOCK1_VIEW (view_column varchar) "
+                + "AS SELECT * FROM TEST.MOCK1 WHERE a.name = 'a'");
+        conn.createStatement().execute("CREATE VIEW TEST.MOCK1_VIEW1 (view_column varchar,"
+                + " zip varchar) AS SELECT * FROM TEST.MOCK1 WHERE a.name = 'a'");
+        conn.createStatement().execute("CREATE VIEW TEST1.MOCK2_VIEW (view_column varchar,"
+                + " state varchar) AS SELECT * FROM TEST1.MOCK2 WHERE name = 'c'");
+        //view-indexes
+        conn.createStatement().execute("CREATE INDEX MOCK1_INDEX1 ON TEST.MOCK1_VIEW1 "
+                + "(view_column)");
+        conn.createStatement().execute("CREATE INDEX MOCK1_INDEX2 ON TEST.MOCK1_VIEW1 "
+                + "(zip)");
+        conn.createStatement().execute("CREATE INDEX MOCK2_INDEX1 ON TEST1.MOCK2_VIEW "
+                + "(state, city)");
+        conn.createStatement().execute("CREATE INDEX MOCK1_INDEX3 ON TEST.MOCK1_VIEW "
+                + "(view_column)");*/
+        //indexes
+        conn.createStatement().execute("CREATE INDEX INDEX1 ON TEST.MOCK1 (sal, a.name)");
+        conn.createStatement().execute("CREATE INDEX INDEX2 ON TEST.MOCK1 (a.name)");
+        conn.createStatement().execute("CREATE INDEX INDEX1 ON TEST1.MOCK2 (city)");
+        conn.createStatement().execute("CREATE INDEX INDEX2 ON TEST1.MOCK2 (phone)");
+        conn.createStatement().execute("CREATE INDEX INDEX3 ON TEST1.MOCK2 (name)");
+        conn.createStatement().execute("CREATE INDEX INDEX3 ON TEST.MOCK3 (age, name)");
+    }
+
+    private void validate(boolean pre) throws IOException {
+        String [] indexList = INDEXES_LIST;
+        String [] tableList = TABLE_LIST;
+        if(isNamespaceEnabled) {
+            indexList = INDEXES_LIST_NAMESPACE;
+            tableList = TABLE_LIST_NAMESPACE;
+        }
+        if (pre) {
+            if (upgrade) {
+                checkOldIndexingCoprocessors(indexList,tableList);
+            } else {
+                checkNewIndexingCoprocessors(indexList,tableList);
+            }
+        } else {
+            if (upgrade) {
+                checkNewIndexingCoprocessors(indexList,tableList);
+            } else {
+                checkOldIndexingCoprocessors(indexList,tableList);
+            }
+        }
+    }
+
+    private void checkNewIndexingCoprocessors(String [] indexList, String [] tableList) throws IOException {
+        if (mutable) {
+            for (String table : tableList) {
+                Assert.assertTrue(admin.getDescriptor(TableName.valueOf(table))
+                        .hasCoprocessor(IndexRegionObserver.class.getName()));
+                Assert.assertFalse(admin.getDescriptor(TableName.valueOf(table))
+                        .hasCoprocessor(Indexer.class.getName()));
+            }
+        }
+        for (String index : indexList) {
+            Assert.assertTrue(admin.getDescriptor(TableName.valueOf(index))
+                    .hasCoprocessor(GlobalIndexChecker.class.getName()));
+        }
+    }
+
+    private void checkOldIndexingCoprocessors(String [] indexList, String [] tableList) throws IOException {
+        if (mutable) {
+            for (String table : tableList) {
+                Assert.assertTrue(admin.getDescriptor(TableName.valueOf(table))
+                        .hasCoprocessor(Indexer.class.getName()));
+                Assert.assertFalse(admin.getDescriptor(TableName.valueOf(table))
+                        .hasCoprocessor(IndexRegionObserver.class.getName()));
+            }
+        }
+        for (String index : indexList) {
+            Assert.assertFalse(admin.getDescriptor(TableName.valueOf(index))
+                    .hasCoprocessor(GlobalIndexChecker.class.getName()));
+        }
+    }
+
+    @Parameters(name ="IndexUpgradeToolIT_mutable={0},upgrade={1},isNamespaceEnabled={2}")
+    public static Collection<Object[]> data() {
+        return Arrays.asList(new Object[][] {
+            {false, false, true},
+            {true, false, true},
+            {false, true, true},
+            {true, true, true},
+            {false, false, false},
+            {true, false, false},
+            {false, true, false},
+            {true, true, false}
+        });
+    }
+
+    public ParameterizedIndexUpgradeToolIT(boolean mutable, boolean upgrade, boolean isNamespaceEnabled) {
+        this.mutable = mutable;
+        this.upgrade = upgrade;
+        this.isNamespaceEnabled = isNamespaceEnabled;
+    }
+
+    @Test
+    public void testNonDryRunToolWithMultiTables() throws Exception {
+        validate(true);
+        iut.setDryRun(false);
+        iut.setLogFile(null);
+        iut.prepareToolSetup();
+        iut.executeTool();
+        //testing actual run
+        validate(false);
+    }
+
+    @Test
+    public void testToolWithIncorrectTables() throws Exception {
+        validate(true);
+        iut.setInputTables("TEST3.TABLE_NOT_PRESENT");
+        iut.prepareToolSetup();
+
+        int status = iut.executeTool();
+        Assert.assertEquals(-1, status);
+        validate(true);
+    }
+
+    @Test
+    public void testToolWithInputFileParameter() throws Exception {
+        BufferedWriter writer = new BufferedWriter(new FileWriter(new File(INPUT_FILE)));
+        writer.write(INPUT_LIST);
+        writer.close();
+
+        validate(true);
+
+        iut.setInputTables(null);
+        iut.setInputFile(INPUT_FILE);
+        iut.prepareToolSetup();
+        iut.executeTool();
+
+        validate(true);
+    }
+
+    @Test
+    public void testCommandLineParsing() {
+
+        String outputFile = "/tmp/index_upgrade_" + UUID.randomUUID().toString();
+        String [] args = {"-o", upgrade ? UPGRADE_OP : ROLLBACK_OP, "-tb",
+                INPUT_LIST, "-lf", outputFile, "-d"};
+        IndexUpgradeTool iut = new IndexUpgradeTool();
+
+        CommandLine cmd = iut.parseOptions(args);
+        iut.initializeTool(cmd);
+        Assert.assertEquals(iut.getDryRun(),true);
+        Assert.assertEquals(iut.getInputTables(), INPUT_LIST);
+        Assert.assertEquals(iut.getOperation(), upgrade ? UPGRADE_OP : ROLLBACK_OP);
+        Assert.assertEquals(iut.getLogFile(), outputFile);
+    }
+
+    @After
+    public void cleanup() throws SQLException {
+        //TEST.MOCK1,TEST1.MOCK2,TEST.MOCK3
+        conn.createStatement().execute("DROP INDEX INDEX1 ON TEST.MOCK1");
+        conn.createStatement().execute("DROP INDEX INDEX2 ON TEST.MOCK1");
+        conn.createStatement().execute("DROP INDEX INDEX1 ON TEST1.MOCK2");
+        conn.createStatement().execute("DROP INDEX INDEX2 ON TEST1.MOCK2");
+        conn.createStatement().execute("DROP INDEX INDEX3 ON TEST1.MOCK2");
+        conn.createStatement().execute("DROP INDEX INDEX3 ON TEST.MOCK3");
+
+        /*
+        Please do not remove/uncomment commented code until PHOENIX-5385 is fixed
+        conn.createStatement().execute("DROP INDEX MOCK1_INDEX3 ON TEST.MOCK1_VIEW");
+        conn.createStatement().execute("DROP INDEX MOCK1_INDEX1 ON TEST.MOCK1_VIEW1");
+        conn.createStatement().execute("DROP INDEX MOCK1_INDEX2 ON TEST.MOCK1_VIEW1");
+        conn.createStatement().execute("DROP INDEX MOCK2_INDEX1 ON TEST1.MOCK2_VIEW");
+
+        conn.createStatement().execute("DROP VIEW TEST.MOCK1_VIEW");
+        conn.createStatement().execute("DROP VIEW TEST.MOCK1_VIEW1");
+        conn.createStatement().execute("DROP VIEW TEST1.MOCK2_VIEW"); */
+
+        conn.createStatement().execute("DROP TABLE TEST.MOCK1");
+        conn.createStatement().execute("DROP TABLE TEST1.MOCK2");
+        conn.createStatement().execute("DROP TABLE TEST.MOCK3");
+
+        if (isNamespaceEnabled) {
+            conn.createStatement().execute("DROP SCHEMA TEST");
+            conn.createStatement().execute("DROP SCHEMA TEST1");
+        }
+        conn.close();
+    }
+}
--- a/phoenix-core/src/main/java/org/apache/phoenix/mapreduce/index/IndexUpgradeTool.java
+++ b/phoenix-core/src/main/java/org/apache/phoenix/mapreduce/index/IndexUpgradeTool.java
@@ -1,0 +1,548 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.phoenix.mapreduce.index;
+
+import com.google.common.annotations.VisibleForTesting;
+import org.apache.commons.cli.CommandLine;
+import org.apache.commons.cli.CommandLineParser;
+import org.apache.commons.cli.DefaultParser;
+import org.apache.commons.cli.HelpFormatter;
+import org.apache.commons.cli.Option;
+import org.apache.commons.cli.Options;
+import org.apache.commons.cli.ParseException;
+import org.apache.hadoop.conf.Configured;
+
+import org.apache.hadoop.hbase.HBaseConfiguration;
+import org.apache.hadoop.hbase.TableName;
+import org.apache.hadoop.hbase.client.Admin;
+import org.apache.hadoop.hbase.client.TableDescriptorBuilder;
+import org.apache.phoenix.hbase.index.IndexRegionObserver;
+import org.apache.phoenix.hbase.index.Indexer;
+import org.apache.phoenix.hbase.index.covered.NonTxIndexBuilder;
+import org.apache.phoenix.index.GlobalIndexChecker;
+import org.apache.phoenix.index.PhoenixIndexBuilder;
+import org.apache.phoenix.index.PhoenixIndexCodec;
+import org.apache.phoenix.jdbc.PhoenixConnection;
+import org.apache.phoenix.mapreduce.util.ConnectionUtil;
+import org.apache.phoenix.query.ConnectionQueryServices;
+
+import org.apache.phoenix.query.QueryServicesOptions;
+import org.apache.phoenix.schema.PTable;
+import org.apache.phoenix.schema.PTableType;
+import org.apache.phoenix.util.MetaDataUtil;
+import org.apache.phoenix.util.PhoenixRuntime;
+
+import java.sql.ResultSet;
+import java.util.Arrays;
+import java.util.HashSet;
+import java.util.List;
+import java.util.logging.Logger;
+import org.apache.hadoop.conf.Configuration;
+import org.apache.phoenix.util.SchemaUtil;
+
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Paths;
+import java.sql.Connection;
+import java.sql.SQLException;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.UUID;
+import java.util.logging.FileHandler;
+import java.util.logging.SimpleFormatter;
+import java.util.stream.Collectors;
+
+import static org.apache.phoenix.query.QueryServicesOptions.GLOBAL_INDEX_CHECKER_ENABLED_MAP_EXPIRATION_MIN;
+
+public class IndexUpgradeTool extends Configured {
+
+    private static final Logger LOGGER = Logger.getLogger(IndexUpgradeTool.class.getName());
+
+    private static final Option OPERATION_OPTION = new Option("o", "operation",
+            true,
+            "[Required]Operation to perform (upgrade/rollback)");
+    private static final Option TABLE_OPTION = new Option("tb", "table", true,
+            "[Required]Tables list ex. table1,table2");
+    private static final Option TABLE_CSV_FILE_OPTION = new Option("f", "file",
+            true,
+            "[Optional]Tables list in a csv file");
+    private static final Option DRY_RUN_OPTION = new Option("d", "dry-run",
+            false,
+            "[Optional]If passed this will output steps that will be executed");
+    private static final Option HELP_OPTION = new Option("h", "help",
+            false, "Help");
+    private static final Option LOG_FILE_OPTION = new Option("lf", "logfile",
+            true,
+            "Log file path where the logs are written");
+    private static final Option INDEX_SYNC_REBUILD_OPTION = new Option("sr", "index-sync-rebuild",
+            false,
+            "[Optional]Whether or not synchronously rebuild the indexes; default rebuild asynchronous");
+
+    public static final String UPGRADE_OP = "upgrade";
+    public static final String ROLLBACK_OP = "rollback";
+    private static final String GLOBAL_INDEX_ID = "#NA#";
+
+    private HashMap<String, HashSet<String>> tablesAndIndexes = new HashMap<>();
+    private HashMap<String, HashMap<String,String>> rebuildMap = new HashMap<>();
+    private HashMap<String, String> prop = new  HashMap<>();
+
+    private boolean dryRun, upgrade, syncRebuild;
+    private String operation;
+    private String inputTables;
+    private String logFile;
+    private String inputFile;
+    private IndexTool indexingTool;
+
+    private boolean test = false;
+
+    public void setDryRun(boolean dryRun) {
+        this.dryRun = dryRun;
+    }
+
+    public void setInputTables(String inputTables) {
+        this.inputTables = inputTables;
+    }
+
+    public void setLogFile(String logFile) {
+        this.logFile = logFile;
+    }
+
+    public void setInputFile(String inputFile) {
+        this.inputFile = inputFile;
+    }
+
+    public void setTest(boolean test) { this.test = test; }
+
+    public boolean getDryRun() {
+        return this.dryRun;
+    }
+
+    public String getInputTables() {
+        return this.inputTables;
+    }
+
+    public String getLogFile() {
+        return this.logFile;
+    }
+
+    public String getOperation() {
+        return operation;
+    }
+
+    public static void main (String[] args) {
+        CommandLine cmdLine = null;
+
+        IndexUpgradeTool iut = new IndexUpgradeTool();
+        try {
+            cmdLine = iut.parseOptions(args);
+            LOGGER.info("Index Upgrade tool initiated: "+ String.join(",", args));
+        } catch (IllegalStateException e) {
+            iut.printHelpAndExit(e.getMessage(), iut.getOptions());
+        }
+        iut.initializeTool(cmdLine);
+        iut.prepareToolSetup();
+        iut.executeTool();
+    }
+
+    public IndexUpgradeTool(String mode, String tables, String inputFile,
+            String outputFile, boolean dryRun, IndexTool indexingTool) {
+        this.operation = mode;
+        this.inputTables = tables;
+        this.inputFile = inputFile;
+        this.logFile = outputFile;
+        this.dryRun = dryRun;
+        this.indexingTool = indexingTool;
+    }
+
+    public IndexUpgradeTool () { }
+
+    /**
+     * Parses the commandline arguments, throws IllegalStateException if mandatory arguments are
+     * missing.
+     * @param args supplied command line arguments
+     * @return the parsed command line
+     */
+    @VisibleForTesting
+    public CommandLine parseOptions(String[] args) {
+
+        final Options options = getOptions();
+
+        CommandLineParser parser = new DefaultParser();
+        CommandLine cmdLine = null;
+        try {
+            cmdLine = parser.parse(options, args);
+        } catch (ParseException e) {
+            printHelpAndExit("severe parsing command line options: " + e.getMessage(),
+                    options);
+        }
+        if (cmdLine.hasOption(HELP_OPTION.getOpt())) {
+            printHelpAndExit(options, 0);
+        }
+        if (!cmdLine.hasOption(OPERATION_OPTION.getOpt())) {
+            throw new IllegalStateException(OPERATION_OPTION.getLongOpt()
+                    +" is a mandatory parameter");
+        }
+        if (cmdLine.hasOption(DRY_RUN_OPTION.getOpt())
+                && !cmdLine.hasOption(LOG_FILE_OPTION.getOpt())) {
+            throw new IllegalStateException("Log file with "+TABLE_OPTION.getLongOpt()
+                    + " is mandatory if " + DRY_RUN_OPTION.getLongOpt() +" is passed");
+        }
+        if (!(cmdLine.hasOption(TABLE_OPTION.getOpt()))
+                && !(cmdLine.hasOption(TABLE_CSV_FILE_OPTION.getOpt()))) {
+            throw new IllegalStateException("Tables list should be passed in either with"
+                    +TABLE_OPTION.getLongOpt() + " or " + TABLE_CSV_FILE_OPTION.getLongOpt());
+        }
+        if ((cmdLine.hasOption(TABLE_OPTION.getOpt()))
+                && (cmdLine.hasOption(TABLE_CSV_FILE_OPTION.getOpt()))) {
+            throw new IllegalStateException("Tables list passed in with"
+                    +TABLE_OPTION.getLongOpt() + " and " + TABLE_CSV_FILE_OPTION.getLongOpt()
+                    + "; specify only one.");
+        }
+        return cmdLine;
+    }
+
+    private void printHelpAndExit(String severeMessage, Options options) {
+        System.err.println(severeMessage);
+        printHelpAndExit(options, 1);
+    }
+
+    private void printHelpAndExit(Options options, int exitCode) {
+        HelpFormatter formatter = new HelpFormatter();
+        formatter.printHelp("help", options);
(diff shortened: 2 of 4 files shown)


phoenix-core/src/main/java/org/apache/phoenix/mapreduce/index/IndexUpgradeTool.java
```diff
@@ -0,0 +1,409 @@
...
+    private static final Option LOG_FILE_OPTION = new Option("lf", "logfile", true,
+            "Log file path where the logs are written");
+    private static final Option HOST_OPTION = new Option("ho", "host", true, "[Required]zookeeper quorum host");
+    private static final Option PORT_OPTION = new Option("po", "port", true, "[Required]zookeeper port");
+
+    private static final String UPGRADE_OP = "upgrade";
+    private static final String ROLLBACK_OP = "rollback";
+
+
+    private static HashMap <String, Boolean> tableMC = new HashMap<>();
```
priyankporwal: Any reason these need to be static members? These are not literals/consts like the ones above. So, let's just use regular class members.

swaroopak: Mainly, it needs to be accessed in more than 2 static methods and as the main method is more than an entry point, all methods needed to be static too. Do you suggest, having to create an object of the class in the main method and have other methods non-static?

priyankporwal: Mainly that we should avoid using statics unless they are constants that never change. Usage of such statics prevents parallelism and leads to race conditions that show-up much later. I noticed that your test you call main() multiple times and all these invocations using these statics. Can we just declare them locally in main() and pass them around to other helper functions?

phoenix-core/src/main/java/org/apache/phoenix/query/ConnectionQueryServicesImpl.java
```diff
@@ -911,7 +911,10 @@ private void addCoprocessors(byte[] tableName, TableDescriptorBuilder builder, P
                     QueryServicesOptions.DEFAULT_INDEX_REGION_OBSERVER_ENABLED);
 
             if (tableType == PTableType.INDEX && !isTransactional) {
-                if (globalIndexerEnabled && !newDesc.hasCoprocessor(GlobalIndexChecker.class.getName())) {
+
```
priyankporwal: A bug fix? Should this fix be done in another JIRA and PR?

phoenix-core/src/it/java/org/apache/phoenix/end2end/IndexUpgradeToolIT.java
```diff
@@ -0,0 +1,506 @@
...
+    @Test
+    public void testToolWithShowParameter() throws IOException, SQLException {
+        try(Connection conn = DriverManager.getConnection(getUrl(), new Properties())) {
+            ConnectionQueryServices queryServices = conn.unwrap(PhoenixConnection.class).getQueryServices();
+            Admin admin = queryServices.getAdmin();
+
+            ArrayList<String> args = new ArrayList<>();
+            args.add("-o");
+            if (upgrade) {
+                args.add("upgrade");
```
priyankporwal: Declare a const for "upgrade" / "rollback" and reuse between tool and test-code. Perhaps do the same for all literals used multiple times.

phoenix-core/src/it/java/org/apache/phoenix/end2end/IndexUpgradeToolIT.java
```diff
@@ -0,0 +1,506 @@
...
+    private String [] tableList = {"TEST.MOCK1","TEST1.MOCK2","TEST.MOCK3"};
+
+    private String inputList = "TEST.MOCK1:INDEX1,INDEX2;TEST1.MOCK2:INDEX3,INDEX2,INDEX1;TEST.MOCK3:INDEX3";
+
+
+    @Before
+    public void setup () throws Exception {
+        optionsBuilder = new StringBuilder();
+
+        if (upgrade) {
```
priyankporwal: I feel this code block can use some comments to help new devs understand that we are trying to restore the cluster state prior to upgrade/rollback operations.

phoenix-core/src/it/java/org/apache/phoenix/end2end/IndexUpgradeToolIT.java
```diff
@@ -0,0 +1,506 @@
...
+                    Assert.assertFalse(admin.getDescriptor(TableName.valueOf(dataTableFullName)).hasCoprocessor(Indexer.class.getName()));
+                }
+                Assert.assertTrue(admin.getDescriptor(TableName.valueOf(schemaName + "." + indexTableName)).hasCoprocessor(GlobalIndexChecker.class.getName()));
+            }
+            args.add("-tb");
+            args.add(dataTableFullName + ":" + indexTableName);
+            args.add("-c");
+            args.add(conn.unwrap(PhoenixConnection.class).getURL());
+            IndexUpgradeTool.main(args.toArray(new String[args.size()]));
+            if (upgrade) {
```
priyankporwal: This validation block from 156-168 looks like a special-case of the same logic in next test fn. Refactor to reuse the same code?

priyankporwal: And looks like all tests can share the code if we passed an array/list of indexes to verify. Will make you test file smaller.

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/index/IndexUpgradeTool.java
```diff
@@ -0,0 +1,459 @@
...
+
+                    String fullTableName = entry.getKey();
+                    PTable dataTable = PhoenixRuntime.getTableNoCache(conn, fullTableName);
+                    LOGGER.fine("Executing upgrade for "+ fullTableName);
+
+                    boolean mutable = !(dataTable.isImmutableRows());
+
+                    checkAndWaitForMajorCompaction(admin, entry);
+
+                    resetMajorCompaction(admin, entry);
```
priyankporwal: Don't we need to disable major compaction first and then wait for currently running ones to end? Otherwise, there's a slim window that another one can start between the two.

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/index/IndexUpgradeTool.java
```diff
@@ -0,0 +1,459 @@
...
+
+        } catch (SQLException e) {
+            LOGGER.severe("Something went wrong in rollbackConsistencyFeature "+ e);
+        }
+    }
+
+    private static void checkAndWaitForMajorCompaction(Admin admin, Map.Entry<String, ArrayList<String>> value) {
+        value.getValue().parallelStream().forEach(table -> {
+            try {
+                CompactionState cs = admin.getCompactionState(TableName.valueOf(table));
```
priyankporwal: Does this return MAJOR when any of the table regions is running Major compaction? What happens when some are running Minor compaction and some are running Major compaction?

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/index/IndexUpgradeTool.java
```diff
@@ -0,0 +1,459 @@
...
+                while (cs.equals(CompactionState.MAJOR)) {
+                    LOGGER.fine("Major compaction state is " + cs.name() +" for "+table+" sleeping for 5 secs before next check");
```
priyankporwal: Nit: the log line should be "Compaction state is MAJOR for ...."

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/index/IndexUpgradeTool.java
```diff
@@ -0,0 +1,459 @@
...
+                if (!dryRun) {
+                    LOGGER.fine("Major compaction state is " + cs.name() + " for " + table);
+                }
+            } catch (InterruptedException | IOException e) {
+                LOGGER.severe("Something went wrong while checking and waiting for Major Compaction for table " + table+e);
+            }
+        });
+    }
+
+    private static void resetMajorCompaction(Admin admin, Map.Entry<String, ArrayList<String>> value) {
```
priyankporwal: This and the below function could be the same with an additional boolean parameter (enable/disable). Call it "setCompactionEnabled" to be in-sync with the admin API.

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/index/IndexUpgradeTool.java
```diff
@@ -0,0 +1,459 @@
...
+            // Need to run this job in foreground for the upgrade to be deterministic
+            argArray.add("-direct");
+            argArray.add("-runfg");
+            argArray.add("-op");
+            argArray.add("/tmp/" + UUID.randomUUID().toString());
+            try {
+                LOGGER.info("Rebuilding index "+ index);
+                if(!dryRun) {
+                    indexingTool.run(argArray.toArray(new String[argArray.size()]));
+                }
```
priyankporwal: Does indexingTool wait for the index rebuild to finish? If not, is there another way to wait?

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/index/IndexUpgradeTool.java
```diff
@@ -0,0 +1,459 @@
...
+            } catch (Exception e) {
+                LOGGER.severe("Something went wrong while building the index "+ index +" " +e);
+            }
+        });
+    }
+
+    private static HashMap<String, ArrayList<String>> getTablesAndIndexes(String tables) {
```
priyankporwal: Nit: rename to parseTablesAndIndexes or extractTablesAndIndexes

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/index/IndexUpgradeTool.java
```diff
@@ -0,0 +1,459 @@
...
+            }
+            if (dryRun) {
+                LOGGER.info("This is the beginning of the tool with dry run.");
+            }
+            LOGGER.info("Index Upgrade tool initiated: "+ String.join(",", args));
+
+            prop.put(Indexer.INDEX_BUILDER_CONF_KEY, PhoenixIndexBuilder.class.getName());
+            prop.put(NonTxIndexBuilder.CODEC_CLASS_NAME_KEY, PhoenixIndexCodec.class.getName());
+
+            connectionURL = cmdLine.getOptionValue(CONNECTION_STRING_OPTION.getOpt());
```
gjacoby126: rather than a connection string, it's easier on the end user to take a ZK quorum. We can create a connection string of our own from a ZK quorum using ConnectionUtil.

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/index/IndexUpgradeTool.java
gjacoby126: @priyankporwal - what's tricky is that we're running on the mnds node and we don't necessarily know if a major compaction's going on in a region, I don't think.

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/index/IndexUpgradeTool.java
```diff
@@ -0,0 +1,456 @@
...
+                LOGGER.severe("Something went wrong while setting Major Compaction to old value "+table+" "+e);
+            }
+        });
+    }
+
+    private static void rebuildIndexes(Map.Entry<String, ArrayList<String>> value, Configuration conf) {
+        String schema = value.getKey().split("\\.")[0];
+        String table = value.getKey().split("\\.")[1];
+
+        value.getValue().stream().forEach(index -> {
```
gjacoby126: @swaroopak - unfortunately, we don't allow JDK 8 and up features like stream() in Phoenix 4.x, because HBase 1.x doesn't allow them. Since this PR is bound for master, which is 5.x, it's technically OK here, but you'd have to rewrite this for the backport, which is probably more trouble than we want to maintain.

phoenix-core/src/it/java/org/apache/phoenix/end2end/IndexUpgradeToolIT.java
```diff
@@ -0,0 +1,506 @@
...
+                        Assert.assertFalse(admin.getDescriptor(TableName.valueOf(table)).hasCoprocessor(IndexRegionObserver.class.getName()));
+                    }
+                }
+                for (String index : indexesList) {
+                    Assert.assertFalse(admin.getDescriptor(TableName.valueOf(index)).hasCoprocessor(GlobalIndexChecker.class.getName()));
+                }
+            }
+        }
+    }
+
```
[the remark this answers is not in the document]
priyankporwal: That's a good one @gjacoby126!

phoenix-core/src/main/java/org/apache/phoenix/query/ConnectionQueryServicesImpl.java
swaroopak: Added new PR: https://github.com/apache/phoenix/pull/522/files

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/index/IndexUpgradeTool.java
priyankporwal: Hmmm ... then I'm not even sure why we should rely on getCompactionState() if it does not give us accurate status of ongoing major compactions.I see there is getCompactionStateForRegion() - but that would require us to enumerate all the regions - and that may itself pose race when region splits and merges are happening.  Is there any better way?

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/index/IndexUpgradeTool.java
swaroopak: Yes, removed this option.

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/index/IndexUpgradeTool.java
swaroopak: Okay. I ll discontinue using lambda function.

phoenix-core/src/it/java/org/apache/phoenix/end2end/IndexUpgradeToolIT.java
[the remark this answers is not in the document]
priyankporwal: If we do this, perhaps we can cut-down the test matrix (data()) to just true/false for upgrade/rollback .. with a single table and 2 mutable and 2 immutable indexes. @swaroopak?

phoenix-core/src/it/java/org/apache/phoenix/end2end/IndexUpgradeToolIT.java
```diff
@@ -0,0 +1,506 @@
...
+    @After
+    public void cleanup() throws SQLException {
+        try (Connection conn = DriverManager.getConnection(getUrl(), new Properties())) {
+            //TEST.MOCK1: INDEX1, INDEX2; TEST1.MOCK2: INDEX3, INDEX2, INDEX1; TEST.MOCK3: INDEX3
```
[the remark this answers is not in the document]
swaroopak: This is to drop objects between the test runs.

phoenix-core/src/it/java/org/apache/phoenix/end2end/IndexUpgradeToolIT.java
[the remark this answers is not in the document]
swaroopak: As I know, immutable index = immutable table + index and similarly for mutable. 
How do we specify two different types of indexes on the same table?

phoenix-core/src/it/java/org/apache/phoenix/end2end/IndexUpgradeToolIT.java
```diff
@@ -0,0 +1,506 @@
...
+        String indexTableName = generateUniqueName();
+        try (Connection conn = DriverManager.getConnection(getUrl(), new Properties())) {
+            conn.setAutoCommit(true);
+            ConnectionQueryServices queryServices = conn.unwrap(PhoenixConnection.class).getQueryServices();
+            Admin admin = queryServices.getAdmin();
+            conn.createStatement().execute("CREATE TABLE " + dataTableFullName + " (id bigint NOT NULL PRIMARY KEY, name varchar)" + tableDDLOptions);
+            conn.createStatement().execute("CREATE INDEX " + indexTableName + " ON " + dataTableFullName + " (name)");
+
+            ArrayList<String> args = new ArrayList<>();
+            args.add("-o");
```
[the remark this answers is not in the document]
swaroopak: ack. ll try to implement such.

phoenix-core/src/it/java/org/apache/phoenix/end2end/IndexUpgradeToolIT.java
```diff
@@ -0,0 +1,506 @@
...
+            conn.setAutoCommit(true);
+            prepareSetup(conn);
+        }
+
+    }
+
+
+    private void prepareSetup(Connection conn) throws SQLException {
+        //inputList is "TEST.MOCK1: INDEX1, INDEX2; TEST1.MOCK2: INDEX3, INDEX2, INDEX1; TEST.MOCK3: INDEX3";
+        conn.createStatement().execute("CREATE TABLE TEST.MOCK1 (id bigint NOT NULL PRIMARY KEY, a.name varchar, sal bigint, address varchar)"+tableDDLOptions);
```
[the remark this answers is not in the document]
swaroopak: Yes, I agree. But here, since I needed to test across different schemas, tables, etc., I chose to have a defined list which can be prepopulated to pass in -tb or -f parameter as per the tests.

phoenix-core/src/it/java/org/apache/phoenix/end2end/IndexUpgradeToolIT.java
```diff
@@ -0,0 +1,506 @@
...
+        optionsBuilder = new StringBuilder();
+
+        if (upgrade) {
+            if ((Boolean.TRUE.toString()).equals(clientProps.get(QueryServices.INDEX_REGION_OBSERVER_ENABLED_ATTRIB))) {
+                tearDownMiniClusterAsync(1);
+            }
+            clientProps.put(QueryServices.INDEX_REGION_OBSERVER_ENABLED_ATTRIB, Boolean.FALSE.toString());
+        } else {
+            if ((Boolean.FALSE.toString()).equals(clientProps.get(QueryServices.INDEX_REGION_OBSERVER_ENABLED_ATTRIB))) {
+                tearDownMiniClusterAsync(1);
```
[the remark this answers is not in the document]
swaroopak: Comparing it with the older value of the parameter, this as well makes sure that there are 2 startups and 2 shutdowns only. 

However, as you pointed out since I am not changing server properties, I don't need a new cluster. Removing it.

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/index/IndexUpgradeTool.java
swaroopak: done.

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/index/IndexUpgradeTool.java
swaroopak: status is MAJOR_AND_MINOR. 
updated the code.

swaroopak:
Thank you for reviewing @priyankporwal  @gjacoby126 
I updated the PR fixing some comments. ll get back to rest of them shortly.

phoenix-core/src/it/java/org/apache/phoenix/end2end/IndexUpgradeToolIT.java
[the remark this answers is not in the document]
priyankporwal: Ahh .. You may be right @swaroopak. After all immutability is the property of the data-table, not the indexes.

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/index/IndexUpgradeTool.java
swaroopak: Yes, I am passing -runfg (run foreground) which makes sure that the program waits till the tool completes. https://github.com/apache/phoenix/blob/master/phoenix-core/src/main/java/org/apache/phoenix/mapreduce/index/IndexTool.java#L690

phoenix-core/src/it/java/org/apache/phoenix/end2end/IndexUpgradeToolIT.java
swaroopak: done.

phoenix-core/src/it/java/org/apache/phoenix/end2end/IndexUpgradeToolIT.java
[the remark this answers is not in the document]
swaroopak: The driver is created using client properties which are changing in this case, hence I 'll have to tear down the cluster used in rollback before moving on to upgrade scenario.

swaroopak:
I updated the PR as per the comments. Please review. In case, I have missed addressing any, please reopen. @gjacoby126  @priyankporwal

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/index/IndexUpgradeTool.java
```diff
@@ -0,0 +1,537 @@
...
+
+            try (Admin admin = queryServices.getAdmin()) {
+
+                PTable dataTable = PhoenixRuntime.getTableNoCache(conn, dataTableFullName);
+                LOGGER.fine("Executing "+operation+" for " + dataTableFullName);
+
+                boolean mutable = !(dataTable.isImmutableRows());
+
+                if (upgrade) {
+                    setCompactionEnabled(admin, entry, false);
```
priyankporwal: Nit: Instead of passing-around entry to all these helper functions, they would be much more intuitive if you passed them (String tableName, ArrayList<String> indexNames) explicitly - one of both of them, as needed by these functions.

priyankporwal:
+1 @swaroopak. The refactoring changes look great! Just a nit and waiting for compaction discussion with @gjacoby126.

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/index/IndexUpgradeTool.java
```diff
@@ -0,0 +1,550 @@
...
+
+        LOGGER.info("Executing "+ operation);
+
+        for (Map.Entry<String, ArrayList<String>> entry: tablesAndIndexes.entrySet()) {
+            String dataTableFullName = entry.getKey();
+            ArrayList<String> indexes = entry.getValue();
+
+            try (Admin admin = queryServices.getAdmin()) {
+
+                PTable dataTable = PhoenixRuntime.getTableNoCache(conn, dataTableFullName);
```
gjacoby126: This logic in general is assuming that the logical table name in Phoenix == the physical table in HBase. This assumption's not always going to be valid. 

Three big cases: 
1. If namespaces are enabled in HBase, and namespace / schema mapping is enabled in Phoenix. In that case, a table named FOO.BAR in Phoenix will be named FOO:BAR in HBase
2. Global views can own indexes, and global views will have different names from the physical table they're located in. Likewise, indexes on views will have names in Phoenix, but be physically located in tables named _IDX_<name of HBase data table> 
3. Tenant-owned views can own indexes, and they can't be seen using a PhoenixRuntime.getTable() command, unfortunately.

gjacoby126: Thinking about this more -- I think for 2 and 3 we just need to automatically upgrade the _IDX_ table when its associated data table is being upgraded, and downgrade it with the associated table too. Because all view indexes of a table share the same physical table, we can't pick and choose like we do for global indexes of a base table.

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/index/IndexUpgradeTool.java
```diff
@@ -0,0 +1,550 @@
...
+                    rebuildIndexes(entry, conf);
+                    if (compaction) {
+                        setCompactionEnabled(admin, entry, true);
+                    }
+                } else {
+                    disableTable(admin, entry);
+                    if(mutable) {
+                        modifyDataTable(admin, dataTableFullName);
+                    }
+                    modifyIndexTable(admin, indexes);
```
gokceni: Could you clean IndexNameGlobalIndexCheckerEnabledMap at this point? My PR is introducing it. It is in indexutil

kadirozde: @gokceni, we should define a constant for this, if it is not defined yet. The upgrade tool should sleep this constant + 1 minutes.

gokceni: Yes, it is called GLOBAL_INDEX_CHECKER_ENABLED_MAP_EXPIRATION_MIN

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/index/IndexUpgradeTool.java
priyankporwal: @swaroopak This was the comment that I mentioned yesterday.

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/index/IndexUpgradeTool.java
priyankporwal: The data-tables and index-tables are disabled while their metadata is changed. Let's make sure the sleeps (if necessary) factor in the time since end of disable. Just to ensure we don't sleep for too long.

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/index/IndexUpgradeTool.java
swaroopak: As discussed offline, PHOENIX-5373 will consider the scenario of stale rows and upgrade can avoid hitting race condition if any due to compaction.

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/index/IndexUpgradeTool.java
swaroopak: Aah! Noted. :)

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/index/IndexUpgradeTool.java
```diff
@@ -0,0 +1,550 @@
...
+            try (Admin admin = queryServices.getAdmin()) {
+
+                PTable dataTable = PhoenixRuntime.getTableNoCache(conn, dataTableFullName);
+                LOGGER.fine("Executing "+operation+" for " + dataTableFullName);
+
+                boolean mutable = !(dataTable.isImmutableRows());
+
+                if (upgrade) {
+                    if (compaction) {
+                        setCompactionEnabled(admin, entry, false);
```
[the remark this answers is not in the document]
swaroopak: We don't need to, but, keeping it around as an optional parameter. I think I should get rid of it as we have a solid design to handle that.

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/index/IndexUpgradeTool.java
```diff
@@ -0,0 +1,550 @@
...
+        }
+    }
+
+    private HashMap<String, ArrayList<String>> extractTablesAndIndexes(String tables) {
+        String [] entries = tables.trim().split(";");
+        HashMap<String, ArrayList<String>> tablesAndIndexes = new HashMap<>();
+        try {
+            for (String entry : entries) {
+                String tableName = entry.split(":")[0];
+                String schema = tableName.split("\\.")[0];
```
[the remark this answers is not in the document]
swaroopak: Done

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/index/IndexUpgradeTool.java
```diff
@@ -0,0 +1,550 @@
...
+                LOGGER.severe("Something went wrong in setCompactionEnabled for "+table+" "+e);
+            }
+        }
+        if (setMC) {
+            tableMC.clear();
+        }
+    }
+
+    private void rebuildIndexes(Map.Entry<String, ArrayList<String>> value, Configuration conf) {
+        String schema = value.getKey().split("\\.")[0];
```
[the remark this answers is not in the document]
swaroopak: done

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/index/IndexUpgradeTool.java
```diff
@@ -0,0 +1,550 @@
...
+                LOGGER.fine("Executing "+operation+" for " + dataTableFullName);
+
+                boolean mutable = !(dataTable.isImmutableRows());
+
+                if (upgrade) {
+                    if (compaction) {
+                        setCompactionEnabled(admin, entry, false);
+                        checkAndWaitForMajorCompaction(admin, entry);
+                    }
+                    disableTable(admin, entry);
```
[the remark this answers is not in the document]
swaroopak: Yes, I am going to use k8s_hbase multi-node cluster to test this tool at scale.

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/index/IndexUpgradeTool.java
```diff
@@ -0,0 +1,550 @@
...
+        iut.prepareToolSetup();
+        iut.executeTool();
+    }
+
+    public IndexUpgradeTool(String mode, String tables, String inputFile, String host, String port,
+            String outputFile, boolean dryRun, boolean compaction) {
+        this.operation = mode;
+        this.inputTables = tables;
+        this.inputFile = inputFile;
+        this.connectionURL = "jdbc:phoenix:" + host + ":" + port;
```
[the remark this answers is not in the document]
swaroopak: any reason we can't use ZK quorum and ConnectionUtil.getInputConnection? 
> Priyank also mentioned this initially. When I tried using it, IT was failing and now I know why! This tool is like a client executing necessary commands similar to sqlline, didn't realize tool is extending configured class which I can make use of and get rid of host/port as well. ll address this. 
Will this be sufficient to talk to a secure cluster? (Where would the user/principal be set?)
> This would change as I use getInputConnection method. I don't have to rely on URL anymore.

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/index/IndexUpgradeTool.java
```diff
@@ -0,0 +1,550 @@
...
+                } else {
+                    disableTable(admin, entry);
+                    if(mutable) {
+                        modifyDataTable(admin, dataTableFullName);
+                    }
+                    modifyIndexTable(admin, indexes);
+                    enableTable(admin, entry);
+                }
+            } catch (IOException | SQLException e) {
+                LOGGER.severe("Something went wrong while executing "+operation+" steps " + e);
```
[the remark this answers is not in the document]
swaroopak: That's a good point. I'll include more safety methods.

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/index/IndexUpgradeTool.java
swaroopak: In Point 2 and 3, the tool accepts the indexes and table list to be upgraded. For consistency, it is better to pass all the indexes as you said. In that case, I think only point one needs to be addressed here. Is that correct?

gjacoby126: @swaroopak , yes, I think just automatically doing all view indexes of a data table passed in would be best. But you'll have to add logic to map the logical view index names to the physical view index table (and make sure you only do the upgrade once) if they provide view index names. 

There should also be error handling logic if the user provides data table and/or index names that don't exist. 

Another safety feature -- we should make sure that any index name passed in is a global index and not a local one. Local indexes are unaffected by the new global index design.

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/index/IndexUpgradeTool.java
```diff
@@ -0,0 +1,550 @@
...
+                && !cmdLine.hasOption(LOG_FILE_OPTION.getOpt())) {
+            throw new IllegalStateException("Log file with "+TABLE_OPTION.getLongOpt()
+                    + " is mandatory if " + DRY_RUN_OPTION.getLongOpt() +" is passed");
+        }
+
+        if (!(cmdLine.hasOption(TABLE_OPTION.getOpt()))
+                && !(cmdLine.hasOption(TABLE_CSV_FILE_OPTION.getOpt()))) {
+            throw new IllegalStateException("Tables and indexes list should be passed in either with"
+                    +TABLE_OPTION.getLongOpt() + " or " + TABLE_CSV_FILE_OPTION.getLongOpt());
+        }
```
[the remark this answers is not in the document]
swaroopak: done

phoenix-core/src/it/java/org/apache/phoenix/end2end/IndexUpgradeToolIT.java
```diff
@@ -0,0 +1,303 @@
...
+        iut.setInputTables(null);
+        iut.setInputFile(INPUT_FILE);
+        iut.prepareToolSetup();
+        iut.executeTool();
+
+        postValidate();
+    }
+
+    @Test
+    public void testToolWithShowParameter() throws Exception {
```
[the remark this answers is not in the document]
swaroopak: done

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/index/IndexUpgradeTool.java
swaroopak: Done!

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/index/IndexUpgradeTool.java
swaroopak: done.

swaroopak: I am thinking about how to check if the view index is local or global. ll update the PR as per my finding. meanwhile, I fixed the rest of the comments on this thread.

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/index/IndexUpgradeTool.java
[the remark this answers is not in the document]
swaroopak: Since this tool is centered around changing the tableDescriptor for index and data table, one way of handling it could be to save older tableDescriptors before performing any operation and then in case of failure revert the table descriptors and some more atomic logic. @gjacoby126  how does this sound?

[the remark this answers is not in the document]
swaroopak: As discussed offline, making sure the tool is completely idempotent to handle this scenario.

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/index/IndexUpgradeTool.java
```diff
@@ -0,0 +1,529 @@
...
+    private void extractTablesAndIndexes(PhoenixConnection conn) {
+        String [] tables = inputTables.trim().split(",");
+        PTable dataTable = null;
+        try {
+            for (String tableName : tables) {
+                HashSet<String> physicalIndexes = new HashSet<>();
+                dataTable = PhoenixRuntime.getTableNoCache(conn, tableName);
+                String physicalTableName = dataTable.getPhysicalName().getString();
+
+                for (PTable indexTable : dataTable.getIndexes()) {
```
kadirozde: This feature is not supported on transactional tables. If a table is transactional, it should be skipped and this should be logged.

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/index/IndexUpgradeTool.java
```diff
@@ -0,0 +1,535 @@
...
+                                        "SELECT DISTINCT TABLE_NAME FROM " + "SYSTEM.CATALOG WHERE COLUMN_FAMILY = \'"
+                                                + viewIndexPhysicalName
+                                                + "\' AND TABLE_TYPE = \'i\'");
+                        while (rs.next()) {
+                            String viewIndexName = rs.getString(1);
+                            rebuildIndexes.add(viewIndexName);
+                        }
+                    }
+                    rebuildMap.put(physicalTableName, rebuildIndexes);
+                    tablesAndIndexes.put(physicalTableName, physicalIndexes);
```
gjacoby126: good, @swaroopak , I think that gets all the indexes.

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/index/IndexUpgradeTool.java
```diff
@@ -0,0 +1,535 @@
...
+        for(String index: rebuildMap.get(dataTable)) {
+            String outFile = "/tmp/index_rebuild_"+index+"_" + UUID.randomUUID().toString();
+            String indexName = SchemaUtil.getTableNameFromFullName(index);
+            conf.set(QueryServices.TRANSACTIONS_ENABLED, Boolean.TRUE.toString());
+            indexingTool.setConf(conf);
+            String [] args = {"-s", schema, "-dt", table, "-it", indexName, "-direct", "-runfg", "-op", outFile};
+            try {
+                LOGGER.info("Rebuilding index "+ index);
+                if(!dryRun) {
+                    indexingTool.run(args);
```
gjacoby126: Since this is running in the foreground, and some indexes take hours to rebuild, this can take a very long time. Should we be running in the background, or multi-threading? (That would of course be more complicated because you'd need callbacks or polling)

gjacoby126: @swaroopak - Since @kadirozde 's recent change to make unknown == UNVERIFIED, should the rebuild step be optional? I can see some operators wanting to separate the coprocessor upgrade process from the rebuild process.

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/index/IndexUpgradeTool.java
```diff
@@ -0,0 +1,535 @@
...
+                }
+            } catch (Exception e) {
+                LOGGER.severe("Something went wrong while building the index "+ index +" "+e);
```
gjacoby126: What should the larger process do if an index rebuild fails? Right now it's just eaten and logged, but seems like the tool would exit with a success.

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/index/IndexUpgradeTool.java
swaroopak: Done

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/index/IndexUpgradeTool.java
```diff
@@ -0,0 +1,535 @@
...
+    }
+
+
+    private void rebuildIndexes(String dataTable, Configuration conf) {
+        String schema = SchemaUtil.getSchemaNameFromFullName(dataTable);
+        String table = SchemaUtil.getTableNameFromFullName(dataTable);
+        for(String index: rebuildMap.get(dataTable)) {
+            String outFile = "/tmp/index_rebuild_"+index+"_" + UUID.randomUUID().toString();
+            String indexName = SchemaUtil.getTableNameFromFullName(index);
+            conf.set(QueryServices.TRANSACTIONS_ENABLED, Boolean.TRUE.toString());
```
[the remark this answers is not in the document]
swaroopak: Do we absolutely need to? Cause I saw IndexRebuildIT does. @kadirozde do you know?

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/index/IndexUpgradeTool.java
swaroopak: Since the new logic will only return global indexes on a view, I don't need to have this additional check -- resolving the comment.

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/index/IndexUpgradeTool.java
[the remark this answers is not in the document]
kadirozde: A test can freely change/set the cluster wide config params to test various scenarios. A tool should not need to set these parameters in general. In this case, the new secondary indexing design is for non-transactional tables so whether TRANSACTIONS_ENABLED is false or true does not matter.

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/index/IndexUpgradeTool.java
swaroopak: Will go with optionally rebuilding as sync or async :)

phoenix-core/src/it/java/org/apache/phoenix/end2end/IndexUpgradeToolIT.java
```diff
@@ -0,0 +1,348 @@
...
+                + "(view_column)");
+        conn.createStatement().execute("CREATE INDEX MOCK1_INDEX2 ON TEST.MOCK1_VIEW1 "
+                + "(zip)");
+        conn.createStatement().execute("CREATE INDEX MOCK2_INDEX1 ON TEST1.MOCK2_VIEW "
+                + "(state, city)");
+        conn.createStatement().execute("CREATE INDEX MOCK1_INDEX3 ON TEST.MOCK1_VIEW "
+                + "(view_column)");
+    }
+
+    @Parameters(name ="IndexUpgradeToolIT_mutable={0},upgrade={1},isNamespaceEnabled={2}")
```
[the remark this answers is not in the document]
swaroopak: Earlier it was taking 33 mins total, but now I made some changes -- running in dry mode for all tests but one. It runs within 16 mins.

phoenix-core/src/it/java/org/apache/phoenix/end2end/IndexUpgradeToolIT.java
```diff
@@ -0,0 +1,348 @@
...
+
+        conn.createStatement().execute("DROP TABLE TEST.MOCK1");
+        conn.createStatement().execute("DROP TABLE TEST1.MOCK2");
+        conn.createStatement().execute("DROP TABLE TEST.MOCK3");
+
+        if (isNamespaceEnabled) {
+            conn.createStatement().execute("DROP SCHEMA TEST");
+            conn.createStatement().execute("DROP SCHEMA TEST1");
+        }
+        conn.close();
```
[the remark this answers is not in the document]
swaroopak: Done

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/index/IndexUpgradeTool.java
swaroopak: exiting with -1 now.

phoenix-core/src/main/java/org/apache/phoenix/query/ConnectionQueryServicesImpl.java
```diff
@@ -1530,8 +1533,8 @@ private void ensureViewIndexTableCreated(byte[] physicalTableName, Map<String, O
         byte[] physicalIndexName = MetaDataUtil.getViewIndexPhysicalName(physicalTableName);
 
         tableProps.put(MetaDataUtil.IS_VIEW_INDEX_TABLE_PROP_NAME, TRUE_BYTES_AS_STRING);
-        TableDescriptor desc = ensureTableCreated(physicalIndexName, PTableType.TABLE, tableProps, families, splits,
-                false, isNamespaceMapped, false);
+        TableDescriptor desc = ensureTableCreated(physicalIndexName, PTableType.INDEX, tableProps, families, splits,
+                true, isNamespaceMapped, false);
```
swaroopak: @kadirozde This is fixing the problem I encountered. The fix makes sense to me. I ll create another jira for this.

phoenix-core/src/main/java/org/apache/phoenix/query/ConnectionQueryServicesImpl.java
```diff
@@ -916,6 +916,9 @@ private void addCoprocessors(byte[] tableName, TableDescriptorBuilder builder, P
                 } else if (indexRegionObserverEnabled && !newDesc.hasCoprocessor(GlobalIndexChecker.class.getName()) &&
                         !isLocalIndexTable(newDesc.getColumnFamilyNames())) {
                     builder.addCoprocessor(GlobalIndexChecker.class.getName(), null, priority - 1, null);
+                    if(newDesc.hasCoprocessor(IndexRegionObserver.class.getName())) {
+                        builder.removeCoprocessor(IndexRegionObserver.class.getName());
```
swaroopak: @kadirozde this was added to remove the data table coprocessor that gets added on an index table (it is similar to having indexer coproc (as per old design) added to all the tables by default). Shall we do this?

phoenix-core/src/main/java/org/apache/phoenix/query/ConnectionQueryServicesImpl.java
swaroopak: @kadirozde for your review https://github.com/apache/phoenix/pull/531

phoenix-core/src/it/java/org/apache/phoenix/end2end/ParameterizedIndexUpgradeToolIT.java
```diff
@@ -0,0 +1,351 @@
...
+                    Assert.assertFalse(admin.getDescriptor(TableName.valueOf(table))
+                            .hasCoprocessor(IndexRegionObserver.class.getName()));
+                }
+            }
+            for (String index : indexList) {
+                Assert.assertFalse(admin.getDescriptor(TableName.valueOf(index))
+                        .hasCoprocessor(GlobalIndexChecker.class.getName()));
+            }
+        }
+    }
```
kadirozde: Can we restructure the pre and post validate methods so that we eliminate repeated code? For example, we can have two more methods, checkNewIndexingCoprocessors() and checkOldIndexingCoprocessors() and call these methods to eliminate the repeated code.

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/index/IndexUpgradeTool.java
```diff
@@ -0,0 +1,547 @@
...
+                    LOGGER.fine("Data table is immutable, waiting for "+GLOBAL_INDEX_CHECKER_ENABLED_MAP_EXPIRATION_MIN + 1+" minutes for client cache to expire");
+                    if(!test) {
+                        Thread.sleep(
+                                (GLOBAL_INDEX_CHECKER_ENABLED_MAP_EXPIRATION_MIN + 1) * 60 * 1000);
+                    }
+                }
+                if (upgrade) {
+                    disableTable(admin, dataTableFullName, indexes);
+                    modifyIndexTable(admin, indexes);
+                    if (mutable) {
```
kadirozde: We should modify the coprocessors on the data table for immutable tables too as eventually we will remove the Indexer coprocessor from the code base.

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/index/IndexUpgradeTool.java
```diff
@@ -0,0 +1,547 @@
...
+                        modifyDataTable(admin, dataTableFullName);
+                    }
+                    enableTable(admin, dataTableFullName, indexes);
+                    rebuildIndexes(dataTableFullName, conf);
+                } else {
+                    disableTable(admin, dataTableFullName, indexes);
+                    if (mutable) {
```
kadirozde: As before, no need to check if table is mutable.

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/index/IndexUpgradeTool.java
```diff
@@ -0,0 +1,547 @@
...
+    }
+
+    private void extractTablesAndIndexes(PhoenixConnection conn) {
+        String [] tables = inputTables.trim().split(",");
+        PTable dataTable = null;
+        try {
+            for (String tableName : tables) {
+                HashSet<String> physicalIndexes = new HashSet<>();
+                dataTable = PhoenixRuntime.getTableNoCache(conn, tableName);
+                String physicalTableName = dataTable.getPhysicalName().getString();
```
kadirozde: We should check if the table is a data table and if not we should skip and log and continue. For example, if an index table is included in the inputTables, this should not be silently ignored.

phoenix-core/src/main/java/org/apache/phoenix/query/ConnectionQueryServicesImpl.java
kadirozde: Ok

phoenix-core/src/main/java/org/apache/phoenix/query/ConnectionQueryServicesImpl.java
kadirozde: Got it

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/index/IndexUpgradeTool.java
```diff
@@ -0,0 +1,553 @@
...
+            LOGGER.severe("Something went wrong "+e);
+            System.exit(-1);
+        }
+    }
+
+    @VisibleForTesting
+    public void executeTool() {
+        try {
+            Configuration conf = HBaseConfiguration.addHbaseResources(getConf());
+            Connection conn = ConnectionUtil.getInputConnection(conf);
```
priyankporwal: Where is this conn getting closed? Should it be used with a try-with-resources?

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/index/IndexUpgradeTool.java
```diff
@@ -0,0 +1,553 @@
...
+                    }
+                    rebuildMap.put(physicalTableName, rebuildIndexes);
+                    tablesAndIndexes.put(physicalTableName, physicalIndexes);
+                } else {
+                    LOGGER.info("Skipping Table " + tableName + " because it is "+
+                            (dataTable.isTransactional() ? "transactional" : "not a data table"));
+                }
+            }
+        } catch (SQLException e) {
+            LOGGER.severe("Exception while splitting tool input "+e);
```
priyankporwal: Nit: Exception message could be something like "Failed to find list of indexes"

phoenix-core/src/it/java/org/apache/phoenix/end2end/ParameterizedIndexUpgradeToolIT.java
```diff
@@ -0,0 +1,330 @@
...
+            conn.createStatement().execute("CREATE SCHEMA TEST1");
+        }
+        conn.createStatement().execute("CREATE TABLE TEST.MOCK1 (id bigint NOT NULL "
+                + "PRIMARY KEY, a.name varchar, sal bigint, address varchar)"+tableDDLOptions);
+        conn.createStatement().execute("CREATE TABLE TEST1.MOCK2 (id bigint NOT NULL "
+                + "PRIMARY KEY, name varchar, city varchar, phone bigint)"+tableDDLOptions);
+        conn.createStatement().execute("CREATE TABLE TEST.MOCK3 (id bigint NOT NULL "
+                + "PRIMARY KEY, name varchar, age bigint)"+tableDDLOptions);
+
+        //views
```
gjacoby126: nit: please mark "disabled until after PHOENIX-5385"

phoenix-core/src/it/java/org/apache/phoenix/end2end/ParameterizedIndexUpgradeToolIT.java
```diff
@@ -0,0 +1,330 @@
...
+//                + " state varchar) AS SELECT * FROM TEST1.MOCK2 WHERE name = 'c'");
+        //indexes
+        conn.createStatement().execute("CREATE INDEX INDEX1 ON TEST.MOCK1 (sal, a.name)");
+        conn.createStatement().execute("CREATE INDEX INDEX2 ON TEST.MOCK1 (a.name)");
+        conn.createStatement().execute("CREATE INDEX INDEX1 ON TEST1.MOCK2 (city)");
+        conn.createStatement().execute("CREATE INDEX INDEX2 ON TEST1.MOCK2 (phone)");
+        conn.createStatement().execute("CREATE INDEX INDEX3 ON TEST1.MOCK2 (name)");
+        conn.createStatement().execute("CREATE INDEX INDEX3 ON TEST.MOCK3 (age, name)");
+
+        //view-indexes
```
gjacoby126: ditto here

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/index/IndexUpgradeTool.java
```diff
@@ -0,0 +1,562 @@
...
+            String dataTableFullName = entry.getKey();
+            HashSet<String> indexes = entry.getValue();
+
+            try (Admin admin = queryServices.getAdmin()) {
+
+                PTable dataTable = PhoenixRuntime.getTableNoCache(conn, dataTableFullName);
+                LOGGER.fine("Executing " + operation + " for " + dataTableFullName);
+
+                boolean mutable = !(dataTable.isImmutableRows());
+                if (!mutable) {
```
gjacoby126: @swaroopak - anything that would prevent an operator from running multiple calls to the tool at once? (e.g, if I have 100 data tables, can I run this 10 times in parallel from the command line to get all tables done faster?)

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/index/IndexUpgradeTool.java
```diff
@@ -0,0 +1,562 @@
...
+        }
+    }
+
+    private void modifyDataTable(Admin admin, String tableName)
+            throws IOException {
+        TableDescriptorBuilder dataTableDescBuilder = TableDescriptorBuilder
+                .newBuilder(admin.getDescriptor(TableName.valueOf(tableName)));
+        if (upgrade) {
+            if (admin.getDescriptor(TableName.valueOf(tableName)).hasCoprocessor(Indexer.class.getName())) {
+                if (!dryRun) {
```
gjacoby126: Could simplify with helper methods for addCoprocessor(Class coprocessor) and removeCoprocessor(Class coprocessor) and then just have two lines for upgrade and two lines for rollback.

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/index/IndexUpgradeTool.java
```diff
@@ -0,0 +1,562 @@
...
+                        physicalIndexes.add(viewIndexPhysicalName);
+
+                        ResultSet
+                                rs =
+                                conn.createStatement().executeQuery(
+                                "SELECT DISTINCT TABLE_NAME FROM " + "SYSTEM.CATALOG WHERE COLUMN_FAMILY = \'"
+                                    + viewIndexPhysicalName +"\' AND TABLE_TYPE = \'i\'");
+                        while (rs.next()) {
+                            String viewIndexName = rs.getString(1);
+                            rebuildIndexes.add(viewIndexName);
```
gjacoby126: For tenant-owned indexes you'll need both the index name (which you have) _and_ the tenant id. Two tenants can create indexes with the same name, so to rebuild them you need to provide the tenant id that created them and now owns them.

phoenix-core/src/it/java/org/apache/phoenix/end2end/ParameterizedIndexUpgradeToolIT.java
swaroopak: I have added the global comment at the beginning of the test class. Should it be repeated?

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/index/IndexUpgradeTool.java
swaroopak: Yes, it should be able to run them in parallel. In fact, using a parallel stream was an easier way to do it than dealing with multiple threads. Do you think we should implement multithreading here?

gjacoby126: If you can easily add multi-threading, go for it. (Remember though that for the 4.x and 4.14.x branches you can't use Java 8's pretty parallel stream APIs)

phoenix-core/src/it/java/org/apache/phoenix/end2end/ParameterizedIndexUpgradeToolIT.java
gjacoby126: Ah, I missed that line. Since other people might miss it too, maybe move it to near the affected comments, rather than at the top?

swaroopak:
@priyankporwal @kadirozde @gjacoby126  Please have a final review. Hoping to close this by today.

PR #523, opened 2019-06-21
PHOENIX-5363: Phoenix parcel artifacts for CDH releases don't pass Cloudera verification tools test

m2je:
Parcel artifacts are created as `tar` files but they should be `tar.gz` files instead


stoty:
This a very old version, and the Apache project lo longer supports vendor Hbase versions.

PR #526, opened 2019-06-26
PHOENIX-5374: Incorrect exception thrown in some cases when client does not have Exec permissions on SYSTEM:CATALOG

ChinmaySKulkarni:


ChinmaySKulkarni:
Please review @abhishek-chouhan @twdsilva

twdsilva:
LGTM

phoenix-core/src/it/java/org/apache/phoenix/end2end/PermissionNSEnabledIT.java
```diff
@@ -50,4 +60,43 @@ public Void run() throws Exception {
             revokeAll();
         }
     }
+
+    @Test
+    public void testConnectionCreationFailsWhenNoExecPermsOnSystemCatalog() throws Throwable {
+        try {
+            grantSystemTableAccess();
+            superUser1.runAs((PrivilegedExceptionAction<Object>) () -> {
```
gjacoby126: Patch looks fine for master, but isn't the "->" lambda operator a Java 8 thing? If so you'll have to modify when backporting to 4.x, because from prior discussions on the dev list the community decided to keep 4.x compatible with JDK 7 because HBase 1.x is.

gjacoby126: Or alternatively do all versions in a JDK 7 compatible way

ChinmaySKulkarni: @gjacoby126 yup, I realized this when I tried applying it to 4.x branches and so I put up a different patch for those branches. Thanks!

gjacoby126:
JIRA has already been committed, closing PR.

PR #527, opened 2019-06-26
PHOENIX-5373 GlobalIndexChecker should treat the rows created by the …

kadirozde:
…previous design as unverified

### DIFF
--- a/phoenix-core/src/main/java/org/apache/phoenix/index/GlobalIndexChecker.java
+++ b/phoenix-core/src/main/java/org/apache/phoenix/index/GlobalIndexChecker.java
@@ -21,7 +21,7 @@ import static org.apache.phoenix.coprocessor.BaseScannerRegionObserver.CHECK_VER
 import static org.apache.phoenix.coprocessor.BaseScannerRegionObserver.EMPTY_COLUMN_FAMILY_NAME;
 import static org.apache.phoenix.coprocessor.BaseScannerRegionObserver.EMPTY_COLUMN_QUALIFIER_NAME;
 import static org.apache.phoenix.coprocessor.BaseScannerRegionObserver.PHYSICAL_DATA_TABLE_NAME;
-import static org.apache.phoenix.hbase.index.IndexRegionObserver.UNVERIFIED_BYTES;
+import static org.apache.phoenix.hbase.index.IndexRegionObserver.VERIFIED_BYTES;
 import static org.apache.phoenix.index.IndexMaintainer.getIndexMaintainer;
 import static org.apache.phoenix.schema.types.PDataType.TRUE_BYTES;
 
@@ -33,7 +33,9 @@ import java.util.Optional;
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
 import org.apache.hadoop.hbase.Cell;
+import org.apache.hadoop.hbase.CoprocessorEnvironment;
 import org.apache.hadoop.hbase.DoNotRetryIOException;
+import org.apache.hadoop.hbase.TableName;
 import org.apache.hadoop.hbase.client.Delete;
 import org.apache.hadoop.hbase.client.Get;
 import org.apache.hadoop.hbase.client.Mutation;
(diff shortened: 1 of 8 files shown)


phoenix-core/src/main/java/org/apache/phoenix/index/GlobalIndexChecker.java
```diff
@@ -315,7 +320,7 @@ private boolean verifyRowAndRemoveEmptyColumn(List<Cell> cellList) throws IOExce
                     // we consider anything that is not UNVERIFIED means VERIFIED. IndexTool should be used to
```
priyankporwal: These comment needs to be changed too to reflect new behavior.

phoenix-core/src/main/java/org/apache/phoenix/index/GlobalIndexChecker.java
```diff
@@ -212,8 +217,7 @@ private void repairIndexRows(byte[] indexRowKey, long ts, List<Cell> row) throws
                 indexScan = new Scan(scan);
                 byte[] dataTableName = scan.getAttribute(PHYSICAL_DATA_TABLE_NAME);
                 byte[] indexTableName = region.getRegionInfo().getTable().getName();
-                dataHTable = ServerUtil.getHTableForCoprocessorScan(env,
-                        SchemaUtil.getPhysicalTableName(dataTableName, env.getConfiguration()));
+                dataHTable = ServerUtil.ConnectionFactory.getConnection(ServerUtil.ConnectionType.DEFAULT_SERVER_CONNECTION, env).getTable(TableName.valueOf(dataTableName));
```
priyankporwal: Nit: Line is too long

phoenix-core/src/main/java/org/apache/phoenix/index/GlobalIndexChecker.java
kadirozde: Will remove the comments

phoenix-core/src/main/java/org/apache/phoenix/index/GlobalIndexChecker.java
kadirozde: Will break into two lines.

phoenix-core/src/main/java/org/apache/phoenix/index/GlobalIndexChecker.java
```diff
@@ -237,6 +241,7 @@ private void repairIndexRows(byte[] indexRowKey, long ts, List<Cell> row) throws
             }
             // Rebuild the index rows from the corresponding the rows in the the data table
             byte[] dataRowKey = indexMaintainer.buildDataRowKey(new ImmutableBytesWritable(indexRowKey), viewConstants);
+            LOG.debug("repairIndexRows : " + Bytes.toStringBinary(dataRowKey) + " --> " + Bytes.toStringBinary(indexRowKey));
```
gjacoby126: Log should be at TRACE, both because it could be quite verbose, and because some operators have restrictions about logging customer data, which can be contained in row keys.

kadirozde: I will remove the log. It was not really important to have it. We can have metrics instead.

phoenix-core/src/main/java/org/apache/phoenix/index/GlobalIndexChecker.java
kadirozde: Done

phoenix-core/src/main/java/org/apache/phoenix/index/GlobalIndexChecker.java
kadirozde: Done

phoenix-core/src/main/java/org/apache/phoenix/index/GlobalIndexChecker.java
kadirozde: Done

PR #528, opened 2019-06-28
PHOENIX-5379 : Avoid possible NPE while closing CSVParser

virajjasani:
CSVFileResultHandler closes CSVParser which could cause NPE if creating CSVParser causes IOException and original IOException can be lost


virajjasani:
Please review @twdsilva

PR #531, opened 2019-07-05
PHOENIX-5385: GlobalIndexChecker coproc doesn't load on view indexes

swaroopak:
…E; does not load IndexCheckerCoproc

### DIFF
--- a/phoenix-core/src/it/java/org/apache/phoenix/end2end/index/DropColumnIT.java
+++ b/phoenix-core/src/it/java/org/apache/phoenix/end2end/index/DropColumnIT.java
@@ -34,6 +34,7 @@ import java.util.Collection;
 import java.util.Properties;
 
 import org.apache.hadoop.hbase.KeyValue;
+import org.apache.hadoop.hbase.TableName;
 import org.apache.hadoop.hbase.client.HTable;
 import org.apache.hadoop.hbase.client.Result;
 import org.apache.hadoop.hbase.client.ResultScanner;
@@ -342,8 +343,6 @@ public class DropColumnIT extends ParallelStatsDisabledIT {
             PhoenixConnection pconn = conn.unwrap(PhoenixConnection.class);
             PTable dataTable = pconn.getTable(new PTableKey(null, dataTableFullName));
             assertEquals("Unexpected number of indexes ", 3, dataTable.getIndexes().size());
-            PTable indexTable = dataTable.getIndexes().get(0);
-            byte[] indexTablePhysicalName = indexTable.getPhysicalName().getBytes();
             PName localIndexTablePhysicalName = dataTable.getIndexes().get(1).getPhysicalName();
             
             // drop v2 which causes the regular index and first local index to be dropped
@@ -373,9 +372,10 @@ public class DropColumnIT extends ParallelStatsDisabledIT {
             
             // verify that the regular index physical table was dropped
             try {
-                conn.unwrap(PhoenixConnection.class).getQueryServices().getTableDescriptor(indexTablePhysicalName);
+                conn.unwrap(PhoenixConnection.class).getQueryServices().getAdmin()
+                        .getTableDescriptor(TableName.valueOf(indexTableName));
                 fail("Index table should have been dropped");
-            } catch (TableNotFoundException e) {
+            } catch (org.apache.hadoop.hbase.TableNotFoundException e) {
             }
             
             // verify that the local index physical table was *not* dropped
--- a/phoenix-core/src/main/java/org/apache/phoenix/execute/MutationState.java
+++ b/phoenix-core/src/main/java/org/apache/phoenix/execute/MutationState.java
@@ -83,6 +83,8 @@ import org.apache.phoenix.schema.PColumn;
 import org.apache.phoenix.schema.PIndexState;
 import org.apache.phoenix.schema.PMetaData;
 import org.apache.phoenix.schema.PName;
+import org.apache.phoenix.schema.PNameFactory;
+import org.apache.phoenix.schema.PNameImpl;
 import org.apache.phoenix.schema.PRow;
 import org.apache.phoenix.schema.PTable;
 import org.apache.phoenix.schema.PTableRef;
@@ -103,6 +105,7 @@ import org.apache.phoenix.util.EnvironmentEdgeManager;
 import org.apache.phoenix.util.IndexUtil;
 import org.apache.phoenix.util.KeyValueUtil;
 import org.apache.phoenix.util.LogUtil;
+import org.apache.phoenix.util.MetaDataUtil;
 import org.apache.phoenix.util.PhoenixRuntime;
 import org.apache.phoenix.util.SQLCloseable;
 import org.apache.phoenix.util.SchemaUtil;
@@ -507,7 +510,7 @@ public class MutationState implements SQLCloseable {
         return ptr;
     }
 
-    private Iterator<Pair<PName, List<Mutation>>> addRowMutations(final TableRef tableRef,
+    private Iterator<Pair<PTable, List<Mutation>>> addRowMutations(final TableRef tableRef,
             final MultiRowMutationState values, final long mutationTimestamp, final long serverTimestamp,
             boolean includeAllIndexes, final boolean sendAll) {
         final PTable table = tableRef.getTable();
@@ -520,7 +523,7 @@ public class MutationState implements SQLCloseable {
                 .<Mutation> newArrayListWithExpectedSize(values.size()) : null;
         generateMutations(tableRef, mutationTimestamp, serverTimestamp, values, mutationList,
                 mutationsPertainingToIndex);
-        return new Iterator<Pair<PName, List<Mutation>>>() {
+        return new Iterator<Pair<PTable, List<Mutation>>>() {
             boolean isFirst = true;
             Map<byte[], List<Mutation>> indexMutationsMap = null;
 
@@ -530,10 +533,10 @@ public class MutationState implements SQLCloseable {
             }
 
             @Override
-            public Pair<PName, List<Mutation>> next() {
+            public Pair<PTable, List<Mutation>> next() {
                 if (isFirst) {
                     isFirst = false;
-                    return new Pair<>(table.getPhysicalName(), mutationList);
+                    return new Pair<>(table, mutationList);
                 }
 
                 PTable index = indexes.next();
@@ -585,7 +588,7 @@ public class MutationState implements SQLCloseable {
                 } catch (SQLException | IOException e) {
                     throw new IllegalDataException(e);
                 }
-                return new Pair<PName, List<Mutation>>(index.getPhysicalName(),
+                return new Pair<PTable, List<Mutation>>(index,
                         indexMutations == null ? Collections.<Mutation> emptyList() : indexMutations);
             }
 
@@ -690,8 +693,9 @@ public class MutationState implements SQLCloseable {
             private Iterator<Pair<byte[], List<Mutation>>> innerIterator = init();
 
             private Iterator<Pair<byte[], List<Mutation>>> init() {
-                final Iterator<Pair<PName, List<Mutation>>> mutationIterator = addRowMutations(current.getKey(),
-                        current.getValue(), mutationTimestamp, serverTimestamp, includeMutableIndexes, true);
+                final Iterator<Pair<PTable, List<Mutation>>> mutationIterator =
+                        addRowMutations(current.getKey(), current.getValue(), mutationTimestamp,
+                                serverTimestamp, includeMutableIndexes, true);
                 return new Iterator<Pair<byte[], List<Mutation>>>() {
                     @Override
                     public boolean hasNext() {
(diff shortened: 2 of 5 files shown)


gjacoby126:
This change has some broader implications -- PTableType is used in JDBC metadata API queries, and changing the PTableType from TABLE to INDEX would mean that the view index table would show up as an index in calls to DatabaseMetadata.getIndexInfo(), or DatabaseMetadata.getTables() that filters on INDEX type. 

But it's not _really_ an index, just a physical container in which other indexes may or may not live. Really, they should be their own PTableType, but because the enum's used on both client and server side, adding an enum value would be really tricky. 

This is also why this small change caused a good number of test failures.

swaroopak:
Thank you @gjacoby126 and @kadirozde. Yes, I have been investigating the test failures. Was able to find the root cause as all failed tests had an immutable base table and the same point of failure in case of view indexes. working on a better approach to tackle this problem.

phoenix-core/src/main/java/org/apache/phoenix/execute/MutationState.java
```diff
@@ -1199,7 +1201,27 @@ private void filterIndexCheckerMutations(Map<TableInfo, List<Mutation>> mutation
             Entry<TableInfo, List<Mutation>> pair = mapIter.next();
             TableInfo tableInfo = pair.getKey();
             if (tableInfo.getOrigTableRef().getTable().isImmutableRows() && IndexUtil.isGlobalIndexCheckerEnabled(connection, tableInfo.getHTableName())) {
-                PTable table = PhoenixRuntime.getTable(connection, tableInfo.getHTableName().getString());
+                HTableDescriptor desc = null;
+                PTable table;
+                try {
+                    desc = connection.getQueryServices().getAdmin().getTableDescriptor(
```
gjacoby126: @swaroopak  This makes a remote call to the HMaster. Probably need to either cache table descriptors or find another way. Is the naming convention on the table not sufficient?

gjacoby126: What about MetaDataUtil.isViewIndex?

phoenix-core/src/it/java/org/apache/phoenix/end2end/index/DropColumnIT.java
```diff
@@ -373,9 +373,9 @@ public void testDroppingIndexedColDropsIndex() throws Exception {
             
             // verify that the regular index physical table was dropped
             try {
-                conn.unwrap(PhoenixConnection.class).getQueryServices().getTableDescriptor(indexTablePhysicalName);
+                conn.unwrap(PhoenixConnection.class).getQueryServices().getAdmin().getTableDescriptor(indexTablePhysicalName);
                 fail("Index table should have been dropped");
-            } catch (TableNotFoundException e) {
+            } catch (org.apache.hadoop.hbase.TableNotFoundException e) {
```
swaroopak: At this point the problem arose due to table name is being cached in tableCache variable at HBase level. Could not find a way to invalidate it, hence the RPC.  any suggestion?

phoenix-core/src/main/java/org/apache/phoenix/execute/MutationState.java
```diff
@@ -1199,7 +1201,20 @@ private void filterIndexCheckerMutations(Map<TableInfo, List<Mutation>> mutation
...
-                PTable table = PhoenixRuntime.getTable(connection, tableInfo.getHTableName().getString());
+                PTable table;
+                if (MetaDataUtil.isViewIndex(tableInfo.getHTableName().getString())) {
+                    PTable origLogicalTable = tableInfo.getOrigTableRef().getTable();
+                    if (origLogicalTable.getTenantId() == null) {
+                        //Global view index
+                        table = origLogicalTable.getIndexes().get(0);
+                    } else {
+                        //tenant specific view index
+                        PTable view = PhoenixRuntime.getTableNoCache(connection, origLogicalTable.getParentName().getString());
```
gjacoby126: for a tenant-specific view index you'll either need a tenant connection to do the PTable lookup, or to use the new API @abhishek-chouhan created in PHOENIX-5180. (You'd have to backport PHOENIX-5180 to 4.14.3 if you go that way, as it's currently just in 4.15 and 5.1)

phoenix-core/src/it/java/org/apache/phoenix/end2end/index/DropColumnIT.java
gjacoby126: There's no problem with a single RPC in a test like here, just in code that gets called every commit like previously in MutationState below :-)

phoenix-core/src/main/java/org/apache/phoenix/execute/MutationState.java
gjacoby126: Or is connection already the tenant-specific connection here?

gjacoby126: Yeah, looks like it's already the tenant connection. Never mind.

phoenix-core/src/main/java/org/apache/phoenix/execute/MutationState.java
```diff
@@ -1199,7 +1201,20 @@ private void filterIndexCheckerMutations(Map<TableInfo, List<Mutation>> mutation
             Entry<TableInfo, List<Mutation>> pair = mapIter.next();
             TableInfo tableInfo = pair.getKey();
             if (tableInfo.getOrigTableRef().getTable().isImmutableRows() && IndexUtil.isGlobalIndexCheckerEnabled(connection, tableInfo.getHTableName())) {
-                PTable table = PhoenixRuntime.getTable(connection, tableInfo.getHTableName().getString());
+                PTable table;
+                if (MetaDataUtil.isViewIndex(tableInfo.getHTableName().getString())) {
+                    PTable origLogicalTable = tableInfo.getOrigTableRef().getTable();
```
swaroopak: can be moved outside of if at line 1203

phoenix-core/src/main/java/org/apache/phoenix/execute/MutationState.java
```diff
@@ -1199,7 +1201,20 @@ private void filterIndexCheckerMutations(Map<TableInfo, List<Mutation>> mutation
...
+                    if (origLogicalTable.getTenantId() == null) {
+                        //Global view index
+                        table = origLogicalTable.getIndexes().get(0);
```
twdsilva: Does it matter if there is more than one index on the view?

phoenix-core/src/main/java/org/apache/phoenix/query/ConnectionQueryServicesImpl.java
```diff
@@ -904,12 +904,15 @@ private void addCoprocessors(byte[] tableName, HTableDescriptor descriptor, PTab
...
 
-            if (tableType == PTableType.INDEX && !isTransactional) {
+            if ((tableType == PTableType.INDEX && !isTransactional) || TRUE_BYTES_AS_STRING.equals(tableProps.get(MetaDataUtil.IS_VIEW_INDEX_TABLE_PROP_NAME))) {
                 if (!indexRegionObserverEnabled && descriptor.hasCoprocessor(GlobalIndexChecker.class.getName())) {
                     descriptor.removeCoprocessor(GlobalIndexChecker.class.getName());
                 } else if (indexRegionObserverEnabled && !descriptor.hasCoprocessor(GlobalIndexChecker.class.getName()) &&
                         !isLocalIndexTable(descriptor.getFamiliesKeys())) {
                     descriptor.addCoprocessor(GlobalIndexChecker.class.getName(), null, priority - 1, null);
+                    if (descriptor.hasCoprocessor(IndexRegionObserver.class.getName())) {
+                        descriptor.removeCoprocessor(IndexRegionObserver.class.getName());
```
twdsilva: Whats the relationship between IndexRegionObserver and GlobalIndexChecker?

swaroopak: IndexRegionObserver is loaded on dataTable and GlobalIndexChecker is loaded on indexTable

phoenix-core/src/main/java/org/apache/phoenix/execute/MutationState.java
swaroopak: I modified to use MetaDataUtil method to get the logical table.

phoenix-core/src/main/java/org/apache/phoenix/query/ConnectionQueryServicesImpl.java
```diff
@@ -958,10 +958,6 @@ private void addCoprocessors(byte[] tableName, HTableDescriptor descriptor, PTab
                             opts.put(NonTxIndexBuilder.CODEC_CLASS_NAME_KEY, PhoenixIndexCodec.class.getName());
                             IndexRegionObserver.enableIndexing(descriptor, PhoenixIndexBuilder.class, opts, priority);
                         }
-                        if (descriptor.hasCoprocessor(Indexer.class.getName())) {
```
swaroopak: @kadirozde Looks like the addendum patch didn't make it to 4.x branches? removing the repeated lines in my patch.

phoenix-core/src/main/java/org/apache/phoenix/execute/MutationState.java
```diff
@@ -894,6 +897,7 @@ private void send(Iterator<TableRef> tableRefIterator) throws SQLException {
 
         MultiRowMutationState multiRowMutationState;
         Map<TableInfo, List<Mutation>> physicalTableMutationMap = Maps.newLinkedHashMap();
+        Map<TableInfo, TableRef> physicalLogicalMap = Maps.newLinkedHashMap();
```
priyankporwal: Nit: Please rename this to logicalToPhysicalMap .. this one is easy to be read as 'physicalToLogicalMap' and being passed to filterIndexCheckerMutations() where logicalToPhysicalMap is expected.

swaroopak: Actually, it's physical to Logical, I renamed it accordingly. :)

swaroopak:
Thank you for the review @priyankporwal.
@gjacoby126  @twdsilva Could you please take a look?

phoenix-core/src/main/java/org/apache/phoenix/execute/MutationState.java
```diff
@@ -837,14 +838,18 @@ public long getBatchCount() {
         private final PName hTableName;
         @Nonnull
         private final TableRef origTableRef;
+        //This is a phoenix/logical table name
+        private String pTableFullName;
```
kadirozde: Should not this be of type PName instead of String in this context to be compatible with the rest of the implementation?

swaroopak: Should be fine, this class is only used in this file. Didn't involve any broader implications.

kadirozde: The question is then why is hTableName of type PName whereas pTableFullName is of type String? Please note that PName is designed to be used as the name type for objects in Phoenix (Phoenix Name for short). So, if you do not have any specific reason to use String then you should use PName, especially here for consistency. The type for pTableFullName is visible via the constructor method, so it is not internal to this class.

swaroopak: Okay.

phoenix-core/src/main/java/org/apache/phoenix/execute/MutationState.java
```diff
@@ -916,17 +928,21 @@ private void send(Iterator<TableRef> tableRefIterator) throws SQLException {
...
-                    Pair<PName, List<Mutation>> pair = mutationsIterator.next();
-                    PName hTableName = pair.getFirst();
+                    Pair<PTable, List<Mutation>> pair = mutationsIterator.next();
+                    PTable logicalTable = pair.getFirst();
                     List<Mutation> mutationList = pair.getSecond();
-                    TableInfo tableInfo = new TableInfo(isDataTable, hTableName, tableRef);
+                    PName logicalFullName = PNameFactory.newName(SchemaUtil.getTableName(
+                            logicalTable.getSchemaName().getString(),
+                            logicalTable.getTableName().getString()));
+                    TableInfo tableInfo = new TableInfo(isDataTable, logicalTable.getPhysicalName()
```
kadirozde: So, we get the table name and schema from the PTable object in order to form a PName object and store the  reference to the PName object in a TableInfo object. Then later we get the PName object from the TableInfo object in order to get the PTable object. Should we store the PTable object reference in the TableInfo object, instead of the PName object reference to eliminate this back-and_forth? @gjacoby126, do you have any comment on this? I think we can even eliminate the PTable Info object and use PTable instead. I am not suggesting that we should do it now but in a separate jira.

kadirozde: I was hopping that you would do the first suggestion which is to replace PName with PTable in this patch.

PR #535, opened 2019-07-08
PHOENIX-5358: Metrics for the GlobalIndexChecker coprocessor

priyankporwal:


phoenix-core/src/main/java/org/apache/phoenix/hbase/index/metrics/IndexCheckerSource.java
```diff
@@ -0,0 +1,67 @@
...
+    // Metrics2 and JMX constants
+    String METRICS_NAME = "PhoenixIndexChecker";
+    String METRICS_CONTEXT = "phoenix";
+    String METRICS_DESCRIPTION = "Metrics about the Phoenix Index Checker";
+    String METRICS_JMX_CONTEXT = "RegionServer,sub=" + METRICS_NAME;
+
+    String INDEX_REPAIR = "indexRepairs";
+    String INDEX_REPAIR_DESC = "The number of index row repairs";
+
+    String INDEX_REPAIR_FAILURE = "indexRepairs";
```
priyankporwal: Fixing this typo to "indexRepairFailures"

phoenix-core/src/main/java/org/apache/phoenix/hbase/index/metrics/IndexCheckerSource.java
```diff
@@ -0,0 +1,67 @@
...
+package org.apache.phoenix.hbase.index.metrics;
+
+import org.apache.hadoop.hbase.metrics.BaseSource;
+
+/**
+ * Interface for metrics from GlobalIndexChecker
+ */
+public interface IndexCheckerSource extends BaseSource {
+    // Metrics2 and JMX constants
+    String METRICS_NAME = "PhoenixIndexChecker";
```
kadirozde: Shouldn't the interface and metrics name be GlobalIndexCheckerSource and PhoenixGlobalIndexChecker as the metrics is not for local indexes.

phoenix-core/src/main/java/org/apache/phoenix/hbase/index/metrics/IndexCheckerSource.java
```diff
@@ -0,0 +1,67 @@
...
+    String METRICS_CONTEXT = "phoenix";
+    String METRICS_DESCRIPTION = "Metrics about the Phoenix Index Checker";
```
kadirozde: Similarly this is about the Phoenix Global Index Checker

phoenix-core/src/main/java/org/apache/phoenix/index/GlobalIndexChecker.java
```diff
@@ -108,6 +107,8 @@ public GlobalIndexScanner(RegionCoprocessorEnvironment env, Scan scan, RegionSca
                     QueryServicesOptions.DEFAULT_GLOBAL_INDEX_REPAIR_COUNT);
             minTimestamp = scan.getTimeRange().getMin();
             maxTimestamp = scan.getTimeRange().getMax();
+
+            metricsSource = MetricsIndexerSourceFactory.getInstance().createForIndexChecker();
```
kadirozde: This should be a member of GlobalIndexChecker (not GlobalIndexScanner).

priyankporwal: It can be. However, currently GlobalIndexChecker has no member fields and entirity of the logic is in GlobalIndexScanner. Also, since the scanner object does not refer back to the checker, the metricsSource would have to be passed via constructor and still be a member of the scanner. Let me know if you feel strongly, and I will do the change.

kadirozde: It does not have a member field in the master branch but has it in the 4.x branches. You do not need a constructor for a RegionObserver but need to implement the start method where you can initialize it. You can just follow what IndexRegionObsever does. Please note that GlobalIndexChecker object is created when the table is loaded whereas GlobalIndexScanner is created for every scan. This is the reason I suggested to move the field to the GlobalIndexChecker class.

priyankporwal: Got it. Will do as suggested. Thanks for providing context!

phoenix-core/src/main/java/org/apache/phoenix/hbase/index/metrics/GlobalIndexCheckerSource.java
```diff
@@ -0,0 +1,67 @@
...
+
+import org.apache.hadoop.hbase.metrics.BaseSource;
+
+/**
+ * Interface for metrics from GlobalIndexChecker
+ */
+public interface GlobalIndexCheckerSource extends BaseSource {
+    // Metrics2 and JMX constants
+    String METRICS_NAME = "GlobalIndexChecker";
+    String METRICS_CONTEXT = "phoenix";
```
swaroopak: Would it be better to be more specific something like phoenixSecondaryIndex?

priyankporwal: Not sure. I just want to avoid repeating 'phoenix' and 'index' everywhere and make the fully qualified metric name extremely long. 'secondary' is sort of implied everywhere when we say index, right?

phoenix-core/src/main/java/org/apache/phoenix/index/GlobalIndexChecker.java
```diff
@@ -356,11 +362,23 @@ private boolean verifyRowAndRepairIfNecessary(List<Cell> cellList) throws IOExce
...
                 byte[] rowKey = new byte[cell.getRowLength()];
                 System.arraycopy(cell.getRowArray(), cell.getRowOffset(), rowKey, 0, cell.getRowLength());
                 long ts = getMaxTimestamp(cellList);
                 cellList.clear();
-                repairIndexRows(rowKey, ts, cellList);
+
+                try {
+                    repairIndexRows(rowKey, ts, cellList);
+                    metricsSource.incrementIndexRepairs();
+                    metricsSource.updateIndexRepairTime(EnvironmentEdgeManager.currentTimeMillis() - repairStart);
```
swaroopak: shall we update the repair time before incrementing the index repairs?

phoenix-core/src/main/java/org/apache/phoenix/index/GlobalIndexChecker.java
```diff
@@ -356,11 +362,23 @@ private boolean verifyRowAndRepairIfNecessary(List<Cell> cellList) throws IOExce
...
+                } catch (IOException e) {
+                    metricsSource.incrementIndexRepairFailures();
+                    metricsSource.updateIndexRepairFailureTime(EnvironmentEdgeManager.currentTimeMillis() - repairStart);
```
swaroopak: same as above

phoenix-core/src/main/java/org/apache/phoenix/index/GlobalIndexChecker.java
priyankporwal: Shouldn't matter. The above increment is likely a single instruction and would take a few thousandths of micro-sec to impact our next ms level metric. If that is not the case, we should be very sad that our metric collection framework itself skews the metrics :(

phoenix-core/src/main/java/org/apache/phoenix/hbase/index/metrics/GlobalIndexCheckerSource.java
priyankporwal: I'll confirm once I get a test run and see the metrics from JMX port. We can decide accordingly.

PR #540, opened 2019-07-12
PHOENIX-4878 Remove SharedTableState and replace with PTable

yanxinyi:


twdsilva:
Thanks for the patch @yanxinyi . Looks pretty good, can you please create a patch and submit it on the JIRA so that we can get a a test run? Also can you file a follow up JIRA to remove SharedTableState in two releases (4.17) ?

yanxinyi:
rebase from the master

PR #547, opened 2019-07-22
PHOENIX-5401 Error in running apache.phoenix.mapreduce.index.automati…

kadirozde:
…on.PhoenixMRJobSubmitter


phoenix-core/src/main/java/org/apache/phoenix/mapreduce/PhoenixServerBuildIndexInputFormat.java
```diff
@@ -99,9 +98,6 @@ protected  QueryPlan getQueryPlan(final JobContext context, final Configuration
             if (txnScnValue != null) {
                 scan.setAttribute(BaseScannerRegionObserver.TX_SCN, Bytes.toBytes(Long.valueOf(txnScnValue)));
             }
-
-            // Initialize the query plan so it sets up the parallel scans
-            queryPlan.iterator(MapReduceParallelScanGrouper.getInstance());
```
priyankporwal: Why is this removed @kadirozde?

kadirozde: The same step is done again after the calling this method so it is not necessary to do here. Please see the super class where this method is called. You can also looked at the default implementation of this method which does not this step and the other step that I removed above this.

swaroopak:
Thank you @kadirozde  for the patch. LGTM, +1

PR #548, opened 2019-07-22
PHOENIX-5104: breaks client backwards compatibility

m2je:
Fixing by adding a new connection property "phoenix.index.longViewIndex.enabled" which is by default false.
Clients can set this property to true to be able to use longViewIndexes.


chrajeshbabu:
+1 @m2je. Thanks.

twdsilva:
+1

m2je:
Merged

PR #550, opened 2019-07-25
PHOENIX-5348: Fix flaky test: testIndexRebuildTask

gokceni:


gokceni:
@lhofhansl @ChinmaySKulkarni

phoenix-core/src/it/java/org/apache/phoenix/end2end/DropTableWithViewsIT.java
```diff
@@ -166,9 +168,14 @@ public static void assertTaskColumns(Connection conn, String expectedStatus, PTa
...
 
+        if (expectedTableName != null) {
+            String tableName = rs.getString(PhoenixDatabaseMetaData.TABLE_NAME);
+            assertEquals(expectedTableName, tableName);
+        }
+
         if (expectedData != null) {
             String data = rs.getString(PhoenixDatabaseMetaData.TASK_DATA);
-            assertEquals(expectedData, data);
+            assertTrue(data.contains(expectedData));
```
ChinmaySKulkarni: I ran `DropTableWithViewsIT` and it looks like `data` just gets the value `{"TaskDetails":"COMPLETED"}`, whereas when running `IndexMetadataIT`, it gets a json such as `{"IndexName":"R_ASYNCIND_N000002","RebuildAll":true,"JobID":"job_local1895022878_0001","TaskDetails":"COMPLETED"}`. 

To make this check more accurate, can we pass in a key and expected value and parse the data accordingly? Just checking `contains` may cause us to miss actual failures or even lead to false alarms if we change values stored in the `TASK_DATA`.

phoenix-core/src/it/java/org/apache/phoenix/end2end/IndexRebuildTaskIT.java
```diff
@@ -129,43 +128,48 @@ public void testIndexRebuildTask() throws Throwable {
...
             Admin admin = queryServices.getAdmin();
             TableName tableName = TableName.valueOf(viewIndexTableName);
             admin.disableTable(tableName);
-            admin.truncateTable(tableName, false);
+            admin.truncateTable(tableName, true);
 
             data = "{IndexName:" + indexName + ", DisableBefore:true}";
 
-            // Add a new task (update status to created) to System.Task to rebuild indexes
+            // Add a new task to System.Task to rebuild indexes
```
ChinmaySKulkarni: Can you move this comment back to the line just before we `addTask`?

phoenix-core/src/it/java/org/apache/phoenix/end2end/IndexRebuildTaskIT.java
```diff
@@ -129,43 +128,48 @@ public void testIndexRebuildTask() throws Throwable {
...
+            conn.createStatement().execute("DELETE " + " FROM " + PhoenixDatabaseMetaData.SYSTEM_TASK_NAME
+                    + " WHERE TABLE_NAME ='" + viewName  + "'");
+            conn.commit();
```
ChinmaySKulkarni: I guess we are doing this to remove any possible tasks that may already be present in SYSTEM.TASK for the `viewName`. If so, can you please add a comment here about this?

phoenix-core/src/it/java/org/apache/phoenix/end2end/IndexRebuildTaskIT.java
```diff
@@ -129,43 +128,48 @@ public void testIndexRebuildTask() throws Throwable {
...
+            Thread.sleep(1);
```
ChinmaySKulkarni: Instead of sleeping without checking, perhaps query SYSTEM.TASK to check that the tasks for that `viewName` have been successfully deleted, in a loop bounded by some number of retries. Then continue if all tasks have been deleted. If some tasks still remain even after said number of retries, we can perhaps fail the test since I believe this would eventually fail the test anyways due to the leftover tasks for that `viewName`, right?

phoenix-core/src/it/java/org/apache/phoenix/end2end/IndexRebuildTaskIT.java
```diff
@@ -129,43 +128,48 @@ public void testIndexRebuildTask() throws Throwable {
...
+            waitForTaskState(conn, PTable.TaskType.INDEX_REBUILD, viewName, PTable.TaskStatus.COMPLETED);
 
             // See that index is rebuilt and confirm index has rows
             Table htable= queryServices.getTable(Bytes.toBytes(viewIndexTableName));
             count = getUtility().countRows(htable);
             assertEquals(numOfValues, count);
         } finally {
-            conn.createStatement().execute("DELETE " + " FROM " + PhoenixDatabaseMetaData.SYSTEM_TASK_NAME);
+            conn.createStatement().execute("DELETE " + " FROM " + PhoenixDatabaseMetaData.SYSTEM_TASK_NAME
+            + " WHERE TABLE_NAME ='" + viewName  + "'");
```
ChinmaySKulkarni: This should all be inside the `if(conn != null)` block. Also, instead of doing this in a finally block, does it make sense to have an `@After` method to clean up the SYSTEM.TASK table? That way, if we add more tests here, we don't have to repeat this code.

phoenix-core/src/it/java/org/apache/phoenix/end2end/DropTableWithViewsIT.java
gokceni: Key is a bit complicated:
PRIMARY KEY (" + TASK_TYPE + "," + TASK_TS + " ROW_TIMESTAMP," + TENANT_ID + "," + TABLE_SCHEM + "," +
            TABLE_NAME + "))
I will try to move to key

ChinmaySKulkarni: I meant, key="IndexName" and value="<name to find" which we then parse from the `TASK_DATA` string, rather than just doing a `contains` within the entire string. I wasn't talking about the primary key, sorry for the confusion.

phoenix-core/src/it/java/org/apache/phoenix/end2end/IndexRebuildTaskIT.java
gokceni: Actually, we don't need to remove this. I realized that Lars changed connection settings to autocommit=false. When I commit after setting status to Created, I don't need to delete the task.

ChinmaySKulkarni: @gokceni I thought the point of this was to remove any possible leftover tasks from any previous operations, so that we ensure that all future checks only correspond to tasks created via the tests. Is this not the case? I don't understand how this is related to autocommit.

phoenix-core/src/it/java/org/apache/phoenix/end2end/IndexRebuildTaskIT.java
gokceni: @ChinmaySKulkarni, there is one task I create before this point in the same test function. I run an IndexRebuild task, and then I go ahead and remove the contents of index via admin api and re-run it to see if it builds. We can do it two ways: delete the task record and create a new record or just update the task record and set its status so that it gets picked up. Same thing really in terms of what we want to achieve.

gokceni:
@dbwong the main change in the actual code is to add a timestamp parameter to queryTaskTable method so that we get the Task with the given timestamp. It is tested in the IndexRebuildTaskIT (in code/test path already). If you have a particular test suggestion, I can try looking to implement that.

phoenix-core/src/main/java/org/apache/phoenix/schema/task/Task.java
```diff
@@ -230,7 +231,13 @@ public static void deleteTask(PhoenixConnection conn, PTable.TaskType taskType,
                 taskQuery += " AND " + PhoenixDatabaseMetaData.TASK_DATA + " LIKE '%" + indexName + "%'";
             }
 
-        return populateTasks(connection, taskQuery);
+            List<TaskRecord> result = populateTasks(connection, taskQuery);
+            if (ts != null) {
+                // Adding TASK_TS to the where clause did not work. It returns empty when directly querying with the timestamp.
+                result.removeIf(p -> !p.getTimeStamp().equals(ts));
```
ChinmaySKulkarni: This might complain for some branches which need to be source compatible with Java 7 instead of 8

phoenix-core/src/it/java/org/apache/phoenix/end2end/IndexRebuildTaskIT.java
```diff
@@ -188,7 +187,17 @@ public static void waitForTaskState(Connection conn, PTable.TaskType taskType, P
 
             String taskStatus = null;
 
-            if (rs.next()) {
+            while (rs.next()) {
```
dbwong: I will attempt to allude to the type of testing I'm referring to here.  I have not yet spent enough time to understand all the logic but this appears to be 2 changes.  1, the data layer may return multiple results for our query.  2, we need to test against the results and only.  However, since the logic today couples these concerns testing that our logic handles is not straight forward.  We should be able to write 3 unit tests, one that our logic handles multiple results and two we look for the expected name and the expected time stamp.  While I agree that your IT tests likely check this behavior the very act of making code unit testable suggests better boundaries.  @gokceni

phoenix-core/src/main/java/org/apache/phoenix/schema/task/Task.java
```diff
@@ -230,7 +232,20 @@ public static void deleteTask(PhoenixConnection conn, PTable.TaskType taskType,
                 taskQuery += " AND " + PhoenixDatabaseMetaData.TASK_DATA + " LIKE '%" + indexName + "%'";
             }
 
-        return populateTasks(connection, taskQuery);
+            List<TaskRecord> taskRecords = populateTasks(connection, taskQuery);
+            List<TaskRecord> result = new ArrayList<TaskRecord>();
+            if (ts != null) {
```
dbwong: Let me try again I accidentally commented in test code.  Here we have a simple logic change where we query against a property.  This is tested in the ITs but it is good to test with a small simple unit test.  However, it might pose a bit more difficult to test if there isn't a boundary between the data layer and the logic.  This may inform how to shape our code.

phoenix-core/src/it/java/org/apache/phoenix/end2end/IndexRebuildTaskIT.java
```diff
@@ -129,45 +128,44 @@ public void testIndexRebuildTask() throws Throwable {
...
             admin.disableTable(tableName);
-            admin.truncateTable(tableName, false);
+            admin.truncateTable(tableName, true);
 
             data = "{IndexName:" + indexName + ", DisableBefore:true}";
 
-            // Add a new task (update status to created) to System.Task to rebuild indexes
+            // Add a new task to System.Task to rebuild indexes with a new timestamp.
+            //Thread.sleep(1);
+            //startTs = new Timestamp(EnvironmentEdgeManager.currentTimeMillis());
```
ChinmaySKulkarni: Can we remove this commented code?

phoenix-core/src/it/java/org/apache/phoenix/end2end/IndexRebuildTaskIT.java
```diff
@@ -129,45 +128,42 @@ public void testIndexRebuildTask() throws Throwable {
             Task.addTask(conn.unwrap(PhoenixConnection.class), PTable.TaskType.INDEX_REBUILD,
                     TENANT1, null, viewName,
                     PTable.TaskStatus.CREATED.toString(), data, null, startTs, null, true);
-
-
+            conn.commit();
```
ChinmaySKulkarni: Why do we need `conn.commit` here?

phoenix-core/src/it/java/org/apache/phoenix/end2end/IndexRebuildTaskIT.java
```diff
@@ -188,7 +185,17 @@ public static void waitForTaskState(Connection conn, PTable.TaskType taskType, P
 
             String taskStatus = null;
 
-            if (rs.next()) {
+            while (rs.next()) {
+                String tableName = rs.getString(PhoenixDatabaseMetaData.TABLE_NAME);
+                if (!tableName.equals(expectedTableName)) {
```
lhofhansl: I think it would be better to filter for the table name in the SELECT statement above (i.e. add it to the WHERE clause)

phoenix-core/src/it/java/org/apache/phoenix/end2end/IndexRebuildTaskIT.java
lhofhansl: Likewise I don't understand why we need the commit here (and below)

phoenix-core/src/it/java/org/apache/phoenix/end2end/IndexRebuildTaskIT.java
```diff
@@ -83,6 +82,7 @@ private String generateDDL(String format) {
     @Test
     public void testIndexRebuildTask() throws Throwable {
         String baseTable = generateUniqueName();
+        String viewName = generateUniqueName();
         Connection conn = null;
         Connection viewConn = null;
```
swaroopak: nit: Could you please rename this to tenantConn, makes it more readable.

phoenix-core/src/it/java/org/apache/phoenix/end2end/IndexRebuildTaskIT.java
```diff
@@ -129,45 +128,42 @@ public void testIndexRebuildTask() throws Throwable {
             Task.addTask(conn.unwrap(PhoenixConnection.class), PTable.TaskType.INDEX_REBUILD,
```
ChinmaySKulkarni: @gokceni Above, we have upserted data into the view. Won't this automatically trigger an index build i.e. won't global view index get these populated rows already? Why do we need to explicitly run IndexRebuildTask?

phoenix-core/src/it/java/org/apache/phoenix/end2end/IndexRebuildTaskIT.java
```diff
@@ -129,45 +128,42 @@ public void testIndexRebuildTask() throws Throwable {
...
-
-
+            conn.commit();
             task.run();
+            waitForTaskState(conn, PTable.TaskType.INDEX_REBUILD, viewName, startTs, PTable.TaskStatus.COMPLETED);
 
             String viewIndexTableName = MetaDataUtil.getViewIndexPhysicalName(baseTable);
             ConnectionQueryServices queryServices = conn.unwrap(PhoenixConnection.class).getQueryServices();
             int count = getUtility().countRows(queryServices.getTable(Bytes.toBytes(viewIndexTableName)));
             assertTrue(count == numOfValues);
```
ChinmaySKulkarni: Not part of your diffs, but we can replace this with `assertEquals` instead

phoenix-core/src/it/java/org/apache/phoenix/end2end/IndexRebuildTaskIT.java
```diff
@@ -129,45 +128,42 @@ public void testIndexRebuildTask() throws Throwable {
...
+            admin.truncateTable(tableName, true);
 
             data = "{IndexName:" + indexName + ", DisableBefore:true}";
 
-            // Add a new task (update status to created) to System.Task to rebuild indexes
+            // Update task to to rebuild indexes.
             Task.addTask(conn.unwrap(PhoenixConnection.class), PTable.TaskType.INDEX_REBUILD,
                     TENANT1, null, viewName,
                     PTable.TaskStatus.CREATED.toString(), data, null, startTs, null, true);
+            conn.commit();
```
ChinmaySKulkarni: Why do we need `conn.commit` here?

phoenix-core/src/it/java/org/apache/phoenix/end2end/IndexRebuildTaskIT.java
ChinmaySKulkarni: +1 to that. I think that would also not require us to have the `continue` statements in case there are no matches.

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/TaskRegionObserver.java
```diff
@@ -199,6 +200,7 @@ public void run() {
                             Task.addTask(connForTask, taskRecord.getTaskType(), taskRecord.getTenantId(), taskRecord.getSchemaName(),
                                     taskRecord.getTableName(), PTable.TaskStatus.STARTED.toString(),
                                     taskRecord.getData(), taskRecord.getPriority(), taskRecord.getTimeStamp(), null, true);
+                            connForTask.commit();
```
ChinmaySKulkarni: nit: null check for `connForTask`

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/TaskRegionObserver.java
```diff
@@ -255,6 +257,7 @@ public static void setEndTaskStatus(PhoenixConnection connForTask, Task.TaskReco
             Task.addTask(connForTask, taskRecord.getTaskType(), taskRecord.getTenantId(), taskRecord.getSchemaName(),
                     taskRecord.getTableName(), taskStatus, data, taskRecord.getPriority(),
                     taskRecord.getTimeStamp(), endTs, true);
+            connForTask.commit();
```
ChinmaySKulkarni: nit: null check for `connForTask`

phoenix-core/src/main/java/org/apache/phoenix/schema/task/Task.java
```diff
@@ -230,7 +232,20 @@ public static void deleteTask(PhoenixConnection conn, PTable.TaskType taskType,
...
+                // Adding TASK_TS to the where clause did not work. It returns empty when directly querying with the timestamp.
+                for (TaskRecord tr : taskRecords) {
```
ChinmaySKulkarni: @gokceni Why do we need these changes for fixing the test?

swaroopak:
@gokceni Let me know when you update this PR.

phoenix-core/src/it/java/org/apache/phoenix/end2end/IndexRebuildTaskIT.java
ChinmaySKulkarni: @gokceni I still don't understand why we are committing here.

phoenix-core/src/it/java/org/apache/phoenix/end2end/IndexRebuildTaskIT.java
ChinmaySKulkarni: @gokceni can you please address these comments?

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/TaskRegionObserver.java
ChinmaySKulkarni: @gokceni why do we need to commit here?

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/TaskRegionObserver.java
ChinmaySKulkarni: Also, this and the previous change in this file seem outside the scope of this JIRA which is just to fix the flaky test.

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/tasks/IndexRebuildTask.java
```diff
@@ -107,6 +107,7 @@
             Task.addTask(conn.unwrap(PhoenixConnection.class ), taskRecord.getTaskType(), taskRecord.getTenantId(), taskRecord.getSchemaName(),
                     taskRecord.getTableName(), PTable.TaskStatus.STARTED.toString(), jsonObject.toString(), taskRecord.getPriority(),
                     taskRecord.getTimeStamp(), null, true);
+            conn.commit();
```
ChinmaySKulkarni: @gokceni what are we committing here? Also, this change seems outside the scope of this JIRA again. Is there any change you made which would now require a `conn.commit` after adding a task?

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/tasks/IndexRebuildTask.java
```diff
@@ -162,7 +163,9 @@ private String getJobID(String data) {
             Cluster cluster = new Cluster(configuration);
 
             Job job = cluster.getJob(org.apache.hadoop.mapreduce.JobID.forName(jobID));
-
+            if (job == null) {
```
ChinmaySKulkarni: @gokceni where is this used?

phoenix-core/src/main/java/org/apache/phoenix/schema/task/Task.java
ChinmaySKulkarni: ping @gokceni please address these review comments from earlier. Also, would this change affect any other tasks wherein timestamp is specified?

phoenix-core/src/it/java/org/apache/phoenix/end2end/IndexRebuildTaskIT.java
gokceni: Removed the commit after adding task. The connection is setAutocommit(false)

phoenix-core/src/it/java/org/apache/phoenix/end2end/IndexRebuildTaskIT.java
gokceni: addressed

phoenix-core/src/it/java/org/apache/phoenix/end2end/IndexRebuildTaskIT.java
gokceni: done

phoenix-core/src/it/java/org/apache/phoenix/end2end/IndexRebuildTaskIT.java
gokceni: removed

phoenix-core/src/it/java/org/apache/phoenix/end2end/IndexRebuildTaskIT.java
gokceni: done

phoenix-core/src/it/java/org/apache/phoenix/end2end/IndexRebuildTaskIT.java
gokceni: removed

phoenix-core/src/main/java/org/apache/phoenix/schema/task/Task.java
gokceni: @ChinmaySKulkarni Your previous comments about uniquely getting a task made me question about task query. I realized that there might be a case where we have tasks for the same index and table but with a different timestamp. I change the code to handle that.

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/TaskRegionObserver.java
gokceni: removed

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/tasks/IndexRebuildTask.java
gokceni: removed

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/tasks/IndexRebuildTask.java
gokceni: When you kick the task and it runs index tool, cluster.getJob doesn't return a job immediately sometimes. We have to come back and requery the job status when we find the job from cluster.getJob

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/TaskRegionObserver.java
ChinmaySKulkarni: @gokceni looks like `connForTask.commit()` is still there.

gjacoby126:
JIRA has already been committed, closing PR.

PR #558, opened 2019-07-31
PHOENIX-5413: Fix union all error when exists literal column type.

IceMimosa:
Fix [PHOENIX-5413](https://issues.apache.org/jira/browse/PHOENIX-5413). The `maxLength` should be null even though there exists literal column type.


phoenix-core/src/it/java/org/apache/phoenix/end2end/UnionAllIT.java
```diff
@@ -1065,4 +1065,41 @@ public void testCoerceExpr() throws Exception {
...
+
+            String dml = "UPSERT INTO " + tableName1 + " VALUES(?, ?)";
+            PreparedStatement stmt = conn.prepareStatement(dml);
+            stmt.setString(1, "a");
+            stmt.setString(2, "11111");
+            stmt.execute();
+            conn.commit();
+
+            String unionAll = "select id, col1 from " + tableName1 + " union all select 'b' id, '2' col1";
+            String select  = "select t.id, case when t.col1='11111' then 'yes' else 'no' end c from ( " + unionAll + " ) t";
```
yanxinyi: Maybe reformat SQL here

yanxinyi:
Overall looks good to me, can you also submit your patch to the jira so that we can get a test run? Thanks.

IceMimosa:
@yanxinyi Thanks for reviewing, i have uploaded the patch to jira.

phoenix-core/src/it/java/org/apache/phoenix/end2end/UnionAllIT.java
```diff
@@ -1065,4 +1065,44 @@ public void testCoerceExpr() throws Exception {
...
     }
+
+    @Test
+    public void testUnionAllWithLiteralColumn() throws Exception {
+        String tableName1 = generateUniqueName();
+        Properties props = PropertiesUtil.deepCopy(TEST_PROPERTIES);
+        Connection conn = DriverManager.getConnection(getUrl(), props);
+        conn.setAutoCommit(false);
+
+        try {
```
swaroopak: Could you please use try-with-resources?

IceMimosa: Done, thanks.

phoenix-core/src/test/java/org/apache/phoenix/compile/QueryCompilerTest.java
```diff
@@ -2465,10 +2465,10 @@ public void testColumnProjectionUnionAll() throws SQLException {
             assertTrue(rowProj.getColumnProjector(1).getExpression().getMaxLength() == 20);
             assertTrue(rowProj.getColumnProjector(2).getExpression().getDataType()
                 instanceof PVarchar);
-            assertTrue(rowProj.getColumnProjector(2).getExpression().getMaxLength() == 30);
+            assertTrue(rowProj.getColumnProjector(2).getExpression().getMaxLength() == null);
```
yanxinyi: `/**
     * @return the actual length of the column. For decimal, it would be its precision. For char or
     * varchar, it would be the maximum length as specified during schema definition.
     */
    Integer getMaxLength();

    /**
     * @return scale of a decimal number.
     */
    Integer getScale();
`
I'm not sure about `getMaxLength` and `getScale` return `null` here, it might be better to change the logic to return the actual number instead.  @swaroopak any input here?

IceMimosa: @yanxinyi Actually, the issue is caused by class [ComparisonExpression](https://github.com/apache/phoenix/blob/e3d3ae5b120189235259280fc074bc7f337aa5b3/phoenix-core/src/main/java/org/apache/phoenix/expression/ComparisonExpression.java#L264). Column type `varchar` union `char(10)` is in compile phase, and it can't determine the maxLength of the result column. API `getMaxLength` and `getScale` may be return `null`, for example, if the input is `varchar` or `decimal` with no limit.
Maybe we should fix the class ComparisonExpression, i will try to look at other possible cases.

PR #559, opened 2019-08-01
PHOENIX-5419: Converting anonymous to inner class in TracingQueryPlan

ankitjain64:


phoenix-core/src/main/java/org/apache/phoenix/compile/TraceQueryPlan.java
```diff
@@ -295,4 +236,71 @@ public Long getEstimateInfoTimestamp() throws SQLException {
     public List<OrderBy> getOutputOrderBys() {
         return Collections.<OrderBy> emptyList();
     }
+
+    private class QueryResultIterator implements ResultIterator {
```
swaroopak: nit: could you please rename the private class to TraceQueryResultIterator?

ankitjain64: @swaroopak : Can you please have a look now. Thanks.

gjacoby126:
@ankitjain64 could you please include the JIRA id PHOENIX-5419 in the commit header.

ankitjain64:
> @ankitjain64 could you please include the JIRA id PHOENIX-5419 in the commit header.

@gjacoby126 : Done. Thanks

gjacoby126:
@ankitjain64 - for future reference, when you squash your commits into a single one, please also edit the commit message so that it only has the one line. 

This commit message had a line for each of the former commits, and since the UI truncated it to just show the first, I didn't notice until after I pushed the merge button. I can fix it when I push to the 4.15 branches, but since this repo doesn't allow force pushes, I think it will have to remain messy in master.

ankitjain64:
> @ankitjain64 - for future reference, when you squash your commits into a single one, please also edit the commit message so that it only has the one line.
> 
> This commit message had a line for each of the former commits, and since the UI truncated it to just show the first, I didn't notice until after I pushed the merge button. I can fix it when I push to the 4.15 branches, but since this repo doesn't allow force pushes, I think it will have to remain messy in master.

Sure, I will make sure about it from next time. Thank you.

PR #564, opened 2019-08-06
PHOENIX-5416: Fix Array2IT testArrayRefToLiteral

ChinmaySKulkarni:
For Arrays of `CHAR` and `BINARY` which are both fixed width and both have getByteSize() null, we run into an NPE.

With this change, inside ArrayIndexFunction#getMaxLength, instead of returning this.children.get(0).getMaxLength(), we inspect the children.get(0) `LiteralExpression` and return the maxLength of the underlying `PCharArray`/`PBinaryArray` data.

### DIFF
--- a/phoenix-core/src/it/java/org/apache/phoenix/end2end/Array2IT.java
+++ b/phoenix-core/src/it/java/org/apache/phoenix/end2end/Array2IT.java
@@ -18,6 +18,7 @@
 package org.apache.phoenix.end2end;
 
 import static org.apache.phoenix.util.TestUtil.TEST_PROPERTIES;
+import static org.junit.Assert.assertArrayEquals;
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertFalse;
 import static org.junit.Assert.assertNull;
@@ -37,10 +38,12 @@ import org.apache.phoenix.schema.types.PhoenixArray;
 import org.apache.phoenix.util.PropertiesUtil;
 import org.apache.phoenix.util.SchemaUtil;
 import org.apache.phoenix.util.StringUtil;
-import org.junit.Ignore;
 import org.junit.Test;
 
 public class Array2IT extends ArrayIT {
+
+    private static final String TEST_QUERY = "select ?[2] from \"SYSTEM\".\"CATALOG\" limit 1";
+
     @Test
     public void testFixedWidthCharArray() throws Exception {
         Connection conn;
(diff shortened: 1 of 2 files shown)


ChinmaySKulkarni:
Please review @twdsilva @gjacoby126 @dbwong

twdsilva:
+1 , LGTM.

gjacoby126:
@ChinmaySKulkarni  @twdsilva  Should this logic be encapsulated into LiteralExpression.getMaxLength() itself? Might be cleaner and prevent errors elsewhere.

twdsilva:
Moving this logic into LiteralExpression.getMaxLength() is a good idea, assuming it doesn't break anything else :)

ChinmaySKulkarni:
Good point @gjacoby126. I have moved the logic to LiteralExpression. Attaching a new patch on the JIRA with this change to make sure it doesn't break anything *fingers crossed*. Updated the PR, please review now @gjacoby126 @twdsilva

phoenix-core/src/main/java/org/apache/phoenix/expression/LiteralExpression.java
```diff
@@ -338,6 +338,17 @@ public PDataType getDataType() {
 
     @Override
     public Integer getMaxLength() {
+        // For literals representing arrays of CHAR or BINARY, the byte size is null and the max
```
BinShi-SecularBird: Isn't the logic from 344 to 350 already covered by public static LiteralExpression newConstant(...)? It seems that the value/maxLength are assigned by newConstant static function  in which a LiteralExpression object is created.

phoenix-core/src/it/java/org/apache/phoenix/end2end/Array2IT.java
```diff
@@ -670,12 +670,13 @@ public void testPKWithArrayNotInEnd() throws Exception {
...
-    @Test // see PHOENIX-5416
-    @Ignore
-    public void testArrayRefToLiteral() throws Exception {
+    @Test
+    public void testArrayRefToLiteralCharArraySameLengths() throws Exception {
         Properties props = PropertiesUtil.deepCopy(TEST_PROPERTIES);
         try (Connection conn = DriverManager.getConnection(getUrl(), props)) {
-            PreparedStatement stmt = conn.prepareStatement("select ?[2] from \"SYSTEM\".\"CATALOG\" limit 1");
+            PreparedStatement stmt = conn.prepareStatement("select ?[2] from "
+                    + "\"SYSTEM\".\"CATALOG\" limit 1");
```
BinShi-SecularBird: define static final String TEST_STATEMENT = "select ?[2] from \"SYSTEM\".\"CATALOG\" limit 1"?

phoenix-core/src/main/java/org/apache/phoenix/expression/LiteralExpression.java
```diff
@@ -338,6 +338,17 @@ public PDataType getDataType() {
...
+        // length of the expression is also null, so we must get the max length of the
+        // actual underlying array
+        if (maxLength == null && getDataType().isArrayType() &&
+                PDataType.arrayBaseType(getDataType()).getByteSize() == null) {
+            Object value = getValue();
+            if (value instanceof PhoenixArray) {
+                // Return the max length of the underlying PhoenixArray data
+                return ((PhoenixArray) value).getMaxLength();
```
BinShi-SecularBird: Such type casting means we don't have good class hierarchy structure; otherwise we should just call value.getMaxLength.

phoenix-core/src/main/java/org/apache/phoenix/expression/LiteralExpression.java
ChinmaySKulkarni: I ran my debugger through this and it looks like [that code](https://github.com/apache/phoenix/blob/master/phoenix-core/src/main/java/org/apache/phoenix/expression/LiteralExpression.java#L206) is never hit. Overall, this class needs to be refactored to follow the builder pattern. There are 5 constructors and 10 overloaded methods for `newConstant` which basically instantiates a new `LiteralExpression`. I am not aware of all of the code paths involved here and whether the `newConstant` method where this logic already exists is getting called in all those cases. As I mentioned, this code is not getting hit for the test cases I've added in this PR, so I don't think it is safe to rely on changing `newConstant`.

phoenix-core/src/main/java/org/apache/phoenix/expression/LiteralExpression.java
ChinmaySKulkarni: I completely agree. There is a lot of scope for improvement in restructuring/refactoring all this code. Unfortunately that is out of scope for the current JIRA so I have to do this (ugly) type casting.

ChinmaySKulkarni:
I have updated the patch and PR as per reviews. Please take a look now @BinShi-SecularBird @gjacoby126 @twdsilva

phoenix-core/src/main/java/org/apache/phoenix/expression/LiteralExpression.java
BinShi-SecularBird: The interface of LiteralExpression is a mess. While it provides newConstant() APIs to build the object, it also provide two public constructors.

Overall, it seems to me that all production and test code paths should call newConstant() APIs to build the LiteralExpression object. Once value/type/byteValue/maxLength are passed through the following private LiteralExpression() constructor, they should never be change and the get accessor should return exactly what is. Having this ad-hoc change in getMaxLength() will make code quality to be worse and hurt the performance of this getter accessor. If this introduce big refactor which is out of scope of this JIRA PHOENIX-5416. Shall we leave getMaxLength() as it was and open a new JIRA to refactor LiteralExpression class and its callers? 

private LiteralExpression(Object value, PDataType type, byte[] byteValue,
            Integer maxLength, Integer scale, SortOrder sortOrder, Determinism deterministic) {

phoenix-core/src/main/java/org/apache/phoenix/expression/LiteralExpression.java
BinShi-SecularBird: I agree this is out of scope for the current JIRA.

ChinmaySKulkarni:
Thanks for the review @BinShi-SecularBird. I have opened https://issues.apache.org/jira/browse/PHOENIX-5432 to introduce this refactoring in `LiteralExpression`. I will commit this change for now.

PR #565, opened 2019-08-06
PHOENIX-5428 Upgrade maven-checkstyle-plugin version

stoty:
This upgrades maven-checkstyle-plugin to 3.1.0 and
updates the checker.xml config files to work with recent
checkstyle versions


joshelser:
Looks like we're skipping checkstyle for the build now, but I can see that this does run validations. Tough to say if the change is "OK" since there are so many violations.

PR #567, opened 2019-08-15
PHOENIX-5134 Phoenix Connection Driver #normalize does not distinguis…

yanxinyi:
…h different url with same ZK quorum but different Properties

### DIFF
--- a/phoenix-core/src/main/java/org/apache/phoenix/jdbc/PhoenixEmbeddedDriver.java
+++ b/phoenix-core/src/main/java/org/apache/phoenix/jdbc/PhoenixEmbeddedDriver.java
@@ -253,11 +253,11 @@ public abstract class PhoenixEmbeddedDriver implements Driver, SQLCloseable {
             if (tokenizer.hasMoreTokens() && !TERMINATOR.equals(token)) {
                 String extraToken = tokenizer.nextToken();
                 if (WINDOWS_SEPARATOR_CHAR == extraToken.charAt(0)) {
-                  String prevToken = tokens[nTokens - 1];
-                  tokens[nTokens - 1] = prevToken + ":" + extraToken;
-                  if (tokenizer.hasMoreTokens() && !(token=tokenizer.nextToken()).equals(TERMINATOR)) {
-                      throw getMalFormedUrlException(url);
-                  }
+                    String prevToken = tokens[nTokens - 1];
+                    tokens[nTokens - 1] = prevToken + ":" + extraToken;
+                    if (tokenizer.hasMoreTokens() && !(token=tokenizer.nextToken()).equals(TERMINATOR)) {
+                        throw getMalFormedUrlException(url);
+                    }
                 } else {
                     throw getMalFormedUrlException(url);
                 }
@@ -304,15 +304,22 @@ public abstract class PhoenixEmbeddedDriver implements Driver, SQLCloseable {
                     }
                 }
             }
-            return new ConnectionInfo(quorum,port,rootNode, principal, keytabFile);
+            return new ConnectionInfo(quorum,port,rootNode, principal, keytabFile,
+                    getTenantIDParser(url));
         }
-        
+
         public ConnectionInfo normalize(ReadOnlyProps props, Properties info) throws SQLException {
+            return normalize(props, info, null);
+        }
+
+        public ConnectionInfo normalize(ReadOnlyProps props, Properties info, String url)
+                throws SQLException {
             String zookeeperQuorum = this.getZookeeperQuorum();
             Integer port = this.getPort();
             String rootNode = this.getRootNode();
             String keytab = this.getKeytab();
             String principal = this.getPrincipal();
+            String tenantID = this.getTenantID();
             // Normalize connInfo so that a url explicitly specifying versus implicitly inheriting
             // the default values will both share the same ConnectionQueryServices.
             if (zookeeperQuorum == null) {
@@ -398,7 +405,7 @@ public abstract class PhoenixEmbeddedDriver implements Driver, SQLCloseable {
                 }
             } // else, no connection, no need to login
             // Will use the current User from UGI
-            return new ConnectionInfo(zookeeperQuorum, port, rootNode, principal, keytab);
+            return new ConnectionInfo(zookeeperQuorum, port, rootNode, principal, keytab, tenantID);
         }
 
         // Visible for testing
@@ -489,8 +496,23 @@ public abstract class PhoenixEmbeddedDriver implements Driver, SQLCloseable {
         private final String principal;
         private final String keytab;
         private final User user;
+        private final String tenantID;
         
         public ConnectionInfo(String zookeeperQuorum, Integer port, String rootNode, String principal, String keytab) {
+            this(zookeeperQuorum, port, rootNode, principal, keytab, null);
+        }
+
+        /**
+         * Constructor ConnectionInfo
+         * @param zookeeperQuorum zookeeperQuorum info
+         * @param port port number
+         * @param rootNode rootNode
+         * @param principal Kerberos user principal
+         * @param tenantID tenantID
+         */
+        public ConnectionInfo(String zookeeperQuorum, Integer port, String rootNode,
+                              String principal, String keytab, String tenantID) {
+
             this.zookeeperQuorum = zookeeperQuorum;
             this.port = port;
             this.rootNode = rootNode;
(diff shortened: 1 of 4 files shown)


phoenix-core/src/test/java/org/apache/phoenix/jdbc/PhoenixEmbeddedDriverTest.java
```diff
@@ -146,7 +149,70 @@ public void testNegativeGetConnectionInfo() throws SQLException {
...
+                "jdbc:phoenix",
+                "jdbc:phoenix;test=true",
+                "jdbc:phoenix:localhost",
+                "localhost:123:/foo-bar",
+                "localhost:123:/foo-bar;TenantId",
+                "localhost:123:/foo-bar;TenantId="
+        };
+        for (int i = 0; i < invalidUrls.length; i++) {
+            try {
+                assertEquals( ConnectionInfo.getTenantID(invalidUrls[i]), null);
```
dbwong: nit: expected is first

phoenix-core/src/test/java/org/apache/phoenix/jdbc/PhoenixEmbeddedDriverTest.java
```diff
@@ -146,7 +149,70 @@ public void testNegativeGetConnectionInfo() throws SQLException {
...
+            } catch (AssertionError e) {
```
dbwong: Why catch and rethrow?

phoenix-core/src/test/java/org/apache/phoenix/jdbc/PhoenixEmbeddedDriverTest.java
```diff
@@ -146,7 +149,70 @@ public void testNegativeGetConnectionInfo() throws SQLException {
...
+                null,
+                "TenantId=1",
+                "jdbc:phoenix;TenantId=1",
+                "jdbc:phoenix;test=true;TenantId=1",
+                "jdbc:phoenix:localhost;TenantId=1",
+                "localhost:123:/foo-bar;TenantId=1"
+        };
+        for (int i = 0; i < validUrls.length; i++) {
+            try {
+                assertEquals( ConnectionInfo.getTenantID(validUrls[i]), null);
```
dbwong: Should't this return valid tenant Ids?

phoenix-core/src/main/java/org/apache/phoenix/jdbc/PhoenixEmbeddedDriver.java
```diff
@@ -304,15 +305,21 @@ public static ConnectionInfo create(String url) throws SQLException {
...
+        }
+
+        public ConnectionInfo normalize(ReadOnlyProps props, Properties info, String url)
+                throws SQLException {
             String zookeeperQuorum = this.getZookeeperQuorum();
             Integer port = this.getPort();
             String rootNode = this.getRootNode();
             String keytab = this.getKeytab();
             String principal = this.getPrincipal();
+            String tenantID = getTenantID(url);
```
dbwong: nit: This variable seems redundant can just pass to the ConnectionInfo constructor.

phoenix-core/src/main/java/org/apache/phoenix/jdbc/PhoenixEmbeddedDriver.java
```diff
@@ -505,8 +519,21 @@ public ConnectionInfo(String zookeeperQuorum, Integer port, String rootNode, Str
             if (null == this.user) {
                 throw new RuntimeException("Acquired null user which should never happen");
             }
+            this.tenantID = tenantID;
         }
-        
+
+        public static String getTenantID(String url) {
```
dbwong: Public function should have javadoc.  Also this clearly returns null Strings.  Consider Optional to enforce callers handle those cases.  While this can be static, why are we treating this differently than getPort, etc and then just returning the tenantID?  Also will make future mocking simpler.

phoenix-core/src/main/java/org/apache/phoenix/jdbc/PhoenixEmbeddedDriver.java
```diff
@@ -505,8 +519,21 @@ public ConnectionInfo(String zookeeperQuorum, Integer port, String rootNode, Str
...
+            if (url != null && url.contains(PhoenixRuntime.TENANT_ID_ATTRIB)) {
+                for (String param : url.split(";")) {
```
dbwong: I believe we have a separator constant for ';'.

phoenix-core/src/main/java/org/apache/phoenix/jdbc/PhoenixEmbeddedDriver.java
```diff
@@ -505,8 +519,21 @@ public ConnectionInfo(String zookeeperQuorum, Integer port, String rootNode, Str
...
+                    if (param.startsWith(PhoenixRuntime.TENANT_ID_ATTRIB) &&
```
dbwong: I think this code may have some errors for a string like "TenantIdXY", is there protection against these type of non-matching?

dbwong: Could also have some weirdness for stuff like TenantId=TenantId=, might not be worth handling currently but the use of split and whether we allow ='s in the tenant Id.

phoenix-core/src/main/java/org/apache/phoenix/jdbc/PhoenixEmbeddedDriver.java
```diff
@@ -489,8 +496,15 @@ private Configuration getConfiguration(ReadOnlyProps props, Properties info, Str
         private final String principal;
         private final String keytab;
         private final User user;
+        private final String tenantID;
         
         public ConnectionInfo(String zookeeperQuorum, Integer port, String rootNode, String principal, String keytab) {
+            this(zookeeperQuorum, port, rootNode, principal, keytab, null);
+        }
+
+        public ConnectionInfo(String zookeeperQuorum, Integer port, String rootNode,
```
dbwong: Great to add javadoc for new parameter.

yanxinyi:
I agree with the static method is not a good design for testing, and we should spend time to refactor the code. Yes, we can create a new URL parsing utility concept class, but no reason for only one method.

phoenix-core/src/main/java/org/apache/phoenix/jdbc/PhoenixEmbeddedDriver.java
```diff
@@ -304,15 +304,22 @@ public static ConnectionInfo create(String url) throws SQLException {
...
+        }
+
+        public ConnectionInfo normalize(ReadOnlyProps props, Properties info, String url)
+                throws SQLException {
             String zookeeperQuorum = this.getZookeeperQuorum();
             Integer port = this.getPort();
             String rootNode = this.getRootNode();
             String keytab = this.getKeytab();
             String principal = this.getPrincipal();
+            String tenantID = this.getTenantID();
```
swaroopak: nit: use tenantId instead of tenantID (ref: https://google.github.io/styleguide/javaguide.html#s5.3-camel-case)

yanxinyi:
Should we add a config and let people decide whether they want to enable and disable this feature? Since HConnection is kind of heavy so that they want to use the same HConnection for all tenents.

@swaroopak @dbwong @ChinmaySKulkarni

dbwong:
> Should we add a config and let people decide whether they want to enable and disable this feature? Since HConnection is kind of heavy so that they want to use the same HConnection for all tenents.
> 
> @swaroopak @dbwong @ChinmaySKulkarni

I don't have the best visibility into use cases outside ours but typically that is the "safest" approach.  Ping @joshelser @twdsilva does this need to be parameterized?

yanxinyi:
Jira has an example of the usecase:` different tenant wants to have different HBase config (such as HBase timeout settings)`.

twdsilva:
I don't think we should create a new HConnection for a tenant specific connection, since its pretty heavy weight, and this could end up creating a lot of HConnections. Its fine for different tenant to use the sam CQSI as long as the rest of the config in the Properties is the same. 
If a client tries to connect to a cluster with a Properties object that is different than what is currently present in the cache, we should create a new CQSI instead of using the cached one that was created with a different configuration. Maybe you can use the Properties object to compute the hashCode of ConnectionInfo so that if a client passes in different Properties config a new HConnection is created?

gjacoby126:
@yanxinyi  Keep in mind that each CQSI has its own HConnection, and each HConnection I believe has its own ZooKeeper connection, and ZK has its own limitations for how many connections it can manage. ZK connections spawned by HConnections also tend to create a lot of threads on the client side. 

In addition, the connection-throttling framework in Phoenix is per-CQSI, so creating lots of CQSIs for similar clients  connecting to the same cluster greatly reduces the effectiveness of the throttle.

PR #569, opened 2019-08-21
PHOENIX-4743: ALTER TABLE ADD COLUMN for global index should not modify HBase metadata if failed

sandeepvinayak:


phoenix-core/src/main/java/org/apache/phoenix/query/ConnectionQueryServicesImpl.java
```diff
@@ -2127,6 +2124,13 @@ public MetaDataResponse call(MetaDataService instance) throws IOException {
                     }
                 }
             }
+            else if (result.getMutationCode() == MutationCode.UNALLOWED_TABLE_MUTATION) {
```
swaroopak: nit : put it on the same line as closing bracket above.

phoenix-core/src/it/java/org/apache/phoenix/end2end/AlterTableIT.java
```diff
@@ -852,6 +854,37 @@ public void testDropMultipleColumns() throws Exception {
...
+    @Test
+    public void testAlterTableOnGlobalIndex() throws Exception {
+        try (Connection conn = DriverManager.getConnection(getUrl())) {
+            conn.setAutoCommit(false);
+            Statement stmt = conn.createStatement();
+            String tableName = generateUniqueName();
+            String globalIndexTableName = generateUniqueName();
+
+            stmt.execute("CREATE TABLE " + tableName + "(k VARCHAR PRIMARY KEY, v1 VARCHAR, v2 VARCHAR)");
+            stmt.execute("DROP TABLE " + tableName);
```
ChinmaySKulkarni: Why do we need to execute these create and drop table statements?

phoenix-core/src/it/java/org/apache/phoenix/end2end/AlterTableIT.java
```diff
@@ -852,6 +854,37 @@ public void testDropMultipleColumns() throws Exception {
...
+            admin.deleteTable(TableName.valueOf(tableName));
+            conn.createStatement().execute("CREATE TABLE " + tableName +
+                " (ID INTEGER PRIMARY KEY, COL1 VARCHAR(10), COL2 BOOLEAN)");
+
+            conn.createStatement().execute("CREATE INDEX " + globalIndexTableName + " on " + tableName + " (COL2)");
+            TableDescriptor originalDesc = admin.getDescriptor(TableName.valueOf(globalIndexTableName));
+            int expectedErrorCode = 0;
+            try {
+                conn.createStatement().execute("ALTER TABLE " + globalIndexTableName + " ADD CF1.AGE INTEGER ");
+                conn.commit();
```
ChinmaySKulkarni: Instead of initializing `expectedErrorCode` to zero to ensure later that the exception happened, you can instead call `Assert.fail()` after `conn.commit()` so it fails in case the SQLException does not happen. Then you can assert the exception code inside the catch statement.

phoenix-core/src/main/java/org/apache/phoenix/query/ConnectionQueryServicesImpl.java
ChinmaySKulkarni: We can combine this check and the bottom `if` condition:

```java
if (modifyHTable && result.getMutationCode() != MutationCode.UNALLOWED_TABLE_MUTATION) {
    sendHBaseMetaData(tableDescriptors, pollingNeeded);
}
```

sandeepvinayak:
@ChinmaySKulkarni Addressed your comments, can you please take a look now?

phoenix-core/src/it/java/org/apache/phoenix/end2end/AlterTableIT.java
```diff
@@ -852,6 +854,39 @@ public void testDropMultipleColumns() throws Exception {
...
+            } catch (SQLException e) {
+                assertEquals(e.getErrorCode(), CANNOT_MUTATE_TABLE.getErrorCode());
+            }
+
+            TableDescriptor finalDesc = admin.getDescriptor(TableName.valueOf(globalIndexTableName));
+            assertTrue(finalDesc.equals(originalDesc));
+
+            // remove the table
+            stmt.execute("DROP TABLE " + tableName);
+            admin.disableTable(TableName.valueOf(tableName));
```
ChinmaySKulkarni: This and `admin.deleteTable` is not necessary since you are dropping the table above. Also, can we move this to a finally block?

phoenix-core/src/it/java/org/apache/phoenix/end2end/AlterTableIT.java
```diff
@@ -852,6 +854,39 @@ public void testDropMultipleColumns() throws Exception {
         conn1.close();
     }
 
+    @Test
+    public void testAlterTableOnGlobalIndex() throws Exception {
+        try (Connection conn = DriverManager.getConnection(getUrl())) {
+            conn.setAutoCommit(false);
+            Admin admin = conn.unwrap(PhoenixConnection.class).getQueryServices().getAdmin();
+            Statement stmt = conn.createStatement();
```
ChinmaySKulkarni: Can we reuse `stmt` throughout this test rather than calling `conn.createStatement()` again? Also, you can move this inside the try-with-resources block so the Statement object also gets closed at the end

phoenix-core/src/it/java/org/apache/phoenix/end2end/AlterTableIT.java
sandeepvinayak: @ChinmaySKulkarni In tests, I saw the pattern of disabling and deleting even after dropping the table. Does the drop alll the operations the delete does ? I can remove it for sure

ChinmaySKulkarni: Ok that's fine. Let's keep it as is then.

ChinmaySKulkarni:
Please also squash all commits and make sure your final commit has just the JIRA number and Title in it. Thanks.

gjacoby126:
@sandeepvinayak , when you get a minute could you please rebase and then fix the commits as @ChinmaySKulkarni suggests above, and we can get this committed. Thanks!

sandeepvinayak:
@gjacoby126 this has already been committed. Closing this PR 
https://github.com/apache/phoenix/commit/197b6e30c894b657758c5d0cb3c6182d6c8d4723 
Thanks

PR #581, opened 2019-09-06
PHOENIX-5274: ConnectionQueryServiceImpl#ensureNamespaceCreated and ensureTableCreated should use HBase APIs that do not require ADMIN permissions for existence checks.

ankitjain64:
Using hbaseAdmin listNamespaces API rather than getNamespaceDescriptor as it does not require ADMIN permissions.

### DIFF
--- a/phoenix-core/src/it/java/org/apache/phoenix/end2end/CreateSchemaIT.java
+++ b/phoenix-core/src/it/java/org/apache/phoenix/end2end/CreateSchemaIT.java
@@ -18,7 +18,7 @@
 package org.apache.phoenix.end2end;
 
 import static org.junit.Assert.assertEquals;
-import static org.junit.Assert.assertNotNull;
+import static org.junit.Assert.assertTrue;
 import static org.junit.Assert.fail;
 
 import java.sql.Connection;
@@ -33,6 +33,7 @@ import org.apache.phoenix.query.QueryServices;
 import org.apache.phoenix.schema.SchemaAlreadyExistsException;
 import org.apache.phoenix.util.PropertiesUtil;
 import org.apache.phoenix.util.SchemaUtil;
+import org.apache.phoenix.util.ServerUtil;
 import org.apache.phoenix.util.TestUtil;
 import org.junit.Test;
 
@@ -54,9 +55,9 @@ public class CreateSchemaIT extends ParallelStatsDisabledIT {
         try (Connection conn = DriverManager.getConnection(getUrl(), props);
                 HBaseAdmin admin = conn.unwrap(PhoenixConnection.class).getQueryServices().getAdmin();) {
             conn.createStatement().execute(ddl1);
-            assertNotNull(admin.getNamespaceDescriptor(schemaName1));
+            assertTrue(ServerUtil.isHbaseNamespaceAvailable(admin, schemaName1));
             conn.createStatement().execute(ddl2);
-            assertNotNull(admin.getNamespaceDescriptor(schemaName2.toUpperCase()));
+            assertTrue(ServerUtil.isHbaseNamespaceAvailable(admin, schemaName2.toUpperCase()));
         }
         // Try creating it again and verify that it throws SchemaAlreadyExistsException
         try (Connection conn = DriverManager.getConnection(getUrl(), props)) {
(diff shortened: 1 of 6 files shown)


phoenix-core/src/it/java/org/apache/phoenix/end2end/CreateSchemaIT.java
```diff
@@ -95,9 +96,33 @@ public void testCreateSchema() throws Exception {
...
     }
+
+    /**
+     * Returns true if HBase namespace exists, else returns false
+     * @param hbaseAdmin HbaseAdmin Object
+     * @param schemaName Phoenix schema name for which we check existence of the HBase namespace
+     * @return true if the HBase namespace exists, else returns false
+     * @throws SQLException If there is an exception checking the HBase namespace
+     */
+    boolean isHbaseNamespaceAvailable(HBaseAdmin hbaseAdmin, String schemaName) throws IOException{
```
gjacoby126: HBaseAdmin is now listed as InterfaceAudience.Private, since it's a concrete class. If possible we should depend on the Admin interface it implements.

gjacoby126: Rather than duplicating the code between several tests, better to have a common place the function lives. For example, can just call the version of it in CQS, or move the version you added in CQS to a util class somewhere.

phoenix-core/src/main/java/org/apache/phoenix/query/ConnectionQueryServicesImpl.java
```diff
@@ -1159,6 +1153,29 @@ boolean ensureNamespaceCreated(String schemaName) throws SQLException {
...
 
+    /**
+     * Returns true if HBase namespace exists, else returns false
+     * @param hbaseAdmin HbaseAdmin Object
+     * @param schemaName Phoenix schema name for which we check existence of the HBase namespace
+     * @return true if the HBase namespace exists, else returns false
+     * @throws SQLException If there is an exception checking the HBase namespace
+     */
+    boolean isHbaseNamespaceAvailable(HBaseAdmin hbaseAdmin, String schemaName) throws IOException{
+        boolean namespaceExists = false;
```
gjacoby126: Need a test for this logic.

gjacoby126: CQS / CQSI is already enormous and not very coherent. Does this need to live here?

phoenix-core/src/it/java/org/apache/phoenix/end2end/CreateSchemaIT.java
ankitjain64: Sure. I did for couple of usages but I saw we are using HbaseAdmin at multiple places in CQSI and other classes. I will file another JIRA to use Admin rather than HBaseAdmin directly.

Also, one more thing that I came across is HConnection is also deprecated and Connection should be used in it's place. (https://hbase.apache.org/1.2/apidocs/org/apache/hadoop/hbase/client/HConnection.html)

Should I go-ahead and file a JIRA for that refactoring as well ?

phoenix-core/src/main/java/org/apache/phoenix/query/ConnectionQueryServicesImpl.java
ankitjain64: Added unit tests in ServerUtilTest class.

ankitjain64: Sure, moved it to ServerUtil class.

phoenix-core/src/it/java/org/apache/phoenix/end2end/SystemCatalogCreationOnConnectionIT.java
```diff
@@ -41,8 +41,8 @@
 import org.apache.hadoop.hbase.HBaseTestingUtility;
 import org.apache.hadoop.hbase.HConstants;
 import org.apache.hadoop.hbase.HTableDescriptor;
-import org.apache.hadoop.hbase.NamespaceNotFoundException;
 import org.apache.hadoop.hbase.TableName;
+import org.apache.hadoop.hbase.client.HBaseAdmin;
```
gjacoby126: Seems to be an unused import

PR #585, opened 2019-09-12
PHOENIX-5470 Tool to find view corruption

yanxinyi:
After 4.14, we enabled CHILD_LINK and changed the behavior of SYSCAT, so 4.15 branches have a different implementation. The logic as following:
1. get all tables to a list.
2. iterate all tables and find direct child/children from CHILD_LINK table.
3. for each child view, we compare SELECT COUNT column from SYSCAT + parent.COLUMN_COUNT with child view COLUMN_COUNT.

This groud buildup logic can help us find a view of corruption with pruning. If we have a corrupted global view with 1000 view on top of it, this algorithm will not waste of time to iterate 1000 views. Instead, it stops immediately and adds a corrupted view to the result list.


yanxinyi:
close this PR since I plan to implement this tool as part of phck.

PR #587, opened 2019-09-12
PHOENIX-5474 IndexTool should report the number of rows built

kadirozde:


kadirozde:
> Logic itself looks good -- could we have a test that verifies that the counters are incremented properly? (Might have to be two unit tests of the new Writable and the mapper since instrumenting MapReduce internals can be tough.)

@gjacoby126, I get the row count from the MR job and verify it in the new version of the PR. Would you please review it? Thanks

kadirozde:
> +1, thanks @kadirozde (And very nice solution to getting the test counter.)

Thank you for pointing out that we need to verify the counter reported by mappers.

gjacoby126:
JIRA is resolved, closing PR.

PR #591, opened 2019-10-01
PHOENIX-5503 IndexTool does not rebuild all the rows

kadirozde:


phoenix-core/src/main/java/org/apache/phoenix/mapreduce/index/IndexTool.java
```diff
@@ -312,7 +312,10 @@ public Job getJob() throws Exception {
                     return configureJobForAysncIndex();
                 }
                 else {
-                    //Local and non-transactional global indexes to be built on the server side
+                    // Local and non-transactional global indexes to be built on the server side
+                    // It is safe to unset CURRENT_SCN_VALUE for server side rebuilds to make sure that
+                    // all the rows that has been inserted so far will be rebuilt
+                    configuration.unset(PhoenixConfigurationUtil.CURRENT_SCN_VALUE);
```
priyankporwal: IMO, this configuration change should be inside configureJobForServerBuildIndex() so that any future calls to it do not miss it. I noticed that configureJobForServerBuildIndex() configuration.set() calls, so this would not be out of place.

kadirozde: Instead of moving this to configureJobForServerBuildIndex(), I restructured the code in this method to convince you that this is the right place to set or not to set SCN. Also, please consider that we may want to introduce the SCN option in future. In that case, it should be handled here.

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/index/IndexTool.java
```diff
@@ -297,24 +297,24 @@ void closeConnection() throws SQLException {
...
-                    return configureJobForAysncIndex();
-                }
-                else {
-                    //Local and non-transactional global indexes to be built on the server side
-                    return configureJobForServerBuildIndex();
-                }
+                return configureJobForAysncIndex();
+            }
+            else {
+                // Local and non-transactional global indexes to be built on the server side
```
gokceni: If this is a direct build, then SCN is not set?

kadirozde: Yes, if this is a direct build on non-transactional tables then SCN is not set

PR #592, opened 2019-10-02
PHOENIX-5505 Index read repair does not repair unverified rows with h…

kadirozde:
…igher timestamp

### DIFF
--- a/phoenix-core/src/it/java/org/apache/phoenix/end2end/index/GlobalIndexCheckerIT.java
+++ b/phoenix-core/src/it/java/org/apache/phoenix/end2end/index/GlobalIndexCheckerIT.java
@@ -30,6 +30,8 @@ import java.util.List;
 import java.util.Map;
 
 import com.google.common.collect.Maps;
+import org.apache.commons.logging.Log;
+import org.apache.commons.logging.LogFactory;
 import org.apache.phoenix.end2end.BaseUniqueNamesOwnClusterIT;
 import org.apache.phoenix.end2end.IndexToolIT;
 import org.apache.phoenix.hbase.index.IndexRegionObserver;
@@ -44,6 +46,7 @@ import com.google.common.collect.Lists;
 
 @RunWith(Parameterized.class)
 public class GlobalIndexCheckerIT extends BaseUniqueNamesOwnClusterIT {
+    private static final Log LOG = LogFactory.getLog(GlobalIndexCheckerIT.class);
     private final boolean async;
     private final String tableDDLOptions;
 
@@ -96,7 +99,7 @@ public class GlobalIndexCheckerIT extends BaseUniqueNamesOwnClusterIT {
     @Test
     public void testSkipPostIndexDeleteUpdate() throws Exception {
         String dataTableName = generateUniqueName();
-        populateTable(dataTableName);
+        populateTable(dataTableName); // with two rows ('a', 'ab', 'abc', 'abcd') and ('b', 'bc', 'bcd', 'bcde')
         Connection conn = DriverManager.getConnection(getUrl());
         String indexName = generateUniqueName();
         conn.createStatement().execute("CREATE INDEX " + indexName + " on " +
@@ -139,7 +142,7 @@ public class GlobalIndexCheckerIT extends BaseUniqueNamesOwnClusterIT {
     @Test
     public void testPartialRowUpdate() throws Exception {
         String dataTableName = generateUniqueName();
-        populateTable(dataTableName);
+        populateTable(dataTableName); // with two rows ('a', 'ab', 'abc', 'abcd') and ('b', 'bc', 'bcd', 'bcde')
         Connection conn = DriverManager.getConnection(getUrl());
         String indexName = generateUniqueName();
         conn.createStatement().execute("CREATE INDEX " + indexName + " on " +
@@ -180,7 +183,7 @@ public class GlobalIndexCheckerIT extends BaseUniqueNamesOwnClusterIT {
     @Test
     public void testSkipPostIndexPartialRowUpdate() throws Exception {
         String dataTableName = generateUniqueName();
-        populateTable(dataTableName);
+        populateTable(dataTableName); // with two rows ('a', 'ab', 'abc', 'abcd') and ('b', 'bc', 'bcd', 'bcde')
         Connection conn = DriverManager.getConnection(getUrl());
         String indexName = generateUniqueName();
         conn.createStatement().execute("CREATE INDEX " + indexName + " on " +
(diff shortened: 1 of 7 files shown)


phoenix-core/src/it/java/org/apache/phoenix/end2end/index/GlobalIndexCheckerIT.java
```diff
@@ -230,14 +233,27 @@ public void testSkipDataTableAndPostIndexPartialRowUpdate() throws Exception {
...
+        assertFalse(rs.next());
+        conn.createStatement().execute("upsert into " + dataTableName + " (id, val3) values ('a', 'abcdd')");
+        conn.commit();
+        selectSql =  "SELECT val2, val3 from " + dataTableName + " WHERE val1  = 'ab'";
+        // Verify that we will read from the first index table
+        assertExplainPlan(conn, selectSql, dataTableName, indexName + "1");
+        rs = conn.createStatement().executeQuery(selectSql);
+        assertTrue(rs.next());
+        assertEquals("abc", rs.getString(1));
         assertEquals("abcdd", rs.getString(2));
```
gokceni: Should not this value be abcd? Because the second and 3rd phase is failing?

kadirozde: They fail only for one upsert and then the behavior is reverted back to normal at line 235 after
IndexRegionObserver.setSkipDataTableUpdatesForTesting(false);
IndexRegionObserver.setSkipPostIndexUpdatesForTesting(false);
You need to look at the file not just the diff to see it all

phoenix-core/src/main/java/org/apache/phoenix/index/GlobalIndexChecker.java
```diff
@@ -247,9 +255,12 @@ private void repairIndexRows(byte[] indexRowKey, long ts, List<Cell> row) throws
                 buildIndexScan.setAttribute(BaseScannerRegionObserver.SKIP_REGION_BOUNDARY_CHECK, Bytes.toBytes(true));
             }
             // Rebuild the index rows from the corresponding the rows in the the data table
+            // Get the data row key from the index row key
             byte[] dataRowKey = indexMaintainer.buildDataRowKey(new ImmutableBytesWritable(indexRowKey), viewConstants);
+            // Rebuild the index rows from the data rows starting with the data row key
             buildIndexScan.withStartRow(dataRowKey, true);
-            buildIndexScan.setTimeRange(ts, maxTimestamp);
+            buildIndexScan.setTimeRange(0, maxTimestamp);
```
gjacoby126: I'm a bit confused by the startTime here. I'd have expected it to be either {0, Long.MAX_VALUE} to look at all rows, {0, EnvironmentEdgeManager.currentTimeMillis()} to look up to the current server timestamp, or {Math.max(0, minTimestamp), Math.min(maxTimestamp, Long.MAX_VALUE}, to look at the timestamps visible to the user Scan, where minTimestamp and maxTimestamp come from, but not a mixture of 0 and maxTimestamp. 

If we do honor the minTimestamp, we need to make sure not to delete from the index based on not finding a data row that might be hidden by the minTimestamp.

phoenix-core/src/main/java/org/apache/phoenix/index/GlobalIndexChecker.java
```diff
@@ -266,25 +277,55 @@ private void repairIndexRows(byte[] indexRowKey, long ts, List<Cell> row) throws
...
+                    return;
                 }
-                return;
+                // The index row is still "unverified" after rebuild. This means either that the data table row timestamp is
+                // lower than than the timestamp of the unverified index row (ts) and the index row that is built from
+                // the data table row is masked by this unverified row, or that the corresponding data table row does
+                // exist
+                // First delete the unverified row from index if it is old enough
+                deleteRowIfAgedEnough(indexRowKey, row, ts);
+                // Now we will do a single row scan to retrieve the verified index row build from the data table row
```
gjacoby126: Where are we verifying that singleRowIndexScan returns a _verified_ row? Isn't it possible to have two unverified index rows with the same key but different timestamps?

phoenix-core/src/main/java/org/apache/phoenix/index/GlobalIndexChecker.java
```diff
@@ -266,25 +277,55 @@ private void repairIndexRows(byte[] indexRowKey, long ts, List<Cell> row) throws
...
-                // Rewind the scanner and let the row be scanned again so that it can be repaired
-                scanner.close();
-                scanner = region.getScanner(indexScan);
+            if (indexRowExists) {
+                // This means there does not exist a data row for the unverified index row. Skip this row. To do that
+                // just return empty row.
                 row.clear();
+                return;
+            } else {
+                // This means the index row is deleted by the rebuild process and we got the next row
```
gjacoby126: nit: else that does nothing

phoenix-core/src/main/java/org/apache/phoenix/index/GlobalIndexChecker.java
```diff
@@ -266,25 +277,55 @@ private void repairIndexRows(byte[] indexRowKey, long ts, List<Cell> row) throws
...
+                ts = getMaxTimestamp(row);
             }
-            // This means the current index row is deleted by the rebuild process and we got the next row.
-            // If it is verified then we are good to go. If not, then we need to repair the new row
-            if (!verifyRowAndRemoveEmptyColumn(row)) {
-                // Rewind the scanner and let the row be scanned again so that it can be repaired
-                scanner.close();
-                scanner = region.getScanner(indexScan);
+            if (indexRowExists) {
+                // This means there does not exist a data row for the unverified index row. Skip this row. To do that
```
gjacoby126: if we found a previous, verified version up in singleRowIndexScan don't we need to return that instead? I assume "scanner"'s not raw so it would miss it if we just go next() elsewhere, right?

phoenix-core/src/main/java/org/apache/phoenix/index/GlobalIndexChecker.java
kadirozde: It is just to comment on the else case. It was intentional.

phoenix-core/src/main/java/org/apache/phoenix/index/GlobalIndexChecker.java
kadirozde: We do that within the while loop.

phoenix-core/src/main/java/org/apache/phoenix/index/GlobalIndexChecker.java
kadirozde: Yes, it is possible and thus we have the loop to cover it. I agree that the code can be optimized to get all versions using one scan. I will look into it.

phoenix-core/src/main/java/org/apache/phoenix/index/GlobalIndexChecker.java
kadirozde: The min and max timestamp are from the original scan on the index (set by the client). I did not want to honor the minTimestamp to cover one case. Assume that there are bunch of overwrites on the same row that made only to the index table within min and max timestamps. If we build index table using [minTimestamp, maxTimestamp], then the data table row will not be visible to buildIndexScan. Also assume that the compaction process has removed the good version of this index row (i.e., verified version). If so, we may end up deleting all the versions of the index row (if we start from minTimestamp).  I think Phoenix client will always set minTimestamp to zero or table create timestamp but I want to be on the safe side.

kadirozde:
> @kadirozde , some questions on the semantics. Also, this is complex with high cyclomatic complexity -- how difficult would a true unit test be with mocked scanners etc? The test changes above are good but I doubt they cover all the code paths.

I prefer real tests. I will try to add more tests to cover more cases.

phoenix-core/src/main/java/org/apache/phoenix/index/GlobalIndexChecker.java
kadirozde: It does not seem it is easy to get all versions in one scan. Phoenix tables are set with one row version currently. For them, readAllVersions() returns only one version. The readAllVersions() with setRaw(true) combination seems return all versions on disk but it also returns delete markers too. (Also I am not sure this is a consistent behavior for all HBase versions as I only checked it for the master branch.) This makes the code complex. Given that we are dealing with a specific failure case (data table write failure when overriding an existing row) , I do not think it is a good idea to optimize the code handling it

phoenix-core/src/it/java/org/apache/phoenix/end2end/index/GlobalIndexCheckerIT.java
kadirozde: @gjacoby126, I added more testing to cover the case where back to back data table row failures happen. Other failures cases were covered in GlobalIndexCheckerIT before. PHOENIX-5055 is to handle data write failures during overwrites.

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/BaseScannerRegionObserver.java
```diff
@@ -104,7 +104,6 @@
     public static final String PHYSICAL_DATA_TABLE_NAME = "_PhysicalDataTableName";
     public static final String EMPTY_COLUMN_FAMILY_NAME = "_EmptyCFName";
     public static final String EMPTY_COLUMN_QUALIFIER_NAME = "_EmptyCQName";
-    public static final String SCAN_LIMIT = "_ScanLimit";
```
kadirozde: @gjacoby126, I have removed the option for multiple index row repair. By default, the code was handling single row repair. The implementation of the multi-row repair was very basic and not effective. In order to simplify failure handling I have removed this option. We need to come up with a better design for it if we feel that we need it.

phoenix-core/src/main/java/org/apache/phoenix/index/GlobalIndexChecker.java
gjacoby126: Right, I get that there are corner cases where honoring minTimestamp would cause problems, so you're choosing 0. Why are you honoring maxTimestamp during the rebuild? Why should the SCN a client has chosen affect the rebuild semantics? (As opposed to the client's query results, which should of course be affected by SCN)

phoenix-core/src/main/java/org/apache/phoenix/index/GlobalIndexChecker.java
gjacoby126: From the diff, misunderstood where the while loop ended. I applied the patch to my local to reread it in my IDE and the logic makes sense.

phoenix-core/src/it/java/org/apache/phoenix/end2end/index/GlobalIndexCheckerIT.java
```diff
@@ -223,21 +226,52 @@ public void testSkipDataTableAndPostIndexPartialRowUpdate() throws Exception {
...
+        assertExplainPlan(conn, selectSql, dataTableName, indexName + "1");
+        rs = conn.createStatement().executeQuery(selectSql);
+        assertTrue(rs.next());
+        assertEquals("abc", rs.getString(1));
+        assertEquals("abcd", rs.getString(2));
+        assertFalse(rs.next());
+        selectSql =  "SELECT val2, val3 from " + dataTableName + " WHERE val2  = 'abcc'";
+        // Verify that we will read from the second index table
+        assertExplainPlan(conn, selectSql, dataTableName, indexName + "2");
+        rs = conn.createStatement().executeQuery(selectSql);
```
gjacoby126: Shouldn't this return a row because updating {a, abcc} -> {a, abccc} happened when we skipped the stage 2 and 3 updates above, and so the initial population on line 232 should still be live in index 2? 

(Also, this would be clearer if before each query you explained what you expected to find and why -- reverse engineering them requires very closely reading the entire function plus the populate helper function)

phoenix-core/src/main/java/org/apache/phoenix/index/GlobalIndexChecker.java
kadirozde: There is no reason to that. If we do that we would rebuild rows unnecessarily. Assume that there are N overwrite attempts on a particular row and all of them failed at phase 2 except the last overwrite. Also assume that just before the successful one, a client scans the index table. This scan will see the index row is unverified and will rebuilt the older version and will get the older successful version (not the last successful version), which is correct behavior. If we do not honor maxTimestamp then we would build both the old and new version but return to the client old version. In other words, we unnecessarily would build the newer versions even though they may not need to be rebuilt.

phoenix-core/src/it/java/org/apache/phoenix/end2end/index/GlobalIndexCheckerIT.java
kadirozde: We start with two rows ('a', 'ab', 'abc', 'abcd') and ('b', 'bc', 'bcd', 'bcde').
The we do three unsuccessful attempts to update the first row with different values (,, abcc,), (,,abccc, ), and (,,abcccc,). These attempts generate index rows with unverified status but do not update the data table. Therefore, the following select statement should return nothing
selectSql =  "SELECT val2, val3 from " + dataTableName + " WHERE val2  = 'abcc'";

phoenix-core/src/it/java/org/apache/phoenix/end2end/index/GlobalIndexCheckerIT.java
```diff
@@ -207,10 +210,120 @@ public void testSkipPostIndexPartialRowUpdate() throws Exception {
...
+            // Verify that we will read from the second index table
+            assertExplainPlan(conn, selectSql, dataTableName, indexTableName + "2");
+            rs = conn.createStatement().executeQuery(selectSql);
+            // Verify that one phase writes have no effect
+            assertFalse(rs.next());
+        }
+    }
+
     @Test
     public void testSkipDataTableAndPostIndexPartialRowUpdate() throws Exception {
```
gjacoby126: Name of test is now inaccurate since we no longer skip the data table and post index write -- this appears to just be testing the happy path where an update succeeds in all 3 phases on multiple indexes. (Which is a useful test, name should just reflect it.)

kadirozde: Great catch! I mistakenly removed the skip part. I will add it back.

kadirozde: Done.

gjacoby126:
Still +1 after @kadirozde 's last test tweak to put the skip phase 2 and 3 + do phantom write logic back in.

PR #593, opened 2019-10-07
PHOENIX-5508 - ALTER INDEX REBUILD removes all rows from a simple glo…

gjacoby126:
…bal index

### DIFF
--- a/phoenix-core/src/it/java/org/apache/phoenix/end2end/index/AlterIndexIT.java
+++ b/phoenix-core/src/it/java/org/apache/phoenix/end2end/index/AlterIndexIT.java
@@ -1,0 +1,73 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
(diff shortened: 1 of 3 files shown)


gjacoby126:
Note that the changes to PostDDLCompiler are just IDE refactoring the two giant anonymous classes to inner classes to make it easier to understand the code flow.

phoenix-core/src/it/java/org/apache/phoenix/end2end/index/AlterIndexIT.java
```diff
@@ -0,0 +1,73 @@
...
+import org.junit.Test;
+
+import java.sql.Connection;
+import java.sql.DriverManager;
+import java.sql.SQLException;
+
+public class AlterIndexIT extends ParallelStatsDisabledIT {
+
+    @Test
+    public void testAlterIndexRebuildNoAsync() throws Exception {
```
gokceni: Do you want to parametrize this test and run with async = true and false parameters?

PR #597, opened 2019-10-11
PHOENIX-5522 If index is disabled, IndexUpgradeTool should not fail d…

gokceni:
…uring indexTool run

### DIFF
--- a/phoenix-core/src/it/java/org/apache/phoenix/end2end/ParameterizedIndexUpgradeToolIT.java
+++ b/phoenix-core/src/it/java/org/apache/phoenix/end2end/ParameterizedIndexUpgradeToolIT.java
@@ -30,7 +30,10 @@ import org.apache.phoenix.mapreduce.index.IndexUpgradeTool;
 import org.apache.phoenix.query.BaseTest;
 import org.apache.phoenix.query.ConnectionQueryServices;
 import org.apache.phoenix.query.QueryServices;
+import org.apache.phoenix.util.PhoenixRuntime;
+import org.apache.phoenix.util.PropertiesUtil;
 import org.apache.phoenix.util.ReadOnlyProps;
+import org.apache.phoenix.util.SchemaUtil;
 import org.junit.After;
 import org.junit.Assert;
 import org.junit.Before;
@@ -56,6 +59,7 @@ import java.util.UUID;
 
 import static org.apache.phoenix.mapreduce.index.IndexUpgradeTool.ROLLBACK_OP;
 import static org.apache.phoenix.mapreduce.index.IndexUpgradeTool.UPGRADE_OP;
+import static org.apache.phoenix.util.TestUtil.TEST_PROPERTIES;
 
 @RunWith(Parameterized.class)
 @Category(NeedsOwnMiniClusterTest.class)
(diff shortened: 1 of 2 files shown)


phoenix-core/src/main/java/org/apache/phoenix/mapreduce/index/IndexUpgradeTool.java
```diff
@@ -491,6 +491,15 @@ private int rebuildIndexes(String dataTable, IndexTool indexingTool) {
             try {
                 LOGGER.info("Rebuilding index: " + String.join(",", args));
                 if (!dryRun) {
+                    // If the index is in DISABLED state, indexTool will fail. First to ALTER REBUILD ASYNC
+                    String stmt = String.format("ALTER INDEX %s ON %s REBUILD ASYNC", indexName,
+                            SchemaUtil.getTableName(schema, baseTable));
+                    try {
+                        conn.createStatement().execute(stmt);
+                    } catch (SQLException ex) {
+                        LOGGER.warning("Something went wrong while running ALTER REBUILD ASYNC the index "
```
gjacoby126: nit: "ASYNC on the index "

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/index/IndexUpgradeTool.java
```diff
@@ -491,6 +491,15 @@ private int rebuildIndexes(String dataTable, IndexTool indexingTool) {
             try {
                 LOGGER.info("Rebuilding index: " + String.join(",", args));
                 if (!dryRun) {
+                    // If the index is in DISABLED state, indexTool will fail. First to ALTER REBUILD ASYNC
```
gjacoby126: nit: Good for the comment to make clear that ALTER INDEX REBUILD ASYNC will set the index state to BUILDING, which is safe to make ACTIVE later.

phoenix-core/src/it/java/org/apache/phoenix/end2end/ParameterizedIndexUpgradeToolIT.java
```diff
@@ -221,25 +235,25 @@ private void checkOldIndexingCoprocessors(String [] indexList, String [] tableLi
...
-            {true, false, true},
-            {false, true, true},
-            {true, true, true},
-            {false, false, false},
-            {true, false, false},
-            {false, true, false},
-            {true, true, false}
-        });
+        return Arrays.asList(new Object[][] { { false, false, true, true },
+                { false, false, true, false }, { true, false, true, true },
```
gjacoby126: how long do all of these take to run now? Are there combinations that are redundant (or orthogonal and can be tested by a single test)?

priyankporwal: +1 .. The test matrix keeps getting doubled for every variant. Perhaps mix these up with some random combination of isNamespaceEnbaled.

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/index/IndexUpgradeTool.java
gokceni: done

phoenix-core/src/it/java/org/apache/phoenix/end2end/ParameterizedIndexUpgradeToolIT.java
gokceni: What do you think @kadirozde ? Should I remove the combination and make just 1 index disabled?

kadirozde: The test already takes 20 min so we should not increase its time any further. Also, we can eliminate some of the existing combinations to reduce the test time as 20 min is already too long for this test.

phoenix-core/src/it/java/org/apache/phoenix/end2end/ParameterizedIndexUpgradeToolIT.java
```diff
@@ -111,18 +119,13 @@ public void setup () throws Exception {
...
-                || Boolean.toString(!isNamespaceEnabled).equals(serverProps
-                .get(QueryServices.IS_NAMESPACE_MAPPING_ENABLED))) {
-            tearDownMiniClusterAsync(1);
-        }
+        tearDownMiniClusterAsync(1);
+
         //setting up properties for namespace
         clientProps.put(QueryServices.IS_NAMESPACE_MAPPING_ENABLED,
-                Boolean.toString(isNamespaceEnabled));
+                Boolean.toString(true));
```
priyankporwal: Namespaces are not enabled by default, right? Should we set these to false instead?

gokceni: Sure but I think it is good to test with namespace just to cover that too. We need to pick either. I thought this is a safer approach to test that scenario too.

priyankporwal: I am not sure which ones gives better coverage. If we are not sure, we can leave the isNameSpaceEnabled and set it to true/false on the rows from lines 235-238, as the last values in tuples: e.g. something like this.
            {false, false, true},
            {true, false, false},
            {false, true, false},
            {true, true, true}

PR #598, opened 2019-10-11
PHOENIX-5515 Able to write indexed value to data table without writin…

kadirozde:
…g to index table

### DIFF
--- a/phoenix-core/src/it/java/org/apache/phoenix/end2end/index/GlobalIndexCheckerIT.java
+++ b/phoenix-core/src/it/java/org/apache/phoenix/end2end/index/GlobalIndexCheckerIT.java
@@ -20,6 +20,7 @@ package org.apache.phoenix.end2end.index;
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertFalse;
 import static org.junit.Assert.assertTrue;
+import static org.junit.Assert.fail;
 
 import java.sql.Connection;
 import java.sql.DriverManager;
@@ -37,6 +38,7 @@ import org.apache.phoenix.end2end.IndexToolIT;
 import org.apache.phoenix.hbase.index.IndexRegionObserver;
 import org.apache.phoenix.util.QueryUtil;
 import org.apache.phoenix.util.ReadOnlyProps;
+import org.junit.After;
 import org.junit.BeforeClass;
 import org.junit.Test;
 import org.junit.runner.RunWith;
@@ -78,6 +80,13 @@ public class GlobalIndexCheckerIT extends BaseUniqueNamesOwnClusterIT {
         return list;
     }
 
+    @After
+    public void unsetFailForTesting() {
+        IndexRegionObserver.setFailPreIndexUpdatesForTesting(false);
+        IndexRegionObserver.setFailDataTableUpdatesForTesting(false);
+        IndexRegionObserver.setFailPostIndexUpdatesForTesting(false);
+    }
+
     public static void assertExplainPlan(Connection conn, String selectSql,
                                          String dataTableFullName, String indexTableFullName) throws SQLException {
         ResultSet rs = conn.createStatement().executeQuery("EXPLAIN " + selectSql);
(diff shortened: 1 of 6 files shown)


phoenix-core/src/it/java/org/apache/phoenix/end2end/index/GlobalIndexCheckerIT.java
```diff
@@ -317,50 +358,94 @@ public void testOnePhaseOverwrite() throws Exception {
...
-        conn.commit();
-        IndexRegionObserver.setSkipDataTableUpdatesForTesting(false);
-        IndexRegionObserver.setSkipPostIndexUpdatesForTesting(false);
-        conn.createStatement().execute("upsert into " + dataTableName + " (id, val3) values ('a', 'abcdd')");
+    }
+
+    static private void commitWithException(Connection conn) {
+        try {
+            conn.commit();
+            IndexRegionObserver.setFailPreIndexUpdatesForTesting(false);
```
gjacoby126: nit: could be useful to have these three lines setting to false in an After method to guarantee that they're always reset between tests.

phoenix-core/src/main/java/org/apache/phoenix/hbase/index/write/TrackingParallelWriterIndexCommitter.java
```diff
@@ -242,8 +243,13 @@ private void throwFailureIfDone() throws SingleIndexWriteFailureException {
         // if any of the tasks failed, then we need to propagate the failure
         if (failures.size() > 0) {
             // make the list unmodifiable to avoid any more synchronization concerns
-            throw new MultiIndexWriteFailureException(Collections.unmodifiableList(failures),
+            MultiIndexWriteFailureException exception = new MultiIndexWriteFailureException(Collections.unmodifiableList(failures),
                     disableIndexOnFailure && PhoenixIndexFailurePolicy.getDisableIndexOnFailure(env));
+            if (disableIndexOnFailure)
+                throw exception;
+            else {
+                throw new DoNotRetryIOException(exception.getMessage());
```
gjacoby126: Shouldn't we wrap the MIWFException in the DNRIE so that higher up can see the stack trace / any inner exceptions rather than just copying the message?

phoenix-core/src/main/java/org/apache/phoenix/hbase/index/write/IndexWriter.java
```diff
@@ -131,25 +131,6 @@ public IndexWriter(IndexCommitter committer, IndexFailurePolicy policy,
...
-   * This method is blocking and could potentially cause the writer to block for a long time as we
-   * write the index updates. When we return depends on the specified {@link IndexCommitter}.
-   * <p>
-   * If update fails, we pass along the failure to the installed {@link IndexFailurePolicy}, which
-   * then decides how to handle the failure.
-   * @param indexUpdates Updates to write
- * @param clientVersion version of the client
- * @throws IOException 
-   */
-    public void writeAndHandleFailure(Collection<Pair<Mutation, byte[]>> indexUpdates,
```
gjacoby126: What about indexes using the old design? Shouldn't they get disabled / use the old failure policies until they're upgraded?

kadirozde: This method was only used by the new design. So, I removed it as it was not necessary anymore.

phoenix-core/src/main/java/org/apache/phoenix/hbase/index/write/TrackingParallelWriterIndexCommitter.java
kadirozde: I will do that.

phoenix-core/src/it/java/org/apache/phoenix/end2end/index/GlobalIndexCheckerIT.java
kadirozde: It sounds good. I will do that.

kadirozde: Done.

phoenix-core/src/main/java/org/apache/phoenix/hbase/index/write/TrackingParallelWriterIndexCommitter.java
kadirozde: Done

vincentpoon:
+1

PR #599, opened 2019-10-18
Create columns in SYSCAT and fields in PTable.proto for view TTL support

jpisaac:


jpisaac:
@ChinmaySKulkarni @yanxinyi

phoenix-core/src/it/java/org/apache/phoenix/end2end/ViewTTLIT.java
```diff
@@ -0,0 +1,111 @@
+package org.apache.phoenix.end2end;
+
+import com.google.common.collect.Lists;
+import org.apache.hadoop.hbase.HConstants;
+import org.apache.hadoop.hbase.client.Result;
+import org.apache.hadoop.hbase.client.ResultScanner;
+import org.apache.hadoop.hbase.client.Scan;
+import org.apache.hadoop.hbase.client.Table;
+import org.apache.hadoop.hbase.filter.*;
```
ChinmaySKulkarni: Avoid wildcard imports. See https://phoenix.apache.org/contributing.html for Contribution guidelines. You can add preference file to your IDE.

phoenix-core/src/it/java/org/apache/phoenix/end2end/ViewTTLIT.java
```diff
@@ -0,0 +1,111 @@
...
+            allRows.setTimeRange(minTimestamp, HConstants.LATEST_TIMESTAMP);
+            allRows.setFilter(filterList);
+            ResultScanner scanner = tbl.getScanner(allRows);
+            int numMatchingRows = 0;
+            for (Result result = scanner.next(); result != null; result = scanner.next()) {
+                numMatchingRows +=
+                        result.containsColumn(QueryConstants.DEFAULT_COLUMN_FAMILY_BYTES,
+                                PhoenixDatabaseMetaData.VIEW_TTL_BYTES) ? 1 : 0;
+            }
+            assertEquals(String.format("Expected rows do match for table = %s at timestamp %d",
```
ChinmaySKulkarni: *do -> don't

phoenix-core/src/it/java/org/apache/phoenix/end2end/ViewTTLIT.java
```diff
@@ -0,0 +1,111 @@
...
+                    Bytes.toString(PhoenixDatabaseMetaData.SYSTEM_CATALOG_NAME_BYTES), minTimestamp), expectedRows, numMatchingRows);
+        }
+
+    }
+
+    /**
+         * -----------------
+         * Test methods
```
ChinmaySKulkarni: nit: Fix indentation

phoenix-core/src/it/java/org/apache/phoenix/end2end/ViewTTLIT.java
```diff
@@ -0,0 +1,111 @@
...
+        tableOptions.getTableColumns().clear();
+        tableOptions.getTableColumnTypes().clear();
+
+        final PhoenixTestBuilder.SchemaBuilder schemaBuilder = new PhoenixTestBuilder.SchemaBuilder(getUrl());
+        schemaBuilder.withTableOptions(tableOptions).build();
+
+        // Define the test data.
+        PhoenixTestBuilder.DataSupplier dataSupplier = new PhoenixTestBuilder.DataSupplier() {
+
+            final String
```
ChinmaySKulkarni: nit: unnecessary line break

phoenix-core/src/it/java/org/apache/phoenix/end2end/ViewTTLIT.java
```diff
@@ -0,0 +1,111 @@
...
+                    orgId =
+                    String.format("00D0x000%s", schemaBuilder.getDataOptions().getUniqueName());
+            final String kp = SchemaUtil.normalizeIdentifier(schemaBuilder.getEntityKeyPrefix());
+
+            @Override public List<Object> getValues(int rowIndex) {
+                String id = String.format("00A0y000%07d", rowIndex);
```
ChinmaySKulkarni: What is this logic for getting the `id`?

phoenix-core/src/it/java/org/apache/phoenix/end2end/ViewTTLIT.java
```diff
@@ -0,0 +1,111 @@
...
+    }
+
+    /**
+         * -----------------
+         * Test methods
+         * -----------------
+         */
+
+    @Test
+    public void testWithBasicTenantViewWithNoViewTTLDefined() throws Exception {
```
ChinmaySKulkarni: Do we also want to add a test for a case where we specify a viewTTL?

phoenix-core/src/main/java/org/apache/phoenix/jdbc/PhoenixDatabaseMetaData.java
```diff
@@ -366,7 +366,12 @@
     public static final byte[] COLUMN_QUALIFIER_COUNTER_BYTES = Bytes.toBytes(COLUMN_QUALIFIER_COUNTER);
     public static final String USE_STATS_FOR_PARALLELIZATION = "USE_STATS_FOR_PARALLELIZATION";
     public static final byte[] USE_STATS_FOR_PARALLELIZATION_BYTES = Bytes.toBytes(USE_STATS_FOR_PARALLELIZATION);
-    
+
+    public static final String VIEW_TTL = "VIEW_TTL";
+    public static final byte[] VIEW_TTL_BYTES = Bytes.toBytes(VIEW_TTL);
+    public static final String VIEW_TTL_HWM = "VIEW_TTL_HWM";
```
ChinmaySKulkarni: Can you add a comment here about what the HighWatermark column is used for?

phoenix-core/src/main/java/org/apache/phoenix/schema/MetaDataClient.java
```diff
@@ -2112,6 +2116,17 @@ private PTable createTableInternal(CreateTableStatement statement, byte[][] spli
...
+
+            Integer viewTTL = null;
+            Long viewTTLHighWaterMark = null;
+            Integer viewTTLProp = null;
+
+            //TODO Look at the VIEW_TTL table property values that are set instead of hardcoded defaults.
+            if (!SchemaUtil.isSystemTable(Bytes.toBytes(SchemaUtil.getTableName(schemaName, tableName))) && viewTTLProp != null) {
+                viewTTL = HConstants.FOREVER;
+                viewTTLHighWaterMark = 0L;
+            }
```
ChinmaySKulkarni: Also, we would need some sanity checks for the passed in viewTTL and viewTTLHWM properties, for example: cannot be negative, some possible max value, etc.

phoenix-core/src/it/java/org/apache/phoenix/end2end/ViewTTLIT.java
ChinmaySKulkarni: I guess that will be a TODO..?

phoenix-core/src/it/java/org/apache/phoenix/end2end/ViewTTLIT.java
```diff
@@ -0,0 +1,111 @@
+package org.apache.phoenix.end2end;
```
yanxinyi: nit: apache driver

phoenix-core/src/main/java/org/apache/phoenix/schema/MetaDataClient.java
```diff
@@ -311,8 +313,10 @@
                     IMMUTABLE_STORAGE_SCHEME + "," +
                     ENCODING_SCHEME + "," +
                     USE_STATS_FOR_PARALLELIZATION +"," +
-                    VIEW_INDEX_ID_DATA_TYPE +
-                    ") VALUES (?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?)";
+                    VIEW_INDEX_ID_DATA_TYPE +"," +
```
yanxinyi: I don't understand the logic of adding a new column to SYSTEM.CATALOG table. for the server-side update, do we drop the table and re-create a table?

ChinmaySKulkarni:
@jpisaac @yanxinyi One thing to keep in mind is, this requires a schema change for SYSTEM.CATALOG, so we should handle this in the `upgradeSystemTables` path as well. 

Also, we only do such schema upgrades during a major or minor version upgrade i.e. 4.x -> 4.(x+1). Now since this isn't in 4.15.0, we can't push it into 4.15.1 either, otherwise 4.15.0 clients won't get the new SYSCAT schema. Which means, this needs to go into 4.16.0.

In case we want this to go into 4.15.x, one (ugly) way to avoid this is to add a tool that adds these extra columns and document that users need to run this before being able to use a 4.15. This will be mandatory probably since this affects the `createTable` and other client side code which expects these columns to be present in SYSTEM.CATALOG.

phoenix-core/src/it/java/org/apache/phoenix/end2end/ViewTTLIT.java
jpisaac: Nothing special! Just trying to get char(15) id, just using the sfdc id format.

phoenix-core/src/it/java/org/apache/phoenix/end2end/ViewTTLIT.java
jpisaac: Yeah, planning to do that as part of my next PR - which is tying in the actual property value specified during the create/alter command.

phoenix-core/src/main/java/org/apache/phoenix/schema/MetaDataClient.java
jpisaac: The view-ttl properties will be stored/persisted as part of the view header columns.

phoenix-core/src/it/java/org/apache/phoenix/end2end/ViewTTLIT.java
ChinmaySKulkarni: Can you add a comment about the format of the generated key here?

yanxinyi: Maybe move this 15 char format to `Util` and named some method like `getRowIndexFormat` so that we can reuse it and have better code readability?

yanxinyi:
LGTM, but I will let @ChinmaySKulkarni to confirm the changes and meger the PR.

phoenix-core/src/it/java/org/apache/phoenix/end2end/ViewTTLIT.java
jpisaac: For the time being, created a static for it.

jpisaac:
@kadirozde This is 1 of n commits for this JIRA https://issues.apache.org/jira/browse/PHOENIX-5501.

Chunking it to smaller PRs so that it is easier for review.

jpisaac:
With the latest commit
- We should be able to set VIEW_TTL TableProperty during CREATE/ALTER table commands.
- Should be able to persist TTL values specified during CREATE/ALTER table commands.
- Ensure that the VIEW_TTL property is set only on the view (no view indexes, parent tables)
- Ensure that when the VIEW_TTL property is set on the view,  the parent table TTL property is not set
- Ensure the view_indexes for this view inherit the VIEW_TTL property
- Ensure that when view VIEW_TTL property is set the child views that have overridden the property do not inherit the new values.

phoenix-core/src/it/java/org/apache/phoenix/end2end/ViewTTLIT.java
```diff
@@ -87,6 +99,27 @@ private void assertViewHeaderRowsHaveViewTTLRelatedCells(String schemaName, long
 
     }
 
+    private void assertSyscatHaveViewTTLRelatedColumns(String tenantId, String schemaName, String tableName, String tableType, long ttlValueExpected)
```
yanxinyi: nin: more than 100 char per line

phoenix-core/src/it/java/org/apache/phoenix/end2end/ViewTTLIT.java
```diff
@@ -107,34 +140,258 @@ public void testWithBasicGlobalViewWithNoViewTTLDefined() throws Exception {
...
+        TenantViewOptions tenantViewOptions = TenantViewOptions.withDefaults();
+        tenantViewOptions.setTableProps("VIEW_TTL=1000");
+        try {
+            schemaBuilder
+                    .withTableOptions(tableOptions)
+                    .withTenantViewOptions(tenantViewOptions)
+                    .buildNewView();
+            fail();
+        } catch (SQLException e) {
+            assertEquals(SQLExceptionCode.CANNOT_SET_OR_ALTER_VIEW_TTL_FOR_TABLE_WITH_TTL.getErrorCode(), e.getErrorCode());
```
yanxinyi: what if we don't get exception?  Maybe using a flag to indicate we do have this exception?

phoenix-core/src/it/java/org/apache/phoenix/end2end/ViewTTLIT.java
```diff
@@ -107,34 +140,258 @@ public void testWithBasicGlobalViewWithNoViewTTLDefined() throws Exception {
...
+        tenantViewIndexOptions.setIndexProps("VIEW_TTL=1000");
+        try {
+            schemaBuilder
+                    .withTableOptions(tableOptions)
+                    .withTenantViewDefaults()
+                    .withTenantViewIndexOptions(tenantViewIndexOptions)
+                    .buildNewView();
+            fail();
+        } catch (SQLException e) {
+            assertEquals(SQLExceptionCode.VIEW_TTL_SUPPORTED_FOR_VIEWS_ONLY.getErrorCode(), e.getErrorCode());
```
yanxinyi: same as above, we need to make sure we do execute the exception code block

phoenix-core/src/it/java/org/apache/phoenix/end2end/ViewTTLIT.java
```diff
@@ -107,34 +140,258 @@ public void testWithBasicGlobalViewWithNoViewTTLDefined() throws Exception {
...
+        // Define the test schema.
+        // 1. Table with default columns => (ORG_ID, KP, COL1, COL2, COL3), PK => (ORG_ID, KP)
+        // 2. Tenant with default columns => (ZID, COL7, COL8, COL9), PK => (ZID)
+        final PhoenixTestBuilder.SchemaBuilder schemaBuilder = new PhoenixTestBuilder.SchemaBuilder(getUrl());
+
+        TableOptions tableOptions = TableOptions.withDefaults();
+        tableOptions.setTableProps("COLUMN_ENCODED_BYTES=0,MULTI_TENANT=true");
+
+        TenantViewOptions tenantViewOptions = TenantViewOptions.withDefaults();
+        tenantViewOptions.setTableProps("VIEW_TTL=0");
```
yanxinyi: do we treat `TTL=0` for never expire or data expired right away?

phoenix-core/src/test/java/org/apache/phoenix/query/PhoenixTestBuilder.java
```diff
@@ -368,7 +368,7 @@ public void upsertRow(int rowIndex) throws SQLException {
      * for specific test requirements.
      */
     public static class SchemaBuilder {
-        private static final AtomicInteger TENANT_COUNTER = new AtomicInteger(0);
+        private static final AtomicInteger TENANT_COUNTER = new AtomicInteger(7);
```
yanxinyi: is there a specific reason that we want to have an initial integer `7` instead of `0`?

phoenix-core/src/main/java/org/apache/phoenix/schema/MetaDataClient.java
yanxinyi: I was asking how to upgrade client and server side changes because it requires a new release or tool to doing this.

phoenix-core/src/main/java/org/apache/phoenix/exception/SQLExceptionCode.java
```diff
@@ -322,6 +322,10 @@ public SQLException newException(SQLExceptionInfo info) {
             + MetaDataUtil.SYNCED_DATA_TABLE_AND_INDEX_COL_FAM_PROPERTIES.toString()),
     CANNOT_SET_OR_ALTER_UPDATE_CACHE_FREQ_FOR_INDEX(10950, "44A31", "Cannot set or alter "
             + PhoenixDatabaseMetaData.UPDATE_CACHE_FREQUENCY + " on an index"),
+    VIEW_TTL_SUPPORTED_FOR_VIEWS_ONLY(10951, "44A32", PhoenixDatabaseMetaData.VIEW_TTL
```
ChinmaySKulkarni: Do you want to mention immutable/mutable view constraint in the exception too?

phoenix-core/src/main/java/org/apache/phoenix/schema/MetaDataClient.java
ChinmaySKulkarni: As I mentioned before, we will need to add steps in the upgrade path `CQSI.upgradeSystemTables` to ensure that new columns are added to syscat when a client upgrades from 4.x metadata to 4.(x+1), assuming the VIEW_TTL feature goes into 4.(x+1)

ChinmaySKulkarni: OR, we will need to create a new tool like the IndexUpgradeTool to do the same syscat table alteration in order to enable using this feature for a patch upgrade i.e. 4.x.y -> 4.x.(y+1)

phoenix-core/src/main/java/org/apache/phoenix/schema/MetaDataClient.java
```diff
@@ -2650,6 +2685,8 @@ else if (!SchemaUtil.isSystemTable(Bytes.toBytes(SchemaUtil.getTableName(schemaN
                         .setIndexes(Collections.<PTable>emptyList())
                         .setPhysicalNames(ImmutableList.<PName>of())
                         .setColumns(columns.values())
+                        .setViewTTL(0L)
```
ChinmaySKulkarni: Why are we setting it to 0L? Shouldn't we set it to `viewTTL` extracted above?
Also, can we extract `0L` as a special constant value with a comment on what it indicates (never expire?)

phoenix-core/src/main/java/org/apache/phoenix/schema/MetaDataClient.java
```diff
@@ -4896,6 +4954,20 @@ else if (metaProperties.getImmutableStorageSchemeProp() != table.getImmutableSto
                 metaProperties.setNonTxToTx(true);
             }
         }
+
+        if (metaProperties.getViewTTL() != null) {
+            if (table.getType() == PTableType.INDEX) {
```
ChinmaySKulkarni: Shouldn't the condition be `if(table.getType() != PTableType.VIEW)`?

phoenix-core/src/test/java/org/apache/phoenix/query/PhoenixTestBuilder.java
jpisaac: Oh!!, It was a leftover from my testing, will reset it to 0

phoenix-core/src/main/java/org/apache/phoenix/schema/MetaDataClient.java
jpisaac: My understanding is this is special handling for SYSTEM table as the comment above says. We do not want to set TTL for any SYSTEM tables. 
But your comment about extracting 0L as a special constant makes sense. I will do that.

phoenix-core/src/main/java/org/apache/phoenix/schema/MetaDataClient.java
jpisaac: evaluateStmtProperties() is typically called by alterIndex and addColumn, but no harm in making it more general.

phoenix-core/src/it/java/org/apache/phoenix/end2end/ViewTTLIT.java
```diff
@@ -0,0 +1,397 @@
...
+                .buildWithNewTenant();
+
+        String tenantId = schemaBuilder.getDataOptions().getTenantId();
+        String schemaName = stripQuotes(SchemaUtil.getSchemaNameFromFullName(schemaBuilder.getEntityTenantViewName()));
+        String globalViewName = stripQuotes(SchemaUtil.getTableNameFromFullName(schemaBuilder.getEntityGlobalViewName()));
+        String tenantViewName = stripQuotes(SchemaUtil.getTableNameFromFullName(schemaBuilder.getEntityTenantViewName()));
+        String indexOnGlobalViewName = String.format("IDX_%s", globalViewName);
+        String indexOnTenantViewName = String.format("IDX_%s", stripQuotes(schemaBuilder.getEntityKeyPrefix()));
+
+
```
ChinmaySKulkarni: nit: Remove extra newlines.

phoenix-core/src/it/java/org/apache/phoenix/end2end/ViewTTLIT.java
ChinmaySKulkarni: @jpisaac can you please address Xinyi's comment above?

phoenix-core/src/it/java/org/apache/phoenix/end2end/ViewTTLIT.java
```diff
@@ -0,0 +1,397 @@
...
+        // Define the test schema.
+        // 1. Table with default columns => (ORG_ID, KP, COL1, COL2, COL3), PK => (ORG_ID, KP)
+        // 2. Tenant with default columns => (ZID, COL7, COL8, COL9), PK => (ZID)
+        final PhoenixTestBuilder.SchemaBuilder schemaBuilder = new PhoenixTestBuilder.SchemaBuilder(getUrl());
+
+        TableOptions tableOptions = TableOptions.withDefaults();
+        tableOptions.setTableProps("COLUMN_ENCODED_BYTES=0,MULTI_TENANT=true");
+
+        TenantViewOptions tenantViewOptions = TenantViewOptions.withDefaults();
+        tenantViewOptions.setTableProps("VIEW_TTL=0");
```
ChinmaySKulkarni: same here

phoenix-core/src/main/java/org/apache/phoenix/schema/MetaDataClient.java
```diff
@@ -1987,10 +1993,37 @@ private PTable createTableInternal(CreateTableStatement statement, byte[][] spli
             int baseTableColumnCount =
                     tableType == PTableType.VIEW ? parent.getColumns().size()
                             : QueryConstants.BASE_TABLE_BASE_COLUMN_COUNT;
+
+            Long viewTTL = VIEW_TTL_NOT_DEFINED;
+            Long viewTTLHighWaterMark = MIN_VIEW_TTL_HWM;
+            Long viewTTLProp = (Long) TableProperty.VIEW_TTL.getValue(tableProps);;
+
+            if (tableType == VIEW && viewTTLProp != null && parentPhysicalName != null) {
```
ChinmaySKulkarni: @jpisaac Shouldn't we also do some checks on the actual value of the `viewTTLProp` provided by the user, such as checking that it is non-negative, other bounds, etc.?

phoenix-core/src/main/java/org/apache/phoenix/schema/TableProperty.java
```diff
@@ -222,6 +224,28 @@ public Object getValue(Object value) {
...
+    },
+
+    VIEW_TTL(PhoenixDatabaseMetaData.VIEW_TTL, true, true, true) {
+        @Override
+        public Object getValue(Object value) {
+            if (value instanceof String) {
+                String strValue = (String) value;
+                if ("FOREVER".equalsIgnoreCase(strValue)) {
+                    return HConstants.LATEST_TIMESTAMP;
+                } else if ("NONE".equalsIgnoreCase(strValue)) {
```
ChinmaySKulkarni: @jpisaac What about passing in zero as viewTTL? What would that mean? Would it have any meaning at all or some special meaning?

ChinmaySKulkarni:
@jpisaac @yanxinyi Will there be any problems making this work for views with indexes? OR, views with views ... (with or without indexes)? I guess for the MR delete job, you would be issuing phoenix deletes rather than HBase deletes and so index updates should be taken care of automatically?

jpisaac:
> @jpisaac @yanxinyi Will there be any problems making this work for views with indexes? OR, views with views ... (with or without indexes)? I guess for the MR delete job, you would be issuing phoenix deletes rather than HBase deletes and so index updates should be taken care of automatically?

@ChinmaySKulkarni 
The current thinking is to issue separate "select count(1) from <view>/<view-index>" and inject a special attribute DELETE_TTL_EXPIRED with value=true. The coprocessor will intercept it and then issue point deletes for the selected rows if expired. We can discuss this at length (pros and cons) in the design doc.

phoenix-core/src/main/java/org/apache/phoenix/schema/MetaDataClient.java
```diff
@@ -1997,26 +1997,35 @@ private PTable createTableInternal(CreateTableStatement statement, byte[][] spli
             Long viewTTL = VIEW_TTL_NOT_DEFINED;
             Long viewTTLHighWaterMark = MIN_VIEW_TTL_HWM;
             Long viewTTLProp = (Long) TableProperty.VIEW_TTL.getValue(tableProps);;
+            // Validate VIEW_TTL prop value if set
+            if (viewTTLProp != null) {
+                if (viewTTLProp < 0) {
```
ChinmaySKulkarni: @jpisaac I guess there is no upper-bound?

ChinmaySKulkarni: Wait, upper bound is `HConstants.LATEST_TIMESTAMP`. Is it possible to assign a larger value than that from the client?

jpisaac: HConstants.LATEST_TIMESTAMP = Long.MAX_VALUE

ChinmaySKulkarni:
@jpisaac Overall lgtm. Please squash all commits and let's get a Hadoop QA run on these changes.

PR #600, opened 2019-10-20
PHOENIX-5535 Index rebuilds via UngroupedAggregateRegionObserver shou…

kadirozde:
…ld replay delete markers

### DIFF
--- a/phoenix-core/src/it/java/org/apache/phoenix/end2end/IndexToolIT.java
+++ b/phoenix-core/src/it/java/org/apache/phoenix/end2end/IndexToolIT.java
@@ -29,6 +29,7 @@ import java.sql.DriverManager;
 import java.sql.PreparedStatement;
 import java.sql.ResultSet;
 import java.sql.SQLException;
+import java.sql.Types;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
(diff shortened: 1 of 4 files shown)


phoenix-core/src/it/java/org/apache/phoenix/end2end/IndexToolIT.java
```diff
@@ -251,6 +252,61 @@ public void testSecondaryIndex() throws Exception {
...
+
+            // Run the index MR job and verify that the index table is built correctly
+            IndexTool indexTool = runIndexTool(directApi, useSnapshot, schemaName, dataTableName, indexTableName, null, 0, new String[0]);
+            assertEquals(NROWS, indexTool.getJob().getCounters().findCounter(INPUT_RECORDS).getValue());
+            long actualRowCount = IndexScrutiny.scrutinizeIndex(conn, dataTableFullName, indexTableFullName);
+            assertEquals(NROWS, actualRowCount);
+        }
+    }
+
+
```
gjacoby126: Do we need a test that shows that NULL rebuilds work after a major compaction if we pass in the deleteall flag to the index tool, or is that not worth the time it would take to run?

kadirozde: @gjacoby126, I am not sure if have seen my comments on the JIRA. I have pointed out that this patch is not sufficient as compaction will remove delete markers. Would you please read my comments there and reply?

phoenix-core/src/it/java/org/apache/phoenix/end2end/IndexToolIT.java
```diff
@@ -251,6 +252,61 @@ public void testSecondaryIndex() throws Exception {
...
+            String stmString1 =
+                    "CREATE TABLE " + dataTableFullName
+                            + " (ID INTEGER NOT NULL PRIMARY KEY, VAL INTEGER, ZIP INTEGER) "
+                            + tableDDLOptions;
+            conn.createStatement().execute(stmString1);
+            String upsertStmt = "UPSERT INTO " + dataTableFullName + " VALUES(?,?,?)";
+            PreparedStatement stmt = conn.prepareStatement(upsertStmt);
+            setWithNull(NROWS, 3, stmt);
+            conn.commit();
+            setWithNull(NROWS, 5, stmt);
```
gokceni: Can we add a comment here what are we trying to do? Insert 3 rows and upsert 2 more new ones?

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/UngroupedAggregateRegionObserver.java
```diff
@@ -1111,7 +1125,7 @@ private RegionScanner rebuildIndices(final RegionScanner innerScanner, final Reg
             synchronized (innerScanner) {
                 do {
                     List<Cell> results = new ArrayList<Cell>();
-                    hasMore = innerScanner.nextRaw(results);
+                    hasMore =  newScanner.nextRaw(results);
```
gjacoby126: is anything else later in the pipeline assuming that innerScanner will get read fully here?

phoenix-core/src/it/java/org/apache/phoenix/end2end/IndexToolIT.java
kadirozde: Done

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/UngroupedAggregateRegionObserver.java
kadirozde: Yes, the above changes will be no-op for partial rebuilds by MetaDataRegionObserver and GlobalIndexChecker. The changes are only for rebuilds by IndexTool and index create (both using ServerIndexBuildComplier).

phoenix-core/src/it/java/org/apache/phoenix/end2end/IndexToolIT.java
kadirozde: I added major compaction.

PR #602, opened 2019-10-22
PHOENIX-5535 Index rebuilds via UngroupedAggregateRegionObserver shou…

kadirozde:
…ld replay delete markers

@gjacoby126, I had to come up with a new PR for PHOENIX-5535 due to some permission IT test failures of the previous PR. The new PR does not change the server side (UngroupedAggregateRegionServer) as it uses essentially the same compiler partial rebuilder uses. I needed to restructure PostDDLCompiler to separate query plan from mutation plan. So there is no change there. This PR passes all tests. I appreciate if you can review it. Thank you!

### DIFF
--- a/phoenix-core/src/it/java/org/apache/phoenix/end2end/IndexToolIT.java
+++ b/phoenix-core/src/it/java/org/apache/phoenix/end2end/IndexToolIT.java
@@ -29,6 +29,7 @@ import java.sql.DriverManager;
 import java.sql.PreparedStatement;
 import java.sql.ResultSet;
 import java.sql.SQLException;
+import java.sql.Types;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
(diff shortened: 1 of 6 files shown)


gjacoby126:
@kadirozde - So, I see that you're now having ServerBuildIndexCompiler using PostDDLCompiler rather than PhoenixStatement to generate its query plan -- and given that that's what you're trying to do, the changes make sense. (The extract method on PostDDLCompiler is a good refactoring change even by itself!) 

What's less clear to me is what the purpose of using PostDDLCompiler's AggregatePlan, since the compilation logic is complex and I'm not sure what aspect of it you're relying on. Could you please explain what's the difference, and how did it help fix a permission test issue?

kadirozde:
@gjacoby126, The rebuilds are done by scanning tables using UngroupedAggregateRegionObserver which implements the server side of AggregatePlan.  A "SELECT COUNT(*)" statement is compiled to an aggregate plan. Before this JIRA, I was generating an aggregate plan using this select statement. This plan uses regular scan. Turning this scan into raw scan did not work as the filter set on the scan for this aggregate plan was not compatible with raw scans. So, I hacked it by using ""SELECT *" and setting scan.setAttribute(BaseScannerRegionObserver.UNGROUPED_AGG, TRUE_BYTES).
This worked with raw scan but caused permission IT failures. I tried to find out the reason for this failure but it was not obvious. So then I followed exactly what MetaDataRegionObserver for partial rebuilds which also uses PostDDLCompiler and raw scan. In a way, we have better unification between partial rebuilds and full rebuilds now. Actually, this was my very first patch that I posted. However, it was more involved than what I wanted (or thought so) and I removed it and came up with the first one you reviewed. So, I made a circle :-)

PR #604, opened 2019-10-23
PHOENIX-5541 - Redundant Global Mutable Index Writes on Legacy Indexer

gjacoby126:


priyankporwal:
We cannot remove old codepath since old indexer is still present and will be in use until indexes are upgraded. Re-baselining does make sense though.

PR #605, opened 2019-10-23
PHOENIX-5540 Full row index write at the last write phase for immutab…

gokceni:
…le global indexes

### DIFF
--- a/phoenix-core/src/it/java/org/apache/phoenix/end2end/index/ImmutableIndexIT.java
+++ b/phoenix-core/src/it/java/org/apache/phoenix/end2end/index/ImmutableIndexIT.java
@@ -47,6 +47,7 @@ import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicInteger;
 
 import org.apache.hadoop.hbase.Cell;
+import org.apache.hadoop.hbase.CellScanner;
 import org.apache.hadoop.hbase.DoNotRetryIOException;
 import org.apache.hadoop.hbase.HConstants;
 import org.apache.hadoop.hbase.TableName;
@@ -58,6 +59,7 @@ import org.apache.hadoop.hbase.client.Put;
 import org.apache.hadoop.hbase.client.Result;
 import org.apache.hadoop.hbase.client.ResultScanner;
 import org.apache.hadoop.hbase.client.Scan;
+import org.apache.hadoop.hbase.client.Table;
 import org.apache.hadoop.hbase.coprocessor.ObserverContext;
 import org.apache.hadoop.hbase.coprocessor.RegionCoprocessorEnvironment;
 import org.apache.hadoop.hbase.coprocessor.SimpleRegionObserver;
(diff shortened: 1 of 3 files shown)


phoenix-core/src/it/java/org/apache/phoenix/end2end/index/ImmutableIndexIT.java
```diff
@@ -417,6 +422,58 @@ public void testGlobalImmutableIndexDelete() throws Exception {
         }
     }
 
+    @Test
+    public void testGlobalImmutableIndexFullWriteOnLast() throws Exception {
```
priyankporwal: Nit: This test only validates phase1 that it does not write the full row and only unverified. The test name should probably be something like "testGlobalImmutableIndexUnverifiedOnlyInPhase1"

BinShi-SecularBird:
Just curious, the main code change is in MutationState which seems to be generally for both mutable and Immutable global index, so how do we ensure this change is specifically for immutable global index?

phoenix-core/src/it/java/org/apache/phoenix/end2end/index/ImmutableIndexIT.java
BinShi-SecularBird: Does the existing test case cover "In the presence of concurrent writes, we skip the last write phase. These writes leave the index writes in unverified status. Similarly, if the first or second phase write fails, we do not proceed with the third phase"?

gokceni: Other tests cover those cases

phoenix-core/src/main/java/org/apache/phoenix/execute/MutationState.java
```diff
@@ -1239,15 +1239,17 @@ private void filterIndexCheckerMutations(Map<TableInfo, List<Mutation>> mutation
...
+                        // Just send empty with Unverified
+                        Put unverifiedPut = new Put(m.getRow());
+                        unverifiedPut.addColumn(emptyCF, emptyCQ, timestamp, IndexRegionObserver.UNVERIFIED_BYTES);
+                        addToMap(unverifiedIndexMutations, tableInfo, unverifiedPut);
 
                         // Phase 3 mutations are verified
-                        Put verifiedPut = new Put(m.getRow());
-                        verifiedPut.addColumn(emptyCF, emptyCQ, timestamp,
+                        // Send entire mutation with verified
+                        ((Put) m).addColumn(emptyCF, emptyCQ, timestamp,
```
kadirozde: The empty column is already part of the mutation at this moment with value "x". Can we remove it first and add the new one as the patch for PHOENIX-5539 does.

phoenix-core/src/main/java/org/apache/phoenix/hbase/index/IndexRegionObserver.java
```diff
@@ -506,7 +506,7 @@ private void populatePendingRows(BatchMutateContext context, long now) {
       return mutations;
   }
 
-  public void removeEmptyColumn(Mutation m, byte[] emptyCF, byte[] emptyCQ) {
+  public static void removeEmptyColumn(Mutation m, byte[] emptyCF, byte[] emptyCQ) {
```
BinShi-SecularBird: @gokceni, @kadirozde, IMPO, we should move the following list from IndexRegionObserver to an MutationUtil class in org.apache.phoenix.hbase.index.util or org.apache.phoenix.util:

IndexRegionObserver.UNVERIFIED_BYTES
IndexRegionObserver.VERIFIED_BYTES
IndexRegionObserver.getMaxTimestamp(...)
IndexRegionObserver.removeEmptyColumn(...)

gjacoby126:
JIRA already committed, closing

PR #607, opened 2019-10-25
PHOENIX-5103: Can't create/drop table using 4.14 client against 4.15 server (Addendum)

ChinmaySKulkarni:


ChinmaySKulkarni:
@vincentpoon @twdsilva @kadirozde @lhofhansl please review.

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/MetaDataEndpointImpl.java
```diff
@@ -2355,9 +2362,16 @@ private MetaDataMutationResult doDropTable(byte[] key, byte[] tenantId, byte[] s
...
-                                    Bytes.toString(schemaName), Bytes.toString(tableName), this.accessCheckEnabled);
+                            if (clientVersion >= MIN_SPLITTABLE_SYSTEM_CATALOG) {
+                                PhoenixConnection conn =
+                                        QueryUtil.getConnectionOnServer(env.getConfiguration())
+                                                .unwrap(PhoenixConnection.class);
+                                Task.addTask(conn, PTable.TaskType.DROP_CHILD_VIEWS,
+                                        Bytes.toString(tenantId), Bytes.toString(schemaName),
+                                        Bytes.toString(tableName), this.accessCheckEnabled);
+                            }
+                            // else: the client version is old, so we cannot add a task to cleanup
```
twdsilva: If the client is older than 4.15 we should just throw an exception if they try to drop a base table that has child views.

ChinmaySKulkarni: @twdsilva makes sense. I will handle that as part of PHOENIX-5544. Do the other changes to support creating/dropping a table look good?

twdsilva: SGTM.

ChinmaySKulkarni:
Thanks for the review @twdsilva

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/MetaDataEndpointImpl.java
```diff
@@ -1714,9 +1716,14 @@ public void createTable(RpcController controller, CreateTableRequest request,
             }
 
             // check if the table was dropped, but had child views that were have not yet been cleaned up
-            if (!Bytes.toString(schemaName).equals(QueryConstants.SYSTEM_SCHEMA_NAME) &&
-                    clientVersion >= MIN_SPLITTABLE_SYSTEM_CATALOG) {
-                ViewUtil.dropChildViews(env, tenantIdBytes, schemaName, tableName);
+            if (!Bytes.toString(schemaName).equals(QueryConstants.SYSTEM_SCHEMA_NAME)) {
+                byte[] sysCatOrSysChildLink = SchemaUtil.getPhysicalTableName(
+                        clientVersion >= MIN_SPLITTABLE_SYSTEM_CATALOG ?
```
gjacoby126: nit: this would be much easier to read if you extracted the clientVersion -> table byte logic into its own method

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/MetaDataEndpointImpl.java
```diff
@@ -2341,10 +2348,10 @@ private MetaDataMutationResult doDropTable(byte[] key, byte[] tenantId, byte[] s
 
             if (tableType == PTableType.TABLE || tableType == PTableType.VIEW || tableType == PTableType.SYSTEM) {
                 // check to see if the table has any child views
-                try (Table hTable =
-                             env.getTable(SchemaUtil.getPhysicalTableName(
-                                     PhoenixDatabaseMetaData.SYSTEM_CHILD_LINK_NAME_BYTES,
-                                     env.getConfiguration()))) {
+                try (Table hTable = env.getTable(SchemaUtil.getPhysicalTableName(
+                        clientVersion >= MIN_SPLITTABLE_SYSTEM_CATALOG ?
```
gjacoby126: ditto extract method

phoenix-core/src/main/java/org/apache/phoenix/query/DelegateConnectionQueryServices.java
```diff
@@ -361,6 +361,11 @@ public boolean isUpgradeRequired() {
         return getDelegate().isUpgradeRequired();
     }
 
+    @Override
+    public void clearUpgradeRequired() {
+        getDelegate().isUpgradeRequired();
```
gjacoby126: Shouldn't this be getDelegate.clearUpgradeRequired(); ?

ChinmaySKulkarni: Good catch! Copy-paste error. Luckily this method is never really used in our upgrade code path since we use the `CQSI.clearUpgradeRequired`. I will add this to my changes for PHOENIX-5544 since it works with relatively the same path and 5103 is already committed.

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/MetaDataEndpointImpl.java
ChinmaySKulkarni: I did this as part of PHOENIX-5544 (https://github.com/apache/phoenix/pull/620)

PR #608, opened 2019-10-25
PHOENIX-5539 Full row index write at the last write phase for mutable…

kadirozde:
… global indexes

### DIFF
--- a/phoenix-core/src/it/java/org/apache/phoenix/end2end/index/GlobalIndexCheckerIT.java
+++ b/phoenix-core/src/it/java/org/apache/phoenix/end2end/index/GlobalIndexCheckerIT.java
@@ -17,6 +17,7 @@
  */
 package org.apache.phoenix.end2end.index;
 
+import static org.apache.phoenix.end2end.index.ImmutableIndexIT.verifyRowsForEmptyColValue;
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertFalse;
 import static org.junit.Assert.assertTrue;
@@ -33,11 +34,19 @@ import java.util.Map;
 import com.google.common.collect.Maps;
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
+import org.apache.hadoop.hbase.CellScanner;
+import org.apache.hadoop.hbase.client.Result;
+import org.apache.hadoop.hbase.client.ResultScanner;
+import org.apache.hadoop.hbase.client.Scan;
+import org.apache.hadoop.hbase.client.Table;
+import org.apache.hadoop.hbase.util.Bytes;
 import org.apache.phoenix.end2end.BaseUniqueNamesOwnClusterIT;
 import org.apache.phoenix.end2end.IndexToolIT;
 import org.apache.phoenix.hbase.index.IndexRegionObserver;
+import org.apache.phoenix.jdbc.PhoenixConnection;
 import org.apache.phoenix.util.QueryUtil;
 import org.apache.phoenix.util.ReadOnlyProps;
+import org.apache.phoenix.util.TestUtil;
 import org.junit.After;
 import org.junit.BeforeClass;
 import org.junit.Test;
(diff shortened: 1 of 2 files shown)


phoenix-core/src/it/java/org/apache/phoenix/end2end/index/GlobalIndexCheckerIT.java
```diff
@@ -257,6 +266,44 @@ public void testFailPostIndexRowUpdate() throws Exception {
...
+            conn.commit();
+            IndexRegionObserver.setFailPostIndexUpdatesForTesting(false);
+            Table hIndexTable = conn.unwrap(PhoenixConnection.class).getQueryServices()
+                    .getTable(Bytes.toBytes(indexTableName));
+            long indexCnt = TestUtil.getRowCount(hIndexTable, false);
+            assertEquals(1, indexCnt);
+            assertEquals(true, verifyRowsForEmptyColValue(conn, indexTableName, IndexRegionObserver.UNVERIFIED_BYTES));
+            // check that in the first phase we don't send the full row.
+            // We count the num of cells for this
+            Scan s = new Scan();
```
gokceni: nit: nice to have is to have this as a separate function to call. But don't worry about it

phoenix-core/src/main/java/org/apache/phoenix/hbase/index/IndexRegionObserver.java
```diff
@@ -555,34 +570,33 @@ private void prepareIndexMutations(ObserverContext<RegionCoprocessorEnvironment>
...
+                      // going through concurrent updates
+                      ImmutableBytesPtr rowKey = new ImmutableBytesPtr(next.getSecond());
+                      if (!context.pendingRows.contains(rowKey)) {
+                          if (m instanceof Put) {
+                              // Remove the empty column prepared by Index codec as we need to change its value
+                              removeEmptyColumn(m, emptyCF, emptyCQ);
+                              ((Put) m).addColumn(emptyCF, emptyCQ, now, VERIFIED_BYTES);
                               context.intermediatePostIndexUpdates.add(next);
+                          } else {
+                              if (!context.pendingRows.contains(rowKey)) {
```
gokceni: Do we need this if? There is if in 590 so it should be ok to remove this

kadirozde: Will do that.

phoenix-core/src/it/java/org/apache/phoenix/end2end/index/GlobalIndexCheckerIT.java
kadirozde: It is a good idea. I will do that.

phoenix-core/src/main/java/org/apache/phoenix/hbase/index/IndexRegionObserver.java
```diff
@@ -555,33 +570,30 @@ private void prepareIndexMutations(ObserverContext<RegionCoprocessorEnvironment>
                   // add the VERIFIED cell, which is the empty cell
                   Mutation m = next.getFirst().getFirst();
                   boolean rebuild = PhoenixIndexMetaData.isIndexRebuild(m.getAttributesMap());
-                  long ts = getMaxTimestamp(m);
                   if (rebuild) {
                       if (m instanceof Put) {
+                          long ts = getMaxTimestamp(m);
+                          // Remove the empty column prepared by Index codec as we need to change its value
+                          removeEmptyColumn(m, emptyCF, emptyCQ);
```
BinShi-SecularBird: It seems that we first remove the existing "empty" column from an array -- the cell list of the column family then add a new "empty" column. Is it necessary have an optimization here - replace the old cell with the new cell in place in the array?

phoenix-core/src/main/java/org/apache/phoenix/hbase/index/IndexRegionObserver.java
```diff
@@ -555,33 +570,30 @@ private void prepareIndexMutations(ObserverContext<RegionCoprocessorEnvironment>
...
+                      context.preIndexUpdates.add(new Pair <>(unverifiedPut, next.getFirst().getSecond()));
+                      // Ignore post index updates (i.e., the third write phase updates) for this row if it is
+                      // going through concurrent updates
+                      ImmutableBytesPtr rowKey = new ImmutableBytesPtr(next.getSecond());
+                      if (!context.pendingRows.contains(rowKey)) {
+                          if (m instanceof Put) {
+                              // Remove the empty column prepared by Index codec as we need to change its value
+                              removeEmptyColumn(m, emptyCF, emptyCQ);
+                              ((Put) m).addColumn(emptyCF, emptyCQ, now, VERIFIED_BYTES);
+                              context.intermediatePostIndexUpdates.add(next);
```
BinShi-SecularBird: Move "context.intermediatePostIndexUpdates.add(next);" outside of if ... else?

kadirozde: ok. I will do that

phoenix-core/src/main/java/org/apache/phoenix/hbase/index/IndexRegionObserver.java
kadirozde: No need to transfer an unnecessary cell mutation over the network. Also, giving the same cell with two different values is confusing, although HBase takes the latest one.

BinShi-SecularBird: That isn't what I meant. I meant overriding the old cell with the new cell in the cell list in place and avoiding removing an element in an array.

kadirozde: There is no HBase API to do that as far as I know, that is why I have to remove it.

gjacoby126:

PR #611, opened 2019-10-28
PHOENIX-5535 Replay delete markers during server side global index re…

kadirozde:
…build

### DIFF
--- a/phoenix-core/src/it/java/org/apache/phoenix/end2end/IndexToolIT.java
+++ b/phoenix-core/src/it/java/org/apache/phoenix/end2end/IndexToolIT.java
@@ -29,6 +29,7 @@ import java.sql.DriverManager;
 import java.sql.PreparedStatement;
 import java.sql.ResultSet;
 import java.sql.SQLException;
+import java.sql.Types;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
@@ -37,8 +38,11 @@ import java.util.Map;
 import java.util.Properties;
 import java.util.UUID;
 
+import org.apache.commons.logging.Log;
+import org.apache.commons.logging.LogFactory;
 import org.apache.hadoop.conf.Configuration;
 import org.apache.hadoop.hbase.Cell;
+import org.apache.hadoop.hbase.CellUtil;
 import org.apache.hadoop.hbase.HTableDescriptor;
 import org.apache.hadoop.hbase.TableName;
 import org.apache.hadoop.hbase.client.*;
@@ -46,6 +50,7 @@ import org.apache.hadoop.hbase.regionserver.HRegion;
 import org.apache.hadoop.hbase.regionserver.RegionScanner;
 import org.apache.hadoop.hbase.util.Bytes;
 import org.apache.hadoop.mapreduce.Job;
+import org.apache.phoenix.end2end.index.GlobalIndexCheckerIT;
 import org.apache.phoenix.jdbc.PhoenixConnection;
 import org.apache.phoenix.mapreduce.index.IndexTool;
 import org.apache.phoenix.query.ConnectionQueryServices;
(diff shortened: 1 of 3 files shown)


phoenix-core/src/main/java/org/apache/phoenix/coprocessor/UngroupedAggregateRegionObserver.java
```diff
@@ -1146,6 +1154,45 @@ private void setMutationAttributes(Mutation m, byte[] uuidValue) {
             m.setDurability(Durability.SKIP_WAL);
         }
 
+        private Delete generateDeleteMarkers(List<Cell> row) {
+            Set<ColumnReference> allColumns = indexMaintainer.getAllColumns();
+            if (row.size() == allColumns.size() + 1) {
```
gokceni: Can we have a comment here why we are returning null when we have 1 more row cell then allColumns?

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/UngroupedAggregateRegionObserver.java
```diff
@@ -1146,6 +1154,45 @@ private void setMutationAttributes(Mutation m, byte[] uuidValue) {
...
+                return null;
+            }
+            Set<ColumnReference> includedColumns = Sets.newLinkedHashSetWithExpectedSize(row.size());
+            long ts = 0;
+            for (Cell cell : row) {
+                includedColumns.add(new ColumnReference(CellUtil.cloneFamily(cell), CellUtil.cloneQualifier(cell)));
+                if (ts < cell.getTimestamp()) {
+                    ts = cell.getTimestamp();
+                }
+            }
```
gokceni: Lines from 1163 to 1169 seem to be a good candidate to have in a separate function

phoenix-core/src/it/java/org/apache/phoenix/end2end/IndexToolIT.java
```diff
@@ -157,6 +162,70 @@ public static void setup() throws Exception {
...
+            conn.commit();
+            actualRowCount = IndexScrutiny.scrutinizeIndex(conn, dataTableFullName, indexTableFullName);
+            assertEquals(NROWS, actualRowCount);
+            setEveryNthRowWithNull(NROWS, 7, stmt);
+            conn.commit();
+            actualRowCount = IndexScrutiny.scrutinizeIndex(conn, dataTableFullName, indexTableFullName);
+            assertEquals(NROWS, actualRowCount);
+            TestUtil.doMajorCompaction(conn, dataTableFullName);
+            actualRowCount = IndexScrutiny.scrutinizeIndex(conn, dataTableFullName, indexTableFullName);
+            assertEquals(NROWS, actualRowCount);
```
gokceni: do you want to add a test to do check for cells with different timestamps?

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/UngroupedAggregateRegionObserver.java
kadirozde: They are already in a function, which is generateDeleteMarkers.

phoenix-core/src/it/java/org/apache/phoenix/end2end/IndexToolIT.java
kadirozde: Not sure I understand the test you are suggesting.

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/UngroupedAggregateRegionObserver.java
kadirozde: I will add a comment line. The reason we return is that we have all the columns for the index table plus the empty column.

phoenix-core/src/it/java/org/apache/phoenix/end2end/IndexToolIT.java
gokceni: In generateDeleteMarkers, you have a loop to find a timestamp. I was suggesting some test related to that code. It is ok if you think it is not necessary

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/UngroupedAggregateRegionObserver.java
```diff
@@ -1121,10 +1124,15 @@ private static PTable deserializeTable(byte[] b) {
                 useProto = false;
                 indexMetaData = scan.getAttribute(PhoenixIndexCodec.INDEX_MD);
             }
+            if (!scan.isRaw()) {
+                List<IndexMaintainer> maintainers = IndexMaintainer.deserialize(indexMetaData, true);
+                indexMaintainer = maintainers.get(0);
```
gokceni: Is there a possibility that  maintainers is an empty list?

phoenix-core/src/it/java/org/apache/phoenix/end2end/IndexToolIT.java
kadirozde: Got it. The loop for getting the maximum timestamp for the mutation of a  given row and using the max timestamp for the delete markers for the missing columns. Since we do not read delete markers, we do not know exactly when the delete markers are added or if the row is created with missing columns without delete markers in the first place. Even if we attempt to get the delete markers, they were there in the first place or removed by compaction.  So reading delete markers does not help either. Now, why we do insert these delete markers here, you may ask. It is to make sure that index rebuild code on IndexRegionObserver passes these delete markers to the index table. If these markers are not there in the index table for missing columns and there is a wrong/stale row in the index table before rebuild, the missing columns will inherit these stale/wrong values. 
Please note that the test for delete markers is pretty involved. It creates 210 rows and overwrites these rows with different null patterns three times and checks if the index table is correct with and without compaction.

PR #612, opened 2019-10-29
PHOENIX-5527 Unverified index rows should not be deleted due to repli…

kadirozde:
…cation lag

### DIFF
--- a/phoenix-core/src/main/java/org/apache/phoenix/hbase/index/IndexRegionObserver.java
+++ b/phoenix-core/src/main/java/org/apache/phoenix/hbase/index/IndexRegionObserver.java
@@ -19,6 +19,7 @@ package org.apache.phoenix.hbase.index;
 
 import static org.apache.phoenix.hbase.index.util.IndexManagementUtil.rethrowIndexingException;
 import static org.apache.phoenix.index.IndexMaintainer.getIndexMaintainer;
+import static org.apache.phoenix.util.ServerUtil.wrapInDoNotRetryIOException;
 
 import java.io.IOException;
 import java.util.ArrayList;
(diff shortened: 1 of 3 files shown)


phoenix-core/src/main/java/org/apache/phoenix/hbase/index/IndexRegionObserver.java
```diff
@@ -582,7 +584,7 @@ private void prepareIndexMutations(ObserverContext<RegionCoprocessorEnvironment>
                       // For this mutation whether it is put or delete, set the status of the index row "unverified"
                       // This will be done before the data table row is updated (i.e., in the first write phase)
                       Put unverifiedPut = new Put(m.getRow());
-                      unverifiedPut.addColumn(emptyCF, emptyCQ, now, UNVERIFIED_BYTES);
+                      unverifiedPut.addColumn(emptyCF, emptyCQ, now - 1, UNVERIFIED_BYTES);
```
gokceni: How about immutable indexes?

BinShi-SecularBird: It's better to add comment to explain the magic "now - 1". Something like "The timestamp of the unverified row is the timestamp of the index full row (and also the data table row) minus 1. This will make sure that if the unverified row is deleted by read repair, it will not mask the verified row."(PHOENIX-5527 https://issues.apache.org/jira/browse/PHOENIX-5527)

phoenix-core/src/main/java/org/apache/phoenix/hbase/index/IndexRegionObserver.java
```diff
@@ -719,6 +722,23 @@ public void postBatchMutateIndispensably(ObserverContext<RegionCoprocessorEnviro
           for (RowLock rowLock : context.rowLocks) {
               rowLock.release();
           }
+          // Sleep for one millisecond if we have done the data table updates in less than 1 ms. The sleep is necessary
+          // not to allow another data table write on the same row within the same ms. The sleep is very rare to happen.
+          // Assume that a data table write completes at timestamp t.  Now assume another write happens on the same data
+          // row at timestamp t+1. The mutation for unverifying the index table row(s) for the previous write will have
```
priyankporwal: Nit: Repharse, this as " On second update to the row, 
the mutation for unverifying the index table row(s) for the previous write ....

priyankporwal:
+1 Thanks for fixing this @kadirozde!

PR #614, opened 2019-10-30
PHOENIX-5553: Transactional views should not get GlobalIndexChecker c…

gokceni:
…oproc

### DIFF
--- a/phoenix-core/src/main/java/org/apache/phoenix/query/ConnectionQueryServicesImpl.java
+++ b/phoenix-core/src/main/java/org/apache/phoenix/query/ConnectionQueryServicesImpl.java
@@ -53,6 +53,7 @@ import static org.apache.phoenix.jdbc.PhoenixDatabaseMetaData.TABLE_NAME;
 import static org.apache.phoenix.jdbc.PhoenixDatabaseMetaData.TABLE_SCHEM;
 import static org.apache.phoenix.jdbc.PhoenixDatabaseMetaData.TASK_TABLE_TTL;
 import static org.apache.phoenix.jdbc.PhoenixDatabaseMetaData.TENANT_ID;
+import static org.apache.phoenix.jdbc.PhoenixDatabaseMetaData.TRANSACTIONAL;
 import static org.apache.phoenix.jdbc.PhoenixDatabaseMetaData.VIEW_CONSTANT;
 import static org.apache.phoenix.monitoring.GlobalClientMetrics.GLOBAL_HCONNECTIONS_COUNTER;
 import static org.apache.phoenix.monitoring.GlobalClientMetrics.GLOBAL_PHOENIX_CONNECTIONS_THROTTLED_COUNTER;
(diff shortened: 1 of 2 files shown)


phoenix-core/src/main/java/org/apache/phoenix/query/ConnectionQueryServicesImpl.java
```diff
@@ -927,7 +928,16 @@ private void addCoprocessors(byte[] tableName, TableDescriptorBuilder builder, P
...
-            if ((tableType == PTableType.INDEX && !isTransactional) || isViewIndex) {
+            boolean isViewBaseTransactional = false;
+            if (!isTransactional && isViewIndex) {
+                if (tableProps.containsKey(TRANSACTIONAL) &&
+                        Boolean.TRUE.equals(tableProps.get(TRANSACTIONAL))) {
+                    isViewBaseTransactional = true;
+                }
+            }
+
+            if (!isTransactional && !isViewBaseTransactional
```
kadirozde: Do we need to introduce a separate variable, i.e., isViewBaseTransactional? Would not the following work?
If (!isTransactional && (PTableType.INDEX || isViewIndex) {

I think we have a test to check coprocessors on a table for the new index design. Can we update the test for view indexes too?

phoenix-core/src/main/java/org/apache/phoenix/query/ConnectionQueryServicesImpl.java
```diff
@@ -927,7 +928,16 @@ private void addCoprocessors(byte[] tableName, TableDescriptorBuilder builder, P
             boolean isViewIndex = TRUE_BYTES_AS_STRING
                     .equals(tableProps.get(MetaDataUtil.IS_VIEW_INDEX_TABLE_PROP_NAME));
 
-            if ((tableType == PTableType.INDEX && !isTransactional) || isViewIndex) {
+            boolean isViewBaseTransactional = false;
+            if (!isTransactional && isViewIndex) {
+                if (tableProps.containsKey(TRANSACTIONAL) &&
+                        Boolean.TRUE.equals(tableProps.get(TRANSACTIONAL))) {
```
BinShi-SecularBird: Is "Boolean.TRUE.equals(tableProps.getOrDefault(TRANSACTIONAL, FALSE))" more efficient?

phoenix-core/src/main/java/org/apache/phoenix/query/ConnectionQueryServicesImpl.java
```diff
@@ -927,7 +928,16 @@ private void addCoprocessors(byte[] tableName, TableDescriptorBuilder builder, P
...
+                    isViewBaseTransactional = true;
```
BinShi-SecularBird: The whole logic is a mess which should be cleaned. I don't understand why we need to check "Boolean.TRUE.equals(tableProps.getOrDefault(TRANSACTIONAL, FALSE))" particularly for viewIndex? Shouldn't "isTransactional = (provider != null) && Boolean.TRUE.equals(tableProps.getOrDefault(TRANSACTIONAL, FALSE))" be applied to both INDEX and viewIndex, and simplify "if (!isTransactional && !isViewBaseTransactional && (tableType == PTableType.INDEX || isViewIndex))" as (!isTransactional && (tableType == PTableType.INDEX || isViewIndex))?

gjacoby126:

PR #616, opened 2019-10-30
PHOENIX-5432: Refactor LiteralExpression to use the builder pattern

christinefeng:


christinefeng:
@ChinmaySKulkarni @yanxinyi @JamesRTaylor @jpisaac

gjacoby126:
@christinefeng - we try to avoid using star imports, and what looks like your IDE's automatic use of them makes this patch bigger than it otherwise would be.

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/MetaDataProtocol.java
```diff
@@ -380,7 +381,7 @@ public PDataType getViewIndexIdType() {
           return viewIndexIdType;
       }
 
-        public static MetaDataMutationResult constructFromProto(MetaDataResponse proto) {
+        public static MetaDataMutationResult constructFromProto(MetaDataResponse proto) throws SQLException {
```
yanxinyi: why we are adding extra `throws SQLException` here?

phoenix-core/src/main/java/org/apache/phoenix/execute/MutationState.java
```diff
@@ -726,7 +726,11 @@ public boolean hasNext() {
             public Pair<byte[], List<Mutation>> next() {
                 if (!innerIterator.hasNext()) {
                     current = iterator.next();
-                    innerIterator = init();
+                    try {
+                        innerIterator = init();
+                    } catch (SQLException e) {
+                        e.printStackTrace();
```
yanxinyi: maybe using `LOGGER.ERROR(e.printStackTrace());`

phoenix-core/src/main/java/org/apache/phoenix/expression/DelegateExpression.java
```diff
@@ -20,6 +20,7 @@
 import java.io.DataInput;
 import java.io.DataOutput;
 import java.io.IOException;
+import java.sql.SQLException;
```
yanxinyi: nit: please remove

phoenix-core/src/main/java/org/apache/phoenix/expression/Expression.java
```diff
@@ -17,6 +17,7 @@
  */
 package org.apache.phoenix.expression;
 
+import java.sql.SQLException;
```
yanxinyi: nit: please remove

gjacoby126:
@christinefeng - looks like some extra stuff got pushed up with your PR.

phoenix-core/src/main/java/org/apache/phoenix/compile/ListJarsQueryPlan.java
```diff
@@ -158,9 +158,9 @@ public Tuple next() throws SQLException {
                     ParseNodeFactory factory = new ParseNodeFactory();
                     LiteralParseNode literal =
                             factory.literal(listFiles.next().getPath().toString());
-                    LiteralExpression expression =
-                            LiteralExpression.newConstant(literal.getValue(), PVarchar.INSTANCE,
-                                Determinism.ALWAYS);
+//                    LiteralExpression expression =
```
yanxinyi: nit: remove

phoenix-core/src/main/java/org/apache/phoenix/compile/ProjectionCompiler.java
```diff
@@ -103,7 +103,8 @@
  * @since 0.1
  */
 public class ProjectionCompiler {
-    private static final Expression NULL_EXPRESSION = LiteralExpression.newConstant(null);
+    private static Expression NULL_EXPRESSION = new LiteralExpression.Builder().buildValueAndDeterminism(false);
```
yanxinyi: why not `final` ?

phoenix-core/src/main/java/org/apache/phoenix/expression/LiteralExpression.java
```diff
@@ -22,6 +22,7 @@
 import java.io.IOException;
 import java.sql.SQLException;
 
+import com.sun.org.apache.xpath.internal.operations.Bool;
```
yanxinyi: what does this do?

phoenix-core/src/main/java/org/apache/phoenix/expression/LiteralExpression.java
```diff
@@ -34,13 +35,12 @@
...
 import org.apache.phoenix.schema.types.PChar;
 import org.apache.phoenix.schema.types.PDataType;
-import org.apache.phoenix.schema.types.PDate;
-import org.apache.phoenix.schema.types.PTime;
 import org.apache.phoenix.schema.types.PTimestamp;
 import org.apache.phoenix.schema.types.PVarchar;
 import org.apache.phoenix.schema.types.PhoenixArray;
-import org.apache.phoenix.util.ByteUtil;
 import org.apache.phoenix.util.StringUtil;
+import org.apache.phoenix.util.ByteUtil;
```
yanxinyi: nit: removed and added the same import statement?

phoenix-core/src/main/java/org/apache/phoenix/expression/LiteralExpression.java
```diff
@@ -54,184 +54,218 @@
...
-        if (maxLength == null) {
-            maxLength = type.isFixedWidth() ? type.getMaxLength(value) : null;
+
+        public LiteralExpression build() throws SQLException{
+            setDefaults();
+            if (this.value == null) {
+                return  (this.type == null) ?  LiteralExpression.getNullLiteralExpression(this.determinism) :
+                        LiteralExpression.getTypedNullLiteralExpression(this.type, this.determinism);
+            }
+            else if (this.value instanceof Boolean){
```
yanxinyi: nit: inconsistent coding style, most of the time you keep `else if` with the same line after `}`, maybe follow the same coding style here?

phoenix-core/src/main/java/org/apache/phoenix/expression/LiteralExpression.java
```diff
@@ -54,184 +54,218 @@
...
+            }
+            return new LiteralExpression(this.value, this.type, this.byteValue, this.maxLength, this.scale, this.sortOrder, this.determinism);
         }
-        return new LiteralExpression(value, type, b, maxLength, scale, sortOrder, determinism);
+
     }
 
-    public LiteralExpression() {
+    public static boolean isFalse(Expression child) {
+        if (child!=null) {
```
yanxinyi: nit: space before and after `!=` ?

phoenix-core/src/main/java/org/apache/phoenix/index/IndexMaintainer.java
```diff
@@ -236,7 +236,7 @@ public static void serialize(PTable dataTable, ImmutableBytesWritable ptr,
     
     /**
      * For client-side to append serialized IndexMaintainers of keyValueIndexes
-     * @param dataTable data table
+     * @param table data table
```
yanxinyi: any reason that u want to change here?

phoenix-core/src/main/java/org/apache/phoenix/util/ExpressionUtil.java
```diff
@@ -26,14 +26,17 @@
...
 import org.apache.phoenix.expression.ColumnExpression;
 import org.apache.phoenix.expression.ComparisonExpression;
 import org.apache.phoenix.expression.Determinism;
 import org.apache.phoenix.expression.Expression;
+import org.apache.phoenix.expression.InListExpression;
 import org.apache.phoenix.expression.IsNullExpression;
 import org.apache.phoenix.expression.LiteralExpression;
 import org.apache.phoenix.expression.OrderByExpression;
 import org.apache.phoenix.expression.RowKeyColumnExpression;
+import org.apache.phoenix.expression.RowValueConstructorExpression;
```
yanxinyi: 3 new import statements without new usage?

christinefeng: For some reason, my IDE isn't marking these as unused imports so I must have missed them. Removing now

phoenix-core/src/main/java/org/apache/phoenix/expression/LiteralExpression.java
```diff
@@ -54,184 +53,216 @@
...
+
 	private static final LiteralExpression[] NULL_EXPRESSIONS = new LiteralExpression[Determinism.values().length];
     private static final LiteralExpression[] TYPED_NULL_EXPRESSIONS = new LiteralExpression[PDataType.values().length * Determinism.values().length];
     private static final LiteralExpression[] BOOLEAN_EXPRESSIONS = new LiteralExpression[2 * Determinism.values().length];
-    
+
     static {
     	for (Determinism determinism : Determinism.values()) {
-    		NULL_EXPRESSIONS[determinism.ordinal()] = new LiteralExpression(null, determinism);
+                NULL_EXPRESSIONS[determinism.ordinal()] = new LiteralExpression(null, null, ByteUtil.EMPTY_BYTE_ARRAY,
```
yanxinyi: nit: weird indentation

phoenix-core/src/main/java/org/apache/phoenix/expression/LiteralExpression.java
```diff
@@ -54,184 +53,216 @@
...
+                        null, null, SortOrder.getDefault(), determinism);
 	        for (int i = 0; i < PDataType.values().length; i++) {
-	            TYPED_NULL_EXPRESSIONS[i+PDataType.values().length*determinism.ordinal()] = new LiteralExpression(PDataType.values()[i], determinism);
-	        }        
-	        BOOLEAN_EXPRESSIONS[determinism.ordinal()] = new LiteralExpression(Boolean.FALSE,
-              PBoolean.INSTANCE, PBoolean.INSTANCE.toBytes(Boolean.FALSE), determinism);
-	        BOOLEAN_EXPRESSIONS[Determinism.values().length+determinism.ordinal()] = new LiteralExpression(Boolean.TRUE, PBoolean.INSTANCE, PBoolean.INSTANCE.toBytes(Boolean.TRUE), determinism);
+	            PDataType type = PDataType.values()[i];
```
yanxinyi: nit: weird indentation

phoenix-core/src/main/java/org/apache/phoenix/expression/LiteralExpression.java
```diff
@@ -54,184 +53,216 @@
...
+                TYPED_NULL_EXPRESSIONS[i+PDataType.values().length*determinism.ordinal()] = new LiteralExpression(null,
+                        type, ByteUtil.EMPTY_BYTE_ARRAY, !type.isFixedWidth() ? null : type.getMaxLength(null), null,
+                        SortOrder.getDefault(), determinism);
+	        }
+                BOOLEAN_EXPRESSIONS[determinism.ordinal()] = new LiteralExpression(Boolean.FALSE, PBoolean.INSTANCE,
```
yanxinyi: nit: weird indentation

phoenix-core/src/main/java/org/apache/phoenix/expression/LiteralExpression.java
```diff
@@ -54,184 +53,216 @@
...
-            if (type == PChar.INSTANCE && maxLength != null  && b.length < maxLength) {
-                if (rowKeyOrderOptimizable) {
-                    b = type.pad(b, maxLength, sortOrder);
-                } else {
-                    b = StringUtil.padChar(b, maxLength);
+
+        public LiteralExpression buildValueAndDeterminism(boolean byteValueOnly) {
+            /***
+             * @param byteValueOnly: true if building expression that only has byteValue, false otherwise
+             */
```
yanxinyi: method documentation should be above of the method.

phoenix-core/src/main/java/org/apache/phoenix/expression/LiteralExpression.java
```diff
@@ -57,181 +56,236 @@
...
+            if (type.isNull(this.byteValue)) {
+                return LiteralExpression.getTypedNullLiteralExpression(type, this.determinism);
+            }
+            if (type == PVarchar.INSTANCE) {
+                String s = (String) this.value;
+                if (s.length() == this.byteValue.length) { // single byte characters only
+                    type = PChar.INSTANCE;
+                }
+            }
+            return new LiteralExpression(this.value, type, this.byteValue, type == null
```
ChinmaySKulkarni: Same here. Why do we need to call a constructor of this any longer? We ideally shouldn't have any more such constructors. We should just have 1 constructor that has a `Builder` object as a parameter (and which is private and only called when `build()` is called). Take a look at how this is done in PTableImpl.

phoenix-core/src/main/java/org/apache/phoenix/expression/LiteralExpression.java
```diff
@@ -242,6 +296,9 @@ private LiteralExpression(Object value, PDataType type, byte[] byteValue,
         this.determinism = deterministic;
     }
 
+    public LiteralExpression () {
```
ChinmaySKulkarni: Where is this constructor used?

christinefeng: I believe it's used in some of the test cases that access LiteralExpression.<init>
At least, some of the unit tests were failing before I added the default constructor

phoenix-core/src/main/java/org/apache/phoenix/expression/LiteralExpression.java
```diff
@@ -57,181 +56,236 @@
 	private static final LiteralExpression[] NULL_EXPRESSIONS = new LiteralExpression[Determinism.values().length];
     private static final LiteralExpression[] TYPED_NULL_EXPRESSIONS = new LiteralExpression[PDataType.values().length * Determinism.values().length];
     private static final LiteralExpression[] BOOLEAN_EXPRESSIONS = new LiteralExpression[2 * Determinism.values().length];
-    
+
     static {
     	for (Determinism determinism : Determinism.values()) {
-    		NULL_EXPRESSIONS[determinism.ordinal()] = new LiteralExpression(null, determinism);
+            NULL_EXPRESSIONS[determinism.ordinal()] = new LiteralExpression(null, null,
```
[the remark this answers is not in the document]
christinefeng: For this whole static block, I decided to call the constructor directly since the original code is calling the constructor directly.
Additionally, using a builder here would necessitate creating a new build method (or adding another path to my existing build method) which I decided not to do since it would only be used twice throughout the code

phoenix-core/src/main/java/org/apache/phoenix/expression/LiteralExpression.java
```diff
@@ -57,181 +56,236 @@
...
-	        BOOLEAN_EXPRESSIONS[determinism.ordinal()] = new LiteralExpression(Boolean.FALSE,
-              PBoolean.INSTANCE, PBoolean.INSTANCE.toBytes(Boolean.FALSE), determinism);
-	        BOOLEAN_EXPRESSIONS[Determinism.values().length+determinism.ordinal()] = new LiteralExpression(Boolean.TRUE, PBoolean.INSTANCE, PBoolean.INSTANCE.toBytes(Boolean.TRUE), determinism);
+	            PDataType type = PDataType.values()[i];
+                TYPED_NULL_EXPRESSIONS[i+PDataType.values().length*determinism.ordinal()] =
+                        new LiteralExpression(null, type, ByteUtil.EMPTY_BYTE_ARRAY,
+                                !type.isFixedWidth() ? null : type.getMaxLength(null), null,
+                                SortOrder.getDefault(), determinism);
+	        }
+            BOOLEAN_EXPRESSIONS[determinism.ordinal()] = new LiteralExpression(Boolean.FALSE,
```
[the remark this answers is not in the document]
christinefeng: Using a Builder here has the potential of throwing an SQL Exception when the build method attempts to derive the `value` field, which isn't necessary since the value has already been provided.

phoenix-core/src/main/java/org/apache/phoenix/expression/LiteralExpression.java
```diff
@@ -57,181 +56,236 @@
...
-    }
+        /**
+         * Build method for Literal Expressions that have only fields value and determinism (can be null)
+         * or only field byteValue
+         * @param byteValueOnly true if LiteralExpression's only field is byteValue; false otherwise
+         * @return LiteralExpression object
+         */
+        public LiteralExpression buildValueAndDeterminism(boolean byteValueOnly) {
+            if (byteValueOnly) {
+                return new LiteralExpression(null, null, this.byteValue = this.byteValue != null
```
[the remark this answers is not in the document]
christinefeng: I see in PTableImpl, there's a public build method that calls the private initDerivedAttributes method based on whether a particular field is null.
I could change my code to do a similar thing, but in my case the public build method would have no function other than calling the method that initializes derived attributes, since it needs to be called in all cases, if that makes sense.
Either way, the actual LiteralExpression constructor I included is private, and the two build methods (build() and buildValueAndDeterminism(), which call the constructor) are public.
Let me know if you feel I should change it so there's a public build method calling the private initDerivedAttributes method.

phoenix-core/src/it/java/org/apache/phoenix/end2end/DescOrderKeysIT.java
```diff
@@ -0,0 +1,441 @@
+/*
```
ChinmaySKulkarni: I think you did not rebase your changes properly, which is why a lot of these extra files have showed up in your PR. Can you please rebase?

phoenix-core/src/main/java/org/apache/hadoop/hbase/regionserver/DataTableLocalIndexRegionScanner.java
```diff
@@ -91,15 +92,19 @@ public boolean next(List<Cell> outResult, ScannerContext scannerContext) throws
     public boolean next(List<Cell> results) throws IOException {
         List<Cell> dataTableResults = new ArrayList<Cell>();
         boolean next = super.next(dataTableResults);
-        addMutations(dataTableResults);
+        try {
+            addMutations(dataTableResults);
+        } catch (SQLException e) {
+            e.printStackTrace();
```
ChinmaySKulkarni: We shouldn't ever be swallowing an exception and printing to stdout/stderr. We should log it at worst. Any reason this exception shouldn't bubble up?

phoenix-core/src/main/java/org/apache/phoenix/expression/function/CountAggregateFunction.java
```diff
@@ -46,10 +47,12 @@
 @BuiltInFunction(name=CountAggregateFunction.NAME, args= {@Argument()} )
 public class CountAggregateFunction extends SingleAggregateFunction {
     public static final String NAME = "COUNT";
-    public static final List<Expression> STAR = Arrays.<Expression>asList(LiteralExpression.newConstant(1, Determinism.ALWAYS));
+//    STAR is never used so I commented it out
+//    public static final List<Expression> STAR = Arrays.<Expression>asList(new LiteralExpression.Builder().setValue(1).setDeterminism(Determinism.ALWAYS).build());
```
siddhimehta: unused code?

PR #617, opened 2019-10-31
PHOENIX-5140 TableNotFoundException occurs when we create local async…

clevertension:
originally, **table.getParentTableName()** will return the parent table as dot seperator such as **foo.table1**, but the physical table name was stored as colon seperator in hbase:meta such as **foo:table1**

so this fix will get parent physical table name with colon seperator


dbwong:
As @gjacoby126 mentioned in the JIRA looks okay but we would want a test.

PR #621, opened 2019-11-07
PHOENIX-5562 Simplify detection of concurrent updates on data tables with indexes

kadirozde:
In the current implementation, we have two pending-rows sets. The first one is at the table region level and the second one is at the batch level. The table region level set is a set of structures, one for each active row. This per row structure includes the row key, the timestamp of the last update on the row, and the number of concurrent updates on the row (i.e., the reference count). The second set is a simply set of row keys such that the corresponding rows for these row keys have at least one pending update at the time this batch is in the first phase. It turns out that we do not actually need the batch level set. All we need is that after acquiring the lock for the second time for a given update, we need to check if there are concurrent updates on the row using the region level set. if so, the last phase is skipped.

### DIFF
--- a/phoenix-core/src/main/java/org/apache/phoenix/hbase/index/IndexRegionObserver.java
+++ b/phoenix-core/src/main/java/org/apache/phoenix/hbase/index/IndexRegionObserver.java
@@ -108,19 +108,12 @@ public class IndexRegionObserver implements RegionObserver, RegionCoprocessor {
    * Class to represent pending data table rows
    */
   private static class PendingRow {
-      private long latestTimestamp;
-      private long count;
+      private boolean concurrent = false;
+      private long count = 1;
 
-      PendingRow(long latestTimestamp) {
-          count = 1;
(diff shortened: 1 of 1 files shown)


phoenix-core/src/main/java/org/apache/phoenix/hbase/index/IndexRegionObserver.java
```diff
@@ -131,8 +124,8 @@ public long getCount() {
           return count;
       }
 
-      public long getLatestTimestamp() {
-          return latestTimestamp;
+      public boolean isConcurrent() {
+          return concurrent;
```
gokceni: Nit: We don't need a boolean. We can just return count > 1;

kadirozde: I cannot do that because the count is a reference count. I would not know if the row has gone through concurrent updates when the reference count drops to 1.

PR #624, opened 2019-11-10
PHOENIX-5564 Restructure read repair to improve readability and corre…

kadirozde:
…ctness

### DIFF
--- a/phoenix-core/src/main/java/org/apache/phoenix/index/GlobalIndexChecker.java
+++ b/phoenix-core/src/main/java/org/apache/phoenix/index/GlobalIndexChecker.java
@@ -70,10 +70,42 @@ import org.apache.phoenix.util.ServerUtil;
  * 
  * Coprocessor that verifies the scanned rows of a non-transactional global index.
  *
+ * If an index row is unverified (i.e., the row status is unverified), the following steps are taken :
+ * (1) We generate the data row key from the index row key, and check if the data row exists. If not, this unverified
+ * index row is skipped (i.e., not returned to the client), and it is deleted if it is old enough. The age check is
+ * necessary in order not to delete the index rows that are currently being updated. If the data row exists,
+ * we continue with the rest of the steps.
+ * (2) The index row is rebuilt from the data row.
+ * (3) The current scanner is closed as the newly rebuilt row will not be visible to the current scanner.
+ * (4) if the data row does not point back to the unverified index row (i.e., the index row key generated from the data
+ * row does not match with the row key of the unverified index row), this unverified row is skipped and and it is
+ * deleted if it is old enough. A new scanner is opened starting form the index row after this unverified index row.
+ * (5) if the data points back to the unverified index row then, a new scanner is opened starting form the index row.
+ * The next row is scanned to check if it is verified. if it is verified, it is returned to the client. If not, then
+ * it means the data table row timestamp is lower than than the timestamp of the unverified index row, and
+ * the index row that has been rebuilt from the data table row is masked by this unverified row. This happens if the
+ * first phase updates (i.e., unverified index row updates) complete but the second phase updates (i.e., data table
+ * row updates) fail. There could be back to back such events so we need to scan older versions to retrieve
+ * the verified version that is masked by the unverified version(s).
+ *
  */
 public class GlobalIndexChecker implements RegionCoprocessor, RegionObserver {
     private static final Log LOG = LogFactory.getLog(GlobalIndexChecker.class);
     private GlobalIndexCheckerSource metricsSource;
+    public enum RebuildReturnCode {
+        NO_DATA_ROW(0),
+        NO_INDEX_ROW(1),
+        INDEX_ROW_EXISTS(2);
+        private int value;
+
+        RebuildReturnCode(int value) {
+            this.value = value;
+        }
+
+        public int getValue() {
+            return value;
+        }
+    }
 
     /**
      * Class that verifies a given row of a non-transactional global index.
@@ -264,8 +296,8 @@ public class GlobalIndexChecker implements RegionCoprocessor, RegionObserver {
             buildIndexScan.withStartRow(dataRowKey, true);
             buildIndexScan.withStopRow(dataRowKey, true);
             buildIndexScan.setTimeRange(0, maxTimestamp);
-            // Pass the index row key to the partial index builder which will build the index row only when the data
-            // table row for dataRowKey matches with this unverified index row.
+            // Pass the index row key to the partial index builder which will rebuild the index row and check if the
+            // row key of this rebuilt index row matches with the passed index row key
             buildIndexScan.setAttribute(BaseScannerRegionObserver.INDEX_ROW_KEY, indexRowKey);
             Result result = null;
             try (ResultScanner resultScanner = dataHTable.getScanner(buildIndexScan)){
@@ -275,9 +307,10 @@ public class GlobalIndexChecker implements RegionCoprocessor, RegionObserver {
             }
             // A single cell will be returned. We decode that here
             byte[] value = result.value();
-            long rowCount = PLong.INSTANCE.getCodec().decodeLong(new ImmutableBytesWritable(value), SortOrder.getDefault());
-            if (rowCount == 0) {
-                // This means there does not exist a data table row for this unverified index row
+            long code = PLong.INSTANCE.getCodec().decodeLong(new ImmutableBytesWritable(value), SortOrder.getDefault());
+            if (code == RebuildReturnCode.NO_DATA_ROW.getValue()) {
+                // This means there does not exist a data table row for the data row key derived from
+                // this unverified index row. So, no index row has been built
                 // Delete the unverified row from index if it is old enough
                 deleteRowIfAgedEnough(indexRowKey, row, ts, false);
                 // Skip this unverified row (i.e., do not return it to the client). Just retuning empty row is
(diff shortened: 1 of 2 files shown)


phoenix-core/src/main/java/org/apache/phoenix/index/GlobalIndexChecker.java
```diff
@@ -322,28 +370,19 @@ private void repairIndexRows(byte[] indexRowKey, long ts, List<Cell> row) throws
                 singleRowScanner.next(row);
                 singleRowScanner.close();
                 if (row.isEmpty()) {
-                    // This means that the data row did not exist, so we need to skip this unverified row (i.e., do not
-                    // return it to the client). Just retuning empty row is sufficient to do that
-                    return;
+                    throw new DoNotRetryIOException("Could not find the newly rebuilt index row " +
```
gokceni: I am not sure for throwing exception here. If we can't build 1 index row, we will error out and skip the next ones. Would not this create more issues down the line?

kadirozde: This and the previous exception are to catch "impossible" cases. Now that you asked, I want to check more time if they really cannot happen. Maybe logging error and skipping could be safer as you suggested.

kadirozde: I have looked into this again. The first case can be encountered due to compaction. But this one is really an exception as we have a can filter for a specific row but we got a different row. I updated the code accordingly. Thx

gokceni: Thank you @kadirozde !

phoenix-core/src/main/java/org/apache/phoenix/index/GlobalIndexChecker.java
```diff
@@ -70,10 +70,31 @@
...
+ * the index row that has been rebuilt from the data table row is masked by this unverified row. This happens if the
+ * first phase updates (i.e., unverified index row updates) complete but the second phase updates (i.e., data table
+ * row updates) fail. There could be back to back such events so we need to scan older versions to retrieve
+ * the verified version that is masked by the unverified version(s).
+ *
  */
 public class GlobalIndexChecker implements RegionCoprocessor, RegionObserver {
     private static final Log LOG = LogFactory.getLog(GlobalIndexChecker.class);
     private GlobalIndexCheckerSource metricsSource;
+    public static final int NO_DATA_ROW = 0;
```
gjacoby126: Should this be an enum, esp since it's used in the Unaggregated coproc as well?

phoenix-core/src/main/java/org/apache/phoenix/index/GlobalIndexChecker.java
```diff
@@ -275,45 +296,72 @@ private void repairIndexRows(byte[] indexRowKey, long ts, List<Cell> row) throws
...
                 }
-                // The index row is still "unverified" after rebuild. This means either that the data table row timestamp is
-                // lower than than the timestamp of the unverified index row (ts) and the index row that is built from
-                // the data table row is masked by this unverified row, or that the corresponding data table row does
-                // exist
+                // The row is "unverified". Rewind the scanner and let the row be scanned again
+                // so that it can be repaired
+                scanner.close();
+                scanner = region.getScanner(indexScan);
+                row.clear();
```
gjacoby126: Do we not need to return here? We're clearing the scanned row but in the very next block we're checking if it still exists after a new rebuild (and I think the next block is assuming that it's an else{} to this one). Otherwise I don't see where the rebuild of the "new" unverified different-keyed-row is happening.

phoenix-core/src/main/java/org/apache/phoenix/index/GlobalIndexChecker.java
gjacoby126: (Note that this is a genuine question, not just requesting a change politely. :-) )

phoenix-core/src/main/java/org/apache/phoenix/index/GlobalIndexChecker.java
kadirozde: The reason this missing return did not cause any issue is because the row is cleared and the next block of code returns when the row is empty (i.e., verifyRowAndRemoveEmptyColumn(row) returns true when the row is empty):
           if (verifyRowAndRemoveEmptyColumn(row)) {
                // The index row status is "verified". This row is good to return to the client. We are done here.
                return;
            }

phoenix-core/src/main/java/org/apache/phoenix/index/GlobalIndexChecker.java
kadirozde: Done

phoenix-core/src/main/java/org/apache/phoenix/index/GlobalIndexChecker.java
kadirozde: Done

phoenix-core/src/main/java/org/apache/phoenix/index/GlobalIndexChecker.java
```diff
@@ -70,10 +70,42 @@
...
+ * first phase updates (i.e., unverified index row updates) complete but the second phase updates (i.e., data table
+ * row updates) fail. There could be back to back such events so we need to scan older versions to retrieve
+ * the verified version that is masked by the unverified version(s).
+ *
  */
 public class GlobalIndexChecker implements RegionCoprocessor, RegionObserver {
     private static final Log LOG = LogFactory.getLog(GlobalIndexChecker.class);
     private GlobalIndexCheckerSource metricsSource;
+    public enum RebuildReturnCode {
+        NO_DATA_ROW(0),
```
gjacoby126: nit: enum fields in Java usually LookLikeThis NOT_LIKE_THIS.

phoenix-core/src/main/java/org/apache/phoenix/index/GlobalIndexChecker.java
```diff
@@ -275,45 +307,73 @@ private void repairIndexRows(byte[] indexRowKey, long ts, List<Cell> row) throws
             }
             // A single cell will be returned. We decode that here
             byte[] value = result.value();
-            long rowCount = PLong.INSTANCE.getCodec().decodeLong(new ImmutableBytesWritable(value), SortOrder.getDefault());
-            if (rowCount == 0) {
-                // This means there does not exist a data table row for this unverified index row
+            long code = PLong.INSTANCE.getCodec().decodeLong(new ImmutableBytesWritable(value), SortOrder.getDefault());
```
gjacoby126: Rather than doing getValue a bunch, probably better to translate code into the enum once and then refer to it directly.

phoenix-core/src/main/java/org/apache/phoenix/index/GlobalIndexChecker.java
```diff
@@ -322,28 +382,27 @@ private void repairIndexRows(byte[] indexRowKey, long ts, List<Cell> row) throws
...
-                // so that it can be repaired
-                scanner = region.getScanner(indexScan);
-                row.clear();
-            }
+                ts = getMaxTimestamp(row);
+            } while (Bytes.compareTo(row.get(0).getRowArray(), row.get(0).getRowOffset(), row.get(0).getRowLength(),
+                    indexRowKey, 0, indexRowKey.length) == 0);
+            // This should not happen at all
+            Cell cell = row.get(0);
+            byte[] rowKey = new byte[cell.getRowLength()];
```
gjacoby126: Can use CellUtil.cloneRow()

phoenix-core/src/main/java/org/apache/phoenix/index/GlobalIndexChecker.java
kadirozde: I have just checked a bunch of enums used in the Phoenix code. All of them follow the style I used.

phoenix-core/src/main/java/org/apache/phoenix/index/GlobalIndexChecker.java
kadirozde: I appreciate if you give me an example here to follow.

phoenix-core/src/main/java/org/apache/phoenix/index/GlobalIndexChecker.java
kadirozde: I will use it

phoenix-core/src/main/java/org/apache/phoenix/index/GlobalIndexChecker.java
gjacoby126: Fair point. Never mind.

phoenix-core/src/main/java/org/apache/phoenix/index/GlobalIndexChecker.java
gjacoby126: So, it turns out it's messier to translate a long to an enum than an int, so I leave it up to you. But something like:

long result = PLong.INSTANCE.getCodec().decodeLong(new ImmutableBytesWritable(value), SortOrder.getDefault());
RebuildReturnCode returnCode = RebuildReturnCode.values[(Integer) result];

The idea was that you'd have a variable of type RebuildReturnCode, so you wouldn't have to keep converting enum values to longs in multiple places in the code and could just compare the result against the enum values directly. But the narrowing cast makes it a bit messy.

phoenix-core/src/main/java/org/apache/phoenix/index/GlobalIndexChecker.java
kadirozde: Done

PR #625, opened 2019-11-10
PHOENIX-5565 Unify index update structures in IndexRegionObserver and…

kadirozde:
… IndexCommitter

### DIFF
--- a/phoenix-core/src/main/java/org/apache/phoenix/hbase/index/IndexRegionObserver.java
+++ b/phoenix-core/src/main/java/org/apache/phoenix/hbase/index/IndexRegionObserver.java
@@ -33,6 +33,8 @@ import java.util.Map;
 import java.util.Optional;
 import java.util.concurrent.ConcurrentHashMap;
 
+import com.google.common.collect.ArrayListMultimap;
+import com.google.common.collect.ListMultimap;
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
 import org.apache.hadoop.conf.Configuration;
@@ -71,6 +73,7 @@ import org.apache.phoenix.hbase.index.builder.IndexBuilder;
 import org.apache.phoenix.hbase.index.covered.IndexMetaData;
 import org.apache.phoenix.hbase.index.metrics.MetricsIndexerSource;
 import org.apache.phoenix.hbase.index.metrics.MetricsIndexerSourceFactory;
+import org.apache.phoenix.hbase.index.table.HTableInterfaceReference;
 import org.apache.phoenix.hbase.index.util.ImmutableBytesPtr;
 import org.apache.phoenix.hbase.index.util.IndexManagementUtil;
 import org.apache.phoenix.hbase.index.write.IndexWriter;
@@ -149,16 +152,16 @@ public class IndexRegionObserver implements RegionObserver, RegionCoprocessor {
       private final int clientVersion;
       // The collection of index mutations that will be applied before the data table mutations. The empty column (i.e.,
       // the verified column) will have the value false ("unverified") on these mutations
-      private Collection<Pair<Mutation, byte[]>> preIndexUpdates = Collections.emptyList();
+      private ListMultimap<HTableInterfaceReference, Mutation> preIndexUpdates;
       // The collection of index mutations that will be applied after the data table mutations. The empty column (i.e.,
       // the verified column) will have the value true ("verified") on the put mutations
-      private Collection<Pair<Mutation, byte[]>> postIndexUpdates = Collections.emptyList();
+      private ListMultimap<HTableInterfaceReference, Mutation> postIndexUpdates;
       // The collection of candidate index mutations that will be applied after the data table mutations
-      private Collection<Pair<Pair<Mutation, byte[]>, byte[]>> intermediatePostIndexUpdates;
+      private ListMultimap<HTableInterfaceReference, Pair<Mutation, byte[]>> intermediatePostIndexUpdates;
       private List<RowLock> rowLocks = Lists.newArrayListWithExpectedSize(QueryServicesOptions.DEFAULT_MUTATE_BATCH_SIZE);
       private HashSet<ImmutableBytesPtr> rowsToLock = new HashSet<>();
-      long dataWriteStartTime;
-
+      private long dataWriteStartTime;
+      private boolean rebuild;
       private BatchMutateContext(int clientVersion) {
           this.clientVersion = clientVersion;
       }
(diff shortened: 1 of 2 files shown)


phoenix-core/src/main/java/org/apache/phoenix/hbase/index/IndexRegionObserver.java
```diff
@@ -512,86 +515,84 @@ public static void removeEmptyColumn(Mutation m, byte[] emptyCF, byte[] emptyCQ)
...
   }
 
+  private void handleLocalIndexUpdates(ObserverContext<RegionCoprocessorEnvironment> c,
+                                       MiniBatchOperationInProgress<Mutation> miniBatchOp,
+                                       ListMultimap<HTableInterfaceReference, Pair<Mutation, byte[]>> indexUpdates) {
+      byte[] tableName = c.getEnvironment().getRegion().getTableDescriptor().getTableName().getName();
+      HTableInterfaceReference hTableInterfaceReference =
+                          new HTableInterfaceReference(new ImmutableBytesPtr(tableName));
+      List<Pair<Mutation, byte[]>> localIndexUpdates = indexUpdates.removeAll(hTableInterfaceReference);
+      if (localIndexUpdates == null) {
```
gokceni: I think removeAll will return an empty list rather than null. Let's add a size check here as well

phoenix-core/src/main/java/org/apache/phoenix/hbase/index/IndexRegionObserver.java
```diff
@@ -512,86 +515,84 @@ public static void removeEmptyColumn(Mutation m, byte[] emptyCF, byte[] emptyCQ)
...
-                  }
-                  byte[] emptyCF = indexMaintainer.getEmptyKeyValueFamily().copyBytesIfNecessary();
-                  byte[] emptyCQ = indexMaintainer.getEmptyKeyValueQualifier();
+          handleLocalIndexUpdates(c, miniBatchOp, context.intermediatePostIndexUpdates);
+          context.preIndexUpdates = ArrayListMultimap.<HTableInterfaceReference, Mutation>create();
+          int updateCount = 0;
+          for (IndexMaintainer indexMaintainer : maintainers) {
+              updateCount++;
+              byte[] emptyCF = indexMaintainer.getEmptyKeyValueFamily().copyBytesIfNecessary();
+              byte[] emptyCQ = indexMaintainer.getEmptyKeyValueQualifier();
```
gokceni: Why have copyBytesIfNecessary on the emptyCF but not on emptyCQ? Just curious

phoenix-core/src/main/java/org/apache/phoenix/hbase/index/IndexRegionObserver.java
```diff
@@ -616,20 +617,23 @@ public void preBatchMutateWithExceptions(ObserverContext<RegionCoprocessorEnviro
...
-      // Add the table rows in the mini batch to the collection of pending rows. This will be used to detect
-      // concurrent updates
-      if (replayWrite == null) {
+          now = EnvironmentEdgeManager.currentTimeMillis();
+          // Add the table rows in the mini batch to the collection of pending rows. This will be used to detect
+          // concurrent updates
           populatePendingRows(context);
       }
+      else {
+          now = EnvironmentEdgeManager.currentTimeMillis();
```
gokceni: Why not move now = Env.. to outside of this if. It would simplify the code.

phoenix-core/src/main/java/org/apache/phoenix/hbase/index/IndexRegionObserver.java
kadirozde: Why do you think so? Please notice the removeAll returns List.

kadirozde: I misunderstood your comment. I thought you implied that we cannot use removeAll here. I will add the size check although it is not needed for correctness as the rest of the code will not do anything if the list is empty.

phoenix-core/src/main/java/org/apache/phoenix/hbase/index/IndexRegionObserver.java
kadirozde: I do not know the reason actually. I must have copied the code from somewhere else. I will look into it.

phoenix-core/src/main/java/org/apache/phoenix/hbase/index/IndexRegionObserver.java
kadirozde: We cannot because we need to get the current time only after acquiring the locks on the regular (non rebuild) code path.

phoenix-core/src/main/java/org/apache/phoenix/hbase/index/IndexRegionObserver.java
```diff
@@ -149,16 +152,16 @@ public static void setFailDataTableUpdatesForTesting(boolean fail) {
       private final int clientVersion;
       // The collection of index mutations that will be applied before the data table mutations. The empty column (i.e.,
       // the verified column) will have the value false ("unverified") on these mutations
-      private Collection<Pair<Mutation, byte[]>> preIndexUpdates = Collections.emptyList();
+      private ListMultimap<HTableInterfaceReference, Mutation> preIndexUpdates;
```
gjacoby126: Doesn't have to be this patch, but we should consider extracting this ListMultimap<HTableInterfaceReference, Mutation> into a class of some kind with named accessor methods so that code dealing with it can be at a higher level of abstraction / more self-documenting.

phoenix-core/src/main/java/org/apache/phoenix/hbase/index/builder/IndexBuildManager.java
```diff
@@ -79,28 +82,20 @@ public IndexMetaData getIndexMetaData(MiniBatchOperationInProgress<Mutation> min
...
       Collection<Pair<Mutation, byte[]>> updates = delegate.getIndexUpdate(m, indexMetaData);
-      if (PhoenixIndexMetaData.isIndexRebuild(m.getAttributesMap())) {
-          for (Pair<Mutation, byte[]> update : updates) {
-            update.getFirst().setAttribute(BaseScannerRegionObserver.REPLAY_WRITES,
-                BaseScannerRegionObserver.REPLAY_INDEX_REBUILD_WRITES);
-          }
-      }
       for (Pair<Mutation, byte[]> update : updates) {
-        results.add(new Pair<>(update, m.getRow()));
+        indexUpdates.put(new HTableInterfaceReference(new ImmutableBytesPtr(update.getSecond())), new Pair<>(update.getFirst(), m.getRow()));
```
gjacoby126: nit: wouldn't it be more efficient to reuse HTableInterfaceReference rather than create N of them when we're likely mutating only a small number of tables?

phoenix-core/src/main/java/org/apache/phoenix/hbase/index/IndexRegionObserver.java
kadirozde: It is necessary as what is returned is an ImmutableBytesPtr which may be pointing to a non-zero offset of another byte array.

phoenix-core/src/main/java/org/apache/phoenix/hbase/index/builder/IndexBuildManager.java
kadirozde: It is not easy to optimize this as for each data table mutation we get updates for all index tables. If we do not create a separate reference for each update, then we need to maintain a hash or list of references and then need to look up on or search them. This may not be more efficient. Let me know if you had something else to suggest here.

phoenix-core/src/main/java/org/apache/phoenix/hbase/index/IndexRegionObserver.java
kadirozde: Done

phoenix-core/src/main/java/org/apache/phoenix/hbase/index/IndexRegionObserver.java
swaroopak: Do we have a jira for this? @gjacoby126  If not, I ll file one.

gjacoby126: There's no JIRA that I know of, @swaroopak. Thanks!

kadirozde: Yes, there is a JIRA on this. It is https://issues.apache.org/jira/browse/PHOENIX-5339

PR #626, opened 2019-11-12
PHOENIX-5531: IndexUpgradeTool crashes for tables without any indexes…

swaroopak:
… + Sleep problems

### DIFF
--- a/phoenix-core/src/main/java/org/apache/phoenix/mapreduce/index/IndexUpgradeTool.java
+++ b/phoenix-core/src/main/java/org/apache/phoenix/mapreduce/index/IndexUpgradeTool.java
@@ -73,8 +73,7 @@ import java.util.UUID;
 import java.util.logging.FileHandler;
 import java.util.logging.SimpleFormatter;
 
-import static org.apache.phoenix.query.QueryServicesOptions.
-        GLOBAL_INDEX_CHECKER_ENABLED_MAP_EXPIRATION_MIN;
+import static org.apache.phoenix.query.QueryServicesOptions.GLOBAL_INDEX_CHECKER_ENABLED_MAP_EXPIRATION_MIN;
 
 public class IndexUpgradeTool extends Configured implements Tool {
 
@@ -96,7 +95,8 @@ public class IndexUpgradeTool extends Configured implements Tool {
     private static final Option LOG_FILE_OPTION = new Option("lf", "logfile",
             true,
             "Log file path where the logs are written");
-    private static final Option INDEX_SYNC_REBUILD_OPTION = new Option("sr","index-sync-rebuild",
+    private static final Option INDEX_SYNC_REBUILD_OPTION = new Option("sr",
+            "index-sync-rebuild",
             false,
             "[Optional]Whether or not synchronously rebuild the indexes; "
                     + "default rebuild asynchronous");
@@ -107,7 +107,6 @@ public class IndexUpgradeTool extends Configured implements Tool {
     private IndexTool indexingTool;
 
     private HashMap<String, HashSet<String>> tablesAndIndexes = new HashMap<>();
-    private HashMap<String, HashMap<String,IndexInfo>> rebuildMap = new HashMap<>();
     private HashMap<String, String> prop = new  HashMap<>();
 
     private boolean dryRun, upgrade, syncRebuild;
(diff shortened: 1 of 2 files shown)


phoenix-core/src/main/java/org/apache/phoenix/mapreduce/index/IndexUpgradeTool.java
```diff
@@ -418,7 +418,9 @@ private void rebuildIndexes(Connection conn, Configuration conf, String dataTabl
                 indexingTool = new IndexTool();
             }
             indexingTool.setConf(conf);
-            rebuildIndexes(conn, dataTableFullName, indexingTool);
+            if(rebuildMap.get(dataTableFullName).size() != 0 ) {
```
priyankporwal: I believe the crash was since rebuildMap didn't have anything for dataTableFullName. This change only moves the crash point.

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/index/IndexUpgradeTool.java
```diff
@@ -333,6 +333,8 @@ private int executeTool(Connection conn, ConnectionQueryServices queryServices,
                 LOGGER.info("Executing " + operation + " for " + dataTableFullName);
 
                 boolean mutable = !(dataTable.isImmutableRows());
+                disableTable(admin, dataTableFullName, indexes);
+                modifyTable(admin, dataTableFullName, indexes);
                 if (!mutable) {
```
priyankporwal: Also, I think we should have a single sleep for the given batch of tables.

swaroopak: Let's say there are 2 immutable tables in the list and after waiting for cache expiration for the first time and while remaining steps are in progress, the second table's GIC boolean is loaded in the cache due to some other read/write request. Now, we have not upgraded the second table yet. However, we have waited for cache to expire before and hence if we won't, now, the old GIC boolean value will be there in the cache (for second table) even though the server has changed the coproc for the second table. Does this make sense? 

On the other side of the code, is it possible to invalidate the cache after server-side values have been updated or update the cache itself via server? @gjacoby126 @priyankporwal

gjacoby126: That would require having all tables in the batch be disabled at once, right? So it's a tradeoff -- the process would be quicker, and the overall pain less, but the pain that's there would be greater.

swaroopak: I would prefer waiting each time there is an immutable table if invalidating cache is not an option. what say?

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/index/IndexUpgradeTool.java
swaroopak: done. Used the file you shared as a starting point.

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/index/IndexUpgradeTool.java
```diff
@@ -321,38 +320,50 @@ public int executeTool() {
...
+                if (dataTable.isImmutableRows()) {
+                    // If the table is immutable, we need to wait for clients to purge
+                    // their caches of table metadata
+                    LOGGER.info(dataTableFullName + " is an immutable table, waiting for "
                             + (GLOBAL_INDEX_CHECKER_ENABLED_MAP_EXPIRATION_MIN + 1)
                             + " minutes for client cache to expire");
-                    if (!test) {
-                        Thread.sleep((GLOBAL_INDEX_CHECKER_ENABLED_MAP_EXPIRATION_MIN + 1)
-                                * 60 * 1000);
+                    if (!(test | dryRun)) {
```
priyankporwal: || instead of |, right?

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/index/IndexUpgradeTool.java
```diff
@@ -321,38 +320,50 @@ public int executeTool() {
...
+                        try {
+                            Thread.sleep((GLOBAL_INDEX_CHECKER_ENABLED_MAP_EXPIRATION_MIN + 1)
```
gjacoby126: Do we need to sleep if the immutable table has no index?

swaroopak: you are right! we should not sleep.

swaroopak: or maybe not! On second thought, it seems okay.

swaroopak:
Thank you @gjacoby126  and @priyankporwal for the approval.

dbwong:
> Can we have a unit tests for these changes?

I saw this was already merged without reply to my comment.  Can we consider more unit test for pieces of code that are easily unit testable?    @priyankporwal @gjacoby126 @swaroopak The index tool is new code.  It should be built in a testable fashion.

swaroopak:
@dbwong The unit test was added as a response to your comment. I didn't see any reply option on your comment.

dbwong:
I don’t see a UT I only see an IT.  Am I missing something?

PR #627, opened 2019-11-12
PHOENIX-5561: Replace Java8 code

gokceni:


gjacoby126:
I see you're checking this into master. Just want to point out that this is only a problem in the 4.x branches; JDK 8 is fine in master. It's also fine if you want to keep the code in sync in both places.

gokceni:
Thanks @gjacoby126 ! Yes I would like to keep them in sync

gjacoby126:
JIRA is resolved, closing PR

PR #628, opened 2019-11-12
PHOENIX-5546: TASK_TS being set as HConstants.LATEST_TIMESTAMP in SYSTEM.TASK table

ChinmaySKulkarni:
Please review @twdsilva @lhofhansl @kadirozde

### DIFF
--- a/phoenix-core/src/main/java/org/apache/phoenix/execute/MutationState.java
+++ b/phoenix-core/src/main/java/org/apache/phoenix/execute/MutationState.java
@@ -42,7 +42,6 @@ import javax.annotation.Nonnull;
 import javax.annotation.concurrent.Immutable;
 
 import org.apache.hadoop.hbase.HConstants;
-import org.apache.hadoop.hbase.TableName;
 import org.apache.hadoop.hbase.client.Delete;
 import org.apache.hadoop.hbase.client.Mutation;
 import org.apache.hadoop.hbase.client.Put;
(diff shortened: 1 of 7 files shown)


phoenix-core/src/main/java/org/apache/phoenix/execute/MutationState.java
```diff
@@ -776,8 +771,14 @@ private long validateAndGetServerTimestamp(TableRef tableRef, MultiRowMutationSt
         // If we're auto committing, we've already validated the schema when we got the ColumnResolver,
         // so no need to do it again here.
         PTable table = tableRef.getTable();
-        MetaDataMutationResult result = client.updateCache(table.getSchemaName().getString(), table.getTableName()
-                .getString());
+
+        // We generally don't re-resolve SYSTEM tables, but if it relies on ROW_TIMESTAMP, we must
```
twdsilva: Would it make sense to change the check we have here for system tables to also check if the system table doesn't have ROW_TIMESTAMP column instead?
https://github.com/apache/phoenix/blob/dd5551b4e8cc17e5bd587e6c4f5f256a8212345a/phoenix-core/src/main/java/org/apache/phoenix/schema/MetaDataClient.java#L606

ChinmaySKulkarni: @twdsilva I think that makes sense. I will update the PR. Thanks!

phoenix-core/src/main/java/org/apache/phoenix/schema/MetaDataClient.java
```diff
@@ -725,6 +719,20 @@ public MetaDataMutationResult updateCache(PName origTenantId, String schemaName,
         return result;
     }
 
+    // Do not make rpc to getTable if
+    // 1. table is a system table that does not have a ROW_TIMESTAMP column OR
+    // 2. table was already resolved as of that timestamp OR
+    // 3. table does not have a ROW_TIMESTAMP column and age is less then UPDATE_CACHE_FREQUENCY
+    private boolean avoidRpcToGetTable(boolean alwaysHitServer, Long resolvedTimestamp,
```
ChinmaySKulkarni: @twdsilva made the change as per your review. Please take a look.

ChinmaySKulkarni:
Thanks for the reviews @twdsilva @kadirozde

gjacoby126:
JIRA resolved, closing

PR #629, opened 2019-11-12
PHOENIX-5559 Fix remaining issues with Long viewIndexIds.

lhofhansl:
See https://issues.apache.org/jira/browse/PHOENIX-5559

### DIFF
--- a/phoenix-core/src/it/java/org/apache/phoenix/end2end/BaseTenantSpecificViewIndexIT.java
+++ b/phoenix-core/src/it/java/org/apache/phoenix/end2end/BaseTenantSpecificViewIndexIT.java
@@ -140,18 +140,18 @@ public class BaseTenantSpecificViewIndexIT extends SplitSystemCatalogIT {
         ResultSet rs = conn.createStatement().executeQuery("EXPLAIN SELECT k1, k2, v2 FROM " + viewName + " WHERE v2='" + valuePrefix + "v2-1'");
         if(localIndex){
             assertEquals(saltBuckets == null ? 
-                    "CLIENT PARALLEL 1-WAY RANGE SCAN OVER " + tableName + " [" + Long.toString(1L + expectedIndexIdOffset) + ",'" + tenantId + "','" + valuePrefix + "v2-1']\n"
+                    "CLIENT PARALLEL 1-WAY RANGE SCAN OVER " + tableName + " [" + (1L + expectedIndexIdOffset) + ",'" + tenantId + "','" + valuePrefix + "v2-1']\n"
                             + "    SERVER FILTER BY FIRST KEY ONLY\n"
                             + "CLIENT MERGE SORT" :
-                    "CLIENT PARALLEL 3-WAY RANGE SCAN OVER " + tableName + " [" + Long.toString(1L + expectedIndexIdOffset) + ",'" + tenantId + "','" + valuePrefix + "v2-1']\n"
+                    "CLIENT PARALLEL 3-WAY RANGE SCAN OVER " + tableName + " [" + (1L + expectedIndexIdOffset) + ",'" + tenantId + "','" + valuePrefix + "v2-1']\n"
                             + "    SERVER FILTER BY FIRST KEY ONLY\n"
                             + "CLIENT MERGE SORT", QueryUtil.getExplainPlan(rs));
         } else {
             String expected = saltBuckets == null ? 
-                    "CLIENT PARALLEL 1-WAY RANGE SCAN OVER _IDX_" + tableName + " [" + Long.toString(Short.MIN_VALUE + expectedIndexIdOffset) + ",'" + tenantId + "','" + valuePrefix + "v2-1']\n"
+                    "CLIENT PARALLEL 1-WAY RANGE SCAN OVER _IDX_" + tableName + " [" + (Short.MIN_VALUE + expectedIndexIdOffset) + ",'" + tenantId + "','" + valuePrefix + "v2-1']\n"
                             + "    SERVER FILTER BY FIRST KEY ONLY" :
-                    "CLIENT PARALLEL 3-WAY RANGE SCAN OVER _IDX_" + tableName + " [0," + Long.toString(Short.MIN_VALUE + expectedIndexIdOffset) + ",'" + tenantId + "','" + valuePrefix +
-                        "v2-1'] - ["+(saltBuckets.intValue()-1)+"," + Long.toString(Short.MIN_VALUE + expectedIndexIdOffset) + ",'" + tenantId + "','" + valuePrefix + "v2-1']\n"
+                    "CLIENT PARALLEL 3-WAY RANGE SCAN OVER _IDX_" + tableName + " [0," + (Short.MIN_VALUE + expectedIndexIdOffset) + ",'" + tenantId + "','" + valuePrefix +
+                        "v2-1'] - ["+(saltBuckets.intValue()-1)+"," + (Short.MIN_VALUE + expectedIndexIdOffset) + ",'" + tenantId + "','" + valuePrefix + "v2-1']\n"
 
                   + "    SERVER FILTER BY FIRST KEY ONLY\n"
                   + "CLIENT MERGE SORT";
--- a/phoenix-core/src/it/java/org/apache/phoenix/end2end/BaseTenantSpecificViewIndexIT.java
+++ b/phoenix-core/src/it/java/org/apache/phoenix/end2end/BaseTenantSpecificViewIndexIT.java
@@ -148,10 +148,10 @@ public class BaseTenantSpecificViewIndexIT extends SplitSystemCatalogIT {
                             + "CLIENT MERGE SORT", QueryUtil.getExplainPlan(rs));
         } else {
             String expected = saltBuckets == null ? 
-                    "CLIENT PARALLEL 1-WAY RANGE SCAN OVER _IDX_" + tableName + " [" + Long.toString(Long.MIN_VALUE + expectedIndexIdOffset) + ",'" + tenantId + "','" + valuePrefix + "v2-1']\n"
+                    "CLIENT PARALLEL 1-WAY RANGE SCAN OVER _IDX_" + tableName + " [" + Long.toString(Short.MIN_VALUE + expectedIndexIdOffset) + ",'" + tenantId + "','" + valuePrefix + "v2-1']\n"
                             + "    SERVER FILTER BY FIRST KEY ONLY" :
-                    "CLIENT PARALLEL 3-WAY RANGE SCAN OVER _IDX_" + tableName + " [0," + Long.toString(Long.MIN_VALUE + expectedIndexIdOffset) + ",'" + tenantId + "','" + valuePrefix +
-                        "v2-1'] - ["+(saltBuckets.intValue()-1)+"," + Long.toString(Long.MIN_VALUE + expectedIndexIdOffset) + ",'" + tenantId + "','" + valuePrefix + "v2-1']\n"
+                    "CLIENT PARALLEL 3-WAY RANGE SCAN OVER _IDX_" + tableName + " [0," + Long.toString(Short.MIN_VALUE + expectedIndexIdOffset) + ",'" + tenantId + "','" + valuePrefix +
+                        "v2-1'] - ["+(saltBuckets.intValue()-1)+"," + Long.toString(Short.MIN_VALUE + expectedIndexIdOffset) + ",'" + tenantId + "','" + valuePrefix + "v2-1']\n"
 
                   + "    SERVER FILTER BY FIRST KEY ONLY\n"
                   + "CLIENT MERGE SORT";
--- a/phoenix-core/src/it/java/org/apache/phoenix/end2end/BaseViewIT.java
+++ b/phoenix-core/src/it/java/org/apache/phoenix/end2end/BaseViewIT.java
@@ -197,8 +197,8 @@ public abstract class BaseViewIT extends ParallelStatsEnabledIT {
                 queryPlan);
         } else {
             assertEquals(saltBuckets == null
-                    ? "CLIENT PARALLEL 1-WAY RANGE SCAN OVER " + viewIndexPhysicalName +" [" + Long.MIN_VALUE + ",51]"
-                            : "CLIENT PARALLEL " + saltBuckets + "-WAY RANGE SCAN OVER " + viewIndexPhysicalName + " [0," + Long.MIN_VALUE + ",51] - ["+(saltBuckets.intValue()-1)+"," + Long.MIN_VALUE + ",51]\nCLIENT MERGE SORT",
+                    ? "CLIENT PARALLEL 1-WAY RANGE SCAN OVER " + viewIndexPhysicalName +" [" + Short.MIN_VALUE + ",51]"
+                            : "CLIENT PARALLEL " + saltBuckets + "-WAY RANGE SCAN OVER " + viewIndexPhysicalName + " [0," + Short.MIN_VALUE + ",51] - ["+(saltBuckets.intValue()-1)+"," + Short.MIN_VALUE + ",51]\nCLIENT MERGE SORT",
                             queryPlan);
         }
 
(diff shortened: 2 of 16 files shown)


phoenix-core/src/main/java/org/apache/phoenix/coprocessor/BaseScannerRegionObserver.java
```diff
@@ -178,8 +178,11 @@ private static void throwIfScanOutOfRegion(Scan scan, Region region) throws DoNo
             byte[] expectedUpperRegionKey =
                     scan.getAttribute(EXPECTED_UPPER_REGION_KEY) == null ? scan.getStopRow() : scan
                             .getAttribute(EXPECTED_UPPER_REGION_KEY);
-            isStaleRegionBoundaries = expectedUpperRegionKey != null &&
-                    Bytes.compareTo(upperExclusiveRegionKey, expectedUpperRegionKey) != 0;
+
```
lhofhansl: The following is the tricky part. Previous in order to detect a stale region boundary cache we only looked at the upper region bound. That work fine as long as do not have an open scanner that has already reached the upper half of the region when we split. When the scan restarts after the split finishes, the scan restarts in the new upper daugther region and the upper might be the same (it definitely is when this was the last region of the table).

That always was a problem but since we're starting with larger number of the viewIndexId this is just more likely. My solution is to use the Scan's expected start key and compare that. For everything I tested that does the right thing.

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/MetaDataEndpointImpl.java
```diff
@@ -2082,7 +2082,7 @@ private long getViewIndexSequenceValue(PhoenixConnection connection, String tena
         long sequenceTimestamp = HConstants.LATEST_TIMESTAMP;
         try {
             connection.getQueryServices().createSequence(key.getTenantId(), key.getSchemaName(), key.getSequenceName(),
-                    Long.MIN_VALUE, 1, 1, Long.MIN_VALUE, Long.MAX_VALUE, false, sequenceTimestamp);
+                    Short.MIN_VALUE, 1, 1, Long.MIN_VALUE, Long.MAX_VALUE, false, sequenceTimestamp);
```
lhofhansl: This is the core change. In order to work with both short and long view index ids, we start with Short.MIN_VALUE, which can be encoded both as short and long.

The bulk of the rest are test fixes for tests that happen to verify the generate index id.

phoenix-core/src/main/java/org/apache/phoenix/util/MetaDataUtil.java
```diff
@@ -941,6 +941,16 @@ public static IndexType getIndexType(List<Mutation> tableMetaData, KeyValueBuild
         return null;
     }
 
+    public static PDataType<?> getIndexDataType(List<Mutation> tableMetaData,
+            KeyValueBuilder builder, ImmutableBytesWritable value) {
+        if (getMutationValue(getPutOnlyTableHeaderRow(tableMetaData),
```
gjacoby126: nit: comment would be really nice here. I _think_ you're looking up the view index id config setting from syscat, and that makes sense, but the code's a mouthful without explanation :-)

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/BaseScannerRegionObserver.java
lhofhansl: The main part I could use a double-check on... Did I understand the purpose of SCAN_ACTUAL_START_ROW correctly?
From looking at the code and going by the name it's the start row of the scan that we can use in Phoenix to restart the scan (and that can happen exactly when a split occurred and the scanner was restarted by HBase with the new region start key - assuming it fell into the second daughter region). If that is indeed the right interpretation that this should be correct.

phoenix-core/src/it/java/org/apache/phoenix/end2end/BaseTenantSpecificViewIndexIT.java
```diff
@@ -148,10 +148,10 @@ private void createAndVerifyIndex(Connection conn, String viewName, String table
                             + "CLIENT MERGE SORT", QueryUtil.getExplainPlan(rs));
         } else {
             String expected = saltBuckets == null ? 
-                    "CLIENT PARALLEL 1-WAY RANGE SCAN OVER _IDX_" + tableName + " [" + Long.toString(Long.MIN_VALUE + expectedIndexIdOffset) + ",'" + tenantId + "','" + valuePrefix + "v2-1']\n"
+                    "CLIENT PARALLEL 1-WAY RANGE SCAN OVER _IDX_" + tableName + " [" + Long.toString(Short.MIN_VALUE + expectedIndexIdOffset) + ",'" + tenantId + "','" + valuePrefix + "v2-1']\n"
```
ChinmaySKulkarni: nit: Use `Short.toString` instead.

phoenix-core/src/it/java/org/apache/phoenix/end2end/BaseTenantSpecificViewIndexIT.java
```diff
@@ -148,10 +148,10 @@ private void createAndVerifyIndex(Connection conn, String viewName, String table
...
                             + "    SERVER FILTER BY FIRST KEY ONLY" :
-                    "CLIENT PARALLEL 3-WAY RANGE SCAN OVER _IDX_" + tableName + " [0," + Long.toString(Long.MIN_VALUE + expectedIndexIdOffset) + ",'" + tenantId + "','" + valuePrefix +
-                        "v2-1'] - ["+(saltBuckets.intValue()-1)+"," + Long.toString(Long.MIN_VALUE + expectedIndexIdOffset) + ",'" + tenantId + "','" + valuePrefix + "v2-1']\n"
+                    "CLIENT PARALLEL 3-WAY RANGE SCAN OVER _IDX_" + tableName + " [0," + Long.toString(Short.MIN_VALUE + expectedIndexIdOffset) + ",'" + tenantId + "','" + valuePrefix +
+                        "v2-1'] - ["+(saltBuckets.intValue()-1)+"," + Long.toString(Short.MIN_VALUE + expectedIndexIdOffset) + ",'" + tenantId + "','" + valuePrefix + "v2-1']\n"
```
ChinmaySKulkarni: ditto here.

phoenix-core/src/it/java/org/apache/phoenix/end2end/BaseViewIT.java
```diff
@@ -238,9 +238,9 @@ protected String testUpdatableView(Integer saltBuckets) throws Exception {
         } else {
             physicalTableName = viewIndexPhysicalName;
             assertEquals(saltBuckets == null
-                    ? "CLIENT PARALLEL 1-WAY RANGE SCAN OVER " + viewIndexPhysicalName +" [" + (Long.MIN_VALUE+1) + ",'foo']\n"
+                    ? "CLIENT PARALLEL 1-WAY RANGE SCAN OVER " + viewIndexPhysicalName +" [" + (Short.MIN_VALUE+1) + ",'foo']\n"
                             + "    SERVER FILTER BY FIRST KEY ONLY"
-                            : "CLIENT PARALLEL " + saltBuckets + "-WAY RANGE SCAN OVER " + viewIndexPhysicalName + " [0," + (Long.MIN_VALUE+1) + ",'foo'] - ["+(saltBuckets.intValue()-1)+"," + (Long.MIN_VALUE+1) + ",'foo']\n"
+                            : "CLIENT PARALLEL " + saltBuckets + "-WAY RANGE SCAN OVER " + viewIndexPhysicalName + " [0," + (Short.MIN_VALUE+1) + ",'foo'] - ["+(saltBuckets.intValue()-1)+"," + (Long.MIN_VALUE+1) + ",'foo']\n"
```
ChinmaySKulkarni: Shouldn't the last `Long.MIN_VALUE` also be changed to `Short.MIN_VALUE`?

phoenix-core/src/it/java/org/apache/phoenix/end2end/UpgradeIT.java
```diff
@@ -188,7 +188,7 @@ public void testMapTableToNamespaceDuringUpgrade()
             PName physicalName = PNameFactory.newName(hbaseTableName);
             String newSchemaName = MetaDataUtil.getViewIndexSequenceSchemaName(physicalName, true);
             String newSequenceName = MetaDataUtil.getViewIndexSequenceName(physicalName, tenantId, true);
-            verifySequenceValue(null, newSequenceName, newSchemaName, -9223372036854775805L);
+            verifySequenceValue(null, newSequenceName, newSchemaName, -32765);
```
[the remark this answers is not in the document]
ChinmaySKulkarni: +1

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/BaseScannerRegionObserver.java
ChinmaySKulkarni: Can you please add a comment here with a short version of the explanation above?

phoenix-core/src/it/java/org/apache/phoenix/end2end/index/ShortViewIndexIdIT.java
```diff
@@ -0,0 +1,104 @@
...
+            // we're expecting exactly one row mapping the column value to the key
+            assertTrue(rs.next());
+            assertEquals(3, rs.getInt(1));
+            assertEquals(1, rs.getInt(2));
+            assertFalse(rs.next());
+            rs.close();
+        }
+    }
+
+}
```
ChinmaySKulkarni: Any way we can add a test for the stale region boundaries change as well?

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/BaseScannerRegionObserver.java
```diff
@@ -178,8 +178,11 @@ private static void throwIfScanOutOfRegion(Scan scan, Region region) throws DoNo
...
+            byte[] actualStartRow = scan.getAttribute(SCAN_ACTUAL_START_ROW);
+            isStaleRegionBoundaries = (expectedUpperRegionKey != null &&
+                    Bytes.compareTo(upperExclusiveRegionKey, expectedUpperRegionKey) != 0) || 
+                    (actualStartRow != null && Bytes.compareTo(actualStartRow, lowerInclusiveRegionKey) < 0);
         } else {
             isStaleRegionBoundaries = Bytes.compareTo(lowerInclusiveScanKey, lowerInclusiveRegionKey) < 0 ||
```
ChinmaySKulkarni: Why is this logic different for the non-local index case? I guess the only thing that should change is how we extract the start and end keys of the scan, right? For ex, in the case of a local index scan, these would be passed as scan attributes and for normal scans they are just the startRow and endRow of the scan, or am I not understanding this correctly?

phoenix-core/src/main/java/org/apache/phoenix/iterate/ExplainTable.java
```diff
@@ -215,7 +215,7 @@ private void appendPKColumnValue(StringBuilder buf, byte[] range, Boolean isNull
     private Long getViewIndexValue(PDataType type, byte[] range) {
         boolean useLongViewIndex = MetaDataUtil.getViewIndexIdDataType().equals(type);
         Object s = type.toObject(range);
-        return (useLongViewIndex ? (Long) s : (Short) s) + (useLongViewIndex ? Long.MAX_VALUE : Short.MAX_VALUE) + 2;
+        return (useLongViewIndex ? (Long) s : (Short) s) + Short.MAX_VALUE + 2;
```
ChinmaySKulkarni: I guess this isn't relevant to your change, but is the `+2` for the 'L#' characters? If yes, can we change this to `LOCAL_INDEX_COLUMN_FAMILY_PREFIX.length` for better readability?

phoenix-core/src/it/java/org/apache/phoenix/end2end/BaseViewIT.java
lhofhansl: Hmm... Good point. Not sure how this test passes.

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/BaseScannerRegionObserver.java
lhofhansl: This is because for local indexes we cannot (ever) proceed when the region was split, we have to restart, since we cannot exactly know where the index keys were split to, they sort differently than the main table's keys.
It's not enough that the new scan is fully "contained" in the new region. Is that end "old" scan boundaries are contained on the lower side and exact on the upper side.

phoenix-core/src/main/java/org/apache/phoenix/iterate/ExplainTable.java
lhofhansl: I actually don't know. :( I just made sure it behaves like before.

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/BaseScannerRegionObserver.java
ChinmaySKulkarni: @lhofhansl I think that makes sense given what's happening when you get a [StaleRegionBoundaryCacheException](https://github.com/apache/phoenix/blob/f8cfa6b8b9f63ab080cdafa486bb4aabf043b1f8/phoenix-core/src/main/java/org/apache/phoenix/iterate/BaseResultIterators.java#L1315) and call BaseResultIterators#[recreateIterators](https://github.com/apache/phoenix/blob/f8cfa6b8b9f63ab080cdafa486bb4aabf043b1f8/phoenix-core/src/main/java/org/apache/phoenix/iterate/BaseResultIterators.java#L1405)

phoenix-core/src/main/java/org/apache/phoenix/util/MetaDataUtil.java
lhofhansl: Done.

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/BaseScannerRegionObserver.java
lhofhansl: Yeah. Good idea.

phoenix-core/src/it/java/org/apache/phoenix/end2end/index/ShortViewIndexIdIT.java
lhofhansl: That's tested in MutableIndexSplitForwardScanIT and MutableIndexSplitReverseScanIT. Those tests fails with the first version of the change.

phoenix-core/src/it/java/org/apache/phoenix/end2end/BaseTenantSpecificViewIndexIT.java
lhofhansl: I mostly restore this part from before PHOENIX-3547.
Realizing now it was all over the place. Using  + Short.MIN_VALUE is some place, -32768 in others, toString in yet others.
Here it does not matter though, the output is identical.

lhofhansl: In fact this is best expressed as "string" + (number1 + number2) + "string", anyway. I'll make that change.

ChinmaySKulkarni: Sounds good, thanks

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/BaseScannerRegionObserver.java
ChinmaySKulkarni: Makes sense. Thanks for the explanation.

ChinmaySKulkarni:
Closing since JIRA is resolved.

stoty:
Already merged.

PR #630, opened 2019-11-13
PHOENIX-5545: DropChildViews Task fails for a base table when its child view has an index

ChinmaySKulkarni:
Please review @kadirozde @lhofhansl @twdsilva


phoenix-core/src/it/java/org/apache/phoenix/end2end/ViewMetadataIT.java
```diff
@@ -186,7 +190,8 @@ public void testCreateViewMappedToExistingHbaseTableWithNamespaceMappingEnabled(
 
             // test for a view that is in non-default schema
             {
-                HTableDescriptor desc = new HTableDescriptor(TableName.valueOf(NS, TBL));
+                tableName = TableName.valueOf(NS, TBL);
```
dbwong: Why did we move the declarations out of the block?  I'm assuming the declarations were local so we would be less likely to have same variable reused with wrong values scenario.

ChinmaySKulkarni: Hmm I think I pushed this change accidentally. I will revert it

ChinmaySKulkarni:
Please review now. I think it is okay to swallow the TNFE inside `dropChildViews` directly. Wdyt @twdsilva?

PR #631, opened 2019-11-13
PHOENIX-5494 Batched, mutable Index updates are unnecessarily run one…

kadirozde:
…-by-one

### DIFF
--- a/phoenix-core/src/main/java/org/apache/phoenix/hbase/index/IndexRegionObserver.java
+++ b/phoenix-core/src/main/java/org/apache/phoenix/hbase/index/IndexRegionObserver.java
@@ -638,7 +638,17 @@ public class IndexRegionObserver implements RegionObserver, RegionCoprocessor {
       }
 
       long start = EnvironmentEdgeManager.currentTimeMillis();
-      prepareIndexMutations(c, miniBatchOp, context, mutations, now, indexMetaData);
+      if (replayWrite == null) {
+          this.builder.scanCurrentRowStates(context.rowsToLock, indexMetaData, now);
+      }
+      try {
+          prepareIndexMutations(c, miniBatchOp, context, mutations, now, indexMetaData);
+      }
+      finally {
+          if (replayWrite == null) {
+              this.builder.removeRowStates(context.rowsToLock);
+          }
+      }
       metricSource.updateIndexPrepareTime(EnvironmentEdgeManager.currentTimeMillis() - start);
 
       // Sleep for one millisecond if we have prepared the index updates in less than 1 ms. The sleep is necessary to
--- a/phoenix-core/src/main/java/org/apache/phoenix/hbase/index/builder/IndexBuilder.java
+++ b/phoenix-core/src/main/java/org/apache/phoenix/hbase/index/builder/IndexBuilder.java
@@ -21,6 +21,7 @@ import java.io.IOException;
 import java.util.Collection;
 import java.util.List;
 import java.util.Map;
+import java.util.Set;
 
 import org.apache.hadoop.hbase.Cell;
 import org.apache.hadoop.hbase.Stoppable;
@@ -34,6 +35,8 @@ import org.apache.hadoop.hbase.util.Pair;
 import org.apache.phoenix.coprocessor.BaseScannerRegionObserver.ReplayWrite;
 import org.apache.phoenix.hbase.index.Indexer;
 import org.apache.phoenix.hbase.index.covered.IndexMetaData;
+import org.apache.phoenix.hbase.index.covered.update.ColumnReference;
+import org.apache.phoenix.hbase.index.util.ImmutableBytesPtr;
 
 /**
  * Interface to build updates ({@link Mutation}s) to the index tables, based on the primary table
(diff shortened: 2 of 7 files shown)


dbwong:
While existing tests may or may not be sufficient for checking behavior if we want to check that they are not running one by one can we at least add a UT?

phoenix-core/src/main/java/org/apache/phoenix/hbase/index/IndexRegionObserver.java
```diff
@@ -638,7 +638,13 @@ public void preBatchMutateWithExceptions(ObserverContext<RegionCoprocessorEnviro
       }
 
       long start = EnvironmentEdgeManager.currentTimeMillis();
+      if (replayWrite == null) {
+          this.builder.scanCurrentRowStates(context.rowsToLock, indexMetaData, now);
+      }
       prepareIndexMutations(c, miniBatchOp, context, mutations, now, indexMetaData);
+      if (replayWrite == null) {
+          this.builder.removeRowStates(context.rowsToLock);
```
gjacoby126: If prepareIndexMutations throws an exception so that this.builder.removeRowStates never gets called, are we still ok, or does this need a finally block? Seems like the LocalTable cache of results would just fill and fill until we OOM. 

Alternately, the results cache could have an upper limit beyond which it auto-removes older items, but that's probably more complex.

phoenix-core/src/main/java/org/apache/phoenix/hbase/index/builder/IndexBuilder.java
```diff
@@ -151,4 +154,8 @@
   public List<Mutation> executeAtomicOp(Increment inc) throws IOException;
 
   public ReplayWrite getReplayWrite(Mutation m);
+
```
gjacoby126: nit: need a javadoc for both new methods

phoenix-core/src/main/java/org/apache/phoenix/hbase/index/covered/data/LocalHBaseState.java
```diff
@@ -46,4 +48,7 @@
   public Result getCurrentRowState(Mutation m, Collection<? extends ColumnReference> toCover, boolean ignoreNewerMutations)
       throws IOException;
 
+  public void scanCurrentRowStates(Set<ImmutableBytesPtr> rows, Collection<? extends ColumnReference> columns, long ts)
```
gjacoby126: nit: javadoc

phoenix-core/src/main/java/org/apache/phoenix/hbase/index/covered/data/LocalTable.java
```diff
@@ -49,15 +61,53 @@
...
+    private Map<ImmutableBytesPtr, Result> results = new HashMap<>();;
 
   public LocalTable(RegionCoprocessorEnvironment env) {
     this.env = env;
   }
 
+  public void scanCurrentRowStates(Set<ImmutableBytesPtr> rows, Collection<? extends ColumnReference> columns, long ts) throws IOException {
+      Scan s = IndexManagementUtil.newLocalStateScan(Collections.singletonList(columns));
+      FilterList filterList = new FilterList(FilterList.Operator.MUST_PASS_ONE);
+      for (ImmutableBytesPtr row : rows) {
```
gjacoby126: just fyi, In the past we've found perf issues with FilterLists with enormous numbers of filters (> 5000)

phoenix-core/src/main/java/org/apache/phoenix/hbase/index/covered/data/LocalTable.java
```diff
@@ -49,15 +61,53 @@
...
+          do {
+              List<Cell> kvs = new ArrayList<Cell>(1);
+              more = scanner.next(kvs);
+              if (kvs.isEmpty()) {
+                  return;
+              }
+              Result r = Result.create(kvs);
+              Cell cell = kvs.get(0);
+              byte[] rowKey = new byte[cell.getRowLength()];
+              System.arraycopy(cell.getRowArray(), cell.getRowOffset(), rowKey, 0, cell.getRowLength());
```
gjacoby126: Can replace with HBase's CellUtil.cloneRow method for this

phoenix-core/src/main/java/org/apache/phoenix/hbase/index/covered/data/LocalTable.java
```diff
@@ -49,15 +61,53 @@
...
+              results.put(new ImmutableBytesPtr(rowKey), r);
+          } while (more);
+      }
+  }
+
+  public void removeRowStates(Set<ImmutableBytesPtr> rows) {
+      for (ImmutableBytesPtr row : rows) {
+          results.remove(row);
```
gjacoby126: nit: looks like we can use results.removeAll(rows) here.

phoenix-core/src/main/java/org/apache/phoenix/hbase/index/covered/data/LocalTable.java
```diff
@@ -49,15 +61,53 @@
...
   @Override
   public Result getCurrentRowState(Mutation m, Collection<? extends ColumnReference> columns, boolean ignoreNewerMutations)
       throws IOException {
-    byte[] row = m.getRow();
+      byte[] row = m.getRow();
+      Result r = results.get(new ImmutableBytesPtr(row));
+      if (r !=null) {
+          return r;
+      }
+
```
gjacoby126: Is there actually a circumstance where the Mutation row we want isn't already in the cache? If so, this is good as is, if not, maybe either add logging or an exception rather than just leaving the old logic here as is.

PR #635, opened 2019-11-15
PHOENIX-5573 Ensure that all Maven dependencies converge

risdenk:
PR to show what this looks like. The `mvn verify` locally blew up immediately :)


stoty:
This cannot be done, as HBase itself does not converge.

PR #640, opened 2019-11-22
PHOENIX-5578 - CREATE TABLE IF NOT EXISTS loads IndexRegionObserver o…

gjacoby126:
…n an existing table

### DIFF
--- a/phoenix-core/src/it/java/org/apache/phoenix/end2end/index/IndexCoprocIT.java
+++ b/phoenix-core/src/it/java/org/apache/phoenix/end2end/index/IndexCoprocIT.java
@@ -62,7 +62,7 @@ public class IndexCoprocIT extends ParallelStatsDisabledIT {
 
     @Test
     public void testCreateCoprocs() throws Exception {
-        String schemaName = "S" + generateUniqueName();
+        String schemaName = "S_" + generateUniqueName();
         String tableName = "T_" + generateUniqueName();
         String indexName = "I_" + generateUniqueName();
         String physicalTableName = SchemaUtil.getPhysicalHBaseTableName(schemaName, tableName,
@@ -77,9 +77,24 @@ public class IndexCoprocIT extends ParallelStatsDisabledIT {
         HTableDescriptor baseDescriptor = admin.getTableDescriptor(TableName.valueOf(physicalTableName));
         HTableDescriptor indexDescriptor = admin.getTableDescriptor(TableName.valueOf(physicalIndexName));
 
-        assertCoprocsContains(IndexRegionObserver.class, baseDescriptor);
-        assertCoprocsContains(GlobalIndexChecker.class, indexDescriptor);
+        assertUsingNewCoprocs(baseDescriptor, indexDescriptor);
 
+        //Simulate an index upgrade rollback by removing coprocs and enabling old Indexer
+        downgradeIndexCoprocs(admin, baseDescriptor, indexDescriptor);
+
+        baseDescriptor = admin.getTableDescriptor(TableName.valueOf(physicalTableName));
+        indexDescriptor = admin.getTableDescriptor(TableName.valueOf(physicalIndexName));
+        assertUsingOldCoprocs(baseDescriptor, indexDescriptor);
+
+        //Now that we've downgraded, we make sure that a create statement won't re-upgrade us
+        createBaseTable(schemaName, tableName, true, 0, null);
+        baseDescriptor = admin.getTableDescriptor(TableName.valueOf(physicalTableName));
+        indexDescriptor = admin.getTableDescriptor(TableName.valueOf(physicalIndexName));
+        assertUsingOldCoprocs(baseDescriptor, indexDescriptor);
+    }
+
+    private void downgradeIndexCoprocs(Admin admin, HTableDescriptor baseDescriptor,
+                                       HTableDescriptor indexDescriptor) throws Exception {
         removeCoproc(IndexRegionObserver.class, baseDescriptor, admin);
         removeCoproc(IndexRegionObserver.class, indexDescriptor, admin);
         removeCoproc(GlobalIndexChecker.class, indexDescriptor, admin);
@@ -89,14 +104,6 @@ public class IndexCoprocIT extends ParallelStatsDisabledIT {
         Indexer.enableIndexing(baseDescriptor, NonTxIndexBuilder.class,
             props, 100);
         admin.modifyTable(baseDescriptor.getTableName(), baseDescriptor);
-        baseDescriptor = admin.getTableDescriptor(TableName.valueOf(physicalTableName));
-        indexDescriptor = admin.getTableDescriptor(TableName.valueOf(physicalIndexName));
-        assertUsingOldCoprocs(baseDescriptor, indexDescriptor);
-
-        createBaseTable(schemaName, tableName, true, 0, null);
-        baseDescriptor = admin.getTableDescriptor(TableName.valueOf(physicalTableName));
-        indexDescriptor = admin.getTableDescriptor(TableName.valueOf(physicalIndexName));
-        assertUsingOldCoprocs(baseDescriptor, indexDescriptor);
     }
 
     @Test
@@ -140,16 +147,14 @@ public class IndexCoprocIT extends ParallelStatsDisabledIT {
         HTableDescriptor baseDescriptor = admin.getTableDescriptor(TableName.valueOf(physicalTableName));
         HTableDescriptor indexDescriptor = admin.getTableDescriptor(TableName.valueOf(physicalIndexName));
 
-        assertCoprocsContains(IndexRegionObserver.class, baseDescriptor);
-        assertCoprocsContains(GlobalIndexChecker.class, indexDescriptor);
+        assertUsingNewCoprocs(baseDescriptor, indexDescriptor);
         String columnName = "foo";
         addColumnToBaseTable(schemaName, tableName, columnName);
-        assertCoprocsContains(IndexRegionObserver.class, baseDescriptor);
-        assertCoprocsContains(GlobalIndexChecker.class, indexDescriptor);
+        assertUsingNewCoprocs(baseDescriptor, indexDescriptor);
         dropColumnToBaseTable(schemaName, tableName, columnName);
-        assertCoprocsContains(IndexRegionObserver.class, baseDescriptor);
-        assertCoprocsContains(GlobalIndexChecker.class, indexDescriptor);
+        assertUsingNewCoprocs(baseDescriptor, indexDescriptor);
     }
+
     private void assertUsingOldCoprocs(HTableDescriptor baseDescriptor,
                                        HTableDescriptor indexDescriptor) {
         assertCoprocsContains(Indexer.class, baseDescriptor);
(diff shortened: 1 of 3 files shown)


phoenix-core/src/main/java/org/apache/phoenix/query/ConnectionQueryServicesImpl.java
```diff
@@ -999,6 +1007,23 @@ private void addCoprocessors(byte[] tableName, HTableDescriptor descriptor, PTab
...
+        }
+        boolean hasScanObserver = existingDesc.hasCoprocessor(ScanRegionObserver.class.getName());
+        boolean hasUnAggObserver = existingDesc.hasCoprocessor(
+            UngroupedAggregateRegionObserver.class.getName());
+        boolean hasGroupedObserver = existingDesc.hasCoprocessor(
+            GroupedAggregateRegionObserver.class.getName());
+        boolean hasIndexObserver = existingDesc.hasCoprocessor(Indexer.class.getName())
+            || existingDesc.hasCoprocessor(IndexRegionObserver.class.getName())
+            || existingDesc.hasCoprocessor(GlobalIndexChecker.class.getName());
+        return hasScanObserver && hasUnAggObserver && hasGroupedObserver && hasIndexObserver;
```
gokceni: Rather than checking coprocessors one by one, why don't we check that the coprocessors on the table length is greater than 0?

phoenix-core/src/it/java/org/apache/phoenix/end2end/index/CreateIndexCoprocIT.java
```diff
@@ -0,0 +1,204 @@
...
+public class CreateIndexCoprocIT extends ParallelStatsDisabledIT {
+    private boolean isNamespaceMapped = false;
+
+    @Test
+    public void testCreateCoprocs() throws Exception {
+        String schemaName = "S" + generateUniqueName();
+        String tableName = "T_" + generateUniqueName();
+        String indexName = "I_" + generateUniqueName();
+        createBaseTable(schemaName, tableName, true, 0, null);
+        createIndexTable(schemaName, tableName, indexName);
```
gokceni: Should we also add view and view index here as well?

phoenix-core/src/it/java/org/apache/phoenix/end2end/index/CreateIndexCoprocIT.java
```diff
@@ -0,0 +1,204 @@
...
+            assertUsingNewCoprocs(baseDescriptor);
+            createIndexTable(schemaName, tableName, indexName);
+            baseDescriptor = admin.getTableDescriptor(table);
+            TableName indexTable = TableName.valueOf(SchemaUtil.getPhysicalHBaseTableName(schemaName,
+                indexName, isNamespaceMapped).getString());
+            HTableDescriptor indexDescriptor = admin.getTableDescriptor(indexTable);
+            assertUsingNewCoprocs(baseDescriptor, indexDescriptor);
+        }
+    }
+
```
gokceni: Do you want to add another test for ALTER?

phoenix-core/src/it/java/org/apache/phoenix/end2end/index/CreateIndexCoprocIT.java
```diff
@@ -0,0 +1,204 @@
...
+        admin.modifyTable(baseDescriptor.getTableName(), baseDescriptor);
+        baseDescriptor = admin.getTableDescriptor(TableName.valueOf(physicalTableName));
+        indexDescriptor = admin.getTableDescriptor(TableName.valueOf(physicalIndexName));
+        assertUsingOldCoprocs(baseDescriptor, indexDescriptor);
+
+        createBaseTable(schemaName, tableName, true, 0, null);
+        baseDescriptor = admin.getTableDescriptor(TableName.valueOf(physicalTableName));
+        indexDescriptor = admin.getTableDescriptor(TableName.valueOf(physicalIndexName));
+        assertUsingOldCoprocs(baseDescriptor, indexDescriptor);
+
```
gokceni: nit: Remove unnecessary empty lines

phoenix-core/src/it/java/org/apache/phoenix/end2end/index/CreateIndexCoprocIT.java
```diff
@@ -0,0 +1,204 @@
...
+        throws SQLException {
+        Connection conn = getConnection();
+        String fullTableName = SchemaUtil.getTableName(schemaName, tableName);
+        conn.createStatement().execute("CREATE INDEX " + indexName + " ON " + fullTableName + "(v1)");
+    }
+
+    private void createBaseTable(String schemaName, String tableName, boolean multiTenant, Integer saltBuckets, String splits)
+        throws SQLException {
+        Connection conn = getConnection();
+        if (isNamespaceMapped) {
```
gokceni: This never seem to be set to true.

phoenix-core/src/main/java/org/apache/phoenix/query/ConnectionQueryServicesImpl.java
gjacoby126: Coprocs could exist for other, non-Phoenix reasons, or be incomplete because of some prior metadata corruption.

phoenix-core/src/it/java/org/apache/phoenix/end2end/index/CreateIndexCoprocIT.java
gjacoby126: That's already checked elsewhere (in PHOENIX-5560, though I apparently never backported it to 4.14, which I can do). Please let me know if between the two you still see a gap

phoenix-core/src/main/java/org/apache/phoenix/query/ConnectionQueryServicesImpl.java
```diff
@@ -2119,7 +2144,7 @@ private void setTransactional(HTableDescriptor tableDescriptor, PTableType table
         } else {
             tableDescriptor.setValue(PhoenixTransactionContext.READ_NON_TX_DATA, txValue);
         }
-        this.addCoprocessors(tableDescriptor.getName(), tableDescriptor, tableType, tableProps);
+        this.addCoprocessors(tableDescriptor.getName(), tableDescriptor, tableType, tableProps, tableDescriptor);
```
swaroopak: curious to know, if table is created for the first time, would tableDescriptor be empty or null?

phoenix-core/src/it/java/org/apache/phoenix/end2end/index/CreateIndexCoprocIT.java
```diff
@@ -0,0 +1,204 @@
...
+
+import java.sql.Connection;
+import java.sql.DriverManager;
+import java.sql.SQLException;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Properties;
+
+public class CreateIndexCoprocIT extends ParallelStatsDisabledIT {
+    private boolean isNamespaceMapped = false;
```
swaroopak: did you intend to make this IT parameterized? Also, I would consider adding one test for the Multitenant table.

phoenix-core/src/main/java/org/apache/phoenix/query/ConnectionQueryServicesImpl.java
priyankporwal: If any of these coprocs are missing, we would still end-up adding the indexing coprocs, possibly under the similar rolling upgrade time as we have recently experienced. Should we confirm if 'CREATE .. IF NOT EXISTS' is supposed to so such 'repairs' or are there other better mechanisms?

phoenix-core/src/it/java/org/apache/phoenix/end2end/index/IndexCoprocIT.java
```diff
@@ -0,0 +1,259 @@
...
+        this.isMultiTenant = isMultiTenant;
+    }
+    @Parameterized.Parameters(name ="CreateIndexCoprocIT_mulitTenant={0}")
+    public static Collection<Object[]> data() {
+        return Arrays.asList(new Object[][]{{true}, {false}});
+    }
+
+    @Test
+    public void testCreateCoprocs() throws Exception {
+        String schemaName = "S" + generateUniqueName();
```
gokceni: nit: S_

phoenix-core/src/it/java/org/apache/phoenix/end2end/index/IndexCoprocIT.java
```diff
@@ -0,0 +1,259 @@
...
+    private void assertCoprocsNotContains(Class clazz, HTableDescriptor descriptor) {
+        String expectedCoprocName = clazz.getName();
+        boolean foundCoproc = isCoprocPresent(descriptor, expectedCoprocName);
+        Assert.assertFalse("Could find coproc " + expectedCoprocName +
+            " in descriptor " + descriptor,foundCoproc);
+    }
+
+    private boolean isCoprocPresent(HTableDescriptor descriptor, String expectedCoprocName) {
+        boolean foundCoproc = false;
+        for (String coprocName : descriptor.getCoprocessors()){
```
gokceni: Doesn't descriptor have hasCoprocessor method? No need the for stmt

phoenix-core/src/it/java/org/apache/phoenix/end2end/index/IndexCoprocIT.java
```diff
@@ -0,0 +1,259 @@
...
+        Admin admin = ((PhoenixConnection) getConnection()).getQueryServices().getAdmin();
+
+        createBaseTable(schemaName, tableName, isMultiTenant, 0, null);
+        createIndexTable(schemaName, tableName, indexName);
+
+        HTableDescriptor baseDescriptor = admin.getTableDescriptor(TableName.valueOf(physicalTableName));
+        HTableDescriptor indexDescriptor = admin.getTableDescriptor(TableName.valueOf(physicalIndexName));
+
+        assertCoprocsContains(IndexRegionObserver.class, baseDescriptor);
+        assertCoprocsContains(GlobalIndexChecker.class, indexDescriptor);
```
gokceni: how about assertNewCoprocessors?

phoenix-core/src/it/java/org/apache/phoenix/end2end/index/IndexCoprocIT.java
```diff
@@ -0,0 +1,259 @@
...
+        removeCoproc(IndexRegionObserver.class, baseDescriptor, admin);
+        removeCoproc(IndexRegionObserver.class, indexDescriptor, admin);
+        removeCoproc(GlobalIndexChecker.class, indexDescriptor, admin);
+
+        Map<String, String> props = new HashMap<String, String>();
+        props.put(NonTxIndexBuilder.CODEC_CLASS_NAME_KEY, PhoenixIndexCodec.class.getName());
+        Indexer.enableIndexing(baseDescriptor, NonTxIndexBuilder.class,
+            props, 100);
+        admin.modifyTable(baseDescriptor.getTableName(), baseDescriptor);
+        baseDescriptor = admin.getTableDescriptor(TableName.valueOf(physicalTableName));
```
gokceni: Is this part to make sure we have the old coprocs? In that case, why not disable new index design, create tables, check old coprocs, create again check old coprocs again?
It is a bit hard to understand like this

gjacoby126: @gokceni - extracted method and added some comments to make it a bit more clear. I'm "downgrading" the existing table to make sure that no matter what happens the create statement won't "re-upgrade" a table again. Creating a separate table wouldn't prove quite the same thing.

phoenix-core/src/it/java/org/apache/phoenix/end2end/index/CreateIndexCoprocIT.java
gjacoby126: Done.

phoenix-core/src/it/java/org/apache/phoenix/end2end/index/CreateIndexCoprocIT.java
gjacoby126: Done

phoenix-core/src/it/java/org/apache/phoenix/end2end/index/IndexCoprocIT.java
gjacoby126: Done

phoenix-core/src/it/java/org/apache/phoenix/end2end/index/IndexCoprocIT.java
gjacoby126: Good catch. Switched to use the existing descriptor method.

phoenix-core/src/it/java/org/apache/phoenix/end2end/index/IndexCoprocIT.java
gjacoby126: Done.

phoenix-core/src/main/java/org/apache/phoenix/query/ConnectionQueryServicesImpl.java
gjacoby126: @priyankporwal - the tricky thing is that it's legal to take a raw HBase table, run a "CREATE TABLE" statement, and expect it to get the coprocs. Proving that that's not what's happening (or is in the middle of happening on some other transaction) is hard, so I'm being conservative + not trying to solve the general problem of making an existent table not touched (which I agree is the ideal behavior.)

phoenix-core/src/it/java/org/apache/phoenix/end2end/index/CreateIndexCoprocIT.java
gokceni: @gjacoby126 I am wondering about this variable. Other than this, LGTM. But why do we have this if we don't set it?

PR #641, opened 2019-11-26
PHOENIX-5589 GlobalIndexChecker does not populate global index relate…

kadirozde:
…d attributes for certain queries

### DIFF
--- a/phoenix-core/src/main/java/org/apache/phoenix/util/IndexUtil.java
+++ b/phoenix-core/src/main/java/org/apache/phoenix/util/IndexUtil.java
@@ -958,6 +958,13 @@ public class IndexUtil {
             // This index table must be being deleted. No need to set the scan attributes
             return;
         }
+        // MetaDataClient modifies the index table name for view indexes if the parent view of an index has a child
+        // view. This, we need to recreate a PTable object with the correct table name for the rest of this code to work
+        if (indexTable.getViewIndexId() != null && indexTable.getName().getString().contains(QueryConstants.CHILD_VIEW_INDEX_NAME_SEPARATOR)) {
+            int lastIndexOf = indexTable.getName().getString().lastIndexOf(QueryConstants.CHILD_VIEW_INDEX_NAME_SEPARATOR);
+            String indexName = indexTable.getName().getString().substring(lastIndexOf + 1);
+            indexTable = PhoenixRuntime.getTable(phoenixConnection, indexName);
+        }
         if (!dataTable.getIndexes().contains(indexTable)) {
             return;
         }
--- a/phoenix-core/src/it/java/org/apache/phoenix/end2end/index/GlobalIndexCheckerIT.java
+++ b/phoenix-core/src/it/java/org/apache/phoenix/end2end/index/GlobalIndexCheckerIT.java
@@ -30,11 +30,14 @@ import java.sql.SQLException;
 import java.util.Collection;
 import java.util.List;
 import java.util.Map;
+import java.util.Properties;
 
 import com.google.common.collect.Maps;
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
 import org.apache.hadoop.hbase.CellScanner;
+import org.apache.hadoop.hbase.client.HBaseAdmin;
+import org.apache.hadoop.hbase.client.HTable;
 import org.apache.hadoop.hbase.client.Result;
 import org.apache.hadoop.hbase.client.ResultScanner;
 import org.apache.hadoop.hbase.client.Scan;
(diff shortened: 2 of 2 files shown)


phoenix-core/src/main/java/org/apache/phoenix/util/IndexUtil.java
```diff
@@ -958,6 +958,13 @@ public static void setScanAttributesForIndexReadRepair(Scan scan, PTable table,
             // This index table must be being deleted. No need to set the scan attributes
             return;
         }
+        // MetaDataClient modifies the index table name for view indexes if the parent view of an index has a child
+        // view. This, we need to recreate a PTable object with the correct table name for the rest of this code to work
+        if (table.getName().getString().contains(QueryConstants.CHILD_VIEW_INDEX_NAME_SEPARATOR)) {
```
gokceni: Should we add a check here that table is a view?
Can a user create a table with # in the name?

phoenix-core/src/main/java/org/apache/phoenix/util/IndexUtil.java
```diff
@@ -958,6 +958,13 @@ public static void setScanAttributesForIndexReadRepair(Scan scan, PTable table,
...
+            int lastIndexOf = table.getName().getString().lastIndexOf(QueryConstants.CHILD_VIEW_INDEX_NAME_SEPARATOR);
+            String indexName = table.getName().getString().substring(lastIndexOf + 1);
```
gokceni: Recommend to check that lastIndexOf < tableName length so that we don't get IndexOutOfBounds exception

kadirozde: We better get exception if the name is malformed.

phoenix-core/src/main/java/org/apache/phoenix/util/IndexUtil.java
kadirozde: I do not think it is allowed. Can you check it?

ChinmaySKulkarni: This should check if table is view index..maybe you can use 'table.getViewIndexId() != null`?
I believe we disallow using '#' in user-defined names..you get:

```
java.lang.IllegalArgumentException: Illegal character code:35, <#> at 3. User-space table qualifiers can only contain 'alphanumeric characters': i.e. [a-zA-Z_0-9-.]:
```

gokceni: Just checked, yes it is not allowed for tables.

gjacoby126: It would be clearer if you used indexTable rather than table (to be consistent with the logic earlier in the method)

gjacoby126: Agree it would be better to not rely on a naming convention if possible to tell if it's a view index.

phoenix-core/src/main/java/org/apache/phoenix/util/IndexUtil.java
```diff
@@ -958,6 +958,13 @@ public static void setScanAttributesForIndexReadRepair(Scan scan, PTable table,
             // This index table must be being deleted. No need to set the scan attributes
             return;
         }
+        // MetaDataClient modifies the index table name for view indexes if the parent view of an index has a child
+        // view. This, we need to recreate a PTable object with the correct table name for the rest of this code to work
+        if (table.getName().getString().contains(QueryConstants.CHILD_VIEW_INDEX_NAME_SEPARATOR)) {
+            int lastIndexOf = table.getName().getString().lastIndexOf(QueryConstants.CHILD_VIEW_INDEX_NAME_SEPARATOR);
```
gjacoby126: Should this logic be encapsulated somewhere? Is this the only place where we need to get the index name from a view index?

phoenix-core/src/main/java/org/apache/phoenix/util/IndexUtil.java
kadirozde: Ok, I will add this extra check.

kadirozde: Done

phoenix-core/src/main/java/org/apache/phoenix/util/IndexUtil.java
kadirozde: This is a workaround. I think we need to fix this issue in the Phoenix core later. I do not want to give the impression that it is a final fix.

phoenix-core/src/main/java/org/apache/phoenix/util/IndexUtil.java
kadirozde: Done

ChinmaySKulkarni:
lgtm. Thanks for changing the test @kadirozde!

PR #642, opened 2019-11-26
PHOENIX-5590 IndexTool always runs with direct

gokceni:


phoenix-core/src/it/java/org/apache/phoenix/end2end/IndexToolIT.java
```diff
@@ -697,20 +691,19 @@ private static void verifyMapper(Job job, boolean directApi, boolean useSnapshot
         }
         try (Connection conn =
                      DriverManager.getConnection(getUrl(), props)) {
-            PTable dataTable = PhoenixRuntime.getTable(conn, SchemaUtil.getTableName(schemaName, dataTableName));
-            PTable indexTable = PhoenixRuntime.getTable(conn, SchemaUtil.getTableName(schemaName, indexTableName));
+            PTable
+                    dataTable =
```
gjacoby126: nit: can we fix the line breaks here?

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/index/IndexTool.java
```diff
@@ -147,7 +146,7 @@
             "To build indexes for a data table from least disabledTimeStamp");
     
     private static final Option DIRECT_API_OPTION = new Option("direct", "direct", false,
-            "If specified, we avoid the bulk load (optional)");
+            "This option is no longer a parameter. Direct is mandatory. Keeping as parameter due to backwards compatibility (optional)");
```
gjacoby126: How about "This parameter is deprecated. Direct mode will be used whether it is set or not. Keeping it for backwards compatibility."

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/index/IndexTool.java
```diff
@@ -234,7 +233,7 @@ private CommandLine parseOptions(String[] args) {
                     + "parameter");
         }
 
-		if (!(cmdLine.hasOption(PARTIAL_REBUILD_OPTION.getOpt()) || cmdLine.hasOption(DIRECT_API_OPTION.getOpt()))
+		if (!cmdLine.hasOption(PARTIAL_REBUILD_OPTION.getOpt())
```
gjacoby126: This looks incorrect -- output path shouldn't ever be required now.

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/index/IndexTool.java
```diff
@@ -248,12 +247,13 @@ private CommandLine parseOptions(String[] args) {
                     + PARTIAL_REBUILD_OPTION.getLongOpt());
         }
         		
-        if (!(cmdLine.hasOption(DIRECT_API_OPTION.getOpt())) && cmdLine.hasOption(INDEX_TABLE_OPTION.getOpt())
+        if (!cmdLine.hasOption(INDEX_TABLE_OPTION.getOpt())
```
gjacoby126: This also looks incorrect. Is this check needed anymore?

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/index/IndexTool.java
```diff
@@ -429,10 +429,7 @@ private long getMaxRebuildAsyncDate(String schemaName, List<String> disableIndex
             
         }
 
-        private Job configureJobForAysncIndex()
-
-                throws Exception {
-
+        private Job configureJobForAysncIndex() throws Exception {
```
gjacoby126: nit: please fix spelling error in "async". (yes I know it was already there but it's been bugging me for months! :-) )

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/index/IndexTool.java
gokceni: If there is no index table parameter supplied, what will runfg do? I think this check is needed.

phoenix-core/src/it/java/org/apache/phoenix/end2end/IndexToolIT.java
priyankporwal: I liked the old way of these lines even if they were a bit longer, were more readable.

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/index/IndexTool.java
gjacoby126: Shouldn't the check be that if there's no index table set and no partial build set, then something is wrong? 

What prevents a partial build's job from being run in the foreground (other than that the logic currently has an isDirect flag, which might also need to be removed?)

gokceni: In the original code, there is no PARTIAL_REBUILD mentioned in this if block.
I just did conversion of previous check like this: !(cmdLine.hasOption(DIRECT_API_OPTION.getOpt()) && cmdLine.hasOption(INDEX_TABLE_OPTION.getOpt()) ==> !(true && cmdLine.hasOption(INDEX_TABLE_OPTION.getOpt()) ==> !cmdLine.hasOption(INDEX_TABLE_OPTION.getOpt()

phoenix-core/src/it/java/org/apache/phoenix/end2end/IndexToolIT.java
```diff
@@ -658,10 +653,9 @@ public static void assertExplainPlan(boolean localIndex, String actualExplainPla
         args.add(indxTable);
         if (directApi) {
```
priyankporwal: Remove if (directApi) block completely?

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/index/IndexTool.java
priyankporwal: @gjacoby126 Are you asking for more validations to be added here? Gokcen's changes seem to be simply replacing cmdLine.hasOption(INDEX_TABLE_OPTION.getOpt() with true.

gjacoby126: @priyankporwal -  Keep in mind that just because we're retiring the -direct flag, not _all_ runs will go through the direct paths now, because partial build exists. Collapsing it down to "true" isn't _quite_ right, and the check as it existed was already kind of weird. 

Most index tool invocations need to specify an index, except for partial build, which apparently will look up all indexes of a base table. Leaving out the index name and not invoking partial build should throw an error. Foreground or not shouldn't affect it. 

There used to be a check much later that only allowed for running in the foreground if -direct was selected, but I don't see a reason you couldn't run the partial build job in the foreground.

gjacoby126: We should probably also look to see if the partial build path still makes sense, and retire it if not, but that can potentially be a separate JIRA.

gokceni: @gjacoby126 Removed the check, thanks for explanation. Can you please re-review?

gjacoby126: @gokceni - don't we still need a check for "no index table set, no partial build set"? Other than that I think is ready to go.

PR #652, opened 2019-12-12
PHOENIX-5615 Index read repair should delete all the cells of an inva…

kadirozde:
…lid unverified row

### DIFF
--- a/phoenix-core/src/main/java/org/apache/phoenix/index/GlobalIndexChecker.java
+++ b/phoenix-core/src/main/java/org/apache/phoenix/index/GlobalIndexChecker.java
@@ -113,10 +113,12 @@ public class GlobalIndexChecker implements RegionCoprocessor, RegionObserver {
      * and used to verify individual rows and rebuild them if they are not valid
      */
     private class GlobalIndexScanner implements RegionScanner {
-        RegionScanner scanner;
+        private RegionScanner scanner;
+        private RegionScanner deleteRowScanner;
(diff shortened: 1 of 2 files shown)


phoenix-core/src/main/java/org/apache/phoenix/index/GlobalIndexChecker.java
```diff
@@ -245,6 +247,14 @@ private void deleteRowIfAgedEnough(byte[] indexRowKey, List<Cell> row, long ts,
             if ((EnvironmentEdgeManager.currentTimeMillis() - ts) > ageThreshold) {
                 Delete del = new Delete(indexRowKey, ts);
                 if (specific) {
+                    // Get all the cells of this row
+                    deleteRowScan.withStartRow(indexRowKey, true);
```
gjacoby126: hmm...if you knew for sure what the CFs were you could probbaly do addFamilyVersion in a loop instead, possibly saving yourself a Scan. You wouldn't have to worry about dynamic columns (can't be indexed), but you'd need a PTable for getColumnFamilies, and to know for sure it was up to date as of the timestamp you're deleting.

I'm going to approve as-is, just thinking out loud here.

kadirozde: It is safer to retrieve the columns to find out their timestamps and delete markers for these timestamps, at least for now. We can think about your suggestion in future.

PR #657, opened 2019-12-17
PHOENIX-5591 Fail indextool if split-merge happens during job run

gokceni:


phoenix-core/src/main/java/org/apache/phoenix/mapreduce/index/PhoenixIndexImportDirectReducer.java
```diff
@@ -49,8 +50,42 @@
...
+
+    private static Integer regionCountForTesting = null;
+
+    public static void setRegionCountForTesting(Integer regionCount) {
+        regionCountForTesting = regionCount;
+    }
+
     @Override
     protected void cleanup(Context context) throws IOException, InterruptedException{
+        String strNumOfRegions = PhoenixConfigurationUtil.getNumOfRegions(context.getConfiguration());
```
priyankporwal: Isn't this the number of mappers? Can the reducer get it instead and compare with it?

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/index/PhoenixIndexImportDirectReducer.java
```diff
@@ -49,8 +50,42 @@
     private static final Logger LOGGER =
             LoggerFactory.getLogger(PhoenixIndexImportDirectReducer.class);
 
+    private static final String SPLIT_MERGE_ERROR_STR_FMT =
```
priyankporwal: This solution tackles Split-Only case. With concurrent merges, this is not a guarantee that region counts would differ.

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/index/PhoenixIndexImportDirectReducer.java
gokceni: It is not mappers. It is table region count. Not sure I understand what you are suggesting

gokceni:
This Jira is not needed. It is fixed in IndexRegionObserver with sending LastCell to Hbase client.

PR #658, opened 2019-12-18
PHOENIX-5632: Use 'phoenix.default.update.cache.frequency' from conne…

ntshmah:
…ction properties at query time

The effective update cache frequency value for a table should be evaluated based on the following precedence order:
Table level > Connection level > Default.


ntshmah:
I accidentally created this Pull Request, kindly ignore this.

PR #659, opened 2019-12-18
PHOENIX-5634: Use 'phoenix.default.update.cache.frequency' from conne…

ntshmah:
…ction properties at query time

### DIFF
--- a/phoenix-core/src/it/java/org/apache/phoenix/rpc/UpdateCacheConnectionLevelPropIT.java
+++ b/phoenix-core/src/it/java/org/apache/phoenix/rpc/UpdateCacheConnectionLevelPropIT.java
@@ -1,0 +1,239 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.phoenix.rpc;
+
+import org.apache.phoenix.end2end.ParallelStatsDisabledIT;
+import org.apache.phoenix.query.ConnectionQueryServices;
+import org.apache.phoenix.query.QueryConstants;
+import org.apache.phoenix.query.QueryServices;
+import org.apache.phoenix.schema.PName;
+import org.apache.phoenix.schema.types.PVarchar;
+import org.apache.phoenix.util.SchemaUtil;
+import static org.apache.phoenix.util.TestUtil.DEFAULT_SCHEMA_NAME;
+
+import org.junit.AfterClass;
+import org.junit.Test;
+import static org.junit.Assert.assertTrue;
+
+import java.sql.Connection;
+import java.sql.DriverManager;
+import java.sql.ResultSet;
+import java.sql.SQLException;
+import java.sql.Statement;
+import java.util.Properties;
+
+import org.mockito.Mockito;
+import static org.mockito.Matchers.anyLong;
+import static org.mockito.Matchers.eq;
+import static org.mockito.Mockito.isNull;
+import static org.mockito.Mockito.reset;
+import static org.mockito.Mockito.times;
+import static org.mockito.Mockito.verify;
+
+/**
+ * Tests for connection level 'Update Cache Frequency' property.
+ *
+ * The tests verify that the 'Update Cache Frequency' is honored in the following precedence order
+ * for SELECTs and UPSERTs:
+ * Table-level property > Connection-level property > Default value
+ */
+public class UpdateCacheConnectionLevelPropIT extends ParallelStatsDisabledIT {
+
+    private static Connection conn1;
+    private static Connection conn2;
+    private static ConnectionQueryServices spyForConn2;
+
+    @AfterClass
+    public static synchronized void freeResources() {
+        try {
+            conn1.close();
+            conn2.close();
+        } catch (Exception e) {
+            /* ignored */
+        }
+    }
+
+    /**
+     * Test 'Update Cache Frequency' property when it is set at connection-level only, and not at
+     * table-level.
+     */
+    @Test
+    public void testWithConnLevelUCFNoTableLevelUCF() throws Exception {
+        final long connUpdateCacheFrequency = 1000;
+        String fullTableName = DEFAULT_SCHEMA_NAME + QueryConstants.NAME_SEPARATOR +
+                generateUniqueName();
+
+        setUpTableAndConnections(fullTableName, null,
+                String.valueOf(connUpdateCacheFrequency));
+
+        // There should only be a single call to getTable() for fetching the table's metadata
+        int numExecutions = 2;
+        int numExpectedGetTableCalls = 1;
+        verifyExpectedGetTableCalls(fullTableName, numExecutions, numExpectedGetTableCalls);
+
+        // Wait for a period of 'connUpdateCacheFrequency' and verify that there was one new call to
+        // getTable() for fetching the table's metadata
+        Thread.sleep(connUpdateCacheFrequency);
+        verifyExpectedGetTableCalls(fullTableName, numExecutions, numExpectedGetTableCalls);
+    }
+
+    /**
+     * Test 'Update Cache Frequency' property when it is set at table-level only, and not at
+     * connection-level.
+     */
+    @Test
+    public void testWithTableLevelUCFNoConnLevelUCF() throws Exception {
+        final long tableUpdateCacheFrequency = 1000;
+        String fullTableName = DEFAULT_SCHEMA_NAME + QueryConstants.NAME_SEPARATOR +
+                generateUniqueName();
+
+        // There should only be a single call to getTable() for fetching the table's metadata
+        int numExecutions = 2;
+        int numExpectedGetTableCalls = 1;
+        setUpTableAndConnections(fullTableName, String.valueOf(tableUpdateCacheFrequency), null);
+        verifyExpectedGetTableCalls(fullTableName, numExecutions, numExpectedGetTableCalls);
+
+        // Wait for a period of 'tableUpdateCacheFrequency' and verify that there was one new call
+        // to getTable() for fetching the table's metadata
+        Thread.sleep(tableUpdateCacheFrequency);
+        verifyExpectedGetTableCalls(fullTableName, numExecutions, numExpectedGetTableCalls);
+    }
+
+    /**
+     * Test 'Update Cache Frequency' property when it is not set at both table-level and
+     * connection-level.
+     */
+    @Test
+    public void testWithNoConnAndTableLevelUCF() throws Exception {
+        String fullTableName = DEFAULT_SCHEMA_NAME + QueryConstants.NAME_SEPARATOR +
+                generateUniqueName();
+
+        // This is the default behavior (i.e. always fetch the latest metadata of the table) when
+        // both connection and table level properties are not set
+        int numExecutions = 2;
+        int numExpectedGetTableCalls = 4; // 2 for SELECTs, and 2 for UPSERTs
+        setUpTableAndConnections(fullTableName, null, null);
+        verifyExpectedGetTableCalls(fullTableName, numExecutions, numExpectedGetTableCalls);
+    }
+
+    /**
+     * Test 'Update Cache Frequency' property when it is set at both table-level and
+     * connection-level.
+     */
+    @Test
+    public void testWithBothConnAndTableLevelUCF() throws Exception {
+        // Set table level property to a much higher value than the connection level property
+        final long tableUpdateCacheFrequency = 5000;
+        final long connUpdateCacheFrequency = 1000;
+        String fullTableName = DEFAULT_SCHEMA_NAME + QueryConstants.NAME_SEPARATOR +
+                generateUniqueName();
+
+        // There should only be a single call to getTable() for fetching the table's metadata
+        int numExecutions = 2;
+        int numExpectedGetTableCalls = 1;
+        setUpTableAndConnections(fullTableName, String.valueOf(tableUpdateCacheFrequency),
+                String.valueOf(connUpdateCacheFrequency));
+        verifyExpectedGetTableCalls(fullTableName, numExecutions, numExpectedGetTableCalls);
+
+        // Wait for a period of 'connUpdateCacheFrequency' and verify that there were no new calls
+        // to getTable() as the table level UCF should come in to effect
+        Thread.sleep(connUpdateCacheFrequency);
+        numExpectedGetTableCalls = 0;
+        verifyExpectedGetTableCalls(fullTableName, numExecutions, numExpectedGetTableCalls);
+
+        // Extend the wait to a period of 'tableUpdateCacheFrequency' and verify that there was one
+        // new call to getTable() for fetching the table's metadata
+        Thread.sleep(tableUpdateCacheFrequency - connUpdateCacheFrequency);
+        numExpectedGetTableCalls = 1;
+        verifyExpectedGetTableCalls(fullTableName, numExecutions, numExpectedGetTableCalls);
+    }
+
+    /**
+     * Helper method that sets up the connections and creates the table to be tested.
+     * @param fullTableName The table's full name
+     * @param tableUpdateCacheFrequency If not null, the table-level value to be set for 'Update
+     *                                  Cache Frequency'
+     * @param connUpdateCacheFrequency If not null, the connection-level value to be set for 'Update
+     *                                 Cache Frequency'
+     */
+    private static void setUpTableAndConnections(String fullTableName,
+            String tableUpdateCacheFrequency, String connUpdateCacheFrequency) throws SQLException {
+        // Create two connections - a connection that we'll use to create the table and the second
+        // one that we will spy on and use to query the table.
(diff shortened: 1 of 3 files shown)


phoenix-core/src/it/java/org/apache/phoenix/rpc/UpdateCacheConnectionLevelPropIT.java
```diff
@@ -0,0 +1,173 @@
...
+
+import java.sql.Connection;
+import java.sql.DriverManager;
+import java.sql.ResultSet;
+import java.sql.SQLException;
+import java.util.Properties;
+
+import static org.apache.phoenix.util.TestUtil.DEFAULT_SCHEMA_NAME;
+import static org.junit.Assert.assertTrue;
+import static org.mockito.Matchers.*;
```
ChinmaySKulkarni: Avoid import .* Checkout https://phoenix.apache.org/contributing.html for an IDE plugin to format your code as per Phoenix coding conventions.

phoenix-core/src/it/java/org/apache/phoenix/rpc/UpdateCacheConnectionLevelPropIT.java
```diff
@@ -0,0 +1,173 @@
...
+import static org.mockito.Matchers.anyLong;
+import static org.mockito.Mockito.*;
+
+public class UpdateCacheConnectionLevelPropIT extends ParallelStatsDisabledIT {
```
ChinmaySKulkarni: Can we add these tests to the existing `UpdateCacheIT` class?

phoenix-core/src/it/java/org/apache/phoenix/rpc/UpdateCacheConnectionLevelPropIT.java
```diff
@@ -0,0 +1,173 @@
...
+        verifyExpectedGetTableCalls(fullTableName, numSelectExecutions, numExpectedGetTableCalls);
+
+        // Extend the wait to a period of 'tableUpdateCacheFrequency' and verify that there was one new call to
+        // getTable() for fetching the table's metadata
+        Thread.sleep(tableUpdateCacheFrequency - connUpdateCacheFrequency);
+        numExpectedGetTableCalls = 1;
+        verifyExpectedGetTableCalls(fullTableName, numSelectExecutions, numExpectedGetTableCalls);
+    }
+
+    // Helper method that sets up the connections and creates the table
```
ChinmaySKulkarni: nit: Use Javadoc comments for method headers

phoenix-core/src/it/java/org/apache/phoenix/rpc/UpdateCacheConnectionLevelPropIT.java
```diff
@@ -0,0 +1,173 @@
...
+    private static void setUpTableAndConnections(String fullTableName, long tableUpdateCacheFrequency, long connUpdateCacheFrequency) throws SQLException {
+        // Create two connections - a connection that we'll use to create the table and the second one that we will
+        // spy on and will have 'phoenix.default.update.cache.frequency' set
+        Properties props = new Properties();
+        conn1 = DriverManager.getConnection(getUrl(), props);
+        conn1.setAutoCommit(true);
+
+        Properties propsWithUCF = new Properties();
+        //if (connUpdateCacheFrequency != -1) {
```
ChinmaySKulkarni: Remove commented code

phoenix-core/src/it/java/org/apache/phoenix/rpc/UpdateCacheConnectionLevelPropIT.java
```diff
@@ -0,0 +1,173 @@
...
+            propsWithUCF.put(QueryServices.DEFAULT_UPDATE_CACHE_FREQUENCY_ATRRIB, "" + connUpdateCacheFrequency);
```
ChinmaySKulkarni: nit: fix indentation

ChinmaySKulkarni: Also instead of "" + long, can we do `String.valueOf(long)`? Makes it more explicit

phoenix-core/src/it/java/org/apache/phoenix/rpc/UpdateCacheConnectionLevelPropIT.java
ChinmaySKulkarni: Or rather, I think you want to uncomment this, no? -1 indicates the value is not set? Can we add a comment for that here?

phoenix-core/src/it/java/org/apache/phoenix/rpc/UpdateCacheConnectionLevelPropIT.java
```diff
@@ -0,0 +1,173 @@
...
+
+        // use a spied ConnectionQueryServices so we can verify calls to getTable()
+        spyForConn2 = Mockito.spy(driver.getConnectionQueryServices(getUrl(), connUpdateCacheFrequency != -1 ? propsWithUCF : props));
+        conn2 = spyForConn2.connect(getUrl(), propsWithUCF);
```
ChinmaySKulkarni: Here we pass in `propsWithUCF` even if `connUpdateCacheFrequency` is -1? Is that intended?

phoenix-core/src/it/java/org/apache/phoenix/rpc/UpdateCacheConnectionLevelPropIT.java
```diff
@@ -0,0 +1,173 @@
...
+        }
+
+        // Create the table over first connection
+        conn1.createStatement().execute(createTableQuery);
+        conn1.createStatement().execute("UPSERT INTO " + fullTableName + " VALUES (1, 2, 3, 4)");
+        conn1.commit();
+    }
+
+    // Helper method that executes a select query on the table for numSelectExecutions times and verifies that
+    // numExpectedGetTableCalls were made to getTable for the table
```
ChinmaySKulkarni: Same nit about Javadoc comments here.

phoenix-core/src/it/java/org/apache/phoenix/rpc/UpdateCacheConnectionLevelPropIT.java
```diff
@@ -0,0 +1,173 @@
...
+
+        // There should only be a single call to getTable() for fetching the table's metadata
+        int numSelectExecutions = 2;
+        int numExpectedGetTableCalls = 1;
+        verifyExpectedGetTableCalls(fullTableName, numSelectExecutions, numExpectedGetTableCalls);
+
+        // Wait for a period of 'connUpdateCacheFrequency' and verify that there was one new call to getTable() for
+        // fetching the table's metadata
+        Thread.sleep(connUpdateCacheFrequency);
+        verifyExpectedGetTableCalls(fullTableName, numSelectExecutions, numExpectedGetTableCalls);
```
ChinmaySKulkarni: We're still passing in `numExpectedGetTableCalls` instead of `numExpectedGetTableCalls + 1` (as per comments). Not sure if that's the intention.

phoenix-core/src/main/java/org/apache/phoenix/schema/MetaDataClient.java
```diff
@@ -722,14 +722,27 @@ public MetaDataMutationResult updateCache(PName origTenantId, String schemaName,
     // 1. table is a system table that does not have a ROW_TIMESTAMP column OR
     // 2. table was already resolved as of that timestamp OR
     // 3. table does not have a ROW_TIMESTAMP column and age is less then UPDATE_CACHE_FREQUENCY
+    // 3a. Get the effective UPDATE_CACHE_FREQUENCY for checking the age in the following precedence order:
+    // Table-level property > Connection-level property > Default value.
```
ChinmaySKulkarni: Please create a follow-up JIRA for this documentation change, particularly about the precedence order. Maybe we can add a couple of lines here https://phoenix.apache.org/language/index.html#options

phoenix-core/src/main/java/org/apache/phoenix/schema/MetaDataClient.java
```diff
@@ -722,14 +722,27 @@ public MetaDataMutationResult updateCache(PName origTenantId, String schemaName,
...
+            if (systemTable && table.getRowTimestampColPos() == -1 || resolvedTimestamp == tableResolvedTimestamp) {
+                return true;
+            }
+
+            // What if the table is created with UPDATE_CACHE_FREQUENCY explicitly set to ALWAYS (= 0)?
+            // We should ideally be checking something like hasUpdateCacheFrequency()
+            long effectiveUpdateCacheFrequency = table.getUpdateCacheFrequency() != 0L ?
+                    table.getUpdateCacheFrequency() :
+                    connection.getQueryServices().getProps().getLong(
+                            QueryServices.DEFAULT_UPDATE_CACHE_FREQUENCY_ATRRIB, QueryServicesOptions.DEFAULT_UPDATE_CACHE_FREQUENCY);
```
ChinmaySKulkarni: Maybe add a debug log line here saying that we are using the table-level property vs the connection-level property and its value and the table tenantId + table schema + table name.

phoenix-core/src/main/java/org/apache/phoenix/schema/MetaDataClient.java
```diff
@@ -722,14 +722,27 @@ public MetaDataMutationResult updateCache(PName origTenantId, String schemaName,
...
+
+            return (table.getRowTimestampColPos() == -1 &&
+                    connection.getMetaDataCache().getAge(tableRef) <
+                            effectiveUpdateCacheFrequency);
+        }
+        return false;
```
ChinmaySKulkarni: Are there any more places where we may not be honoring the UCF property value before retrieving the schema of a table? Does this change handle all query paths, upsert path, etc.? I guess for DDL operations, we force the client to get the latest metadata so we don't need to make any changes there. Perhaps we can add more tests to reflect all these different code paths.

phoenix-core/src/it/java/org/apache/phoenix/rpc/UpdateCacheConnectionLevelPropIT.java
ntshmah: Done.

phoenix-core/src/it/java/org/apache/phoenix/rpc/UpdateCacheConnectionLevelPropIT.java
ntshmah: I could, but only if you have a strong preference for that. I would prefer to keep these tests in a separate class as I have some private methods and members which won't have a purpose for the methods in `UpdateCacheIT` class.

phoenix-core/src/it/java/org/apache/phoenix/rpc/UpdateCacheConnectionLevelPropIT.java
ntshmah: Noted, will update the headers.

phoenix-core/src/it/java/org/apache/phoenix/rpc/UpdateCacheConnectionLevelPropIT.java
ntshmah: Right, I've updated this. Added some comments to this method.

phoenix-core/src/it/java/org/apache/phoenix/rpc/UpdateCacheConnectionLevelPropIT.java
ntshmah: Done.

phoenix-core/src/it/java/org/apache/phoenix/rpc/UpdateCacheConnectionLevelPropIT.java
ntshmah: I can do with a single `props` variable instead, that makes it clearer.

phoenix-core/src/it/java/org/apache/phoenix/rpc/UpdateCacheConnectionLevelPropIT.java
ntshmah: Noted, will update the headers.

phoenix-core/src/it/java/org/apache/phoenix/rpc/UpdateCacheConnectionLevelPropIT.java
ntshmah: The call to `verifyExpectedGetTableCalls()` before this block of code will also reset the spy object. This code will verify the number of expected `getTable()` calls after the reset. I'll add this information in the method header for `verifyExpectedGetTableCalls()`

phoenix-core/src/main/java/org/apache/phoenix/schema/MetaDataClient.java
ntshmah: Done. Created [PHOENIX-5653](https://issues.apache.org/jira/browse/PHOENIX-5653)

phoenix-core/src/main/java/org/apache/phoenix/schema/MetaDataClient.java
ntshmah: Done, added a debug log line here.

phoenix-core/src/main/java/org/apache/phoenix/schema/MetaDataClient.java
ntshmah: Would you prefer this investigation and adding of tests be done under a separate JIRA? Or this one itself?

phoenix-core/src/it/java/org/apache/phoenix/rpc/UpdateCacheConnectionLevelPropIT.java
```diff
@@ -0,0 +1,211 @@
+package org.apache.phoenix.rpc;
```
yanxinyi: nit: apache open-source license

ntshmah: Noted, will add that.

phoenix-core/src/main/java/org/apache/phoenix/schema/MetaDataClient.java
ChinmaySKulkarni: I think it should be covered under this JIRA since the understanding is that once you set the connection-level property, that would be honored for any actions taken with that connection. Good to go through various code paths where we `getTable` and see that we only get the latest metadata based on the UCF value

ntshmah: I've modified the tests to also check for the UPSERT path. UPSERT follows the same behavior as SELECT.

> I guess for DDL operations, we force the client to get the latest metadata
That's not true. We have logic in PhoenixStatement#executeMutation() to [force update cache if we run into meta not found exceptions](https://github.com/apache/phoenix/blob/master/phoenix-core/src/main/java/org/apache/phoenix/jdbc/PhoenixStatement.java#L427-L437).

phoenix-core/src/it/java/org/apache/phoenix/rpc/UpdateCacheConnectionLevelPropIT.java
```diff
@@ -0,0 +1,211 @@
...
+    }
+
+    /**
+     * Test 'Update Cache Frequency' property when it is set at connection-level only, and not at
+     * table-level.
+     */
+    @Test
+    public void testWithConnLevelUCFNoTableLevelUCF() throws Exception {
+        long tableUpdateCacheFrequency = -1;
+        long connUpdateCacheFrequency = 1000;
```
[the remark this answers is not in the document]
ntshmah: Done.

phoenix-core/src/it/java/org/apache/phoenix/rpc/UpdateCacheConnectionLevelPropIT.java
```diff
@@ -0,0 +1,211 @@
...
+    }
+
+    /**
+     * Test 'Update Cache Frequency' property when it is set at table-level only, and not at
+     * connection-level.
+     */
+    @Test
+    public void testWithTableLevelUCFNoConnLevelUCF() throws Exception {
+        long tableUpdateCacheFrequency = 1000;
+        long connUpdateCacheFrequency = -1;
```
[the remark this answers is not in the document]
ntshmah: Done.

phoenix-core/src/it/java/org/apache/phoenix/rpc/UpdateCacheConnectionLevelPropIT.java
```diff
@@ -0,0 +1,211 @@
...
+        // There should only be a single call to getTable() for fetching the table's metadata
+        int numSelectExecutions = 2;
+        int numExpectedGetTableCalls = 1;
+        setUpTableAndConnections(fullTableName, tableUpdateCacheFrequency, connUpdateCacheFrequency);
+        verifyExpectedGetTableCalls(fullTableName, numSelectExecutions, numExpectedGetTableCalls);
+
+        // Wait for a period of 'tableUpdateCacheFrequency' and verify that there was one new call to getTable() for
+        // fetching the table's metadata
+        Thread.sleep(tableUpdateCacheFrequency);
+        verifyExpectedGetTableCalls(fullTableName, numSelectExecutions, numExpectedGetTableCalls);
```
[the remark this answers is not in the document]
ntshmah: I've not made any changes related to this behavior here, and also the initial commits for UPDATE_CACHE_FREQUENCY feature were added a while ago so I'm assuming that works as expected. I could add new tests under a separate JIRA if you would like that.

ntshmah:
Hi @ckulkarni
I will be squashing my commits together and force pushing here since quite a few things changed from the time this was last reviewed. I hope that is fine.

phoenix-core/src/main/java/org/apache/phoenix/schema/MetaDataClient.java
ChinmaySKulkarni: Cool. Thanks for checking.

ntshmah:
Thank you, @yanxinyi and @ChinmaySKulkarni for the reviews and for committing the patches. I'll close this PR.

PR #660, opened 2019-12-19
PHOENIX-5644: IndexUpgradeTool should sleep only once if there is at …

swaroopak:
…least one immutable table provided

### DIFF
--- a/phoenix-core/src/main/java/org/apache/phoenix/mapreduce/index/IndexUpgradeTool.java
+++ b/phoenix-core/src/main/java/org/apache/phoenix/mapreduce/index/IndexUpgradeTool.java
@@ -49,6 +49,8 @@ import org.apache.phoenix.query.QueryServicesOptions;
 import org.apache.phoenix.schema.PIndexState;
 import org.apache.phoenix.schema.PTable;
 import org.apache.phoenix.schema.PTableType;
+import org.apache.phoenix.util.EnvironmentEdge;
+import org.apache.phoenix.util.EnvironmentEdgeManager;
 import org.apache.phoenix.util.MetaDataUtil;
 import org.apache.phoenix.util.PhoenixRuntime;
 
@@ -72,7 +74,8 @@ import java.util.UUID;
 import java.util.logging.FileHandler;
 import java.util.logging.SimpleFormatter;
 
-import static org.apache.phoenix.query.QueryServicesOptions.GLOBAL_INDEX_CHECKER_ENABLED_MAP_EXPIRATION_MIN;
+import static org.apache.phoenix.query.QueryServicesOptions.
+        GLOBAL_INDEX_CHECKER_ENABLED_MAP_EXPIRATION_MIN;
 
 public class IndexUpgradeTool extends Configured implements Tool {
 
@@ -116,6 +119,7 @@ public class IndexUpgradeTool extends Configured implements Tool {
     private String inputFile;
 
     private boolean test = false;
+    private boolean waited = false;
 
     public void setDryRun(boolean dryRun) {
         this.dryRun = dryRun;
@@ -134,6 +138,7 @@ public class IndexUpgradeTool extends Configured implements Tool {
     }
 
     public void setTest(boolean test) { this.test = test; }
+    public boolean getWaited() { return this.waited; }
     public boolean getDryRun() {
         return this.dryRun;
     }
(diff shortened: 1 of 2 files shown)


phoenix-core/src/main/java/org/apache/phoenix/mapreduce/index/IndexUpgradeTool.java
```diff
@@ -367,6 +358,53 @@ private int executeTool(Connection conn, ConnectionQueryServices queryServices,
...
     }
 
+    private void executeToolForImmutableTables(ConnectionQueryServices queryServices,
+            ArrayList<String> immutableList) {
+
+        boolean atLeastOneHasIndexes=false;
+
+        for (String dataTableFullName : immutableList) {
+            HashSet<String> indexes = tablesAndIndexes.get(dataTableFullName);
+            if (!(indexes.isEmpty() || atLeastOneHasIndexes)) {
```
gjacoby126: Don't think you need the "|| atLeastOneHasIndexes" here. If atLeastOneHasIndexes is already true, no need to set it to true again.

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/index/IndexUpgradeTool.java
```diff
@@ -367,6 +358,53 @@ private int executeTool(Connection conn, ConnectionQueryServices queryServices,
...
+        // their caches of table metadata
+        if (!(test || dryRun) && atLeastOneHasIndexes) {
+            LOGGER.info("The input list has at least one immutable table with indexes, "
+                    + "waiting for " + (GLOBAL_INDEX_CHECKER_ENABLED_MAP_EXPIRATION_MIN + 1)
+                    + " minutes for client cache to expire");
+            try {
+                Thread.sleep((GLOBAL_INDEX_CHECKER_ENABLED_MAP_EXPIRATION_MIN + 1)
+                        * 60 * 1000);
+            } catch (InterruptedException e) {
+                LOGGER.warning("Sleep before starting index rebuild is interrupted. "
```
gjacoby126: If we get interrupted here before we've slept our full time, shouldn't we take some action rather than just logging and swallowing the exception? If we proceed with enabling the table before the full span is done, we risk the cache coherency problems we were trying to fix in the first place.

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/index/IndexUpgradeTool.java
```diff
@@ -367,6 +358,53 @@ private int executeTool(Connection conn, ConnectionQueryServices queryServices,
...
+            ArrayList<String> immutableList) {
+
+        boolean atLeastOneHasIndexes=false;
+
+        for (String dataTableFullName : immutableList) {
+            HashSet<String> indexes = tablesAndIndexes.get(dataTableFullName);
+            if (!(indexes.isEmpty() || atLeastOneHasIndexes)) {
+                atLeastOneHasIndexes = true;
+            }
+            try (Admin admin = queryServices.getAdmin()) {
```
gjacoby126: Only need to create one Admin for this whole method, not two per data table. Docs say they're lightweight, but no need to churn.

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/index/IndexUpgradeTool.java
```diff
@@ -367,6 +358,53 @@ private int executeTool(Connection conn, ConnectionQueryServices queryServices,
...
+                HashSet<String> indexes = tablesAndIndexes.get(dataTableFullName);
+                enableTable(admin, dataTableFullName, indexes);
+            } catch (IOException | SQLException e) {
+                LOGGER.severe("Something went wrong while enabling immutable table "
+                        + dataTableFullName + " " + e);
+            }
+            LOGGER.info("Completed " + operation + " of " + dataTableFullName
+                    + " (immutable)");
+        }
+    }
```
gokceni: Consider adding tests with less sleep time than 11 mins

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/index/IndexUpgradeTool.java
```diff
@@ -367,6 +358,53 @@ private int executeTool(Connection conn, ConnectionQueryServices queryServices,
         return 0;
     }
 
+    private void executeToolForImmutableTables(ConnectionQueryServices queryServices,
```
kadirozde: I suggest having a separate list, i.e., mutableList, for mutable tables.  We should disable and modify immutable tables first and then record the current time, say as startWaitTime. Then we can upgrade mutable tables (disable, modify and rebuildIndexes). After that we can  sleep for 11 minutes - (current time - startWaitTime) and call rebuildIndexes for mutable tables. This eliminates unnecessary sleep time better.

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/index/IndexUpgradeTool.java
swaroopak: One thing I can think of is we can retry sleeping. I just added that logic.

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/index/IndexUpgradeTool.java
swaroopak: Changed the code.

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/index/IndexUpgradeTool.java
```diff
@@ -318,53 +322,123 @@ public int executeTool() {
...
+            for (String dataTableFullName : immutableList) {
+                HashSet<String> indexes = tablesAndIndexes.get(dataTableFullName);
+                LOGGER.info("Executing " + operation + " of " + dataTableFullName
+                        + " (immutable)");
+                disableTable(admin, dataTableFullName, indexes);
+                modifyTable(admin, dataTableFullName, indexes);
+            }
+        } catch (IOException | SQLException e) {
+            LOGGER.severe("Something went wrong while disabling "
+                    + "or modifying immutable table " + e);
```
kadirozde: I think we cannot simply ignore the failure here. We can downgrade the data table, remove it from the list and continue with the rest. If the downgrade fails too, then we can throw an exception.

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/index/IndexUpgradeTool.java
```diff
@@ -318,53 +322,123 @@ public int executeTool() {
...
-                    }
-                }
                 enableTable(admin, dataTableFullName, indexes);
                 LOGGER.info("Completed " + operation + " of " + dataTableFullName);
 
             } catch (IOException | SQLException e) {
-                LOGGER.severe("Something went wrong while executing " + operation
-                        + " steps " + e);
-                return -1;
+                LOGGER.severe("Something went wrong while executing "
```
kadirozde: Please see the above comment for immutable tables. I think we should not ignore the failures. If this happens during downgrade/rollback, we can attempt to upgrade the table back. Maybe we can track where the failure happened and just rollback the executed operation(s). If this gets too complicated to implement, then we can re-throw the exception with a proper message (instead of ignoring the exception).

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/index/IndexUpgradeTool.java
swaroopak: Thank you for the suggestion, kadir. I made the changes in the code.

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/index/IndexUpgradeTool.java
swaroopak: took the first approach. Thank you for the suggestion!

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/index/IndexUpgradeTool.java
```diff
@@ -318,53 +323,155 @@ public int executeTool() {
...
+                        + " (immutable)");
+                disableTable(admin, dataTableFullName, indexes);
+                modifyTable(admin, dataTableFullName, indexes);
+            }
+        } catch (IOException | SQLException e) {
+            LOGGER.severe("Something went wrong while disabling "
+                + "or modifying immutable table " + e);
+            handleFailure(queryServices, dataTableFullNameGlobal, immutableList);
+        }
+        long startWaitTime = System.currentTimeMillis();
```
gjacoby126: nit: should be EnvironmentEdgeManager.currentTime

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/index/IndexUpgradeTool.java
```diff
@@ -318,53 +323,155 @@ public int executeTool() {
...
+        executeToolForMutableTables(conn, queryServices, conf, mutableList);
+        enableImmutableTables(queryServices, immutableList, startWaitTime);
+        rebuildIndexes(conn, conf, immutableList);
+        return 0;
+    }
 
+    private long executeToolForImmutableTables(ConnectionQueryServices queryServices,
+            ArrayList<String> immutableList) {
+        LOGGER.info("Started " + operation + " for immutable tables");
+        try (Admin admin = queryServices.getAdmin()) {
```
priyankporwal: Do you want to put try/catch to be inside the for loop? It seems like handleFailure() is called for just the the table that hit an exception, and the rest of tables in the list are not handled (upgraded/rolledback).

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/index/IndexUpgradeTool.java
```diff
@@ -318,53 +323,155 @@ public int executeTool() {
...
-
-            } catch (IOException | SQLException e) {
-                LOGGER.severe("Something went wrong while executing " + operation
-                        + " steps " + e);
-                return -1;
             }
+        } catch (IOException | SQLException e) {
+            LOGGER.severe("Something went wrong while executing "
+                    + operation + " steps for "+ dataTableFullNameGlobal + " " + e);
+            handleFailure(queryServices, dataTableFullNameGlobal, mutableTables);
```
priyankporwal: Similar problem here too .. exception handling seems to be misplaced.

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/index/IndexUpgradeTool.java
```diff
@@ -318,53 +323,155 @@ public int executeTool() {
...
+        rebuildIndexes(conn, conf, mutableTables);
+    }
+
+    private void enableImmutableTables(ConnectionQueryServices queryServices,
+            ArrayList<String> immutableList,
+            long startWaitTime) {
+        int retryCount = 0;
+        long endWaitTime = System.currentTimeMillis();
+        long waitMore = getWaitMoreTime(endWaitTime, startWaitTime);
+        while  (retryCount < 5 && !waited) {
```
priyankporwal: const for 5 .. use here and on line 445

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/index/IndexUpgradeTool.java
```diff
@@ -318,53 +323,155 @@ public int executeTool() {
...
+            LOGGER.info("Sleep was interrupted multiple times for immutable tables, "
+                    + "we'll still go ahead and enable the table.");
+        }
+        try (Admin admin = queryServices.getAdmin()) {
+            for (String dataTableFullName: immutableList) {
+                dataTableFullNameGlobal = dataTableFullName;
+                HashSet<String> indexes = tablesAndIndexes.get(dataTableFullName);
+                enableTable(admin, dataTableFullName, indexes);
+            }
+        } catch (IOException | SQLException e) {
```
priyankporwal: Similar exception handling problem

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/index/IndexUpgradeTool.java
```diff
@@ -116,6 +117,8 @@
     private String inputFile;
 
     private boolean test = false;
+    private boolean waited = false;
+    private String dataTableFullNameGlobal;
```
priyankporwal: Why do we need this global?

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/index/IndexUpgradeTool.java
```diff
@@ -318,53 +323,155 @@ public int executeTool() {
...
-                LOGGER.info("Started index rebuild post " + operation + " of "
-                        + dataTableFullName);
+        rebuildIndexes(conn, conf, mutableTables);
+    }
+
+    private void enableImmutableTables(ConnectionQueryServices queryServices,
+            ArrayList<String> immutableList,
+            long startWaitTime) {
+        int retryCount = 0;
+        long endWaitTime = System.currentTimeMillis();
```
gjacoby126: likewise, should use environmentedgemanager

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/index/IndexUpgradeTool.java
```diff
@@ -318,53 +323,155 @@ public int executeTool() {
...
+            // their caches of table metadata
+
+            LOGGER.info("waiting for more " + waitMore + " ms for client cache "
+                    + "to expire for immutable tables");
+            try {
+                startWaitTime = System.currentTimeMillis();
+                Thread.sleep(waitMore);
+                waited = true;
+            } catch (InterruptedException e) {
+                endWaitTime = System.currentTimeMillis();
```
gjacoby126: ditto

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/index/IndexUpgradeTool.java
```diff
@@ -318,53 +323,155 @@ public int executeTool() {
...
+                waitMore = getWaitMoreTime(endWaitTime, startWaitTime);
+                retryCount++;
+                LOGGER.warning("Sleep before starting index rebuild is interrupted. "
+                        + "Attempting to sleep again! " + e.getMessage());
             }
         }
-        return 0;
+        if (retryCount == 5) {
+            LOGGER.info("Sleep was interrupted multiple times for immutable tables, "
```
gjacoby126: why do we eat this exception? Do we need the max retries since we already have a max timeout to prevent infinite loops?

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/index/IndexUpgradeTool.java
```diff
@@ -318,53 +323,155 @@ public int executeTool() {
...
+                HashSet<String> indexes = tablesAndIndexes.get(dataTableFullName);
+                enableTable(admin, dataTableFullName, indexes);
+            }
+        } catch (IOException | SQLException e) {
+            LOGGER.severe("Something went wrong while enabling immutable table "
+                    + dataTableFullNameGlobal);
+            //removing to avoid any rebuilds after upgrade
+            tablesAndIndexes.remove(dataTableFullNameGlobal);
+            immutableList.remove(dataTableFullNameGlobal);
+            throw new RuntimeException("Manually enable the table " + dataTableFullNameGlobal
```
gjacoby126: If enabling table 2 of 5 fails, you need to manually enable tables 2,3,4, and 5, and the error message should say so.

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/index/IndexUpgradeTool.java
swaroopak: We want to retry in case of interruption during sleep. And only try a finite number of times. I think for this corner case scenario, we are okay if we see any problem with cache incoherency. What say?

priyankporwal: I agree with not having the needless retry-count logic. Whether the sleep is interrupted or not, as long as the clock moved beyond initial time + GLOBAL_INDEX_CHECKER_ENABLED_MAP_EXPIRATION_MIN + 1 minute, it shouldn't matter how many retries of sleep were needed. Safer than not sleeping enough and let caches be stale.

kadirozde: I agree with @gjacoby126, no need to have the max retries here. I suggest removing retryCount and having 
while (waitMore  > 0) instead.

swaroopak: Thank you, everyone! Initially, the logic wasn't there and I was concerned about stale cache then. Looks like we are okay there.

PR #662, opened 2019-12-20
PHOENIX-5645 - GlobalIndexChecker should prevent compaction from purg…

gjacoby126:
…ing very recently deleted cells

### DIFF
--- a/phoenix-core/src/it/java/org/apache/phoenix/end2end/MaxLookbackIT.java
+++ b/phoenix-core/src/it/java/org/apache/phoenix/end2end/MaxLookbackIT.java
@@ -1,0 +1,351 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.phoenix.end2end;
+
+
+import com.google.common.collect.Maps;
+import org.apache.commons.logging.Log;
+import org.apache.commons.logging.LogFactory;
+import org.apache.hadoop.conf.Configuration;
+import org.apache.hadoop.hbase.TableName;
+import org.apache.hadoop.hbase.client.Admin;
+import org.apache.hadoop.hbase.protobuf.generated.AdminProtos;
+import org.apache.hadoop.hbase.regionserver.HRegion;
+import org.apache.hadoop.hbase.regionserver.ScanInfoUtil;
+import org.apache.hadoop.hbase.util.Bytes;
+import org.apache.hadoop.hbase.util.EnvironmentEdgeManager;
+import org.apache.phoenix.exception.SQLExceptionCode;
+import org.apache.phoenix.query.QueryServices;
+import org.apache.phoenix.util.PhoenixRuntime;
+import org.apache.phoenix.util.QueryUtil;
+import org.apache.phoenix.util.ReadOnlyProps;
+import org.apache.phoenix.util.TestUtil;
+import org.junit.Assert;
+import org.junit.Before;
+import org.junit.BeforeClass;
+import org.junit.Test;
+
+import java.io.IOException;
+import java.sql.Connection;
+import java.sql.DriverManager;
+import java.sql.ResultSet;
+import java.sql.SQLException;
+import java.sql.Statement;
+import java.util.Map;
+import java.util.Properties;
+
+import static org.apache.phoenix.util.TestUtil.assertRawCellCount;
+import static org.apache.phoenix.util.TestUtil.assertRawRowCount;
+import static org.apache.phoenix.util.TestUtil.assertRowExistsAtSCN;
+import static org.apache.phoenix.util.TestUtil.assertRowHasExpectedValueAtSCN;
+import static org.apache.phoenix.util.TestUtil.assertTableHasTtl;
+import static org.apache.phoenix.util.TestUtil.assertTableHasVersions;
+
+@NeedsOwnMiniClusterTest
+public class MaxLookbackIT extends BaseUniqueNamesOwnClusterIT {
+    private static final Log LOG = LogFactory.getLog(MaxLookbackIT.class);
+    private static final int MAX_LOOKBACK_AGE = 10;
+    private static final int ROWS_POPULATED = 2;
+    private String tableDDLOptions;
+    private StringBuilder optionBuilder;
+
+    @BeforeClass
+    public static synchronized void doSetup() throws Exception {
+        Map<String, String> props = Maps.newHashMapWithExpectedSize(1);
+        props.put(QueryServices.GLOBAL_INDEX_ROW_AGE_THRESHOLD_TO_DELETE_MS_ATTRIB, Long.toString(0));
+        props.put(ScanInfoUtil.PHOENIX_MAX_LOOKBACK_AGE_CONF_KEY, Integer.toString(MAX_LOOKBACK_AGE));
+        setUpTestDriver(new ReadOnlyProps(props.entrySet().iterator()));
+    }
+
+    @Before
+    public void beforeTest(){
+        optionBuilder = new StringBuilder();
+        this.tableDDLOptions = optionBuilder.toString();
+    }
+
+    @Test
+    public void testTooLowSCNWithMaxLookbackAge() throws Exception {
+        try (Connection conn = DriverManager.getConnection(getUrl())) {
+            String dataTableName = generateUniqueName();
+            String indexStem = generateUniqueName();
+            createTableAndIndexes(conn, dataTableName, indexStem);
+            //need to sleep long enough for the SCN to still find the syscat row for the table
+            Thread.sleep(MAX_LOOKBACK_AGE * 1000 + 1000);
+            Properties props = new Properties();
+            props.setProperty(PhoenixRuntime.CURRENT_SCN_ATTRIB,
+                Long.toString(EnvironmentEdgeManager.currentTime() - (MAX_LOOKBACK_AGE + 1) * 1000));
+            try (Connection connscn = DriverManager.getConnection(getUrl(), props)) {
+                connscn.createStatement().executeQuery("select * from " + dataTableName);
+            } catch (SQLException se) {
+                SQLExceptionCode code =
+                    SQLExceptionCode.CANNOT_QUERY_TABLE_WITH_SCN_OLDER_THAN_MAX_LOOKBACK_AGE;
+                TestUtil.assertSqlExceptionCode(code, se);
+                return;
+            }
+        }
+        Assert.fail("We should have thrown an exception for the too-early SCN");
+    }
+
+    @Test(timeout=120000L)
+    public void testRecentlyDeletedRowsNotCompactedAway() throws Exception {
+        try (Connection conn = DriverManager.getConnection(getUrl())) {
+            String dataTableName = generateUniqueName();
+            String indexStem = generateUniqueName();
+            createTableAndIndexes(conn, dataTableName, indexStem);
+            String fullIndexName = indexStem + "1";
+            TableName dataTable = TableName.valueOf(dataTableName);
+            TableName indexTable = TableName.valueOf(fullIndexName);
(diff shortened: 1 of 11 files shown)


gjacoby126:
Note that the new test I just pushed up to verify TTL behavior is currently failing.

phoenix-core/src/main/java/org/apache/phoenix/index/GlobalIndexChecker.java
```diff
@@ -48,9 +52,16 @@
 import org.apache.hadoop.hbase.coprocessor.ObserverContext;
 import org.apache.hadoop.hbase.coprocessor.RegionCoprocessorEnvironment;
 import org.apache.hadoop.hbase.io.ImmutableBytesWritable;
+import org.apache.hadoop.hbase.regionserver.InternalScanner;
+import org.apache.hadoop.hbase.regionserver.KeyValueScanner;
 import org.apache.hadoop.hbase.regionserver.Region;
 import org.apache.hadoop.hbase.regionserver.RegionScanner;
+import org.apache.hadoop.hbase.regionserver.ScanInfo;
```
swaroopak: nit: are these imports used in this file?

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/BaseScannerRegionObserver.java
```diff
@@ -374,20 +382,111 @@ RegionScanner getWrappedScanner(final ObserverContext<RegionCoprocessorEnvironme
...
+                " and region " + store.getRegionInfo().getRegionNameAsString() +
+                " and cf " + store.getColumnFamilyName());
+        }
+        return new StoreScanner(store, scanInfo, scan, Collections.singletonList(memstoreScanner),
+            scanType, store.getSmallestReadPoint(),
+            HConstants.LATEST_TIMESTAMP);
+    }
+
+    @Override
+    public InternalScanner preCompactScannerOpen(
```
swaroopak: There is a lot of commonality in the function body of the above two hooks. Is it possible to throw that in a common function and invoke it from both the hooks?

phoenix-core/src/it/java/org/apache/phoenix/end2end/SCNIT.java
```diff
@@ -82,28 +85,58 @@ public void testReadBeforeDelete() throws Exception {
...
+	@Test
+	public void testTooLowSCNWithTTL() throws Exception {
+		//if scn is for an older time than a table's ttl, it should throw a SQLException
+		int ttl = 2;
+		String fullTableName = createTableWithTTL(ttl);
+		//need to sleep long enough for the SCN to still find the syscat row for the table
+		Thread.sleep(ttl * 1000 + 1000);
+		Properties props = new Properties();
+		props.setProperty(PhoenixRuntime.CURRENT_SCN_ATTRIB,
+			Long.toString(EnvironmentEdgeManager.currentTime() - (ttl + 1)* 1000));
```
swaroopak: nit: (ttl+1)*1000 can be a new variable and reused at 108 and 111

phoenix-core/src/it/java/org/apache/phoenix/end2end/SCNIT.java
```diff
@@ -82,28 +85,58 @@ public void testReadBeforeDelete() throws Exception {
...
+		try (Connection connscn = DriverManager.getConnection(getUrl(), props)) {
+			connscn.createStatement().executeQuery("select * from "
```
swaroopak: nit : use format string? 113 and 96

phoenix-core/src/it/java/org/apache/phoenix/end2end/index/GlobalIndexCheckerIT.java
```diff
@@ -385,21 +402,36 @@ public void testOnePhaseOverwrite() throws Exception {
...
+        conn.commit();
+        if (async) {
+            // run the index MR job.
+            IndexToolIT.runIndexTool(true, false, null, dataTableName, indexTableName + "1");
+            IndexToolIT.runIndexTool(true, false, null, dataTableName, indexTableName + "2");
+        }
+    }
+
+    private void assertTableHasTtl(Connection conn, TableName tableName, int ttl)
+        throws Exception{
```
swaroopak: nit: space begore '{'

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/BaseScannerRegionObserver.java
```diff
@@ -374,20 +382,111 @@ RegionScanner getWrappedScanner(final ObserverContext<RegionCoprocessorEnvironme
...
+            ScanInfoUtil.getScanInfoForFlushesAndCompactions(conf, oldScanInfo,
+                store, scanType);
+        return new StoreScanner(store, scanInfo, scan, scanners, scanType,
+            store.getSmallestReadPoint(),
+            earliestPutTs);
+    }
+
+
+
+
```
swaroopak: nit: multiple empty lines

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/BaseScannerRegionObserver.java
```diff
@@ -374,20 +382,111 @@ RegionScanner getWrappedScanner(final ObserverContext<RegionCoprocessorEnvironme
...
 
+
     @Override
     public KeyValueScanner preStoreScannerOpen(final ObserverContext<RegionCoprocessorEnvironment> c,
         final Store store, final Scan scan, final NavigableSet<byte[]> targetCols,
         final KeyValueScanner s) throws IOException {
-
-      if (scan.isRaw() || ScanInfoUtil.isKeepDeletedCells(store.getScanInfo()) || scan.getTimeRange().getMax() == HConstants.LATEST_TIMESTAMP || TransactionUtil.isTransactionalTimestamp(scan.getTimeRange().getMax())) {
+      if (storeFileScanDoesntNeedAlteration(store, scan)) {
         return s;
```
swaroopak: nit: is something weird on this indentation?

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/BaseScannerRegionObserver.java
```diff
@@ -374,20 +382,111 @@ RegionScanner getWrappedScanner(final ObserverContext<RegionCoprocessorEnvironme
...
+                store.getTableName().getNameAsString() + " and region " +
+                store.getRegionInfo().getRegionNameAsString() +
+                " and cf " + store.getColumnFamilyName());
+        }
+
+        Configuration conf = c.getEnvironment().getConfiguration();
+        ScanInfo scanInfo =
+            ScanInfoUtil.getScanInfoForFlushesAndCompactions(conf, oldScanInfo,
+                store, scanType);
+        return new StoreScanner(store, scanInfo, scan, scanners, scanType,
```
swaroopak: unrelated to this PR. @gjacoby126 @abhishek-chouhan Looking at StoreScanner, seems like a scope to refactor the class into using the builder pattern. Any thoughts?

phoenix-core/src/main/java/org/apache/hadoop/hbase/regionserver/ScanInfoUtil.java
```diff
@@ -20,28 +20,107 @@
...
+    private static int getMinVersions(ScanInfo oldScanInfo, final Store store) {
+        return oldScanInfo.getTtl() != Long.MAX_VALUE ? store.getFamily().getMinVersions()
+            : Math.max(store.getFamily().getMinVersions(), 1);
+    }
+
+    public static ScanInfo getScanInfoForFlushesAndCompactions(Configuration conf,
+                                                               ScanInfo oldScanInfo,
+                                                         final Store store,
+                                                         ScanType type) {
+        long ttl = ScanInfoUtil.getTimeToLiveForCompactions(store.getFamily(), oldScanInfo);
```
swaroopak: nit: no need to provide the class name.

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/BaseScannerRegionObserver.java
gjacoby126: The builder pattern would improve the StoreScanner's readability, yes, though since StoreScanner's IA.Private and really only supposed to be used within HBase, it would be to improve the HBase codebase more than Phoenix. Hopefully in the medium term we can get HBASE-23602 done so that we have a higher-level, supported way to do this.

phoenix-core/src/main/java/org/apache/phoenix/index/GlobalIndexChecker.java
gjacoby126: Thanks, removed them.

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/BaseScannerRegionObserver.java
gjacoby126: I thought of that, but they're just a bit different and some of the differences were in the middle, so I didn't see an easy way to consolidate. If you do please let me know -- always nice to eliminate dupes.

phoenix-core/src/it/java/org/apache/phoenix/end2end/SCNIT.java
gjacoby126: Good catch, done.

phoenix-core/src/it/java/org/apache/phoenix/end2end/SCNIT.java
gjacoby126: Done.

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/BaseScannerRegionObserver.java
gjacoby126: Done

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/BaseScannerRegionObserver.java
gjacoby126: Fixed indentation.

phoenix-core/src/it/java/org/apache/phoenix/end2end/index/GlobalIndexCheckerIT.java
gjacoby126: Done.

phoenix-core/src/main/java/org/apache/hadoop/hbase/regionserver/ScanInfoUtil.java
gjacoby126: Done

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/BaseScannerRegionObserver.java
```diff
@@ -374,20 +382,111 @@ RegionScanner getWrappedScanner(final ObserverContext<RegionCoprocessorEnvironme
...
+        }
+
+        // Called during flushing the memstore to disk.
+        // Need to retain all the delete markers & all the versions
+        Scan scan = new Scan();
+        scan.setMaxVersions(Integer.MAX_VALUE);
+        ScanInfo oldScanInfo = store.getScanInfo();
+
+        Configuration conf = c.getEnvironment().getConfiguration();
+        //minor compactions and flushes both use "compact retain deletes"
```
swaroopak: as per the comment here, should line 440 also be in other hook?

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/BaseScannerRegionObserver.java
swaroopak: I think it is more like readability vs DRY. We could put everything into a function and have a boolean passed but I think conditionals make things more complicated.

phoenix-core/src/it/java/org/apache/phoenix/end2end/MaxLookbackIT.java
```diff
@@ -0,0 +1,351 @@
...
+        }
+        Assert.fail("We should have thrown an exception for the too-early SCN");
+    }
+
+    @Test(timeout=120000L)
+    public void testRecentlyDeletedRowsNotCompactedAway() throws Exception {
+        try (Connection conn = DriverManager.getConnection(getUrl())) {
+            String dataTableName = generateUniqueName();
+            String indexStem = generateUniqueName();
+            createTableAndIndexes(conn, dataTableName, indexStem);
```
gokceni: nit: You seem to be creating dataTable and indexes in each of these tests. Recommend you to move it to a @Before method

gjacoby126: Unfortunately the tables are slightly different each time.

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/BaseScannerRegionObserver.java
gjacoby126: Compactions already set the scan type in the hook.

PR #665, opened 2019-12-25
PHOENIX-5650: IndexUpgradeTool does not rebuild view indexes

swaroopak:


phoenix-core/src/main/java/org/apache/phoenix/mapreduce/index/IndexUpgradeTool.java
```diff
@@ -648,17 +648,18 @@ private boolean extractTablesAndIndexes(PhoenixConnection conn) {
...
                     + (!StringUtil.EMPTY_STRING.equals(schemaName) ? "AND TABLE_SCHEM = \'"
                     + schemaName + "\' " : "")
                     + "AND LINK_TYPE = "
-                    + PTable.LinkType.PHYSICAL_TABLE.getSerializedValue();
+                    + PTable.LinkType.CHILD_TABLE.getSerializedValue();
 
             ResultSet rs = conn.createStatement().executeQuery(viewSql);
             while (rs.next()) {
-                String viewName = rs.getString(1);
+                String viewFullName = rs.getString(1);
```
yanxinyi: nit: maybe better to name `viewName` here?
I think it's better to have String `viewFullName` = SchemaUtil.getTableName(schemaName, `viewName`), since this is a full name that contains schema and view name.

phoenix-core/src/it/java/org/apache/phoenix/end2end/ParameterizedIndexUpgradeToolIT.java
```diff
@@ -346,6 +358,26 @@ public void testDryRunAndFailures() throws Exception {
         }
     }
 
+    @Test
+    public void verifyViewAndViewIndexes() throws SQLException {
+        ResultSet rs = conn.createStatement().executeQuery(VIEWS_QUERY);
+        int countViews = -1, indexCount = -1;
+        while(rs.next()) {
```
yanxinyi: nit: `if (rs.next())` is better? We technically can't go there more than 1 once, since we just `SELECT COUNT`

yanxinyi:
overall looks great! Just two small nits from my perspective. Thank you for the patch and happy holiday!

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/index/IndexUpgradeTool.java
swaroopak: I meant to use getTableNameFromFullName(); thanks for pointing out though. :)

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/index/IndexUpgradeTool.java
```diff
@@ -648,17 +648,18 @@ private boolean extractTablesAndIndexes(PhoenixConnection conn) {
         }
 
         if (hasViewIndex) {
-            String viewSql = "SELECT DISTINCT TABLE_NAME, TENANT_ID FROM "
-                    + "SYSTEM.CATALOG "
-                    + "WHERE COLUMN_FAMILY = \'" + dataTableFullName + "\' "
+            String viewSql = "SELECT DISTINCT COLUMN_FAMILY, TENANT_ID FROM "
```
gjacoby126: can this be extracted into a reusable function that verifyViewAndViewIndexes can also use? Right now the test is verifying that its _own_ query works as expected, but not the actual upgrade tool's. Imagine if someone changed IndexUpgradeTool:651 and broke it -- the test would still pass.

phoenix-core/src/it/java/org/apache/phoenix/end2end/ParameterizedIndexUpgradeToolIT.java
```diff
@@ -346,6 +362,28 @@ public void testDryRunAndFailures() throws Exception {
         }
     }
 
+    @Test
+    public void verifyViewAndViewIndexes() throws SQLException {
```
gjacoby126: Does this test belong in a different class? Doesn't actually seem related to index upgrades, but to looking up metadata from syscat.

swaroopak: Sure. :)

swaroopak: I kept it in this file, as it is testing the code in IUT.

swaroopak: @gjacoby126 Where do you propose to keep this test? I was considering "NonParameterizedIndexUpgradeToolIT". any thoughts?

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/index/IndexUpgradeTool.java
```diff
@@ -692,6 +692,17 @@ private boolean extractTablesAndIndexes(PhoenixConnection conn) {
         return viewIndexes;
     }
 
+    @VisibleForTesting
+    public String getViewIndexesSql(String viewName, String schemaName, String tenantId) {
+        return "SELECT DISTINCT COLUMN_FAMILY FROM "
+                + "SYSTEM.CATALOG "
+                + "WHERE TABLE_NAME = \'" + viewName + "\'"
+                + (!StringUtil.EMPTY_STRING.equals(schemaName) ? "AND TABLE_SCHEM = \'"
```
priyankporwal: Nit: Add space before AND

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/index/IndexUpgradeTool.java
```diff
@@ -692,6 +692,17 @@ private boolean extractTablesAndIndexes(PhoenixConnection conn) {
...
+                + schemaName + "\' " : "")
+                + "AND LINK_TYPE = " + PTable.LinkType.INDEX_TABLE.getSerializedValue()
```
priyankporwal: Likewise, add space before AND

phoenix-core/src/it/java/org/apache/phoenix/end2end/ParameterizedIndexUpgradeToolIT.java
```diff
@@ -86,6 +91,8 @@
     private static Map<String, String> serverProps = Maps.newHashMapWithExpectedSize(1),
             clientProps = Maps.newHashMapWithExpectedSize(1);
 
+    public static final String ASSERT_MESSAGE = "view-index count in system table doesn't match";
```
gjacoby126: nit: better constant name.

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/index/IndexUpgradeTool.java
```diff
@@ -673,16 +668,21 @@ private boolean extractTablesAndIndexes(PhoenixConnection conn) {
         return indexInfos;
     }
 
+    @VisibleForTesting
+    public String getViewSql(String tableName, String schemaName) {
+       return "SELECT DISTINCT COLUMN_FAMILY, TENANT_ID FROM "
+                + "SYSTEM.CHILD_LINK "
+                + "WHERE TABLE_NAME = \'" + tableName + "\' "
+                + (!StringUtil.EMPTY_STRING.equals(schemaName) ? "AND TABLE_SCHEM = \'"
```
priyankporwal: That seems like a very long way to check emptiness of a string. Perhaps add an isEmpty() fn in StringUtil class, if schemaName.isEmpty() is not sufficient for whatever reason.

priyankporwal: Also, would be much simpler to write as schemaName.isEmpty() ? "" : "AND TABLE_SCHEM = \'" + schemaName + "\' " .. instead of negating it first.

priyankporwal: Please add spaces before AND's

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/index/IndexUpgradeTool.java
swaroopak: There is a space after quote but I think before AND makes it more readable. ll change it.

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/index/IndexUpgradeTool.java
swaroopak: to avoid any NPE due to schemaName being null, it's better to check .equals on a constant. It reads longer but it is not that long :)

swaroopak: There are spaces after the quote, but before AND makes it more readable. ll change it.

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/index/IndexUpgradeTool.java
```diff
@@ -673,16 +668,21 @@ private boolean extractTablesAndIndexes(PhoenixConnection conn) {
         return indexInfos;
     }
 
+    @VisibleForTesting
+    public String getViewSql(String tableName, String schemaName) {
+       return "SELECT DISTINCT COLUMN_FAMILY, TENANT_ID FROM "
+                + "SYSTEM.CHILD_LINK "
+                + "WHERE TABLE_NAME = \'" + tableName + "\'"
+                + (StringUtil.EMPTY_STRING.equals(schemaName) ? " AND TABLE_SCHEM = \'"
```
priyankporwal: The order got messed up here (bug). Also, can you please add isNullOrEmpty() in StringUtil class and use it here? Your argument to avoid NPE is sort of incomplete since null string is being simply concatenated in the query string as if it was more valid empty schemaName - no difference functionally.

swaroopak: It wasn't a complete push. I updated the commit. I think you are right about using isNullOrEmpty.

phoenix-core/src/it/java/org/apache/phoenix/end2end/ParameterizedIndexUpgradeToolIT.java
swaroopak: Is this one better, VERIFY_COUNT_ASSERT_MESSAGE?

PR #668, opened 2019-12-31
PHOENIX-5654: String values (ALWAYS and NEVER) don't work for connect…

ntshmah:
…ion level config phoenix.default.update.cache.frequency.

- Introduced a new enum ConnectionProperty (similar to the existing TableProperty)
for connection level properties.
- Added validation code for connection properties at connection creation time.
- Changed the type of DEFAULT_UPDATE_CACHE_FREQUENCY from int to long.

Other changes made as part of this JIRA:
- Updated TableProperty.UPDATE_CACHE_FREQUENCY to raise relevant error message
for invalid table level property
- More tests for valid and invalid table/connection level property.

### DIFF
--- a/phoenix-core/src/it/java/org/apache/phoenix/end2end/CreateTableIT.java
+++ b/phoenix-core/src/it/java/org/apache/phoenix/end2end/CreateTableIT.java
@@ -32,6 +32,8 @@ import java.sql.PreparedStatement;
 import java.sql.ResultSet;
 import java.sql.SQLException;
 import java.sql.Statement;
+import java.util.ArrayList;
+import java.util.HashMap;
 import java.util.List;
 import java.util.Properties;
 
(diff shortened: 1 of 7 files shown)


phoenix-core/src/it/java/org/apache/phoenix/end2end/CreateTableIT.java
```diff
@@ -621,55 +623,96 @@ public void testMultiTenantImmutableTableMetadata() throws Exception {
         }
     }
 
+    private void verifyUCFValueInSysCat(String tableName, String createTableString, Properties props, long expectedUCFInSysCat) throws SQLException {
+        String readSysCatQuery = "SELECT TABLE_NAME, UPDATE_CACHE_FREQUENCY FROM SYSTEM.CATALOG WHERE "
+                + "TABLE_NAME = '" + tableName + "'  AND TABLE_TYPE='u'";
+
+        Connection connection = DriverManager.getConnection(getUrl(), props);
```
ChinmaySKulkarni: nit: Put connection object in a try-with-resources

phoenix-core/src/it/java/org/apache/phoenix/end2end/CreateTableIT.java
```diff
@@ -621,55 +623,96 @@ public void testMultiTenantImmutableTableMetadata() throws Exception {
...
+
+        for (String tableLevelUCF : invalidUCF) {
+            String createTableString = "CREATE TABLE " + tableName + " (k VARCHAR PRIMARY KEY,"
+                    + "v1 VARCHAR, v2 VARCHAR) UPDATE_CACHE_FREQUENCY = " + tableLevelUCF;
+            try {
+                verifyUCFValueInSysCat(tableName, createTableString, props, -1L);
+                fail();
+            } catch (IllegalArgumentException e) {
+                // expected
+                assertEquals("Table's UPDATE_CACHE_FREQUENCY can only be set to 'ALWAYS', 'NEVER' or a millisecond numeric value.",
```
ChinmaySKulkarni: I dislike comparing actual string messages in the exception, since a subtle change can cause test failures. Is there any other way you can confirm this via a sqlCode comparison for example?

phoenix-core/src/it/java/org/apache/phoenix/rpc/UpdateCacheIT.java
```diff
@@ -244,4 +248,25 @@ private static void validateSelectRowKeyCols(Connection conn, String selectSql,
...
+        invalidUCF.add("10000.6");
+
+        for (String connLevelUCF : invalidUCF) {
+            props.put(QueryServices.DEFAULT_UPDATE_CACHE_FREQUENCY_ATRRIB, connLevelUCF);
+            try {
+                DriverManager.getConnection(getUrl(), props);
+                fail();
+            } catch (IllegalArgumentException e) {
+                // expected
+                assertEquals("Connection's phoenix.default.update.cache.frequency can only be set to 'ALWAYS', 'NEVER' or a millisecond numeric value.",
```
ChinmaySKulkarni: Same with comparing the message string here.

phoenix-core/src/main/java/org/apache/phoenix/schema/ConnectionProperty.java
```diff
@@ -0,0 +1,50 @@
...
+public enum ConnectionProperty {
+    /**
+     * Connection level property phoenix.default.update.cache.frequency
+     */
+    UPDATE_CACHE_FREQUENCY() {
+        @Override
+        public Object getValue(String value) {
+            if (value == null) {
+                return QueryServicesOptions.DEFAULT_UPDATE_CACHE_FREQUENCY;
+            } else {
```
ChinmaySKulkarni: nit: Don't need an `else` since we return immediately inside the `if` block.

phoenix-core/src/main/java/org/apache/phoenix/schema/ConnectionProperty.java
```diff
@@ -0,0 +1,50 @@
...
+                if ("ALWAYS".equalsIgnoreCase(value)) {
+                    return 0L;
```
ChinmaySKulkarni: nit: ditto comment about `else` conditions

phoenix-core/src/main/java/org/apache/phoenix/schema/ConnectionProperty.java
```diff
@@ -0,0 +1,50 @@
...
+                } else if ("NEVER".equalsIgnoreCase(value)) {
+                    return Long.MAX_VALUE;
+                } else {
+                    try {
+                        return Long.parseLong(value);
+                    } catch (NumberFormatException e) {
+                        throw new IllegalArgumentException("Connection's phoenix.default.update.cache.frequency can only be set to 'ALWAYS', 'NEVER' or a millisecond numeric value.");
+                    }
```
ChinmaySKulkarni: nit: Replace the string `phoenix.default.update.cache.frequency` with the static variable name

phoenix-core/src/it/java/org/apache/phoenix/end2end/CreateTableIT.java
ntshmah: Done.

phoenix-core/src/it/java/org/apache/phoenix/end2end/CreateTableIT.java
ntshmah: The message comparison was probably an overhead in the first place, I only need to verify that an IllegalArgumentException was thrown. Will remove the message comparison.

phoenix-core/src/it/java/org/apache/phoenix/rpc/UpdateCacheIT.java
ntshmah: Okay, same here.

phoenix-core/src/main/java/org/apache/phoenix/schema/ConnectionProperty.java
ntshmah: Done.

phoenix-core/src/main/java/org/apache/phoenix/schema/ConnectionProperty.java
ntshmah: Done.

phoenix-core/src/main/java/org/apache/phoenix/schema/ConnectionProperty.java
ntshmah: Done.

ntshmah:
Thanks for the review, @ChinmaySKulkarni. I have addressed your comments, can you please take a look again?

ntshmah:
Thanks @ChinmaySKulkarni. Sure, on it.

PR #669, opened 2020-01-02
PHOENIX-5651: IndexScrutiny does not handle TTL/row-expiry

swaroopak:
1. Added new test parameter
2. Added new custom clock 
3. Added an Integration Test.

### DIFF
--- a/phoenix-core/src/it/java/org/apache/phoenix/end2end/IndexScrutinyToolBaseIT.java
+++ b/phoenix-core/src/it/java/org/apache/phoenix/end2end/IndexScrutinyToolBaseIT.java
@@ -16,6 +16,7 @@ import org.apache.hadoop.conf.Configuration;
 import org.apache.hadoop.hbase.HConstants;
 import org.apache.hadoop.mapreduce.Counters;
 import org.apache.hadoop.mapreduce.Job;
+import org.apache.phoenix.mapreduce.index.IndexScrutinyMapperForTest;
 import org.apache.phoenix.mapreduce.index.IndexScrutinyTool;
 import org.apache.phoenix.mapreduce.index.IndexScrutinyTool.OutputFormat;
 import org.apache.phoenix.mapreduce.index.IndexScrutinyTool.SourceTable;
@@ -54,7 +55,7 @@ public class IndexScrutinyToolBaseIT extends BaseTest {
     }
 
     protected List<Job> runScrutiny(String[] cmdArgs) throws Exception {
-        IndexScrutinyTool scrutiny = new IndexScrutinyTool();
+        IndexScrutinyTool scrutiny = new IndexScrutinyTool(IndexScrutinyMapperForTest.class);
         Configuration conf = new Configuration(getUtility().getConfiguration());
         scrutiny.setConf(conf);
         int status = scrutiny.run(cmdArgs);
--- a/phoenix-core/src/main/java/org/apache/phoenix/mapreduce/index/IndexScrutinyMapper.java
+++ b/phoenix-core/src/main/java/org/apache/phoenix/mapreduce/index/IndexScrutinyMapper.java
@@ -27,16 +27,22 @@ import java.sql.SQLException;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.HashMap;
+import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
 import java.util.Properties;
+import java.util.Set;
 
 import org.apache.commons.codec.binary.Hex;
 import org.apache.hadoop.conf.Configuration;
+import org.apache.hadoop.hbase.HTableDescriptor;
+import org.apache.hadoop.hbase.TableName;
+import org.apache.hadoop.hbase.client.Admin;
 import org.apache.hadoop.hbase.util.Pair;
 import org.apache.hadoop.io.NullWritable;
 import org.apache.hadoop.io.Text;
 import org.apache.hadoop.mapreduce.Mapper;
+import org.apache.phoenix.jdbc.PhoenixConnection;
 import org.apache.phoenix.jdbc.PhoenixResultSet;
 import org.apache.phoenix.mapreduce.PhoenixJobCounters;
 import org.apache.phoenix.mapreduce.index.IndexScrutinyTool.OutputFormat;
@@ -45,7 +51,10 @@ import org.apache.phoenix.mapreduce.util.ConnectionUtil;
 import org.apache.phoenix.mapreduce.util.PhoenixConfigurationUtil;
 import org.apache.phoenix.parse.HintNode.Hint;
 import org.apache.phoenix.schema.PTable;
+import org.apache.phoenix.schema.PTableType;
 import org.apache.phoenix.util.ColumnInfo;
+import org.apache.phoenix.util.EnvironmentEdgeManager;
+import org.apache.phoenix.util.MetaDataUtil;
 import org.apache.phoenix.util.PhoenixRuntime;
 import org.apache.phoenix.util.QueryUtil;
 import org.apache.phoenix.util.SchemaUtil;
@@ -54,22 +63,23 @@ import org.slf4j.LoggerFactory;
 
 import com.google.common.base.Joiner;
 
+
 /**
  * Mapper that reads from the data table and checks the rows against the index table
  */
 public class IndexScrutinyMapper extends Mapper<NullWritable, PhoenixIndexDBWritable, Text, Text> {
 
     private static final Logger LOGGER = LoggerFactory.getLogger(IndexScrutinyMapper.class);
-    private Connection connection;
+    protected Connection connection;
     private List<ColumnInfo> targetTblColumnMetadata;
     private long batchSize;
     // holds a batch of rows from the table the mapper is iterating over
     // Each row is a pair - the row TS, and the row values
-    private List<Pair<Long, List<Object>>> currentBatchValues = new ArrayList<>();
-    private String targetTableQuery;
-    private int numTargetPkCols;
-    private boolean outputInvalidRows;
-    private OutputFormat outputFormat = OutputFormat.FILE;
+    protected List<Pair<Long, List<Object>>> currentBatchValues = new ArrayList<>();
+    protected String targetTableQuery;
+    protected int numTargetPkCols;
+    protected boolean outputInvalidRows;
+    protected OutputFormat outputFormat = OutputFormat.FILE;
     private String qSourceTable;
     private String qTargetTable;
     private long executeTimestamp;
@@ -78,10 +88,11 @@ public class IndexScrutinyMapper extends Mapper<NullWritable, PhoenixIndexDBWrit
     private List<ColumnInfo> sourceTblColumnMetadata;
 
     // used to write results to the output table
-    private Connection outputConn;
-    private PreparedStatement outputUpsertStmt;
+    protected Connection outputConn;
+    protected PreparedStatement outputUpsertStmt;
     private long outputMaxRows;
     private MessageDigest md5;
+    private long ttl;
 
     @Override
     protected void setup(final Context context) throws IOException, InterruptedException {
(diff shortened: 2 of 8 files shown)


phoenix-core/src/main/java/org/apache/phoenix/mapreduce/index/IndexScrutinyMapper.java
```diff
@@ -254,6 +291,47 @@ private void processBatch(Context context)
...
+        }
+    }
+
+    private long getTableTtl() throws SQLException, IOException {
+        PTable psourceTable = PhoenixRuntime.getTable(connection, qSourceTable);
+        Admin admin = connection.unwrap(PhoenixConnection.class).getQueryServices().getAdmin();
+        String fullTableName = SchemaUtil.getQualifiedTableName(psourceTable.getSchemaName().toString(),
+                psourceTable.getTableName().toString());
+        HTableDescriptor tableDesc = admin.getTableDescriptor(TableName.valueOf(fullTableName));
+        long ttl = tableDesc.getFamily(SchemaUtil.getEmptyColumnFamily(psourceTable)).getTimeToLive();
```
priyankporwal: What is the TTL for tables that don't have it set explicitly? If it's LONG_MAX, then please make sure hasRowExpiredOnSource() does not overflow when doing sourceTS + ttl.

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/index/IndexScrutinyMapper.java
```diff
@@ -254,6 +291,47 @@ private void processBatch(Context context)
...
 
+    private void checkIfInvalidRowsExpired(Context context,
+            Map<String, Pair<Long,
+            List<Object>>> targetPkToSourceValues) throws SQLException, IOException {
+        long ttl = getTableTtl();
+        Set<Map.Entry<String, Pair<Long,
+                                List<Object>>>> entrySet = targetPkToSourceValues.entrySet();
+
+        Iterator<Map.Entry<String, Pair<Long,
+                        List<Object>>>> itr = entrySet.iterator();
```
priyankporwal: nit: Splitting of these templatized types across multiple lines is making it worse than long lines. Missing typedefs in Java :|

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/util/PhoenixConfigurationUtil.java
```diff
@@ -157,6 +156,8 @@
 
     public static final String MAPREDUCE_TENANT_ID = "phoenix.mapreduce.tenantid";
 
+    public static final String CUSTOM_TEST = "phoenix.mr.custom.test";
```
priyankporwal: Why is this phoenix.mr instead of phoenix.mapreduce?

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/index/IndexScrutinyMapper.java
```diff
@@ -228,8 +256,17 @@ private void processBatch(Context context)
                     buildTargetStatement(targetStatement);
 
             // fetch results from the target table and output invalid rows
+            if (isTest) {
+                // change the current time past ttl
+                ScrutinyTestClock clock = new ScrutinyTestClock(3600000);
```
priyankporwal: How does the isTest guarantee that TTL is 360000? Perhpas make a const that is used here as well as in the test that creates table with this TTL.

priyankporwal: Looks like Line 94 in the file above.

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/util/PhoenixConfigurationUtil.java
swaroopak: The majority of the properties start with "phoenix.mr" except for a couple of them as phoenix.mapreduce.

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/index/IndexScrutinyMapper.java
swaroopak: I ll extract it to a constant.

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/index/IndexScrutinyMapper.java
```diff
@@ -228,8 +257,17 @@ private void processBatch(Context context)
                     buildTargetStatement(targetStatement);
 
             // fetch results from the target table and output invalid rows
+            if (isTest) {
```
gjacoby126: Usually not a good idea to put test logic that directly in production code. Injection would be cleaner, but given the M/R framework reflectively instantiates the mapper, hard to do. One option would be to have a protected hook that a test-only subclass could override that will set up the clock, and have your test use the subclass.

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/index/IndexScrutinyMapper.java
```diff
@@ -254,6 +292,41 @@ private void processBatch(Context context)
...
+        long currentTS = EnvironmentEdgeManager.currentTimeMillis();
+        return ttl != Integer.MAX_VALUE && sourceTS + ttl*1000 < currentTS;
+    }
+
+    private long getTableTtl() throws SQLException, IOException {
+        PTable psourceTable = PhoenixRuntime.getTable(connection, qSourceTable);
+        Admin admin = connection.unwrap(PhoenixConnection.class).getQueryServices().getAdmin();
+        String fullTableName = SchemaUtil.getQualifiedTableName(psourceTable.getSchemaName().toString(),
+                psourceTable.getTableName().toString());
+        HTableDescriptor tableDesc = admin.getTableDescriptor(TableName.valueOf(fullTableName));
```
gjacoby126: If we have a lot of invalid rows, that's a lot of calls to getTableDescriptor, which sends an RPC to the HMaster. Is caching appropriate here? ("Weird stuff happens to your scrutiny run results if you change TTL during it" is an OK limitation, I think, since scrutiny mappers would probably fail during the TTL change anyway.)

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/util/PhoenixConfigurationUtil.java
```diff
@@ -290,6 +291,18 @@ public static void setOutputCluster(final Configuration configuration,
     public static Class<?> getInputClass(final Configuration configuration) {
         return configuration.getClass(INPUT_CLASS, NullDBWritable.class);
     }
+
+    public static boolean getCustomTest(final Configuration configuration) {
```
gjacoby126: Likewise, this is letting test details leak into production code a bit too much.

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/index/IndexScrutinyMapper.java
gjacoby126: @swaroopak  Hmm...turns out I also need a frequent TableDescriptor lookup over at PHOENIX-5645 to verify that scn doesn't overshoot the ttl. Wonder if there's a need for a shared cache, similar to the PTable cache.

swaroopak: In this case I can call getTableTtl in setup, isnt it?

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/index/IndexScrutinyMapper.java
```diff
@@ -307,15 +324,26 @@ private void queryTargetTable(Context context, PreparedStatement targetStatement
...
+                        outputInvalidRow(context, sourceValues, targetValues, sourceTS, targetTS);
+                    }
                 }
             }
             targetPkToSourceValues.remove(targetPk);
         }
     }
 
+    private boolean hasRowExpiredOnSource(Long sourceTS, Long ttl) {
+        long currentTS = System.currentTimeMillis();
```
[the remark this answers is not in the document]
swaroopak: This is an old version I guess.

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/index/IndexScrutinyTool.java
```diff
@@ -114,18 +115,27 @@
             "If specified, uses Tenant connection for tenant view index scrutiny (optional)");
     public static final String INDEX_JOB_NAME_TEMPLATE = "PHOENIX_SCRUTINY_[%s]_[%s]";
 
+    @Inject
+    Class<IndexScrutinyMapperForTest> mapperClass = null;
+
+    public IndexScrutinyTool(Class<IndexScrutinyMapperForTest> indexScrutinyMapperForTestClass) {
```
swaroopak: I want to use Class<?> instead of providing Test class . However, while invoking setMapperClass with this attribute, it complains that setMapperClass cant be applied to capture<?> Any other ideas? @gjacoby126

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/index/IndexScrutinyMapperForTest.java
```diff
@@ -0,0 +1,89 @@
...
+        }
+
+        @Override
+        public long currentTime() {
+            return System.currentTimeMillis() + delta;
+        }
+    }
+
+    @Override
+    protected void processBatch(Context context)
```
gjacoby126: I think the method you're overriding here is too broad -- you've copied much of the logic you're testing, which makes the test less useful. (For example, if someone changes IndexScrutinyMapper.processBatch and breaks it, any tests using this mapper class will never notice.) 

My suggestion is to have a very narrow hook that just lets you insert the logic that needs to be different, and use the "real" code for everything else. Think of how HBase coprocs work, where HBase is doing most of the work, but occasionally has hooks that can be overriden for extra / different behavior. 

For example, an overridable, no-op "beforeQueryTargetTable" method in IndexScrutinyMapper would let you override here with just your current lines 53 and 54 setting up the different clock values.

gjacoby126: Basically, that if-block in IndexScrutinyMapper that I objected to last time around -- that's pretty close to what your overridden method in the subclass would look like. :-)

phoenix-core/src/it/java/org/apache/phoenix/end2end/IndexScrutinyToolBaseIT.java
```diff
@@ -54,7 +55,7 @@
     }
 
     protected List<Job> runScrutiny(String[] cmdArgs) throws Exception {
-        IndexScrutinyTool scrutiny = new IndexScrutinyTool();
+        IndexScrutinyTool scrutiny = new IndexScrutinyTool(IndexScrutinyMapperForTest.class);
```
priyankporwal: I would suggest either of these approached 1. Just override hasRowExpiredOnSource() in IndexScrutinyMapperForTest class and do IndexScrutinyTool scrutiny = new IndexScrutinyMapperForTest(), OR 2. Use Callable<> to pass in a function implementation. I like the former approach.

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/index/IndexScrutinyMapperForTest.java
swaroopak: Thanks @gjacoby126 Your comments really helped me get more insights on various things on this PR. :)

phoenix-core/src/it/java/org/apache/phoenix/end2end/IndexScrutinyToolBaseIT.java
swaroopak: I made changes to override only a small part of the function. Please review.

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/index/IndexScrutinyMapper.java
```diff
@@ -54,22 +61,24 @@
...
 
+
 /**
  * Mapper that reads from the data table and checks the rows against the index table
  */
 public class IndexScrutinyMapper extends Mapper<NullWritable, PhoenixIndexDBWritable, Text, Text> {
 
     private static final Logger LOGGER = LoggerFactory.getLogger(IndexScrutinyMapper.class);
-    private Connection connection;
+    public static final int TEST_TABLE_TTL = 3600;
```
priyankporwal: Move this const into the test mapper class.

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/index/IndexScrutinyMapper.java
```diff
@@ -227,9 +236,14 @@ private void processBatch(Context context)
             Map<String, Pair<Long, List<Object>>> targetPkToSourceValues =
                     buildTargetStatement(targetStatement);
 
+            preQueryTargetTable();
```
priyankporwal: Nit: This fn implementation has nothing specific to target table. Also, the clock change applies to source table too. Is there any better name? e.g. preBatchScrutiny(), preScrutinyQueries()?

swaroopak: this clock change has to happen before the target table is queried and hence the name

PR #670, opened 2020-01-03
PHOENIX-5630: MAX_MUTATION_SIZE_EXCEEDED and MAX_MUTATION_SIZE_BYTES_EXCEEDED SQLExceptions should print existing size

nsgupta1:
Hi @ChinmaySKulkarni , @yanxinyi  can you please review this patch ?

### DIFF
--- a/phoenix-core/src/main/java/org/apache/phoenix/exception/SQLExceptionInfo.java
+++ b/phoenix-core/src/main/java/org/apache/phoenix/exception/SQLExceptionInfo.java
@@ -38,6 +38,10 @@ public class SQLExceptionInfo {
     public static final String FAMILY_NAME = "familyName";
     public static final String COLUMN_NAME = "columnName";
     public static final String FUNCTION_NAME = "functionName";
+    public static final String MAX_MUTATION_SIZE = "maxMutationSize";
+    public static final String MUTATION_SIZE = "mutationSize";
+    public static final String MAX_MUTATION_SIZE_BYTES = "maxMutationSizeBytes";
+    public static final String MUTATION_SIZE_BYTES = "mutationSizeBytes";
 
     private final Throwable rootCause;
     private final SQLExceptionCode code; // Should always have one.
@@ -47,6 +51,10 @@ public class SQLExceptionInfo {
     private final String familyName;
     private final String columnName;
     private final String functionName;
+    private final int maxMutationSize;
+    private final int mutationSize;
+    private final long maxMutationSizeBytes;
+    private final long mutationSizeBytes;
 
     public static class Builder {
 
@@ -58,6 +66,10 @@ public class SQLExceptionInfo {
         private String familyName;
         private String columnName;
         private String functionName;
+        private int maxMutationSize;
+        private int mutationSize;
+        private long maxMutationSizeBytes;
+        private long mutationSizeBytes;
 
         public Builder(SQLExceptionCode code) {
             this.code = code;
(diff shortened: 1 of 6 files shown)


phoenix-core/src/main/java/org/apache/phoenix/schema/MaxMutationSizeBytesExceededException.java
```diff
@@ -0,0 +1,31 @@
+package org.apache.phoenix.schema;
```
yanxinyi: please add the Apache license

phoenix-core/src/main/java/org/apache/phoenix/schema/MaxMutationSizeExceededException.java
```diff
@@ -0,0 +1,28 @@
+package org.apache.phoenix.schema;
```
yanxinyi: same as above, please add the Apache license

phoenix-core/src/main/java/org/apache/phoenix/exception/SQLExceptionInfo.java
```diff
@@ -47,6 +51,10 @@
     private final String familyName;
     private final String columnName;
     private final String functionName;
+    private final long maxMutationSize;
+    private final int mutationSize;
```
mihir6692: This should be "long" if variable maxMutationSize is also long or vice-a-versa

mihir6692: Same for Builder class too

yanxinyi: nice catch here @mihir6692

nsgupta1: Thanks @mihir6692 @yanxinyi I had kept it "long" but mutationSize(numRows) in MutationState class is an integer ([here](https://github.com/apache/phoenix/blob/356f4cd7d43bccb9538a5a2b94863b1c52cd9aad/phoenix-core/src/main/java/org/apache/phoenix/execute/MutationState.java#L139)), hence I changed it to int.

mihir6692: I think this is the relevant [code path ](https://github.com/apache/phoenix/blob/356f4cd7d43bccb9538a5a2b94863b1c52cd9aad/phoenix-core/src/main/java/org/apache/phoenix/execute/MutationState.java#L130). 

But based on this [QueryServicesOptions](https://github.com/apache/phoenix/blob/78cfb911a157fa74fd9868099e20df3eae6f62ba/phoenix-core/src/main/java/org/apache/phoenix/query/QueryServicesOptions.java#L147) Number of max rows in single batch are limited to int range.

What do you think? @yanxinyi @ChinmaySKulkarni

yanxinyi: since we have a `DEFAULT_MAX_MUTATION_SIZE = 500000`, it's ok to use `int` instead of `long` from my point of view. What do you think @ChinmaySKulkarni

ChinmaySKulkarni: @nsgupta1 I think it's fine to keep both to `int` too. But, then for consistency, also better to make the data type change in other relevant places too as @mihir6692 mentioned.

nsgupta1: Changes maxSize variable to int from long for consistency.

mihir6692:
lgtm +1 @nsgupta1

yanxinyi:
LGTM +1

better to create a new JIRA that making the data type changes in another relevant place for consistency.

nsgupta1:
Thanks @mihir6692 and @yanxinyi, I added those data type changes in the same patch DEFAULT_MAX_MUTATION_SIZE is used as an integer only in other places.

nsgupta1:
Hi @mihir6692 @ChinmaySKulkarni  can you please merge the changes and close this PR ?

PR #672, opened 2020-01-07
PHOENIX-5658 IndexTool to verify index rows inline

kadirozde:
IndexTool uses UngroupedAggregateRegionObserver and IndexRegionObserver coprocessors to build an index table on the server side. UngroupedAggregateRegionObserver scans a data table regions and reconstructs the put mutations for each scanned row and applies them on the data table with the REPLAY_INDEX_REBUILD_WRITES option. These mutations are then only used to rebuild index rows by IndexRegionObserver. This PR adds an inline verification feature to UngroupedAggregateRegionObserver which reads these index rows and verifies their content to make sure they are build correctly.

### DIFF
--- a/phoenix-core/src/it/java/org/apache/phoenix/end2end/IndexToolIT.java
+++ b/phoenix-core/src/it/java/org/apache/phoenix/end2end/IndexToolIT.java
@@ -42,11 +42,16 @@ import org.apache.hadoop.conf.Configuration;
 import org.apache.hadoop.hbase.Cell;
 import org.apache.hadoop.hbase.HTableDescriptor;
 import org.apache.hadoop.hbase.TableName;
-import org.apache.hadoop.hbase.client.*;
+import org.apache.hadoop.hbase.client.Admin;
+import org.apache.hadoop.hbase.client.Result;
+import org.apache.hadoop.hbase.client.ResultScanner;
+import org.apache.hadoop.hbase.client.Scan;
+import org.apache.hadoop.hbase.client.Table;
 import org.apache.hadoop.hbase.regionserver.HRegion;
 import org.apache.hadoop.hbase.regionserver.RegionScanner;
 import org.apache.hadoop.hbase.util.Bytes;
 import org.apache.hadoop.mapreduce.Job;
+import org.apache.phoenix.hbase.index.IndexRegionObserver;
 import org.apache.phoenix.jdbc.PhoenixConnection;
 import org.apache.phoenix.mapreduce.index.IndexTool;
 import org.apache.phoenix.query.ConnectionQueryServices;
@@ -366,6 +371,139 @@ public class IndexToolIT extends BaseUniqueNamesOwnClusterIT {
         }
     }
 
+    private Cell getErrorMessageFromIndexToolOutputTable(Connection conn, String dataTableFullName, String indexTableFullName)
+            throws Exception {
+        byte[] indexTableFullNameBytes = Bytes.toBytes(indexTableFullName);
+        byte[] dataTableFullNameBytes = Bytes.toBytes(dataTableFullName);
+        Table hIndexTable = conn.unwrap(PhoenixConnection.class).getQueryServices()
+                .getTable(IndexTool.OUTPUT_TABLE_NAME_BYTES);
+        Scan scan = new Scan();
+        ResultScanner scanner = hIndexTable.getScanner(scan);
+        boolean dataTableNameCheck = false;
+        boolean indexTableNameCheck = false;
+        Cell errorMessageCell = null;
+        Result result = scanner.next();
+        if (result != null) {
+            for (Cell cell : result.rawCells()) {
+                assertTrue(Bytes.compareTo(cell.getFamilyArray(), cell.getFamilyOffset(), cell.getFamilyLength(),
+                        IndexTool.OUTPUT_TABLE_COLUMN_FAMILY, 0,
+                        IndexTool.OUTPUT_TABLE_COLUMN_FAMILY.length) == 0);
+                if (Bytes.compareTo(cell.getQualifierArray(), cell.getQualifierOffset(), cell.getQualifierLength(),
+                        IndexTool.DATA_TABLE_NAME_BYTES, 0, IndexTool.DATA_TABLE_NAME_BYTES.length) == 0) {
+                    dataTableNameCheck = true;
+                    assertTrue(Bytes.compareTo(cell.getValueArray(), cell.getValueOffset(), cell.getValueLength(),
+                            dataTableFullNameBytes, 0, dataTableFullNameBytes.length) == 0);
+                } else if (Bytes.compareTo(cell.getQualifierArray(), cell.getQualifierOffset(), cell.getQualifierLength(),
+                        IndexTool.INDEX_TABLE_NAME_BYTES, 0, IndexTool.INDEX_TABLE_NAME_BYTES.length) == 0) {
+                    indexTableNameCheck = true;
+                    assertTrue(Bytes.compareTo(cell.getValueArray(), cell.getValueOffset(), cell.getValueLength(),
+                            indexTableFullNameBytes, 0, indexTableFullNameBytes.length) == 0);
+                } else if (Bytes.compareTo(cell.getQualifierArray(), cell.getQualifierOffset(), cell.getQualifierLength(),
+                        IndexTool.ERROR_MESSAGE_BYTES, 0, IndexTool.ERROR_MESSAGE_BYTES.length) == 0) {
+                    errorMessageCell = cell;
+                }
+            }
+        }
+        assertTrue(dataTableNameCheck && indexTableNameCheck && errorMessageCell != null);
+        return errorMessageCell;
+    }
+
+    @Test
+    public void testIndexToolVerifyOption() throws Exception {
+        // This test is for building non-transactional global indexes with direct api
+        if (localIndex || transactional || !directApi || useSnapshot) {
+            return;
+        }
+        Properties props = PropertiesUtil.deepCopy(TEST_PROPERTIES);
+        try (Connection conn = DriverManager.getConnection(getUrl(), props)) {
+            String schemaName = generateUniqueName();
+            String dataTableName = generateUniqueName();
+            String dataTableFullName = SchemaUtil.getTableName(schemaName, dataTableName);
+            String indexTableName = generateUniqueName();
+            String viewName = generateUniqueName();
+            String viewFullName = SchemaUtil.getTableName(schemaName, viewName);
+            conn.createStatement().execute("CREATE TABLE " + dataTableFullName
+                    + " (ID INTEGER NOT NULL PRIMARY KEY, NAME VARCHAR, ZIP INTEGER) "
+                    + tableDDLOptions);
+            conn.commit();
+            conn.createStatement().execute("CREATE VIEW " + viewFullName + " AS SELECT * FROM " + dataTableFullName);
+            conn.commit();
+            // Insert a row
+            conn.createStatement().execute("upsert into " + viewFullName + " values (1, 'Phoenix', 12345)");
+            conn.commit();
+            // Configure IndexRegionObserver to fail the first write phase. This should not
+            // lead to any change on index and thus the index verify during index rebuild should fail
+            IndexRegionObserver.setIgnoreIndexRebuildForTesting(true);
+            conn.createStatement().execute(String.format(
+                    "CREATE INDEX %s ON %s (NAME) INCLUDE (ZIP) ASYNC", indexTableName, viewFullName));
+            // Run the index MR job and verify that the index table rebuild fails
+            runIndexTool(directApi, useSnapshot, schemaName, viewName, indexTableName,
+                    null, -1, true, false);
+            // The index tool output table should report that there is a missing index row
+            Cell cell = getErrorMessageFromIndexToolOutputTable(conn, dataTableFullName, "_IDX_" + dataTableFullName);
+            byte[] expectedValueBytes = Bytes.toBytes("Missing index rows - Expected: 1 Actual: 0");
+            assertTrue(Bytes.compareTo(cell.getValueArray(), cell.getValueOffset(), cell.getValueLength(),
+                    expectedValueBytes, 0, expectedValueBytes.length) == 0);
+            IndexRegionObserver.setIgnoreIndexRebuildForTesting(false);
+            Admin admin = conn.unwrap(PhoenixConnection.class).getQueryServices().getAdmin();
+            TableName indexToolOutputTable = TableName.valueOf(IndexTool.OUTPUT_TABLE_NAME_BYTES);
+            admin.disableTable(indexToolOutputTable);
+            admin.deleteTable(indexToolOutputTable);
+        }
+    }
+
+    @Test
+    public void testIndexToolOnlyVerifyOption() throws Exception {
+        // This test is for building non-transactional global indexes with direct api
+        if (localIndex || transactional || !directApi || useSnapshot) {
+            return;
+        }
+        String schemaName = generateUniqueName();
+        String dataTableName = generateUniqueName();
+        String dataTableFullName = SchemaUtil.getTableName(schemaName, dataTableName);
+        String indexTableName = generateUniqueName();
+        String indexTableFullName = SchemaUtil.getTableName(schemaName, indexTableName);
+        Properties props = PropertiesUtil.deepCopy(TEST_PROPERTIES);
+        try (Connection conn = DriverManager.getConnection(getUrl(), props)) {
+            conn.createStatement().execute("CREATE TABLE " + dataTableFullName
+                    + " (ID INTEGER NOT NULL PRIMARY KEY, NAME VARCHAR, CODE VARCHAR) COLUMN_ENCODED_BYTES=0");
+            // Insert a row
+            conn.createStatement().execute("upsert into " + dataTableFullName + " values (1, 'Phoenix', 'A')");
+            conn.commit();
+            conn.createStatement().execute(String.format(
+                    "CREATE INDEX %s ON %s (NAME) INCLUDE (CODE) ASYNC", indexTableName, dataTableFullName));
+            // Run the index MR job to only verify that each data table row has a corresponding index row
+            // IndexTool will go through each data table row and record the mismatches in the output table
+            // called PHOENIX_INDEX_TOOL
+            runIndexTool(directApi, useSnapshot, schemaName, dataTableName, indexTableName,
+                    null, 0, false, true);
+            Cell cell = getErrorMessageFromIndexToolOutputTable(conn, dataTableFullName, indexTableFullName);
+            byte[] expectedValueBytes = Bytes.toBytes("Missing index rows - Expected: 1 Actual: 0");
+            assertTrue(Bytes.compareTo(cell.getValueArray(), cell.getValueOffset(), cell.getValueLength(),
+                    expectedValueBytes, 0, expectedValueBytes.length) == 0);
+            // Delete the output table for the next test
+            Admin admin = conn.unwrap(PhoenixConnection.class).getQueryServices().getAdmin();
+            TableName indexToolOutputTable = TableName.valueOf(IndexTool.OUTPUT_TABLE_NAME_BYTES);
+            admin.disableTable(indexToolOutputTable);
+            admin.deleteTable(indexToolOutputTable);
+            // Run the index tool to populate the index while verifying rows
+            runIndexTool(directApi, useSnapshot, schemaName, dataTableName, indexTableName,
+                    null, 0, true, false);
+            // Corrupt one cell by writing directly into the index table
+            conn.createStatement().execute("upsert into " + indexTableFullName + " values ('Phoenix', 1, 'B')");
+            conn.commit();
+            // Run the index tool using the only-verify option to detect this mismatch between the data and index table
+            runIndexTool(directApi, useSnapshot, schemaName, dataTableName, indexTableName,
+                    null, 0, false, true);
+            cell = getErrorMessageFromIndexToolOutputTable(conn, dataTableFullName, indexTableFullName);
+            expectedValueBytes = Bytes.toBytes("Not matching cell value - 0:0:CODE - Expected: A - Actual: B");
+            assertTrue(Bytes.compareTo(cell.getValueArray(), cell.getValueOffset(), cell.getValueLength(),
+                    expectedValueBytes, 0, expectedValueBytes.length) == 0);
+            admin.disableTable(indexToolOutputTable);
+            admin.deleteTable(indexToolOutputTable);
+        }
+    }
+
     @Test
     public void testIndexToolWithTenantId() throws Exception {
         if (!useTenantId) { return;}
@@ -640,9 +778,10 @@ public class IndexToolIT extends BaseUniqueNamesOwnClusterIT {
             actualExplainPlan.contains(expectedExplainPlan));
     }
 
-    public static String[] getArgValues(boolean directApi, boolean useSnapshot, String schemaName,
-            String dataTable, String indxTable, String tenantId) {
-        final List<String> args = Lists.newArrayList();
+    private static List<String> getArgList (boolean directApi, boolean useSnapshot, String schemaName,
+                                            String dataTable, String indxTable, String tenantId,
+                                            boolean verify, boolean onlyVerify) {
+        List<String> args = Lists.newArrayList();
         if (schemaName != null) {
             args.add("-s");
             args.add(schemaName);
@@ -653,6 +792,11 @@ public class IndexToolIT extends BaseUniqueNamesOwnClusterIT {
         args.add(indxTable);
         if (directApi) {
             args.add("-direct");
+        }
+        if (verify) {
+            args.add("-v"); // verify index rows inline
+        } else if (onlyVerify) {
+            args.add("-ov");
         }
         // Need to run this job in foreground for the test to be deterministic
         args.add("-runfg");
@@ -667,6 +811,13 @@ public class IndexToolIT extends BaseUniqueNamesOwnClusterIT {
 
         args.add("-op");
         args.add("/tmp/" + UUID.randomUUID().toString());
+        return args;
+    }
+
+    public static String[] getArgValues(boolean directApi, boolean useSnapshot, String schemaName,
+            String dataTable, String indexTable, String tenantId, boolean verify, boolean onlyVerify) {
+        List<String> args = getArgList(directApi, useSnapshot, schemaName, dataTable, indexTable, tenantId, verify, onlyVerify);
+
         return args.toArray(new String[0]);
     }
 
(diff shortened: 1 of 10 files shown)


phoenix-core/src/main/java/org/apache/phoenix/coprocessor/UngroupedAggregateRegionObserver.java
```diff
@@ -1291,6 +1470,38 @@ public boolean next(List<Cell> results) throws IOException {
             } finally {
                 region.closeRegionOperation();
             }
+            if (verify) {
+                ArrayList<KeyRange> keys = new ArrayList<>(rowCountPerTask);
+                for (byte[] key : dataPutMap.keySet()) {
+                    keys.add(PVarbinary.INSTANCE.getKeyRange(key));
```
gjacoby126: Is there a difference between KeyRange.getKeyRange() and PDataType.getKeyRange() as you're using here?

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/UngroupedAggregateRegionObserver.java
```diff
@@ -1189,40 +1237,168 @@ private Delete generateDeleteMarkers(List<Cell> row) {
...
+                        throw new IOException(exceptionMessage);
+                    }
+                    verifySingleIndexRow(result, dataPut);
+                    rowCount++;
+                }
+            } catch (Throwable t) {
+                ServerUtil.throwIOException(indexHTable.getName().toString(), t);
+            }
+            if (rowCount != expectedRowCount) {
+                exceptionMessage = "Index verify failed - Missing index rows Expected : " + expectedRowCount +
```
swaroopak: I was wondering if we can have a test for this expected message

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/UngroupedAggregateRegionObserver.java
```diff
@@ -1189,40 +1237,168 @@ private Delete generateDeleteMarkers(List<Cell> row) {
...
+                        + indexRow.rawCells().length + " cells - " + indexHTable.getName();
+                throw new IOException(exceptionMessage);
+            }
+        }
+
+        private void verifyIndexRows(ArrayList<KeyRange> keys) throws IOException {
+            int expectedRowCount = keys.size();
+            ScanRanges scanRanges = ScanRanges.createPointLookup(keys);
+            Scan indexScan = new Scan();
+            indexScan.setTimeRange(scan.getTimeRange().getMin(), scan.getTimeRange().getMax());
```
priyankporwal: What is the index scan time-range? The max should be index rebuild start time (or whatever timestamp index is getting built at), right?

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/UngroupedAggregateRegionObserver.java
```diff
@@ -1189,40 +1237,168 @@ private Delete generateDeleteMarkers(List<Cell> row) {
...
+            scanRanges.initializeScan(indexScan);
+            SkipScanFilter skipScanFilter = scanRanges.getSkipScanFilter();
+            indexScan.setFilter(skipScanFilter);
+            int rowCount = 0;
+            try (ResultScanner resultScanner = indexHTable.getScanner(indexScan)) {
+                for (Result result = resultScanner.next(); (result != null); result = resultScanner.next()) {
+                    Put dataPut = dataPutMap.get(result.getRow());
+                    if (dataPut == null) {
+                        exceptionMessage = "Index verify failed - Missing data row - " + indexHTable.getName();
```
priyankporwal: Can we log the index/data row key?

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/UngroupedAggregateRegionObserver.java
priyankporwal: Likewise, can we log the first & last data row key here?

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/UngroupedAggregateRegionObserver.java
```diff
@@ -1189,40 +1237,168 @@ private Delete generateDeleteMarkers(List<Cell> row) {
...
+        private boolean checkIndexRow(final byte[] indexRowKey, final Put put) throws IOException {
+            byte[] builtIndexRowKey = getIndexRowKey(put);
             if (Bytes.compareTo(builtIndexRowKey, 0, builtIndexRowKey.length,
                     indexRowKey, 0, indexRowKey.length) != 0) {
                 return false;
             }
             return true;
         }
 
+        private void verifySingleIndexRow(Result indexRow, final Put dataRow) throws IOException {
```
priyankporwal: @kadirozde Wouldn't this data/index row comparison code already exist in scrutiny or some other shared class?

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/UngroupedAggregateRegionObserver.java
```diff
@@ -1189,40 +1237,168 @@ private Delete generateDeleteMarkers(List<Cell> row) {
...
+                        verifyIndexRows(keys);
+                    } catch (Exception e) {
+                        throw e;
+                    }
+                    return Boolean.TRUE;
+                }
+            });
+        }
+
+        private void addToBeVerifiedIndexRows() throws IOException {
```
swaroopak: nit: do you think moving this method close to the caller makes it more readable?

phoenix-core/src/main/java/org/apache/phoenix/query/QueryServicesOptions.java
```diff
@@ -334,7 +334,7 @@
 
     public static final long DEFAULT_GLOBAL_INDEX_ROW_AGE_THRESHOLD_TO_DELETE_MS = 7*24*60*60*1000; /* 7 days */
     public static final boolean DEFAULT_INDEX_REGION_OBSERVER_ENABLED = true;
-    public static final long DEFAULT_INDEX_REBUILD_PAGE_SIZE_IN_ROWS = 16*1024;
+    public static final long DEFAULT_INDEX_REBUILD_PAGE_SIZE_IN_ROWS = 32*1024;
```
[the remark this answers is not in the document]
gokceni: +1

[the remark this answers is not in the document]
kadirozde: The increase is to have more rows for each skip scan to reduce the overall overhead of scan operations. The rebuild page size is the number of rows to be rebuilt at a time per data table region. Immediately after these rows are rebuilt on the server side, a number of scan threads start verifying them using skip scans again on the server side. Having more rows to be rebuilt at a time increases the number of rows per skip scan. This makes skip scan more effective.

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/UngroupedAggregateRegionObserver.java
```diff
@@ -1089,30 +1109,40 @@ private static PTable deserializeTable(byte[] b) {
...
         private IndexMaintainer indexMaintainer;
         private byte[] indexRowKey = null;
+        private Table indexHTable = null;
+        private boolean verify = false;
+        Map<byte[], Put> dataPutMap;
+        private TaskRunner pool;
+        TaskBatch<Boolean> tasks;
+        String exceptionMessage;
 
         IndexRebuildRegionScanner (final RegionScanner innerScanner, final Region region, final Scan scan,
```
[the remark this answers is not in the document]
kadirozde: I will make IndexRebuildRegionScanner a separate class.

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/UngroupedAggregateRegionObserver.java
```diff
@@ -1189,40 +1237,168 @@ private Delete generateDeleteMarkers(List<Cell> row) {
...
+            // Add the empty column
+            indexPut.addColumn(indexMaintainer.getEmptyKeyValueFamily().copyBytesIfNecessary(),
+                    indexMaintainer.getEmptyKeyValueQualifier(), ts, VERIFIED_BYTES);
+            int cellCount = 0;
+            for (List<Cell> cells : indexPut.getFamilyCellMap().values()) {
+                if (cells == null) {
+                    break;
+                }
+                for (Cell expectedCell : cells) {
+                    Cell actualCell = indexRow.getColumnLatestCell(CellUtil.cloneFamily(expectedCell),
```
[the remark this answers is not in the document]
kadirozde: I could not see how I can use caching here. I do not know the column name or column family name of the cell scanned from the index table row. The scan is set up to retrieve all the cells to make sure there is nothing more or less in every index row. Please note this check is done for the covered columns and empty column only.

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/UngroupedAggregateRegionObserver.java
```diff
@@ -1089,30 +1109,40 @@ private static PTable deserializeTable(byte[] b) {
...
         private byte[] indexMetaData;
         private boolean useProto = true;
         private Scan scan;
         private RegionScanner innerScanner;
         private Region region;
         private IndexMaintainer indexMaintainer;
         private byte[] indexRowKey = null;
+        private Table indexHTable = null;
+        private boolean verify = false;
+        Map<byte[], Put> dataPutMap;
```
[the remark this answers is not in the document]
kadirozde: dataPutMap is a map from an index row key to the corresponding data table row put mutation. Is it still misleading?

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/UngroupedAggregateRegionObserver.java
kadirozde: They are the same thing except that PDataType.getKeyRange() is a wrapper method of KeyRange.getKeyRange() and adjusts the key range to "
* Force lower bound to be inclusive for fixed width keys because it makes comparisons less expensive when you
* can count on one bound or the other being inclusive."

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/UngroupedAggregateRegionObserver.java
kadirozde: It is the start time.

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/UngroupedAggregateRegionObserver.java
kadirozde: No. The scrutiny tool uses Phoenix API. IndexTool uses HBase API.

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/UngroupedAggregateRegionObserver.java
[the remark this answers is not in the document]
gjacoby126: indexKeyToDataPutMap is a mouthful, but explains better what it is. Happy to hear alternatives.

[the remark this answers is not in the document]
kadirozde: I will rename it as you suggested

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/UngroupedAggregateRegionObserver.java
kadirozde: I will do that

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/index/IndexTool.java
```diff
@@ -145,6 +146,9 @@
     private static final Option DIRECT_API_OPTION = new Option("direct", "direct", false,
             "This parameter is deprecated. Direct mode will be used whether it is set or not. Keeping it for backwards compatibility.");
 
+    private static final Option VERIFY_OPTION = new Option("v", "verify", false,
+            "To verify every index row is rebuilt correctly. It is supported with the direct option");
```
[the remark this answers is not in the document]
kadirozde: ok

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/UngroupedAggregateRegionObserver.java
kadirozde: It is called in two places and it is closer to the second one. I will move closer to the first one.

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/index/IndexTool.java
```diff
@@ -113,13 +117,31 @@
...
+    public final static byte[] DATA_TABLE_NAME_BYTES = Bytes.toBytes(DATA_TABLE_NAME);
+    public static String INDEX_TABLE_NAME = "ITName";
+    public final static byte[] INDEX_TABLE_NAME_BYTES = Bytes.toBytes(INDEX_TABLE_NAME);
+    public static String DATA_TABLE_ROW_KEY = "DTRowKey";
+    public final static byte[] DATA_TABLE_ROW_KEY_BYTES = Bytes.toBytes(DATA_TABLE_ROW_KEY);
+    public static String INDEX_TABLE_ROW_KEY = "ITRowKey";
+    public final static byte[] INDEX_TABLE_ROW_KEY_BYTES = Bytes.toBytes(INDEX_TABLE_ROW_KEY);
+    public static String DATA_TABLE_TS = "DTTS";
+    public final static byte[] DATA_TABLE_TS_BYTES = Bytes.toBytes(DATA_TABLE_TS);
+    public static String INDEX_TABLE_TS = "TTTS";
```
priyankporwal: ITTS

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/IndexRebuildRegionScanner.java
```diff
@@ -0,0 +1,559 @@
...
+            put.addColumn(IndexTool.OUTPUT_TABLE_COLUMN_FAMILY, IndexTool.INDEX_TABLE_NAME_BYTES,
+                    scanMaxTs, indexMaintainer.getIndexTableName());
+            if (dataRowKey != null) {
+                put.addColumn(IndexTool.OUTPUT_TABLE_COLUMN_FAMILY, IndexTool.DATA_TABLE_ROW_KEY_BYTES,
+                        scanMaxTs, dataRowKey);
+                put.addColumn(IndexTool.OUTPUT_TABLE_COLUMN_FAMILY, IndexTool.DATA_TABLE_TS_BYTES,
+                        scanMaxTs, Bytes.toBytes(dataRowTs));
+            }
+            if (indexRowKey != null) {
+                put.addColumn(IndexTool.OUTPUT_TABLE_COLUMN_FAMILY, IndexTool.DATA_TABLE_ROW_KEY_BYTES,
```
priyankporwal: INDEX_TABLE_ROW_KEY_BYTES

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/IndexRebuildRegionScanner.java
```diff
@@ -0,0 +1,559 @@
...
+                        scanMaxTs, indexRowKey);
+                put.addColumn(IndexTool.OUTPUT_TABLE_COLUMN_FAMILY, IndexTool.DATA_TABLE_TS_BYTES,
```
priyankporwal: INDEX_TABLE_TS_BYTES

kadirozde: Yes, I will correct it

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/index/IndexTool.java
kadirozde: Yes, I will correct it

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/IndexRebuildRegionScanner.java
```diff
@@ -0,0 +1,583 @@
...
+            return false;
+        }
+        return true;
+    }
+
+    private void logToIndexToolOutputTable(byte[] dataRowKey, byte[] indexRowKey, long dataRowTs, long indexRowTs,
+                                           String errorMsg) {
+        try (Table hTable = ServerUtil.ConnectionFactory.getConnection(ServerUtil.ConnectionType.INDEX_WRITER_CONNECTION,
+                env).getTable(TableName.valueOf(IndexTool.OUTPUT_TABLE_NAME))) {
+            byte[] rowKey = new byte[Long.BYTES];
```
gjacoby126: Shouldn't the row key also have the table being rebuilt? What if rebuilds are going on simultaneously on two different tables? Unlikely they'd be exact same ms, but possible. 

Also, if the row key is just a timestamp, you'll hotspot (though hopefully this table will be low volume)

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/IndexRebuildRegionScanner.java
```diff
@@ -0,0 +1,583 @@
...
+        for (List<Cell> cells : indexPut.getFamilyCellMap().values()) {
+            if (cells == null) {
+                break;
+            }
+            for (Cell expectedCell : cells) {
+                byte[] family = CellUtil.cloneFamily(expectedCell);
+                byte[] qualifier = CellUtil.cloneQualifier(expectedCell);
+                Cell actualCell = indexRow.getColumnLatestCell(family, qualifier);
+                if (actualCell == null) {
+                    exceptionMessage = "Index verify failed - Missing cell " + indexHTable.getName();
```
gjacoby126: nit: some extract methods around here would make this easier to read. maybe one function for cell not found and another for checking all the columns?

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/PhoenixServerBuildIndexInputFormat.java
```diff
@@ -92,6 +94,12 @@ protected  QueryPlan getQueryPlan(final JobContext context, final Configuration
             try {
                 scan.setTimeRange(0, scn);
                 scan.setAttribute(BaseScannerRegionObserver.INDEX_REBUILD_PAGING, TRUE_BYTES);
+                if (getVerifyIndex(configuration)) {
+                    scan.setAttribute(BaseScannerRegionObserver.INDEX_REBUILD_VERIFY, TRUE_BYTES);
```
gjacoby126: what happens if both set?

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/index/IndexTool.java
```diff
@@ -510,7 +542,12 @@ private Job configureJobForServerBuildIndex() throws Exception {
 
             PhoenixConfigurationUtil.setIndexToolDataTableName(configuration, qDataTable);
             PhoenixConfigurationUtil.setIndexToolIndexTableName(configuration, qIndexTable);
-
+            if (verify) {
+                PhoenixConfigurationUtil.setVerifyIndex(configuration, true);
+            }
```
gjacoby126: what if both set?

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/index/IndexTool.java
```diff
@@ -573,6 +610,21 @@ public Job getJob() {
...
 
+    private void createIndexToolOutputTable(Connection connection) throws Exception {
+        ConnectionQueryServices queryServices = connection.unwrap(PhoenixConnection.class).getQueryServices();
+        Admin admin = queryServices.getAdmin();
+        if (admin.tableExists(TableName.valueOf(OUTPUT_TABLE_NAME))) {
+            return;
+        }
+        HTableDescriptor tableDescriptor = new
+                HTableDescriptor(TableName.valueOf(OUTPUT_TABLE_NAME));
+        tableDescriptor.setValue("DISABLE_TABLE_SOR", "true");
```
gjacoby126: Not a property that open source HBase or Phoenix recognizes.

phoenix-core/src/it/java/org/apache/phoenix/end2end/IndexToolIT.java
```diff
@@ -366,6 +371,140 @@ public void testBuildSecondaryIndexAndScrutinize() throws Exception {
...
+                    null, -1, true, false);
+            // The index tool output table should report that there is a missing index row
+            Cell cell = getErrorMessageFromIndexToolOutputTable(conn, dataTableFullName, "_IDX_" + dataTableFullName);
+            byte[] expectedValueBytes = Bytes.toBytes("Missing index rows - Expected: 1 Actual: 0");
+            assertTrue(Bytes.compareTo(cell.getValueArray(), cell.getValueOffset(), cell.getValueLength(),
+                    expectedValueBytes, 0, expectedValueBytes.length) == 0);
+            IndexRegionObserver.setIgnoreIndexRebuildForTesting(false);
+            Admin admin = conn.unwrap(PhoenixConnection.class).getQueryServices().getAdmin();
+            TableName indexToolOutputTable = TableName.valueOf(IndexTool.OUTPUT_TABLE_NAME_BYTES);
+            admin.disableTable(indexToolOutputTable);
```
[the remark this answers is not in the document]
kadirozde: Having an empty output table to start with simplifies the tests.

phoenix-core/src/it/java/org/apache/phoenix/end2end/IndexToolIT.java
```diff
@@ -654,6 +794,11 @@ public static void assertExplainPlan(boolean localIndex, String actualExplainPla
         if (directApi) {
             args.add("-direct");
         }
+        if (verify) {
```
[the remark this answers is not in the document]
kadirozde: I can add an assert here.

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/IndexRebuildRegionScanner.java
kadirozde: Yes, it is possible but very unlikely to get the same timestamp especially when IndexTool runs are not automated. I intentionally chose a very simple row key to make the table human readable and easy to query using the HBase shell. There will be one table row for each mismatch row. The excepted volume will be low. I suggest not to optimize this for very unlikely scenarios.

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/PhoenixServerBuildIndexInputFormat.java
kadirozde: Verify wins

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/index/IndexTool.java
kadirozde: if both "verify" and "only-verify" are set then as the help for the command says the only-verify option will be ignored: "To verify every data table row has the corresponding index row with the correct content (without building the index table). If the verify option is set then the only-verify option will be ignored");

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/index/IndexTool.java
kadirozde: I will remove it in the open source versions.

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/IndexRebuildRegionScanner.java
kadirozde: I thought about this again. Because of the "only-verify" option where the tool reports every mismatch row instead of just one, I need to make the row key unique. So, I am thinking about including the data row key in the in the row key for the output table.

phoenix-core/src/it/java/org/apache/phoenix/end2end/IndexToolIT.java
```diff
@@ -366,6 +371,140 @@ public void testBuildSecondaryIndexAndScrutinize() throws Exception {
...
+            // lead to any change on index and thus the index verify during index rebuild should fail
+            IndexRegionObserver.setIgnoreIndexRebuildForTesting(true);
+            conn.createStatement().execute(String.format(
+                    "CREATE INDEX %s ON %s (NAME) INCLUDE (ZIP) ASYNC", indexTableName, viewFullName));
+            // Run the index MR job and verify that the index table rebuild fails
+            runIndexTool(directApi, useSnapshot, schemaName, viewName, indexTableName,
+                    null, -1, true, false);
+            // The index tool output table should report that there is a missing index row
+            Cell cell = getErrorMessageFromIndexToolOutputTable(conn, dataTableFullName, "_IDX_" + dataTableFullName);
+            byte[] expectedValueBytes = Bytes.toBytes("Missing index rows - Expected: 1 Actual: 0");
```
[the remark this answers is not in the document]
kadirozde: I have a test for the mismatch on the column value in the following test for the only-verify option. It is almost impossible to generate missing columns or wrong column values with the verify option since index rows are corrected during rebuild

phoenix-core/src/it/java/org/apache/phoenix/end2end/IndexToolIT.java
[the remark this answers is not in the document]
kadirozde: After more thinking on this, I should not add an assert here as it is acceptable for the implementation to set both at this moment

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/IndexRebuildRegionScanner.java
```diff
@@ -285,17 +286,25 @@ private void logToIndexToolOutputTable(byte[] dataRowKey, byte[] indexRowKey, lo
...
         try (Table hTable = ServerUtil.ConnectionFactory.getConnection(ServerUtil.ConnectionType.INDEX_WRITER_CONNECTION,
                 env).getTable(TableName.valueOf(IndexTool.OUTPUT_TABLE_NAME))) {
-            byte[] rowKey = new byte[Long.BYTES];
-            Bytes.putLong(rowKey, 0, scan.getTimeRange().getMax());
+            byte[] rowKey;
+            if (dataRowKey != null) {
+                int length = Long.BYTES + dataRowKey.length;
+                rowKey = new byte[length];
+                Bytes.putLong(rowKey, 0, scan.getTimeRange().getMax());
+                for (int i = Long.BYTES, j = 0; i < length; i++, j++) {
```
priyankporwal: Nit: System.arraycopy() might perform better than byte-by-byte copy. I assume it's java equivalent of memcpy.

gjacoby126: See also Bytes.putBytes

kadirozde: Both suggestions work. I will go with Bytes.putBytes as it makes the code more readable.

PR #676, opened 2020-01-10
PHOENIX-5636: Improve the error message when client connects to server with higher major version

christinefeng:


christinefeng:
@ChinmaySKulkarni @dbwong @yanxinyi @abhishek-chouhan

phoenix-core/src/main/java/org/apache/phoenix/query/ConnectionQueryServicesImpl.java
```diff
@@ -1571,15 +1576,28 @@ public GetVersionResponse call(MetaDataService instance)
...
+            }
         }
+
         if (systemCatalogTimestamp < MIN_SYSTEM_TABLE_TIMESTAMP) {
             throw new UpgradeRequiredException(systemCatalogTimestamp);
         }
     }
 
+    private String displayClientandServerVersions(long serverJarVersion) {
+        String clientVersion = PHOENIX_MAJOR_VERSION + "." + PHOENIX_MINOR_VERSION + "." + PHOENIX_PATCH_NUMBER;
```
sguggilam: This is already present , you can probably just use it : https://github.com/apache/phoenix/blob/master/phoenix-core/src/main/java/org/apache/phoenix/coprocessor/MetaDataProtocol.java#L151

phoenix-core/src/main/java/org/apache/phoenix/query/ConnectionQueryServicesImpl.java
```diff
@@ -1571,15 +1576,28 @@ public GetVersionResponse call(MetaDataService instance)
                 }
             }
         }
-        if (isIncompatible) {
-            buf.setLength(buf.length()-1);
-            throw new SQLExceptionInfo.Builder(SQLExceptionCode.OUTDATED_JARS).setMessage(buf.toString()).build().buildException();
+        if (!(errorCode.toString().equals(new StringBuilder().toString()))) {
```
sguggilam: is this doing an empty string comparision ?

phoenix-core/src/main/java/org/apache/phoenix/util/MetaDataUtil.java
```diff
@@ -122,22 +123,29 @@
...
     // Default scope for testing
-    static boolean areClientAndServerCompatible(int serverVersion, int clientMajorVersion, int clientMinorVersion) {
+    static boolean areClientAndServerCompatible(int serverVersion, int clientMajorVersion, int clientMinorVersion, StringBuilder buf) {
         // A server and client with the same major and minor version number must be compatible.
         // So it's important that we roll the PHOENIX_MAJOR_VERSION or PHOENIX_MINOR_VERSION
         // when we make an incompatible change.
-        return VersionUtil.encodeMinPatchVersion(clientMajorVersion, clientMinorVersion) <= serverVersion && // Minor major and minor cannot be ahead of server
-                VersionUtil.encodeMaxMinorVersion(clientMajorVersion) >= serverVersion; // Major version must at least be up to server version
+        boolean isCompatible = true;
+        if (!(VersionUtil.encodeMinPatchVersion(clientMajorVersion, clientMinorVersion) <= serverVersion)) {
```
ChinmaySKulkarni: nit: simplify this condition to `>`

phoenix-core/src/main/java/org/apache/phoenix/util/MetaDataUtil.java
```diff
@@ -122,22 +123,29 @@
...
+            isCompatible = false;
+            buf.append(SQLExceptionCode.OUTDATED_JARS.getErrorCode());
+        } else if (!(VersionUtil.encodeMaxMinorVersion(clientMajorVersion) >= serverVersion)) {
```
ChinmaySKulkarni: same here. Change `!>=` to `<`

phoenix-core/src/main/java/org/apache/phoenix/query/ConnectionQueryServicesImpl.java
```diff
@@ -1487,20 +1487,17 @@ private static boolean hasIndexWALCodec(Long serverVersion) {
         return MetaDataUtil.decodeHasIndexWALCodec(serverVersion);
     }
 
-    private static boolean isCompatible(Long serverVersion) {
+    private static boolean isCompatible(Long serverVersion, StringBuilder buf) {
```
ChinmaySKulkarni: This might be out of scope of your JIRA, but it's a potentially small change. The method `isCompatible` looks pretty useless to me. Can't all the logic be moved down to `MetaDataUtil.areClientAndServerCompatible`?

phoenix-core/src/main/java/org/apache/phoenix/query/ConnectionQueryServicesImpl.java
```diff
@@ -1487,20 +1487,17 @@ private static boolean hasIndexWALCodec(Long serverVersion) {
...
         if (serverVersion == null) {
             return false;
         }
-        return MetaDataUtil.areClientAndServerCompatible(serverVersion);
+        return MetaDataUtil.areClientAndServerCompatible(serverVersion, buf);
```
ChinmaySKulkarni: I'm also not a big fan of passing the StringBuilder param everywhere. It might be just cleaner to have a small inner class to represent clientServerCompatibility and which contains perhaps the following:
- client version
- server version
- an is compatible check based on the above 2 values
- error message and/or code if there is an incompatibility
- getters/setters of error message/code and versions

That also makes it obvious what the class does and what the buffer contains.
It shouldn't be too much work to add this small inner class. At least, we should rename "buf" to "errorString" or something.

phoenix-core/src/main/java/org/apache/phoenix/query/ConnectionQueryServicesImpl.java
```diff
@@ -1588,15 +1593,27 @@ public GetVersionResponse call(MetaDataService instance)
...
+                throw new SQLExceptionInfo.Builder(SQLExceptionCode.INCOMPATIBLE_CLIENT_SERVER_JAR).setMessage(errorMessage.toString()).build().buildException();
+            }
         }
+
         if (systemCatalogTimestamp < MIN_SYSTEM_TABLE_TIMESTAMP) {
             throw new UpgradeRequiredException(systemCatalogTimestamp);
         }
     }
 
+    private String displayClientandServerVersions(long serverJarVersion) {
```
ChinmaySKulkarni: nit: Rename method to "getClientAndServerVersions" since we're not really printing to stdout or logging the versions for it to be "display<xyz>".

phoenix-core/src/main/java/org/apache/phoenix/query/ConnectionQueryServicesImpl.java
```diff
@@ -1588,15 +1596,19 @@ public GetVersionResponse call(MetaDataService instance)
...
-            buf.setLength(buf.length()-1);
-            throw new SQLExceptionInfo.Builder(SQLExceptionCode.OUTDATED_JARS).setMessage(buf.toString()).build().buildException();
-        }
+
         if (systemCatalogTimestamp < MIN_SYSTEM_TABLE_TIMESTAMP) {
             throw new UpgradeRequiredException(systemCatalogTimestamp);
         }
     }
 
+    private String getClientandServerVersions(long serverJarVersion) {
```
ChinmaySKulkarni: I guess this method can just be `getServerVersion` right? Since the client version is basically `MetaDataProtocol.CURRENT_CLIENT_VERSION`. That way, you're also not making it mandatory for this info to be returned as a string "Client version ... ".

phoenix-core/src/main/java/org/apache/phoenix/util/MetaDataUtil.java
```diff
@@ -122,7 +123,38 @@
...
-    public static boolean areClientAndServerCompatible(long serverHBaseAndPhoenixVersion) {
+    public static class ClientServerCompatibility {
+
+        long serverVersion;
+        StringBuilder errorCode;
+        boolean isCompatible;
+
+        ClientServerCompatibility(long serverVersion) {
+            this.serverVersion = serverVersion;
+            this.errorCode = new StringBuilder();
```
ChinmaySKulkarni: Why is `errorCode` a StringBuilder? SQLExceptionCode error code types are integers.

phoenix-core/src/main/java/org/apache/phoenix/util/MetaDataUtil.java
```diff
@@ -122,7 +123,38 @@
...
+            this.isCompatible = true;
+        }
+
+        public int getErrorCode() {
+            int errorCode;
+            try {
+                errorCode = Integer.parseInt(this.errorCode.toString());
```
ChinmaySKulkarni: I don't understand why we are converting this to string. Do we need to store the error code as a string for some reason?

phoenix-core/src/main/java/org/apache/phoenix/util/MetaDataUtil.java
```diff
@@ -132,12 +164,20 @@ public static boolean areClientAndServerCompatible(long serverHBaseAndPhoenixVer
     }
 
     // Default scope for testing
```
ChinmaySKulkarni: nit: Add `@VisibleForTesting` annotation here.

phoenix-core/src/main/java/org/apache/phoenix/query/ConnectionQueryServicesImpl.java
```diff
@@ -1553,11 +1543,20 @@ public GetVersionResponse call(MetaDataService instance)
...
+                    if (compatibility.getErrorCode() == SQLExceptionCode.OUTDATED_JARS.getErrorCode()) {
+                        HRegionLocation name = regionMap.get(result.getKey());
+                        errorMessage.append("Newer Phoenix clients can't communicate with older "
+                                + "Phoenix servers. " + getClientandServerVersions(serverJarVersion)
+                                + " The following servers require an updated "
+                                + QueryConstants.DEFAULT_COPROCESS_JAR_NAME
+                                + " to be put in the classpath of HBase: ");
+                        errorMessage.append(name);
+                        errorMessage.append(';');
+                    } else {
```
ChinmaySKulkarni: Make this more specific with an `else if` rather than an `else`? Otherwise this breaks if tomorrow we have other cases for incompatibility.

phoenix-core/src/main/java/org/apache/phoenix/query/ConnectionQueryServicesImpl.java
```diff
@@ -1570,6 +1575,15 @@ public GetVersionResponse call(MetaDataService instance)
                     systemCatalogTimestamp = systemCatalogTimestamp < versionResponse.getSystemCatalogTimestamp() ?
                       systemCatalogTimestamp: versionResponse.getSystemCatalogTimestamp();
                 }
+
+                if (compatibility.getErrorCode() != 0) {
+                    if (compatibility.getErrorCode() == SQLExceptionCode.OUTDATED_JARS.getErrorCode()) {
+                        errorMessage.setLength(errorMessage.length()-1);
```
ChinmaySKulkarni: Do we need to setLength on the StringBuilder object? We can just directly call toString right?

phoenix-core/src/main/java/org/apache/phoenix/util/MetaDataUtil.java
```diff
@@ -122,7 +124,36 @@
...
+    public static class ClientServerCompatibility {
+
+        long serverVersion;
+        int errorCode;
+        boolean isCompatible;
+
+        ClientServerCompatibility(long serverVersion) {
+            this.serverVersion = serverVersion;
+            this.errorCode = 0;
+            this.isCompatible = true;
```
ChinmaySKulkarni: Why is this true by default? I think it's safer for it to default to false (you don't have to assign it since primitive boolean defaults to false).

phoenix-core/src/main/java/org/apache/phoenix/util/MetaDataUtil.java
```diff
@@ -122,7 +124,36 @@
             ColumnFamilyDescriptorBuilder.KEEP_DELETED_CELLS,
             ColumnFamilyDescriptorBuilder.REPLICATION_SCOPE);
 
-    public static boolean areClientAndServerCompatible(long serverHBaseAndPhoenixVersion) {
+    public static class ClientServerCompatibility {
+
+        long serverVersion;
+        int errorCode;
+        boolean isCompatible;
```
ChinmaySKulkarni: nit: make all 3 private and serverVersion can also be final

phoenix-core/src/main/java/org/apache/phoenix/util/MetaDataUtil.java
```diff
@@ -132,12 +163,21 @@ public static boolean areClientAndServerCompatible(long serverHBaseAndPhoenixVer
...
-    static boolean areClientAndServerCompatible(int serverVersion, int clientMajorVersion, int clientMinorVersion) {
+    @VisibleForTesting
+    static ClientServerCompatibility areClientAndServerCompatible(int serverVersion, int clientMajorVersion, int clientMinorVersion) {
         // A server and client with the same major and minor version number must be compatible.
         // So it's important that we roll the PHOENIX_MAJOR_VERSION or PHOENIX_MINOR_VERSION
         // when we make an incompatible change.
-        return VersionUtil.encodeMinPatchVersion(clientMajorVersion, clientMinorVersion) <= serverVersion && // Minor major and minor cannot be ahead of server
-                VersionUtil.encodeMaxMinorVersion(clientMajorVersion) >= serverVersion; // Major version must at least be up to server version
+        ClientServerCompatibility compatibility = new ClientServerCompatibility(serverVersion);
+        if (VersionUtil.encodeMinPatchVersion(clientMajorVersion, clientMinorVersion) > serverVersion) {
```
ChinmaySKulkarni: You've removed the original comments here.

phoenix-core/src/main/java/org/apache/phoenix/util/MetaDataUtil.java
```diff
@@ -132,12 +163,21 @@ public static boolean areClientAndServerCompatible(long serverHBaseAndPhoenixVer
...
+            compatibility.setCompatible(false);
+            compatibility.setErrorCode(SQLExceptionCode.OUTDATED_JARS.getErrorCode());
+        } else if (VersionUtil.encodeMaxMinorVersion(clientMajorVersion) < serverVersion) {
+            compatibility.setCompatible(false);
+            compatibility.setErrorCode(SQLExceptionCode.INCOMPATIBLE_CLIENT_SERVER_JAR.getErrorCode());
+        }
+        return compatibility; // Major version must at least be up to server version
```
ChinmaySKulkarni: nit: Move this comment back to where it was near the if condition

phoenix-core/src/main/java/org/apache/phoenix/query/ConnectionQueryServicesImpl.java
christinefeng: The setLength(length-1) gets rid of a trailing semicolon if there are multiple outdated jars error messages strung together (for loop at line 1540)

ChinmaySKulkarni: If there aren't multiple messages, and the loop runs just once, does it print the correct thing still or leave out the last character?

phoenix-core/src/main/java/org/apache/phoenix/util/MetaDataUtil.java
```diff
@@ -132,12 +160,22 @@ public static boolean areClientAndServerCompatible(long serverHBaseAndPhoenixVer
...
         // when we make an incompatible change.
-        return VersionUtil.encodeMinPatchVersion(clientMajorVersion, clientMinorVersion) <= serverVersion && // Minor major and minor cannot be ahead of server
-                VersionUtil.encodeMaxMinorVersion(clientMajorVersion) >= serverVersion; // Major version must at least be up to server version
+        ClientServerCompatibility compatibility = new ClientServerCompatibility();
+        if (VersionUtil.encodeMinPatchVersion(clientMajorVersion, clientMinorVersion) > serverVersion) { // Client major and minor cannot be ahead of server
+            compatibility.setErrorCode(SQLExceptionCode.OUTDATED_JARS.getErrorCode());
+            return compatibility;
+        } else if (VersionUtil.encodeMaxMinorVersion(clientMajorVersion) < serverVersion) { // Client major version must at least be up to server major version
+            compatibility.setErrorCode(SQLExceptionCode.INCOMPATIBLE_CLIENT_SERVER_JAR.getErrorCode());
+            return compatibility;
```
ChinmaySKulkarni: I know we said that we don't need to set boolean default value to false when initializing compatibility variable, but here it should be set before returning for clarity.

christinefeng: makes sense !

phoenix-core/src/main/java/org/apache/phoenix/query/ConnectionQueryServicesImpl.java
ChinmaySKulkarni: @christinefeng  ping on this one^

christinefeng: If it runs just once, there's still that trailing semicolon from line 1559, so the set length-1 gets rid of that semicolon
Not sure if I should add something similar to the incompatible_client_server_jar error message in case of multiple messages?

ChinmaySKulkarni: That's more for cases when we have SYSCAT on more than 1 region. Your change should be fine.

PR #678, opened 2020-01-13
PHOENIX-5644 and PHOENIX-5651 addendum patch

swaroopak:


phoenix-core/src/main/java/org/apache/phoenix/mapreduce/index/IndexUpgradeTool.java
```diff
@@ -424,20 +424,21 @@ private void enableImmutableTables(ConnectionQueryServices queryServices,
...
         long endWaitTime = EnvironmentEdgeManager.currentTimeMillis();
         long waitMore = getWaitMoreTime(endWaitTime, startWaitTime);
-        while (waitMore>0) {
+        while (waitMore>0 && !isWaitComplete) {
             // If the table is immutable, we need to wait for clients to purge
             // their caches of table metadata
             LOGGER.info("waiting for more " + waitMore + " ms for client cache "
                     + "to expire for immutable tables");
             try {
                 startWaitTime = EnvironmentEdgeManager.currentTimeMillis();
```
priyankporwal: Why change startWaitTime here?

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/index/IndexUpgradeTool.java
```diff
@@ -424,20 +424,21 @@ private void enableImmutableTables(ConnectionQueryServices queryServices,
...
                 Thread.sleep(waitMore);
-                waited = true;
+                isWaitComplete = true;
             } catch (InterruptedException e) {
                 endWaitTime = EnvironmentEdgeManager.currentTimeMillis();
```
priyankporwal: @swaroopak This logic of splitting endWaitTime and startWaitTime and two booleans is getting more bug prone now. 
I would suggest the below reorg
while(true) { 
long endWaitTime = EnvironmentEdgeManager.currentTimeMillis();
long waitMore = getWaitMoreTime(endWaitTime, startWaitTime);
if (waitMore <= 0) 
{
isWaitComplete = true; break;
}
try {
Thread.sleep(waitMore);
isWaitComplete = true;
}
catch((InterruptedException e) {
LOGGER.warning("Sleep before starting index rebuild is interrupted. "
                        + "Attempting to sleep again! " + e.getMessage());
}
}

swaroopak: This is exactly what I was doing before considering @gjacoby126 's suggestion to not ignore sleep interruptions and avoid cache incoherency.

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/index/IndexUpgradeTool.java
swaroopak: in case the sleep is interrupted.

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/index/IndexUpgradeTool.java
swaroopak: I think this patch is good to overcome the problem.

priyankporwal: In fact, remove endWaitTime entirely and let getWaitMoreTime() use EnvironmentEdgeManager.currentTimeMillis() directly.

swaroopak: I personally don't like while (true) :(

priyankporwal: Alternative to appease your dislike:  
do { ... } while(waitMore > 0) by pulling out waitMore out of the body

PR #679, opened 2020-01-13
PHOENIX-5645 - BaseScannerRegionObserver should prevent compaction from purg…

gjacoby126:
…ing very recently deleted cells (addendum)

### DIFF
--- a/phoenix-core/src/it/java/org/apache/phoenix/end2end/MaxLookbackIT.java
+++ b/phoenix-core/src/it/java/org/apache/phoenix/end2end/MaxLookbackIT.java
@@ -17,7 +17,6 @@
  */
 package org.apache.phoenix.end2end;
 
-
 import com.google.common.collect.Maps;
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
@@ -28,11 +27,11 @@ import org.apache.hadoop.hbase.protobuf.generated.AdminProtos;
 import org.apache.hadoop.hbase.regionserver.HRegion;
 import org.apache.hadoop.hbase.regionserver.ScanInfoUtil;
 import org.apache.hadoop.hbase.util.Bytes;
-import org.apache.hadoop.hbase.util.EnvironmentEdgeManager;
 import org.apache.phoenix.exception.SQLExceptionCode;
 import org.apache.phoenix.query.QueryServices;
+import org.apache.phoenix.util.EnvironmentEdge;
+import org.apache.phoenix.util.EnvironmentEdgeManager;
 import org.apache.phoenix.util.PhoenixRuntime;
-import org.apache.phoenix.util.QueryUtil;
 import org.apache.phoenix.util.ReadOnlyProps;
 import org.apache.phoenix.util.TestUtil;
 import org.junit.Assert;
@@ -43,7 +42,6 @@ import org.junit.Test;
 import java.io.IOException;
 import java.sql.Connection;
 import java.sql.DriverManager;
-import java.sql.ResultSet;
 import java.sql.SQLException;
 import java.sql.Statement;
 import java.util.Map;
@@ -59,10 +57,31 @@ import static org.apache.phoenix.util.TestUtil.assertTableHasVersions;
 @NeedsOwnMiniClusterTest
 public class MaxLookbackIT extends BaseUniqueNamesOwnClusterIT {
     private static final Log LOG = LogFactory.getLog(MaxLookbackIT.class);
-    private static final int MAX_LOOKBACK_AGE = 10;
+    private static final int MAX_LOOKBACK_AGE = 15;
     private static final int ROWS_POPULATED = 2;
+    public static final int WAIT_AFTER_TABLE_CREATION = 600000;
     private String tableDDLOptions;
     private StringBuilder optionBuilder;
+    ManualEnvironmentEdge injectEdge;
+    private int ttl;
+
+    private class ManualEnvironmentEdge extends EnvironmentEdge {
+        // Sometimes 0 ts might have a special value, so lets start with 1
+        protected long value = 1L;
+
+        public void setValue(long newValue) {
+            value = newValue;
+        }
+
+        public void incValue(long addedValue) {
+            value += addedValue;
+        }
+
+        @Override
+        public long currentTime() {
+            return this.value;
+        }
+    }
 
     @BeforeClass
     public static synchronized void doSetup() throws Exception {
@@ -74,29 +93,34 @@ public class MaxLookbackIT extends BaseUniqueNamesOwnClusterIT {
 
     @Before
     public void beforeTest(){
+        EnvironmentEdgeManager.reset();
         optionBuilder = new StringBuilder();
         this.tableDDLOptions = optionBuilder.toString();
+        ttl = 0;
+        injectEdge = new ManualEnvironmentEdge();
+        injectEdge.setValue(EnvironmentEdgeManager.currentTimeMillis());
+        EnvironmentEdgeManager.injectEdge(injectEdge);
     }
 
     @Test
     public void testTooLowSCNWithMaxLookbackAge() throws Exception {
-        try (Connection conn = DriverManager.getConnection(getUrl())) {
-            String dataTableName = generateUniqueName();
-            String indexStem = generateUniqueName();
-            createTableAndIndexes(conn, dataTableName, indexStem);
-            //need to sleep long enough for the SCN to still find the syscat row for the table
-            Thread.sleep(MAX_LOOKBACK_AGE * 1000 + 1000);
-            Properties props = new Properties();
-            props.setProperty(PhoenixRuntime.CURRENT_SCN_ATTRIB,
-                Long.toString(EnvironmentEdgeManager.currentTime() - (MAX_LOOKBACK_AGE + 1) * 1000));
-            try (Connection connscn = DriverManager.getConnection(getUrl(), props)) {
-                connscn.createStatement().executeQuery("select * from " + dataTableName);
-            } catch (SQLException se) {
-                SQLExceptionCode code =
-                    SQLExceptionCode.CANNOT_QUERY_TABLE_WITH_SCN_OLDER_THAN_MAX_LOOKBACK_AGE;
-                TestUtil.assertSqlExceptionCode(code, se);
-                return;
-            }
+        String dataTableName = generateUniqueName();
+        createTable(dataTableName);
+        //increase long enough to make sure we can find the syscat row for the table
+        injectEdge.incValue(WAIT_AFTER_TABLE_CREATION);
+        populateTable(dataTableName);
+        long populateTime = EnvironmentEdgeManager.currentTimeMillis();
+        injectEdge.incValue(MAX_LOOKBACK_AGE * 1000 + 1000);
+        Properties props = new Properties();
+        props.setProperty(PhoenixRuntime.CURRENT_SCN_ATTRIB,
+            Long.toString(populateTime));
+        try (Connection connscn = DriverManager.getConnection(getUrl(), props)) {
+            connscn.createStatement().executeQuery("select * from " + dataTableName);
+        } catch (SQLException se) {
+            SQLExceptionCode code =
+                SQLExceptionCode.CANNOT_QUERY_TABLE_WITH_SCN_OLDER_THAN_MAX_LOOKBACK_AGE;
+            TestUtil.assertSqlExceptionCode(code, se);
+            return;
         }
         Assert.fail("We should have thrown an exception for the too-early SCN");
     }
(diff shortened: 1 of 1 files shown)


phoenix-core/src/it/java/org/apache/phoenix/end2end/MaxLookbackIT.java
```diff
@@ -59,10 +57,31 @@
 @NeedsOwnMiniClusterTest
 public class MaxLookbackIT extends BaseUniqueNamesOwnClusterIT {
     private static final Log LOG = LogFactory.getLog(MaxLookbackIT.class);
-    private static final int MAX_LOOKBACK_AGE = 10;
+    private static final int MAX_LOOKBACK_AGE = 15;
     private static final int ROWS_POPULATED = 2;
+    public static final int WAIT_AFTER_TABLE_CREATION = 600000;
```
gokceni: Can we write the unit as part of const name?

phoenix-core/src/it/java/org/apache/phoenix/end2end/MaxLookbackIT.java
```diff
@@ -168,36 +184,48 @@ public void testTTLAndMaxLookbackAge() throws Exception {
...
             assertTableHasTtl(conn, dataTable, ttl);
-            String fullIndexName = indexStem + "1";
-            TableName indexTable = TableName.valueOf(fullIndexName);
-            assertTableHasTtl(conn, indexTable, ttl);
-
             //first make sure we inserted correctly
-            String sql = String.format("SELECT val2 FROM %s WHERE val1 = 'ab'", dataTableName);
-            assertExplainPlan(conn, sql, dataTableName, fullIndexName);
+            String sql = String.format("SELECT val2 FROM %s WHERE id = 'a'", dataTableName);
+          //  assertExplainPlan(conn, sql, dataTableName, fullIndexName);
```
gokceni: nit: remove commented code

gjacoby126:
Some notes:
1. Everything now uses the EnvironmentEdgeManager rather than Thread.sleeps to fix the non-deterministic behavior of the original version
2. The TTL test now has a longer TTL than max lookback age, which is correct (the original test was written before we decided that   scn age <= max lookback age < ttl
3. All tests pass in < 30s on my laptop, when before most of the tests took 30s each. 
4. I removed the index assertions; they were making things more complicated and since this functionality is part of all Phoenix tables now, didn't seem to have much benefit.

phoenix-core/src/it/java/org/apache/phoenix/end2end/MaxLookbackIT.java
```diff
@@ -168,36 +184,48 @@ public void testTTLAndMaxLookbackAge() throws Exception {
...
             int originalRowCount = 2;
-            assertRawRowCount(conn, indexTable, originalRowCount);
+            assertRawRowCount(conn, dataTable, originalRowCount);
             //force a flush
-            flush(indexTable);
+            flush(dataTable);
             //flush shouldn't have changed it
-            assertRawRowCount(conn, indexTable, originalRowCount);
+            assertRawRowCount(conn, dataTable, originalRowCount);
+                      // assertExplainPlan(conn, sql, dataTableName, fullIndexName);
```
gokceni: nit: remove commented code

phoenix-core/src/it/java/org/apache/phoenix/end2end/MaxLookbackIT.java
```diff
@@ -168,36 +184,48 @@ public void testTTLAndMaxLookbackAge() throws Exception {
...
+            long timeToSleep = (MAX_LOOKBACK_AGE * 1000) -
+                (EnvironmentEdgeManager.currentTimeMillis() - afterFirstInsertSCN);
+            if (timeToSleep > 0) {
+                injectEdge.incValue(timeToSleep);
+                //Thread.sleep(timeToSleep);
```
gokceni: nit: remove commented code

gokceni:
Couple of nits. Don't we want  a test on index tables as well?

phoenix-core/src/it/java/org/apache/phoenix/end2end/MaxLookbackIT.java
```diff
@@ -105,60 +129,52 @@ public void testTooLowSCNWithMaxLookbackAge() throws Exception {
...
             //deleted row should be gone, but not deleted row should still be there.
             assertRowExistsAtSCN(getUrl(), sql, beforeSecondCompactSCN, false);
             assertRowExistsAtSCN(getUrl(), notDeletedRowSql, beforeSecondCompactSCN, true);
             //1 deleted row should be gone
-            assertRawRowCount(conn, indexTable, ROWS_POPULATED);
+            assertRawRowCount(conn, dataTable, ROWS_POPULATED);
         }
     }
 
     @Test(timeout=60000L)
```
swaroopak: does this value change now?

phoenix-core/src/it/java/org/apache/phoenix/end2end/MaxLookbackIT.java
```diff
@@ -168,36 +184,48 @@ public void testTTLAndMaxLookbackAge() throws Exception {
...
-            long afterTTLExpiresSCN = org.apache.phoenix.util.EnvironmentEdgeManager.currentTimeMillis();
-            assertExplainPlan(conn, sql, dataTableName, fullIndexName);
-            //make sure we can't see it after expiration from masking
-            assertRowExistsAtSCN(getUrl(), sql, afterTTLExpiresSCN, false);
-            //but it's still on disk
-            assertRawRowCount(conn, indexTable, originalRowCount);
-            long beforeMajorCompactSCN = org.apache.phoenix.util.EnvironmentEdgeManager.currentTimeMillis();
-            majorCompact(indexTable, beforeMajorCompactSCN);
-            assertRawRowCount(conn, indexTable, 0);
+            timeToSleep = (ttl * 1000) -
```
kadirozde: nit : timeToAdvance sounds better than timeToSleep to me

phoenix-core/src/it/java/org/apache/phoenix/end2end/MaxLookbackIT.java
```diff
@@ -105,60 +129,52 @@ public void testTooLowSCNWithMaxLookbackAge() throws Exception {
     public void testRecentlyDeletedRowsNotCompactedAway() throws Exception {
         try (Connection conn = DriverManager.getConnection(getUrl())) {
             String dataTableName = generateUniqueName();
-            String indexStem = generateUniqueName();
-            createTableAndIndexes(conn, dataTableName, indexStem);
-            String fullIndexName = indexStem + "1";
+            createTable(dataTableName);
+            injectEdge.incValue(WAIT_AFTER_TABLE_CREATION);
```
swaroopak: nit: increamentValue

phoenix-core/src/it/java/org/apache/phoenix/end2end/MaxLookbackIT.java
gjacoby126: Not sure exactly what the question means? Before it was testing the cleanup of the delete markers in an index, now it's testing the cleanup of the delete markers in a base table. The same code's getting exercised either way.

phoenix-core/src/it/java/org/apache/phoenix/end2end/MaxLookbackIT.java
priyankporwal: Suggest setting WAIT_AFTER_TABLE_CREATION = MAX_LOOKBACK_AGE * 100 ... just in case someone increases MAX_LOOKBACK_AGE and leaves WAIT_AFTER_TABLE_CREATION as is.

phoenix-core/src/it/java/org/apache/phoenix/end2end/MaxLookbackIT.java
```diff
@@ -213,62 +241,49 @@ public void testRecentMaxVersionsNotCompactedAway() throws Exception {
...
+            //increment by 10 min to make sure we don't "look back" past table creation
+            injectEdge.incValue(WAIT_AFTER_TABLE_CREATION);
+            populateTable(dataTableName);
+            injectEdge.incValue(1); //increment by 1 so we can see our write
+            long afterInsertSCN = EnvironmentEdgeManager.currentTimeMillis();
             //make sure table and index metadata is set up right for versions
             TableName dataTable = TableName.valueOf(dataTableName);
             assertTableHasVersions(conn, dataTable, versions);
-            String fullIndexName = indexStem + "1";
-            TableName indexTable = TableName.valueOf(fullIndexName);
```
kadirozde: What was the reason for eliminating index tables from the tests? Was not this JIRA originally about index tables?

phoenix-core/src/it/java/org/apache/phoenix/end2end/MaxLookbackIT.java
```diff
@@ -168,36 +184,48 @@ public void testTTLAndMaxLookbackAge() throws Exception {
         conf.setLong(HRegion.MEMSTORE_PERIODIC_FLUSH_INTERVAL, 0L);
         try (Connection conn = DriverManager.getConnection(getUrl())) {
             String dataTableName = generateUniqueName();
-            String indexStem = generateUniqueName();
-            createTableAndIndexes(conn, dataTableName, indexStem);
-            long afterFirstInsertSCN = org.apache.phoenix.util.EnvironmentEdgeManager.currentTimeMillis();
+            createTable(dataTableName);
+            //increment by 10 min to make sure we don't "look back" past table creation
```
priyankporwal: Nit: consistent comment here

phoenix-core/src/it/java/org/apache/phoenix/end2end/MaxLookbackIT.java
gjacoby126: The ManualEnvironmentEdge is copied from HBase code, where that's the method name. I couldn't use the HBase class because Phoenix's EEM is odd and requires a specific abstract subclass of Edge

swaroopak: Given that code does not override the method, it makes sense to rename the method.

phoenix-core/src/it/java/org/apache/phoenix/end2end/MaxLookbackIT.java
swaroopak: aah! my bad, I misunderstood this.

phoenix-core/src/it/java/org/apache/phoenix/end2end/MaxLookbackIT.java
gjacoby126: It was originally about index tables, but I eventually realized that this needed to apply to all tables in order for SCN to work properly, so there's nothing "index specific" about the functionality anymore.

If the consensus is that we should have specific index cases in here to make sure nothing in the indexing coprocs break the general behavior, I'll put some (back) in.

phoenix-core/src/it/java/org/apache/phoenix/end2end/MaxLookbackIT.java
gjacoby126: WAIT_AFTER_TABLE_CREATION isn't meant to be a factor of MAX_LOOKBACK_AGE -- it's just an arbitrarily large number to make sure that all the metadata is older than the current timestamp. It could probably be lower, but since the size doesn't cost us anything I didn't spend too much time trying to optimize it.

phoenix-core/src/it/java/org/apache/phoenix/end2end/MaxLookbackIT.java
kadirozde: I suggest including indexes back and applying the same operations/checks to the index tables along with their data tables.

phoenix-core/src/it/java/org/apache/phoenix/end2end/MaxLookbackIT.java
gjacoby126: Turns out any value >= 1 works. I'll just switch it to 1

priyankporwal: The suggestion was not to have WAIT_AFTER_TABLE_CREATION be a multiple of MAX_LOOKBACK_AGE necessarily, but something larger than MAX_LOOKBACK_AGE  -- just something so that if you did (now - MAX_LOOKBACK_AGE) it didn't go before table creation time.

phoenix-core/src/it/java/org/apache/phoenix/end2end/MaxLookbackIT.java
gjacoby126: All right, will do.

phoenix-core/src/it/java/org/apache/phoenix/end2end/MaxLookbackIT.java
gjacoby126: Not commented out anymore. :-)

phoenix-core/src/it/java/org/apache/phoenix/end2end/MaxLookbackIT.java
gjacoby126: Not commented out anymore. :-)

phoenix-core/src/it/java/org/apache/phoenix/end2end/MaxLookbackIT.java
gjacoby126: That's not necessary because we always advance time by MAX_LOOKBACK_AGE before trying to look back that far.

phoenix-core/src/it/java/org/apache/phoenix/end2end/MaxLookbackIT.java
```diff
@@ -213,51 +274,57 @@ public void testRecentMaxVersionsNotCompactedAway() throws Exception {
         String thirdValue = "ghi";
         try (Connection conn = DriverManager.getConnection(getUrl())) {
             String dataTableName = generateUniqueName();
-            String indexStem = generateUniqueName();
-            createTableAndIndexes(conn, dataTableName, indexStem, versions);
-            long afterInsertSCN = org.apache.phoenix.util.EnvironmentEdgeManager.currentTimeMillis();
+            String indexName = generateUniqueName();
+            createTable(dataTableName);
+            //increment by 10 min to make sure we don't "look back" past table creation
```
priyankporwal: Nit: the comment is not right (nothing here equals 10mins).. wait is only 1 ms now

gjacoby126: Thanks, good catch. Fixed.

phoenix-core/src/it/java/org/apache/phoenix/end2end/MaxLookbackIT.java
gjacoby126: Done.

phoenix-core/src/it/java/org/apache/phoenix/end2end/MaxLookbackIT.java
gjacoby126: And now we're testing both.

PR #682, opened 2020-01-14
PHOENIX-5675: IndexUpgradeTool should allow verify options for IndexTool run

swaroopak:


phoenix-core/src/main/java/org/apache/phoenix/mapreduce/index/IndexUpgradeTool.java
```diff
@@ -138,10 +144,12 @@ public void setInputFile(String inputFile) {
...
     public void setTest(boolean test) { this.test = test; }
+
     public boolean getIsWaitComplete() { return this.isWaitComplete; }
-    public boolean getDryRun() {
-        return this.dryRun;
-    }
+
+    public boolean getDryRun() { return this.dryRun; }
+
+    public String getVerify() { return verify; }
```
priyankporwal: Nit: this.verify to be consistent with other accessors

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/index/IndexUpgradeTool.java
```diff
@@ -704,6 +714,10 @@ private int startIndexRebuilds(Connection conn,
         if (syncRebuild) {
             list.add("-runfg");
         }
+        if(verify != null) {
```
priyankporwal: Replace with IsNullOrEmpty check

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/index/IndexUpgradeTool.java
```diff
@@ -117,6 +122,7 @@
     private String inputTables;
     private String logFile;
     private String inputFile;
+    private String verify = null;
```
priyankporwal: Remove this initial value of null .. looks somewhat special and bug prone

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/index/IndexUpgradeTool.java
```diff
@@ -103,6 +103,11 @@
             "[Optional]Whether or not synchronously rebuild the indexes; "
                     + "default rebuild asynchronous");
 
+    private static final Option INDEX_VERIFY_OPTION = new Option("v",
+            "verify",
+            true,
+            "[Optional] mode to run indexTool with verify options [BEFORE|AFTER|NONE|ONLY|BOTH]");
```
priyankporwal: For discussion: I would have preferred not to add this list here since we would every time we add new options in IndexTool, we would have to keep adapting IndexUpgradeTool. Not a big ask, but something that the community has to remember to do each time. Alternative is to keep these opaque and not even list here.

phoenix-core/src/test/java/org/apache/phoenix/index/IndexUpgradeToolTest.java
```diff
@@ -27,34 +27,57 @@
...
-        Assert.assertEquals(iut.getInputTables(), INPUT_LIST);
-        Assert.assertEquals(iut.getOperation(), upgrade ? UPGRADE_OP : ROLLBACK_OP);
-        Assert.assertEquals(iut.getLogFile(), outputFile);
+    @Test
+    public void testIfVerifyOptionIsPassedToTool() {
+        if (!upgrade) {
+            return;
+        }
+        Assert.assertEquals("", VERIFY_VALUE, indexUpgradeTool.getVerify());
+        String [] values = indexUpgradeTool.getIndexToolArgValues(Mockito.anyString(),
```
gjacoby126: It's unusual to use Mockito.anyString() in a call to something that isn't setting up a mock. Any reason these can't just be dummy values?

phoenix-core/src/test/java/org/apache/phoenix/index/IndexUpgradeToolTest.java
```diff
@@ -27,34 +27,57 @@
...
+                Mockito.anyString(), Mockito.anyString(),
+                Mockito.anyString(), Mockito.anyString());
+        Assert.assertTrue(Arrays.asList(values).contains(VERIFY_VALUE));
+        Assert.assertTrue(Arrays.asList(values).contains("-v"));
+        Assert.assertEquals("verify option and value is not passed consecutively", 1,
+                Arrays.asList(values).indexOf(VERIFY_VALUE) - Arrays.asList(values).indexOf("-v"));
```
gjacoby126: nit: No need to make the same list via Arrays.asList(values) twice.

phoenix-core/src/test/java/org/apache/phoenix/index/IndexUpgradeToolTest.java
```diff
@@ -27,34 +27,57 @@
...
 import org.junit.runners.Parameterized;
 import org.junit.runners.Parameterized.Parameters;
+import org.mockito.Mock;
+import org.mockito.Mockito;
 
 @RunWith(Parameterized.class)
 public class IndexUpgradeToolTest {
     private static final String INPUT_LIST = "TEST.MOCK1,TEST1.MOCK2,TEST.MOCK3";
     private final boolean upgrade;
+    private static final String VERIFY_VALUE = "AFTER";
```
priyankporwal: Nit/TDD guidance:  I would keep the value as "SomeVerifyMode" instead of "AFTER" .. then the test ensures the tool is not actually checking against one of the valid verification modes. Also, highlights the flip side that If the tool is supposed to validate the modes then we have missing tests.

phoenix-core/src/test/java/org/apache/phoenix/index/IndexUpgradeToolTest.java
swaroopak: I can create a dummy value for the entire test and use it wherever necessary. Mock.anyString() was the way I thought to use it as a dummy. :)

gjacoby126:
Already committed in JIRA, so I'm closing

PR #683, opened 2020-01-14
PHOENIX-5674 IndexTool to not write already correct index rows/CFs

kadirozde:


phoenix-core/src/main/java/org/apache/phoenix/coprocessor/IndexRebuildRegionScanner.java
```diff
@@ -504,6 +539,75 @@ public Boolean call() throws Exception {
...
+            mutations.clear();
+            for (Map.Entry<byte[], Put> entry: dataKeyToDataPutMap.entrySet()) {
+                mutations.add(entry.getValue());
+            }
+            rebuildIndexRows(mutations);
+        }
+
+        if (verifyType == IndexTool.IndexVerifyType.AFTER || verifyType == IndexTool.IndexVerifyType.BOTH) {
+            // We have rebuilt index row and now we need to verify them
+            onlyVerify = false; // Stop at the first mismatch
```
gjacoby126: if onlyVerify now really means "don't stop at first mismatch" whether we're doing rebuilds or not, then the variable name is confusing.

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/IndexRebuildRegionScanner.java
```diff
@@ -521,55 +625,58 @@ public boolean next(List<Cell> results) throws IOException {
...
                             }
                             break;
                         }
                         rowCount++;
                     }
-
                 } while (hasMore && rowCount < pageSizeInRows);
-                if (!mutations.isEmpty() && !onlyVerify) {
+            }
+            if (!scan.isRaw() && indexRowKey == null) {
```
gjacoby126: what does the scan being raw indicate? A comment would be good.

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/IndexRebuildRegionScanner.java
```diff
@@ -385,13 +388,13 @@ private void verifySingleIndexRow(Result indexRow, final Put dataRow) throws IOE
             // This means the index row does not have any covered columns. We just need to check if the index row
             // has only one cell (which is the empty column cell)
             if (indexRow.rawCells().length == 1) {
-                return;
+                return false;
```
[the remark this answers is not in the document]
kadirozde: Yes, this is a success path and it should be true. Good catch.

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/IndexRebuildRegionScanner.java
```diff
@@ -495,7 +524,13 @@ public Boolean call() throws Exception {
                         exceptionMessage = "Pool closed, not attempting to verify index rows! " + indexHTable.getName();
                         throw new IOException(exceptionMessage);
                     }
-                    verifyIndexRows(keys);
+                    verifyIndexRows(keys, perTaskRowKeyToDataPutMap);
+                    if (verifyType == IndexTool.IndexVerifyType.BEFORE || verifyType == IndexTool.IndexVerifyType.BOTH) {
+                        synchronized (dataKeyToDataPutMap) {
```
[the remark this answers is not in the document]
kadirozde: There is no concurrent map for tree maps (i.e., sorted maps). Please note that this is called after verifying all rows per task. So, it should not lead to a lock contention issue as it will not be a frequent operation.

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/index/IndexTool.java
```diff
@@ -205,11 +205,12 @@ public static IndexVerifyType fromValue(byte[] value) {
...
-            "To verify every data row has a corresponding row. The accepted values are NONE, ONLY, BEFORE," +
-                    " AFTER, and BOTH. NONE is for no inline verification, which is also the default for this option. " +
-                    "ONLY is for verifying without rebuilding index rows. The rest for verifying before, after, and " +
-                    "both before and after rebuilding row. If the verification is done before rebuilding rows and " +
-                    "the correct index rows are not rebuilt. Currently supported values are NONE, ONLY and AFTER ");
+            "To verify every data row has a corresponding row of a global index. For other types of indexes, " +
+                    "this option will be silently ignored. The accepted values are NONE, ONLY, BEFORE,  AFTER, and BOTH. " +
+                    "NONE is for no inline verification, which is also the default for this option. ONLY is for " +
+                    "verifying without rebuilding index rows. The rest for verifying before, after, and both before " +
+                    "and after rebuilding row. If the verification is done before rebuilding rows and the correct " +
```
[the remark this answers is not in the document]
kadirozde: I will reword

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/IndexRebuildRegionScanner.java
kadirozde: I rename it to doNotFail

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/IndexRebuildRegionScanner.java
```diff
@@ -504,11 +537,82 @@ public Boolean call() throws Exception {
...
+        for (Boolean result : taskResultList) {
+            if (result == null) {
+                // there was a failure
+                throw new IOException(exceptionMessage);
+            }
+        }
+    }
+
+    private void verifyAndOrRebuildIndex() throws IOException {
+        if (verifyType == IndexTool.IndexVerifyType.AFTER || verifyType == IndexTool.IndexVerifyType.NONE ||
```
priyankporwal: 2nd and 3rd OR'ed expr are the same. Should the second OR'ed expr be verifyType == IndexTool.IndexVerifyType.BOTH?

priyankporwal: Hmm .. just one of them needs to be removed

priyankporwal: Readability suggestion : (shortened the names)

if (vt == AFTER || vt == NONE) {

   if (vt == NONE) {
       return;
   }
}
else if (vt == BEFORE || vt == BOTH || vt == ONLY) {
   ...
   if (vt == ONLY)  {
      return;
   }
   ....
}

kadirozde: Yes, I will remove one of them

kadirozde: Is "vt"  more readable than "verifyType"? :-))

priyankporwal: Nope .. that's why I said "shortened the names" for my comment purpose.
Readability suggestion was about organization of the various if() blocks - and it sucks how it shows in the comment above.

PR #685, opened 2020-01-15
PHOENIX-5676 Inline-verification from IndexTool does not handle TTL/r…

abhishek-chouhan:
…ow-expiry

### DIFF
--- a/phoenix-core/src/it/java/org/apache/phoenix/end2end/IndexToolIT.java
+++ b/phoenix-core/src/it/java/org/apache/phoenix/end2end/IndexToolIT.java
@@ -43,6 +43,7 @@ import org.apache.hadoop.conf.Configuration;
 import org.apache.hadoop.hbase.Cell;
 import org.apache.hadoop.hbase.CellUtil;
 import org.apache.hadoop.hbase.HBaseIOException;
+import org.apache.hadoop.hbase.HColumnDescriptor;
 import org.apache.hadoop.hbase.HTableDescriptor;
 import org.apache.hadoop.hbase.TableName;
 import org.apache.hadoop.hbase.client.Admin;
@@ -59,14 +60,14 @@ import org.apache.hadoop.hbase.regionserver.HRegion;
 import org.apache.hadoop.hbase.regionserver.MiniBatchOperationInProgress;
 import org.apache.hadoop.hbase.regionserver.RegionScanner;
 import org.apache.hadoop.hbase.util.Bytes;
+import org.apache.hadoop.hbase.util.Pair;
 import org.apache.hadoop.mapreduce.Job;
 import org.apache.phoenix.hbase.index.IndexRegionObserver;
 import org.apache.phoenix.jdbc.PhoenixConnection;
 import org.apache.phoenix.mapreduce.index.IndexTool;
-import org.apache.phoenix.query.ConnectionQueryServices;
 import org.apache.phoenix.mapreduce.index.PhoenixIndexImportDirectMapper;
 import org.apache.phoenix.mapreduce.index.PhoenixServerBuildIndexMapper;
-
+import org.apache.phoenix.query.ConnectionQueryServices;
 import org.apache.phoenix.query.QueryServices;
 import org.apache.phoenix.query.QueryServicesOptions;
 import org.apache.phoenix.schema.PIndexState;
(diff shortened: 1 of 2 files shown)


phoenix-core/src/main/java/org/apache/phoenix/coprocessor/IndexRebuildRegionScanner.java
```diff
@@ -584,16 +633,19 @@ public boolean next(List<Cell> results) throws IOException {
             if (onlyVerify) {
                 addToBeVerifiedIndexRows();
             }
-            ArrayList<KeyRange> keys = new ArrayList<>(rowCountPerTask);
+            ArrayList<KeyRange> keyRanges = new ArrayList<>(rowCountPerTask);
+            Set<byte[]> indexRowKeys = new HashSet<>(rowCountPerTask);
```
kadirozde: A byte array hash set would not work here. The byte array needs to be wrapped with equals and hashCode, or Map<byte[], Put> created with Maps.newTreeMap(Bytes.BYTES_COMPARATOR) can be used instead.

abhishek-chouhan:
Sorry for that miss. Have updated to use TreeSet with appropriate comparator @kadirozde and verified the expected behavior.

phoenix-core/src/it/java/org/apache/phoenix/end2end/IndexToolIT.java
```diff
@@ -510,6 +508,72 @@ public void testIndexToolOnlyVerifyOption() throws Exception {
...
+            runIndexTool(directApi, useSnapshot, schemaName, dataTableName, indexTableName, null, 0,
+                IndexTool.IndexVerifyType.AFTER);
+
+            // Set ttl of index table ridiculously low so that all data is expired
+            Admin admin = conn.unwrap(PhoenixConnection.class).getQueryServices().getAdmin();
+            TableName indexTable = TableName.valueOf(indexTableFullName);
+            HColumnDescriptor desc = admin.getTableDescriptor(indexTable).getColumnFamilies()[0];
+            desc.setTimeToLive(1);
+            admin.modifyColumn(indexTable, desc);
+            Thread.sleep(1000);
```
swaroopak: why not use EEM to advance time instead of sleeping?

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/IndexRebuildRegionScanner.java
```diff
@@ -465,17 +490,32 @@ private void verifyIndexRows(ArrayList<KeyRange> keys) throws IOException {
...
                 }
                 verifySingleIndexRow(result, dataPut);
+                indexRowKeys.remove(result.getRow());
                 rowCount++;
             }
         } catch (Throwable t) {
             ServerUtil.throwIOException(indexHTable.getName().toString(), t);
         }
+        // Check if any expected rows from index(which we didn't get) are already expired due to TTL
+        // TODO: metrics for expired rows
```
swaroopak: Could you please also create a jira for this so that we don't lose it in the comments? Thanks.

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/IndexRebuildRegionScanner.java
```diff
@@ -465,17 +490,32 @@ private void verifyIndexRows(ArrayList<KeyRange> keys) throws IOException {
...
+        if (!indexRowKeys.isEmpty()) {
+            Iterator<byte[]> itr = indexRowKeys.iterator();
+            long currentTime = EnvironmentEdgeManager.currentTime();
+            while(itr.hasNext()) {
+                byte[] indexKey = itr.next();
+                long ts = getMaxTimestampFromPut(indexKeyToDataPutMap.get(indexKey));
+                if (indexTableTTL != HConstants.FOREVER && ts < (currentTime - (long)indexTableTTL * 1000)) {
```
swaroopak: This can be replaced with the function at line 530

phoenix-core/src/it/java/org/apache/phoenix/end2end/IndexToolIT.java
abhishek-chouhan: modify is async operation, which needs to actually disable/enable regions. Faking a skip in time would not help.

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/IndexRebuildRegionScanner.java
abhishek-chouhan: Sure, will do.

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/IndexRebuildRegionScanner.java
abhishek-chouhan: The logic was a little different from checking a cellTs directly, in the sense that it requires finding the max ts first amongst different cells in the put, just that the comparison part is common. Didn't feel very strongly about creating a different function for that.

phoenix-core/src/it/java/org/apache/phoenix/end2end/IndexToolIT.java
swaroopak: is it not sleeping because the test would want to expire the row?

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/IndexRebuildRegionScanner.java
swaroopak: I think "isCellExpired" function can expect "ts" as a parameter and then that will solve the problem.

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/IndexRebuildRegionScanner.java
swaroopak: It would be nice to link it to the code's comment.

phoenix-core/src/it/java/org/apache/phoenix/end2end/IndexToolIT.java
abhishek-chouhan: It is mainly for the admin action but solves both the things. For majority of the cases the admin operation finishes in 1 sec so the retry loop is skipped altogether.

gjacoby126: 1. Why do you need the first sleep for 1s if you're going to loop to wait later anyway?
2. Why is each sleep so long (2s) rather than something short, like 10 or 100ms?

abhishek-chouhan: The first sleep was handling both the things of ttl expiry as well as giving reasonable time for most cases of the admin operation to succeed. The alternative is to use edge just for the ttl of 1 second but still have the retries in place since we actually need to wait for the admin operation. No particular reason of having the retry interval large, thought 40 seconds overall for the test should be enough to have it running reasonably deterministic on the most flaky of hardwares. Let me know if you strongly feel about changing these and i can work on them.

abhishek-chouhan:
Thanks for the reviews!

PR #687, opened 2020-01-22
Hbase compat buildtime

stoty:
This is the proposed solution for HBase compatibility modules for Phoenix.

run maven with -Dhbase.compat.version=2.0.1 or -Dhbase.compat.version=2.2.1 to choose the version it is built against.


stoty:
I have reworked the build system.

You need to specify both the hbase compatibility module's version, as well as the HBase version that you build/test/shade against.

f.e:
mvn -Dhbase.version=2.2.3 -Dhbase.compat.version=2.2
(the default is 2.0.1 and 2.0, the old behaviour)

I have also reset the project version to 5.1-SNAPSHOT, and added the hbase compatbility information as a classifier to the shaded artifacts.

This should play nice with maven repos, as there is one set of unshaded artifacts, and the shaded artifacts can be selected by classifers.

phoenix-core/src/it/java/org/apache/phoenix/end2end/index/MutableIndexSplitIT.java
```diff
@@ -135,9 +135,11 @@ private void createTableAndLoadData(Connection conn1, String tableName, String i
         List<RegionInfo> regionsOfUserTable = null;
         for(int i = 0; i <=1; i++) {
             boolean split = false;
-            for (int j = 0; j < 60 && !split; j++) {
+            for (int j = 0; j < 150 && !split; j++) {
                 try {
                     if (localIndex) {
+                        //With Hbase 2.2 the local index splits trigger longCompactions, and have to wait for an
```
joshelser: Can we change the period of this discharger? That would prevent this wait+retry to be copied to the N+1th test case.

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/PhoenixAccessController.java
```diff
@@ -464,12 +464,14 @@ public void preIndexUpdate(ObserverContext<PhoenixMetaDataControllerEnvironment>
...
        }
 
+     //FIXME This seems to have no effect at all
      private void getUserDefinedPermissions(final TableName tableName,
              final List<UserPermission> userPermissions) throws IOException {
           User.runAsLoginUser(new PrivilegedExceptionAction<List<UserPermission>>() {
               @Override
               public List<UserPermission> run() throws Exception {
-                  final List<UserPermission> userPermissions = new ArrayList<UserPermission>();
+                 //FIXME We are masking the parameter list that we are supposed to add to
```
joshelser: A todo you left for yourself? :)

phoenix-hbase-compat-2.0/pom.xml
```diff
@@ -0,0 +1,95 @@
...
+-->
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+  xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+  <modelVersion>4.0.0</modelVersion>
+    <parent>
+    <groupId>org.apache</groupId>
+    <artifactId>apache</artifactId>
+    <version>21</version>
+  </parent>
+<!--   <parent> -->
```
joshelser: We don't want this declaring phoenix as the parent?

phoenix-hbase-compat-2.0/src/main/java/org/apache/phoenix/compat/hbase/CompatUtil.java
```diff
@@ -0,0 +1,36 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more contributor license
+ * agreements. See the NOTICE file distributed with this work for additional information regarding
```
joshelser: Formatting on this license header, please.

phoenix-hbase-compat-2.0/src/main/java/org/apache/phoenix/compat/hbase/CompatUtil.java
```diff
@@ -0,0 +1,36 @@
...
+import java.util.List;
+
+import org.apache.hadoop.hbase.Cell;
+import org.apache.hadoop.hbase.MetaTableAccessor;
+import org.apache.hadoop.hbase.client.Connection;
+import org.apache.hadoop.hbase.client.RegionInfo;
+import org.apache.hadoop.hbase.util.Pair;
+
+public class CompatUtil {
+
```
joshelser: Nit: private constructor to prevent instantiation.

phoenix-hbase-compat-2.2/pom.xml
```diff
@@ -0,0 +1,94 @@
...
+              <artifactId>maven-project-info-reports-plugin</artifactId>
+          </plugin>
+          <plugin>
+              <groupId>org.codehaus.mojo</groupId>
+              <artifactId>findbugs-maven-plugin</artifactId>
+          </plugin>
+      </plugins>
+  </reporting>
+  
+<!--         <profiles> -->
```
joshelser: Nit: remove

phoenix-hbase-compat-2.0/pom.xml
```diff
@@ -0,0 +1,95 @@
...
+              <artifactId>maven-project-info-reports-plugin</artifactId>
+          </plugin>
+          <plugin>
+              <groupId>org.codehaus.mojo</groupId>
+              <artifactId>findbugs-maven-plugin</artifactId>
+          </plugin>
+      </plugins>
+  </reporting>
+  
+      <profiles>
```
joshelser: These don't seem to have any consequence. Remove them?

pom.xml
```diff
@@ -1031,6 +1054,42 @@
...
     </profile>
+    <!-- Choose which hbase compatibility module we compile, test, and shade against.
+    Shouldn't have an effect on the generated code and artifacts apart from the contents of the the shaded jars
+    When specifying a non-default compatibility module, you also have to specify a
+    compatible hbase.version i.e mvn clean install -P phoenix-hbase-compat-2.2 -Dhbase.version=2.2.3
+    -->
+    <profile>
+      <id>phoenix-hbase-compat-2.0</id>
+      <activation>
+        <activeByDefault>true</activeByDefault>
```
joshelser: I thought this didn't actually work, but poking at it, maybe it does? I was thinking that there was some weirdness around the activeByDefault not doing what you are intending. Using `mvn help:active-profiles` looks liek the appropriate profiles are getting activated so maybe I just think life has to be harder than it needs to be ;)

pom.xml
```diff
@@ -1031,6 +1054,42 @@
         </plugins>
       </build>
     </profile>
+    <!-- Choose which hbase compatibility module we compile, test, and shade against.
+    Shouldn't have an effect on the generated code and artifacts apart from the contents of the the shaded jars
+    When specifying a non-default compatibility module, you also have to specify a
+    compatible hbase.version i.e mvn clean install -P phoenix-hbase-compat-2.2 -Dhbase.version=2.2.3
```
joshelser: Instead of forcing the user to activate the module and install it locally, you can simply add the appropriate compat module to the build reactor (specifying it with a new `<modules><module/></modules` element).

Yes, it forces us to rebuild the compat, but they both look very small at this point. The trick will be to make sure the appropriate one shows up in the assembly, not both of them (might automatically happen if all dependencies are getting pulled in properly).

pom.xml
joshelser: Talked to Busbey about this because I figured he would be the only other Maven guru I know who remembers this. He pointed me back to https://issues.apache.org/jira/browse/MNG-4917

The question is why, when I specified a profile, did I not see these semantics? Sounds like https://issues.apache.org/jira/browse/MNG-6787 might be affecting my 3.6.3 Maven version.

I do think this needs to be changed from 2 profiles activating 2 states to 3 profiles activating 2 states (as dumb as that is):
* profile1 activated by `! hbase.compat.version` which selects the hbase 2.0 compat module
* profile 2 activated by `hbase.compat.version=2.0` which selects the hbase 2.0 compat module
* profile 3 activated by `hbase.compat.version=2.2` which selects the hbase 2.2 compat module

phoenix-assembly/src/build/components/all-common-jars.xml
```diff
@@ -27,9 +27,9 @@
       <directory>${project.basedir}/../phoenix-client/target</directory>
       <outputDirectory>/</outputDirectory>
       <includes>
-        <include>phoenix-client-${project.version}.jar</include>
+        <include>phoenix-client-${project.version}-HBase-${hbase.compat.version}.jar</include>
```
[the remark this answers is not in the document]
stoty: I've extracted the version string to property to make sure the spelling is consistent.

phoenix-core/src/it/java/org/apache/phoenix/end2end/index/MutableIndexSplitIT.java
stoty: I'd prefer to keep the changes minimal for this patch.
But I'll make a separate ticket for that.

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/PhoenixAccessController.java
stoty: Yes, I don't want fix unrelated bugs in this commit, to make testing easier.
Also, writing a test case for this promises to be quite hairy, which is another reason not to tackle this right now.
I'll make a separate ticket for this as well.

phoenix-hbase-compat-2.0/src/main/java/org/apache/phoenix/compat/hbase/CompatUtil.java
stoty: Done

phoenix-hbase-compat-2.2/pom.xml
stoty: Done

phoenix-hbase-compat-2.0/pom.xml
stoty: Done

phoenix-hbase-compat-2.0/pom.xml
stoty: We don't. Having phoenix as parent caused problems with the module depending on itself. 
We wouldn't really win anything by that, but we'd have to dance around the circular dependency issue.

phoenix-hbase-compat-2.0/src/main/java/org/apache/phoenix/compat/hbase/CompatUtil.java
stoty: Done

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/PhoenixAccessController.java
joshelser: Ah, I see now. I just read it too fast. That sounds fine!

pom.xml
stoty: Right now I am using `hbase.compat.version` for two things
- selecting the right profile
- building the filenames and classifiers.

So the build is not designed to work correctly if you directly specify the profile.

We could probably make that work, but we'd have to push down profile handling to each module and set the filenames/classifiers directly, which would be rather ugly. 

Actually, MNG-6787 even helps us, as it means that you we can specify other profiles (namely release), and still get the implicit 2.0.1 profile as well.

However, to make behaviour more consistent with a future maven release where MNG-6787 is fixed, I agree that adding a third entry is the best option.

pom.xml
stoty: It already works that way, we always build and install both modules (maven artifacts)

However, I haven't thought of the /lib dir in the assembly, and the compat dependency doesn't get added automatically. I have fixed that now.

pom.xml
stoty: I've just realized that the example in the comment is bad (uses -P ). I'll fix that tomorrow.

phoenix-core/src/it/java/org/apache/phoenix/end2end/index/MutableIndexSplitIT.java
stoty: https://issues.apache.org/jira/browse/PHOENIX-5699

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/PhoenixAccessController.java
stoty: https://issues.apache.org/jira/browse/PHOENIX-5700

pom.xml
stoty: fixed in last version

pom.xml
```diff
@@ -1031,6 +1049,55 @@
...
+          <artifactId>phoenix-hbase-compat-2.0.1</artifactId>
+        </dependency>
+      </dependencies>
+    </profile>
+    <profile>
+      <id>phoenix-hbase-compat-2.2.1</id>
+      <activation>
+        <property>
+          <name>hbase.compat.version</name>
+          <value>2.2.1</value>
```
infraio: Need to specify 2.2.1? I thought it should compact with 2.2.x too if it compact with 2.2.1?

stoty: Phoenix uses semi-public APIs that tend to change between patch releases. Case in point: 2.2.0 does not have the multi-merge feature that 2.2.1 added, while late 2.0 and 2.1 releases have also added it. We don't support 2.0.0 either, though I don't know if that's due to API changes or some critical bug.

So the modules are actually supporting version ranges, and I named them after the first supported release.

Using the full version means that if there is a change again at say 2.2.5, we can also handle that more elegantly. (by having 2.2.1 and 2.2.5 modules, as opposed to 2.2 and 2.2.5 modules)

There are actually four API combinations we could support (2.0 v. 2.2 base / 2 region merge v multi-merge). In this patch I only added support for latest 2.2.1+ versions, and kept the support for the old 2.0 and 2.1 versions, though it wouldn't take much code to support all four current combinations.

stoty: I've just run some quick tests, the 2.0.1 compatibility module (and consequently, the current master branch) in fact only compiles with HBase 2.0.1 - 2.0.4 and 2.1.0 - 2.1.2. Maybe adding a module for current 2.0 versions would be useful as well.

joshelser: I think having a known list of "working" versions within a minor-release is sufficient to start. Long-term, we don't want to encourage users to stay on the non-recent bugfix releases anyways. Phoenix can be a forcing function :)

joshelser:
Ok, I see a bit better how we have this working. Re-stating both for my own benefit and anyone else who comes along :)

* We introduce compatibility modules tied to HBase "breaking" versions
* * these are generally aligned to HBase minor versions, 2.x
* We have profiles, activated by a system property, which set a dependency on exactly one of those compatibility modules
* Each profile activates a specific hbase version (e.g. the earliest in that minor version which works with Phoenix, 2.2.1 for 2.x)
* * Users can override that hbase version via system property to compile against a specific hbase version in that minor release line.

We should get the docs updated on building to include this. It looks like build.txt is woefully out of date. Maybe put it in README.md or introduce a BUILDING.md? (something that github will nicely render)

stoty:
Yes, that is correct.

However, we'd better think of the compatibility modules not as supporting minor releases, but as supporting HBase version ranges, which may not even be continuous (e.g. the 2.0.1 module supports both 2.0 and 2.1 releases made in the same timeframe, but breaks on a change introduced in both branches).
 
What @joshelser  is describing may well end being the way we handle breaking HBase API changes (i.e. we support the latest patch releases of each minor version), but that would not be a consequence of how the module system works, but rather decision made by us to limit the number of HBase compatibility modules we support/test/maintain, and as suggested, to nudge users to update to the latest HBase point release.

We will also definitely need to flesh out the build instructions, and document how to download the HBase specific artifacts/assemblies, though I am not sure wether to put this in the repo, or add it to web site (or both). Maybe keeping it on the website, and adding a direct link to the build instructions in README.md would be the best.

joshelser:
It looks like omid is bringing in a custom version of hbase-endpoint to the assembly (looking at the bin tarball's lib directory):

```
[INFO] --- maven-dependency-plugin:3.1.1:tree (default-cli) @ phoenix-assembly ---
[INFO] org.apache.phoenix:phoenix-assembly:pom:5.1.0-SNAPSHOT
[INFO] +- org.apache.phoenix:phoenix-core:jar:5.1.0-SNAPSHOT:compile
[INFO] |  +- org.apache.omid:omid-hbase-shims-hbase2.x:jar:1.0.1:compile
[INFO] |  |  +- org.apache.hbase:hbase-endpoint:jar:2.0.1:compile
```

In the above, I had overwritten the version of the hbase dependency to be 2.1.7, but hbase-endpoint was still at 2.0.1. We can get rid of that my just setting the dependency on hbase-endpoint in phoenix-core, I think. At least, I think that would be the cleanest way to work around it.

I think I've convinced myself that renaming the compat modules from 2.y.z to 2.y is a good idea. If there is some .z release which doesn't work for Phoenix, let's just preclude that value from being set by creating a maven-enforcer-plugin in that compat module's profile to fail on that value.

The builtin rule RequireProperty lets us do exactly that (sadly, not via the nice version range syntax) with a regex like `2\.2\.[1-9]`. Recognizing that this will fail if we get to `2.2.10`, we can also write a custom enforcer rule which would do exactly what we want. I just think that's overkill for now :)

> supporting HBase version ranges, which may not even be continuous (e.g. the 2.0.1 module supports both 2.0 and 2.1 releases made in the same timeframe, but breaks on a change introduced in both branches).

I was thinking more about this. We don't need separate compat modules for (most of) 2.1.x and 2.2, but would it reduce user-error to make a profile that just activates `phoenix-compat-hbase-2.2` when the user sets `-Dhbase.compat.version=2.1`?

In HBase, when dealing with Hadoop dependencies, we have a property named `hadoop.profile` which is doing the same thing you're doing for HBase now. It would be nice for consistency to rename `hbase.compat.version` to be `hbase.profile`.

I'll go through the rest of the code-changes, but I think, with the above, I am happy with the build-system level changes :) Awesome work.

phoenix-hbase-compat-2.0.1/pom.xml
```diff
@@ -0,0 +1,76 @@
...
+      </plugin>
+    </plugins>
+  </build>
+
+  <dependencies>
+       <!-- HBase dependencies -->
+      <dependency>
+        <groupId>org.apache.hbase</groupId>
+        <artifactId>hbase-server</artifactId>
+        <version>2.0.1</version>
```
joshelser: Should be `hbase.compat.version`?

phoenix-hbase-compat-2.0/pom.xml
joshelser: Note for later -- when we go to make the official release, we should make sure that nothing "funny" happens because the compat modules have a different parent pom. I don't think that will affect what gets deployed to the staging repository, but I also wouldnt' be surprised if that does affect things :)

phoenix-hbase-compat-2.0.1/pom.xml
joshelser: Thinking through this some more, I think this goes to my `hbase.profile` suggestion in the other thread.

`hbase.profile` activates the 2.y release "compat line" we want, and pull in that code. `hbase.version` sets the specific version of hbase we depend on. We'll automatically set one for the compat module which can be the latest in that 2.y.z release line (we should just use the latest .z -- often times hbase will fix things which we would see breaking in our tests).

Is that what you see? Stating it out lout to make sure we're on the same page :)

stoty:
> It looks like omid is bringing in a custom version of hbase-endpoint to the assembly (looking at the bin tarball's lib directory):
> 
> ```
> [INFO] --- maven-dependency-plugin:3.1.1:tree (default-cli) @ phoenix-assembly ---
> [INFO] org.apache.phoenix:phoenix-assembly:pom:5.1.0-SNAPSHOT
> [INFO] +- org.apache.phoenix:phoenix-core:jar:5.1.0-SNAPSHOT:compile
> [INFO] |  +- org.apache.omid:omid-hbase-shims-hbase2.x:jar:1.0.1:compile
> [INFO] |  |  +- org.apache.hbase:hbase-endpoint:jar:2.0.1:compile
> ```

Actually, just adding it to parent DependencyManagement is sufficient.

Speaking of transactions, we know that tephra does not support HBase 2.2. 
This is something that we will have to resolve, preferably after this is commited 
(i.e. in the initial version, I think it's OK not to have working tephra for 2.2)

phoenix-hbase-compat-2.0.1/pom.xml
stoty: > Should be `hbase.compat.version`?

My intent is to always build (and make available in the maven repo) each of the compatibility modules, so that one can use the version-independent unshaded JARs/artifacts plus the right compatibility module, to get a working phoenix classpath. (even though it is not really a use case that we expect in the real world)

specifying hbase.compat.version here would break building the compatibility module for the non-selected versions.

stoty:
> In HBase, when dealing with Hadoop dependencies, we have a property named `hadoop.profile` which is doing the same thing you're doing for HBase now. It would be nice for consistency to rename `hbase.compat.version` to be `hbase.profile`.

Sure, this is a no-brainer. I need some more time to digest your suggested minor vs point release suggestion, and how it can be implemented.

stoty:
This version includes most of your suggestions @joshelser.

You can set hbase.profile to 2.0 2.1 or 2.2 , and get a build with the latest supported HBase patch release (except for 2.0, which is EOM, and I kept it at the 2.0.1 API level to be directly comparable with the current master version)
 
I kept the full version number in the compatibility modules names, as I am uncomfortable with changing their semantics when following the Hbase API changes, and I'd prefer to keep the door open for supporting multiple HBase patch versions for some time in case the API changes.

I've left the maven qualifier equal to the hbase.profile, to ease CI/CD, and to keep the number of  download assemblys / official maven artifact versions manageable. We could use compat.hbase.version or hbase.version as well here, none of the choices is perfect.

Yo can still override the defaults that come from the -Dhbase.profile setting, i.e.

mvn clean package -Dhbase.profile=2.0 -Dhbase.version=2.1.1 will build a version with the 2.0.1 compatibility module, but include the 2.1.1 runtime.

stoty:
I still need to write BUILDING.md, but I wanted to get this out for review.

joshelser:
> I kept the full version number in the compatibility modules names, as I am uncomfortable with changing their semantics when following the Hbase API changes, and I'd prefer to keep the door open for supporting multiple HBase patch versions for some time in case the API changes.

Sounds fine. I think I just keep hoping for the best in terms of stability in a minor release, but we know that we hit issues around this :). Your approach is good.

> Actually, just adding it to parent DependencyManagement is sufficient.

👍 

> in the initial version, I think it's OK not to have working tephra for 2.2

:+1:

joshelser:
> mvn clean package -Dhbase.profile=2.0 -Dhbase.version=2.1.1 will build a version with the 2.0.1 compatibility module, but include the 2.1.1 runtime.

This looks really nice and is (sufficiently) clear to understand. Like you point out, there are examples where there are edge cases that can be misleading to someone who doesn't know our history (_but HBase 2.y.z doesn't actually work with the 2.y profile!_). Acknowledging this as something we can't fix after the fact (we have to get better at catching that _before_ the HBase release happens), I think your solution is ideal.

phoenix-hbase-compat-2.0.1/pom.xml
joshelser: Right you are. Thanks for catching that :)

pom.xml
```diff
@@ -367,6 +380,45 @@
...
+              <goal>enforce</goal>
+            </goals>
+            <configuration>
+              <rules>
+                <evaluateBeanshell>
+                 <condition>
+                   import java.util.regex.Pattern;
+                   import java.lang.Integer;
+
+                   versionPattern = Pattern.compile("(\\d+)\\.(\\d+)\\.(\\d+)");
```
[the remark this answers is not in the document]
stoty: Yes. I have just checked with -Dhbase.version=2.2.4-SNAPSHOT.

I have also updated the regex in the meantime to also handle some downstream release numbering schemes (CDP) as well, by binding to the end of the version string.

pom.xml
```diff
@@ -367,6 +380,45 @@
...
+            <goals>
+              <goal>enforce</goal>
+            </goals>
+            <configuration>
+              <rules>
+                <evaluateBeanshell>
+                 <condition>
+                   import java.util.regex.Pattern;
+                   import java.lang.Integer;
+
```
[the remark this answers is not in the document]
stoty: -Denforcer.skip=true disables all enforcer rules.
Do you think we need a dedicated property to disable this specific check (we don't have any others ATM) ?

pom.xml
```diff
@@ -367,6 +380,45 @@
...
+                   versionPattern = Pattern.compile("(\\d+)\\.(\\d+)\\.(\\d+)");
+                   versionMatcher = versionPattern.matcher("${hbase.version}");
+                   versionMatcher.find();
+
+                   hbaseMajor = Integer.parseInt(versionMatcher.group(1));
+                   hbaseMinor = Integer.parseInt(versionMatcher.group(2));
+                   hbasePatch = Integer.parseInt(versionMatcher.group(3));
+
+                   hbaseMajor == 2 &amp;&amp; (
```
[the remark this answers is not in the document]
stoty: No, I am eternal optimist, so I didn't put an upper bound on the latest supported version ranges.
The rules only exclude the known not working/unsupported combinations.

pom.xml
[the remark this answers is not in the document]
joshelser: Nope, that's sufficient!

Can you make sure that's mentioned in BUILDING.md too, please?

pom.xml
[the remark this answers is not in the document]
joshelser: You did exactly what I would have suggested. I just misread initially due to the urlencoded operators :)

pom.xml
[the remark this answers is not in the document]
joshelser: Perfect!

stoty:
Merged a slightly amended version.

PR #688, opened 2020-01-23
PHOENIX-5629 Phoenix Function to return HBase row timestamp

tkhurana:


tkhurana:
@priyankporwal @gokceni

phoenix-core/src/it/java/org/apache/phoenix/end2end/RowTimestampEpochFunctionIT.java
```diff
@@ -0,0 +1,128 @@
...
+    private void verifyHbaseRowTimestamp(String tableName, ResultSet rs) throws Exception {
+
+        Scan scan = new Scan();
+        byte[] emptyKVQualifier = EncodedColumnsUtil.getEmptyKeyValueInfo(true).getFirst();
+        try (org.apache.hadoop.hbase.client.Connection hconn = ConnectionFactory.createConnection(config)) {
+            Table table = hconn.getTable(TableName.valueOf(tableName));
+            ResultScanner resultScanner = table.getScanner(scan);
+            while (rs.next()) {
+                Result result = resultScanner.next();
+                long timeStamp = result.getColumnLatestCell(QueryConstants.DEFAULT_COLUMN_FAMILY_BYTES, emptyKVQualifier).getTimestamp();
```
gjacoby126: Should also assert the expected number of rows are returned.

phoenix-core/src/it/java/org/apache/phoenix/end2end/RowTimestampEpochFunctionIT.java
```diff
@@ -0,0 +1,128 @@
...
+
+import java.sql.Connection;
+import java.sql.Date;
+import java.sql.DriverManager;
+import java.sql.PreparedStatement;
+import java.sql.ResultSet;
+
+import static org.junit.Assert.assertEquals;
+
+public class RowTimestampEpochFunctionIT extends ParallelStatsDisabledIT {
```
gjacoby126: Since this is dependent on the Cell level, do we need to parameterize the tests so they verify all the supported column encodings work with this feature?

phoenix-core/src/main/java/org/apache/phoenix/expression/function/RowTimestampEpochFunction.java
```diff
@@ -0,0 +1,43 @@
...
+    }
+
+    public RowTimestampEpochFunction(List<Expression> children) { super(children); }
+
+    @Override
+    public String getName() { return NAME; }
+
+    @Override
+    public boolean evaluate(Tuple tuple, ImmutableBytesWritable ptr) {
+        assert (tuple.size() >= 1);
```
gjacoby126: Do we use asserts in other functions? Agree there needs to be some validation logic here, just unsure of what the right exception is.

phoenix-core/src/main/java/org/apache/phoenix/expression/function/RowTimestampEpochFunction.java
```diff
@@ -0,0 +1,43 @@
...
+import java.util.List;
+
+/**
+ * Function to return the timestamp of the empty column which functions as the row timestamp. The
+ * result returned can be used for debugging(eg. using HBase shell), logging etc.
+ */
+@BuiltInFunction(name = RowTimestampEpochFunction.NAME, args = {})
+public class RowTimestampEpochFunction extends ScalarFunction {
+
+    public static final String NAME = "ROW_TIMESTAMP_EPOCH";
```
priyankporwal: "Epoch" is generally used as beginning/origin of a time-period. Is there another name we can use for this one? .. e.g. ROW_TIMESTAMP_STRING

phoenix-core/src/main/java/org/apache/phoenix/expression/function/RowTimestampEpochFunction.java
tkhurana: I didn't find any asserts being used in other functions. Also didn't find any functions which explicitly raise exceptions. Intuitively, an assert made most sense here because you will always expect a non empty result tuple and that's what you are asserting on.

phoenix-core/src/it/java/org/apache/phoenix/end2end/RowTimestampEpochFunctionIT.java
```diff
@@ -0,0 +1,161 @@
...
+            for (int id = 0; id < count; ++id) {
+                stmt.setInt(1, id);
+                stmt.setDate(2, rowTimestampDate);
+                stmt.setString(3, "KV1_" + id);
+                stmt.setString(4, "KV2_" + id);
+                stmt.executeUpdate();
+            }
+            conn.commit();
+
+            String dql = "SELECT ROW_TIMESTAMP_EPOCH() FROM " + tableName;
```
swaroopak: could you please add one test for using this function in the WHERE clause?

tkhurana: sure

phoenix-core/src/main/java/org/apache/phoenix/expression/function/RowTimestampStringFunction.java
```diff
@@ -0,0 +1,43 @@
+package org.apache.phoenix.expression.function;
```
swaroopak: nit: please include the license here.

PR #689, opened 2020-01-23
PHOENIX-5677 - Replace System.currentTimeMillis with EnvironmentEdgeM…

gjacoby126:
…anager in non-test code (Addendum to fix IndexScrutinyIT)

### DIFF
--- a/phoenix-core/src/it/java/org/apache/phoenix/end2end/index/PartialIndexRebuilderIT.java
+++ b/phoenix-core/src/it/java/org/apache/phoenix/end2end/index/PartialIndexRebuilderIT.java
@@ -65,20 +65,17 @@ import org.apache.phoenix.util.IndexUtil;
 import org.apache.phoenix.util.MetaDataUtil;
 import org.apache.phoenix.util.ReadOnlyProps;
 import org.apache.phoenix.util.Repeat;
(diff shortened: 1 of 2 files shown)


gjacoby126:
So, the errors came from 2 sources. 
1. IndexScrutinyMapperForTest is really test code, inheriting from EnvironmentEdge, but is in production code, so 5677 converted it over to use EnvironmentEdgeManger incorrectly, causing infinite recursion. 

2. PartialIndexRebuilderIT has a strange test, testIndexActiveIfRegionMovesWhileRebuilding, which was apparently passing before these changes but is failing now. It's multi-threaded, delicately pieced together using both Thread sleeps AND the edge manager. More importantly, it seems to be asserting something incorrect: that the partial rebuilder should succeed and make a disabled (old-style) index ACTIVE even if a region gets unassigned (and not re-assigned) in the middle of the rebuild run. But then the MR job should fail and the index shouldn't be reactivated. So since I don't understand the purpose of the test, I've gone ahead and removed it. If someone sees a valid use case to test there, please let me know and I'll try again to make it work. 

These would have been caught pre-commit, but the Jenkins job was broken for a few days OOMing on each run. 

@kadirozde  @lhofhansl

PR #691, opened 2020-01-24
PHOENIX-5694 Add MR job counters for IndexTool inline verification

kadirozde:


phoenix-core/src/main/java/org/apache/phoenix/mapreduce/index/IndexTool.java
```diff
@@ -641,18 +668,27 @@ public Job getJob() {
...
+            admin.createTable(tableDescriptor);
+        }
+        if (!admin.tableExists(TableName.valueOf(RESULT_TABLE_NAME))) {
+            HTableDescriptor tableDescriptor = new
+                    HTableDescriptor(TableName.valueOf(RESULT_TABLE_NAME));
+            tableDescriptor.setValue("DISABLE_TABLE_SOR", "true");
+            tableDescriptor.setValue(HColumnDescriptor.TTL, String.valueOf(MetaDataProtocol.DEFAULT_LOG_TTL));
+            HColumnDescriptor columnDescriptor = new HColumnDescriptor(RESULT_TABLE_COLUMN_FAMILY);
+            tableDescriptor.addFamily(columnDescriptor);
+            admin.createTable(tableDescriptor);
```
kadirozde: I will remove tableDescriptor.setValue("DISABLE_TABLE_SOR", "true") lines above.

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/IndexRebuildRegionScanner.java
```diff
@@ -89,6 +94,319 @@
...
+            return false;
+        }
+
+        private long getValue(Cell cell) {
+            return Long.parseLong(Bytes.toString(cell.getValueArray(),
+                    cell.getValueOffset(), cell.getValueLength()));
+        }
+
+        private void update(Cell cell) {
+            if (isScannedDataRowCount(cell)) {
```
gjacoby126: if this method knew what phase it was in you could eliminate half the comparisons, right?

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/IndexRebuildRegionScanner.java
```diff
@@ -187,13 +509,56 @@ public RegionInfo getRegionInfo() {
     @Override
     public boolean isFilterDone() { return hasMore; }
 
+    private void logToIndexToolResultTable() throws IOException {
+        long scanMaxTs = scan.getTimeRange().getMax();
+        byte[] keyPrefix = Bytes.toBytes(Long.toString(scanMaxTs));
+        byte[] regionName = Bytes.toBytes(region.getRegionInfo().getRegionNameAsString());
+        byte[] rowKey = new byte[keyPrefix.length + regionName.length];
+        Bytes.putBytes(rowKey, 0, keyPrefix, 0, keyPrefix.length);
```
gjacoby126: Comment here that PK is scanMaxTs + regionName would be helpful. Or alternately, extract key building into a helper method.

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/IndexRebuildRegionScanner.java
```diff
@@ -298,69 +655,71 @@ private boolean checkIndexRow(final byte[] indexRowKey, final Put put) throws IO
...
         final byte[] E_VALUE_PREFIX_BYTES = Bytes.toBytes(" E:");
         final byte[] A_VALUE_PREFIX_BYTES = Bytes.toBytes(" A:");
         final int PREFIX_LENGTH = 3;
         final int TOTAL_PREFIX_LENGTH = 6;
+        final byte[] PHASE_BEFORE_VALUE = Bytes.toBytes("BEFORE");
+        final byte[] PHASE_AFTER_VALUE = Bytes.toBytes("AFTER");
+        long scanMaxTs = scan.getTimeRange().getMax();
+        byte[] keyPrefix = Bytes.toBytes(Long.toString(scanMaxTs));
+        byte[] rowKey;
+        if (dataRowKey != null) {
```
gjacoby126: Again, good to document what the PK schema is.

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/index/IndexTool.java
```diff
@@ -641,18 +668,27 @@ public Job getJob() {
         return job;
     }
 
-    private void createIndexToolOutputTable(Connection connection) throws Exception {
+    private void createIndexToolTables(Connection connection) throws Exception {
         ConnectionQueryServices queryServices = connection.unwrap(PhoenixConnection.class).getQueryServices();
         Admin admin = queryServices.getAdmin();
-        if (admin.tableExists(TableName.valueOf(OUTPUT_TABLE_NAME))) {
-            return;
+        if (!admin.tableExists(TableName.valueOf(OUTPUT_TABLE_NAME))) {
```
gjacoby126: will this work on a cluster with namespaces enabled?

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/index/PhoenixIndexImportDirectReducer.java
```diff
@@ -40,15 +43,79 @@
...
+        }
+    }
+
+    @Override
+    protected void reduce(ImmutableBytesWritable arg0, Iterable<IntWritable> arg1,
+                          Reducer<ImmutableBytesWritable, IntWritable, NullWritable, NullWritable>.Context context)
+            throws IOException, InterruptedException
+
+    {
+        synchronized (this) {
```
gjacoby126: can use an AtomicBoolean instead of a synchronized block.

phoenix-core/src/it/java/org/apache/phoenix/end2end/IndexToolIT.java
```diff
@@ -521,17 +552,14 @@ public void testIndexToolVerifyAfterOption() throws Exception {
             assertTrue(Bytes.compareTo(cell.getValueArray(), cell.getValueOffset(), cell.getValueLength(),
                     expectedValueBytes, 0, expectedValueBytes.length) == 0);
             IndexRegionObserver.setIgnoreIndexRebuildForTesting(false);
-            Admin admin = conn.unwrap(PhoenixConnection.class).getQueryServices().getAdmin();
-            TableName indexToolOutputTable = TableName.valueOf(IndexTool.OUTPUT_TABLE_NAME_BYTES);
-            admin.disableTable(indexToolOutputTable);
-            admin.deleteTable(indexToolOutputTable);
+            dropIndexToolTables(conn);
```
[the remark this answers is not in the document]
kadirozde: Some tests do nothing for some parameters and these tables will not be created for them.

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/IndexRebuildRegionScanner.java
```diff
@@ -89,6 +94,319 @@
...
+        private void addAfterRebuildMissingIndexRowCount(long count) {
+            after.missingIndexRowCount += count;
+        }
+
+        private void addAfterRebuildInvalidIndexRowCount(long count) {
+            after.invalidIndexRowCount += count;
+        }
+
+        private static boolean isScannedDataRowCount(Cell cell) {
+            if (Bytes.compareTo(cell.getQualifierArray(), cell.getQualifierOffset(), cell.getQualifierLength(),
```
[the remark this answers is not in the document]
kadirozde: Good suggestion. I will also eliminate isXXXX(Cell cell) methods altogether.

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/IndexRebuildRegionScanner.java
```diff
@@ -89,6 +94,319 @@
...
+                addAfterRebuildMissingIndexRowCount(getValue(cell));
+            } else if (isAfterRebuildInvalidIndexRowCount(cell)) {
+                addAfterRebuildInvalidIndexRowCount(getValue(cell));
+            }
+        }
+
+        public static byte[] calculateTheClosestNextRowKeyForPrefix(byte[] rowKeyPrefix) {
+            // Essentially we are treating it like an 'unsigned very very long' and doing +1 manually.
+            // Search for the place where the trailing 0xFFs start
+            int offset = rowKeyPrefix.length;
```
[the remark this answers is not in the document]
kadirozde: I got the calculateTheClosestNextRowKeyForPrefix method from HBase 3.0. When we start using HBase 3.0, we can remove this local copy. It seems Bytes.incrementBytes does something similar but not exactly the same, I think.

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/IndexRebuildRegionScanner.java
```diff
@@ -89,6 +94,319 @@
...
+            for (Result result = scanner.next(); result != null; result = scanner.next()) {
+                for (Cell cell : result.rawCells()) {
+                    verificationResult.update(cell);
+                }
+            }
+            return verificationResult;
+        }
+
+        public boolean isVerificationFailed(IndexTool.IndexVerifyType verifyType) {
+            if (verifyType == IndexTool.IndexVerifyType.BEFORE || verifyType == IndexTool.IndexVerifyType.NONE) {
```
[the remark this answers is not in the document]
kadirozde: BEFORE means build only invalid or missing rows. Here the verification is not for verifying the rows that have been rebuilt. Instead, it is for identifying the rows to be rebuilt.

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/IndexRebuildRegionScanner.java
kadirozde: This method is called only once and by the reducer. It is not called by the mapper which implements the phases.

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/IndexRebuildRegionScanner.java
kadirozde: I will add the comment you suggested

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/IndexRebuildRegionScanner.java
kadirozde: I will do that

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/index/IndexTool.java
kadirozde: It should work. Tables can be created without specifying a namespace. Then they are created within the default namespace and the default namespace is not specified in commands/operations as far as I understand.

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/index/PhoenixIndexImportDirectReducer.java
kadirozde: I will use it

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/IndexRebuildRegionScanner.java
```diff
@@ -680,20 +609,21 @@ private void logToIndexToolOutputTable(byte[] dataRowKey, byte[] indexRowKey, lo
             rowKey = new byte[keyPrefix.length];
             Bytes.putBytes(rowKey, 0, keyPrefix, 0, keyPrefix.length);
         }
+        //TODO Write timestamps as strings instead of longs
```
gjacoby126: Do you still intend to do this TODO and change the longs to strings? If so, why not do it now, and if not, why not get rid of the TODO? :-)

kadirozde: I have done with it but forgot to remove the comment. I will remove it

PR #697, opened 2020-01-31
PHOENIX-5706 IndexTool verification reports failure when data row has…

kadirozde:
… no covered column values

This bug was introduced in the previous patch. So, I revert it and added a test to catch such a problem.


phoenix-core/src/main/java/org/apache/phoenix/coprocessor/IndexRebuildRegionScanner.java
```diff
@@ -682,9 +682,8 @@ private boolean verifySingleIndexRow(Result indexRow, final Put dataRow) throws
         Put indexPut = indexMaintainer.buildUpdateMutation(GenericKeyValueBuilder.INSTANCE,
                 valueGetter, new ImmutableBytesWritable(dataRow.getRow()), ts, null, null);
         if (indexPut == null) {
-            String errorMsg = "Empty index update";
-            logToIndexToolOutputTable(dataRow.getRow(), indexRow.getRow(), ts, getMaxTimestamp(indexRow), errorMsg);
-            return false;
+            // This means the dat row does not have any covered column values
```
tkhurana: Nit: the comment should be `data row`

PR #699, opened 2020-02-01
PHOENIX-5496 Ensure that we handle all server-side mutation codes on the client

nsgupta1:
Hi @abhishek-chouhan , @sandeep I've added initial patch for handling server side mutation code on Phoenix client for create table method. Can you please review this PR ?
cc: @ChinmaySKulkarni


phoenix-core/src/test/java/org/apache/phoenix/schema/MetaDataClientTest.java
```diff
@@ -0,0 +1,54 @@
...
+public class MetaDataClientTest extends BaseConnectionlessQueryTest {
+    private static final Logger LOGGER = LoggerFactory.getLogger(MetaDataClientTest.class);
+
+    @Test
+    public void testCreateTable() throws SQLException {
+        String schema = generateUniqueName();
+        String baseTable = generateUniqueName();
+        PhoenixConnection phxConn = (PhoenixConnection) DriverManager.getConnection(getUrl());
+        MetaDataClient client = new MetaDataClient(phxConn);
+        MetaDataClient spyClient = Mockito.spy(client);
```
swaroopak: using spy implies code smell and hence should be avoided. Could you try with dependency injection?

phoenix-core/src/main/java/org/apache/phoenix/exception/SQLExceptionCode.java
```diff
@@ -422,7 +422,8 @@ public SQLException newException(SQLExceptionInfo info) {
     UNABLE_TO_UPDATE_PARENT_TABLE(1143, "XCL43", "Error Updating the parent table"),
     UNABLE_TO_DELETE_CHILD_LINK(1144, "XCL44", "Error deleting parent-child link (Link type=" +
             PTable.LinkType.CHILD_TABLE + ") for view"),
-
+    TABLE_NOT_IN_REGION(1145, "XCL45", "Modifying this table is not allowed. "
```
swaroopak: nit: "No modifications allowed on this table"

phoenix-core/src/main/java/org/apache/phoenix/schema/MetaDataClient.java
```diff
@@ -1959,7 +1959,7 @@ private void populateFamilyPropsList(Map<String, PName> familyNames, Map<String,
         }
     }
 
-    private PTable createTableInternal(CreateTableStatement statement, byte[][] splits,
+    public PTable createTableInternal(CreateTableStatement statement, byte[][] splits,
```
swaroopak: why was this made public? if for the test, please add an annotation.

phoenix-core/src/main/java/org/apache/phoenix/schema/MetaDataClient.java
```diff
@@ -2980,126 +3086,45 @@ public boolean isViewReferenced() {
...
             case AUTO_PARTITION_SEQUENCE_NOT_FOUND:
-                throw new SQLExceptionInfo.Builder(SQLExceptionCode.AUTO_PARTITION_SEQUENCE_UNDEFINED)
-                .setSchemaName(schemaName).setTableName(tableName).build().buildException();
             case CANNOT_COERCE_AUTO_PARTITION_ID:
-                throw new SQLExceptionInfo.Builder(SQLExceptionCode.CANNOT_COERCE_AUTO_PARTITION_ID)
-                .setSchemaName(schemaName).setTableName(tableName).build().buildException();
+            case UNABLE_TO_CREATE_CHILD_LINK:
+            case PARENT_TABLE_NOT_FOUND:
+            case TABLE_NOT_IN_REGION:
+                throwsSQLExceptionUtil(String.valueOf(code), schemaName, tableName);
```
swaroopak: nice simplification!

phoenix-core/src/main/java/org/apache/phoenix/schema/MetaDataClient.java
```diff
@@ -2980,126 +3086,45 @@ public boolean isViewReferenced() {
...
-                addTableToCache(result);
-                return table;
-            }
-        } finally {
-            connection.setAutoCommit(wasAutoCommit);
-            deleteMutexCells(parentPhysicalSchemaName, parentPhysicalTableName, acquiredColumnMutexSet);
+                // Cannot use SQLExecptionInfo here since not all mutation codes have their
+                // corresponding codes in the enum SQLExceptionCode
+                throw new SQLException("Exception occurred while creating table Mutation code:"
+                        + code.toString());
```
swaroopak: nit: space and ,

phoenix-core/src/test/java/org/apache/phoenix/schema/MetaDataClientTest.java
```diff
@@ -7,48 +7,45 @@
...
-import static org.mockito.Matchers.any;
+import static junit.framework.TestCase.assertTrue;
+import static junit.framework.TestCase.fail;
 
 public class MetaDataClientTest extends BaseConnectionlessQueryTest {
-    private static final Logger LOGGER = LoggerFactory.getLogger(MetaDataClientTest.class);
 
     @Test
-    public void testCreateTable() throws SQLException {
+    public void TestHandleCreateTableMutationCode() throws SQLException {
```
swaroopak: nit: small t

phoenix-core/src/test/java/org/apache/phoenix/schema/MetaDataClientTest.java
```diff
@@ -7,48 +7,45 @@
...
         CreateTableStatement stmt = (CreateTableStatement)new SQLParser((ddlFormat)).parseStatement();
 
         try {
-            Mockito.when(spyClient.getCreateTableMutationResult(Mockito.anyList(),
-                any(PTable.ViewType.class),Mockito.anyBoolean(), Mockito.anyList(),
-                any(PTableType.class), Mockito.anyMap(), Mockito.anyList(), any(byte[][].class),
-                Mockito.anyBoolean(), any(PTable.class))).thenReturn(result);
-            spyClient.createTableInternal(stmt, null, null, null,null, PDataType.fromLiteral("BIGINT"),
-               null, null, false, null, null, new HashMap<>(), new HashMap<>());
+            mockClient.handleCreateTableMutationCode(result, result.getMutationCode(), stmt, schema, baseTable, null);
```
swaroopak: Is this line and line 44 longer than 100?

phoenix-core/src/main/java/org/apache/phoenix/schema/MetaDataClient.java
swaroopak: I mean` , `in place of `+` but that's fine since this is not an exception stack.

phoenix-core/src/main/java/org/apache/phoenix/schema/MetaDataClient.java
```diff
@@ -2964,18 +2965,11 @@ public boolean isViewReferenced() {
...
-                    if(result.getTable() != null){
-                        addTableToCache(result);
-                    }
-                    if(statement.ifNotExists()) {
-                        throw new TableAlreadyExistsException(schemaName, tableName, result.getTable());
-                    }
+                boolean mutationResult =  handleCreateTableMutationCode(result, code, statement,
+                        schemaName, tableName, parent);
+                if(mutationResult)
                     return null;
```
swaroopak: nit: please put braces around the `if` block

phoenix-core/src/main/java/org/apache/phoenix/schema/MetaDataClient.java
```diff
@@ -3074,23 +3068,36 @@ public boolean isViewReferenced() {
...
                 }
+                return false;
             case UNALLOWED_TABLE_MUTATION:
                 throw new SQLExceptionInfo.Builder(SQLExceptionCode.CANNOT_MUTATE_TABLE)
                        .setSchemaName(schemaName).setTableName(tableName).build().buildException();
             case CONCURRENT_TABLE_MUTATION:
                 addTableToCache(result);
                 throw new ConcurrentTableMutationException(schemaName, tableName);
             case AUTO_PARTITION_SEQUENCE_NOT_FOUND:
+                throw new SQLExceptionInfo.Builder(SQLExceptionCode.AUTO_PARTITION_SEQUENCE_UNDEFINED)
```
swaroopak: also, why not use throwsSQLExceptionUtil method for this and other cases? If it is increasing the scope of your jira at least for the cases that are not added/modified in this PR,  could you please file a new Jira?

phoenix-core/src/test/java/org/apache/phoenix/schema/MetaDataClientTest.java
swaroopak: I would break this test into two parts, may consider further refactoring to check the exception code passed in as a parameter to a function used for both the tests.

phoenix-core/src/test/java/org/apache/phoenix/schema/MetaDataClientTest.java
nsgupta1: Yes, I'll fix that.

phoenix-core/src/main/java/org/apache/phoenix/schema/MetaDataClient.java
nsgupta1: Created new [JIRA](https://issues.apache.org/jira/browse/PHOENIX-5722) to fix underlying issue in SQLExceptionCode.java

phoenix-core/src/main/java/org/apache/phoenix/schema/MetaDataClient.java
```diff
@@ -2957,149 +2958,180 @@ public boolean isViewReferenced() {
...
                         QueryServices.FORCE_ROW_KEY_ORDER_ATTRIB, QueryServicesOptions.DEFAULT_FORCE_ROW_KEY_ORDER));
             }
-            MetaDataMutationResult result = connection.getQueryServices().createTable(
-                    tableMetaData,
-                    viewType == ViewType.MAPPED || allocateIndexId ? physicalNames.get(0).getBytes() : null,
-                    tableType, tableProps, familyPropList, splits, isNamespaceMapped, allocateIndexId,
-                    UpgradeUtil.isNoUpgradeSet(connection.getClientInfo()), parent);
+
+			// Modularized this code for unit testing
+            MetaDataMutationResult result = getCreateTableMutationResult(tableMetaData, viewType,
```
abhishek-chouhan: Why do we need a separate public method for this? Are there any other places we envision it being used?

phoenix-core/src/main/java/org/apache/phoenix/schema/MetaDataClient.java
```diff
@@ -2957,149 +2958,180 @@ public boolean isViewReferenced() {
...
+                    allocateIndexId, physicalNames, tableType, tableProps, familyPropList, splits,
+                    isNamespaceMapped, parent);
             MutationCode code = result.getMutationCode();
-            switch(code) {
+            if (code != MutationCode.TABLE_NOT_FOUND) {
+                boolean mutationResult =  handleCreateTableMutationCode(result, code, statement,
```
abhishek-chouhan: mutationResult doesn't look like the correct name from readability perspective. Something like tableAreadyExists might be better. From a glance at the function it looks like it returns true if thats the case.

phoenix-core/src/main/java/org/apache/phoenix/schema/MetaDataClient.java
```diff
@@ -2957,149 +2958,180 @@ public boolean isViewReferenced() {
...
+            addTableToCache(result);
+            return table;
+        } finally {
+            connection.setAutoCommit(wasAutoCommit);
+            deleteMutexCells(parentPhysicalSchemaName, parentPhysicalTableName, acquiredColumnMutexSet);
+        }
+    }
+
+    @VisibleForTesting
+    public boolean handleCreateTableMutationCode(MetaDataMutationResult result, MutationCode code,
```
abhishek-chouhan: Please add a couple of comments explaining the behavior of this method and when does it return true/false and whats the expected behavior after that.

phoenix-core/src/main/java/org/apache/phoenix/schema/MetaDataClient.java
```diff
@@ -2957,149 +2958,180 @@ public boolean isViewReferenced() {
...
-                result = new MetaDataMutationResult(code, result.getMutationTime(), table, true);
-                addTableToCache(result);
-                return table;
-            }
-        } finally {
-            connection.setAutoCommit(wasAutoCommit);
-            deleteMutexCells(parentPhysicalSchemaName, parentPhysicalTableName, acquiredColumnMutexSet);
+                // Cannot use SQLExecptionInfo here since not all mutation codes have their
+                // corresponding codes in the enum SQLExceptionCode
+                throw new SQLException("Exception occurred while creating table Mutation code, "
```
abhishek-chouhan: We're throwing this exception if none of the cases cover this mutation code, right? The exception message is incoorect in that case, since we're not creating mutation code.

phoenix-core/src/main/java/org/apache/phoenix/schema/MetaDataClient.java
nsgupta1: I was stubbing this function earlier in the unit test MetaDataClientTest class. I've changed the test now I'll remove the separate public method.

nsgupta1:
Hi @abhishek-chouhan @swaroopak I've resolved all comments and attached the patch for Hadoop QA, It runs fine. Can I get approval for this PR ? Also the patch isn't getting apply for other branches, should I open new PR for 4.x branches ?

phoenix-core/src/main/java/org/apache/phoenix/schema/MetaDataClient.java
```diff
@@ -2957,149 +2958,186 @@ public boolean isViewReferenced() {
...
-                        .build();
-                result = new MetaDataMutationResult(code, result.getMutationTime(), table, true);
-                addTableToCache(result);
-                return table;
-            }
-        } finally {
-            connection.setAutoCommit(wasAutoCommit);
-            deleteMutexCells(parentPhysicalSchemaName, parentPhysicalTableName, acquiredColumnMutexSet);
+                // Cannot use SQLExecptionInfo here since not all mutation codes have their
+                // corresponding codes in the enum SQLExceptionCode
```
abhishek-chouhan: Just noticed we already have a sqlexceptioncode for this. Checkout https://github.com/apache/phoenix/blob/master/phoenix-core/src/main/java/org/apache/phoenix/schema/MetaDataClient.java#L3412 . Can we use the same code here?

phoenix-core/src/test/java/org/apache/phoenix/schema/MetaDataClientTest.java
```diff
@@ -0,0 +1,67 @@
+package org.apache.phoenix.schema;
```
abhishek-chouhan: Can you add the apache license to the file? Lets run a QA after that.

phoenix-core/src/main/java/org/apache/phoenix/schema/MetaDataClient.java
```diff
@@ -2957,149 +2958,187 @@ public boolean isViewReferenced() {
...
+                throw new SQLExceptionInfo.Builder(SQLExceptionCode.UNEXPECTED_MUTATION_CODE)
```
abhishek-chouhan: Would be good to tighten the test and assert on this specific exception being thrown.

abhishek-chouhan:
@nsgupta1 Thanks for seeing this through. Almost there :)

nsgupta1:
@abhishek-chouhan @swaroopak Hadoop QA ran fine for [Master](https://issues.apache.org/jira/browse/PHOENIX-5496?focusedCommentId=17045887&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-17045887) (It says test failed but there are no underlying test failures) and [4.x-Hbase-1.3](https://issues.apache.org/jira/browse/PHOENIX-5496?focusedCommentId=17047143&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-17047143).

nsgupta1:
As discussed offline @abhishek-chouhan I attached patch to master branch twice but it still gives -1 for [tests](https://issues.apache.org/jira/browse/PHOENIX-5496?focusedCommentId=17048455&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-17048455) and there are no test failures when I check the test report.

PR #701, opened 2020-02-02
PHOENIX-5709 Simplify index update generation code for consistent glo…

kadirozde:
…bal indexes

### DIFF
--- a/phoenix-core/src/it/java/org/apache/phoenix/end2end/ConcurrentMutationsExtendedIT.java
+++ b/phoenix-core/src/it/java/org/apache/phoenix/end2end/ConcurrentMutationsExtendedIT.java
@@ -53,6 +53,21 @@ public class ConcurrentMutationsExtendedIT extends ParallelStatsDisabledIT {
 
     private final Object lock = new Object();
 
+    private long verifyIndexTable(String tableName, String indexName, Connection conn) throws Exception {
+        // This checks the state of every raw index row without rebuilding any row
+        IndexToolIT.runIndexTool(true, false, "", tableName, indexName, null,
+                0, IndexTool.IndexVerifyType.ONLY);
+        // This checks the state of an index row after it is repaired
+        long actualRowCount = IndexScrutiny.scrutinizeIndex(conn, tableName, indexName);
+        // We want to check the index rows again as they may be modified by the read repair
+        IndexToolIT.runIndexTool(true, false, "", tableName, indexName, null,
+                0, IndexTool.IndexVerifyType.ONLY);
+        // Now we rebuild the entire index table and expect that it is still good after the full rebuild
+        IndexToolIT.runIndexTool(true, false, "", tableName, indexName, null,
+                0, IndexTool.IndexVerifyType.AFTER);
+        return actualRowCount;
+    }
+
     @Test
     public void testSynchronousDeletesAndUpsertValues() throws Exception {
         final String tableName = generateUniqueName();
@@ -132,11 +147,7 @@ public class ConcurrentMutationsExtendedIT extends ParallelStatsDisabledIT {
         t2.start();
 
         doneSignal.await(60, TimeUnit.SECONDS);
-        IndexToolIT.runIndexTool(true, false, "", tableName, indexName, null,
-                0, IndexTool.IndexVerifyType.ONLY);
-        IndexScrutiny.scrutinizeIndex(conn, tableName, indexName);
-        IndexToolIT.runIndexTool(true, false, "", tableName, indexName, null,
-                0, IndexTool.IndexVerifyType.ONLY);
+        verifyIndexTable(tableName, indexName, conn);
     }
 
     @Test
@@ -197,11 +208,7 @@ public class ConcurrentMutationsExtendedIT extends ParallelStatsDisabledIT {
         t2.start();
 
         doneSignal.await(60, TimeUnit.SECONDS);
-        IndexToolIT.runIndexTool(true, false, "", tableName, indexName, null,
-                0, IndexTool.IndexVerifyType.ONLY);
-        IndexScrutiny.scrutinizeIndex(conn, tableName, indexName);
-        IndexToolIT.runIndexTool(true, false, "", tableName, indexName, null,
-                0, IndexTool.IndexVerifyType.ONLY);
+        verifyIndexTable(tableName, indexName, conn);
     }
 
     @Test @Repeat(5)
@@ -214,10 +221,10 @@ public class ConcurrentMutationsExtendedIT extends ParallelStatsDisabledIT {
         final String indexName = generateUniqueName();
         Connection conn = DriverManager.getConnection(getUrl());
         conn.createStatement().execute("CREATE TABLE " + tableName
-                + "(k1 INTEGER NOT NULL, k2 INTEGER NOT NULL, v1 INTEGER, v2 INTEGER, v3 INTEGER, " +
+                + "(k1 INTEGER NOT NULL, k2 INTEGER NOT NULL, a.v1 INTEGER, b.v2 INTEGER, c.v3 INTEGER, d.v4 INTEGER," +
                 "CONSTRAINT pk PRIMARY KEY (k1,k2))  COLUMN_ENCODED_BYTES = 0, VERSIONS=1");
         TestUtil.addCoprocessor(conn, tableName, DelayingRegionObserver.class);
-        conn.createStatement().execute("CREATE INDEX " + indexName + " ON " + tableName + "(v1) INCLUDE(v2)");
+        conn.createStatement().execute("CREATE INDEX " + indexName + " ON " + tableName + "(v1) INCLUDE(v2, v3)");
         final CountDownLatch doneSignal = new CountDownLatch(nThreads);
         Runnable[] runnables = new Runnable[nThreads];
         for (int i = 0; i < nThreads; i++) {
@@ -232,7 +239,9 @@ public class ConcurrentMutationsExtendedIT extends ParallelStatsDisabledIT {
                             conn.createStatement().execute(
                                     "UPSERT INTO " + tableName + " VALUES (" + (i % nRows) + ", 0, "
                                             + (RAND.nextBoolean() ? null : (RAND.nextInt() % nIndexValues)) + ", "
-                                            + (RAND.nextBoolean() ? null : (RAND.nextInt() % nIndexValues)) + ")");
+                                            + (RAND.nextBoolean() ? null : RAND.nextInt()) + ", "
+                                            + (RAND.nextBoolean() ? null : RAND.nextInt()) + ", "
+                                            + (RAND.nextBoolean() ? null : RAND.nextInt()) + ")");
                             if ((i % batchSize) == 0) {
                                 conn.commit();
                             }
@@ -253,12 +262,8 @@ public class ConcurrentMutationsExtendedIT extends ParallelStatsDisabledIT {
         }
 
         assertTrue("Ran out of time", doneSignal.await(120, TimeUnit.SECONDS));
-        IndexToolIT.runIndexTool(true, false, "", tableName, indexName, null,
-                0, IndexTool.IndexVerifyType.ONLY);
-        long actualRowCount = IndexScrutiny.scrutinizeIndex(conn, tableName, indexName);
+        long actualRowCount = verifyIndexTable(tableName, indexName, conn);
         assertEquals(nRows, actualRowCount);
-        IndexToolIT.runIndexTool(true, false, "", tableName, indexName, null,
-                0, IndexTool.IndexVerifyType.ONLY);
     }
 
     @Test
--- a/phoenix-core/src/it/java/org/apache/phoenix/end2end/ConcurrentMutationsExtendedIT.java
+++ b/phoenix-core/src/it/java/org/apache/phoenix/end2end/ConcurrentMutationsExtendedIT.java
@@ -132,7 +132,11 @@ public class ConcurrentMutationsExtendedIT extends ParallelStatsDisabledIT {
         t2.start();
 
         doneSignal.await(60, TimeUnit.SECONDS);
+        IndexToolIT.runIndexTool(true, false, "", tableName, indexName, null,
+                0, IndexTool.IndexVerifyType.ONLY);
         IndexScrutiny.scrutinizeIndex(conn, tableName, indexName);
+        IndexToolIT.runIndexTool(true, false, "", tableName, indexName, null,
+                0, IndexTool.IndexVerifyType.ONLY);
     }
 
     @Test
@@ -193,7 +197,11 @@ public class ConcurrentMutationsExtendedIT extends ParallelStatsDisabledIT {
         t2.start();
 
         doneSignal.await(60, TimeUnit.SECONDS);
+        IndexToolIT.runIndexTool(true, false, "", tableName, indexName, null,
+                0, IndexTool.IndexVerifyType.ONLY);
         IndexScrutiny.scrutinizeIndex(conn, tableName, indexName);
+        IndexToolIT.runIndexTool(true, false, "", tableName, indexName, null,
+                0, IndexTool.IndexVerifyType.ONLY);
     }
 
     @Test @Repeat(5)
@@ -206,9 +214,10 @@ public class ConcurrentMutationsExtendedIT extends ParallelStatsDisabledIT {
         final String indexName = generateUniqueName();
         Connection conn = DriverManager.getConnection(getUrl());
         conn.createStatement().execute("CREATE TABLE " + tableName
-                + "(k1 INTEGER NOT NULL, k2 INTEGER NOT NULL, v1 INTEGER, CONSTRAINT pk PRIMARY KEY (k1,k2))  COLUMN_ENCODED_BYTES = 0, VERSIONS=1");
+                + "(k1 INTEGER NOT NULL, k2 INTEGER NOT NULL, v1 INTEGER, v2 INTEGER, v3 INTEGER, " +
+                "CONSTRAINT pk PRIMARY KEY (k1,k2))  COLUMN_ENCODED_BYTES = 0, VERSIONS=1");
         TestUtil.addCoprocessor(conn, tableName, DelayingRegionObserver.class);
-        conn.createStatement().execute("CREATE INDEX " + indexName + " ON " + tableName + "(v1)");
+        conn.createStatement().execute("CREATE INDEX " + indexName + " ON " + tableName + "(v1) INCLUDE(v2)");
         final CountDownLatch doneSignal = new CountDownLatch(nThreads);
         Runnable[] runnables = new Runnable[nThreads];
         for (int i = 0; i < nThreads; i++) {
@@ -222,7 +231,8 @@ public class ConcurrentMutationsExtendedIT extends ParallelStatsDisabledIT {
                             int randInt = RAND.nextInt() % nIndexValues;
                             conn.createStatement().execute(
                                     "UPSERT INTO " + tableName + " VALUES (" + (i % nRows) + ", 0, "
-                                            + (isNull ? null : randInt) + ")");
+                                            + (RAND.nextBoolean() ? null : (RAND.nextInt() % nIndexValues)) + ", "
+                                            + (RAND.nextBoolean() ? null : (RAND.nextInt() % nIndexValues)) + ")");
                             if ((i % batchSize) == 0) {
                                 conn.commit();
                             }
@@ -243,6 +253,8 @@ public class ConcurrentMutationsExtendedIT extends ParallelStatsDisabledIT {
         }
 
         assertTrue("Ran out of time", doneSignal.await(120, TimeUnit.SECONDS));
+        IndexToolIT.runIndexTool(true, false, "", tableName, indexName, null,
+                0, IndexTool.IndexVerifyType.ONLY);
         long actualRowCount = IndexScrutiny.scrutinizeIndex(conn, tableName, indexName);
         assertEquals(nRows, actualRowCount);
         IndexToolIT.runIndexTool(true, false, "", tableName, indexName, null,
--- a/phoenix-core/src/it/java/org/apache/phoenix/end2end/ConcurrentMutationsExtendedIT.java
+++ b/phoenix-core/src/it/java/org/apache/phoenix/end2end/ConcurrentMutationsExtendedIT.java
@@ -234,8 +234,6 @@ public class ConcurrentMutationsExtendedIT extends ParallelStatsDisabledIT {
                     try {
                         Connection conn = DriverManager.getConnection(getUrl());
                         for (int i = 0; i < 10000; i++) {
-                            boolean isNull = RAND.nextBoolean();
-                            int randInt = RAND.nextInt() % nIndexValues;
                             conn.createStatement().execute(
                                     "UPSERT INTO " + tableName + " VALUES (" + (i % nRows) + ", 0, "
                                             + (RAND.nextBoolean() ? null : (RAND.nextInt() % nIndexValues)) + ", "
--- a/phoenix-core/src/it/java/org/apache/phoenix/end2end/ConcurrentMutationsExtendedIT.java
+++ b/phoenix-core/src/it/java/org/apache/phoenix/end2end/ConcurrentMutationsExtendedIT.java
@@ -25,7 +25,9 @@ import org.apache.hadoop.hbase.coprocessor.RegionCoprocessorEnvironment;
 import org.apache.hadoop.hbase.coprocessor.SimpleRegionObserver;
 import org.apache.hadoop.hbase.regionserver.MiniBatchOperationInProgress;
 import org.apache.hadoop.hbase.util.Bytes;
+
 import org.apache.phoenix.jdbc.PhoenixConnection;
+import org.apache.phoenix.mapreduce.index.IndexTool;
 import org.apache.phoenix.util.*;
 import org.junit.Ignore;
 import org.junit.Test;
@@ -243,6 +245,8 @@ public class ConcurrentMutationsExtendedIT extends ParallelStatsDisabledIT {
         assertTrue("Ran out of time", doneSignal.await(120, TimeUnit.SECONDS));
         long actualRowCount = IndexScrutiny.scrutinizeIndex(conn, tableName, indexName);
         assertEquals(nRows, actualRowCount);
+        IndexToolIT.runIndexTool(true, false, "", tableName, indexName, null,
+                0, IndexTool.IndexVerifyType.ONLY);
     }
 
     @Test
--- a/phoenix-core/src/it/java/org/apache/phoenix/end2end/ConcurrentMutationsIT.java
+++ b/phoenix-core/src/it/java/org/apache/phoenix/end2end/ConcurrentMutationsIT.java
@@ -266,7 +266,8 @@ public class ConcurrentMutationsIT extends ParallelStatsDisabledIT {
             EnvironmentEdgeManager.injectEdge(null);
         }
     }
-
+    @Ignore ("It is not possible to assign the same timestamp two separately committed mutations in the current model\n" +
+            " except when the server time goes backward. In that case, the behavior is not deterministic")
     @Test
     public void testDeleteRowAndUpsertValueAtSameTS1() throws Exception {
         try {
--- a/phoenix-core/src/main/java/org/apache/phoenix/coprocessor/IndexRebuildRegionScanner.java
+++ b/phoenix-core/src/main/java/org/apache/phoenix/coprocessor/IndexRebuildRegionScanner.java
@@ -19,8 +19,6 @@ package org.apache.phoenix.coprocessor;
 
 import static org.apache.phoenix.hbase.index.IndexRegionObserver.UNVERIFIED_BYTES;
 import static org.apache.phoenix.hbase.index.IndexRegionObserver.VERIFIED_BYTES;
-import static org.apache.phoenix.hbase.index.IndexRegionObserver.getMutationsWithSameTS;
-import static org.apache.phoenix.hbase.index.IndexRegionObserver.prepareIndexMutationsForRebuild;
 import static org.apache.phoenix.hbase.index.write.AbstractParallelWriterIndexCommitter.INDEX_WRITER_KEEP_ALIVE_TIME_CONF_KEY;
 import static org.apache.phoenix.mapreduce.index.IndexTool.AFTER_REBUILD_EXPIRED_INDEX_ROW_COUNT_BYTES;
 import static org.apache.phoenix.mapreduce.index.IndexTool.AFTER_REBUILD_INVALID_INDEX_ROW_COUNT_BYTES;
@@ -52,6 +50,7 @@ import java.util.Map;
 import java.util.NavigableSet;
 import java.util.concurrent.ExecutionException;
 
+import com.google.common.collect.Lists;
 import org.apache.hadoop.conf.Configuration;
 import org.apache.hadoop.hbase.Cell;
 import org.apache.hadoop.hbase.CellUtil;
@@ -87,6 +86,9 @@ import org.apache.phoenix.hbase.index.parallel.TaskRunner;
 import org.apache.phoenix.hbase.index.parallel.ThreadPoolBuilder;
 import org.apache.phoenix.hbase.index.parallel.ThreadPoolManager;
 import org.apache.phoenix.hbase.index.parallel.WaitForCompletionTaskRunner;
+import org.apache.phoenix.hbase.index.util.GenericKeyValueBuilder;
+import org.apache.phoenix.hbase.index.util.ImmutableBytesPtr;
+import org.apache.phoenix.hbase.index.util.IndexManagementUtil;
 import org.apache.phoenix.index.GlobalIndexChecker;
 import org.apache.phoenix.index.IndexMaintainer;
 import org.apache.phoenix.index.PhoenixIndexCodec;
@@ -366,9 +368,9 @@ public class IndexRebuildRegionScanner extends BaseRegionScanner {
     private int indexTableTTL;
     private VerificationResult verificationResult;
     private boolean isBeforeRebuilt = true;
-    boolean partialRebuild = false;
-    long singleRowRebuildReturnCode;
-    Map<byte[], NavigableSet<byte[]>> familyMap;
+    private boolean partialRebuild = false;
+    private int  singleRowRebuildReturnCode;
+    private Map<byte[], NavigableSet<byte[]>> familyMap;
     private byte[][] viewConstants;
 
     IndexRebuildRegionScanner(final RegionScanner innerScanner, final Region region, final Scan scan,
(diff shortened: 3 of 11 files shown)


phoenix-core/src/main/java/org/apache/phoenix/hbase/index/IndexRegionObserver.java
```diff
@@ -533,13 +666,139 @@ private void handleLocalIndexUpdates(ObserverContext<RegionCoprocessorEnvironmen
...
+                    ValueGetter mergedRowVG = new IndexRebuildRegionScanner.SimpleValueGetter(mergedRow);
+                    Put indexPut = indexMaintainer.buildUpdateMutation(GenericKeyValueBuilder.INSTANCE,
+                            mergedRowVG, rowKeyPtr, ts, regionStartKey, regionEndKey);
+                    if (indexPut == null) {
+                        // No covered column. Just prepare an index row with the empty column
+                        byte[] indexRowKey = indexMaintainer.buildRowKey(mergedRowVG, rowKeyPtr,
+                                regionStartKey, regionEndKey, HConstants.LATEST_TIMESTAMP);
+                        indexPut = new Put(indexRowKey);
+                    }
+                    indexPut.addColumn(indexMaintainer.getEmptyKeyValueFamily().copyBytesIfNecessary(),
```
swaroopak: why shouldn't this be `dataEmptyKeyValueCF` from IndexMaintainer?

phoenix-core/src/main/java/org/apache/phoenix/hbase/index/IndexRegionObserver.java
```diff
@@ -565,12 +829,6 @@ private void prepareIndexMutations(
                   // add the VERIFIED cell, which is the empty cell
```
swaroopak: Since the verified cell is getting added in `prepareIndexMutationsForIndexRebuild`, do you mind moving this comment inside else block or changing it?

phoenix-core/src/main/java/org/apache/phoenix/hbase/index/IndexRegionObserver.java
```diff
@@ -672,15 +927,15 @@ public void preBatchMutateWithExceptions(ObserverContext<RegionCoprocessorEnviro
...
+      Collection<? extends Mutation> pendingMutations = mergeMutations(c, miniBatchOp, context, now, context.rebuild);
       // early exit if it turns out we don't have any edits
-      if (mutations == null) {
+      if (pendingMutations == null || pendingMutations.isEmpty()) {
           return;
       }
 
       long start = EnvironmentEdgeManager.currentTimeMillis();
-      prepareIndexMutations(c, miniBatchOp, context, mutations, now, indexMetaData);
+      preparePreIndexMutations(c, miniBatchOp, context, pendingMutations, now, indexMetaData);
```
gjacoby126: comment specifying that this is preparing the phase 1 unverified index writes would be helpful. Otherwise this sounds like you're preparing mutations for before the index write, which is a bit confusing.

gjacoby126: It's also  a little confusing that preparePreIndexMutations calls prepareIndexMutations. :-)

phoenix-core/src/main/java/org/apache/phoenix/hbase/index/IndexRegionObserver.java
```diff
@@ -409,91 +413,220 @@ private void populatePendingRows(BatchMutateContext context) {
...
-      if (copyMutations || replayWrite != null) {
-          mutations = IndexManagementUtil.flattenMutationsByTimestamp(mutations);
+      Collection<Mutation> pendingMutations = Lists.newArrayListWithExpectedSize(miniBatchOp.size());
+      // Add put mutations into the collection first. This ordering of puts first and then deletes is important for
+      // obtaining correct index updates
+      for (int i = 0; i < miniBatchOp.size(); i++) {
+          Mutation m = miniBatchOp.getOperation(i);
+          if (m instanceof Put) {
+              if (miniBatchOp.getOperationStatus(i) != IGNORE &&
+                      miniBatchOp.getOperationStatus(i) != NOWRITE && this.builder.isEnabled(m)) {
```
gjacoby126: can extract this to a helper method called by both puts and deletes.

phoenix-core/src/main/java/org/apache/phoenix/index/IndexMaintainer.java
```diff
@@ -1592,7 +1592,7 @@ private int estimateIndexRowKeyByteSize(int indexColByteSize) {
      */
     private void initCachedState() {
         byte[] emptyKvQualifier = EncodedColumnsUtil.getEmptyKeyValueInfo(encodingScheme).getFirst();
-        dataEmptyKeyValueRef = new ColumnReference(emptyKeyValueCFPtr.copyBytesIfNecessary(), emptyKvQualifier);
+        dataEmptyKeyValueRef = new ColumnReference(dataEmptyKeyValueCF, emptyKvQualifier);
```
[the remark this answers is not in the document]
kadirozde: @gokceni, if you look at a bit further, you would see the other constructor sets only two attributes while IndexMaintainer has tens of attributes. That constructor is not by itself sufficient to initialize an IndexMaintainer object. It is used to get a starting object then the rest of the attributes is set by the caller of this constructor. So, even if this constructor is used, dataEmptyKeyValueCF is initialized at the end.

phoenix-core/src/main/java/org/apache/phoenix/hbase/index/IndexRegionObserver.java
```diff
@@ -409,91 +413,220 @@ private void populatePendingRows(BatchMutateContext context) {
...
+      }
+  }
+
+    public static long getTimestamp(Mutation m) throws IOException {
+        for (List<Cell> cells : m.getFamilyCellMap().values()) {
+            for (Cell cell : cells) {
+                return cell.getTimestamp();
+            }
+        }
+        return 0;
```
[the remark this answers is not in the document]
kadirozde: Every mutation must have at least one cell. Yes, it is a good idea to throw exception.

phoenix-core/src/main/java/org/apache/phoenix/hbase/index/IndexRegionObserver.java
```diff
@@ -409,91 +413,220 @@ private void populatePendingRows(BatchMutateContext context) {
...
+  private static void removeColumn(Put put, Cell deleteCell) {
+      byte[] family = CellUtil.cloneFamily(deleteCell);
+      List<Cell> cellList = put.getFamilyCellMap().get(family);
+      if (cellList == null) {
+          return;
+      }
+      Iterator<Cell> cellIterator = cellList.iterator();
+      while (cellIterator.hasNext()) {
+          Cell cell = cellIterator.next();
+          if (Bytes.compareTo(cell.getQualifierArray(), cell.getQualifierOffset(), cell.getQualifierLength(),
```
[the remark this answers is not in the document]
kadirozde: Yes, I will use it.

phoenix-core/src/main/java/org/apache/phoenix/hbase/index/IndexRegionObserver.java
```diff
@@ -409,91 +413,220 @@ private void populatePendingRows(BatchMutateContext context) {
...
+                  deleteCell.getQualifierArray(), deleteCell.getQualifierOffset(), deleteCell.getQualifierLength()) == 0) {
+              cellIterator.remove();
+              if (cellList.isEmpty()) {
+                  put.getFamilyCellMap().remove(family);
+              }
+              return;
+          }
+      }
```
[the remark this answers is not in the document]
kadirozde: We do not know upfront if the family or column delete cell refers to exists in the mutation. It may or may not exist.

phoenix-core/src/main/java/org/apache/phoenix/hbase/index/IndexRegionObserver.java
```diff
@@ -409,91 +413,220 @@ private void populatePendingRows(BatchMutateContext context) {
...
+  }
+
+  private Put mergeNew(Put current, Put previous) throws IOException {
+      Put next = new Put(current);
+      for (List<Cell> cells : previous.getFamilyCellMap().values()) {
+          for (Cell cell : cells) {
+              if (!current.has(CellUtil.cloneFamily(cell), CellUtil.cloneQualifier(cell))) {
+                  next.add(cell);
+              }
+          }
```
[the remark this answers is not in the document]
kadirozde: You cannot reuse it without changing the signature of the method.

phoenix-core/src/main/java/org/apache/phoenix/hbase/index/IndexRegionObserver.java
```diff
@@ -409,91 +413,220 @@ private void populatePendingRows(BatchMutateContext context) {
...
+      }
+      return next;
+  }
+
+  /**
+   * When there are multiple put mutations on the data same row within the same batch, this method merged them into
```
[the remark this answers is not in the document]
kadirozde: ok

phoenix-core/src/main/java/org/apache/phoenix/hbase/index/IndexRegionObserver.java
```diff
@@ -409,91 +413,220 @@ private void populatePendingRows(BatchMutateContext context) {
...
+  private void mergePendingPutMutations(MiniBatchOperationInProgress<Mutation> miniBatchOp,
+                                        Map<ImmutableBytesPtr, Integer> pendingPuts,
+                                        long now) throws IOException {
+      for (Integer i = 0; i < miniBatchOp.size(); i++) {
           if (miniBatchOp.getOperationStatus(i) == IGNORE) {
               continue;
           }
           Mutation m = miniBatchOp.getOperation(i);
+          // skip this mutation if we aren't enabling indexing
           if (this.builder.isEnabled(m)) {
```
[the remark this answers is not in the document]
kadirozde: No, it is defined per mutation not per batch.

phoenix-core/src/main/java/org/apache/phoenix/hbase/index/IndexRegionObserver.java
```diff
@@ -409,91 +413,220 @@ private void populatePendingRows(BatchMutateContext context) {
...
-              // Track whether or not we need to
-              ImmutableBytesPtr row = new ImmutableBytesPtr(m.getRow());
-              if (mutationsMap.containsKey(row)) {
-                  copyMutations = true;
-              } else {
-                  mutationsMap.put(row, null);
+              // Unless we're replaying edits to rebuild the index, we update the time stamp
+              // of the data table to prevent overlapping time stamps (which prevents index
+              // inconsistencies as this case isn't handled correctly currently).
```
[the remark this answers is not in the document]
kadirozde: This method is not called for rebuild mutations.

phoenix-core/src/main/java/org/apache/phoenix/hbase/index/IndexRegionObserver.java
```diff
@@ -409,91 +413,220 @@ private void populatePendingRows(BatchMutateContext context) {
...
+              setTimestamp(m, now);
+              if (m instanceof Put) {
+                  ImmutableBytesPtr rowKeyPtr = new ImmutableBytesPtr(m.getRow());
+                  Integer opIndex = pendingPuts.get(rowKeyPtr);
+                  pendingPuts.put(rowKeyPtr, i);
```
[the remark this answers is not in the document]
kadirozde: It is not safer to wait as this is done under row lock. However, it is more readable if it is done after the merge. So, I will change the order.

phoenix-core/src/main/java/org/apache/phoenix/hbase/index/IndexRegionObserver.java
```diff
@@ -409,91 +413,220 @@ private void populatePendingRows(BatchMutateContext context) {
...
-          // skip this mutation if we aren't enabling indexing
-          // unfortunately, we really should ask if the raw mutation (rather than the combined mutation)
-          // should be indexed, which means we need to expose another method on the builder. Such is the
-          // way optimization go though.
-          if (miniBatchOp.getOperationStatus(i) != IGNORE && this.builder.isEnabled(m)) {
-              if (resetTimeStamp) {
-                  // Unless we're replaying edits to rebuild the index, we update the time stamp
-                  // of the data table to prevent overlapping time stamps (which prevents index
-                  // inconsistencies as this case isn't handled correctly currently).
+          if (!this.builder.isEnabled(m)) {
```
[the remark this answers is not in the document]
kadirozde: See my answer above.

phoenix-core/src/main/java/org/apache/phoenix/hbase/index/IndexRegionObserver.java
```diff
@@ -409,91 +413,220 @@ private void populatePendingRows(BatchMutateContext context) {
...
+              if (opIndex != null || mergedRow != null) {
+                  Put put = (Put) miniBatchOp.getOperation(opIndex);
                   for (List<Cell> cells : m.getFamilyCellMap().values()) {
                       for (Cell cell : cells) {
-                          CellUtil.setTimestamp(cell, now);
+                          switch (cell.getType()) {
+                              case DeleteFamily:
+                              case DeleteFamilyVersion:
+                                  if (put != null)
+                                      put.getFamilyCellMap().remove(CellUtil.cloneFamily(cell));
```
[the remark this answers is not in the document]
kadirozde: Agree. I will change it

phoenix-core/src/main/java/org/apache/phoenix/hbase/index/IndexRegionObserver.java
```diff
@@ -409,91 +413,220 @@ private void populatePendingRows(BatchMutateContext context) {
...
+                                  if (mergedRow != null)
+                                      mergedRow.getFamilyCellMap().remove(CellUtil.cloneFamily(cell));
```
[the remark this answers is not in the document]
kadirozde: No. Because in each case a different object (put vs mergedRow) is updated.

phoenix-core/src/main/java/org/apache/phoenix/hbase/index/IndexRegionObserver.java
```diff
@@ -409,91 +413,220 @@ private void populatePendingRows(BatchMutateContext context) {
...
+              Mutation m = miniBatchOp.getOperation(i);
+              if (this.builder.isEnabled(m)) {
+                  if (m instanceof Put) {
+                      ImmutableBytesPtr rowKeyPtr = new ImmutableBytesPtr(m.getRow());
+                      Put currentRow = context.currentRowStates.get(rowKeyPtr);
+                      Put mergedRow;
+                      if (currentRow != null) {
+                          mergedRow = mergeNew((Put) m, currentRow);
+                      }
+                      else {
```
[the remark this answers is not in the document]
kadirozde: ok

phoenix-core/src/main/java/org/apache/phoenix/hbase/index/IndexRegionObserver.java
kadirozde: ok

phoenix-core/src/main/java/org/apache/phoenix/hbase/index/IndexRegionObserver.java
```diff
@@ -409,91 +413,220 @@ private void populatePendingRows(BatchMutateContext context) {
...
+          Mutation m = miniBatchOp.getOperation(i);
+          if (m instanceof Delete) {
+              if (miniBatchOp.getOperationStatus(i) != IGNORE &&
+                      miniBatchOp.getOperationStatus(i) != NOWRITE && this.builder.isEnabled(m)) {
+                  if (rebuild) {
+                      miniBatchOp.setOperationStatus(i, NOWRITE);
+                  }
+                  pendingMutations.add(m);
+              }
+          }
```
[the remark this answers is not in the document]
kadirozde: ok.

phoenix-core/src/main/java/org/apache/phoenix/hbase/index/IndexRegionObserver.java
```diff
@@ -533,13 +666,139 @@ private void handleLocalIndexUpdates(ObserverContext<RegionCoprocessorEnvironmen
...
+        SkipScanFilter skipScanFilter = scanRanges.getSkipScanFilter();
+        scan.setFilter(skipScanFilter);
+        context.currentRowStates = new HashMap<ImmutableBytesPtr, Put>(context.rowsToLock.size());
+        try (RegionScanner scanner = c.getEnvironment().getRegion().getScanner(scan)) {
+            boolean more = true;
+            while(more) {
+                List<Cell> cells = new ArrayList<Cell>();
+                more = scanner.next(cells);
+                if (cells.isEmpty()) {
+                    continue;
```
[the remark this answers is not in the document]
kadirozde: It is possible only if there are infinitely many rows :-). I am not doing anything here different than many other places in Phoenix where the local table region is scanned.

phoenix-core/src/main/java/org/apache/phoenix/hbase/index/IndexRegionObserver.java
```diff
@@ -533,13 +666,139 @@ private void handleLocalIndexUpdates(ObserverContext<RegionCoprocessorEnvironmen
...
+            HTableInterfaceReference hTableInterfaceReference =
+                    new HTableInterfaceReference(new ImmutableBytesPtr(indexMaintainer.getIndexTableName()));
+            for (Mutation mutation : pendingMutations) {
+                ImmutableBytesPtr rowKeyPtr = new ImmutableBytesPtr(mutation.getRow());
+                ValueGetter valueGetter = new IndexRebuildRegionScanner.SimpleValueGetter((Put) mutation);
+                Put indexPut = indexMaintainer.buildUpdateMutation(GenericKeyValueBuilder.INSTANCE, valueGetter,
+                        rowKeyPtr, getMaxTimestamp(mutation), null, null);
+                if (indexPut == null) {
+                    byte[] indexRowKey = indexMaintainer.buildRowKey(valueGetter, rowKeyPtr,
+                            null, null, HConstants.LATEST_TIMESTAMP);
```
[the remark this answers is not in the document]
kadirozde: The timestamp is eventually passed to valueGetter by buildRowKey. valueGetter ignores timestamp here as we are getting cells from one mutation, see the above statement : 
ValueGetter valueGetter = new IndexRebuildRegionScanner.SimpleValueGetter((Put) mutation);

phoenix-core/src/main/java/org/apache/phoenix/hbase/index/IndexRegionObserver.java
```diff
@@ -533,13 +666,139 @@ private void handleLocalIndexUpdates(ObserverContext<RegionCoprocessorEnvironmen
...
+                    indexPut = new Put(indexRowKey);
+                }
+                indexPut.addColumn(indexMaintainer.getEmptyKeyValueFamily().copyBytesIfNecessary(),
+                            indexMaintainer.getEmptyKeyValueQualifier(), getMaxTimestamp(mutation), VERIFIED_BYTES);
```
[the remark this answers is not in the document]
kadirozde: No, they are not related.

phoenix-core/src/main/java/org/apache/phoenix/hbase/index/IndexRegionObserver.java
```diff
@@ -533,13 +666,139 @@ private void handleLocalIndexUpdates(ObserverContext<RegionCoprocessorEnvironmen
...
+     * Generate the index update for a data row from the mutation that are obtained by merging the previous data row
+     * state with the pending row mutation.
+     */
+    private void prepareIndexMutations(ObserverContext<RegionCoprocessorEnvironment> c,
+                                       BatchMutateContext context,
+                                       List<IndexMaintainer> maintainers,
+                                       Collection<? extends Mutation> pendingMutations) throws IOException {
+        for (IndexMaintainer indexMaintainer : maintainers) {
+            HTableInterfaceReference hTableInterfaceReference =
+                    new HTableInterfaceReference(new ImmutableBytesPtr(indexMaintainer.getIndexTableName()));
```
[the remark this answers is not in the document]
kadirozde: I will check it.

phoenix-core/src/main/java/org/apache/phoenix/hbase/index/IndexRegionObserver.java
```diff
@@ -533,13 +666,139 @@ private void handleLocalIndexUpdates(ObserverContext<RegionCoprocessorEnvironmen
...
+            if (indexMaintainer.isLocalIndex()) {
+                regionStartKey = c.getEnvironment().getRegion().getRegionInfo().getStartKey();
+                regionEndKey = c.getEnvironment().getRegion().getRegionInfo().getEndKey();
+            }
+            for (Mutation mutation : pendingMutations) {
+                ImmutableBytesPtr rowKeyPtr = new ImmutableBytesPtr(mutation.getRow());
+                Put currentRow = context.currentRowStates.get(rowKeyPtr);
+                ValueGetter currentRowVG = (currentRow == null) ? null :
+                        new IndexRebuildRegionScanner.SimpleValueGetter(currentRow);
+                long ts = getTimestamp(mutation);
```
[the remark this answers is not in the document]
kadirozde: No. Here all the cell within a mutation have the same timestamps.

phoenix-core/src/main/java/org/apache/phoenix/hbase/index/IndexRegionObserver.java
kadirozde: The location of the comment is correct. I will change the comment to:
// Add an empty column cell with the verify status

phoenix-core/src/main/java/org/apache/phoenix/hbase/index/IndexRegionObserver.java
kadirozde: ok

phoenix-core/src/main/java/org/apache/phoenix/hbase/index/IndexRegionObserver.java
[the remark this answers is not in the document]
gokceni: Sorry, did not get this one. Why do we need a signature change?
My suggestion is this:
Put next = new Put(current);
merge(next, previous);
return next;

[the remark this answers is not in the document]
kadirozde: Got it. Thx. I will do that.

phoenix-core/src/main/java/org/apache/phoenix/hbase/index/IndexRegionObserver.java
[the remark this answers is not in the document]
kadirozde: Yes, we have tests

phoenix-core/src/main/java/org/apache/phoenix/hbase/index/IndexRegionObserver.java
```diff
@@ -409,93 +416,262 @@ private void populatePendingRows(BatchMutateContext context) {
...
-                                                        long now, ReplayWrite replayWrite) throws IOException {
-      Map<ImmutableBytesPtr, MultiMutation> mutationsMap = new HashMap<>();
-      boolean copyMutations = false;
-      for (int i = 0; i < miniBatchOp.size(); i++) {
-          if (miniBatchOp.getOperationStatus(i) == IGNORE) {
-              continue;
+  public static void setTimestamp(Mutation m, long ts) throws IOException {
+      for (List<Cell> cells : m.getFamilyCellMap().values()) {
+          for (Cell cell : cells) {
+              CellUtil.setTimestamp(cell, ts);
```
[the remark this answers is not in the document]
kadirozde: In preBatchMutate, we not only set the timestamps for the cells but also set the status of some mutations to NO_WRITE. This has been the case all along (in the old indexer and new indexer). The timestamp for these cell has been set yet. So, they cannot be in Memstore as far as I know. @abhishek-chouhan, would you be able to verify this?

phoenix-core/src/main/java/org/apache/phoenix/hbase/index/IndexRegionObserver.java
```diff
@@ -409,93 +416,262 @@ private void populatePendingRows(BatchMutateContext context) {
...
+      // Add delete mutations into the collection (after the put mutations if any)
+      populateMutationList(miniBatchOp, pendingMutations, false);
+      return pendingMutations;
   }
 
+  /**
+   * This is to reorder the mutations in ascending order by the tuple of row key, timestamp and mutation type where
+   * delete comes before put
+   */
+  public static final Comparator<Mutation> MUTATION_COMPARATOR = new Comparator<Mutation>() {
```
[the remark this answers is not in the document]
kadirozde: The ordering is very specific to the algorithm here. All I do a simple sorting here using Java sorted list. I am not sure what else I can leverage more.

phoenix-core/src/main/java/org/apache/phoenix/hbase/index/IndexRegionObserver.java
```diff
@@ -533,13 +709,202 @@ private void handleLocalIndexUpdates(ObserverContext<RegionCoprocessorEnvironmen
...
+                }
+                context.currentRowStates.put(new ImmutableBytesPtr(rowKey), put);
+            }
+        }
+    }
+
+    private Collection<Cell> getCollection(Mutation m) {
+        Collection<Cell> collection = new ArrayList<>();
+        for (List<Cell> cells : m.getFamilyCellMap().values()) {
+            collection.addAll(cells);
```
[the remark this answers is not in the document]
kadirozde: IndexMaintainer does not touch the input cells.

phoenix-core/src/it/java/org/apache/phoenix/end2end/ConcurrentMutationsIT.java
```diff
@@ -266,7 +266,8 @@ public void testSetIndexedColumnToNullAndValueAtSameTSWithStoreNulls2() throws E
             EnvironmentEdgeManager.injectEdge(null);
         }
     }
-
+    @Ignore ("It is not possible to assign the same timestamp two separately committed mutations in the current model\n" +
```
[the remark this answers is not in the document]
kadirozde: I want to give the reviewers a chance to review this before removing it. This test is still applicable to the old design. I suggest we remove it when we remove the old design.

phoenix-core/src/main/java/org/apache/phoenix/hbase/index/IndexRegionObserver.java
```diff
@@ -565,12 +829,6 @@ private void prepareIndexMutations(
                   // add the VERIFIED cell, which is the empty cell
                   Mutation m = update.getFirst();
                   if (context.rebuild) {
-                      if (m instanceof Put) {
-                          long ts = getMaxTimestamp(m);
-                          // Remove the empty column prepared by Index codec as we need to change its value
-                          removeEmptyColumn(m, emptyCF, emptyCQ);
-                          ((Put) m).addColumn(emptyCF, emptyCQ, ts, VERIFIED_BYTES);
```
swaroopak: Now that this snippet is added in the next commit, the above comment makes sense too. Thanks.

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/IndexRebuildRegionScanner.java
```diff
@@ -1092,6 +1093,181 @@ private boolean isColumnIncluded(Cell cell) {
...
+            mutationList.add(del);
+        }
+        // Group the cells within a mutation based on their timestamps and create a separate mutation for each group
+        mutationList = (List<Mutation>) IndexManagementUtil.flattenMutationsByTimestamp(mutationList);
+        // Reorder the mutations on the same row so that delete comes before put when they have the same timestamp
+        Collections.sort(mutationList, MUTATION_TS_COMPARATOR);
+        return mutationList;
+    }
+
+    private static Put prepareIndexPut(IndexMaintainer indexMaintainer, ImmutableBytesPtr rowKeyPtr,
```
gjacoby126: any reason not to put this method in the index maintainer itself, maybe as part of buildUpdateMutation? Seems like the maintainer's job is to create index puts, and this logic is part of creating those puts. Are there cases where we would call buildUpdateMutation and _not_ want this logic?

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/IndexRebuildRegionScanner.java
```diff
@@ -1092,6 +1093,181 @@ private boolean isColumnIncluded(Cell cell) {
...
+            ValueGetter currentDataRowVG = (currentDataRow == null) ? null : new IndexRebuildRegionScanner.SimpleValueGetter(currentDataRow);
+            long ts = getTimestamp(mutation);
+            if (mutation instanceof Put) {
+                // Add this put on top of the current data row state to get the next data row state
+                Put nextDataRow = (currentDataRow == null) ? new Put((Put)mutation) : mergeNew((Put)mutation, currentDataRow);
+                ValueGetter nextDataRowVG = new IndexRebuildRegionScanner.SimpleValueGetter(nextDataRow);
+                Put indexPut = prepareIndexPut(indexMaintainer, rowKeyPtr, nextDataRowVG, ts, true, null, null);
+                indexMutations.add(indexPut);
+                // Delete the current index row if the new index key is different than the current one
+                if (IndexRowKeyForCurrentDataRow != null) {
```
gjacoby126: nit: java variable names start with lower case.

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/IndexRebuildRegionScanner.java
```diff
@@ -1092,6 +1093,181 @@ private boolean isColumnIncluded(Cell cell) {
...
+                        del.addFamily(indexMaintainer.getEmptyKeyValueFamily().copyBytesIfNecessary(), ts);
+                        indexMutations.add(del);
+                    }
+                }
+                // For the next iteration
+                currentDataRow = nextDataRow;
+                IndexRowKeyForCurrentDataRow = indexPut.getRow();
+            } else {
+                if (currentDataRow != null) {
+                    // We apply the delete column mutations only on the current data row state here as they are already
```
gjacoby126: really nice comment!

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/IndexRebuildRegionScanner.java
kadirozde: I am not sure how I got the upper case variable name here, probably it happened when I did bulk rename. I will change it.

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/IndexRebuildRegionScanner.java
kadirozde: I can use this method only for rebuilds but not for regular writes so I kept it here.

phoenix-core/src/it/java/org/apache/phoenix/end2end/ConcurrentMutationsExtendedIT.java
```diff
@@ -132,7 +132,11 @@ public void testSynchronousDeletesAndUpsertValues() throws Exception {
         t2.start();
 
         doneSignal.await(60, TimeUnit.SECONDS);
+        IndexToolIT.runIndexTool(true, false, "", tableName, indexName, null,
+                0, IndexTool.IndexVerifyType.ONLY);
```
gjacoby126: Curious why we're verifying twice? IndexScrutiny is read-only.

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/IndexRebuildRegionScanner.java
```diff
@@ -719,9 +721,6 @@ private boolean isMatchingMutation(Mutation expected, Mutation actual) throws IO
     }
 
     private boolean isVerified(Mutation mutation) throws IOException {
-        if (mutation instanceof Delete) {
```
gjacoby126: can deletes be verified? They won't have an empty column, will they?

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/IndexRebuildRegionScanner.java
```diff
@@ -750,16 +749,38 @@ public int compare(Mutation o1, Mutation o2) {
...
+                    return true;
+                }
+            }
+        }
+        return false;
+    }
+
+    /**
+     * indexRow is the set of all cells of all the row version of an index row from the index table. These are actual
+     * actual cells. We group these cells based on timestamp and type (put vs delete), and form the actual set of
```
gjacoby126: nit: "actual actual"

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/IndexRebuildRegionScanner.java
```diff
@@ -823,6 +844,42 @@ private boolean verifySingleIndexRow(Result indexRow, VerificationResult.PhaseRe
...
+                // read the existing row state, and the second is to write to the data table. The processing of
+                // concurrent data mutations are serialized once for reading the existing row states, and then
+                // serialized again for updating data table. In other words, they go through locking twice, i.e.,
+                // [lock, read, unlock] and [lock, write, unlock]. Because of this two phase locking, for a pair of
+                // concurrent mutations(for the same row), the same row state is read from the data table. This means
+                // the same existing index row can be made unverified twice with different timestamps, one for each
+                // the concurrent mutation. So skipping newer expected unverified put mutation are safe. One of these
+                // unverified put mutations is later cleaned up by delete markers by the read repair process.
+                // We can skipped these delete markers safely above. When the data table rows are rebuild, the rebuild
+                // process generates the delete family markers. The timestamp of delete markers are the timestamp of
```
gjacoby126: nit: are rebuilt

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/IndexRebuildRegionScanner.java
```diff
@@ -823,6 +844,42 @@ private boolean verifySingleIndexRow(Result indexRow, VerificationResult.PhaseRe
...
+                // the data table mutation for which the delete marker is added. This means their timestamp will be
+                // higher than the timestamp of actual index row to be deleted.
+                continue;
+            }
+            // For each expected delete (family) mutation there has to be the corresponding actual unverified put
+            // mutation. The corresponding delete mutation may not be there if the last write phase does not
+            // make it. The following two patterns are possible
+            // {(expected delete, actual delete), (expected unverified put, actual unverified put)},
+            // {(expected delete, actual unverified put), (expected unverified put, no corresponding actual put)}
+            if (expectedMutation instanceof Delete && isDeleteFamily(expectedMutation) &&
```
gjacoby126: shouldn't this be isDeleteFamilyVersion, after PHOENIX-5743?

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/IndexRebuildRegionScanner.java
kadirozde: I will change the signature of this method. It is used only for Put mutations.

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/IndexRebuildRegionScanner.java
kadirozde: A DeleteFamilyVersion delete marker is only added by the read repair part implemented by GlobalIndexChecker. It is added only if the unverified put is a new version of an existing index row. The (single row or all row) rebuild adds delete markers only of type DeleteFamily.

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/IndexRebuildRegionScanner.java
kadirozde: ok

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/IndexRebuildRegionScanner.java
kadirozde: ok

phoenix-core/src/it/java/org/apache/phoenix/end2end/ConcurrentMutationsExtendedIT.java
kadirozde: No, IndexScrutiny is not read only anymore since it triggers the read repair now.

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/UngroupedAggregateRegionObserver.java
```diff
@@ -1110,9 +1111,20 @@ private static PTable deserializeTable(byte[] b) {
 
     private RegionScanner rebuildIndices(final RegionScanner innerScanner, final Region region, final Scan scan,
                                          final RegionCoprocessorEnvironment env) throws IOException {
-
-        RegionScanner scanner = new IndexRebuildRegionScanner(innerScanner, region, scan, env, this);
-        return scanner;
+        if (!scan.isRaw()) {
```
swaroopak: nit: assuming this is for rebuilding the history from data, a comment stating so would be useful.

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/IndexRebuildRegionScanner.java
```diff
@@ -927,109 +1123,375 @@ private void verifyAndOrRebuildIndex() throws IOException {
...
+        return collection;
+    }
+
+
+    /**
+     * Generate the index update for a data row from the mutation that are obtained by merging the previous data row
+     * state with the pending row mutation for index rebuild. This method is called only for global indexes.
+     * pendingMutations is a sorted list of data table mutations that are used to replay index table mutations.
+     * This list is sorted in ascending order by the tuple of row key, timestamp and mutation type where delete comes
+     * after put.
```
swaroopak: Little confused here, MUTATION_TS_COMPARATOR puts deletes before puts, comment here says otherwise?

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/IndexRebuildRegionScanner.java
```diff
@@ -1272,6 +1272,9 @@ private static Put prepareIndexPutForRebuid(IndexMaintainer indexMaintainer, Imm
             byte[] indexRowKey = indexMaintainer.buildRowKey(mergedRowVG, rowKeyPtr,
                     null, null, HConstants.LATEST_TIMESTAMP);
             indexPut = new Put(indexRowKey);
+        } else {
+            removeEmptyColumn(indexPut, indexMaintainer.getEmptyKeyValueFamily().copyBytesIfNecessary(),
```
gokceni: Can we add a comment here why we remove the empty column?

kadirozde: There is no HBase client API to change the value of a cell. In order to change the value of the empty column cell, we need to remove it from the mutation and add a new cell with the new value (here the verify status). I will add a comment like this.

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/IndexRebuildRegionScanner.java
kadirozde: I have corrected the comment on MUTATION_TS_COMPARATOR.

PR #704, opened 2020-02-04
PHOENIX-5713: Incorrectly handled view and view indexes with/without …

swaroopak:
…namespace in IndexScrutinyMapper#getTtl()

### DIFF
--- a/phoenix-core/src/it/java/org/apache/phoenix/end2end/NonParameterizedIndexScrutinyToolIT.java
+++ b/phoenix-core/src/it/java/org/apache/phoenix/end2end/NonParameterizedIndexScrutinyToolIT.java
@@ -23,6 +23,7 @@ import org.apache.phoenix.mapreduce.index.IndexScrutinyMapperForTest;
 import org.apache.phoenix.mapreduce.index.IndexScrutinyTool;
 import org.apache.phoenix.util.EnvironmentEdgeManager;
 import org.apache.phoenix.util.PropertiesUtil;
+import org.apache.phoenix.util.SchemaUtil;
 import org.junit.Test;
 
 import java.sql.Connection;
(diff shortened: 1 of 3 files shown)


phoenix-core/src/test/java/org/apache/phoenix/index/IndexScrutinyMapperTest.java
```diff
@@ -0,0 +1,117 @@
...
+    String schema, tableName, viewName, viewIndex, indexName;
+    boolean isNamespaceEnabled;
+    PTable inputTable;
+
+    @Before
+    public void setup() {
+        schema = "S_"+generateUniqueName();
+        tableName = "T_"+generateUniqueName();
+        viewName = "V_"+generateUniqueName();
+        viewIndex = "VI_"+generateUniqueName();
```
priyankporwal: viewName and viewIndex don't seem to be used in the tests below - probably testGetSourceTableName_index() and testGetSourceTableName_viewIndex() ought to use these instead of tableName. If not, consider removing them.

phoenix-core/src/it/java/org/apache/phoenix/end2end/NonParameterizedIndexScrutinyToolIT.java
```diff
@@ -156,20 +159,63 @@ public void testScrutinyOnRowsNearExpiry() throws Exception {
...
+        }
+    }
+
+    @Test
+    public void testScrutinyOnRowsNearExpiry_viewIndex() throws Exception {
+        String schemaName = "S"+generateUniqueName();
+        String dataTableName = "T"+generateUniqueName();
+        String dataTableFullName = SchemaUtil.getTableName(schemaName,dataTableName);
+        String viewIndexName = "VI"+generateUniqueName();
+        String viewName = "V"+generateUniqueName();
```
gokceni: Nit: spacing between + and recommend adding it like "V_" +

phoenix-core/src/it/java/org/apache/phoenix/end2end/NonParameterizedIndexScrutinyToolIT.java
```diff
@@ -156,20 +159,63 @@ public void testScrutinyOnRowsNearExpiry() throws Exception {
...
+        String viewFullName = SchemaUtil.getTableName(schemaName,viewName);
+        String dataTableDDL = "CREATE TABLE %s (ID INTEGER NOT NULL PRIMARY KEY, NAME VARCHAR, "
+                + "ZIP INTEGER) TTL="+TEST_TABLE_TTL;
```
gokceni: Is test table ttl around 5 sec is that why initialDelta is 5 below? If yes, you can use this constant

phoenix-core/src/test/java/org/apache/phoenix/index/IndexScrutinyMapperTest.java
```diff
@@ -0,0 +1,117 @@
...
+
+        //test
+        String output = IndexScrutinyMapper.getSourceTableName(inputTable, isNamespaceEnabled);
+        //assert
+        Assert.assertEquals(expectedName, output);
+    }
+
+    @Test
+    public void testGetSourceTableName_viewIndex() {
+        PName physicalTableName = SchemaUtil.getPhysicalHBaseTableName(schema, tableName,
```
gokceni: I would suggest creating actual views, indexes, etc rather than mocking so that we are sure there is no test bug and the name convention is indeed as we expect.

phoenix-core/src/it/java/org/apache/phoenix/end2end/NonParameterizedIndexScrutinyToolIT.java
swaroopak: It was an unnecessary code. Removed it.

phoenix-core/src/test/java/org/apache/phoenix/index/IndexScrutinyMapperTest.java
swaroopak: This is a unit test. Creating a view/table would involve setting up and tearing down mini-cluster which will lead to more time in build phase.

PR #708, opened 2020-02-14
PHOENIX-5698 Phoenix Query with RVC IN list expression generates wron…

yanxinyi:
…g scan with non-pk ordered pks


yanxinyi:
While Daniel Wong and I were debugging and discussing the potential way to fix this RVC problem by query rewrite; however, the current WhereOptimizer generates the wrong scan, see PHOENIX-5740. So the current fix is adding an extra flag to identify whether the ordering of conditions matters or not. 

This PR is only addressing the fixing logic, I will add more tests tomorrow.

phoenix-core/src/it/java/org/apache/phoenix/end2end/InListIT.java
```diff
@@ -537,4 +539,267 @@ public void testInListExpressionWithDesc() throws Exception {
...
 
+    @Test
+    public void testInListExpressionWithDescOrderWithRightQueryPlan() throws Exception {
+        String fullTableName = generateUniqueName();
+        String fullViewName = generateUniqueName();
+        String tenantView = generateUniqueName();
+
+        try (Connection conn = DriverManager.getConnection(getUrl())) {
+            conn.setAutoCommit(true);
+            Statement stmt = conn.createStatement();
```
dbwong: nit: statements should be in try-with-resources.

phoenix-core/src/it/java/org/apache/phoenix/end2end/InListIT.java
```diff
@@ -537,4 +539,267 @@ public void testInListExpressionWithDesc() throws Exception {
...
+                    " CONSTRAINT PK PRIMARY KEY (\n" + " TENANT_ID," + " KEY_PREFIX" + ")) MULTI_TENANT=TRUE");
+            stmt.execute("CREATE VIEW " + fullViewName + "(\n" + " ID1 VARCHAR NOT NULL,\n" + " ID2 VARCHAR NOT NULL,\n" + " EVENT_DATE DATE NOT NULL,\n" +
+                    " CONSTRAINT PKVIEW PRIMARY KEY\n" + " (\n" + " ID1, ID2 DESC, EVENT_DATE DESC\n" + ")) " +
+                    "AS SELECT * FROM " + fullTableName + " WHERE KEY_PREFIX = '0CY'");
+        }
+
+
+        try (Connection viewConn = DriverManager.getConnection(TENANT_SPECIFIC_URL1)) {
+            viewConn.setAutoCommit(true);
+            Statement stmt = viewConn.createStatement();
```
dbwong: nit: statements should be in try-with-resources.

phoenix-core/src/it/java/org/apache/phoenix/end2end/InListIT.java
```diff
@@ -537,4 +539,267 @@ public void testInListExpressionWithDesc() throws Exception {
...
+            rs = stmt.executeQuery("EXPLAIN DELETE FROM " + tenantView + " WHERE (EVENT_DATE, ID2, ID1) IN " +
+                    "((1532458240000, '000000000000400', '005xx000001Sv6o')," +
+                    "(1532458230000, '000000000000300', '005xx000001Sv6o'))");
+            assertTrue(getExplainPlanString(rs).contains("CLIENT PARALLEL 1-WAY RANGE SCAN"));
+
+            stmt.execute("DELETE FROM " + tenantView + " WHERE (EVENT_DATE, ID2, ID1) IN " +
+                    "((1532458240000, '000000000000400', '005xx000001Sv6o')," +
+                    "(1532458230000, '000000000000300', '005xx000001Sv6o'))");
+
+            rs = stmt.executeQuery("SELECT ID2 FROM " + tenantView);
```
dbwong: Do we need this select what is this testing to the this test case?

phoenix-core/src/it/java/org/apache/phoenix/end2end/InListIT.java
```diff
@@ -537,4 +539,267 @@ public void testInListExpressionWithDesc() throws Exception {
...
+                    "(('000000000000400', 1532458240000, '005xx000001Sv6o')," +
+                    "('000000000000300', 1532458230000, '005xx000001Sv6o'))");
+            assertTrue(getExplainPlanString(rs).contains("CLIENT PARALLEL 1-WAY RANGE SCAN"));
+
+            rs = stmt.executeQuery("EXPLAIN DELETE FROM " + tenantView + " WHERE (EVENT_DATE, ID2, ID1) IN " +
+                    "((1532458240000, '000000000000400', '005xx000001Sv6o')," +
+                    "(1532458230000, '000000000000300', '005xx000001Sv6o'))");
+            assertTrue(getExplainPlanString(rs).contains("CLIENT PARALLEL 1-WAY RANGE SCAN"));
+
+            stmt.execute("DELETE FROM " + tenantView + " WHERE (EVENT_DATE, ID2, ID1) IN " +
```
dbwong: Do we need this delete what is this testing to the this test case?

phoenix-core/src/it/java/org/apache/phoenix/end2end/InListIT.java
```diff
@@ -537,4 +539,267 @@ public void testInListExpressionWithDesc() throws Exception {
...
+                    " CONSTRAINT PKVIEW PRIMARY KEY\n" + " (\n" + " ID1, ID2 DESC, EVENT_DATE DESC\n" + ")) " +
+                    "AS SELECT * FROM " + fullTableName + " WHERE KEY_PREFIX = '0CY'");
+        }
+
+
+        try (Connection viewConn = DriverManager.getConnection(TENANT_SPECIFIC_URL1)) {
+            viewConn.setAutoCommit(true);
+            Statement stmt = viewConn.createStatement();
+            stmt.execute("CREATE VIEW IF NOT EXISTS " + tenantView + " AS SELECT * FROM " + fullViewName);
+            viewConn.createStatement().execute("UPSERT INTO " + tenantView + "(ID1, ID2, EVENT_DATE) VALUES " +
```
dbwong: If this is testing plans we likely do not need individual data upserted.

phoenix-core/src/it/java/org/apache/phoenix/end2end/InListIT.java
```diff
@@ -537,4 +539,267 @@ public void testInListExpressionWithDesc() throws Exception {
...
+                    "(1532458230000, '000000000000300', '005xx000001Sv6o'))");
+
+            rs = stmt.executeQuery("SELECT ID2 FROM " + tenantView);
+            assertTrue(rs.next());
+            assertEquals("000000000000500", rs.getString(1));
+        }
+    }
+
+    @Test
+    public void testInListExpressionWithRightQueryPlan() throws Exception {
```
dbwong: Same comments as above, can also make these tests shared methods with a schema attribute to reuse code.

phoenix-core/src/it/java/org/apache/phoenix/end2end/InListIT.java
```diff
@@ -537,4 +539,267 @@ public void testInListExpressionWithDesc() throws Exception {
...
+            stmt.execute("DELETE FROM " + tenantView + " WHERE (ID3, ID2, ID1) IN " +
+                    "((1, '000000000000300', '005xx000001Sv6o'))");
+
+            rs = stmt.executeQuery("SELECT ID2 FROM " + tenantView);
+            assertTrue(rs.next());
+            assertEquals("000000000000400", rs.getString(1));
+        }
+    }
+
+    private String getExplainPlanString(ResultSet rs) throws SQLException {
```
dbwong: Doesn't a similar method already exist in PhoenixRuntime? If not consider adding to TestUtils or another class.

phoenix-core/src/it/java/org/apache/phoenix/end2end/InListIT.java
```diff
@@ -537,4 +539,267 @@ public void testInListExpressionWithDesc() throws Exception {
...
+        }
+    }
+
+    private void testNonPkListPlan(String tenantView) throws Exception {
+        try (Connection viewConn = DriverManager.getConnection(TENANT_SPECIFIC_URL1) ) {
+            PreparedStatement preparedStmt = viewConn.prepareStatement("SELECT * FROM " + tenantView + " WHERE (ID3, ID4) IN " +
+                    "((1, 1)," +
+                    "(2, 2))");
+            QueryPlan queryPlan = PhoenixRuntime.getOptimizedQueryPlan(preparedStmt);
+            assertTrue(queryPlan.getExplainPlan().toString().contains("CLIENT PARALLEL 1-WAY RANGE SCAN OVER"));
```
dbwong: Maybe add a comment mentioning this is only a range scan due to the tenant connection.

phoenix-core/src/main/java/org/apache/phoenix/compile/WhereOptimizer.java
```diff
@@ -546,6 +545,8 @@ public Expression visitLeave(AndExpression node, List<Expression> l) {
      * operators, CASE statements, and string concatenation.
      */
     public static class KeyExpressionVisitor extends StatelessTraverseNoExpressionVisitor<KeyExpressionVisitor.KeySlots> {
+        private boolean orderMatter;
```
dbwong: nit: maybe rename, something like orderDependent

phoenix-core/src/main/java/org/apache/phoenix/compile/WhereOptimizer.java
```diff
@@ -590,11 +591,35 @@ private KeySlots newKeyParts(KeySlot slot, List<Expression> extractNodes, List<K
...
+        public KeySlots newRowValueConstructorKeyParts(RowValueConstructorExpression rvc, List<KeySlots> childSlots) {
             if (childSlots.isEmpty() || rvc.isStateless()) {
                 return null;
             }
-            
+            if (!this.orderMatter) {
+                Collections.sort(childSlots, new Comparator<KeySlots>() {
+                    @Override
+                    public int compare(KeySlots o1, KeySlots o2) {
+                        return o1.getSlots().get(0).getPKPosition() -
```
dbwong: I think get(0) is valid because we don't allow nesting on RVCs which is the only case I'm currently aware of that uses multiple slots.  Is that the case?  And if so we should comment.

phoenix-core/src/main/java/org/apache/phoenix/compile/WhereOptimizer.java
```diff
@@ -661,6 +686,17 @@ private KeySlots newCoerceKeyPart(KeySlot slot, final CoerceExpression node) {
                     childPart, ptr, node, extractNodes), slot.getPKPosition(), slot.getKeyRanges());
         }
 
+        public boolean isOrderMatter() {
+            return this.orderMatter;
+        }
+
+        public void setOrderMatterToTrue() {
```
dbwong: nit: I prefer a generic boolean setter, also rename with above.

phoenix-core/src/main/java/org/apache/phoenix/compile/WhereOptimizer.java
```diff
@@ -1952,7 +1991,7 @@ public PTable getTable() {
             }
         }
 
-        private class RowValueConstructorKeyPart implements KeyPart {
+        public class RowValueConstructorKeyPart implements KeyPart {
```
dbwong: On these and the other public methods consider VisibleForTesting annotation.

phoenix-core/src/main/java/org/apache/phoenix/expression/ComparisonExpression.java
```diff
@@ -367,6 +368,13 @@ public void write(DataOutput output) throws IOException {
 
     @Override
     public final <T> T accept(ExpressionVisitor<T> visitor) {
+//        if (visitor instanceof WhereOptimizer.KeyExpressionVisitor) {
```
dbwong: Remove dead comments.

phoenix-core/src/test/java/org/apache/phoenix/expression/InListExpressionTest.java
```diff
@@ -19,13 +19,19 @@
...
+
 import org.apache.hadoop.hbase.util.Bytes;
+import org.apache.phoenix.compile.WhereOptimizer;
 import org.apache.phoenix.hbase.index.util.ImmutableBytesPtr;
+import org.apache.phoenix.schema.PColumn;
+import org.apache.phoenix.schema.PTable;
 import org.junit.Test;
+import org.mockito.Mockito;
 
 public class InListExpressionTest {
```
dbwong: The tests for this should not live in this class.  This class should be only for InListExpression changes, of which you have none.

phoenix-core/src/test/java/org/apache/phoenix/expression/InListExpressionTest.java
```diff
@@ -62,4 +68,83 @@ public void testHashCode() throws Exception {
                     + hashCode, firstHashCode, hashCode);
         }
     }
+
+    @Test
+    public void testKeyPartsForOrderMatters() {
```
dbwong: This is not a great unit test as its testing multiple things.  Consider reducing the scope and adding additional tests.  Same for below.

phoenix-core/src/it/java/org/apache/phoenix/end2end/InListIT.java
yanxinyi: Yes. I do want to check if my changes still delete the right row.

phoenix-core/src/it/java/org/apache/phoenix/end2end/InListIT.java
yanxinyi: I want to check after deletion behavior since its end to end testing.

phoenix-core/src/it/java/org/apache/phoenix/end2end/InListIT.java
yanxinyi: the previous one tests DESC order by the PK, this one is ASC, which is not using the same schema.

phoenix-core/src/main/java/org/apache/phoenix/compile/WhereOptimizer.java
yanxinyi: comment added

phoenix-core/src/main/java/org/apache/phoenix/expression/ComparisonExpression.java
yanxinyi: Actually, this is not a dead comment. I just found that we need to have this additional check for the RVC comparison

phoenix-core/src/it/java/org/apache/phoenix/end2end/InListIT.java
dbwong: Right but what schema used can be an attribute as i mentioned.

phoenix-core/src/main/java/org/apache/phoenix/expression/ComparisonExpression.java
dbwong: Can we add a UT for this then?

phoenix-core/src/it/java/org/apache/phoenix/end2end/InListIT.java
```diff
@@ -537,4 +539,267 @@ public void testInListExpressionWithDesc() throws Exception {
...
+            stmt.execute("CREATE VIEW IF NOT EXISTS " + tenantView + " AS SELECT * FROM " + fullViewName);
+            viewConn.createStatement().execute("UPSERT INTO " + tenantView + "(ID1, ID2, EVENT_DATE) VALUES " +
+                    "('005xx000001Sv6o', '000000000000300', 1532458230000)");
+            viewConn.createStatement().execute("UPSERT INTO " + tenantView + "(ID1, ID2, EVENT_DATE) VALUES " +
+                    "('005xx000001Sv6o', '000000000000400', 1532458240000)");
+            viewConn.createStatement().execute("UPSERT INTO " + tenantView + "(ID1, ID2, EVENT_DATE) VALUES " +
+                    "('005xx000001Sv6o', '000000000000500', 1532458250000)");
+            viewConn.commit();
+
+            ResultSet rs = stmt.executeQuery("EXPLAIN DELETE FROM " + tenantView + " WHERE (ID1, ID2, EVENT_DATE) " +
```
ChinmaySKulkarni: @yanxinyi So here the where condition is basically on the full row key: <TENANT_ID, KEY_PREFIX, ID1, ID2, EVENT_DATE> such that:

```sql
TENANT_ID=<fixed constant> AND 
KEY_PREFIX= '0CY' AND 
((ID1='005xx000001Sv6o' AND ID2='000000000000400' AND EVENT_DATE= 1532458240000) OR 
(ID1='005xx000001Sv6o' AND ID2='000000000000300' AND EVENT_DATE= 1532458230000))
```
In this case, instead of a range scan, shouldn't the client be able to issue 2 deletes since this is basically 2 point lookups?

phoenix-core/src/it/java/org/apache/phoenix/end2end/InListIT.java
```diff
@@ -537,4 +539,267 @@ public void testInListExpressionWithDesc() throws Exception {
...
+            conn.setAutoCommit(true);
+            Statement stmt = conn.createStatement();
+            stmt.execute("CREATE TABLE " + fullTableName + "(\n" + " TENANT_ID CHAR(15) NOT NULL,\n" + " KEY_PREFIX CHAR(3) NOT NULL,\n" +
+                    " CONSTRAINT PK PRIMARY KEY (\n" + " TENANT_ID," + " KEY_PREFIX" + ")) MULTI_TENANT=TRUE");
+            stmt.execute("CREATE VIEW " + fullViewName + "(\n" + " ID1 VARCHAR NOT NULL,\n" + " ID2 VARCHAR NOT NULL,\n" + " EVENT_DATE DATE, ID3 BIGINT NOT NULL \n" +
+                    " CONSTRAINT PKVIEW PRIMARY KEY\n" + " (ID1, ID2, ID3)) " +
+                    "AS SELECT * FROM " + fullTableName + " WHERE KEY_PREFIX = '0CY'");
+        }
+
+        try (Connection viewConn = DriverManager.getConnection(TENANT_SPECIFIC_URL1)) {
```
ChinmaySKulkarni: I feel like a lot of the lines of code can be shared between various tests with respect table/view creation/upserting data, etc. It will make the tests much more readable as well.

phoenix-core/src/it/java/org/apache/phoenix/end2end/InListIT.java
```diff
@@ -537,4 +539,267 @@ public void testInListExpressionWithDesc() throws Exception {
...
+
+    private void testFullPkPlusNonPkInListPlan(String tenantView) throws Exception {
+        Long numberOfRowsToScan = new Long(2);
+        try (Connection viewConn = DriverManager.getConnection(TENANT_SPECIFIC_URL1) ) {
+            PreparedStatement preparedStmt = viewConn.prepareStatement("SELECT * FROM " + tenantView + " WHERE (ID1, ID2, ID3) IN " +
+                    "(('005xx000001Sv6o', '000000000000500', 1)," +
+                    "('005xx000001Sv6o', '000000000000400', 2))");
+            QueryPlan queryPlan = PhoenixRuntime.getOptimizedQueryPlan(preparedStmt);
+            assertEquals(numberOfRowsToScan, queryPlan.getEstimatedRowsToScan());
+            assertTrue(queryPlan.getExplainPlan().toString().contains(ExplainTable.POINT_LOOKUP_ON_STRING));
```
ChinmaySKulkarni: Any reason why we only extracted `POINT_LOOKUP_ON_STRING` string out in `ExplainTable` and not any of the other ones?

yanxinyi: other strings are not easy to abstract due to the way how to generates the diff query plan

phoenix-core/src/it/java/org/apache/phoenix/end2end/InListIT.java
yanxinyi: removed

phoenix-core/src/test/java/org/apache/phoenix/expression/InListExpressionTest.java
yanxinyi: @dbwong this is only testing one thing here, which is whether we have pk ordering or not.
Due to the way how Phoenix implemented, it's very hard to do the unit tests.

@ChinmaySKulkarni sure will do.

phoenix-core/src/it/java/org/apache/phoenix/end2end/InListIT.java
yanxinyi: created two views at the initial time, so we don't need to recreate all tables, global views again and again, and this will make our IT tests much faster.

phoenix-core/src/it/java/org/apache/phoenix/end2end/InListIT.java
```diff
@@ -539,151 +551,180 @@ public void testInListExpressionWithDesc() throws Exception {
...
+                assertEquals("000000000000500", rs.getString(1));
+            }
+        }
+    }
 
-            rs = stmt.executeQuery("EXPLAIN DELETE FROM " + tenantView + " WHERE (EVENT_DATE, ID2, ID1) IN " +
-                    "((1532458240000, '000000000000400', '005xx000001Sv6o')," +
-                    "(1532458230000, '000000000000300', '005xx000001Sv6o'))");
-            assertTrue(getExplainPlanString(rs).contains("CLIENT PARALLEL 1-WAY RANGE SCAN"));
+    @Ignore
```
dbwong: Please comment why we are ignoring.

phoenix-core/src/it/java/org/apache/phoenix/end2end/InListIT.java
```diff
@@ -539,151 +551,180 @@ public void testInListExpressionWithDesc() throws Exception {
...
+    @Test
+    public void testInListExpressionGeneratesRightScanForAsc() throws Exception {
+        testFullPkListPlan(this.ascViewName);
+        testFullPkPlusNonPkInListPlan(this.ascViewName);
+        testPartialPkListPlan(this.ascViewName);
+        testPartialPkPlusNonPkListPlan(this.ascViewName);
+        testNonPkListPlan(this.ascViewName);
     }
 
+    @Ignore
```
dbwong: Same on why we are ignoring.

yanxinyi:
@dbwong @ChinmaySKulkarni 
The previous PR doesn't have a full set of tests, so I added tests to cover:
1. view on the table
2. view on the global view
3. index on view
4. in list expression with function 

For the implementation, the InListExpression does handle function as well, it might be easier to rewrite visitors, but I'm uncertain of all restrictions. Please provide more insights here, thanks.

PR #711, opened 2020-02-20
 PHOENIX-5743 Concurrent read repairs on the same index row should be…

kadirozde:
… idempotent


phoenix-core/src/main/java/org/apache/phoenix/index/GlobalIndexChecker.java
```diff
@@ -247,18 +247,10 @@ private void deleteRowIfAgedEnough(byte[] indexRowKey, List<Cell> row, long ts,
...
-                    row.clear();
-                    deleteRowScanner.next(row);
-                    deleteRowScanner.close();
-                    // We are deleting a specific version of a row so the flowing loop is for that
-                    for (Cell cell : row) {
-                        del.addColumn(CellUtil.cloneFamily(cell), CellUtil.cloneQualifier(cell), cell.getTimestamp());
-                    }
+                    del.addFamilyVersion(indexMaintainer.getEmptyKeyValueFamily().copyBytesIfNecessary(), ts);
+                }
+                else {
```
gokceni: nit: else should be next to } like
} else {

PR #714, opened 2020-02-26
PHOENIX-5607 Client-server backward compatibility tests

sguggilam:


phoenix-core/src/main/java/org/apache/phoenix/coprocessor/MetaDataProtocol.java
```diff
@@ -145,6 +146,8 @@
     }
     
     public static final String CURRENT_CLIENT_VERSION = PHOENIX_MAJOR_VERSION + "." + PHOENIX_MINOR_VERSION + "." + PHOENIX_PATCH_NUMBER;
+    public static final List<String> COMPATIBLE_CLIENT_VERSIONS =
```
ChinmaySKulkarni: Is there no way to get this info programatically? We will have to keep updating this version with each release right? If not, let's document it in the release webpage.

phoenix-core/src/it/java/org/apache/phoenix/end2end/BackwardCompatibilityIT.java
```diff
@@ -0,0 +1,396 @@
...
+        assertTrue(compareOutput(ADD_DELETE, QUERY_ADD_DELETE));
+    }
+    
+    private void checkForPreConditions() throws Exception {
+        // For the first code cut of any major version, there wouldn't be any backward compatible
+        // clients. Hence the test wouldn't run and just return true when the client  
+        // version to be tested is same as current version
+        assumeFalse(compatibleClientVersion.contains(MetaDataProtocol.CURRENT_CLIENT_VERSION));
+        // Make sure that cluster is clean before test execution with no system tables
+        org.apache.hadoop.hbase.client.Connection conn = ConnectionFactory.createConnection(conf);
```
ChinmaySKulkarni: Put inside try-with-resources

phoenix-core/src/it/java/org/apache/phoenix/end2end/BackwardCompatibilityIT.java
```diff
@@ -0,0 +1,396 @@
...
+        executeQueriesWithCurrentVersion(CREATE_ADD);
+        executeQueryWithClientVersion(compatibleClientVersion, QUERY);
+        assertTrue(compareOutput(CREATE_ADD, QUERY));
+
+        // Deletes with the new client
+        executeQueriesWithCurrentVersion(ADD_DELETE);
+        executeQueriesWithCurrentVersion(QUERY_ADD_DELETE);
+        assertTrue(compareOutput(ADD_DELETE, QUERY_ADD_DELETE));
+    }
+
```
ChinmaySKulkarni: Most backward compatibility issues we found during the 4.15 release were related to old clients unable to perform DDL operations once the cluster was upgraded or old clients unable to query views/indexes/view indexes created by a new clients. Can we add more tests like that?

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/MetaDataProtocol.java
sguggilam: There is no easy way to get this programatically as the number of backward compatible versions can depend on the current versions. Sometimes the first code cut of major version might not have any backward compatible clients as per the doc. I have filed a JIRA to add it to the release web page : https://issues.apache.org/jira/browse/PHOENIX-5746

phoenix-core/src/it/java/org/apache/phoenix/end2end/BackwardCompatibilityIT.java
sguggilam: Done

phoenix-core/src/it/java/org/apache/phoenix/end2end/BackwardCompatibilityIT.java
sguggilam: I think what you meant is creating views etc with a older client once the new client connects to the updated server. Yes, that makes sense and I have added more DDL to the add_data to have all these scenarios covered

stoty:
Already merged.

PR #717, opened 2020-02-27
PHOENIX-5750: Upsert on immutable table fails with AccessDeniedException

swaroopak:


phoenix-core/src/it/java/org/apache/phoenix/end2end/BasePermissionsIT.java
```diff
@@ -863,6 +863,40 @@ private void revokeSystemTableAccess(User superUser, User... users) throws Excep
         }
     }
 
+    AccessTestAction onlyCreateTable(final String tableName) throws SQLException {
```
ChinmaySKulkarni: nit: Rename to `onlyCreateImmutableTable`

PR #719, opened 2020-03-02
PHOENIX-5753

comnetwork:
The mainly modification of this PR are in `WhereOptimizer.pushKeyExpressionsToScan`:

1. for `(pk3,pk4) < (5,7), pk3` is DESC, it is not semantically equals to `pk3 <= 5 and pk4 < 7` , so when we clipped `(pk3,pk4) < (5,7)` to `pk3 <= 5 and pk4<7` , we could only add leading `pk3 <=5` to `cnf`, and `pk4 <7` should not be added in `cnf`, which may lead to `ScanRanges` contains `pk4 < 7` 

2. Whether or not `useSkipScan` should only depend on features of `keySlots`, not depend on `stopExtracting`. Only when the `Scan.startRow` and `Scan.stopRow` or `SkipScanFilter` could not exactly represent the `WhereClause`, we should `stopExtracting`


phoenix-core/src/test/java/org/apache/phoenix/compile/WhereCompilerTest.java
```diff
@@ -1026,61 +1026,4 @@ public void testScanCaching_CustomFetchSizeOnStatement() throws SQLException {
         assertEquals(FETCH_SIZE, pstmt.getFetchSize());
         assertEquals(FETCH_SIZE, scan.getCaching());
     }
-
```
comnetwork: This test is moved to WhereOptimizerTest.testLastPkColumnIsVariableLengthAndDescBug5307

PR #721, opened 2020-03-04
PHOENIX-5755 Ensure that Hadoop and HBase dependencies are not overri…

stoty:
…dden with older versions

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -89,6 +89,7 @@
     <commons-cli.version>1.4</commons-cli.version>
     <jackson.version>1.9.2</jackson.version>
     <antlr.version>3.5.2</antlr.version>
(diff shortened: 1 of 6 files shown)


pom.xml
```diff
@@ -153,6 +154,16 @@
             <target>1.8</target>
           </configuration>
         </plugin>
+        <plugin>
+          <groupId>org.antlr</groupId>
+          <artifactId>antlr3-maven-plugin</artifactId>
+          <version>3.5.2</version>
```
mihir6692: nit: can we use version as antlr.version property? It's already defined in this pom.xml

mihir6692:
other than one nit suggestions, LGTM. +1

PR #723, opened 2020-03-05
PHOENIX-5757 IndexUpgrade tool set Phoenix query and Hbase rpc timeouts

tkhurana:


tkhurana:
@swaroopak @priyankporwal @gokceni

priyankporwal:
+1. thanks @tkhurana!

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/index/IndexUpgradeTool.java
```diff
@@ -314,12 +316,39 @@ public void prepareToolSetup() {
         }
     }
 
+    private static void setRpcRetriesAndTimeouts(Configuration conf) {
+        long indexRebuildQueryTimeoutMs =
+                conf.getLong(QueryServices.INDEX_REBUILD_QUERY_TIMEOUT_ATTRIB,
```
swaroopak: This is not using the timeout explicitly, it will end up using default only if not specified in config which is not the case in our environment

We want to specify an explicit timeout like 10 minutes whether or not it is set. This will apply to all timeout changes.

tkhurana: INDEX_REBUILD_* config values are phoenix configuration values which are different from the corresponding hbase corresponding values. For example there is a `hbase.client.scanner.timeout.period` defined as 1 minute in our environment but this change is using the value of `phoenix.index.rebuild.client.scanner.timeout` to set the corresponding hbase scan timeout. I don't see these values being defined so they will fallback to their defaults. At the same time it gives us flexibility to tune these values via config overrides.

phoenix-core/src/it/java/org/apache/phoenix/end2end/IndexUpgradeToolIT.java
```diff
@@ -137,4 +141,33 @@ private void prepareForTest(Connection conn, String schemaName, String tableName
             }
         }
     }
+
+    @Test
+    public void testToolConnectionProperties() throws Exception {
```
swaroopak: should this be in a Unit test? it does not need a cluster.

phoenix-core/src/it/java/org/apache/phoenix/end2end/IndexUpgradeToolIT.java
```diff
@@ -137,4 +141,33 @@ private void prepareForTest(Connection conn, String schemaName, String tableName
             }
         }
     }
+
+    @Test
```
swaroopak: another test to see if defaults are getting passed when conf does not set anything.

swaroopak:
+1, Please upload the patch to Jira so that hadoop QA runs.

PR #724, opened 2020-03-05
PHOENIX-5629 Phoenix Function to Return HBase row timestamp

jpisaac:
Implements a Function **PHOENIX_ROW_TIMESTAMP** that returns the phoenix_row_timestamp - which as defined as the row_timestamp of the empty column.

- Clients can then issue queries using the newly minted function **PHOENIX_ROW_TIMESTAMP**.
for e.g To get the phoenix_row_timestamp on the client side, one would include it in the SELECT clause as such, **SELECT PHOENIX_ROW_TIMESTAMP(), ...other columns FROM table WHERE where-clause**

- Can also be used as a predicate to filter the rows that need to be fetched from the table. for e.g **SELECT columns ... FROM table where PHOENIX_ROW_TIMESTAMP() < time**.


jpisaac:
@ChinmaySKulkarni  @abhishek-chouhan  @gjacoby126  @tkhurana  Please review.

phoenix-core/src/it/java/org/apache/phoenix/end2end/PhoenixRowTimestampFunctionIT.java
```diff
@@ -191,4 +192,158 @@ public void testRowTimestampColumn() throws Exception {
...
+        try (Connection conn = DriverManager.getConnection(getUrl())) {
+            String ddl = "CREATE TABLE IF NOT EXISTS " + tableName
+                    + " (PK1 INTEGER NOT NULL, PK2 DATE NOT NULL, KV1 VARCHAR, KV2 VARCHAR"
+                    + " CONSTRAINT PK PRIMARY KEY(PK1, PK2))" + this.tableDDLOptions;
+            conn.createStatement().execute(ddl);
+
+            String dml = "UPSERT INTO " + tableName + " (PK1, PK2, KV1, KV2) VALUES (?, ?, ?, ?)";
+
+            long rowTimestamp = EnvironmentEdgeManager.currentTimeMillis();
+            Date rowTimestampDate1 = new Date(rowTimestamp - 120000);
```
yanxinyi: why 120000? some random number? can you make it as constant?

phoenix-core/src/it/java/org/apache/phoenix/end2end/PhoenixRowTimestampFunctionIT.java
```diff
@@ -191,4 +192,158 @@ public void testRowTimestampColumn() throws Exception {
...
+                assertTrue(rs3.getDate(1).before(rowTimestampDate2));
+                actualCount3++;
+            }
+            assertEquals(count, actualCount3);
+            rs3.close();
+        }
+    }
+
+    @Ignore("Fails when encoding=NON_ENCODED_QUALIFIERS and PHOENIX_ROW_TIMESTAMP "
+            + "is the only select column"
```
yanxinyi: why we can't get the right behavior of `SELECT PHOENIX_ROW_TIMESTAMP()` only here?

phoenix-core/src/main/java/org/apache/phoenix/parse/PhoenixRowTimestampParseNode.java
```diff
@@ -0,0 +1,89 @@
...
+            }
+            @Override
+            public SortOrder getSortOrder() {
+                return SortOrder.getDefault();
+            }
+        }, emptyColumnFamilyName, emptyColumnName)});
+
+        // Add the empty column to the projection list.
+        // According to PHOENIX-4179 this will then return the timestamp of the empty column cell.
+        IndexUtil.addEmptyColumnToScan(context.getScan(), emptyColumnFamilyName, emptyColumnName);
```
yanxinyi: does IndexUtil handle the non-index table here?

phoenix-core/src/it/java/org/apache/phoenix/end2end/PhoenixRowTimestampFunctionIT.java
```diff
@@ -191,4 +192,158 @@ public void testRowTimestampColumn() throws Exception {
...
     }
 
+    @Test
+    public void testRowTimestampColumnAndPredicates() throws Exception {
+        String tableName =  generateUniqueName();
+        try (Connection conn = DriverManager.getConnection(getUrl())) {
+            String ddl = "CREATE TABLE IF NOT EXISTS " + tableName
+                    + " (PK1 INTEGER NOT NULL, PK2 DATE NOT NULL, KV1 VARCHAR, KV2 VARCHAR"
+                    + " CONSTRAINT PK PRIMARY KEY(PK1, PK2))" + this.tableDDLOptions;
+            conn.createStatement().execute(ddl);
```
ChinmaySKulkarni: nit: Add statement creation to try-with-resources as well.

phoenix-core/src/it/java/org/apache/phoenix/end2end/PhoenixRowTimestampFunctionIT.java
```diff
@@ -191,4 +192,158 @@ public void testRowTimestampColumn() throws Exception {
...
+            }
+            conn.commit();
+
+            // case: No rows should be selected
+            // Since we used a past date for column PK2
+            String dql1 = "SELECT PHOENIX_ROW_TIMESTAMP(), KV1 FROM " + tableName +
+                    " WHERE PHOENIX_ROW_TIMESTAMP() = PK2";
+
+            ResultSet rs1 = conn.createStatement().executeQuery(dql1);
+            assertTrue(!rs1.next());
```
ChinmaySKulkarni: Use assertFalse instead

phoenix-core/src/it/java/org/apache/phoenix/end2end/PhoenixRowTimestampFunctionIT.java
```diff
@@ -191,4 +192,158 @@ public void testRowTimestampColumn() throws Exception {
...
+    @Ignore("Fails when encoding=NON_ENCODED_QUALIFIERS and PHOENIX_ROW_TIMESTAMP "
+            + "is the only select column"
+            + "Expected rows do match expected:<5> but was:<0>")
+    public void testRowTimestampColumnOnlyAndPredicates() throws Exception {
+        String tableName =  generateUniqueName();
+        try (Connection conn = DriverManager.getConnection(getUrl())) {
+            String ddl = "CREATE TABLE IF NOT EXISTS " + tableName
+                    + " (PK1 INTEGER NOT NULL, PK2 DATE NOT NULL, KV1 VARCHAR, KV2 VARCHAR"
+                    + " CONSTRAINT PK PRIMARY KEY(PK1, PK2))" + this.tableDDLOptions;
+            conn.createStatement().execute(ddl);
```
ChinmaySKulkarni: nit: Add statement creation to try-with-resources

phoenix-core/src/it/java/org/apache/phoenix/end2end/PhoenixRowTimestampFunctionIT.java
```diff
@@ -191,4 +192,158 @@ public void testRowTimestampColumn() throws Exception {
...
+            }
+            conn.commit();
+
+            // case: No rows should be selected
+            // Since we used a past date for column PK2
+            String dql1 = "SELECT PHOENIX_ROW_TIMESTAMP() FROM " + tableName +
+                    " WHERE PHOENIX_ROW_TIMESTAMP() = PK2";
+
+            ResultSet rs1 = conn.createStatement().executeQuery(dql1);
+            assertTrue(!rs1.next());
```
ChinmaySKulkarni: Use assertFalse instead

phoenix-core/src/it/java/org/apache/phoenix/end2end/PhoenixRowTimestampFunctionIT.java
```diff
@@ -191,4 +192,158 @@ public void testRowTimestampColumn() throws Exception {
...
+
+            ResultSet rs3 = conn.createStatement().executeQuery(dql3);
+            int actualCount3 = 0;
+            while(rs3.next()) {
+                assertTrue(rs3.getDate(1).before(rowTimestampDate2));
+                actualCount3++;
+            }
+            assertEquals(count, actualCount3);
+            rs3.close();
+        }
```
ChinmaySKulkarni: This test case seems to have 90% of the code common as the first one. Can we refactor to use a common helper method for both to reduce repetition?

phoenix-core/src/it/java/org/apache/phoenix/end2end/PhoenixRowTimestampFunctionIT.java
```diff
@@ -191,4 +192,158 @@ public void testRowTimestampColumn() throws Exception {
...
+    }
+
 }
```
ChinmaySKulkarni: @jpisaac  Since we rely on the empty column timestamp to be reflected as the `PHOENIX_ROW_TIMESTAMP()`, should we add tests to confirm this? Also perhaps add tests that confirm that this is modified whenever a row is updated

phoenix-core/src/main/java/org/apache/phoenix/compile/ProjectionCompiler.java
```diff
@@ -507,6 +507,9 @@ public static RowProjector compile(StatementContext context, SelectStatement sta
                 } else {
                     for (byte[] cq : entry.getValue()) {
                             PColumn column = family.getPColumnForColumnQualifier(cq);
+                            // Continue: If an EMPTY_COLUMN is in the projection list, since
+                            // no new KeyValue is returned.
+                            if (column == null) continue;
```
ChinmaySKulkarni: Don't we omit the empty column when projecting? Under which circumstances will this condition be true?

phoenix-core/src/main/java/org/apache/phoenix/expression/function/PhoenixRowTimestampFunction.java
```diff
@@ -0,0 +1,136 @@
...
+/**
+ * Function to return the timestamp of the empty column which functions as the row timestamp. The
+ * result returned can be used for debugging(eg. using HBase shell), logging etc.
+ * Can also be used in sql predicates.
+ */
+@BuiltInFunction(name = PhoenixRowTimestampFunction.NAME,
+        nodeClass= PhoenixRowTimestampParseNode.class,
+        args = {})
+public class PhoenixRowTimestampFunction extends ScalarFunction {
+    private static final Logger LOGGER = LoggerFactory.getLogger(PhoenixRowTimestampFunction.class);
```
ChinmaySKulkarni: nit: Remove this logger since it's not used. Also remove corresponding imports.

phoenix-core/src/main/java/org/apache/phoenix/parse/PhoenixRowTimestampParseNode.java
```diff
@@ -0,0 +1,89 @@
...
+    PhoenixRowTimestampParseNode(String name, List<ParseNode> children,
+                                 BuiltInFunctionInfo info) {
+        super(name, children, info);
+    }
+
+    @Override
+    public FunctionExpression create(List<Expression> children, StatementContext context) throws SQLException {
+
+        // PhoenixRowTimestampFunction does not take any parameters.
+        assert children.size() == 0;
```
ChinmaySKulkarni: This is a user error, so it will be better to throw an exception so the user understands the correct usage of PHOENIX_ROW_TIMESTAMP. Can we do that instead?

phoenix-core/src/main/java/org/apache/phoenix/parse/PhoenixRowTimestampParseNode.java
```diff
@@ -0,0 +1,89 @@
...
+                return SortOrder.getDefault();
+            }
+        }, emptyColumnFamilyName, emptyColumnName)});
+
+        // Add the empty column to the projection list.
+        // According to PHOENIX-4179 this will then return the timestamp of the empty column cell.
+        IndexUtil.addEmptyColumnToScan(context.getScan(), emptyColumnFamilyName, emptyColumnName);
+        return new PhoenixRowTimestampFunction(emptyColumnExpression, emptyColumnFamilyName, emptyColumnName);
+    }
+}
```
ChinmaySKulkarni: @jpisaac let's add unit tests for the newly added classes, however small the test classes may be. Good for code coverage.

phoenix-core/src/it/java/org/apache/phoenix/end2end/PhoenixRowTimestampFunctionIT.java
```diff
@@ -191,4 +192,158 @@ public void testRowTimestampColumn() throws Exception {
         }
     }
 
+    @Test
+    public void testRowTimestampColumnAndPredicates() throws Exception {
```
swaroopak: all these cases deserve different test functions, please split them. extracting common code will make the test suite more maintainable.

phoenix-core/src/main/java/org/apache/phoenix/parse/PhoenixRowTimestampParseNode.java
jpisaac: addEmptyColumnToScan should handle all cases, But this will probably need refactoring to fix the test failure that we are seeing in the IT tests.

phoenix-core/src/it/java/org/apache/phoenix/end2end/PhoenixRowTimestampFunctionIT.java
jpisaac: Had kept the test just as a reminder for me - that I need to fix this case. Will update the IT with the proper tests.

phoenix-core/src/it/java/org/apache/phoenix/end2end/PhoenixRowTimestampFunctionIT.java
jpisaac: The EmptyColumnIT tests do an exhaustive verification around that condition. Let me know if there is anything missing?

phoenix-core/src/main/java/org/apache/phoenix/compile/ProjectionCompiler.java
jpisaac: This check is for the case when we add empty column in the projection list and since empty column is not part of the table column list, we need to bypass further processing.

phoenix-core/src/it/java/org/apache/phoenix/end2end/PhoenixRowTimestampFunctionIT.java
jpisaac: Fixed it in the new commit. A FirstKeyOnlyFilter filter was getting added which was causing the rows to get filtered out.

phoenix-core/src/main/java/org/apache/phoenix/parse/PhoenixRowTimestampParseNode.java
jpisaac: Found that when a param is passed into PHOENIX_ROW_TIMESTAMP it gets parsed as UDFParseNode. But added the check and exception anyways.

phoenix-core/src/main/java/org/apache/phoenix/parse/PhoenixRowTimestampParseNode.java
jpisaac: Added it the unit tests

phoenix-core/src/it/java/org/apache/phoenix/end2end/PhoenixRowTimestampFunctionIT.java
```diff
@@ -31,6 +31,7 @@
 import org.apache.phoenix.schema.PTable.QualifierEncodingScheme;
 import org.apache.phoenix.util.EncodedColumnsUtil;
 import org.apache.phoenix.util.EnvironmentEdgeManager;
+import org.junit.Ignore;
```
yanxinyi, line 34: ignore import is not needed

phoenix-core/src/it/java/org/apache/phoenix/end2end/PhoenixRowTimestampFunctionIT.java
```diff
@@ -191,4 +225,130 @@ public void testRowTimestampColumn() throws Exception {
...
+                    rs.getString(2);
+                    assertFalse(rs.wasNull());
+                    actualCount++;
+                }
+                assertEquals(NUM_ROWS, actualCount);
+                rs.close();
+            }
+        }
+    }
+
```
yanxinyi, line 353: better to support test for `SELECT COLUMN_ONLY WHERE PHOENIX_ROW_TIMESTAMP()` predicate.

jpisaac:
@swaroopak @ChinmaySKulkarni @yanxinyi Please review the updates.

phoenix-core/src/it/java/org/apache/phoenix/end2end/PhoenixRowTimestampFunctionIT.java
```diff
@@ -0,0 +1,398 @@
...
+        String tableName =  generateUniqueName();
+        try (Connection conn = DriverManager.getConnection(getUrl())) {
+            String ddl = "CREATE TABLE IF NOT EXISTS " + tableName
+                    + " (PK INTEGER NOT NULL PRIMARY KEY, KV1 VARCHAR, KV2 VARCHAR)"
+                    + this.tableDDLOptions;
+            conn.createStatement().execute(ddl);
+
+            String dml = "UPSERT INTO " + tableName + " (PK, KV1, KV2) VALUES (?, ?, ?)";
+            PreparedStatement stmt = conn.prepareStatement(dml);
```
ChinmaySKulkarni, line 157: nit: Use try-with-resources to close this statement?

phoenix-core/src/it/java/org/apache/phoenix/end2end/PhoenixRowTimestampFunctionIT.java
```diff
@@ -0,0 +1,398 @@
...
+            conn.commit();
+
+            String dql = "SELECT PHOENIX_ROW_TIMESTAMP() FROM " + tableName;
+
+            ResultSet rs = conn.createStatement().executeQuery(dql);
+            // verify row timestamp returned by the query matches the empty column cell timestamp
+            verifyHbaseAllRowsTimestamp(tableName, rs, count);
+
+            // update one row
+            conn.createStatement().execute("UPSERT INTO " + tableName
```
ChinmaySKulkarni, line 175: Same with this statement

phoenix-core/src/it/java/org/apache/phoenix/end2end/PhoenixRowTimestampFunctionIT.java
```diff
@@ -0,0 +1,398 @@
...
+                    + " (PK, KV1) VALUES (2, 'KV1_foo')");
+            conn.commit();
+
+            rs = conn.createStatement().executeQuery(dql);
```
ChinmaySKulkarni, line 179: ditto

phoenix-core/src/it/java/org/apache/phoenix/end2end/PhoenixRowTimestampFunctionIT.java
```diff
@@ -0,0 +1,398 @@
...
+            // verify again after update
+            verifyHbaseAllRowsTimestamp(tableName, rs, count);
+
+            dql = "SELECT ROWKEY_BYTES_STRING(), PHOENIX_ROW_TIMESTAMP() FROM " + tableName
+                    + " WHERE PK >= 1 AND PK <=3 ";
+            rs = conn.createStatement().executeQuery(dql);
```
ChinmaySKulkarni, line 185: ditto

phoenix-core/src/it/java/org/apache/phoenix/end2end/PhoenixRowTimestampFunctionIT.java
```diff
@@ -0,0 +1,398 @@
...
+    }
+
+    @Test
+    public void testRowTimestampColumn() throws Exception {
+        String tableName =  generateUniqueName();
+        try (Connection conn = DriverManager.getConnection(getUrl())) {
+            String ddl = "CREATE TABLE IF NOT EXISTS " + tableName
+                    + " (PK1 INTEGER NOT NULL, PK2 DATE NOT NULL, KV1 VARCHAR, KV2 VARCHAR"
+                    + " CONSTRAINT PK PRIMARY KEY(PK1, PK2 ROW_TIMESTAMP))" + this.tableDDLOptions;
+            conn.createStatement().execute(ddl);
```
ChinmaySKulkarni, line 200: Same here. Same comment throughout for all other cases where we don't close the opened statement

phoenix-core/src/main/java/org/apache/phoenix/compile/ProjectionCompiler.java
```diff
@@ -507,6 +507,10 @@ public static RowProjector compile(StatementContext context, SelectStatement sta
                 } else {
                     for (byte[] cq : entry.getValue()) {
                             PColumn column = family.getPColumnForColumnQualifier(cq);
+                            // Continue: If an EMPTY_COLUMN is in the projection list,
+                            // since the table column list does not contain the EMPTY_COLUMN
+                            // no value is returned.
+                            if (column == null) continue;
```
ChinmaySKulkarni, line 513: nit: Can we use braces and put the `continue` on a separate line to be consistent with formatting?

phoenix-core/src/main/java/org/apache/phoenix/expression/function/PhoenixRowTimestampFunction.java
```diff
@@ -0,0 +1,125 @@
...
+
+    @Override
+    public String getName() {
+        return NAME;
+    }
+
+    @Override
+    public boolean evaluate(Tuple tuple, ImmutableBytesWritable ptr) {
+
+        if (tuple == null || tuple.size() == 0) return false;
```
ChinmaySKulkarni, line 68: nit: Use braces and move return to next line

phoenix-core/src/it/java/org/apache/phoenix/end2end/PhoenixRowTimestampFunctionIT.java
```diff
@@ -0,0 +1,398 @@
...
+                while(rs.next()) {
+                    int rowCount = rs.getInt(1);
+                    assertFalse(rs.wasNull());
+                    assertTrue(rowCount == NUM_ROWS);
+                }
+                rs.close();
+            }
+        }
+    }
+
```
ChinmaySKulkarni, line 397: Can we add a test with a `WHERE PHOENIX_ROW_TIMESTAMP() < TO_DATE("date string')` or comparison with `TO_TIMESTAMP('timestamp string')`?

phoenix-core/src/main/java/org/apache/phoenix/expression/function/PhoenixRowTimestampFunction.java
```diff
@@ -0,0 +1,125 @@
...
+import java.util.List;
+
+/**
+ * Function to return the timestamp of the empty column which functions as the row timestamp. The
+ * result returned can be used for debugging(eg. using HBase shell), logging etc.
+ * Can also be used in sql predicates.
+ */
+@BuiltInFunction(name = PhoenixRowTimestampFunction.NAME,
+        nodeClass= PhoenixRowTimestampParseNode.class,
+        args = {})
```
ChinmaySKulkarni, line 45: What is the behavior if we were to call `PHOENIX_ROW_TIMESTAMP(<arg(s)>)`? Does specifying `args = {}` cut it, or do we need to throw a specific exception to indicate improper usage?

phoenix-core/src/main/java/org/apache/phoenix/expression/function/PhoenixRowTimestampFunction.java
```diff
@@ -0,0 +1,125 @@
...
+    public String getName() {
+        return NAME;
+    }
+
+    @Override
+    public boolean evaluate(Tuple tuple, ImmutableBytesWritable ptr) {
+
+        if (tuple == null || tuple.size() == 0) return false;
+
+        long ts = tuple.getValue(0).getTimestamp();
```
ChinmaySKulkarni, line 70: Why do we do this and also later resolve the timestamp from `emptyColumnKV`? Shouldn't the below `if` condition always succeed and `ts` be evaluated that way?

phoenix-core/src/test/java/org/apache/phoenix/parse/PhoenixRowTimestampFunctionTest.java
```diff
@@ -0,0 +1,45 @@
...
+
+public class PhoenixRowTimestampFunctionTest {
+
+    @Test
+    public void testExpressionWithPhoenixRowTimestamp() throws Exception {
+        ParseNode parseNode = SQLParser.parseCondition("(PHOENIX_ROW_TIMESTAMP() = PK2)");
+        boolean hasPhoenixRowTimestampParseNode = false;
+        for (ParseNode childNode : parseNode.getChildren()) {
+            if (childNode.getClass().isAssignableFrom(PhoenixRowTimestampParseNode.class)) {
+                assertEquals("", childNode.getChildren().size(), 0);
```
ChinmaySKulkarni, line 19: nit: Remove empty string print in assertion and also fix if condition formatting

phoenix-core/src/main/java/org/apache/phoenix/parse/PhoenixRowTimestampParseNode.java
```diff
@@ -0,0 +1,92 @@
...
+        // Get the empty column family and qualifier for the context.
+        PTable table = context.getCurrentTable().getTable();
+        byte[] emptyColumnFamilyName = SchemaUtil.getEmptyColumnFamily(table);
+        byte[] emptyColumnName = table.getEncodingScheme() == PTable.QualifierEncodingScheme.NON_ENCODED_QUALIFIERS ?
+                QueryConstants.EMPTY_COLUMN_BYTES :
+                table.getEncodingScheme().encode(QueryConstants.ENCODED_EMPTY_COLUMN_NAME);
+
+        // Create an empty column key value expression.
+        // This will cause the empty column key value to evaluated during scan filter processing.
+        List<Expression> emptyColumnExpression = Arrays.asList(new Expression[] {new KeyValueColumnExpression(new PDatum() {
```
ChinmaySKulkarni, line 64: Why do we need this `emptyColumnExpression`? This gets passed as a `List<Expression> children` and passed to the parent class (`ScalarFunction`) constructor, but is never used since we technically don't have any arguments for this function. Is it possible to pass in null instead of creating this expression? Please let me know if I'm missing something.

phoenix-core/src/main/java/org/apache/phoenix/expression/function/PhoenixRowTimestampFunction.java
```diff
@@ -0,0 +1,125 @@
...
+public class PhoenixRowTimestampFunction extends ScalarFunction {
+    public static final String NAME = "PHOENIX_ROW_TIMESTAMP";
+    private byte[] emptyCF;
+    private byte[] emptyCQ;
+
+    public PhoenixRowTimestampFunction() {
+    }
+
+    public PhoenixRowTimestampFunction(List<Expression> children, byte[] emptyCF, byte[] emptyCQ) {
+        super(children);
```
ChinmaySKulkarni, line 55: Where are we using `children`? We don't expect any args for this UDF and so this wouldn't really be used anywhere right?

phoenix-core/src/main/java/org/apache/phoenix/expression/function/PhoenixRowTimestampFunction.java
ChinmaySKulkarni: If so, this initialization of `ts` would always be overwritten right? Or is there some case wherein the below `if` condition is false and so we use this first initialization?

ChinmaySKulkarni: Discussed offline with @jpisaac about the reasoning behind this. Let's add a comment here to explain these nuances.

phoenix-core/src/main/java/org/apache/phoenix/parse/PhoenixRowTimestampParseNode.java
jpisaac: As discussed offline, adding more comments.

phoenix-core/src/main/java/org/apache/phoenix/expression/function/PhoenixRowTimestampFunction.java
jpisaac: Will add more comments as discussed offline.

jpisaac:
Thanks @ChinmaySKulkarni @yanxinyi @swaroopak @tkhurana for the reviews. Closing it for now.

PR #725, opened 2020-03-06
PHOENIX-5749: Add unit tests for verifySingleIndexRow() of IndexRebui…

swaroopak:
…ldRegionScanner

### DIFF
--- a/phoenix-core/src/main/java/org/apache/phoenix/coprocessor/IndexRebuildRegionScanner.java
+++ b/phoenix-core/src/main/java/org/apache/phoenix/coprocessor/IndexRebuildRegionScanner.java
@@ -48,6 +48,7 @@ import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
 import java.util.NavigableSet;
+import java.util.TreeMap;
 import java.util.concurrent.ExecutionException;
 
 import com.google.common.annotations.VisibleForTesting;
@@ -130,16 +131,14 @@ public class IndexRebuildRegionScanner extends BaseRegionScanner {
     private Scan scan;
     private RegionScanner innerScanner;
     private Region region;
-    @VisibleForTesting
-    public IndexMaintainer indexMaintainer;
+    private IndexMaintainer indexMaintainer;
     private byte[] indexRowKey = null;
     private Table indexHTable = null;
     private Table outputHTable = null;
     private Table resultHTable = null;
     private IndexTool.IndexVerifyType verifyType = IndexTool.IndexVerifyType.NONE;
     private boolean verify = false;
-    @VisibleForTesting
-    public Map<byte[], List<Mutation>> indexKeyToMutationMap;
+    private  Map<byte[], List<Mutation>> indexKeyToMutationMap;
     private Map<byte[], Pair<Put, Delete>> dataKeyToMutationMap;
     private TaskRunner pool;
     private TaskBatch<Boolean> tasks;
@@ -147,8 +146,7 @@ public class IndexRebuildRegionScanner extends BaseRegionScanner {
     private UngroupedAggregateRegionObserver ungroupedAggregateRegionObserver;
     private RegionCoprocessorEnvironment env;
     private HTableFactory hTableFactory;
-    @VisibleForTesting
-    public int indexTableTTL = 0;
+    private int indexTableTTL = 0;
     private IndexToolVerificationResult verificationResult;
     private boolean isBeforeRebuilt = true;
     private boolean partialRebuild = false;
@@ -332,6 +330,24 @@ public class IndexRebuildRegionScanner extends BaseRegionScanner {
             mutationList.clear();
         }
         return uuidValue;
+    }
+
+    @VisibleForTesting
+    public int setIndexTableTTL(int forever) {
+        indexTableTTL = forever;
+        return 0;
+    }
+
+    @VisibleForTesting
+    public int setIndexMaintainer(IndexMaintainer indexMaintainer) {
+        this.indexMaintainer = indexMaintainer;
+        return 0;
+    }
+
+    @VisibleForTesting
+    public int setIndexKeyToMutationMap(Map<byte[], List<Mutation>> newTreeMap) {
+        this.indexKeyToMutationMap = newTreeMap;
+        return 0;
     }
 
     public static class SimpleValueGetter implements ValueGetter {
--- a/phoenix-core/src/main/java/org/apache/phoenix/coprocessor/IndexRebuildRegionScanner.java
+++ b/phoenix-core/src/main/java/org/apache/phoenix/coprocessor/IndexRebuildRegionScanner.java
@@ -42,7 +42,6 @@ import static org.apache.phoenix.query.QueryServices.MUTATE_BATCH_SIZE_BYTES_ATT
 
 import java.io.IOException;
 import java.util.ArrayList;
-import java.util.Arrays;
 import java.util.Collections;
 import java.util.Comparator;
 import java.util.Iterator;
@@ -115,6 +114,9 @@ public class IndexRebuildRegionScanner extends BaseRegionScanner {
     private static final int DEFAULT_CONCURRENT_INDEX_VERIFY_THREADS = 17;
     public static final String INDEX_VERIFY_ROW_COUNTS_PER_TASK_CONF_KEY = "index.verify.threads.max";
     private static final int DEFAULT_INDEX_VERIFY_ROW_COUNTS_PER_TASK = 2048;
+    public static final String NO_EXPECTED_MUTATION = "No expected mutation";
+    public static final String
+            ACTUAL_MUTATION_IS_NULL_OR_EMPTY = "actualMutationList is null or empty";
     private long pageSizeInRows = Long.MAX_VALUE;
     private int rowCountPerTask;
     private boolean hasMore;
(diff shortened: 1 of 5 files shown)


phoenix-core/src/main/java/org/apache/phoenix/coprocessor/IndexRebuildRegionScanner.java
```diff
@@ -109,231 +110,6 @@
 
 public class IndexRebuildRegionScanner extends BaseRegionScanner {
 
-    public static class VerificationResult {
```
kadirozde: Why do you need to move this code out from this file? It is a static helper class and only meaningful in the context of index verification and not by itself.

phoenix-core/src/test/java/org/apache/phoenix/index/IndexToolSingleRowVerifyTest.java
```diff
@@ -0,0 +1,461 @@
...
+        familyCellMap.put(Bytes.toBytes(0), cellList);
+        m.setFamilyCellMap(familyCellMap);
+        actualMutations.removeAll(actualMutations);
+        actualMutations.add(m);
+        return actualMutations;
+    }
+
+    private void assertVerificationPhaseResult(IndexToolVerificationResult.PhaseResult actualPR,
+            IndexToolVerificationResult.PhaseResult expectedPR) {
+        Assert.assertTrue(actualPR.toString().equalsIgnoreCase(expectedPR.toString()));
```
priyankporwal: Comparing one PhaseResult to another would be more intuitive and best implemented by overriding equals() instead of comparing output of toString() functions. You wouldn't even need assertVerificationPhaseResult() function then and Assert.Equals() would suffice.

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/IndexRebuildRegionScanner.java
```diff
@@ -594,14 +376,16 @@ private boolean checkIndexRow(final byte[] indexRowKey, final Put put) throws IO
         return true;
     }
 
-    private void logToIndexToolOutputTable(byte[] dataRowKey, byte[] indexRowKey, long dataRowTs, long indexRowTs,
+    @VisibleForTesting
```
wangweiming800: Do we still need this annotation if it is a public function?

swaroopak: Nope!

phoenix-core/src/test/java/org/apache/phoenix/index/IndexToolSingleRowVerifyTest.java
```diff
@@ -0,0 +1,634 @@
...
+                if (delete == null) {
+                    delete = new Delete(key);
+                }
+                delete.addDeleteMarker(cell);
+            }
+        }
+    }
+
+    private void initializeRebuildScannerAttributes() {
+        when(rebuildScanner.setIndexTableTTL(Matchers.anyInt())).thenCallRealMethod();
```
gjacoby126: All of this partial mocking in initializeRebuildScannerAttributes and initalizeGlobalMockitoSetup should be taken as a "code smell" that the rebuild scanner needs to be broken apart into multiple classes. Generally when mocking you have one class under test which isn't mocked at all, and other dependent classes that aren't being tested that are completely mocked. 

Again, don't have to fix this in _this_ PR, because without the tests first it's hard to refactor this safely. But the refactoring should happen.

phoenix-core/src/test/java/org/apache/phoenix/index/IndexToolSingleRowVerifyTest.java
```diff
@@ -0,0 +1,634 @@
...
+    @Test
+    public void testVerifySingleIndexRow_validIndexRowCount_nonZero() throws IOException {
+        IndexToolVerificationResult.PhaseResult expectedPR = getValidPhaseResult();
+        for (Map.Entry<byte[], List<Mutation>>
+                entry : indexKeyToMutationMapLocal.entrySet()) {
+            initializeLocalMockitoSetup(entry, TestType.VALID);
+            //test code
+            rebuildScanner.verifySingleIndexRow(indexRow, actualPR);
+            //assert
+            actualPR.equals(expectedPR);
```
gjacoby126: forgot to assert?

phoenix-core/src/test/java/org/apache/phoenix/index/IndexToolSingleRowVerifyTest.java
```diff
@@ -0,0 +1,634 @@
...
+
+    @Test
+    public void testVerifySingleIndexRow_validIndexRowCount_moreActual() throws IOException {
+        IndexToolVerificationResult.PhaseResult expectedPR = getValidPhaseResult();
+        for (Map.Entry<byte[], List<Mutation>>
+                entry : indexKeyToMutationMapLocal.entrySet()) {
+            initializeLocalMockitoSetup(entry, TestType.VALID_MORE_MUTATIONS);
+            //test code
+            rebuildScanner.verifySingleIndexRow(indexRow, actualPR);
+            //assert
```
gjacoby126: nit: don't have to tell me you assert just before a line called assert. :-) Ditto for the other places this got pasted.

phoenix-core/src/test/java/org/apache/phoenix/index/IndexToolSingleRowVerifyTest.java
```diff
@@ -0,0 +1,634 @@
...
+        if(testType.toString().startsWith("INVALID")) {
+            return getInvalidActualMutations(testType, actualMutations);
+        }
+        return null;
+    }
+
+    private List <Mutation> getValidActualMutations(TestType testType,
+            List<Mutation> actualMutations) {
+        List <Mutation> newActualMutations = new ArrayList<>();
+        if(testType.equals(TestType.VALID)) {
```
gjacoby126: should this be VALID_EMPTY?

phoenix-core/src/test/java/org/apache/phoenix/index/IndexToolSingleRowVerifyTest.java
```diff
@@ -0,0 +1,634 @@
...
+        List <Mutation> newActualMutations = new ArrayList<>();
+        newActualMutations.addAll(actualMutations);
+        for (Mutation m : actualMutations) {
+            newActualMutations.remove(m);
+            NavigableMap<byte[], List<Cell>> familyCellMap = m.getFamilyCellMap();
+            List<Cell> cellList = familyCellMap.firstEntry().getValue();
+            List<Cell> newCellList = new ArrayList<>();
+            byte[] fam = CellUtil.cloneFamily(cellList.get(0));
+            for (Cell c : cellList) {
+                infiltrateCell(c, newCellList, testType);
```
gjacoby126: Nice method name. :-)

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/IndexRebuildRegionScanner.java
swaroopak: Also, for tests, it was 3 level indirection to access the PhaseResult, so I thought of putting it into its own class. :)

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/IndexToolVerificationResult.java
```diff
@@ -0,0 +1,294 @@
...
+            return "PhaseResult{" +
+                    "validIndexRowCount=" + validIndexRowCount +
+                    ", expiredIndexRowCount=" + expiredIndexRowCount +
+                    ", missingIndexRowCount=" + missingIndexRowCount +
+                    ", invalidIndexRowCount=" + invalidIndexRowCount +
+                    '}';
+        }
+
+        @Override
+        public boolean equals(Object o) {
```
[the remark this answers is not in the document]
swaroopak: aah! ll do it with 17 and 31 way. :)

phoenix-core/src/test/java/org/apache/phoenix/index/IndexToolSingleRowVerifyTest.java
swaroopak: making it VALID_EXACT_MATCH.

phoenix-core/src/test/java/org/apache/phoenix/index/IndexToolSingleRowVerifyTest.java
```diff
@@ -0,0 +1,634 @@
...
+    public static final String completeRowUpsert = "UPSERT INTO %s VALUES (?,?,?,?)";
+    public static final String partialRowUpsert1 = "UPSERT INTO %s (%s, %s, %s) VALUES (?,?,?)";
+    public static final String deleteRowDML = "DELETE FROM %s WHERE %s = ?  AND %s = ?";
+    public static final String INCLUDED_COLUMN = "0:FIRST_VALUE";
+
+    @Rule
+    public ExpectedException exceptionRule = ExpectedException.none();
+
+    private enum TestType {
+        VALID,
```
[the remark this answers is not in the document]
swaroopak: added comment wherever felt necessary

PR #727, opened 2020-03-09
PHOENIX-5762 Update jackson

stoty:


mihir6692:
lgtm +1

pom.xml
```diff
@@ -87,7 +87,7 @@
 
     <!-- Dependency versions -->
     <commons-cli.version>1.4</commons-cli.version>
-    <jackson.version>1.9.2</jackson.version>
+    <jackson.version>2.10.3</jackson.version>
```
joshelser: We don't store any jackson-serialized data anywhere, right? The file change list looks like it's not persisted (just used in processing and then thrown away).

I'd be concerned about serde compatibility issues from Jackson 1.x to 2.x, but hopefully that is not a concern at all if we're not holding on to that data anywhere.

stoty: We only use jackson in the three files that are touched in this commit.
For deserializing JSON data that we bulk load, and for serializing the output of the trace servlet for the tracing webapp.

Neither is persisted anywhere (well, not in JSON form anyway)

stoty:
Already merged

PR #729, opened 2020-03-10
PHOENIX-5766 PhoenixMetricsIT failure in 4.x for HBase 1.3

stoty:
disable HBase global metrics test for HBase 1.3 profile

### DIFF
--- a/phoenix-core/src/it/java/org/apache/phoenix/monitoring/PhoenixMetricsIT.java
+++ b/phoenix-core/src/it/java/org/apache/phoenix/monitoring/PhoenixMetricsIT.java
@@ -62,6 +62,7 @@ import java.util.concurrent.Executors;
 import java.util.concurrent.Future;
 
 import org.apache.hadoop.metrics2.AbstractMetric;
+import org.apache.phoenix.compat.hbase.HbaseCompatCapabilities;
 import org.apache.phoenix.compile.StatementContext;
 import org.apache.phoenix.exception.SQLExceptionCode;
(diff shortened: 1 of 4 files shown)


dbwong:
Can you explain the need for this I thought 1.3 had metrics?  Is it not referring the the availability of https://issues.apache.org/jira/browse/HBASE-12911 ?

stoty:
@dbwong you are right, the comments are not specific enough.
These tests are for https://issues.apache.org/jira/browse/PHOENIX-3655 
https://issues.apache.org/jira/browse/PHOENIX-4857 details which modules are missing from 1.3

If you compare with the 1.3 branch, the bypassed tests are not even present there.

This was one of the things that had to compartmentalised for the 4.x unification, but I somehow missed these tests.

stoty:
renamed the variable, and made comments more explicit

dbwong:
Okay, I think I understand whats going on more here.  I do like the branch merging work but I don't really think this is the right approach in our tests.  Either we should break out the tests and exclude them in your newly created profiles.  Or have a hbase-compat interface that we can query about functionality.  As HBase/Phoenix continue to evolve something like that would likely be useful for maintaining the versions/changes.  Essentially a class that has isHbaseMetricsApiAdapterAvailable and then underneath you could have this class loader approach.  The tests, or other classes could then call into this interface for determining code flow.  Thoughts @stoty  ?  Maybe you already opened/have a JIRA on this I haven't been keeping up in detail.

stoty:
That's a good idea.
I've implemented it, please review the new version.

PR #730, opened 2020-03-10
PHOENIX-5767 Add 4.x branch to the precommit script configuration

stoty:


stoty:
AFAIK, only master, 4.x, and the 4.15.x branches are active. However, I did not want risk delaying the patch if someone disagrees, so I've just them as is.
This is one more area that could use an update.

stoty:
I've just realised that the 4.15.x branches are not even in the list.

PR #733, opened 2020-03-12
PHOENIX-5732: Implement starttime, endtime in IndexTool for rebuild a…

swaroopak:
…nd verification

### DIFF
--- a/phoenix-core/src/it/java/org/apache/phoenix/end2end/IndexToolTimeRangeIT.java
+++ b/phoenix-core/src/it/java/org/apache/phoenix/end2end/IndexToolTimeRangeIT.java
@@ -1,0 +1,218 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.phoenix.end2end;
+
+import com.google.common.collect.Maps;
+import org.apache.phoenix.mapreduce.index.IndexTool;
+import org.apache.phoenix.query.QueryServices;
+import org.apache.phoenix.query.QueryServicesOptions;
+import org.apache.phoenix.util.EnvironmentEdge;
+import org.apache.phoenix.util.EnvironmentEdgeManager;
+import org.apache.phoenix.util.PropertiesUtil;
+import org.apache.phoenix.util.ReadOnlyProps;
+import org.apache.phoenix.util.SchemaUtil;
+import org.junit.AfterClass;
+import org.junit.Assert;
+import org.junit.Before;
+import org.junit.BeforeClass;
+import org.junit.Ignore;
+import org.junit.Rule;
+import org.junit.Test;
+import org.junit.rules.ExpectedException;
+
+import java.sql.Connection;
+import java.sql.DriverManager;
+import java.sql.PreparedStatement;
+import java.sql.ResultSet;
+import java.sql.SQLException;
+import java.util.Map;
+import static org.apache.phoenix.util.TestUtil.TEST_PROPERTIES;
+
+public class IndexToolTimeRangeIT extends BaseUniqueNamesOwnClusterIT {
+    private static final String
+            CREATE_TABLE_DDL = "CREATE TABLE %s (ID INTEGER NOT NULL PRIMARY KEY, "
+            + "VAL1 INTEGER, VAL2 INTEGER) COLUMN_ENCODED_BYTES=0";
+    public static final String CREATE_INDEX_DDL = "CREATE INDEX %s ON %s (VAL1) INCLUDE (VAL2)";
+    private static final String UPSERT_TABLE_DML = "UPSERT INTO %s VALUES(?,?,?)";
+
+    private static String dataTableFullName, indexTableFullName,
+            schemaName, dataTableName, indexTableName;
+    static CustomEnvironmentEdge customEdge = new CustomEnvironmentEdge();
+
+    @BeforeClass
+    public static synchronized void setup() throws Exception {
+        setupMiniCluster();
+        createTableAndIndex();
+        populateDataTable();
+    }
+
+    private static void createTableAndIndex() throws SQLException {
+        schemaName = generateUniqueName();
+        dataTableName = "D_"+generateUniqueName();
+        dataTableFullName = SchemaUtil.getTableName(schemaName, dataTableName);
+        indexTableName = "I_"+generateUniqueName();
+        indexTableFullName = SchemaUtil.getTableName(schemaName, indexTableName);
+        try (Connection conn = getConnection()) {
+            incrementEdgeByOne();
+            conn.createStatement().execute(String.format(CREATE_TABLE_DDL, dataTableFullName));
+            conn.commit();
+            incrementEdgeByOne();
+            conn.createStatement().execute(String.format(CREATE_INDEX_DDL, indexTableName,
+                    dataTableFullName));
+            conn.commit();
+        }
+    }
+
+    private static void incrementEdgeByOne() {
+        customEdge.incrementValue(1);
+        EnvironmentEdgeManager.injectEdge(customEdge);
+    }
+
+    private static void populateDataTable() throws SQLException {
+        try (Connection conn = getConnection()) {
+            //row 1-> time 4, row 2-> time 5, row 3-> time 6, row 4-> time 7, row 5-> time 8
+            for (int i=0; i<5; i++) {
+                incrementEdgeByOne();
+                PreparedStatement ps = conn.prepareStatement(
+                        String.format(UPSERT_TABLE_DML, dataTableFullName));
+                ps.setInt(1, i+1);
+                ps.setInt(2,(i+1)*10);
+                ps.setInt(3, (i+1)*100);
+                ps.execute();
+                conn.commit();
+            }
+        }
+    }
+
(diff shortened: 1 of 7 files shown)


phoenix-core/src/it/java/org/apache/phoenix/end2end/IndexToolTimeRangeIT.java
```diff
@@ -0,0 +1,215 @@
...
+        String [] args = {"--deleteall"};
+        runIndexTool(args, 0);
+        // all rows should be rebuilt
+        Assert.assertEquals(5, countRowsInIndex());
+    }
+
+    @Test
+    public void testValidTimeRange_onlyStartTimePassed() throws Exception {
+        //starttime passed beyond last upsert time
+        String [] args = {"--deleteall", "--starttime", "10"};
```
priyankporwal: Perhaps use starttime 8 and assert 1 row rebuilt?

phoenix-core/src/it/java/org/apache/phoenix/end2end/IndexToolTimeRangeIT.java
```diff
@@ -0,0 +1,215 @@
...
+                PropertiesUtil.deepCopy(TEST_PROPERTIES));
+    }
+
+    @Before
+    public void beforeTest() {
+        incrementEdgeByOne();
+    }
+
+    @Test
+    public void testValidTimeRange() throws Exception {
```
priyankporwal: Nice set of tests @swaroopak! Is there an error-case test where starttime > endtime?

swaroopak: Thanks! :) Yes, I have added them to the IndexToolTest class.

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/index/IndexTool.java
```diff
@@ -801,40 +741,137 @@ public int run(String[] args) throws Exception {
...
+        basePath = cmdLine.getOptionValue(OUTPUT_PATH_OPTION.getOpt());
+        isForeground = cmdLine.hasOption(RUN_FOREGROUND_OPTION.getOpt());
+        useSnapshot = cmdLine.hasOption(SNAPSHOT_OPTION.getOpt());
+        shouldDeleteBeforeRebuild = cmdLine.hasOption(DELETE_ALL_AND_REBUILD_OPTION.getOpt());
+    }
+
+    private void validateTimeRange() {
+        Long currentTime = EnvironmentEdgeManager.currentTimeMillis();
+        if (startTime != null && (startTime.compareTo(currentTime) > 0 ||
+                (endTime != null && startTime.compareTo(endTime) >= 0))) {
```
kadirozde: We should add a check to make sure the end time is not in the future by comparing it with the current time.

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/PhoenixServerBuildIndexInputFormat.java
```diff
@@ -85,14 +89,17 @@ protected  QueryPlan getQueryPlan(final JobContext context, final Configuration
             Long scn = (currentScnValue != null) ? Long.valueOf(currentScnValue) : EnvironmentEdgeManager.currentTimeMillis();
             configuration.set(PhoenixConfigurationUtil.CURRENT_SCN_VALUE,
                     Long.toString(scn));
+
```
kadirozde: We should also change the line 89 above and use the current time - 60 seconds  instead of using the current time for setting scn (as we do in IndexScrutiny). Then we need to sleep 60 second in the index tool before starting the MR job. If you prefer to do it in a different jira, that is also fine with me.

priyankporwal: Good idea @kadirozde! Looks like a small enough change to include in this change.

swaroopak: I would prefer new Jira as it needs a better review covering every scenario and lots of ground for testing.

phoenix-core/src/it/java/org/apache/phoenix/end2end/IndexToolTimeRangeIT.java
```diff
@@ -0,0 +1,215 @@
...
+    public void testNoTimeRangePassed() throws Exception {
+        String [] args = {"--deleteall"};
+        runIndexTool(args, 0);
+        // all rows should be rebuilt
+        Assert.assertEquals(5, countRowsInIndex());
+    }
+
+    @Test
+    public void testValidTimeRange_onlyStartTimePassed() throws Exception {
+        //starttime passed beyond last upsert time
```
priyankporwal: Nit: this comments needs to be updated to reflect the change.

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/index/IndexTool.java
```diff
@@ -801,40 +741,139 @@ public int run(String[] args) throws Exception {
...
+        qDataTable = SchemaUtil.getQualifiedTableName(schemaName, dataTable);
+        basePath = cmdLine.getOptionValue(OUTPUT_PATH_OPTION.getOpt());
+        isForeground = cmdLine.hasOption(RUN_FOREGROUND_OPTION.getOpt());
+        useSnapshot = cmdLine.hasOption(SNAPSHOT_OPTION.getOpt());
+        shouldDeleteBeforeRebuild = cmdLine.hasOption(DELETE_ALL_AND_REBUILD_OPTION.getOpt());
+    }
+
+    private void validateTimeRange() {
+        Long currentTime = EnvironmentEdgeManager.currentTimeMillis();
+        if (startTime != null
```
priyankporwal: The if checks are missing the case when startTime == null and endTime is in future. When you see opening/closing parenthesis like (((( or )))), please treat those as a sign of the code needing some improvements.

e.g. of simplification:
Long st = (startTime == null) ? 0 : startTime;
Long et = (endTime == null) ? Long.MAX : endTime;
if (st.compareTo(currentTime) > 0 || 
    et.compareTo(currentTime) > 0 || 
    st..compareTo(et) >= et) {
    throw new RuntimeException(INVALID_TIME_RANGE_EXCEPTION_MESSAGE);
}

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/index/IndexTool.java
```diff
@@ -271,8 +275,17 @@ public static IndexVerifyType fromValue(byte[] value) {
...
+    private static final Option START_TIME_OPTION = new Option("st", "starttime",
+            true, "Start time for indextool rebuild or verify");
+    private static final Option END_TIME_OPTION = new Option("et", "endtime",
+            true, "End time for indextool rebuild or verify");
+
     public static final String INDEX_JOB_NAME_TEMPLATE = "PHOENIX_%s.%s_INDX_%s";
 
+    public static final String INVALID_TIME_RANGE_EXCEPTION_MESSAGE = "startTime is greater than "
+            + "or equal to endTime "
+            + "or startTime is set in future; IndexTool can't proceed.";
```
priyankporwal: or either of them are in the future??

phoenix-core/src/test/java/org/apache/phoenix/index/IndexToolTest.java
```diff
@@ -0,0 +1,159 @@
...
+        CommandLine cmdLine = it.parseOptions(args);
+        exceptionRule.expect(RuntimeException.class);
+        exceptionRule.expectMessage(INVALID_TIME_RANGE_EXCEPTION_MESSAGE);
+        it.populateIndexToolAttributes(cmdLine);
+    }
+
+    @Test(timeout = 10000 /* 10 secs */)
+    public void testParseOptions_timeRange_endTimeInFuture() {
+        Long startTime = EnvironmentEdgeManager.currentTimeMillis();
+        Long endTime = EnvironmentEdgeManager.currentTimeMillis() + 100000;
```
priyankporwal: This test didn't catch the bug in the time verification code since startTime is non-null here. TDD approach would be make startTime null here and observe the test failing then fix the verification code :)

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/index/IndexTool.java
swaroopak: That's a very apt suggestion! Noting it down to take it long way.
Thank you for catching the missed case :)

swaroopak: btw, setting `et` to long max won't allow passing valid startTime with endTime = null. I have updated the PR. Please review, @priyankporwal

phoenix-core/src/it/java/org/apache/phoenix/end2end/IndexToolTimeRangeIT.java
```diff
@@ -0,0 +1,215 @@
...
+        setupMiniCluster();
+        createTableAndIndex();
+        populateDataTable();
+    }
+
+    private static void createTableAndIndex() throws SQLException {
+        schemaName = generateUniqueName();
+        dataTableName = generateUniqueName();
+        dataTableFullName = SchemaUtil.getTableName(schemaName, dataTableName);
+        indexTableName = generateUniqueName();
```
gokceni: nit: Add "I_" "D_" prefix

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/index/IndexTool.java
```diff
@@ -801,40 +747,155 @@ public int run(String[] args) throws Exception {
...
+        pIndexTable = PhoenixRuntime.getTable(connection, schemaName != null && !schemaName.isEmpty()
+                ? SchemaUtil.getQualifiedTableName(schemaName, indexTable) : indexTable);
+        indexType = pIndexTable.getIndexType();
+        if (schemaName != null && !schemaName.isEmpty()) {
+            qIndexTable = SchemaUtil.getQualifiedTableName(schemaName, indexTable);
+        } else {
+            qIndexTable = indexTable;
+        }
+        if (IndexType.LOCAL.equals(indexType)) {
+            isLocalIndexBuild = true;
```
gokceni: nit: I recommend adding this. for class level variables

swaroopak: This is prevalent in the file. Let's get to this in the next change in the file.

PR #735, opened 2020-03-17
PHOENIX-5734 - IndexScrutinyTool should not report rows beyond maxLoo…

gjacoby126:
…kBack age

### DIFF
--- a/phoenix-core/src/it/java/org/apache/phoenix/end2end/IndexScrutinyToolBaseIT.java
+++ b/phoenix-core/src/it/java/org/apache/phoenix/end2end/IndexScrutinyToolBaseIT.java
@@ -16,7 +16,7 @@ import org.apache.hadoop.conf.Configuration;
 import org.apache.hadoop.hbase.HConstants;
 import org.apache.hadoop.mapreduce.Counters;
 import org.apache.hadoop.mapreduce.Job;
-import org.apache.phoenix.mapreduce.index.IndexScrutinyMapperForTest;
+import org.apache.phoenix.mapreduce.index.IndexScrutinyMapper;
 import org.apache.phoenix.mapreduce.index.IndexScrutinyTool;
 import org.apache.phoenix.mapreduce.index.IndexScrutinyTool.OutputFormat;
 import org.apache.phoenix.mapreduce.index.IndexScrutinyTool.SourceTable;
@@ -54,8 +54,9 @@ public class IndexScrutinyToolBaseIT extends BaseTest {
                 new ReadOnlyProps(clientProps.entrySet().iterator()));
     }
 
-    protected List<Job> runScrutiny(String[] cmdArgs) throws Exception {
-        IndexScrutinyTool scrutiny = new IndexScrutinyTool(IndexScrutinyMapperForTest.class);
+    protected List<Job> runScrutiny(Class<? extends IndexScrutinyMapper> mapperClass,
+                                    String[] cmdArgs) throws Exception {
+        IndexScrutinyTool scrutiny = new IndexScrutinyTool(mapperClass);
         Configuration conf = new Configuration(getUtility().getConfiguration());
         scrutiny.setConf(conf);
         int status = scrutiny.run(cmdArgs);
--- a/phoenix-core/src/it/java/org/apache/phoenix/end2end/IndexScrutinyWithMaxLookbackIT.java
+++ b/phoenix-core/src/it/java/org/apache/phoenix/end2end/IndexScrutinyWithMaxLookbackIT.java
@@ -1,0 +1,304 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.phoenix.end2end;
+
+import com.google.common.collect.Maps;
+import org.apache.hadoop.hbase.TableName;
+import org.apache.hadoop.hbase.regionserver.ScanInfoUtil;
+import org.apache.hadoop.mapreduce.Counters;
+import org.apache.hadoop.mapreduce.Job;
+import org.apache.phoenix.mapreduce.index.IndexScrutinyMapper;
+import org.apache.phoenix.mapreduce.index.IndexScrutinyTableOutput;
+import org.apache.phoenix.mapreduce.index.IndexScrutinyTool;
+import org.apache.phoenix.query.QueryServices;
+import org.apache.phoenix.schema.TableNotFoundException;
+import org.apache.phoenix.util.EnvironmentEdgeManager;
+import org.apache.phoenix.util.ManualEnvironmentEdge;
+import org.apache.phoenix.util.MetaDataUtil;
+import org.apache.phoenix.util.PropertiesUtil;
+import org.apache.phoenix.util.ReadOnlyProps;
+import org.apache.phoenix.util.SchemaUtil;
+import org.apache.phoenix.util.TestUtil;
+import org.junit.Before;
+import org.junit.BeforeClass;
+import org.junit.Test;
+
+import java.sql.Connection;
+import java.sql.DriverManager;
+import java.sql.PreparedStatement;
+import java.sql.ResultSet;
+import java.sql.SQLException;
+import java.util.List;
+import java.util.Map;
+
+import static org.apache.phoenix.mapreduce.index.PhoenixScrutinyJobCounters.INVALID_ROW_COUNT;
+import static org.apache.phoenix.mapreduce.index.PhoenixScrutinyJobCounters.BEYOND_MAX_LOOKBACK_COUNT;
+import static org.apache.phoenix.mapreduce.index.PhoenixScrutinyJobCounters.VALID_ROW_COUNT;
+import static org.apache.phoenix.util.TestUtil.TEST_PROPERTIES;
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertNotNull;
+import static org.junit.Assert.assertTrue;
+
+public class IndexScrutinyWithMaxLookbackIT extends IndexScrutinyToolBaseIT {
+
+    private static PreparedStatement upsertDataStmt;
+    private static String dataTableFullName;
+    private static String schema;
+    private static String dataTableName;
+    private static String indexTableName;
+    private static String indexTableFullName;
+    private static String viewName;
+    private static boolean isViewIndex;
+    private static ManualEnvironmentEdge testClock;
+    public static final String UPSERT_DATA = "UPSERT INTO %s VALUES (?, ?, ?)";
+    public static final String DELETE_SCRUTINY_METADATA = "DELETE FROM "
+        + IndexScrutinyTableOutput.OUTPUT_METADATA_TABLE_NAME;
+    public static final String DELETE_SCRUTINY_OUTPUT = "DELETE FROM " +
+        IndexScrutinyTableOutput.OUTPUT_TABLE_NAME;
+    public static final int MAX_LOOKBACK = 6;
+    private long scrutinyTs;
+
+
+    @BeforeClass
+    public static synchronized void doSetup() throws Exception {
+        Map<String, String> props = Maps.newHashMapWithExpectedSize(2);
+        props.put(QueryServices.GLOBAL_INDEX_ROW_AGE_THRESHOLD_TO_DELETE_MS_ATTRIB, Long.toString(0));
+        props.put(ScanInfoUtil.PHOENIX_MAX_LOOKBACK_AGE_CONF_KEY,
+            Integer.toString(MAX_LOOKBACK));
+        setUpTestDriver(new ReadOnlyProps(props.entrySet().iterator()));
+    }
+
+    @Before
+    public void setupTest() throws SQLException {
+        try(Connection conn = DriverManager.getConnection(getUrl(),
+            PropertiesUtil.deepCopy(TEST_PROPERTIES))){
+            conn.createStatement().execute(DELETE_SCRUTINY_METADATA);
+            conn.createStatement().execute(DELETE_SCRUTINY_OUTPUT);
+            conn.commit();
+        } catch (TableNotFoundException tnfe){
+            //This will happen the first time
+        }
+    }
+
+    @Test
+    public void testScrutinyOnRowsBeyondMaxLookBack() throws Exception {
+        setupTables();
+        upsertDataAndScrutinize(dataTableName, dataTableFullName, testClock);
+        assertBeyondMaxLookbackOutput(dataTableFullName, indexTableFullName);
+
+    }
+
+    @Test
+    public void testScrutinyOnRowsBeyondMaxLookback_viewIndex() throws Exception {
+        schema = "S"+generateUniqueName();
+        dataTableName = "T"+generateUniqueName();
+        dataTableFullName = SchemaUtil.getTableName(schema,dataTableName);
+        indexTableName = "VI"+generateUniqueName();
+        isViewIndex = true;
+        viewName = "V"+generateUniqueName();
+        String viewFullName = SchemaUtil.getTableName(schema,viewName);
+        String indexFullName = SchemaUtil.getTableName(schema, indexTableName);
+        String dataTableDDL = "CREATE TABLE %s (ID INTEGER NOT NULL PRIMARY KEY, NAME VARCHAR, "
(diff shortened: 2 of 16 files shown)


phoenix-core/src/it/java/org/apache/phoenix/end2end/IndexScrutinyToolBaseIT.java
```diff
@@ -54,8 +54,9 @@
                 new ReadOnlyProps(clientProps.entrySet().iterator()));
     }
 
-    protected List<Job> runScrutiny(String[] cmdArgs) throws Exception {
-        IndexScrutinyTool scrutiny = new IndexScrutinyTool(IndexScrutinyMapperForTest.class);
+    protected List<Job> runScrutiny(Class<? extends IndexScrutinyMapper> mapperClass,
```
swaroopak: aah! nice :)

phoenix-core/src/it/java/org/apache/phoenix/end2end/IndexScrutinyWithMaxLookbackIT.java
```diff
@@ -0,0 +1,206 @@
...
+    private static String dataTableName;
+    private static String indexTableName;
+    private static String viewName;
+    private static boolean isViewIndex;
+    private static ManualEnvironmentEdge testClock;
+    public static final String UPSERT_DATA = "UPSERT INTO %s VALUES (?, ?, ?)";
+
+    @BeforeClass
+    public static synchronized void doSetup() throws Exception {
+        Map<String, String> props = Maps.newHashMapWithExpectedSize(1);
```
gokceni: nit : consider expected size = 2

phoenix-core/src/it/java/org/apache/phoenix/end2end/IndexScrutinyWithMaxLookbackIT.java
```diff
@@ -0,0 +1,206 @@
...
+        props.put(ScanInfoUtil.PHOENIX_MAX_LOOKBACK_AGE_CONF_KEY,
+            Integer.toString(MAX_LOOKBACK));
+        setUpTestDriver(new ReadOnlyProps(props.entrySet().iterator()));
+    }
+
+    @Test
+    public void testScrutinyOnRowsBeyondMaxLookBack() throws Exception {
+        schema = generateUniqueName();
+        dataTableName = generateUniqueName();
+        indexTableName = generateUniqueName();
```
gokceni: nit: consider prepending "I_" and "D_" to names

phoenix-core/src/it/java/org/apache/phoenix/end2end/IndexScrutinyWithMaxLookbackIT.java
```diff
@@ -0,0 +1,206 @@
...
+            // insert two rows
+            upsertDataStmt = getUpsertDataStmt(tableFullName, conn);
+
+            NonParameterizedIndexScrutinyToolIT.upsertRow(upsertDataStmt, 1, "name-1", 98051);
+            NonParameterizedIndexScrutinyToolIT.upsertRow(upsertDataStmt, 2, "name-2", 98052);
+            conn.commit();
+            long afterInsertSCN = EnvironmentEdgeManager.currentTimeMillis() + 1;
+            testClock.setValue(afterInsertSCN);
+            EnvironmentEdgeManager.injectEdge(testClock);
+            testClock.incrementValue(1);
```
gokceni: nit: Would be good to have a comment on what the clock increments are doing.

phoenix-core/src/it/java/org/apache/phoenix/end2end/IndexScrutinyWithMaxLookbackIT.java
```diff
@@ -0,0 +1,206 @@
...
+            long scrutinyTs = EnvironmentEdgeManager.currentTimeMillis();
+            changeIndexValue();
+            testClock.incrementValue(MAX_LOOKBACK /2  * 1000);
+            List<Job> completedJobs = runScrutiny(schema, tableName, indexTableName, scrutinyTs);
+            Job job = completedJobs.get(0);
+            assertTrue(job.isSuccessful());
+            assertCounters(job.getCounters());
+        }
+    }
+    private void changeIndexValue() throws SQLException {
```
gokceni: Suggest to rename this function to be add index row

phoenix-core/src/test/java/org/apache/phoenix/mapreduce/index/IndexScrutinyMapperForTest.java
```diff
@@ -23,6 +23,8 @@
 public class IndexScrutinyMapperForTest extends IndexScrutinyMapper {
 
     public static final int TEST_TABLE_TTL = 3600;
+    public static final int MAX_LOOKBACK = 6;
```
swaroopak: nit: unnecessary change

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/index/IndexScrutinyMapper.java
```diff
@@ -288,8 +301,15 @@ protected void checkIfInvalidRowsExpired(Context context,
             Pair<Long, List<Object>> sourceValues = entry.getValue();
             Long sourceTS = sourceValues.getFirst();
             if (hasRowExpiredOnSource(sourceTS, ttl)) {
-                context.getCounter(PhoenixScrutinyJobCounters.EXPIRED_ROW_COUNT).increment(1);
-                itr.remove();
+                context.getCounter(PhoenixScrutinyJobCounters.EXPIRED_ROW_COUNT).increment(1L);
+                itr.remove(); //don't output to the scrutiny table
+            } else if (isRowOlderThanMaxLookback(sourceTS)){
+                context.getCounter(PhoenixScrutinyJobCounters.BEYOND_MAX_LOOKBACK).increment(1L);
+                //still output to the scrutiny table just in case it's useful
```
swaroopak: why do we need them around?

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/index/IndexScrutinyMapper.java
```diff
@@ -299,7 +319,16 @@ protected boolean hasRowExpiredOnSource(Long sourceTS, Long ttl) {
         return ttl != Integer.MAX_VALUE && sourceTS + ttl*1000 < currentTS;
     }
 
-    private long getTableTtl() throws SQLException, IOException {
+    protected boolean isRowOlderThanMaxLookback(Long sourceTS){
```
swaroopak: nit: space before '{' this and some other places.

phoenix-core/src/it/java/org/apache/phoenix/end2end/IndexScrutinyWithMaxLookbackIT.java
```diff
@@ -0,0 +1,206 @@
...
+    }
+
+    private static class MaxLookbackIndexScrutinyMapper extends IndexScrutinyMapper {
+        @Override
+        public void postSetup(){
+            try {
+                String tableToCompact;
+                if (isViewIndex){
+                    String physicalDataTableName =
+                        SchemaUtil.getPhysicalHBaseTableName(schema, dataTableName, false).getString();
```
swaroopak: why not take isNamespaceEnabled property from the config?

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/index/IndexScrutinyMapper.java
```diff
@@ -161,12 +168,22 @@ protected void setup(final Context context) throws IOException, InterruptedExcep
...
         } catch (SQLException | NoSuchAlgorithmException e) {
             tryClosingResourceSilently(this.outputUpsertStmt);
             tryClosingResourceSilently(this.connection);
             tryClosingResourceSilently(this.outputConn);
             throw new RuntimeException(e);
         }
+        postSetup();
+    }
+
+    protected void preSetup() {
```
gokceni: Any reasons to have these empty methods?

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/index/IndexScrutinyTool.java
```diff
@@ -505,6 +509,18 @@ public int run(String[] args) throws Exception {
         }
     }
 
+    private void validateTimestamp(Configuration configuration, long ts) {
+        long maxLookBackAge = ScanInfoUtil.getMaxLookbackInMillis(configuration);
+        if (maxLookBackAge != ScanInfoUtil.DEFAULT_PHOENIX_MAX_LOOKBACK_AGE * 1000L) {
```
gokceni: Is there a chance that people coincidentally pick the same Max lookback age as the default one?

gjacoby126: The default age is the same as "turn off max lookback". It happens to be 0, which makes the 1000L not necessary, but I wanted to keep the discipline of _always_ converting to millis from seconds to avoid accidentally missing it elsewhere.

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/index/IndexScrutinyMapper.java
gjacoby126: Humans need to evaluate beyond max lookback counters to see if they're valid issues or not.

phoenix-core/src/it/java/org/apache/phoenix/end2end/IndexScrutinyWithMaxLookbackIT.java
```diff
@@ -0,0 +1,209 @@
...
+            conn.createStatement().execute(String.format(dataTableDDL, dataTableFullName));
+            conn.createStatement().execute(String.format(viewDDL, viewFullName, dataTableFullName));
+            conn.createStatement().execute(String.format(indexTableDDL, indexTableName,
+                viewFullName));
+            conn.commit();
+        }
+        upsertDataAndScrutinize(viewName, viewFullName, testClock);
+    }
+
+    private void upsertDataAndScrutinize(String tableName, String tableFullName,
```
kadirozde: Can we have tests for deletes in addition to upserts? I am also interested in the cases where rows are deleted within the max loopback window and before and after the scn.

phoenix-core/src/it/java/org/apache/phoenix/end2end/MaxLookbackIT.java
```diff
@@ -317,16 +300,16 @@ public void testRecentMaxVersionsNotCompactedAway() throws Exception {
...
             //after major compaction, check to make sure we can see all three versions
             // at the appropriate times
             assertMultiVersionLookbacks(dataTableSelectSql, allValues, allSCNs);
             assertMultiVersionLookbacks(indexTableSelectSql, allValues, allSCNs);
             injectEdge.incrementValue(MAX_LOOKBACK_AGE * 1000);
             long afterLookbackAgeSCN = EnvironmentEdgeManager.currentTimeMillis();
-            majorCompact(dataTable, afterLookbackAgeSCN);
-            majorCompact(indexTable, afterLookbackAgeSCN);
+            majorCompact(dataTable);
+            majorCompact(indexTable);
```
kadirozde: In the tests above, major compactions run on both the data and index table. Can we pick only one table for one test, the other for another test, and so on? The interesting problems arise when compaction runs on one table but not on both

gjacoby126: First, this test suite only was modified because I needed to steal its major compaction code and put it into TestUtil. It's otherwise unrelated to this JIRA, and if it needs changes, that should be a separate issue. 

Second, this isn't really a test of indexing per se, but of whether the max lookback coproc hooks correctly prevent data from being purged during the lookback window.

gjacoby126: But if there are additional indexing tests we need, definitely please suggest them in a JIRA!

phoenix-core/src/it/java/org/apache/phoenix/end2end/IndexScrutinyWithMaxLookbackIT.java
gjacoby126: Good idea, I will add a test for deletes.

phoenix-core/src/it/java/org/apache/phoenix/end2end/MaxLookbackIT.java
kadirozde: That is fair. Thanks.

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/index/PhoenixScrutinyJobCounters.java
```diff
@@ -41,5 +41,11 @@
...
      * Number of batches processed
      */
-    BATCHES_PROCESSED_COUNT;
+    BATCHES_PROCESSED_COUNT,
+    /**
+     * Number of rows in source that became older than the max lookback age while scrutiny
+     * was comparing them with the target, and didn't match. We break these out separately because
+     * they could be due to extra versions being compacted, and are harmless.
+     */
+    BEYOND_MAX_LOOKBACK;
```
priyankporwal: All counters end with '_COUNT' here ... Perhaps we should remain consistent. That said, 'BEYOND_MAX_LOOKBACK_ROW_COUNT' looks too long. How about 'NO_LOOKBACK_ROW_COUNT'?

phoenix-core/src/main/java/org/apache/phoenix/util/SchemaUtil.java
```diff
@@ -1183,6 +1186,22 @@ public static int getIsNullableInt(boolean isNullable) {
...
+    }
+
+    public static int getTimeToLive(PhoenixConnection conn, byte[] tableQualifier)
+     throws SQLException {
+        HTableDescriptor td = conn.getQueryServices().getTableDescriptor(tableQualifier);
+        HColumnDescriptor[] cds = td.getColumnFamilies();
+        if (cds.length > 0){
+            return cds[0].getTimeToLive();
+        } else {
+            return HConstants.FOREVER;
```
priyankporwal: when would this be the case? i.e. #CF==0

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/index/PhoenixScrutinyJobCounters.java
gjacoby126: No objection to appending _COUNT to the existing enum value, but NO_LOOKBACK_ROW_COUNT seems more confusing to me.

phoenix-core/src/main/java/org/apache/phoenix/util/SchemaUtil.java
gjacoby126: I couldn't think of one, but decided to code the function defensively.

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/index/PhoenixScrutinyJobCounters.java
priyankporwal: True. They are similar to 'EXPIRED' but can't use the same name... 'UNREADABLE' can also be confusing. Anyways, pick what sounds best to you.

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/index/IndexScrutinyMapper.java
priyankporwal: @gjacoby126 Any reason to be concerned about bloating the scrutiny table? This is an expected race condition that can happen pretty regularly if rows are updated at the cadence of around MaxLookbackAge. Humans may not necessarily have bandwidth to look at potentially large number of rows. If we are going to emit to scrutiny table, perhaps we should add a way to filter these out as well to make them manageable.

phoenix-core/src/it/java/org/apache/phoenix/end2end/IndexScrutinyWithMaxLookbackIT.java
swaroopak: @gjacoby126

gjacoby126: I know it's going to be false.

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/index/IndexScrutinyTableOutput.java
```diff
@@ -68,14 +68,18 @@
...
+            "    BEYOND_MAX_LOOKBACK BOOLEAN,\n" +
             "    CONSTRAINT PK PRIMARY KEY\n" +
             "    (\n" +
             "        " + SOURCE_TABLE_COL_NAME + ",\n" +
             "        " + TARGET_TABLE_COL_NAME + ",\n" +
             "        " + SCRUTINY_EXECUTE_TIME_COL_NAME + ",\n" + // time at which the scrutiny ran
             "        SOURCE_ROW_PK_HASH\n" + //  this hash makes the PK unique
             "    )\n" + // dynamic columns consisting of the source and target columns will follow
-            ")";
+            ")  COLUMN_ENCODED_BYTES = 0 "; //column encoding not supported with dyn columns (PHOENIX-5107)
```
priyankporwal: Do we need to do anything to disable COLUMN_ENCODIG for clusters with this table already existing?

gjacoby126:
Thanks for the reviews @swaroopak @priyankporwal @kadirozde

PR #737, opened 2020-03-18
PHOENIX-5773 Index tool output tables should support multiple simulta…

kadirozde:
…neous rebuilds

### DIFF
--- a/phoenix-core/src/main/java/org/apache/phoenix/coprocessor/IndexRebuildRegionScanner.java
+++ b/phoenix-core/src/main/java/org/apache/phoenix/coprocessor/IndexRebuildRegionScanner.java
@@ -118,6 +118,7 @@ public class IndexRebuildRegionScanner extends BaseRegionScanner {
     public static final String NO_EXPECTED_MUTATION = "No expected mutation";
     public static final String
             ACTUAL_MUTATION_IS_NULL_OR_EMPTY = "actualMutationList is null or empty";
+    public static final byte[] ROW_KEY_SEPARATOR_BYTE = Bytes.toBytes("|");
     private long pageSizeInRows = Long.MAX_VALUE;
     private int rowCountPerTask;
     private boolean hasMore;
@@ -257,16 +258,29 @@ public class IndexRebuildRegionScanner extends BaseRegionScanner {
         long scanMaxTs = scan.getTimeRange().getMax();
         byte[] keyPrefix = Bytes.toBytes(Long.toString(scanMaxTs));
         byte[] regionName = Bytes.toBytes(region.getRegionInfo().getRegionNameAsString());
-        // The row key for the result table is the max timestamp of the scan + the table region name + scan start row
-        // + scan stop row
-        byte[] rowKey = new byte[keyPrefix.length + regionName.length + scan.getStartRow().length +
-                scan.getStopRow().length];
-        Bytes.putBytes(rowKey, 0, keyPrefix, 0, keyPrefix.length);
-        Bytes.putBytes(rowKey, keyPrefix.length, regionName, 0, regionName.length);
-        Bytes.putBytes(rowKey, keyPrefix.length + regionName.length, scan.getStartRow(), 0,
-                scan.getStartRow().length);
-        Bytes.putBytes(rowKey, keyPrefix.length + regionName.length + scan.getStartRow().length,
-                scan.getStopRow(), 0, scan.getStopRow().length);
+        int targetOffset = 0;
+        // The row key for the result table : timestamp | index table name | datable table region name |
+        //                                    scan start row | scan stop row
+        byte[] rowKey = new byte[keyPrefix.length + ROW_KEY_SEPARATOR_BYTE.length + indexHTable.getName().toBytes().length +
+                ROW_KEY_SEPARATOR_BYTE.length + regionName.length + ROW_KEY_SEPARATOR_BYTE.length +
+                scan.getStartRow().length + ROW_KEY_SEPARATOR_BYTE.length + scan.getStopRow().length];
+        Bytes.putBytes(rowKey, targetOffset, keyPrefix, 0, keyPrefix.length);
+        targetOffset += keyPrefix.length;
+        Bytes.putBytes(rowKey, targetOffset, ROW_KEY_SEPARATOR_BYTE, 0, ROW_KEY_SEPARATOR_BYTE.length);
+        targetOffset += ROW_KEY_SEPARATOR_BYTE.length;
+        Bytes.putBytes(rowKey, targetOffset, indexHTable.getName().toBytes(), 0, indexHTable.getName().toBytes().length);
+        targetOffset += indexHTable.getName().toBytes().length;
+        Bytes.putBytes(rowKey, targetOffset, ROW_KEY_SEPARATOR_BYTE, 0, ROW_KEY_SEPARATOR_BYTE.length);
+        targetOffset += ROW_KEY_SEPARATOR_BYTE.length;
+        Bytes.putBytes(rowKey, targetOffset, regionName, 0, regionName.length);
+        targetOffset += regionName.length;
+        Bytes.putBytes(rowKey, targetOffset, ROW_KEY_SEPARATOR_BYTE, 0, ROW_KEY_SEPARATOR_BYTE.length);
+        targetOffset += ROW_KEY_SEPARATOR_BYTE.length;
+        Bytes.putBytes(rowKey, targetOffset, scan.getStartRow(), 0, scan.getStartRow().length);
+        targetOffset += scan.getStartRow().length;
+        Bytes.putBytes(rowKey, targetOffset, ROW_KEY_SEPARATOR_BYTE, 0, ROW_KEY_SEPARATOR_BYTE.length);
+        targetOffset += ROW_KEY_SEPARATOR_BYTE.length;
+        Bytes.putBytes(rowKey, targetOffset, scan.getStopRow(), 0, scan.getStopRow().length);
         Put put = new Put(rowKey);
         put.addColumn(RESULT_TABLE_COLUMN_FAMILY, SCANNED_DATA_ROW_COUNT_BYTES,
                 scanMaxTs, Bytes.toBytes(Long.toString(verificationResult.scannedDataRowCount)));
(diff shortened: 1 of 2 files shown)


phoenix-core/src/main/java/org/apache/phoenix/coprocessor/IndexRebuildRegionScanner.java
```diff
@@ -257,16 +258,29 @@ private void logToIndexToolResultTable() throws IOException {
...
-        Bytes.putBytes(rowKey, 0, keyPrefix, 0, keyPrefix.length);
-        Bytes.putBytes(rowKey, keyPrefix.length, regionName, 0, regionName.length);
-        Bytes.putBytes(rowKey, keyPrefix.length + regionName.length, scan.getStartRow(), 0,
-                scan.getStartRow().length);
-        Bytes.putBytes(rowKey, keyPrefix.length + regionName.length + scan.getStartRow().length,
-                scan.getStopRow(), 0, scan.getStopRow().length);
+        int targetOffset = 0;
+        // The row key for the result table : timestamp | index table name | datable table region name |
+        //                                    scan start row | scan stop row
+        byte[] rowKey = new byte[keyPrefix.length + ROW_KEY_SEPARATOR_BYTE.length + indexHTable.getName().toBytes().length +
```
gjacoby126: Would be good to extract row key building into its own function

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/IndexRebuildRegionScanner.java
```diff
@@ -257,16 +258,29 @@ private void logToIndexToolResultTable() throws IOException {
...
+        Bytes.putBytes(rowKey, targetOffset, indexHTable.getName().toBytes(), 0, indexHTable.getName().toBytes().length);
+        targetOffset += indexHTable.getName().toBytes().length;
+        Bytes.putBytes(rowKey, targetOffset, ROW_KEY_SEPARATOR_BYTE, 0, ROW_KEY_SEPARATOR_BYTE.length);
+        targetOffset += ROW_KEY_SEPARATOR_BYTE.length;
+        Bytes.putBytes(rowKey, targetOffset, regionName, 0, regionName.length);
+        targetOffset += regionName.length;
+        Bytes.putBytes(rowKey, targetOffset, ROW_KEY_SEPARATOR_BYTE, 0, ROW_KEY_SEPARATOR_BYTE.length);
+        targetOffset += ROW_KEY_SEPARATOR_BYTE.length;
+        Bytes.putBytes(rowKey, targetOffset, scan.getStartRow(), 0, scan.getStartRow().length);
+        targetOffset += scan.getStartRow().length;
```
gjacoby126: What happens if the Scan start row contains the bytes for "|"? Does this need to be length encoded instead of using separators? Or perhaps escaped somehow?

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/IndexRebuildRegionScanner.java
```diff
@@ -412,30 +426,31 @@ public void logToIndexToolOutputTable(byte[] dataRowKey, byte[] indexRowKey, lon
...
-        } else {
-            rowKey = new byte[keyPrefix.length];
-            Bytes.putBytes(rowKey, 0, keyPrefix, 0, keyPrefix.length);
-        }
+        int targetOffset = 0;
+        // The row key for the output table : timestamp | index table name | data row key
+        rowKey = new byte[keyPrefix.length + ROW_KEY_SEPARATOR_BYTE.length + indexHTable.getName().toBytes().length +
+                ROW_KEY_SEPARATOR_BYTE.length + dataRowKey.length];
+        Bytes.putBytes(rowKey, targetOffset, keyPrefix, 0, keyPrefix.length);
+        targetOffset += keyPrefix.length;
```
gjacoby126: likewise, extracting to a method to build the output table row key would be good.

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/IndexRebuildRegionScanner.java
```diff
@@ -257,16 +258,29 @@ private void logToIndexToolResultTable() throws IOException {
...
-        // + scan stop row
-        byte[] rowKey = new byte[keyPrefix.length + regionName.length + scan.getStartRow().length +
-                scan.getStopRow().length];
-        Bytes.putBytes(rowKey, 0, keyPrefix, 0, keyPrefix.length);
-        Bytes.putBytes(rowKey, keyPrefix.length, regionName, 0, regionName.length);
-        Bytes.putBytes(rowKey, keyPrefix.length + regionName.length, scan.getStartRow(), 0,
-                scan.getStartRow().length);
-        Bytes.putBytes(rowKey, keyPrefix.length + regionName.length + scan.getStartRow().length,
-                scan.getStopRow(), 0, scan.getStopRow().length);
+        int targetOffset = 0;
```
gjacoby126: Extracting the index tool result and index tool output writing code into separate classes (with read methods!) would facilitate testing -- I agree with @swaroopak that this needs some tests, I'd say probably for both insert and read. Adding read methods would also make life easier for utilities, whether in Phoenix or outside, to be able to programmatically parse the result and output tables. Otherwise a whole lot of boilerplate byte parsing has to be rewritten each time.

gjacoby126: If you don't have the bandwidth for the read part, I'm happy to help out with it, either in this JIRA or another. If it doesn't get put into Phoenix here it's code I have to write anyway. :-)

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/IndexRebuildRegionScanner.java
kadirozde: Nothing. The separator is just for helping the human debugger copy subfields within a row key. @priyankporwal suggested this separator.

gjacoby126: Assume that a non-human has to parse this key into fields, and that this is not just for use manually in HBase shell. What happens if the Scan start row contains the bytes for "|"?

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/IndexRebuildRegionScanner.java
kadirozde: I will extract the code out to a separate method.

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/IndexRebuildRegionScanner.java
kadirozde: Will do that

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/IndexRebuildRegionScanner.java
kadirozde: Please feel free to open a JIRA to extract  index tool output writing code into separate classes and  write a unit test for them.

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/IndexRebuildRegionScanner.java
kadirozde: @gjacoby126, I am not sure what you are really asking. I did not say these separators are for programs. I said they were just there to help humans. The previous version did not have any separators. There were no requirement in this jira to add such separators. I added them based on Priyank's suggestion as he claimed it would improve debugging. If there is a requirement to parse the row key to obtain subfields then we need to have a separate jira to implement that. Please also note that the row key is constructed in a way to help debugging. For example, the timestamp values are written as variable strings instead of fixed-size long values. We need to redesign the row key and maybe to have two versions of it, one for human and the other for programs.  That deserves a separate jira.

priyankporwal: Valid points valid on both sides. I suggested a separator and pipe character as an aid for humans reading these keys. Although not perfect, it will work for programmatic access as well, assuming timestamp and index-names cannot have '|' character - by looking at first 2 occurrences only. The key may have '|' but that's a problem for another day.

kadirozde:
> A small test to cover the changed part of the code would be highly appreciated. Thanks.

@swaroopak, @gjacoby126 has offered to extract the code for index tool tables out and write unit tests for it. As part of that effort, the unit tests for index tool table row keys can be added.

gjacoby126:
@kadirozde - I offered (and am still offering) to split out the code into a separate class and write read / field parsing logic in a separate JIRA, but I believe @swaroopak 's request for a test to cover the write logic you've changed is still needed for this JIRA.

kadirozde:
> @kadirozde - I offered (and am still offering) to split out the code into a separate class and write read / field parsing logic in a separate JIRA, but I believe @swaroopak 's request for a test to cover the write logic you've changed is still needed for this JIRA.

@gjacoby126, the change is covered by many existing tests already. You cannot get verification counters correct if the key is not constructed properly. As I said before, this JIRA does not offer anything for parsing or extracting subfields within a row key.

kadirozde:
> > @kadirozde - I offered (and am still offering) to split out the code into a separate class and write read / field parsing logic in a separate JIRA, but I believe @swaroopak 's request for a test to cover the write logic you've changed is still needed for this JIRA.
> 
> @gjacoby126, the change is covered by many existing tests already. You cannot get verification counters correct if the key is not constructed properly. As I said before, this JIRA does not offer anything for parsing or extracting subfields within a row key.



> > @kadirozde - I offered (and am still offering) to split out the code into a separate class and write read / field parsing logic in a separate JIRA, but I believe @swaroopak 's request for a test to cover the write logic you've changed is still needed for this JIRA.
> 
> @gjacoby126, the change is covered by many existing tests already. You cannot get verification counters correct if the key is not constructed properly. As I said before, this JIRA does not offer anything for parsing or extracting subfields within a row key.

Okay. I have added testing for row keys.

PR #739, opened 2020-03-21
PHOENIX-5788 Handling deletes on non-existing rows

kadirozde:


phoenix-core/src/main/java/org/apache/phoenix/hbase/index/IndexRegionObserver.java
```diff
@@ -517,6 +517,11 @@ private void applyPendingDeleteMutations(MiniBatchOperationInProgress<Mutation>
             }
             Put nextDataRowState = dataRowState.getSecond();
             if (nextDataRowState == null) {
+                if (dataRowState.getFirst() == null) {
```
priyankporwal: @kadirozde Are the immutable table/indexes inherently doing these delete mutations on both data and index since they don't check for presence anyways?

kadirozde: Initially, I thought deleting a non-existing row would lead to false invalid row detection and filed this JIRA. Later I found that actually deleting a non-existing row does not confuse the IndexTool verification. This PR is just not to confuse humans when they inspect the raw mutations for mutable data and index tables as they would see a delete mutation on the data table and but no corresponding mutation on the index table. The check for non-existing row is done on the server side and thus it is not for immutable tables as you observed. The potential confusion for immutable table will be still there. I thought we should at least remove this possible confusion for mutable tables.

PR #740, opened 2020-03-23
PHOENIX-5795 Supporting selective queries for index rows updated conc…

kadirozde:
…urrently

### DIFF
--- a/phoenix-core/src/it/java/org/apache/phoenix/end2end/ConcurrentMutationsExtendedIT.java
+++ b/phoenix-core/src/it/java/org/apache/phoenix/end2end/ConcurrentMutationsExtendedIT.java
@@ -19,7 +19,10 @@ package org.apache.phoenix.end2end;
 
 import org.apache.hadoop.hbase.DoNotRetryIOException;
 import org.apache.hadoop.hbase.HBaseIOException;
+import org.apache.hadoop.hbase.TableName;
+import org.apache.hadoop.hbase.client.Admin;
 import org.apache.hadoop.hbase.client.Mutation;
+import org.apache.hadoop.hbase.client.Table;
 import org.apache.hadoop.hbase.coprocessor.ObserverContext;
 import org.apache.hadoop.hbase.coprocessor.RegionCoprocessorEnvironment;
 import org.apache.hadoop.hbase.coprocessor.SimpleRegionObserver;
@@ -28,6 +31,7 @@ import org.apache.hadoop.hbase.util.Bytes;
 
 import org.apache.phoenix.jdbc.PhoenixConnection;
 import org.apache.phoenix.mapreduce.index.IndexTool;
+import org.apache.phoenix.schema.PTable;
 import org.apache.phoenix.util.*;
 import org.junit.Ignore;
 import org.junit.Test;
@@ -65,6 +69,19 @@ public class ConcurrentMutationsExtendedIT extends ParallelStatsDisabledIT {
         // Now we rebuild the entire index table and expect that it is still good after the full rebuild
         IndexToolIT.runIndexTool(true, false, "", tableName, indexName, null,
                 0, IndexTool.IndexVerifyType.AFTER);
+        // Truncate, rebuild and verify the index table
+        PTable pIndexTable = PhoenixRuntime.getTable(conn, indexName);
+        TableName physicalTableName =
+                org.apache.hadoop.hbase.TableName.valueOf(pIndexTable.getPhysicalName().getBytes());
+        PhoenixConnection pConn = conn.unwrap(PhoenixConnection.class);
+        try (Admin admin = pConn.getQueryServices().getAdmin()) {
+            admin.disableTable(physicalTableName);
+            admin.truncateTable(physicalTableName, true);
+        }
+        IndexToolIT.runIndexTool(true, false, "", tableName, indexName, null,
+                0, IndexTool.IndexVerifyType.AFTER);
+        long actualRowCountAfterCompaction = IndexScrutiny.scrutinizeIndex(conn, tableName, indexName);
+        assertEquals(actualRowCount, actualRowCountAfterCompaction);
         return actualRowCount;
     }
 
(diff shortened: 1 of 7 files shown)


phoenix-core/src/it/java/org/apache/phoenix/end2end/ConcurrentMutationsExtendedIT.java
```diff
@@ -264,6 +281,67 @@ public void testConcurrentUpserts() throws Exception {
...
+                    }
+                }
+
+            };
+        }
+        for (int i = 0; i < nThreads; i++) {
+            Thread t = new Thread(runnables[i]);
+            t.start();
+        }
+
```
gokceni: Shouldn't you wait for the threads to finish running?

phoenix-core/src/main/java/org/apache/phoenix/hbase/index/IndexRegionObserver.java
```diff
@@ -803,7 +766,31 @@ protected PhoenixIndexMetaData getPhoenixIndexMetaData(ObserverContext<RegionCop
...
+            byte[] family = columnReference.getFamily();
+            List<Cell> cellList = familyMap.get(family);
+            if (cellList == null) {
+                return false;
+            }
+            boolean has = false;
+            for (Cell cell : cellList) {
+                if (Bytes.compareTo(CellUtil.cloneFamily(cell), family) == 0 &&
+                        Bytes.compareTo(CellUtil.cloneQualifier(cell), columnReference.getQualifier() ) == 0) {
+                    has = true;
```
gokceni: Suggestion: Change cell names and family to indicate whether they belong to index or data table.

I am not sure I understand this loop. If we have index and and data cell having same family and qualifier we just break? Should not it be if they are not equal break?

kadirozde: Please see line 177. The multiMutationMap attribute holds the data table pending rows. Also indexMaintainer.getIndexedColumns() returns the data column references for indexed columns. The data columns are grouped into three classes, pk columns (data table pk columns), the indexed columns (the columns for which we want to have indexing; they form the prefix for the primary key for the index table (after salt and tenant id)) and covered columns. The purpose of the loop is to find out if all the indexed columns are included in the pending data table mutation pointed by multiMutation. I will add these comments to the code.

phoenix-core/src/it/java/org/apache/phoenix/end2end/ConcurrentMutationsExtendedIT.java
kadirozde: It is there in the next line

phoenix-core/src/main/java/org/apache/phoenix/hbase/index/IndexRegionObserver.java
```diff
@@ -409,78 +415,35 @@ private void populatePendingRows(BatchMutateContext context) {
...
-                    }
-                }
-                // Only copy mutations if we found duplicate rows
-                // which only occurs when we're partially rebuilding
-                // the index (since we'll potentially have both a
-                // Put and a Delete mutation for the same row).
-                if (copyMutations) {
-                    // Add the mutation to the batch set
-                    ImmutableBytesPtr row = new ImmutableBytesPtr(m.getRow());
-                    MultiMutation stored = mutationsMap.get(row);
```
kadirozde: Please note I removed these lines for setting the timestamp on cells as they were redundant. It is already done before calling this method. Also note that this method is not called for the rebuild path. This another reason for restructuring. I think I also made the code for a bit more readable.

phoenix-core/src/main/java/org/apache/phoenix/hbase/index/IndexRegionObserver.java
```diff
@@ -803,7 +766,31 @@ protected PhoenixIndexMetaData getPhoenixIndexMetaData(ObserverContext<RegionCop
...
+        Map<byte[], List<Cell>> familyMap = multiMutation.getFamilyCellMap();
+        for (ColumnReference columnReference : indexMaintainer.getIndexedColumns()) {
+            byte[] family = columnReference.getFamily();
+            List<Cell> cellList = familyMap.get(family);
+            if (cellList == null) {
+                return false;
+            }
+            boolean has = false;
+            for (Cell cell : cellList) {
+                if (Bytes.compareTo(CellUtil.cloneFamily(cell), family) == 0 &&
```
gjacoby126: Can be replaced with CellUtil.matchingColumn(Cell, family, qualifier)

phoenix-core/src/main/java/org/apache/phoenix/hbase/index/IndexRegionObserver.java
```diff
@@ -826,13 +813,31 @@ private void preparePostIndexMutations(BatchMutateContext context, long now, Pho
...
+                        // locks for the rows of this batch
+                        removePendingRows(context);
+                        context.indexUpdates.clear();
+                        for (RowLock rowLock : context.rowLocks) {
+                            rowLock.release();
+                        }
+                        context.rowLocks.clear();
+                        throw new IOException("One of the concurrent mutations does not have all indexed columns. " +
+                                "The batch needs to be retried " +
+                                c.getEnvironment().getRegion().getRegionInfo().getTable().getNameAsString());
```
gjacoby126: Better to pass the TableName in rather than the whole ObserverContext, which has a lot of important dependencies that don't relate to this method.

phoenix-core/src/main/java/org/apache/phoenix/hbase/index/IndexRegionObserver.java
kadirozde: I will use it. Thx

phoenix-core/src/it/java/org/apache/phoenix/end2end/ConcurrentMutationsExtendedIT.java
```diff
@@ -65,6 +69,19 @@ private long verifyIndexTable(String tableName, String indexName, Connection con
         // Now we rebuild the entire index table and expect that it is still good after the full rebuild
         IndexToolIT.runIndexTool(true, false, "", tableName, indexName, null,
                 0, IndexTool.IndexVerifyType.AFTER);
+        // Truncate, rebuild and verify the index table
+        PTable pIndexTable = PhoenixRuntime.getTable(conn, indexName);
+        TableName physicalTableName =
+                org.apache.hadoop.hbase.TableName.valueOf(pIndexTable.getPhysicalName().getBytes());
```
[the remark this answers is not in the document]
kadirozde: ok

PR #741, opened 2020-03-24
PHOENIX-5791 Eliminate false invalid row detection due to concurrent …

kadirozde:
…updates

### DIFF
--- a/phoenix-core/src/main/java/org/apache/phoenix/coprocessor/IndexRebuildRegionScanner.java
+++ b/phoenix-core/src/main/java/org/apache/phoenix/coprocessor/IndexRebuildRegionScanner.java
@@ -605,43 +605,47 @@ public class IndexRebuildRegionScanner extends BaseRegionScanner {
         }
         return getMutationsWithSameTS(put, del);
     }
-
+    /**
+     * In this method, the actual list is repaired in memory using the expected list which is actually the output of
+     * rebuilding the index table row. The result of this repair is used only for verification.
+     */
     private void repairActualMutationList(List<Mutation> actualMutationList, List<Mutation> expectedMutationList)
             throws IOException {
-        // find the first (latest) actual unverified put mutation
-        Mutation actual = null;
-        for (Mutation mutation : actualMutationList) {
-            if (mutation instanceof Put && !isVerified((Put) mutation)) {
-                actual = mutation;
-                break;
-            }
-        }
-        if (actual == null) {
-            return;
-        }
-        long ts = getTimestamp(actual);
-        int expectedIndex;
-        int expectedListSize = expectedMutationList.size();
-        for (expectedIndex = 0; expectedIndex < expectedListSize; expectedIndex++) {
-            if (getTimestamp(expectedMutationList.get(expectedIndex)) <= ts) {
-                if (expectedIndex > 0) {
-                    expectedIndex--;
+        // Find the first (latest) actual unverified put mutation
+        List<Mutation> repairedMutationList = new ArrayList<>(expectedMutationList.size());
+        for (Mutation actual : actualMutationList) {
+            if (actual instanceof Put && !isVerified((Put) actual)) {
+                long ts = getTimestamp(actual);
+                int expectedIndex;
+                int expectedListSize = expectedMutationList.size();
+                for (expectedIndex = 0; expectedIndex < expectedListSize; expectedIndex++) {
+                    if (getTimestamp(expectedMutationList.get(expectedIndex)) <= ts) {
+                        if (expectedIndex > 0) {
+                            expectedIndex--;
+                        }
+                        break;
+                    }
                 }
+                if (expectedIndex == expectedListSize) {
+                    continue;
+                }
+                for (; expectedIndex < expectedListSize; expectedIndex++) {
+                    Mutation mutation = expectedMutationList.get(expectedIndex);
+                    if (mutation instanceof Put) {
+                        mutation = new Put((Put) mutation);
+                    } else {
+                        mutation = new Delete((Delete) mutation);
+                    }
+                    repairedMutationList.add(mutation);
+                }
+                // Since we repair the entire history, there is no need to more than once
                 break;
             }
         }
-        if (expectedIndex == expectedListSize) {
+        if (repairedMutationList.isEmpty()) {
             return;
         }
-        for (; expectedIndex < expectedListSize; expectedIndex++) {
-            Mutation mutation = expectedMutationList.get(expectedIndex);
-            if (mutation instanceof Put) {
-                mutation = new Put((Put) mutation);
-            } else {
-                mutation = new Delete((Delete) mutation);
-            }
-            actualMutationList.add(mutation);
-        }
+        actualMutationList.addAll(repairedMutationList);
         Collections.sort(actualMutationList, MUTATION_TS_DESC_COMPARATOR);
     }
 
@@ -665,7 +669,7 @@ public class IndexRebuildRegionScanner extends BaseRegionScanner {
             }
         }
     }
-    
+
     /**
      * There are two types of verification: without repair and with repair. Without-repair verification is done before
      * or after index rebuild. It is done before index rebuild to identify the rows to be rebuilt. It is done after
--- a/phoenix-core/src/main/java/org/apache/phoenix/coprocessor/IndexRebuildRegionScanner.java
+++ b/phoenix-core/src/main/java/org/apache/phoenix/coprocessor/IndexRebuildRegionScanner.java
@@ -498,11 +498,13 @@ public class IndexRebuildRegionScanner extends BaseRegionScanner {
                     errorMsg, null, null);
             return false;
         }
+        int expectedCellCount = 0;
         for (List<Cell> cells : expected.getFamilyCellMap().values()) {
             if (cells == null) {
                 continue;
             }
             for (Cell expectedCell : cells) {
+                expectedCellCount++;
                 byte[] family = CellUtil.cloneFamily(expectedCell);
                 byte[] qualifier = CellUtil.cloneQualifier(expectedCell);
                 Cell actualCell = getCell(actual, family, qualifier);
@@ -521,6 +523,20 @@ public class IndexRebuildRegionScanner extends BaseRegionScanner {
                     return false;
                 }
             }
+        }
+        int actualCellCount = 0;
+        for (List<Cell> cells : actual.getFamilyCellMap().values()) {
+            if (cells == null) {
+                continue;
+            }
+            actualCellCount += cells.size();
+        }
+        if (expectedCellCount != actualCellCount) {
+            String errorMsg = "Index has extra cells (in iteration " + iteration + ")";
+            byte[] dataKey = indexMaintainer.buildDataRowKey(new ImmutableBytesWritable(expected.getRow()), viewConstants);
+            logToIndexToolOutputTable(dataKey, expected.getRow(), getTimestamp(expected), getTimestamp(actual),
+                    errorMsg);
+            return false;
         }
         return true;
     }
@@ -674,7 +690,8 @@ public class IndexRebuildRegionScanner extends BaseRegionScanner {
      * There are two types of verification: without repair and with repair. Without-repair verification is done before
      * or after index rebuild. It is done before index rebuild to identify the rows to be rebuilt. It is done after
      * index rebuild to verify the rows that have been rebuilt. With-repair verification can be done anytime using
-     * the “-v ONLY” option to check the consistency of the index table.
+     * the “-v ONLY” option to check the consistency of the index table. Note that with-repair verification simulates
+     * read repair in-memory for the purpose of verification, but does not actually repair the data in the index.
      *
      * Unverified Rows
      *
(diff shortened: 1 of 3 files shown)


phoenix-core/src/main/java/org/apache/phoenix/coprocessor/IndexRebuildRegionScanner.java
```diff
@@ -614,38 +606,132 @@ private boolean isDeleteFamilyVersion(Mutation mutation) {
...
+     * twice with different timestamps, one for each concurrent mutation. These unverified mutations can be repaired
+     * from the data table later during HBase scans using the index read repair process. This is one of the reasons
+     * for having extra unverified rows in the index table. The other reason is the data table write failures.
+     * When a data table write fails, it leaves an unverified index row behind. These rows are never returned to clients,
+     * instead they are repaired, which means either they are rebuilt from their data table rows or they are deleted if
+     * their data table rows do not exist.
+     *
+     * Delete Family Version Markers
+     *
+     * The family version delete markers are generated by the read repair to remove extra unverified rows. They only
```
wangweiming800: delete marks is not only generated by the read repair. If there is a column delete on the indexed data, there will be family version delete marker as well.

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/IndexRebuildRegionScanner.java
```diff
@@ -614,38 +606,132 @@ private boolean isDeleteFamilyVersion(Mutation mutation) {
...
+     * For both verification types, after the expected list of index mutations is constructed for a given data table,
+     * another list called the actual list of index mutations is constructed by reading the index table row using HBase
+     * raw scan and all versions of the cells of the row are retrieved.
+     *
+     * As in the construction for the expected list, the cells are grouped into a put and a delete set. The put and
+     * delete sets for a given row are further grouped based on their timestamps into put and delete mutations such that
+     * all the cells in a mutation have the timestamps. The put and delete mutations are then sorted within a single
+     * list. Mutations in this list are sorted in ascending order of their timestamp. This list is the actual list.
+     *
+     * For the without-repair verification, unverified mutations and family version delete markers are removed from
```
wangweiming800: will it remove the innocent delete markers as my previous comment?

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/IndexRebuildRegionScanner.java
```diff
@@ -614,38 +606,132 @@ private boolean isDeleteFamilyVersion(Mutation mutation) {
...
+                mutation = new Put((Put) mutation);
+            } else {
+                mutation = new Delete((Delete) mutation);
+            }
+            actualMutationList.add(mutation);
+        }
+        Collections.sort(actualMutationList, MUTATION_TS_DESC_COMPARATOR);
+    }
+
+    private void cleanUpActualMutationList(List<Mutation> actualMutationList)
```
wangweiming800: Before this PR, I think it makes sense to ignore these mutations, as it only targets the ones with timestamp > expectedMutation.timestamp.

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/IndexRebuildRegionScanner.java
```diff
@@ -614,38 +606,132 @@ private boolean isDeleteFamilyVersion(Mutation mutation) {
         return getMutationsWithSameTS(put, del);
     }
 
+    private void repairActualMutationList(List<Mutation> actualMutationList, List<Mutation> expectedMutationList)
```
wangweiming800: I am just thinking aloud here. Read repair has similar logic to generate the expected index mutation, but only for specific timestamp. Is it possible to leverage the code there or unify them?
One side question. From the code of repairIndexRows, it looks like opening a lot of scan. If there are multiple concurrent read repair happen (may be caused by some bug int the future or network issue), will it overload the hbase region server and cause the cluster not be able to recover from such network hitcup?

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/IndexRebuildRegionScanner.java
swaroopak: I don't think that would be a problem. Because the data row would have already been updated (as index row has a delete marker that means data row delete i.e. the second phase was successful) and the expected index list would be built on that data row.

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/IndexRebuildRegionScanner.java
kadirozde: This simulates the read repair given that the list of revisions for the actual and the expected (repaired form) mutations. It is much simpler than the read repair process in GlobalIndexChecker which works on one index row at a time and leverages the HBase scans on the actual table, which is not the case here. Also, GlobalIndexChecker also purges the invalid rows from HBase in addition rebuilding unverified rows. So, they are very different in nature and not sure how they can be unified.

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/IndexRebuildRegionScanner.java
kadirozde: @wangweiming800, Please see the updated the code above and let me know if you have a question on this

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/IndexRebuildRegionScanner.java
kadirozde: @wangweiming800, "If there is a column delete on the indexed column in data table, there will be family version delete marker as well" is not correct. We do not convert delete column markers to delete family markers. Actually, delete column markers translate to missing columns in the index row mutations.

wangweiming800: thanks for explanation.

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/IndexRebuildRegionScanner.java
wangweiming800: thanks

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/IndexRebuildRegionScanner.java
wangweiming800: @kadirozde , I am thinking whether it is possible to replay the actual mutations by ignoring the unverified mutations and compare them with the expected mutations. 

If I understand correctly, there are only 4 types of mutations as a result of raw scan on index table.
unverified put: it is introduced by failure on 2nd or 3rd phase data mutation, successful data row deletion or ongoing data mutation.  
verified put: it is introduced by successful data put mutation or read repair
deleteFamily: it is introduced by successful data mutation which leads to the change or remove of the index row key. T
deleteFamilyVersion: it is introduced by the read repair.

Since verified put has the full information of the index mutation, even it is generated by the read repair. So if we replay all these 3 kinds of mutations by the time order, we are expected to get the expected mutation list. 

Do I misunderstand something here?

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/IndexRebuildRegionScanner.java
```diff
@@ -613,39 +605,132 @@ private boolean isDeleteFamilyVersion(Mutation mutation) {
...
+                    continue;
+                }
+                for (; expectedIndex < expectedListSize; expectedIndex++) {
+                    Mutation mutation = expectedMutationList.get(expectedIndex);
+                    if (mutation instanceof Put) {
+                        mutation = new Put((Put) mutation);
+                    } else {
+                        mutation = new Delete((Delete) mutation);
+                    }
+                    actualMutationList.add(mutation);
```
wangweiming800: will it be a problem to modify the list when iterating it?
https://stackoverflow.com/questions/6866238/concurrent-modification-exception-adding-to-an-arraylist
https://www.baeldung.com/java-concurrentmodificationexception

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/IndexRebuildRegionScanner.java
kadirozde: @wangweiming800, your understanding of mutation types is correct. However, I have no idea what replaying an index mutation means. Let's discuss this offline.

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/IndexRebuildRegionScanner.java
kadirozde: Yes, this should be a problem. Good catch! I will fix it.

kadirozde: This is in general a problem. I was thinking why it did not cause an issue in my tests. The reason is that we are not actually modifying the list here while iterating over it. We modify the list and then terminate the loop. Therefore, it is not a problem. Regardless, I will change the code to eliminate the confusion.

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/IndexRebuildRegionScanner.java
```diff
@@ -613,39 +605,137 @@ private boolean isDeleteFamilyVersion(Mutation mutation) {
...
-     * using the rebuild process (i.e., by replaying raw data table mutations). These sets are sets of expected
-     * index mutations, one set for each index row key. Since not all mutations in the index table have both phase
-     * (i.e., pre and post data phase) mutations, we cannot compare actual index mutations with expected one by one
-     * and expect to find them identical. We need to consider concurrent data mutation effects, data table row write
-     * failures, post index write failures. Thus, we need to allow some expected and actual mutations to be skipped
-     * during comparing actual mutations to index mutations.
+     * There are two types of verification: without repair and with repair. Without-repair verification is done before
+     * or after index rebuild. It is done before index rebuild to identify the rows to be rebuilt. It is done after
+     * index rebuild to verify the rows that have been rebuilt. With-repair verification can be done anytime using
+     * the “-v ONLY” option to check the consistency of the index table.
```
gjacoby126: "Note that with-repair verification simulates read repair in-memory for the purpose of verification, but does not actually repair the data in the index."

gjacoby126: Otherwise a very nice explanation

phoenix-core/src/test/java/org/apache/phoenix/index/VerifySingleIndexRowTest.java
```diff
@@ -350,6 +351,7 @@ public void testVerifySingleIndexRow_allUnverified() throws IOException {
         }
     }
 
+    @Ignore
```
gjacoby126: Why are we ignoring tests rather than removing or fixing them?

kadirozde: They will be fixed later. I will either fix it in this PR or let @swaroopak fix them in a separate jira. I needed to disable it to submit the patch for pre-commit testing

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/IndexRebuildRegionScanner.java
```diff
@@ -614,38 +606,132 @@ private boolean isDeleteFamilyVersion(Mutation mutation) {
...
+
+    private void cleanUpActualMutationList(List<Mutation> actualMutationList)
+            throws IOException {
+        Iterator<Mutation> iterator = actualMutationList.iterator();
+        Mutation previous = null;
+        while (iterator.hasNext()) {
+            Mutation mutation = iterator.next();
+            if ((mutation instanceof Put && !isVerified((Put) mutation)) ||
+                    (mutation instanceof Delete && isDeleteFamilyVersion(mutation))) {
+                iterator.remove();
```
wangweiming800: I am wondering whether it will cause false alarm when the last data mutation is put and fails on the 3rd phase and there is no read repair to fix this unverified row before inline validation. In this case, the unverified index row will be included in the expected mutation list, but cannot find in the actual mutation list. 
Following is the link I come up a scenario to simulate this case. We may discuss it offline
https://docs.google.com/spreadsheets/d/1FDKkcGrJobR5Jtd73vNSVfC1yDDrisyuqUdYWLde5rw/edit?usp=sharing

kadirozde: Since the update is made on the data table, the expected list will include the verified index row for this unverified row. So, it will not be missed.

wangweiming800: Yes, this corresponding index mutation will be in the expected mutation list, but it will not find the matching one in the actual list, as the unverified mutation will be removed here.

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/IndexRebuildRegionScanner.java
kadirozde: I will add the suggest comment

phoenix-core/src/test/java/org/apache/phoenix/index/VerifySingleIndexRowTest.java
kadirozde: @swaroopak and I worked together and fixed the failed unit tests. We needed to do some changes in the unit tests to adjust them for the changes of this PR. The unit tests caught couple issues in the code and we fixed them too. I now better appreciate this unit test :-)

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/IndexRebuildRegionScanner.java
kadirozde: @wangweiming800 and I discussed and decided to resolve this conversation

PR #743, opened 2020-03-25
PHOENIX-5798 IndexUpgrade tool command line improvements

tkhurana:
1. Add an option to IndexUpgrade tool to optionally rebuild indexes. By default, rebuilding is skipped.
2. Add a placeholder option to IndexUpgrade tool to passthrough the option value to the IndexTool

### DIFF
--- a/phoenix-core/src/it/java/org/apache/phoenix/end2end/ParameterizedIndexUpgradeToolIT.java
+++ b/phoenix-core/src/it/java/org/apache/phoenix/end2end/ParameterizedIndexUpgradeToolIT.java
@@ -97,6 +97,7 @@ public class ParameterizedIndexUpgradeToolIT extends BaseTest {
     private final boolean mutable;
     private final boolean upgrade;
     private final boolean isNamespaceEnabled;
+    private final boolean rebuild;
 
     private StringBuilder optionsBuilder;
     private String tableDDLOptions;
@@ -135,7 +136,7 @@ public class ParameterizedIndexUpgradeToolIT extends BaseTest {
         admin = queryServices.getAdmin();
         iut = new IndexUpgradeTool(upgrade ? UPGRADE_OP : ROLLBACK_OP, INPUT_LIST,
                 null, "/tmp/index_upgrade_" + UUID.randomUUID().toString(),
-                true, indexToolMock);
+                true, indexToolMock, rebuild);
         iut.setConf(getUtility().getConfiguration());
         iut.setTest(true);
         if (!mutable) {
@@ -298,20 +299,22 @@ public class ParameterizedIndexUpgradeToolIT extends BaseTest {
         }
     }
 
-    @Parameters(name ="IndexUpgradeToolIT_mutable={0},upgrade={1},isNamespaceEnabled={2}")
+    @Parameters(name ="IndexUpgradeToolIT_mutable={0},upgrade={1},isNamespaceEnabled={2}, rebuild={3}")
     public static synchronized Collection<Object[]> data() {
         return Arrays.asList(new Object[][] {
-            {false, false, true},
-            {true, false, false},
-            {false, true, false},
-            {true, true, true}
+            {false, false, true, false},
+            {true, false, false, true},
+            {false, true, false, false},
+            {true, true, true, true}
         });
     }
 
-    public ParameterizedIndexUpgradeToolIT(boolean mutable, boolean upgrade, boolean isNamespaceEnabled) {
+    public ParameterizedIndexUpgradeToolIT(boolean mutable, boolean upgrade,
+                                           boolean isNamespaceEnabled, boolean rebuild) {
         this.mutable = mutable;
         this.upgrade = upgrade;
         this.isNamespaceEnabled = isNamespaceEnabled;
+        this.rebuild = rebuild;
     }
 
     @Test
@@ -331,7 +334,7 @@ public class ParameterizedIndexUpgradeToolIT extends BaseTest {
             Assert.assertEquals("Index upgrade tool waited for client cache to expire "
                     + "for mutable tables", false, iut.getIsWaitComplete());
         }
-        if(upgrade) {
+        if(upgrade && rebuild) {
             //verifying if index tool was started
             Mockito.verify(indexToolMock,
                     times(11)) // for every index/view-index except index on transaction table
--- a/phoenix-core/src/main/java/org/apache/phoenix/mapreduce/index/IndexUpgradeTool.java
+++ b/phoenix-core/src/main/java/org/apache/phoenix/mapreduce/index/IndexUpgradeTool.java
@@ -98,16 +98,14 @@ public class IndexUpgradeTool extends Configured implements Tool {
     private static final Option LOG_FILE_OPTION = new Option("lf", "logfile",
             true,
             "[Optional] Log file path where the logs are written");
-    private static final Option INDEX_SYNC_REBUILD_OPTION = new Option("sr",
-            "index-sync-rebuild",
+    private static final Option INDEX_REBUILD_OPTION = new Option("rb",
+            "index-rebuild",
             false,
-            "[Optional] Whether or not synchronously rebuild the indexes; "
-                    + "default rebuild asynchronous");
-
-    private static final Option INDEX_VERIFY_OPTION = new Option("v",
-            "verify",
+            "[Optional] Rebuild the indexes");
+    private static final Option INDEX_TOOL_OPTION = new Option("tool",
+            "index-tool",
             true,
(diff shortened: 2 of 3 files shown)


tkhurana:
Please review @swaroopak @priyankporwal @gokceni

phoenix-core/src/it/java/org/apache/phoenix/end2end/ParameterizedIndexUpgradeToolIT.java
```diff
@@ -298,20 +299,22 @@ private void checkOldIndexingCoprocessors(String [] indexList, String [] tableLi
...
 
-    @Parameters(name ="IndexUpgradeToolIT_mutable={0},upgrade={1},isNamespaceEnabled={2}")
+    @Parameters(name ="IndexUpgradeToolIT_mutable={0},upgrade={1},isNamespaceEnabled={2}, rebuild={3}")
     public static synchronized Collection<Object[]> data() {
         return Arrays.asList(new Object[][] {
-            {false, false, true},
-            {true, false, false},
-            {false, true, false},
-            {true, true, true}
+            {false, false, true, false},
```
gokceni: Don't you want to test false, false, true, true  and
true, false, false, false?

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/index/IndexUpgradeTool.java
```diff
@@ -98,16 +98,14 @@
...
-            "index-sync-rebuild",
+    private static final Option INDEX_REBUILD_OPTION = new Option("rb",
+            "index-rebuild",
             false,
-            "[Optional] Whether or not synchronously rebuild the indexes; "
-                    + "default rebuild asynchronous");
-
-    private static final Option INDEX_VERIFY_OPTION = new Option("v",
-            "verify",
+            "[Optional] Rebuild the indexes");
```
gokceni: Can you change this description to include what type of build this is (async)?

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/index/IndexUpgradeTool.java
```diff
@@ -577,6 +582,10 @@ private void rebuildIndexes(Connection conn, Configuration conf, ArrayList<Strin
         if (!upgrade) {
             return;
         }
+        if (!rebuild) {
+            return;
+        }
+
```
gokceni: Do you want to log what type of options provided to IndexUpgradeTool for debugging purposes?

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/index/IndexUpgradeTool.java
```diff
@@ -707,12 +716,12 @@ private int startIndexRebuilds(HashMap<String, IndexInfo> indexInfos,
...
             list.add(tenantId);
         }
-        if (syncRebuild) {
-            list.add("-runfg");
-        }
-        if(!Strings.isNullOrEmpty(verify)) {
-            list.add("-v");
-            list.add(verify);
+
+        if(!Strings.isNullOrEmpty(indexToolOpts)) {
```
swaroopak: nit: spaces instead of tabs

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/index/IndexUpgradeTool.java
```diff
@@ -707,12 +716,12 @@ private int startIndexRebuilds(HashMap<String, IndexInfo> indexInfos,
...
+            String[] options = indexToolOpts.split("\\s+");
```
swaroopak: how would this work in a bash script? Does bash identify consecutive options like this?
-tool -v ONLY

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/index/IndexUpgradeTool.java
```diff
@@ -577,6 +582,10 @@ private void rebuildIndexes(Connection conn, Configuration conf, ArrayList<Strin
         if (!upgrade) {
             return;
         }
+        if (!rebuild) {
```
swaroopak: better to combine in one` if`?

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/index/IndexUpgradeTool.java
swaroopak: should be fine, but good to double-check?

tkhurana: They will have to be quoted. Like -tool '-v ONLY -runfg -starttime 10000'

phoenix-core/src/it/java/org/apache/phoenix/end2end/ParameterizedIndexUpgradeToolIT.java
tkhurana: rebuild parameter is related to the upgrade parameter. So I covered all the 4 permutations of upgrade and rebuild.

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/index/IndexUpgradeTool.java
tkhurana: We are already logging the options passed to IndexUpgradeTool https://github.com/apache/phoenix/blob/master/phoenix-core/src/main/java/org/apache/phoenix/mapreduce/index/IndexUpgradeTool.java#L186 and IndexTool https://github.com/apache/phoenix/blob/master/phoenix-core/src/main/java/org/apache/phoenix/mapreduce/index/IndexUpgradeTool.java#L688

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/index/IndexUpgradeTool.java
tkhurana: I double checked by running the tool on the shell

swaroopak: Is this still a tab and not space? diff looks different from `238 to 242`. Please consider using phoenix code template if not already: https://phoenix.apache.org/develop.html

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/index/IndexUpgradeTool.java
```diff
@@ -235,6 +240,11 @@ public CommandLine parseOptions(String[] args) {
                     +TABLE_OPTION.getLongOpt() + " and " + TABLE_CSV_FILE_OPTION.getLongOpt()
                     + "; specify only one.");
         }
+        if ((cmdLine.hasOption(INDEX_TOOL_OPTION.getOpt()))
```
priyankporwal: Just a thought, would it be simpler to have just one IndexUpgradeTool option which when present indicates both do-rebuild as well as give IndexTool options? We wouldn't need this check then.

phoenix-core/src/test/java/org/apache/phoenix/index/IndexUpgradeToolTest.java
```diff
@@ -55,40 +56,97 @@
...
+        Assert.assertTrue(argList.contains("-runfg"));
+        Assert.assertTrue(argList.contains("-st"));
+
+        // ensure that index tool can parse the options
+        IndexTool it = new IndexTool();
+        it.parseOptions(values);
+    }
+
+    @Test
+    public void testMalformedSpacingOptionsArePassedToIndexTool() {
```
priyankporwal: Perhaps we should add a test for garbage indexTool options being passed so that we can be certain of the tool behavior.

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/index/IndexUpgradeTool.java
tkhurana: @priyankporwal Are you suggesting something like this `-rebuild '-v ONLY'` ? I agree that would simplify some things. The only thing is that a value will always have to be passed to that option even if you want to run IndexTool with default options. Options can either always take a value or never take a value.

priyankporwal: Yeah, that's what I had in mind, but I see the need to have some value for that option would an empty string in quotes. Anyways, current solution works too.

tkhurana:
@priyankporwal added the testcase

PR #745, opened 2020-03-27
PHOENIX-5801: Connection leak when creating a view with a where condition

ChinmaySKulkarni:


ChinmaySKulkarni:
Please review @dbwong @yanxinyi

phoenix-core/src/it/java/org/apache/phoenix/monitoring/PhoenixMetricsIT.java
```diff
@@ -648,6 +648,21 @@ public void testMutationMetricsWhenUpsertingToMultipleTables() throws Exception
         }
     }
 
+    @Test
+    public void createViewWithWhereConditionNoConnLeak() throws SQLException {
```
swaroopak: Can this be a unit test by extending BaseConnectionlessqueryTest instead of IT?

ChinmaySKulkarni: Any suggestions on which UT to add this to? I can't really find one that extends `BaseConnectionlessqueryTest` and is relevant. Should be something related to metrics/view creation etc.

dbwong: I would be not consider what @swaroopak is suggesting a UT, this is  a different type of functional or integration test it sounds like.

dbwong: But on the note of UTs is it possible to add a UT to WhereConstantParser.java?  We may need a static factory and additional refactoring to do so.

ChinmaySKulkarni: I think that'll need a larger refactoring Jira. I also think WhereConstantParser should be in the util package rather than coprocessor. Once refactored, it will be possible to add unit tests for that class. Does that sound good?

ChinmaySKulkarni: @dbwong created https://issues.apache.org/jira/browse/PHOENIX-5803 for the same.

ChinmaySKulkarni:
Thanks for the reviews @gjacoby126 @swaroopak and @dbwong

PR #746, opened 2020-03-27
PHOENIX-5802: Connection leaks in UPSERT SELECT/DELETE paths due to MutatingParallelIteratorFactory iterator not being closed

ChinmaySKulkarni:


ChinmaySKulkarni:
Please review @dbwong @gjacoby126 @yanxinyi

phoenix-core/src/it/java/org/apache/phoenix/end2end/UpsertSelectIT.java
```diff
@@ -1242,11 +1384,13 @@ public void testRowTimestampColWithViewsIndexesAndSaltedTables() throws Exceptio
...
         }
 
-        // Verify that the data upserted using the tenant view can now be queried using base table and the base table index
+        // Verify that the data upserted using the tenant view can now be queried using base table
+        // and the base table index
         Date upsertedDate;
         try (Connection conn = DriverManager.getConnection(getUrl(), props)) {
             // Query the base table
-            PreparedStatement stmt = conn.prepareStatement("SELECT * FROM  " + baseTable + " WHERE TENANT_ID = ? AND PK2 >= ? AND PK2 <= ? AND PK3 = ? ");
+            PreparedStatement stmt = conn.prepareStatement("SELECT * FROM  " + baseTable +
```
swaroopak: why not use try-with-resources for statement also?

phoenix-core/src/it/java/org/apache/phoenix/end2end/UpsertSelectIT.java
```diff
@@ -1353,49 +1504,58 @@ public void testUpsertSelectWithFixedWidthNullByteSizeArray() throws Exception {
...
+        try (Connection conn = DriverManager.getConnection(getUrl(), props)) {
+            conn.createStatement().execute("upsert into " + t2 + " (id, ba) select id, " +
+                    "ARRAY[54, 1024] from " + t2 + " WHERE id = 2");
+            conn.commit();
+        }
 
-        conn = DriverManager.getConnection(getUrl(), props);
-        rs = conn.createStatement().executeQuery("select * from " + t2);
+        try (Connection conn = DriverManager.getConnection(getUrl(), props)) {
+            rs = conn.createStatement().executeQuery("select * from " + t2);
```
swaroopak: why is this space diff different from other space diffs?

ChinmaySKulkarni: Looks like a git diff quirk. This isn't doing anything different from the other lines where try-with-resources is added.

swaroopak: could this be a tab and not a space?

yanxinyi: > could this be a tab and not a space?

I think his IntelliJ automatically did it. not on the purpose of whitespace or tab hater.

phoenix-core/src/it/java/org/apache/phoenix/end2end/UpsertSelectIT.java
ChinmaySKulkarni: Will do

phoenix-core/src/it/java/org/apache/phoenix/end2end/UpsertSelectIT.java
```diff
@@ -101,145 +122,163 @@ private void testUpsertSelect(boolean createIndex, boolean saltTable) throws Exc
...
+            assertEquals(A_VALUE, rs.getString(3));
+
+            assertTrue (rs.next());
+            assertEquals("00A", rs.getString(1));
+            assertEquals("4", rs.getString(2));
+            assertEquals(A_VALUE, rs.getString(3));
 
-        assertFalse(rs.next());
-        conn.close();
+            assertFalse(rs.next());
```
ChinmaySKulkarni: looks like this was a tab before and now a space (which is consistent) so valid diff imo

phoenix-core/src/it/java/org/apache/phoenix/end2end/UpsertSelectIT.java
```diff
@@ -252,158 +291,164 @@ public void testUpsertSelectEmptyPKColumn() throws Exception {
...
-        assertEquals(null, rs.getString(1));
-        assertEquals(ROW9, rs.getString(2));
-        assertTrue(rs.getDate(3).after(now) && rs.getDate(3).before(then));
-        assertTrue(BigDecimal.valueOf(50).compareTo(rs.getBigDecimal(4)) == 0);
-        
-        assertFalse(rs.next());
-        conn.close();
-        
+
+        try (Connection conn = DriverManager.getConnection(getUrl(), props);
```
ChinmaySKulkarni: Here too, spaces added because of additional indentation because of try-with-resources.

ChinmaySKulkarni:
I have added statements to try-with-resources too. Also confirmed that whitespace diffs are valid. They're all there because of the try-with-resources block adding an additional level of indentation. Please take a look now @swaroopak @yanxinyi

phoenix-core/src/main/java/org/apache/phoenix/compile/MutatingParallelIteratorFactory.java
```diff
@@ -50,62 +51,80 @@ protected MutatingParallelIteratorFactory(PhoenixConnection connection) {
...
+                    return done ? null : tuple;
+                }
+            };
+        } catch (Throwable ex) {
+            // Catch just to make sure we close the cloned connection and then rethrow
+            try {
+                // closeQuietly only handles IOException
+                clonedConnection.close();
+            } catch (SQLException ignore) {
+                // ignore
```
dbwong: Do you think we should log here?

phoenix-core/src/main/java/org/apache/phoenix/compile/UpsertCompiler.java
```diff
@@ -175,8 +175,10 @@ private static void setValues(byte[][] values, int[] pkSlotIndex, int[] columnIn
         mutation.put(ptr, new RowMutationState(columnValues, columnValueSize, statement.getConnection().getStatementExecutionCounter(), rowTsColInfo, onDupKeyBytes));
     }
     
-    public static MutationState upsertSelect(StatementContext childContext, TableRef tableRef, RowProjector projector,
-            ResultIterator iterator, int[] columnIndexes, int[] pkSlotIndexes, boolean useServerTimestamp, boolean prefixSysColValues) throws SQLException {
+    public static MutationState upsertSelect(StatementContext childContext, TableRef tableRef,
```
dbwong: This class was just formating only correct?

dbwong:
Overall looks okay, maybe we should consider this one on the refacotring JIRA as well as we have no UTs added.

phoenix-core/src/main/java/org/apache/phoenix/compile/MutatingParallelIteratorFactory.java
ChinmaySKulkarni: Will add

phoenix-core/src/main/java/org/apache/phoenix/compile/UpsertCompiler.java
ChinmaySKulkarni: Mostly. Apart from formatting, removed some dead variable initialization before a return (MutationState `state` variable) as well.

ChinmaySKulkarni:
Added TODO for unit testing and refactoring of classes modified here in https://issues.apache.org/jira/browse/PHOENIX-5803 as well.

ChinmaySKulkarni:
Thanks for the reviews @swaroopak @dbwong @yanxinyi

PR #747, opened 2020-03-30
PHOENIX-5807 Index rows without empty column should be treated as unv…

kadirozde:
…erified

### DIFF
--- a/phoenix-core/src/it/java/org/apache/phoenix/end2end/index/GlobalIndexCheckerIT.java
+++ b/phoenix-core/src/it/java/org/apache/phoenix/end2end/index/GlobalIndexCheckerIT.java
@@ -40,9 +40,11 @@ import org.apache.phoenix.hbase.index.IndexRegionObserver;
 import org.apache.phoenix.mapreduce.index.IndexTool;
 import org.apache.phoenix.query.QueryServices;
 import org.apache.phoenix.schema.PTableImpl;
+import org.apache.phoenix.util.EnvironmentEdgeManager;
 import org.apache.phoenix.util.PhoenixRuntime;
 import org.apache.phoenix.util.QueryUtil;
 import org.apache.phoenix.util.ReadOnlyProps;
+import org.apache.phoenix.util.TestUtil;
 import org.junit.After;
 import org.junit.Before;
 import org.junit.BeforeClass;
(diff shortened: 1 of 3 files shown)


phoenix-core/src/main/java/org/apache/phoenix/index/GlobalIndexChecker.java
```diff
@@ -494,8 +476,7 @@ private boolean verifyRowAndRemoveEmptyColumn(List<Cell> cellList) throws IOExce
                     return true;
                 }
             }
-            byte[] rowKey = CellUtil.cloneRow(cell);
-            return verifyRow(rowKey);
+            return false;
```
gjacoby126: nit: comment to point out that empty column wasn't found so we're considering it unverified.

phoenix-core/src/it/java/org/apache/phoenix/end2end/index/GlobalIndexCheckerIT.java
```diff
@@ -169,6 +171,48 @@ public void testDeleteNonExistingRow() throws Exception {
...
+        if (async) {
+            // No need to run the same test twice one for async = true and the other for async = false
+            return;
+        }
+        long scn;
+        String dataTableName = generateUniqueName();
+        String indexTableName = generateUniqueName();
+        try (Connection conn = DriverManager.getConnection(getUrl())) {
+            populateTable(dataTableName); // with two rows ('a', 'ab', 'abc', 'abcd') and ('b', 'bc', 'bcd', 'bcde')
+            Thread.sleep(1000);
```
gjacoby126: This can be done with EnvironmentEdgeManager rather than a sleep, and that's faster/safer than a sleep. Just remember to increment by at least one after any DDL or DML to be able to read it.

phoenix-core/src/main/java/org/apache/phoenix/index/GlobalIndexChecker.java
kadirozde: Sure

phoenix-core/src/it/java/org/apache/phoenix/end2end/index/GlobalIndexCheckerIT.java
kadirozde: Sleep is not necessary. I forgot to remove it. Thank you for catching it

PR #750, opened 2020-04-01
PHOENIX-5799 - Inline Index Verification Output API

gjacoby126:
… 5748 rebase

### DIFF
--- a/phoenix-core/src/main/java/org/apache/phoenix/coprocessor/IndexRebuildRegionScanner.java
+++ b/phoenix-core/src/main/java/org/apache/phoenix/coprocessor/IndexRebuildRegionScanner.java
@@ -20,7 +20,6 @@ package org.apache.phoenix.coprocessor;
 import static org.apache.phoenix.hbase.index.IndexRegionObserver.VERIFIED_BYTES;
 import static org.apache.phoenix.hbase.index.IndexRegionObserver.removeEmptyColumn;
 import static org.apache.phoenix.hbase.index.write.AbstractParallelWriterIndexCommitter.INDEX_WRITER_KEEP_ALIVE_TIME_CONF_KEY;
-import static org.apache.phoenix.mapreduce.index.IndexTool.*;
 import static org.apache.phoenix.query.QueryConstants.AGG_TIMESTAMP;
 import static org.apache.phoenix.query.QueryConstants.SINGLE_COLUMN;
 import static org.apache.phoenix.query.QueryConstants.SINGLE_COLUMN_FAMILY;
@@ -82,6 +81,8 @@ import org.apache.phoenix.hbase.index.util.ImmutableBytesPtr;
 import org.apache.phoenix.hbase.index.util.IndexManagementUtil;
 import org.apache.phoenix.index.GlobalIndexChecker;
 import org.apache.phoenix.index.IndexMaintainer;
+import org.apache.phoenix.mapreduce.index.IndexVerificationOutputRepository;
+import org.apache.phoenix.mapreduce.index.IndexVerificationResultRepository;
 import org.apache.phoenix.index.PhoenixIndexCodec;
 import org.apache.phoenix.mapreduce.index.IndexTool;
 import org.apache.phoenix.query.KeyRange;
@@ -108,7 +109,6 @@ public class IndexRebuildRegionScanner extends BaseRegionScanner {
     public static final String NO_EXPECTED_MUTATION = "No expected mutation";
     public static final String
             ACTUAL_MUTATION_IS_NULL_OR_EMPTY = "actualMutationList is null or empty";
-    public static final byte[] ROW_KEY_SEPARATOR_BYTE = Bytes.toBytes("|");
     private long pageSizeInRows = Long.MAX_VALUE;
     private int rowCountPerTask;
     private boolean hasMore;
@@ -125,8 +125,6 @@ public class IndexRebuildRegionScanner extends BaseRegionScanner {
     private IndexMaintainer indexMaintainer;
     private byte[] indexRowKey = null;
     private Table indexHTable = null;
-    private Table outputHTable = null;
-    private Table resultHTable = null;
     private IndexTool.IndexVerifyType verifyType = IndexTool.IndexVerifyType.NONE;
     private boolean verify = false;
     private Map<byte[], List<Mutation>> indexKeyToMutationMap;
@@ -135,7 +133,6 @@ public class IndexRebuildRegionScanner extends BaseRegionScanner {
     private TaskBatch<Boolean> tasks;
     private String exceptionMessage;
     private UngroupedAggregateRegionObserver ungroupedAggregateRegionObserver;
-    private RegionCoprocessorEnvironment env;
     private HTableFactory hTableFactory;
     private int indexTableTTL = 0;
     private IndexToolVerificationResult verificationResult;
@@ -144,6 +141,8 @@ public class IndexRebuildRegionScanner extends BaseRegionScanner {
     private int  singleRowRebuildReturnCode;
     private Map<byte[], NavigableSet<byte[]>> familyMap;
     private byte[][] viewConstants;
+    private IndexVerificationResultRepository verificationResultRepository;
+    private IndexVerificationOutputRepository verificationOutputRepository;
     private long maxLookBackInMills;
 
     @VisibleForTesting
@@ -179,7 +178,6 @@ public class IndexRebuildRegionScanner extends BaseRegionScanner {
 
         this.innerScanner = innerScanner;
         this.region = region;
-        this.env = env;
         this.ungroupedAggregateRegionObserver = ungroupedAggregateRegionObserver;
         indexRowKey = scan.getAttribute(BaseScannerRegionObserver.INDEX_ROW_KEY);
         if (indexRowKey != null) {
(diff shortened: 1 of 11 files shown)


phoenix-core/src/main/java/org/apache/phoenix/coprocessor/IndexRebuildRegionScanner.java
```diff
@@ -154,6 +143,8 @@
     private int  singleRowRebuildReturnCode;
     private Map<byte[], NavigableSet<byte[]>> familyMap;
     private byte[][] viewConstants;
+    private IndexVerificationResultRepository verificationResultRepository;
```
swaroopak, line 146: How these two repositories differ from each other? Could the names be more explicit?

gjacoby126: Each repo controls insert/select operations for a particular table: the result table (PHOENIX_INDEX_TOOL_RESULT) that stores summary information about the run (counters, etc), and the output table (PHOENIX_INDEX_TOOL) that contains the row-level information about the failed rows. 

Please let me know if you have suggestions on clearer names. 

(The word "Repository", btw, is the name of the design pattern being used -- https://martinfowler.com/eaaCatalog/repository.html , from Martin Fowler's Patterns of Enterprise Architecture and Eric Evans' Domain Driven Design)

phoenix-core/src/it/java/org/apache/phoenix/end2end/index/IndexVerificationOutputRepositoryTest.java
```diff
@@ -0,0 +1,172 @@
...
+import java.util.Collections;
+import java.util.List;
+import java.util.Map;
+
+import static org.apache.phoenix.mapreduce.index.IndexVerificationOutputRepository.PHASE_AFTER_VALUE;
+import static org.apache.phoenix.mapreduce.index.IndexVerificationOutputRepository.PHASE_BEFORE_VALUE;
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertNotNull;
+
+public class IndexVerificationOutputRepositoryTest extends ParallelStatsDisabledIT {
```
kadirozde, line 47: Should not this be called IndexVerificationOutputRepositoryIT?

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/index/IndexVerificationOutputRow.java
```diff
@@ -0,0 +1,221 @@
...
+ * limitations under the License.
+ */
+package org.apache.phoenix.mapreduce.index;
+
+import org.apache.hadoop.hbase.util.Bytes;
+
+import java.util.Arrays;
+import java.util.Objects;
+
+public class IndexVerificationOutputRow {
```
kadirozde, line 25: Is this class is used only for IT? If so, it should not be under src/main, should it?

phoenix-core/src/it/java/org/apache/phoenix/end2end/index/IndexVerificationResultRepositoryTest.java
```diff
@@ -0,0 +1,131 @@
...
+
+import java.sql.Connection;
+import java.sql.DriverManager;
+import java.util.Collections;
+import java.util.Map;
+
+import static org.junit.Assert.assertArrayEquals;
+import static org.junit.Assert.assertEquals;
+
+public class IndexVerificationResultRepositoryTest extends ParallelStatsDisabledIT {
```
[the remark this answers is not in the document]
gjacoby126: Fair point. It's sort of between a true unit test and an IT, because it needs a minicluster but is exercising a particular class rather than something broader. I'll change the names.

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/index/IndexVerificationOutputRow.java
gjacoby126: The purpose of the patch is to expose the output row and repo classes so they can be used both for tests and also externally in tools that programmatically query and analyze the output table. So it needs to be in main.

kadirozde: It makes sense. Thx

phoenix-core/src/it/java/org/apache/phoenix/end2end/index/IndexVerificationResultRepositoryTest.java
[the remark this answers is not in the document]
gjacoby126: Fixed

phoenix-core/src/it/java/org/apache/phoenix/end2end/index/IndexVerificationOutputRepositoryTest.java
gjacoby126: Fixed

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/index/IndexVerificationOutputRepository.java
```diff
@@ -0,0 +1,285 @@
...
+    public static final byte[] A_VALUE_PREFIX_BYTES = Bytes.toBytes(" A:");
+    public static final int PREFIX_LENGTH = 3;
+    public static final int TOTAL_PREFIX_LENGTH = 6;
+    public static final byte[] PHASE_BEFORE_VALUE = Bytes.toBytes("BEFORE");
+    public static final byte[] PHASE_AFTER_VALUE = Bytes.toBytes("AFTER");
+
+    /**
+     * Only usable for the read path or for testing
+     */
+    public IndexVerificationOutputRepository(){
```
abhishek-chouhan, line 86: Do we need this constructor? I see that read (non static)methods make use of outputHTable which would not work with an instance created through this.

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/index/IndexVerificationOutputRepository.java
```diff
@@ -0,0 +1,285 @@
...
+        builder.setExpectedValue(result.getValue(OUTPUT_TABLE_COLUMN_FAMILY, EXPECTED_VALUE_BYTES));
+        builder.setActualValue(result.getValue(OUTPUT_TABLE_COLUMN_FAMILY, ACTUAL_VALUE_BYTES));
+        builder.setPhaseValue(result.getValue(OUTPUT_TABLE_COLUMN_FAMILY, VERIFICATION_PHASE_BYTES));
+        return builder.build();
+    }
+
+    public void close() throws IOException {
+        if (outputHTable != null) {
+            outputHTable.close();
+        }
```
abhishek-chouhan, line 256: Should we close indexHTable too?

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/index/IndexVerificationResultRepository.java
```diff
@@ -0,0 +1,209 @@
...
+                    isFirst = false;
+                }
+                verificationResult.update(cell);
+            }
+        }
+        return verificationResult;
+    }
+
+    public void close() throws IOException {
+        if (resultHTable != null) {
```
abhishek-chouhan, line 204: Check for indexHtable and close here too?

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/index/IndexVerificationOutputRow.java
```diff
@@ -0,0 +1,221 @@
...
+    @Override
+    public int hashCode(){
+        return Objects.hashCode(scanMaxTimestamp) ^ Objects.hashCode(indexTableName) ^
+            Arrays.hashCode(dataTableRowKey);
+    }
+
+    @Override
+    public String toString() {
+        StringBuilder sb = new StringBuilder();
+        sb.append("DataTableName: ").append(dataTableName).append(",");
```
swaroopak, line 133: Why not reuse the same strings from table values? like DTName, ITName. Would be easier to track and reuse.

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/index/IndexVerificationOutputRepository.java
gjacoby126: Good catch, thanks. Done.

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/index/IndexVerificationResultRepository.java
gjacoby126: Done.

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/index/IndexVerificationOutputRow.java
gjacoby126: Good idea. Done.

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/index/IndexVerificationOutputRepository.java
gjacoby126: The constructor's needed for a case in IndexTool where we're just creating the tables but not reading or writing the tables. I've added setters so that a developer can use the default constructor and still write after using the setters. Is that sufficient?

abhishek-chouhan:
LGTM +1

PR #754, opened 2020-04-03
PHOENIX-5817 Prepatch script is a psycho killer

stoty:


dbwong:
I was looking at what HBase was doing and they don't appear to use similar killer type code.  They also have onboarded on to yetus/docker for test framework/requirements so likely they don't need to.  
Long term we should likely prioritize https://issues.apache.org/jira/browse/PHOENIX-5032. 

In the short term, i'm not sure we have similar output to https://github.com/apache/hbase/blob/master/dev-support/flaky-tests/findHangingTests.py so we may rely on committers to tear down stuff if we misbehave.  @ChinmaySKulkarni  FYI 
Code changes look good though.

stoty:
merged

PR #759, opened 2020-04-14
Initial code push for TableLevel metrics.

vmeka2020:
https://issues.apache.org/jira/browse/PHOENIX-5837


phoenix-core/src/main/java/org/apache/phoenix/monitoring/PhoenixTableRegistry.java
```diff
@@ -0,0 +1,87 @@
...
+ */
+
+public class PhoenixTableRegistry  {
+    private static final Logger LOGGER = LoggerFactory.getLogger(PhoenixTableRegistry.class);
+    private static final boolean isTableLevelMetricsEnabled = QueryServicesOptions.withDefaults().isTableLevelMetricsEnabled();
+
+    private static ConcurrentMap<String,TableLevelMetricRegistry>tablePhoenixMapping;
+
+    static MetricRegistry metricRegistry;
+    static {
```
jpisaac, line 30: Need to rethink the use of singleton in the read/write path.

jpisaac:
@vmeka2020 Take a look at PhoenixRuntime.getRequestReadMetricInfo() which uses the ReadMetricQueue() to transform/aggregate the metrics on the table basis. You may be able to use that style to populate your MetricsSource for JMX, much like the QueryLogger uses it to write it to the logs.

Using the ReadMetricsQueue and MutationMetricQueue will allow us to reuse already instrumented code. It also properly takes care of failures, retries ...

Additional info here - https://phoenix.apache.org/metrics.html

FYI @dbwong @ChinmaySKulkarni

PR #760, opened 2020-04-14
PHOENIX-5840: IndexTool inline verification should not fail with -v O…

swaroopak:
…NLY option

### DIFF
--- a/phoenix-core/src/main/java/org/apache/phoenix/coprocessor/IndexToolVerificationResult.java
+++ b/phoenix-core/src/main/java/org/apache/phoenix/coprocessor/IndexToolVerificationResult.java
@@ -412,19 +412,10 @@ public class IndexToolVerificationResult {
         }
     }
 
-    public boolean isVerificationFailed(IndexTool.IndexVerifyType verifyType) {
-        if (verifyType == IndexTool.IndexVerifyType.BEFORE || verifyType == IndexTool.IndexVerifyType.NONE) {
-            return false;
-        } else if (verifyType == IndexTool.IndexVerifyType.ONLY) {
-            if (getBefore().getInvalidIndexRowCount() + getBefore().getMissingIndexRowCount() +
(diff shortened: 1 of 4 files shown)


phoenix-core/src/main/java/org/apache/phoenix/coprocessor/IndexToolVerificationResult.java
```diff
@@ -413,13 +413,8 @@ public void update(Cell cell) {
...
     public boolean isVerificationFailed(IndexTool.IndexVerifyType verifyType) {
-        if (verifyType == IndexTool.IndexVerifyType.BEFORE || verifyType == IndexTool.IndexVerifyType.NONE) {
+        if (verifyType == IndexTool.IndexVerifyType.BEFORE || verifyType == IndexTool.IndexVerifyType.NONE || verifyType == IndexTool.IndexVerifyType.ONLY) {
             return false;
-        } else if (verifyType == IndexTool.IndexVerifyType.ONLY) {
-            if (getBefore().getInvalidIndexRowCount() + getBefore().getMissingIndexRowCount() +
-                before.getBeyondMaxLookBackInvalidIndexRowCount() + before.getBeyondMaxLookBackMissingIndexRowCount() > 0) {
-                return true;
-            }
         } else if (verifyType == IndexTool.IndexVerifyType.BOTH || verifyType == IndexTool.IndexVerifyType.AFTER) {
```
kadirozde, line 418: I suggest the following instead of above
public boolean isVerificationFailed() {
  if (getAfter().getInvalidIndexRowCount() + getAfter().getMissingIndexRowCount() +
                after.getBeyondMaxLookBackInvalidIndexRowCount() + after.getBeyondMaxLookBackMissingIndexRowCount() > 0) {
                return true;
            }
  return false;
}

swaroopak: nice, however, we ll still need a check on "both" and "after"

PR #762, opened 2020-04-16
PHOENIX-5592 MapReduce job to asynchronously delete rows where the VIEW_TTL has expired

yanxinyi:
…EW_TTL has expired

### DIFF
--- a/phoenix-core/src/it/java/org/apache/phoenix/end2end/ViewTTLToolIT.java
+++ b/phoenix-core/src/it/java/org/apache/phoenix/end2end/ViewTTLToolIT.java
@@ -1,0 +1,610 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.phoenix.end2end;
+
+import org.apache.hadoop.conf.Configuration;
+import org.apache.hadoop.hbase.client.Scan;
+import org.apache.hadoop.hbase.client.Table;
+import org.apache.hadoop.hbase.client.Result;
+import org.apache.hadoop.hbase.client.ResultScanner;
+import org.apache.hadoop.hbase.filter.Filter;
+import org.apache.hadoop.hbase.filter.RowFilter;
+import org.apache.hadoop.hbase.filter.CompareFilter;
+import org.apache.hadoop.hbase.filter.RegexStringComparator;
+import org.apache.phoenix.mapreduce.ViewTTLTool;
+import org.apache.phoenix.mapreduce.util.PhoenixViewTtlUtil;
+import org.apache.phoenix.query.HBaseFactoryProvider;
+import org.junit.Test;
+
+import java.sql.SQLException;
+import java.sql.Connection;
+import java.sql.PreparedStatement;
+import java.sql.DriverManager;
+import java.sql.ResultSet;
+import java.sql.Statement;
+
+import static junit.framework.TestCase.assertTrue;
+import static org.junit.Assert.assertEquals;
+
+public class ViewTTLToolIT extends ParallelStatsDisabledIT {
+
+    private final int NUMBER_OF_UPSERT_ROWS = 10;
+    private final long VIEW_TTL_EXPIRE_IN_A_MILLISECOND = 1;
+    private final long VIEW_TTL_EXPIRE_IN_A_DAY = 1000 * 60 * 60 * 24;
+
+    private void alterViewTtl(Connection conn, String viewName, long view_ttl_value)
+            throws SQLException {
+        conn.createStatement().execute(
+                String.format("ALTER VIEW %s SET VIEW_TTL= %d", viewName, view_ttl_value));
+    }
+
+    private void createViewAndUpsertData(Connection conn, String tableName, String viewName,
+                                         long view_ttl_value) throws SQLException {
+        String ddl = "CREATE VIEW " + viewName +
+                " (PK1 BIGINT NOT NULL, PK2 BIGINT NOT NULL CONSTRAINT PKVIEW PRIMARY KEY(PK1,PK2)) " +
+                "AS SELECT * FROM " + tableName + " VIEW_TTL=" + view_ttl_value;
+
+        conn.createStatement().execute(ddl);
+
+        for (int i = 0; i < NUMBER_OF_UPSERT_ROWS; i++) {
+            PreparedStatement stmt = conn.prepareStatement(
+                    "UPSERT INTO " + viewName + " (ID, PK1, PK2) VALUES(?,?,?)");
+            stmt.setString(1, generateUniqueName());
+            stmt.setInt(2, i);
+            stmt.setInt(3, i);
+            stmt.execute();
+            stmt.close();
+        }
+        conn.commit();
+    }
+
+    private void createViewOnGlobalViewAndUpsertData(Connection conn, String tableName, String viewName,
+                                                     long view_ttl_value) throws SQLException {
+        String ddl = "CREATE VIEW " + viewName + " AS SELECT * FROM " + tableName + " VIEW_TTL= " + view_ttl_value;
+        conn.createStatement().execute(ddl);
+
+        for (int i = 0; i < NUMBER_OF_UPSERT_ROWS; i++) {
+            PreparedStatement stmt = conn.prepareStatement(
+                    "UPSERT INTO " + viewName + " (ID, PK1, PK2, NUM) VALUES(?,?,?,?)");
+            stmt.setString(1, generateUniqueName());
+            stmt.setInt(2, i);
+            stmt.setInt(3, i);
+            stmt.setInt(4, i);
+            stmt.execute();
+            stmt.close();
+        }
+        conn.commit();
+    }
+
+    private void verifyGlobalTableNumberOfRows(String tableName, int expectedRows) throws Exception {
+        try (Table table  = HBaseFactoryProvider.getHConnectionFactory().createConnection(config).getTable(tableName)) {
+            assertEquals(expectedRows, getRowCount(table, new Scan()));
+        }
+    }
+
+    private void verifyIndexTableNumberOfRowsForATenant(String tableName, String regrex, int expectedRows)
+            throws Exception {
+        try (Table table  = HBaseFactoryProvider.getHConnectionFactory().createConnection(config).getTable(tableName)) {
+            Filter filter = new RowFilter(CompareFilter.CompareOp.EQUAL, new RegexStringComparator(regrex));
+            Scan scan = new Scan();
+            scan.setFilter(filter);
+            assertEquals(expectedRows, getRowCount(table,scan));
+        } catch (Exception e) {
+            throw e;
+        }
+    }
+
+    private void verifyMultiTenantTableNumberOfRows(String tableName, String tenantId, int expectedRows,
+                                                    Connection conn) throws Exception {
+        String query = "SELECT COUNT(*) FROM " + tableName;
+        if (tenantId != null) {
+            query = query + " WHERE TENANT_ID = '" + tenantId + "'";
+        }
+        try (Statement stm = conn.createStatement()) {
+
+            ResultSet rs = stm.executeQuery(query);
+            assertTrue(rs.next());
+            assertEquals(expectedRows, rs.getInt(1));
+        }
+    }
+
+    private long getRowCount(Table table, Scan scan) throws Exception {
+        ResultScanner scanner = table.getScanner(scan);
+        int count = 0;
+        for (Result dummy : scanner) {
+            count++;
+        }
+        scanner.close();
+        return count;
+    }
+
+    private void createMultiTenantTable(Connection conn, String tableName) throws Exception {
+        String ddl = "CREATE TABLE " + tableName +
+                " (TENANT_ID CHAR(10) NOT NULL, ID CHAR(10) NOT NULL, NUM BIGINT CONSTRAINT " +
+                "PK PRIMARY KEY (TENANT_ID, ID)) MULTI_TENANT=true, COLUMN_ENCODED_BYTES = 0";
+
+        conn.createStatement().execute(ddl);
+    }
+
+    private void createGlobalViewWithPk(Connection conn, String tableName, String globalViewName) throws Exception {
+        String ddl = "CREATE VIEW " + globalViewName +
+                " (PK1 BIGINT NOT NULL, PK2 BIGINT NOT NULL CONSTRAINT PKVIEW PRIMARY KEY(PK1,PK2)) AS SELECT * FROM "
+                + tableName;
+        conn.createStatement().execute(ddl);
+    }
+
+    @Test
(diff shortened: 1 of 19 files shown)


phoenix-core/src/main/java/org/apache/phoenix/mapreduce/PhoenixMultiViewInputFormat.java
```diff
@@ -0,0 +1,85 @@
...
+            Class<?> defaultViewTtlDeletionSplitStrategyClazz = DefaultMultiViewSplitStrategy.class;
+            if (configuration.get(PhoenixConfigurationUtil.MAPREDUCE_VIEW_TTL_SPLIT_STRATEGY_CLAZZ) != null) {
+                defaultViewTtlDeletionSplitStrategyClazz = Class.forName(
+                        configuration.get(PhoenixConfigurationUtil.MAPREDUCE_VIEW_TTL_SPLIT_STRATEGY_CLAZZ));
+            }
+            MultiViewSplitStrategy multiViewSplitStrategy =
+                    (MultiViewSplitStrategy) defaultViewTtlDeletionSplitStrategyClazz.newInstance();
+            listOfInputSplit = multiViewSplitStrategy.generateSplits(viewsWithTTL, configuration);
+        } catch (Exception e) {
+            LOGGER.debug(e.getStackTrace().toString());
```
jpisaac, line 68: Don't swallow the exception and log with error.

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/PhoenixMultiViewReader.java
```diff
@@ -0,0 +1,74 @@
...
+    Iterator<ViewInfoWritable> it;
+
+    public PhoenixMultiViewReader(final Class<T> inputClass, final Configuration configuration) {
+        this.configuration = configuration;
+        this.inputClass = inputClass;
+    }
+
+    @Override public void initialize(InputSplit split, TaskAttemptContext context)
+            throws IOException, InterruptedException {
+        final PhoenixMultiViewInputSplit pSplit = (PhoenixMultiViewInputSplit)split;
```
jpisaac, line 46: Maybe a good idea to check if "split" is assignable to "PhoenixMultiViewInputSplit" and then cast, else return an exception.

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/ViewTTLDeleteJobMapper.java
```diff
@@ -0,0 +1,120 @@
...
+                LOGGER.info("Using customized tracker class : " +
+                        config.get(PhoenixConfigurationUtil.MAPREDUCE_VIEW_TTL_MAPPER_TRACKER_CLAZZ));
+                defaultViewDeletionTrackerClass = Class.forName(
+                        config.get(PhoenixConfigurationUtil.MAPREDUCE_VIEW_TTL_MAPPER_TRACKER_CLAZZ));
+            } else {
+                LOGGER.info("Using default tracker class ");
+            }
+            this.multiViewJobStatusTracker = (MultiViewJobStatusTracker) defaultViewDeletionTrackerClass.newInstance();
+        } catch (Exception e) {
+            LOGGER.info("exception " + e.getMessage());
```
jpisaac, line 55: throw an appropriate exception and log with error.

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/ViewTTLDeleteJobMapper.java
```diff
@@ -0,0 +1,120 @@
...
+    @Override
+    protected void map(NullWritable key, ViewInfoTracker value,
+                       Context context) {
+        final Configuration config = context.getConfiguration();
+
+        if (this.multiViewJobStatusTracker == null) {
+            initMultiViewJobStatusTracker(config);
+        }
+
+        LOGGER.debug(String.format("Deleting from view %s, TenantID %s, and TTL value: %d",
```
jpisaac, line 69: Should we log at debug level only if debug level enabled?

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/ViewTTLDeleteJobMapper.java
```diff
@@ -0,0 +1,120 @@
...
+            this.multiViewJobStatusTracker = (MultiViewJobStatusTracker) defaultViewDeletionTrackerClass.newInstance();
+        } catch (Exception e) {
+            LOGGER.info("exception " + e.getMessage());
+            LOGGER.info("stack trace" + e.getStackTrace().toString());
+        }
+    }
+
+    @Override
+    protected void map(NullWritable key, ViewInfoTracker value,
+                       Context context) {
```
jpisaac, line 62: Check for value not null?

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/PhoenixMultiViewInputFormat.java
yanxinyi: in this block, it might occur `ClassNotFoundException`, `InstantiationException`, and `IllegalAccessException`; however, the base class `InputFormat` doesn't throw a list above. It has 2 options here.
1. wrapper exception to be either `IOException` or `InterruptedException` and throw it.
2. keep current the implementation, which logs exception message.

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/util/DefaultMultiViewSplitStrategy.java
```diff
@@ -0,0 +1,55 @@
...
+import com.google.common.collect.Lists;
+import org.apache.hadoop.conf.Configuration;
+import org.apache.hadoop.mapreduce.InputSplit;
+import org.apache.phoenix.mapreduce.PhoenixMultiViewInputSplit;
+
+import java.util.List;
+
+public class DefaultMultiViewSplitStrategy implements MultiViewSplitStrategy {
+
+    public List<InputSplit> generateSplits(List<ViewInfoWritable> viewsWithTTL, Configuration configuration) {
```
jpisaac, line 29: This class can be made generic by removing all the viewWithTTL variable naming. Thus it can be used by other mappers when they want SizeBased split strategy.

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/util/MultiViewSplitStrategy.java
```diff
@@ -0,0 +1,27 @@
...
+ */
+package org.apache.phoenix.mapreduce.util;
+
+import org.apache.hadoop.conf.Configuration;
+import org.apache.hadoop.mapreduce.InputSplit;
+
+import java.util.List;
+
+public interface MultiViewSplitStrategy {
+    List<InputSplit> generateSplits(List<ViewInfoWritable> viewsWithTTL, Configuration configuration);
```
jpisaac, line 26: nit: remove references to TTL so that it can be generic

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/util/PhoenixViewTtlUtil.java
```diff
@@ -0,0 +1,92 @@
...
+import static org.apache.phoenix.jdbc.PhoenixDatabaseMetaData.SYSTEM_CATALOG_NAME;
+import static org.apache.phoenix.jdbc.PhoenixDatabaseMetaData.TABLE_TYPE;
+import static org.apache.phoenix.jdbc.PhoenixDatabaseMetaData.VIEW_TTL;
+import static org.apache.phoenix.jdbc.PhoenixDatabaseMetaData.VIEW_TTL_NOT_DEFINED;
+import static org.apache.phoenix.jdbc.PhoenixDatabaseMetaData.VIEW_TYPE;
+import static org.apache.phoenix.jdbc.PhoenixDatabaseMetaData.TENANT_ID;
+
+public class PhoenixViewTtlUtil {
+    public static final String SELECT_ALL_VIEW_METADATA_FROM_SYSCAT_QUERY =
+            "SELECT TENANT_ID, TABLE_SCHEM, TABLE_NAME, VIEW_TTL FROM " +
```
jpisaac, line 43: Shouldn't these be queries be in QUERY_MORE style?

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/util/DefaultPhoenixMultiViewListProvider.java
```diff
@@ -0,0 +1,108 @@
...
+import java.util.List;
+import java.util.Properties;
+
+
+public class DefaultPhoenixMultiViewListProvider implements PhoenixMultiViewListProvider {
+    private static final Logger LOGGER = LoggerFactory.getLogger(DefaultPhoenixMultiViewListProvider.class);
+
+    public List<ViewInfoWritable> getPhoenixMultiViewList(Configuration configuration) {
+        List<ViewInfoWritable> viewInfoWritables = new ArrayList<>();
+        String query = getFetchViewQuery(configuration);
```
jpisaac, line 42: Shouldn't the query handling be QUERY_MORE style?

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/ViewTTLDeleteJobMapper.java
yanxinyi: without the config of debug-level enabled, it will omit the logging for the debug level logging.

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/ViewTTLDeleteJobMapper.java
yanxinyi: the context will never be null, but I will put it wihtin the try-catch block

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/util/DefaultMultiViewSplitStrategy.java
```diff
@@ -26,28 +26,28 @@
 
 public class DefaultMultiViewSplitStrategy implements MultiViewSplitStrategy {
 
-    public List<InputSplit> generateSplits(List<ViewInfoWritable> viewsWithTTL, Configuration configuration) {
-        int numViewsInSplit = PhoenixConfigurationUtil.getViewTTLDeleteMapperSplitSize(configuration);
-        int numberOfMappers = viewsWithTTL.size() / numViewsInSplit;
-        if (Math.ceil(viewsWithTTL.size() % numViewsInSplit) > 0) {
+    public List<InputSplit> generateSplits(List<ViewInfoWritable> views, Configuration configuration) {
+        int numViewsInSplit = PhoenixConfigurationUtil.getMultiViewSplitSize(configuration);
```
jpisaac, line 30: Can u add check that numViewsInSplit is non-zero.?

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/util/ViewInfoWritable.java
```diff
@@ -24,6 +24,25 @@
 import java.io.IOException;
 
 public interface ViewInfoWritable extends Writable {
+    public enum ViewInfoJobState {
```
jpisaac, line 27: Good job on reafctoring it out to be more generic.

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/util/ViewInfoTracker.java
```diff
@@ -24,24 +24,6 @@
 import java.io.IOException;
 
 public class ViewInfoTracker implements ViewInfoWritable {
```
jpisaac, line 26: nit: Good job on refactoring out it to an interface. Now that you have refactored it out it can be called ViewTTLInfoWritable? Since it contains TTL related info

jpisaac:
@Xinyi Another suggestion I have is whether u want to split this into 2 PRs? 
1. A general framework to use multiple input views in a MR jobs.
2. MR job using the above framework to delete VIEW TTL expired rows.

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/ViewTTLDeleteJobMapper.java
```diff
@@ -0,0 +1,128 @@
...
+
+        try {
+            this.multiViewJobStatusTracker.updateJobStatus(view, 0,
+                    ViewInfoJobState.PREP.getValue(), config, 0);
+
+            long startTime = System.currentTimeMillis();
+            this.multiViewJobStatusTracker.updateJobStatus(view, 0,
+                    ViewInfoJobState.RUNNING.getValue(), config, 0 );
+            connection.setAutoCommit(true);
+            int numberOfDeletedRows = connection.createStatement().executeUpdate(deleteIfExpiredStatement);
```
ChinmaySKulkarni, line 108: create statement inside try-with-resources. We have to very careful not to leak any connections/statements, etc. from our own framework. Can you please ensure that throughout? I may have missed some.

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/util/DefaultPhoenixMultiViewListProvider.java
```diff
@@ -0,0 +1,106 @@
...
+        int limit = PhoenixConfigurationUtil.getMultiViewQueryMoreSplitSize(configuration);
+        try (PhoenixConnection connection = (PhoenixConnection)
+                ConnectionUtil.getInputConnection(configuration, new Properties())){
+            TableName catalogOrChildTableName = ViewUtil.getSystemTableForChildLinks(0, configuration);
+            Table catalogOrChildTable = connection.getQueryServices()
+                    .getTable(SchemaUtil.getPhysicalName(catalogOrChildTableName.toBytes(),
+                            connection.getQueryServices().getProps())
+                            .getName());
+            do {
+                ResultSet viewRs = connection.createStatement().executeQuery(query);
```
ChinmaySKulkarni, line 54: create stmt inside try-with-resources

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/PhoenixMultiViewInputFormat.java
ChinmaySKulkarni: I think we should wrap as an IOException and rethrow

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/ViewTTLDeleteJobMapper.java
```diff
@@ -0,0 +1,120 @@
...
+                        config.get(PhoenixConfigurationUtil.MAPREDUCE_VIEW_TTL_MAPPER_TRACKER_CLAZZ));
+                defaultViewDeletionTrackerClass = Class.forName(
+                        config.get(PhoenixConfigurationUtil.MAPREDUCE_VIEW_TTL_MAPPER_TRACKER_CLAZZ));
+            } else {
+                LOGGER.info("Using default tracker class ");
+            }
+            this.multiViewJobStatusTracker = (MultiViewJobStatusTracker) defaultViewDeletionTrackerClass.newInstance();
+        } catch (Exception e) {
+            LOGGER.info("exception " + e.getMessage());
+            LOGGER.info("stack trace" + e.getStackTrace().toString());
```
ChinmaySKulkarni: Why are we swallowing the exception here? Can it be rethrown? Also, in case we DO want to swallow the exception and just log it, this should be `error` level.

phoenix-core/src/it/java/org/apache/phoenix/end2end/ViewTTLToolIT.java
```diff
@@ -0,0 +1,612 @@
...
+import java.sql.Statement;
+
+import static junit.framework.TestCase.assertTrue;
+import static org.junit.Assert.assertEquals;
+
+public class ViewTTLToolIT extends ParallelStatsDisabledIT {
+
+    private final int ONE = 1;
+    private final int ZERO = 0;
+    private final int NUMBER_OF_UPSERT_ROWS = 202;
```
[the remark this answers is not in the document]
yanxinyi: since the default split size is 100, it will have an uneven split in this case.

stoty:
Already merged.

PR #763, opened 2020-04-16
PHOENIX-5839 CompatUtil#setStopRow semantics problem

stoty:
replace setStopRow with setSingleRow which has consistent semantics


richardantal:
LGTM

PR #764, opened 2020-04-16
PHOENIX-5808 Improve shaded artifact naming convetions

stoty:
provide unversioned symlinks in the assembly for -server and -client
drop the deprecated phoenix-<version>-client/server symlinks/jars
update the python lib to look for the unversioned client jar

### DIFF
--- a/phoenix-assembly/pom.xml
+++ b/phoenix-assembly/pom.xml
@@ -42,7 +42,27 @@
         <groupId>org.codehaus.mojo</groupId>
         <executions>
           <execution>
-            <id>Symlink to deprecated client jar name</id>
+            <id>client without classifier</id>
+            <phase>compile</phase>
+            <goals>
+              <goal>exec</goal>
+            </goals>
+            <configuration>
+              <executable>ln</executable>
+              <workingDirectory></workingDirectory>
+              <arguments>
+                <argument>-fnsv</argument>
+                <argument>
+                  phoenix-client-${project.version}-${hbase.profile.string}.jar
+                </argument>
+                <argument>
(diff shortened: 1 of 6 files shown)


phoenix-assembly/pom.xml
```diff
@@ -42,7 +42,27 @@
...
           <execution>
-            <id>Symlink to deprecated client jar name</id>
+            <id>client without classifier</id>
+            <phase>compile</phase>
+            <goals>
+              <goal>exec</goal>
+            </goals>
+            <configuration>
+              <executable>ln</executable>
+              <workingDirectory></workingDirectory>
```
joshelser, line 52: Should be `${project.basedir}/../phoenix-client/target`? I think this leaves it in `phoenix-assembly` and is then not reflected in the binary tarball.

```
$ ll phoenix-5.1.0-SNAPSHOT-hbase-2.2
...
-rw-r--r--   1 jelser  staff   133M Apr 23 11:47 phoenix-client-5.1.0-SNAPSHOT-hbase-2.2.jar
-rw-r--r--   1 jelser  staff    18K Apr 23 11:46 phoenix-client-5.1.0-SNAPSHOT.jar
lrwxr-xr-x   1 jelser  staff    43B Apr 23 11:47 phoenix-client.jar -> phoenix-client-5.1.0-SNAPSHOT-hbase-2.2.jar
```

stoty: Thanks for the catch  @joshelser 
Fixed.
Also added explicit dependency on phoenix-client and phoenix-server so that partial builds work.

PR #766, opened 2020-04-17
Initial checkin for PHOENIX-5821

jpisaac:


phoenix-core/src/it/java/org/apache/phoenix/end2end/ViewTTLIT.java
```diff
@@ -391,7 +402,1390 @@ public void testViewTTLWithAlterView() throws Exception {
...
+                getUrl() + ';' + TENANT_ID_ATTRIB + '=' + schemaBuilder.getDataOptions()
+                        .getTenantId();
+
+        try (Connection deleteConnection = DriverManager.getConnection(tenantConnectUrl, props);
+                final Statement statement = deleteConnection.createStatement()) {
+            deleteConnection.setAutoCommit(true);
+
+            final String deleteIfExpiredStatement = String.format("delete from %s where %s",
+                    viewName,
+                    String.format("((TO_NUMBER(CURRENT_DATE()) - TO_NUMBER(phoenix_row_timestamp())) < %d)" , viewTTL));
```
yanxinyi, line 1761: what's diff btw `CURRENT_DATE()` and `NOW()`?

yanxinyi: I went to check on the code, it seems that `CURRENT_DATE` is looking for the client-side date instead of the server-side. Should we use the consistency of server-side time because I believe phoenix row ts is server-side

phoenix-core/src/main/java/org/apache/phoenix/util/ViewUtil.java
```diff
@@ -629,4 +631,37 @@ public static void addTagsToPutsForViewAlteredProperties(List<Mutation> tableMet
...
+
+    public static ParseNode getViewWhereWithViewTTL(StatementContext context, PTable table)
+            throws SQLException {
+        ParseNode viewWhere = null;
+
+        Long nowTS = context.getConnection().getSCN() == null ?
+                EnvironmentEdgeManager.currentTimeMillis() : context.getConnection().getSCN();
+        String viewStatement = null;
+        String viewTTLFilterClause = String.format(
+                "((%d - TO_NUMBER(phoenix_row_timestamp())) < %d)", nowTS, table.getViewTTL());
```
yanxinyi, line 643: same here. If phoenix row ts is getting server-side ts, it's better to compare with server-side ts instead of client-side

jpisaac: Will change here to use the CURRENT_DATE.

phoenix-core/src/test/java/org/apache/phoenix/query/PhoenixTestBuilder.java
```diff
@@ -212,66 +254,142 @@ private static String getFQColumnsAsString(List<String> columns, List<String> fa
...
+        }
 
-        public static List<String> TABLE_COLUMN_FAMILIES = asList(null, null, null);
-        public static List<String> GLOBAL_VIEW_COLUMN_FAMILIES = asList(null, null, null);
-        public static List<String> TENANT_VIEW_COLUMN_FAMILIES = asList(null, null, null);
+        // Read batch of rows
+        public void readRows() throws SQLException {
+            dataTable.clear();
+            String sql = getDML();
+            Connection connection = getConnection();
```
ChinmaySKulkarni, line 289: Add try-with-resources for `connection`

phoenix-core/src/test/java/org/apache/phoenix/query/PhoenixTestBuilder.java
```diff
@@ -305,16 +422,83 @@ public void upsertRow(int rowIndex) throws SQLException {
...
+            for (int i = 0; i < upsertColumns.size(); i++) {
+                buf.append("?,");
+                if (rowKeys.contains(upsertColumns.get(i))) {
+                    rowKeyPositions.add(i);
+                }
+            }
+            buf.setCharAt(buf.length() - 1, ')');
+            LOGGER.info(buf.toString());
+
+            Connection connection = getConnection();
```
ChinmaySKulkarni, line 461: Add try-with-resources for `connection`

phoenix-core/src/test/java/org/apache/phoenix/query/PhoenixTestBuilder.java
```diff
@@ -294,7 +412,6 @@ public void upsertRow(int rowIndex) throws SQLException {
                 buf.append("?,");
             }
             buf.setCharAt(buf.length() - 1, ')');
-
             LOGGER.info(buf.toString());
 
             Connection connection = getConnection();
```
ChinmaySKulkarni, line 417: Add try-with-resources for `connection`

phoenix-core/src/main/java/org/apache/phoenix/util/ViewUtil.java
```diff
@@ -629,4 +631,37 @@ public static void addTagsToPutsForViewAlteredProperties(List<Mutation> tableMet
 
         }
     }
+
+    public static ParseNode getViewWhereWithViewTTL(StatementContext context, PTable table)
```
ChinmaySKulkarni, line 635: Let's add a unit test for this new method. It's fine for ViewUtilTest to just contain a test for this new method for now.

phoenix-core/src/main/java/org/apache/phoenix/compile/QueryCompiler.java
```diff
@@ -653,9 +653,12 @@ protected QueryPlan compileSingleFlatQuery(
         PTable table = tableRef.getTable();
 
         ParseNode viewWhere = null;
-        if (table.getViewStatement() != null) {
+        if (table.getViewTTL() > 0) {
```
ChinmaySKulkarni, line 656: For clarity, can you please add a comment here where we say that viewTTL is expected to be non-zero only for views or view indexes?

phoenix-core/src/main/java/org/apache/phoenix/util/ViewUtil.java
ChinmaySKulkarni: +1, though maybe use `CURRENT_TIME` or `NOW` instead since that indicates datetime better than `CURRENT_DATE` (though all 3 may be doing the same thing).

phoenix-core/src/test/java/org/apache/phoenix/query/PhoenixTestBuilder.java
```diff
@@ -212,66 +254,142 @@ private static String getFQColumnsAsString(List<String> columns, List<String> fa
...
+    }
 
-        public static List<String> GLOBAL_VIEW_INDEX_COLUMNS = asList("COL4");
-        public static List<String> GLOBAL_VIEW_INCLUDE_COLUMNS = asList("COL6");
+    // An implementation of the DataReader.
+    public static class BasicDataReader extends AbstractDataReader {
 
-        public static List<String> TENANT_VIEW_INDEX_COLUMNS = asList("COL9");
-        public static List<String> TENANT_VIEW_INCLUDE_COLUMNS = asList("COL7");
+        Connection connection;
```
ChinmaySKulkarni, line 332: I might be missing something, but it looks like we create a local variable `Connection connection = ... ` everywhere. Is this class variable required?

phoenix-core/src/it/java/org/apache/phoenix/end2end/ViewTTLIT.java
```diff
@@ -391,7 +402,1390 @@ public void testViewTTLWithAlterView() throws Exception {
...
         assertSyscatHaveViewTTLRelatedColumns(tenantId, schemaName, tenantViewName, PTableType.VIEW.getSerializedValue(), 1000);
         assertSyscatHaveViewTTLRelatedColumns(tenantId, schemaName, indexOnTenantViewName, PTableType.INDEX.getSerializedValue(), 1000);
+    }
+
+    @Test
+    public void testViewTTLForLevelTwoViewWithNoIndexes() throws Exception {
+        long startTime = System.currentTimeMillis();
+
+        // Define the test schema.
+        // 1. Table with default columns => (ORG_ID, KP, COL1, COL2, COL3), PK => (ORG_ID, KP)
```
ChinmaySKulkarni, line 412: what do you mean by "default" columns?

phoenix-core/src/it/java/org/apache/phoenix/end2end/ViewTTLIT.java
ChinmaySKulkarni: It looks like `CURRENT_TIME`, `CURRENT_DATE` and `NOW` are all the same and all hit the server with "alwaysUpdateServer=true" and update the cache and get the timestamp from that.

phoenix-core/src/it/java/org/apache/phoenix/end2end/ViewTTLIT.java
```diff
@@ -391,7 +402,1390 @@ public void testViewTTLWithAlterView() throws Exception {
...
+            deleteConnection.setAutoCommit(true);
+
+            final String deleteIfExpiredStatement = String.format("delete from %s where %s",
+                    viewName,
+                    String.format("((TO_NUMBER(CURRENT_DATE()) - TO_NUMBER(phoenix_row_timestamp())) < %d)" , viewTTL));
+            Preconditions.checkNotNull(deleteIfExpiredStatement);
+
+            final PhoenixStatement pstmt = statement.unwrap(PhoenixStatement.class);
+            int deleteCount = pstmt.executeUpdate(deleteIfExpiredStatement);
+            /*
```
ChinmaySKulkarni, line 1766: Can we remove this commented code?

phoenix-core/src/it/java/org/apache/phoenix/end2end/ViewTTLIT.java
```diff
@@ -391,7 +402,1390 @@ public void testViewTTLWithAlterView() throws Exception {
...
+                            String
+                                    col6 =
+                                    String.format("f%05d", rowIndex + rnd.nextInt(MAX_ROWS));
+                            String
+                                    col7 =
+                                    String.format("g%05d", rowIndex + rnd.nextInt(MAX_ROWS));
+                            String
+                                    col8 =
+                                    String.format("h%05d", rowIndex + rnd.nextInt(MAX_ROWS));
+                            String
```
ChinmaySKulkarni, line 1240: This formatting still seems really odd to me. Are you sure you are using the IDE preferences setting in `<phoenix_project_home>/dev/`? See https://phoenix.apache.org/contributing.html

phoenix-core/src/it/java/org/apache/phoenix/end2end/ViewTTLIT.java
```diff
@@ -391,7 +402,1390 @@ public void testViewTTLWithAlterView() throws Exception {
...
+                            .withGlobalViewIndexOptions(globalViewIndexOptions)
+                            .withTenantViewOptions(tenantViewOptions)
+                            .withTenantViewIndexOptions(tenantViewIndexOptions)
+                            .withOtherOptions(testCaseWhenAllCFMatchAndAllDefault)
+                            .buildWithNewTenant();
+
+                    // Define the test data.
+                    DataSupplier dataSupplier = new DataSupplier() {
+
+                        @Override public List<Object> getValues(int rowIndex) {
```
ChinmaySKulkarni, line 1212: I see a bunch of anonymous classes overriding `DataSupplier` and similar classes. Can we extract them out and re-use some code?

phoenix-core/src/test/java/org/apache/phoenix/query/PhoenixTestBuilder.java
jpisaac: Since the connection object is set by user of this class (typically the tests cases), expect the caller to close the connection. Will add doc to that effect in the setConnection() setter method.

phoenix-core/src/test/java/org/apache/phoenix/query/PhoenixTestBuilder.java
jpisaac: Same here

phoenix-core/src/test/java/org/apache/phoenix/query/PhoenixTestBuilder.java
jpisaac: The test cases create the connection and call the setter method (setConnection)

phoenix-core/src/it/java/org/apache/phoenix/end2end/ViewTTLIT.java
jpisaac: Default columns as defined in PhoenixTestBuilder.DDLDefaults, will remove the "default" reference from the doc lines, as it is confusing folks that are reading it.

phoenix-core/src/it/java/org/apache/phoenix/end2end/ViewTTLIT.java
jpisaac: Would like to keep the anonymous data supplier classes for readability and context. Many of the classes have different sets of columns. Would agree there are some duplicates. But let me know if you feel strongly.

phoenix-core/src/it/java/org/apache/phoenix/end2end/ViewTTLIT.java
jpisaac: Not sure what happened, was using the Phoenix template, but have reformatted it. Seems better!!

phoenix-core/src/test/java/org/apache/phoenix/query/PhoenixTestBuilder.java
jpisaac: Same here

phoenix-core/src/test/java/org/apache/phoenix/query/PhoenixTestBuilder.java
ChinmaySKulkarni: I see, missed that. Thanks for the clarification

phoenix-core/src/test/java/org/apache/phoenix/query/PhoenixTestBuilder.java
ChinmaySKulkarni: Got it. I misread as a call to DriverManager.getConnection() instead of getConnection.

PR #768, opened 2020-04-22
PHOENIX-5580 Wrong values seen when updating a view for a table that …

sguggilam:
…has an index

### DIFF
--- a/phoenix-core/src/main/java/org/apache/phoenix/coprocessor/generated/ServerCachingProtos.java
+++ b/phoenix-core/src/main/java/org/apache/phoenix/coprocessor/generated/ServerCachingProtos.java
@@ -2177,6 +2177,21 @@ public final class ServerCachingProtos {
      * <code>optional int32 indexDataColumnCount = 23 [default = -1];</code>
      */
     int getIndexDataColumnCount();
+
+    // optional string parentTableType = 24;
+    /**
+     * <code>optional string parentTableType = 24;</code>
+     */
+    boolean hasParentTableType();
+    /**
+     * <code>optional string parentTableType = 24;</code>
+     */
+    java.lang.String getParentTableType();
+    /**
+     * <code>optional string parentTableType = 24;</code>
+     */
+    com.google.protobuf.ByteString
+        getParentTableTypeBytes();
   }
   /**
    * Protobuf type {@code IndexMaintainer}
@@ -2378,6 +2393,11 @@ public final class ServerCachingProtos {
             case 184: {
               bitField0_ |= 0x00020000;
               indexDataColumnCount_ = input.readInt32();
+              break;
+            }
+            case 194: {
+              bitField0_ |= 0x00040000;
+              parentTableType_ = input.readBytes();
               break;
             }
           }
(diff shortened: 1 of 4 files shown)


phoenix-core/src/main/java/org/apache/phoenix/coprocessor/generated/ServerCachingProtos.java
```diff
@@ -2177,6 +2177,21 @@ public Builder setColumnNameBytes(
      * <code>optional int32 indexDataColumnCount = 23 [default = -1];</code>
      */
     int getIndexDataColumnCount();
+
+    // optional string parentTableType = 24;
+    /**
+     * <code>optional string parentTableType = 24;</code>
```
swaroopak, line 2183: nit: Comments are not in sync with code.

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/generated/ServerCachingProtos.java
```diff
@@ -2896,6 +2916,49 @@ public int getIndexDataColumnCount() {
       return indexDataColumnCount_;
     }
 
+    // optional string parentTableType = 24;
+    public static final int PARENTTABLETYPE_FIELD_NUMBER = 24;
```
swaroopak, line 2920: nit: PARENT_TABLE_TYPE_FIELD_NUMBER

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/generated/ServerCachingProtos.java
```diff
@@ -5759,6 +5850,80 @@ public Builder clearIndexDataColumnCount() {
...
+          return s;
+        } else {
+          return (java.lang.String) ref;
+        }
+      }
+      /**
+       * <code>optional string parentTableType = 24;</code>
+       */
+      public com.google.protobuf.ByteString
+          getParentTableTypeBytes() {
```
swaroopak, line 5879: this can fit in one line, any reason to break to multiple? applies to other places as well

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/generated/ServerCachingProtos.java
```diff
@@ -5759,6 +5850,80 @@ public Builder clearIndexDataColumnCount() {
...
+          parentTableType_ = b;
+          return b;
+        } else {
+          return (com.google.protobuf.ByteString) ref;
+        }
+      }
+      /**
+       * <code>optional string parentTableType = 24;</code>
+       */
+      public Builder setParentTableType(
```
swaroopak, line 5894: Indent looks little off in this function

phoenix-core/src/main/java/org/apache/phoenix/index/IndexMaintainer.java
```diff
@@ -1887,6 +1899,11 @@ public boolean isImmutableRows() {
         return immutableRows;
     }
     
+    public boolean isTableIndex() {
+        if (parentTableType == null) return false;
```
swaroopak, line 1903: braces in `if` block

phoenix-core/src/main/java/org/apache/phoenix/index/IndexMaintainer.java
```diff
@@ -1887,6 +1899,11 @@ public boolean isImmutableRows() {
         return immutableRows;
     }
     
+    public boolean isTableIndex() {
```
swaroopak, line 1902: I haven't come across any think as table index and might confuse while reading as global index comes with its own table. How about isViewIndex() with a negation?

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/generated/ServerCachingProtos.java
sguggilam: This is auto generated file

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/generated/ServerCachingProtos.java
sguggilam: This is auto generated file

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/generated/ServerCachingProtos.java
sguggilam: This is auto generated file

phoenix-core/src/main/java/org/apache/phoenix/index/IndexMaintainer.java
sguggilam: Done

phoenix-core/src/main/java/org/apache/phoenix/index/IndexMaintainer.java
sguggilam: Makes sense, made the change and pushed in latest commit

phoenix-core/src/main/java/org/apache/phoenix/index/IndexMaintainer.java
```diff
@@ -1887,6 +1899,13 @@ public boolean isImmutableRows() {
         return immutableRows;
     }
     
+    public boolean isViewIndex() {
+        if (parentTableType == null) {
```
gjacoby126, line 1903: @sguggilam - In the upgrade scenario where the clients don't have the new proto and hence parentTableType isn't defined, will this still work OK to have isViewIndex incorrectly return true?

sguggilam: Yes @gjacoby126 , the logic that is added in the change makes sure that the behavior is unchanged for the clients that doesn't pass the parentTableType field (the logic uses a OR condition with !isViewIndex() )

stoty:
Already merged.

PR #769, opened 2020-04-22
PHOENIX-5867 Add support for property overrides in IndexTool

abhishek-chouhan:


phoenix-core/src/test/java/org/apache/phoenix/index/IndexUpgradeToolTest.java
```diff
@@ -125,6 +125,30 @@ public void testIfOptionsArePassedToIndexTool() {
         it.populateIndexToolAttributes(commandLine);
     }
 
+    @Test
```
swaroopak, line 128: If you prefer to have a new test for this, please consider removing duplicate code in a function that can be called from both tests.

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/index/IndexTool.java
```diff
@@ -667,6 +676,7 @@ public int run(String[] args) throws Exception {
         }
         populateIndexToolAttributes(cmdLine);
         Configuration configuration = getConfiguration(tenantId);
+        decorateConfWithProperties(configuration, cmdLine);
```
swaroopak, line 679: Instead of calling this function here, why not do it inside the 678 to follow the setup code in the tool? You can extract properties in the function call at line no. 677 and use/pass in at 678.

abhishek-chouhan: Updated the patch to do this.

phoenix-core/src/test/java/org/apache/phoenix/index/IndexUpgradeToolTest.java
abhishek-chouhan: The test is to ensure that we can override the properties through IndexUpgradeTool too and that the properties are passed correctly to IndexTool and can be parsed there. I've taken out the index tool parsing part which was common with already written tests.

PR #771, opened 2020-04-27
PHOENIX-5872 Close Internal Phoenix Connections that were running during cancel

dbwong:
https://issues.apache.org/jira/browse/PHOENIX-5872

### DIFF
--- a/phoenix-core/src/it/java/org/apache/phoenix/query/MaxConcurrentConnectionsIT.java
+++ b/phoenix-core/src/it/java/org/apache/phoenix/query/MaxConcurrentConnectionsIT.java
@@ -1,0 +1,132 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.phoenix.query;
+
+import org.apache.hadoop.hbase.HBaseTestingUtility;
+import org.apache.phoenix.end2end.BaseUniqueNamesOwnClusterIT;
+import org.apache.phoenix.jdbc.PhoenixDriver;
+import org.apache.phoenix.util.DelayedRegionServer;
+import org.apache.phoenix.util.PhoenixRuntime;
+import org.junit.AfterClass;
+import org.junit.BeforeClass;
+import org.junit.Test;
+
+import java.sql.Connection;
+import java.sql.DriverManager;
+import java.sql.SQLException;
+import java.sql.Statement;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Properties;
+
+import static org.apache.phoenix.exception.SQLExceptionCode.NEW_CONNECTION_THROTTLED;
+import static org.apache.phoenix.exception.SQLExceptionCode.NEW_INTERNAL_CONNECTION_THROTTLED;
+import static org.apache.phoenix.monitoring.GlobalClientMetrics.GLOBAL_OPEN_INTERNAL_PHOENIX_CONNECTIONS;
+import static org.apache.phoenix.monitoring.GlobalClientMetrics.GLOBAL_OPEN_PHOENIX_CONNECTIONS;
+import static org.apache.phoenix.query.QueryServices.CLIENT_CONNECTION_MAX_ALLOWED_CONNECTIONS;
+import static org.apache.phoenix.query.QueryServices.INTERNAL_CONNECTION_MAX_ALLOWED_CONNECTIONS;
+import static org.apache.phoenix.query.QueryServices.RENEW_LEASE_ENABLED;
+import static org.apache.phoenix.query.QueryServices.TASK_HANDLING_INITIAL_DELAY_MS_ATTRIB;
+import static org.apache.phoenix.query.QueryServices.TASK_HANDLING_INTERVAL_MS_ATTRIB;
+import static org.apache.phoenix.util.PhoenixRuntime.JDBC_PROTOCOL_SEPARATOR;
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertTrue;
+import static org.junit.Assert.fail;
+
+/**
+ * Note that some tests for concurrentConnections live in PhoenixMetricsIT.java which also test the metric emission
+ */
+public class MaxConcurrentConnectionsIT extends BaseUniqueNamesOwnClusterIT {
+
+    private static HBaseTestingUtility hbaseTestUtil;
+
+    @BeforeClass
+    public static void setUp() throws Exception {
+        hbaseTestUtil = new HBaseTestingUtility();
+
+        hbaseTestUtil.startMiniCluster(1,1,null,null,DelayedRegionServer.class);
+        // establish url and quorum. Need to use PhoenixDriver and not PhoenixTestDriver
+        String zkQuorum = "localhost:" + hbaseTestUtil.getZkCluster().getClientPort();
+        url = PhoenixRuntime.JDBC_PROTOCOL + JDBC_PROTOCOL_SEPARATOR + zkQuorum +
+                JDBC_PROTOCOL_SEPARATOR + "uniqueConn=A";
+        DriverManager.registerDriver(PhoenixDriver.INSTANCE);
+    }
+
+    private String getUniqueUrl() {
+        return url + generateUniqueName();
+    }
+
+    //Have to shutdown our special delayed region server
+    @AfterClass
+    public static void tearDown() throws Exception {
+        hbaseTestUtil.shutdownMiniCluster();
+    }
+
+    /**
+     * This tests the delete path which creates a internal phoenix connection per region
+     * @throws Exception
+     */
+    @Test
+    public void testDeleteRuntimeFailureClosesConnections() throws Exception {
+        String tableName = generateUniqueName();
+        String connectionUrl = getUniqueUrl();
+        //table with lots of regions
+        String ddl = "create table " + tableName +  "  (i integer not null primary key, j integer) SALT_BUCKETS=256 ";
+
+        Properties props = new Properties();
+        props.setProperty(CLIENT_CONNECTION_MAX_ALLOWED_CONNECTIONS,String.valueOf(10));
+        props.setProperty(INTERNAL_CONNECTION_MAX_ALLOWED_CONNECTIONS,String.valueOf(10));
+
+        //delay any task handeling as that causes additional connections
+        props.setProperty(TASK_HANDLING_INTERVAL_MS_ATTRIB,String.valueOf(600000));
+        props.setProperty(TASK_HANDLING_INITIAL_DELAY_MS_ATTRIB,String.valueOf(600000));
+
+        String deleteStmt = "DELETE FROM " + tableName + " WHERE 20 = j";
+
+        try(Connection conn = DriverManager.getConnection(connectionUrl, props); Statement statement = conn.createStatement()) {
+            statement.execute(ddl);
+        }
+
+        assertEquals(0, GLOBAL_OPEN_PHOENIX_CONNECTIONS.getMetric().getValue());
+        assertEquals(0, GLOBAL_OPEN_INTERNAL_PHOENIX_CONNECTIONS.getMetric().getValue());
+        Connection conn = null;
+        try {
+            conn = DriverManager.getConnection(connectionUrl, props);
+            //Enable delay for the delete
+            DelayedRegionServer.setDelayEnabled(true);
+            try (Statement statement = conn.createStatement()) {
+                statement.execute(deleteStmt);
+            }
+            fail();
+        } catch (SQLException e) {
+            assertEquals(NEW_INTERNAL_CONNECTION_THROTTLED.getErrorCode(), e.getErrorCode());
+            assertEquals(NEW_INTERNAL_CONNECTION_THROTTLED.getSQLState(), e.getSQLState());
+        } finally {
+            DelayedRegionServer.setDelayEnabled(false);
+            if (conn != null) {
+                conn.close();
+            }
+            long connections = GLOBAL_OPEN_PHOENIX_CONNECTIONS.getMetric().getValue();
+            assertEquals(String.format("Found %d connections still open.", connections),0,connections);
+            connections = GLOBAL_OPEN_INTERNAL_PHOENIX_CONNECTIONS.getMetric().getValue();
+            assertEquals(String.format("Found %d internal connections still open.", connections),0 ,connections);
+        }
+
+    }
+}
--- a/phoenix-core/src/it/java/org/apache/phoenix/util/DelayedRegionServer.java
+++ b/phoenix-core/src/it/java/org/apache/phoenix/util/DelayedRegionServer.java
@@ -1,0 +1,119 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.phoenix.util;
+
+import com.google.protobuf.RpcController;
+import com.google.protobuf.ServiceException;
+import org.apache.hadoop.conf.Configuration;
+import org.apache.hadoop.hbase.CoordinatedStateManager;
+import org.apache.hadoop.hbase.MiniHBaseCluster;
+import org.apache.hadoop.hbase.protobuf.generated.ClientProtos;
+import org.apache.hadoop.hbase.regionserver.HRegionServer;
+import org.apache.hadoop.hbase.regionserver.RSRpcServices;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.io.IOException;
+
+/**
+ * This is a extended MiniHbaseCluster Region Server whcih allows developer/tester to inject
+ * delay into specific server side operations for testing.
+ */
+public class DelayedRegionServer extends MiniHBaseCluster.MiniHBaseClusterRegionServer {
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(DelayedRegionServer.class);
+
+    static boolean doDelay = false;
+    // Activate the delays after table creation to test get/scan/put
+    private static int DELAY_GET = 0;
+    private static int DELAY_SCAN = 30000;
+    private static int DELAY_MUTATE = 0;
+
+    public static void setDelayEnabled(boolean delay) {
+        doDelay = delay;
+    }
+
+    public static void setDelayGet(int delayGet) {
+        DELAY_GET = delayGet;
+    }
+
+    public static void setDelayScan(int delayScan) {
+        DELAY_SCAN = delayScan;
+    }
+
+    public static void setDelayMutate(int delayMutate) {
+        DELAY_MUTATE = delayMutate;
+    }
+
+    public DelayedRegionServer(Configuration conf, CoordinatedStateManager cp)
+            throws IOException, InterruptedException {
+        super(conf, cp);
+    }
+
+    @Override protected RSRpcServices createRpcServices() throws IOException {
+        return new DelayedRSRpcServices(this);
+    }
+
+    /**
+     * This class injects delay for Rpc calls and after executes super methods is delay is set.
+     */
+    public static class DelayedRSRpcServices extends RSRpcServices {
+
+        DelayedRSRpcServices(HRegionServer rs) throws IOException {
+            super(rs);
+        }
+
+        @Override public ClientProtos.GetResponse get(RpcController controller,
+                ClientProtos.GetRequest request) throws ServiceException {
+            try {
+                if (doDelay) {
+                    Thread.sleep(DELAY_GET);
+                }
+            } catch (InterruptedException e) {
+                LOGGER.error("Sleep interrupted during get operation", e);
+            }
+            return super.get(controller, request);
+        }
+
+        @Override public ClientProtos.MutateResponse mutate(RpcController rpcc,
+                ClientProtos.MutateRequest request) throws ServiceException {
+            try {
+                if (doDelay) {
+                    Thread.sleep(DELAY_MUTATE);
+                }
+            } catch (InterruptedException e) {
+                LOGGER.error("Sleep interrupted during mutate operation", e);
+            }
+            return super.mutate(rpcc, request);
+        }
+
+        @Override public ClientProtos.ScanResponse scan(RpcController controller,
+                ClientProtos.ScanRequest request) throws ServiceException {
+            try {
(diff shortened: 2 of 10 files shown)


dbwong:
First pass @ChinmaySKulkarni

phoenix-core/src/it/java/org/apache/phoenix/query/MaxConcurrentConnectionsIT.java
```diff
@@ -0,0 +1,152 @@
...
+    }
+
+    //Have to shutdown our special delayed region server
+    @AfterClass
+    public static void tearDown() throws Exception {
+        hbaseTestUtil.shutdownMiniCluster();
+    }
+
+    @Test
+    public void testCannotCreateOverMaximumConnections() throws Exception {
```
ChinmaySKulkarni, line 75: It looks like `PhoenixMetricsIT` already has tests for checking the working on client-connection throttling. Do we need this extra test?

phoenix-core/src/it/java/org/apache/phoenix/query/MaxConcurrentConnectionsIT.java
```diff
@@ -0,0 +1,152 @@
...
+     */
+    @Test
+    public void testDeleteRuntimeFailureClosesConnections() throws Exception {
+        String tableName = generateUniqueName();
+
+        //table with lots of regions
+        String ddl = "create table " + tableName +  "  (i integer not null primary key, j integer) SALT_BUCKETS=256 ";
+
+        Properties props = new Properties();
+        props.setProperty(QueryServices.ENABLE_SERVER_SIDE_DELETE_MUTATIONS,
```
ChinmaySKulkarni, line 117: What happens if server-side mutations are enabled? If enabled, since we no longer would need to `rs.next` and read all the data from the server + generate delete mutations for them, is there any difference in behavior w.r.t. whether we create cloned connections?

phoenix-core/src/it/java/org/apache/phoenix/query/MaxConcurrentConnectionsIT.java
```diff
@@ -0,0 +1,152 @@
...
+                statement.execute(ddl);
+            }
+            //Enable delay for the delete
+            DelayedRegionServer.setDelayEnabled(true);
+            try(Statement statement = conn.createStatement()) {
+                statement.execute(deleteStmt);
+            }
+            fail();
+        } catch (SQLException e) {
+            assertEquals(NEW_CONNECTION_THROTTLED.getErrorCode(),e.getErrorCode());
```
ChinmaySKulkarni, line 140: So here, the client opens just 1 connection and the delete path creates 1 cloned conn per region and so this leads to client throttling? I thought internal conns wouldn't get counted in the client conn throttling?

phoenix-core/src/it/java/org/apache/phoenix/query/MaxConcurrentConnectionsIT.java
```diff
@@ -0,0 +1,152 @@
...
+            assertEquals(NEW_CONNECTION_THROTTLED.getSQLState(),e.getSQLState());
+        } finally {
+            DelayedRegionServer.setDelayEnabled(false);
+            if(conn != null) {
+                conn.close();
+            }
+            long connections = GLOBAL_OPEN_PHOENIX_CONNECTIONS.getMetric().getValue();
+            assertTrue(String.format("Found %d connections still open.",connections),2 >= connections);
+            connections = GLOBAL_OPEN_INTERNAL_PHOENIX_CONNECTIONS.getMetric().getValue();
```
ChinmaySKulkarni, line 149: I'm not sure I understand this condition. Aren't we expecting 10 conns?

phoenix-core/src/it/java/org/apache/phoenix/util/DelayedRegionServer.java
```diff
@@ -0,0 +1,109 @@
...
+import org.apache.hadoop.conf.Configuration;
+import org.apache.hadoop.hbase.CoordinatedStateManager;
+import org.apache.hadoop.hbase.MiniHBaseCluster;
+import org.apache.hadoop.hbase.protobuf.generated.ClientProtos;
+import org.apache.hadoop.hbase.regionserver.HRegionServer;
+import org.apache.hadoop.hbase.regionserver.RSRpcServices;
+
+import java.io.IOException;
+
+public class DelayedRegionServer extends MiniHBaseCluster.MiniHBaseClusterRegionServer {
```
ChinmaySKulkarni, line 32: nit: Add class-level comment

dbwong:
Accidentally added the where optimizer changes from another PR. I'll address shortly but that was the only failing test.

phoenix-core/src/it/java/org/apache/phoenix/util/DelayedRegionServer.java
```diff
@@ -0,0 +1,109 @@
...
+        }
+
+        @Override public ClientProtos.GetResponse get(RpcController controller,
+                ClientProtos.GetRequest request) throws ServiceException {
+            try {
+                if (doDelay) {
+                    Thread.sleep(DELAY_GET);
+                }
+            } catch (InterruptedException e) {
+                LOG.error("Sleep interrupted during get operation", e);
```
ChinmaySKulkarni, line 80: Shouldn't this class use its own logger?

phoenix-core/src/main/java/org/apache/phoenix/query/ConnectionQueryServicesImpl.java
```diff
@@ -4957,12 +4964,22 @@ private void returnAllSequences(ConcurrentMap<SequenceKey,Sequence> sequenceMap)
     public void addConnection(PhoenixConnection connection) throws SQLException {
         if (returnSequenceValues || shouldThrottleNumConnections) {
             synchronized (connectionCountLock) {
-                if (shouldThrottleNumConnections && connectionCount + 1 > maxConnectionsAllowed){
-                    GLOBAL_PHOENIX_CONNECTIONS_THROTTLED_COUNTER.increment();
-                    throw new SQLExceptionInfo.Builder(SQLExceptionCode.NEW_CONNECTION_THROTTLED).
-                        build().buildException();
+
+                if(shouldThrottleNumConnections) {
```
ChinmaySKulkarni, line 4968: The value of `shouldThrottleNumConnections` should now be set based on this internal conn config and the max num conns from a client right?

phoenix-core/src/main/java/org/apache/phoenix/query/ConnectionQueryServicesImpl.java
```diff
@@ -4957,12 +4964,22 @@ private void returnAllSequences(ConcurrentMap<SequenceKey,Sequence> sequenceMap)
...
+                    int futureConnections = 1 + ( connection.isInternalConnection() ? internalConnectionCount : connectionCount);
+                    int allowedConnections = connection.isInternalConnection() ? maxInternalConnectionsAllowed : maxConnectionsAllowed;
+                    if(allowedConnections != 0 && futureConnections > allowedConnections) {
+                        GLOBAL_PHOENIX_CONNECTIONS_THROTTLED_COUNTER.increment();
```
ChinmaySKulkarni, line 4972: Better to add a comment somewhere around here about the expected working of counting internal phoenix conns and how/if this works against client conns, etc.

phoenix-core/src/main/java/org/apache/phoenix/query/ConnectionQueryServicesImpl.java
```diff
@@ -4957,12 +4964,22 @@ private void returnAllSequences(ConcurrentMap<SequenceKey,Sequence> sequenceMap)
...
+                        throw new SQLExceptionInfo.Builder(SQLExceptionCode.NEW_CONNECTION_THROTTLED).
```
ChinmaySKulkarni, line 4973: Should we throw a different error for cases where internal conns exceed the max config vs. when external do?

phoenix-core/src/main/java/org/apache/phoenix/query/ConnectionQueryServicesImpl.java
```diff
@@ -4977,25 +4994,27 @@ public void removeConnection(PhoenixConnection connection) throws SQLException {
         if (returnSequenceValues) {
             ConcurrentMap<SequenceKey,Sequence> formerSequenceMap = null;
             synchronized (connectionCountLock) {
-                if (--connectionCount <= 0) {
-                    if (!this.sequenceMap.isEmpty()) {
-                        formerSequenceMap = this.sequenceMap;
-                        this.sequenceMap = Maps.newConcurrentMap();
+                if(!connection.isInternalConnection()) {
```
ChinmaySKulkarni, line 4997: We used to decrement `connectionCount` here before. Just commenting to make sure we are still  decrementing correctly.

phoenix-core/src/main/java/org/apache/phoenix/query/ConnectionQueryServicesImpl.java
```diff
@@ -4977,25 +4994,27 @@ public void removeConnection(PhoenixConnection connection) throws SQLException {
...
+                    if (connectionCount + internalConnectionCount - 1 <= 0) {
```
ChinmaySKulkarni, line 4998: hmm, can you explain this condition? Why does internalConnectionCount come into the picture here?

phoenix-core/src/main/java/org/apache/phoenix/query/ConnectionQueryServicesImpl.java
dbwong: We decrement in the next block.

phoenix-core/src/main/java/org/apache/phoenix/query/ConnectionQueryServicesImpl.java
dbwong: I was trying to leave this condition the same as before thus combining both connection counts.  This appears to me to be sequence handling for no connection open.

phoenix-core/src/main/java/org/apache/phoenix/query/ConnectionQueryServicesImpl.java
dbwong: We could but I'm not sure there is much benefit as we already get different line number and the end client can't really act differently.

phoenix-core/src/main/java/org/apache/phoenix/query/ConnectionQueryServicesImpl.java
dbwong: Good point, added.

phoenix-core/src/it/java/org/apache/phoenix/util/DelayedRegionServer.java
dbwong: Sure

phoenix-core/src/it/java/org/apache/phoenix/util/DelayedRegionServer.java
dbwong: done

phoenix-core/src/it/java/org/apache/phoenix/query/MaxConcurrentConnectionsIT.java
dbwong: I think we should move those tests to here.  Metrics tests should not overlap with behavior imo.  Are you okay with that change?

phoenix-core/src/it/java/org/apache/phoenix/query/MaxConcurrentConnectionsIT.java
```diff
@@ -0,0 +1,152 @@
...
+import static org.junit.Assert.fail;
+
+public class MaxConcurrentConnectionsIT extends BaseUniqueNamesOwnClusterIT {
+
+    private static HBaseTestingUtility hbaseTestUtil;
+
+    @BeforeClass
+    public static void setUp() throws Exception {
+        hbaseTestUtil = new HBaseTestingUtility();
+        hbaseTestUtil.startMiniCluster(1,1,null,null,DelayedRegionServer.class);
```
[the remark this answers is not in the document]
dbwong: That API you mention was added in 2.2.  Do you want me to split the implementations to handle that?  I don't think the 4.x handles hbase 2.X yet.

phoenix-core/src/it/java/org/apache/phoenix/end2end/QueryTimeoutIT.java
```diff
@@ -139,4 +165,9 @@ public void testQueryTimeout() throws Exception {
         }
         conn.close();
     }
+
+
+
+
+
```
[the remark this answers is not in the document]
dbwong: We don't need.  I was using this as a test for my delayed region server as it addresses a testing issue in that file.  I'll revert this file.

phoenix-core/src/it/java/org/apache/phoenix/query/MaxConcurrentConnectionsIT.java
dbwong: This hits the internal throttling defined as 10.

phoenix-core/src/it/java/org/apache/phoenix/query/MaxConcurrentConnectionsIT.java
dbwong: No we are closing most of the connections now. The 2 was initially chosen as there were occasionally background tasks in phoenix making the assert inconsistent.

phoenix-core/src/main/java/org/apache/phoenix/query/ConnectionQueryServicesImpl.java
dbwong: Sure

phoenix-core/src/it/java/org/apache/phoenix/query/MaxConcurrentConnectionsIT.java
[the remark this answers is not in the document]
gjacoby126: @dbwong - Phoenix 4.x supports 1.3, 1.4, and 1.5. Phoenix 5.0 supports 2.0, and the forthcoming 5.1 supports 2.0, 2.1 and 2.2.

phoenix-core/src/it/java/org/apache/phoenix/query/MaxConcurrentConnectionsIT.java
gjacoby126: Let's be careful not to create a flapper -- do we need a wait step for the background tasks to finish?

phoenix-core/src/main/java/org/apache/phoenix/query/ConnectionQueryServicesImpl.java
gjacoby126: I can see a different error message being useful, because it keeps the client developer from wasting a lot of time looking for a non-existent leak on their side.

phoenix-core/src/it/java/org/apache/phoenix/query/MaxConcurrentConnectionsIT.java
dbwong: Tested locally and does not appear to be a behavior change in terms of leaked internal phoenix connections.

phoenix-core/src/main/java/org/apache/phoenix/query/ConnectionQueryServicesImpl.java
dbwong: Added another sql exception code and surface different exception based on connection type.

phoenix-core/src/it/java/org/apache/phoenix/query/MaxConcurrentConnectionsIT.java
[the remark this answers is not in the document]
dbwong: Thoughts @ChinmaySKulkarni do you want me to pull out this interface to our hbase-compat?

phoenix-core/src/it/java/org/apache/phoenix/query/MaxConcurrentConnectionsIT.java
dbwong: Waits are by nature flappers as well unless carefully constructed with controls.  The system task table is the main one i'm concerned with.  Maybe there is a way to disable system.task for this integration test?

phoenix-core/src/it/java/org/apache/phoenix/query/MaxConcurrentConnectionsIT.java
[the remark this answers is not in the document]
ChinmaySKulkarni: We should use the new API for the master branch. The 4.x patch can continue to use the old one.

phoenix-core/src/it/java/org/apache/phoenix/query/MaxConcurrentConnectionsIT.java
ChinmaySKulkarni: That's fine. In that case, can you remove repeated tests in `PhoenixMetricsIT`?

phoenix-core/src/it/java/org/apache/phoenix/query/MaxConcurrentConnectionsIT.java
ChinmaySKulkarni: In that case, is setting this property necessary?

phoenix-core/src/it/java/org/apache/phoenix/query/MaxConcurrentConnectionsIT.java
ChinmaySKulkarni: Ok, as I think I've mentioned elsewhere, we should have a separate exception for internal vs external conn throttling.

phoenix-core/src/it/java/org/apache/phoenix/query/MaxConcurrentConnectionsIT.java
ChinmaySKulkarni: I believe there is a way to change the task handling interval (set it to Long.MAX_VALUE or something for the sake of your tests). See https://github.com/apache/phoenix/blob/master/phoenix-core/src/main/java/org/apache/phoenix/coprocessor/TaskRegionObserver.java#L67
FYI @kadirozde is there a cleaner way to do this?

phoenix-core/src/main/java/org/apache/phoenix/query/ConnectionQueryServicesImpl.java
ChinmaySKulkarni: @dbwong I don't see this change. Can you please check again?

phoenix-core/src/it/java/org/apache/phoenix/query/MaxConcurrentConnectionsIT.java
```diff
@@ -0,0 +1,155 @@
...
+        } catch (SQLException e) {
+            assertEquals(NEW_INTERNAL_CONNECTION_THROTTLED.getErrorCode(), e.getErrorCode());
+            assertEquals(NEW_INTERNAL_CONNECTION_THROTTLED.getSQLState(), e.getSQLState());
+        } finally {
+            DelayedRegionServer.setDelayEnabled(false);
+            if (conn != null) {
+                conn.close();
+            }
+            long connections = GLOBAL_OPEN_PHOENIX_CONNECTIONS.getMetric().getValue();
+            assertTrue(String.format("Found %d connections still open.", connections), 0 == connections);
```
ChinmaySKulkarni, line 148: nit: Use assertEquals. Same with below check

phoenix-core/src/main/java/org/apache/phoenix/compile/MutatingParallelIteratorFactory.java
```diff
@@ -64,6 +64,8 @@ public PeekingResultIterator newIterator(final StatementContext parentContext,
             QueryPlan plan) throws SQLException {
 
         final PhoenixConnection clonedConnection = new PhoenixConnection(this.connection);
+        connection.addChildConnection(clonedConnection);
+        System.out.println("MutatingParallelIteratorFactor making cloned connections.");
```
ChinmaySKulkarni, line 68: @dbwong Left this print here.

phoenix-core/src/it/java/org/apache/phoenix/query/MaxConcurrentConnectionsIT.java
dbwong: done

phoenix-core/src/it/java/org/apache/phoenix/query/MaxConcurrentConnectionsIT.java
dbwong: Took a closer look and the metrics emission is tied to the behavior.  For now I've removed the test and made a note to see PhoenixMetricsIT

phoenix-core/src/it/java/org/apache/phoenix/query/MaxConcurrentConnectionsIT.java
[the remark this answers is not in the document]
dbwong: Sure will do

phoenix-core/src/it/java/org/apache/phoenix/query/MaxConcurrentConnectionsIT.java
dbwong: Done

phoenix-core/src/it/java/org/apache/phoenix/query/MaxConcurrentConnectionsIT.java
dbwong: Spoke with Kadir offline, he also recommended setting the initial task to a high value.  I've incorporated both into my IT.

phoenix-core/src/main/java/org/apache/phoenix/jdbc/PhoenixConnection.java
```diff
@@ -441,6 +453,17 @@ private static Properties filterKnownNonProperties(Properties info) {
         return result.build();
     }
 
+    public boolean isInternalConnection() {
+        return isInternalConnection;
+    }
+
+    public void addChildConnection(PhoenixConnection connection) {
```
ChinmaySKulkarni, line 460: If you're worried about multiple threads calling `addChildConnection` on the same connection object, then this method is still not thread safe. You need to synchronize the null check and element addition as an entire atomic operation.

phoenix-core/src/main/java/org/apache/phoenix/jdbc/PhoenixConnection.java
```diff
@@ -655,13 +678,20 @@ public void close() throws SQLException {
                     traceScope.close();
                 }
                 closeStatements();
+                if(childConnections != null) {
+                    SQLCloseables.closeAllQuietly(childConnections);
```
ChinmaySKulkarni, line 682: This is guaranteed to only ever happen from 1 thread at a time I suppose? Which is why we wouldn't care about thread-safety of `closeAllQuietly`?

ChinmaySKulkarni: But the check above it should also be synchronized i.e. `if(childConnections != null) {` no?

phoenix-core/src/main/java/org/apache/phoenix/jdbc/PhoenixConnection.java
ChinmaySKulkarni: And then, if we end up synchronizing over all accesses of the list anyways, we can probably do with a normal ArrayList too

dbwong: I'm hoping to only synchronize of the construction of the iterable, as I hope most threads will be able to add themselves in parallel as this is a possible new slowdown, think the effort is worth it over normal ArrayList?

phoenix-core/src/main/java/org/apache/phoenix/jdbc/PhoenixConnection.java
dbwong: Yep agreed, did this too late at night.  I've synchronized the creation/creation checks of this structure.

phoenix-core/src/main/java/org/apache/phoenix/jdbc/PhoenixConnection.java
```diff
@@ -441,6 +453,23 @@ private static Properties filterKnownNonProperties(Properties info) {
...
+    public boolean isInternalConnection() {
+        return isInternalConnection;
+    }
+
+    /**
+     * This method, and *only* this method is thread safe
+     * @param connection
+     */
+    public void addChildConnection(PhoenixConnection connection) {
+        synchronized (this) {
```
ChinmaySKulkarni, line 465: Avoid synchronizing on `this`. Let's have a separate variable which we synchronize on which only relates to the childConnections and use `@GuardedBy` syntax to denote this. Today we don't have any other methods synchronized on `this` inside PhoenixConnection, but if we do tomorrow, there will be unnecessary contention

dbwong:
Merged closing.

PR #772, opened 2020-04-29
PHOENIX-5878 - IndexTool should not fail jobs when the only errors ar…

gjacoby126:
…e due to rows out of MaxLookback


gjacoby126:
Thanks @priyankporwal and @kadirozde for the review.

PR #779, opened 2020-05-14
PHOENIX-5896: Implement incremental rebuild along the failed regions …

swaroopak:
…in IndexTool

IT and unit tests in progress. Sending out for early feedback.

### DIFF
--- a/phoenix-core/src/main/java/org/apache/phoenix/coprocessor/IndexRebuildRegionScanner.java
+++ b/phoenix-core/src/main/java/org/apache/phoenix/coprocessor/IndexRebuildRegionScanner.java
@@ -38,6 +38,7 @@ import java.util.concurrent.ExecutionException;
 
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.collect.Lists;
+import com.sun.org.apache.xpath.internal.operations.Bool;
 import org.apache.hadoop.conf.Configuration;
 import org.apache.hadoop.hbase.Cell;
 import org.apache.hadoop.hbase.CellUtil;
@@ -47,6 +48,7 @@ import org.apache.hadoop.hbase.HRegionInfo;
 import org.apache.hadoop.hbase.KeyValue;
 import org.apache.hadoop.hbase.client.Delete;
 import org.apache.hadoop.hbase.client.Durability;
+import org.apache.hadoop.hbase.client.Get;
 import org.apache.hadoop.hbase.client.Mutation;
 import org.apache.hadoop.hbase.client.Put;
 import org.apache.hadoop.hbase.client.Result;
@@ -57,6 +59,7 @@ import org.apache.hadoop.hbase.io.ImmutableBytesWritable;
 import org.apache.hadoop.hbase.regionserver.Region;
 import org.apache.hadoop.hbase.regionserver.RegionScanner;
 import org.apache.hadoop.hbase.util.Bytes;
+import org.apache.phoenix.mapreduce.index.IndexVerificationResultRepository;
 import org.apache.phoenix.util.EnvironmentEdgeManager;
 import org.apache.hadoop.hbase.util.Pair;
 import org.apache.phoenix.cache.ServerCacheClient;
@@ -111,6 +114,7 @@ public class IndexRebuildRegionScanner extends GlobalIndexRegionScanner {
     private Map<byte[], NavigableSet<byte[]>> familyMap;
     private byte[][] viewConstants;
     private IndexVerificationOutputRepository verificationOutputRepository;
+    private boolean skipped = false;
 
     @VisibleForTesting
     public IndexRebuildRegionScanner(final RegionScanner innerScanner, final Region region, final Scan scan,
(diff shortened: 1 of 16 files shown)


phoenix-core/src/main/java/org/apache/phoenix/coprocessor/IndexRebuildRegionScanner.java
```diff
@@ -159,6 +161,16 @@ public IndexRebuildRegionScanner(final RegionScanner innerScanner, final Region
         }
     }
 
+    private boolean shouldRebuildOrVerify() throws IOException {
```
gokceni, line 164: Please rename as shouldVerify

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/IndexRebuildRegionScanner.java
```diff
@@ -159,6 +161,16 @@ public IndexRebuildRegionScanner(final RegionScanner innerScanner, final Region
...
+        if(verifyType == IndexTool.IndexVerifyType.ONLY) {
+            return true;
+        }
+        byte [] rowKey = generateResultTableRowKey(scan.getTimeRange().getMax(),
```
gjacoby126, line 168: rather than importing the generateResultTableRowKey here, might be cleaner to have a ResultRepository "exists" overload that takes Scan, indexMaintainer, and region name. That keeps the internal details of the table from "leaking" back into the region scanner.

gjacoby126: or Scan, index table name, and region name. Whichever you prefer.

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/IndexRebuildRegionScanner.java
```diff
@@ -1191,6 +1203,9 @@ public boolean next(List<Cell> results) throws IOException {
         }
         Cell lastCell = null;
         int rowCount = 0;
+        if(!shouldRebuildOrVerify()) {
```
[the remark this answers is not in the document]
swaroopak: This was one way without taking any additional parameters. I think the read repair won't be impacted as it would have different ts than the key of PIT result and this `if` will not get executed. We can wrap this under a parameter if you think that's the right thing to do.

phoenix-core/src/test/java/org/apache/phoenix/index/ShouldVerifyTest.java
```diff
@@ -0,0 +1,91 @@
...
+import org.mockito.Matchers;
+import org.mockito.Mock;
+import org.mockito.MockitoAnnotations;
+
+import java.io.IOException;
+
+import static org.mockito.Matchers.any;
+import static org.mockito.Mockito.when;
+
+public class ShouldVerifyTest {
```
kadirozde, line 38: This test class is written to test a tiny method which does a simple check. Given how this test class is constructed, I do not see any value of these tests and the entire class. I am not sure if you notice that these tests verifies that the following block of code is correct
if(a != null || b == IndexTool.IndexVerifyType.ONLY) {
            return true;
        }
        return !c;
where a, b and c are supplied by the test. 

I suggest adding a test to an existing integration test to test this incremental verification feature end to end. The jira does not explain how this feature is activated. It seems one needs to provide the end time parameter. If so, the same end time parameter needs to be used for the base run and the incremental runs.  How we do differentiate the result table entries of one run from those for another run when these runs use the same end time? Do we need to differentiate it? How about the start time? What happens if the end times are the same but the start times are different for these runs. All these questions would have been considered if an integration test were written. Have you considered them?

gjacoby126: @kadirozde - while I agree an IT test is also necessary here, it's useful to have a unit test that can quickly verify behavior given a complicated state machine (even if the lines of code under test are few). What makes it tricky is that we're essentially verifying a private method here, which is a code smell. I don't have a great refactor off the top of my head.

swaroopak: @kadirozde Thank you for explaining the missing piece in the PR. I will add an IT in this one. I agree with @gjacoby126  and I faced the same challenge of refactoring it. Some unit tests are better than no tests I guess :(

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/index/IndexTool.java
```diff
@@ -235,15 +235,22 @@ public static IndexVerifyType fromValue(byte[] value) {
     private static final Option END_TIME_OPTION = new Option("et", "endtime",
             true, "End time for indextool rebuild or verify");
 
+    private static final Option INCREMENTAL_VERIFY_OPTION = new Option("iv", "incre-verify",
```
priyankporwal, line 238: Nit: Please change this to full "incremental-verify"

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/index/IndexTool.java
```diff
@@ -235,15 +235,22 @@ public static IndexVerifyType fromValue(byte[] value) {
     private static final Option END_TIME_OPTION = new Option("et", "endtime",
```
priyankporwal, line 235: For consistency: These should both use "start-time" / "end-time" .. just like all other composite-word options.

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/index/IndexTool.java
priyankporwal: Should we call this as "retry-verify" since that's what the implementation looks like?

PR #785, opened 2020-05-21
PHOENIX-5910 IndexTool verification-only runs should have counters fo…

kadirozde:
…r unverified rows


phoenix-core/src/main/java/org/apache/phoenix/coprocessor/IndexRebuildRegionScanner.java
```diff
@@ -635,6 +672,9 @@ public boolean verifySingleIndexRow(Result indexRow, IndexToolVerificationResult
         Collections.sort(expectedMutationList, MUTATION_TS_DESC_COMPARATOR);
         Collections.sort(actualMutationList, MUTATION_TS_DESC_COMPARATOR);
         if (verifyType == IndexTool.IndexVerifyType.ONLY) {
+            if (actualMutationList.get(0) instanceof Put) {
```
swaroopak, line 675: this means only the latest row right?
in that case, a comment would be helpful here 
"Read repair would happen on the latest row hence considering the latest mutation to update the Unverified counters" or something on the similar lines?

phoenix-core/src/it/java/org/apache/phoenix/end2end/index/GlobalIndexCheckerIT.java
```diff
@@ -446,6 +458,19 @@ public void testFailPostIndexRowUpdate() throws Exception {
...
+            assertEquals(3, indexTool.getJob().getCounters().findCounter(BEFORE_REBUILD_VALID_INDEX_ROW_COUNT).getValue());
+            assertEquals(0, indexTool.getJob().getCounters().findCounter(BEFORE_REBUILD_EXPIRED_INDEX_ROW_COUNT).getValue());
+            assertEquals(0, indexTool.getJob().getCounters().findCounter(BEFORE_REBUILD_INVALID_INDEX_ROW_COUNT).getValue());
+            assertEquals(0, indexTool.getJob().getCounters().findCounter(BEFORE_REBUILD_MISSING_INDEX_ROW_COUNT).getValue());
+            assertEquals(0, indexTool.getJob().getCounters().findCounter(BEFORE_REBUILD_BEYOND_MAXLOOKBACK_MISSING_INDEX_ROW_COUNT).getValue());
+            assertEquals(0, indexTool.getJob().getCounters().findCounter(BEFORE_REBUILD_BEYOND_MAXLOOKBACK_INVALID_INDEX_ROW_COUNT).getValue());
+            assertEquals(2, indexTool.getJob().getCounters().findCounter(BEFORE_REBUILD_UNVERIFIED_INDEX_ROW_COUNT).getValue());
+            assertEquals(0, indexTool.getJob().getCounters().findCounter(BEFORE_REBUILD_OLD_INDEX_ROW_COUNT).getValue());
+            assertEquals(0, indexTool.getJob().getCounters().findCounter(BEFORE_REBUILD_UNKNOWN_INDEX_ROW_COUNT).getValue());
```
swaroopak, line 473: Would it be easier to generate a row with empty CF 'x' or not there so that we get the other two counters also?

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/IndexRebuildRegionScanner.java
```diff
@@ -489,6 +490,42 @@ private boolean isDeleteFamily(Mutation mutation) {
...
+        byte[] indexRowKey = getIndexRowKey(indexMaintainer, put);
+        if (Bytes.compareTo(actual.getRow(), 0, actual.getRow().length,
+                indexRowKey, 0, indexRowKey.length) != 0) {
+            return;
+        }
+        List<Cell> cellList = actual.get(indexMaintainer.getEmptyKeyValueFamily().copyBytesIfNecessary(),
+                indexMaintainer.getEmptyKeyValueQualifier());
+        Cell cell = (cellList != null && !cellList.isEmpty()) ? cellList.get(0) : null;
+        if (cell == null) {
+            verificationPhaseResult.setUnknownIndexRowCount(verificationPhaseResult.getUnknownIndexRowCount() + 1);
```
swaroopak, line 515: does this mean no empty CF?

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/IndexRebuildRegionScanner.java
kadirozde: Yes, look at the previous line : 
Collections.sort(actualMutationList, MUTATION_TS_DESC_COMPARATOR);

Such a comment would be misleading as read repair can happen on any version. If you query with a SCN connection, then read repair will happen on older version. We do read repair only on the latest version as older versions will be always be unverified before newer versions are inserted.

phoenix-core/src/it/java/org/apache/phoenix/end2end/index/GlobalIndexCheckerIT.java
kadirozde: The same code works for one value and works for the others too.  Nothing special for one value vs another here. To have an empty column with "x", I need to update the row with HBase API.

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/IndexRebuildRegionScanner.java
kadirozde: Yes. Is not it clear?

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/IndexRebuildRegionScanner.java
kadirozde: I added a comment to emphasize that we verify only the latest version as you suggested

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/IndexRebuildRegionScanner.java
priyankporwal: I had to read a few lines of code above to understand that, but I also haven't worked on it. Maybe a comment will ease quick code surfing.

kadirozde: Ok. I added a bunch of comments

PR #786, opened 2020-05-22
PHOENIX-5909 - Table and index-level metrics for indexing coprocs

gjacoby126:


abhishek-chouhan:
@gjacoby126 Would the counters corresponding to a particular table exist even after lets say the table doesn't have any region on that rs anymore(table dropped, region moved etc.)? Should we use different registries for each table along with HBase's ref counting support for those?

PR #787, opened 2020-05-22
PHOENIX-5890: Port Phoenix-5799 to master

swaroopak:


phoenix-core/src/main/java/org/apache/phoenix/mapreduce/index/IndexVerificationResultRepository.java
```diff
@@ -0,0 +1,218 @@
...
+                verificationResult.update(cell);
+            }
+        }
+        return verificationResult;
+    }
+
+    public void close() throws IOException {
+        if (resultHTable != null) {
+            resultHTable.close();
+        }
```
gjacoby126, line 215: @swaroopak - need to close indexHTable if not null as well. (The original 5799 did so too)

swaroopak: The patch attached in 5799 didn't have the close on the index table. Anyway, I added them to both (Result/Output)  the repositories. Please take a look.

PR #789, opened 2020-05-27
PHOENIX-5783: Implement starttime in IndexTool for rebuild and verifi…

swaroopak:
…cation

### DIFF
--- a/phoenix-core/src/it/java/org/apache/phoenix/end2end/IndexToolForNonTxGlobalIndexIT.java
+++ b/phoenix-core/src/it/java/org/apache/phoenix/end2end/IndexToolForNonTxGlobalIndexIT.java
@@ -48,6 +48,7 @@ import org.apache.phoenix.query.QueryServicesOptions;
 import org.apache.phoenix.schema.PIndexState;
 import org.apache.phoenix.util.EnvironmentEdgeManager;
 import org.apache.phoenix.util.IndexScrutiny;
+import org.apache.phoenix.util.ManualEnvironmentEdge;
 import org.apache.phoenix.util.PropertiesUtil;
 import org.apache.phoenix.util.ReadOnlyProps;
 import org.apache.phoenix.util.SchemaUtil;
@@ -159,6 +160,7 @@ public class IndexToolForNonTxGlobalIndexIT extends BaseUniqueNamesOwnClusterIT 
             deleteAllRows(conn,
                 TableName.valueOf(IndexVerificationResultRepository.RESULT_TABLE_NAME));
         }
+        EnvironmentEdgeManager.reset();
     }
 
     @Test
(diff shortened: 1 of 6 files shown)


phoenix-core/src/main/java/org/apache/phoenix/coprocessor/UngroupedAggregateRegionObserver.java
```diff
@@ -1081,11 +1086,39 @@ private RegionScanner rebuildIndices(final RegionScanner innerScanner, final Reg
             rawScan.setRaw(true);
             rawScan.setMaxVersions();
             rawScan.getFamilyMap().clear();
-            rawScan.setFilter(null);
             rawScan.setCacheBlocks(false);
             for (byte[] family : scan.getFamilyMap().keySet()) {
                 rawScan.addFamily(family);
             }
+            rawScan.setFilter(null);
+            //override the filter to skip scan when lower bound of timerange is passed
```
gjacoby126, line 1094: More comments please about why we're scanning through everything twice. I figured it out (grabbing the keys that existed at minTime, then doing a skip scan of just those keys), but I had to stare at it for a few min. :-)

phoenix-core/src/it/java/org/apache/phoenix/end2end/IndexToolForNonTxGlobalIndexIT.java
```diff
@@ -588,9 +588,101 @@ public void testIndexToolForIncrementalRebuild() throws Exception {
...
+        Properties props = PropertiesUtil.deepCopy(TEST_PROPERTIES);
+        try (Connection conn = DriverManager.getConnection(getUrl(), props)) {
+            conn.setAutoCommit(true);
+            conn.createStatement().execute("CREATE TABLE "+dataTableFullName+" "
+                    + "(key1 BIGINT NOT NULL, key2 BIGINT NOT NULL, val1 VARCHAR, val2 BIGINT, "
+                    + "val3 BIGINT, val4 DOUBLE, val5 BIGINT, val6 VARCHAR "
+                    + "CONSTRAINT my_pk PRIMARY KEY(key1, key2)) "+tableDDLOptions);
+            conn.createStatement().execute(String.format(
+                    "CREATE INDEX "+indexTableName+" ON "+dataTableFullName+" (val3) INCLUDE(val5)"));
+            long t0 = EnvironmentEdgeManager.currentTimeMillis();
```
gjacoby126, line 606: Is the edge manager injected somewhere? Otherwise this seems prone to flapping if the wall clock milliseconds don't act quite the way you're expecting. Suggest either injecting a manual edge, or picking fixed timestamps for each time variable.

phoenix-core/src/it/java/org/apache/phoenix/end2end/IndexToolForNonTxGlobalIndexIT.java
```diff
@@ -587,6 +589,113 @@ public void testIndexToolForIncrementalRebuild() throws Exception {
         }
     }
 
+    @Test
+    public void testIndexToolForIncrementalVerify() throws Exception {
```
gjacoby126, line 593: Let's also have a test where we incrementally rebuild a view index. Since the copy-constructor of Scan also copies over filters, I _expect_ they're OK, but better to test rather than assume. :-)

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/IndexRebuildRegionScanner.java
```diff
@@ -1398,7 +1410,64 @@ public boolean next(List<Cell> results) throws IOException {
...
+                return null;
+            }
+            if (!keys.isEmpty()) {
+                nextStartKey = ByteUtil.calculateTheClosestNextRowKeyForPrefix(keys.get(keys.size() - 1).getLowerRange());
+            }
+            try {
+                ScanRanges scanRanges = ScanRanges.createPointLookup(keys);
+                scanRanges.initializeScan(incrScan);
+                SkipScanFilter skipScanFilter = scanRanges.getSkipScanFilter();
+                incrScan.setFilter(new SkipScanFilter(skipScanFilter, true));
```
gjacoby126, line 1458: What if there's already a Filter on that scan (such as if we're rebuilding a view index). I _think_ that's OK, because we generated our key list for the SkipScanFilter using the pre-existing view filter on incrScan which we got from the original rebuild Scan, but let's make sure we test and are really sure about that.

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/IndexRebuildRegionScanner.java
```diff
@@ -1398,7 +1410,64 @@ public boolean next(List<Cell> results) throws IOException {
...
+            incrScan.setTimeRange(minTimestamp, scan.getTimeRange().getMax());
+            incrScan.setRaw(true);
+            incrScan.setMaxVersions();
+            incrScan.getFamilyMap().clear();
+            incrScan.setCacheBlocks(false);
+            for (byte[] family : scan.getFamilyMap().keySet()) {
+                incrScan.addFamily(family);
+            }
+            if(nextStartKey != null) {
+                incrScan.setStartRow(nextStartKey);
```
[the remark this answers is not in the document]
gjacoby126: If we don't have a test for index rebuild with a DESC pk, (incremental or not), we really should.

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/IndexRebuildRegionScanner.java
swaroopak: Yes, I will add a test for view index

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/IndexRebuildRegionScanner.java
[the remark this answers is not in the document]
swaroopak: Yes, that's the idea.  Let me poke around if there is a test on DESC pk.

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/IndexRebuildRegionScanner.java
```diff
@@ -1398,7 +1410,64 @@ public boolean next(List<Cell> results) throws IOException {
...
+        // when lower bound of timerange is passed or newStartKey was populated
+        // from previous call to next()
+        if(minTimestamp!= 0) {
+            Scan incrScan = new Scan(scan);
+            incrScan.setTimeRange(minTimestamp, scan.getTimeRange().getMax());
+            incrScan.setRaw(true);
+            incrScan.setMaxVersions();
+            incrScan.getFamilyMap().clear();
+            incrScan.setCacheBlocks(false);
+            for (byte[] family : scan.getFamilyMap().keySet()) {
```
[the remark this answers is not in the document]
kadirozde: We can have families in raw scans but not column qualifiers as HBase returns exception when columns are specified.

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/IndexRebuildRegionScanner.java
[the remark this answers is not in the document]
kadirozde: This is at the HBase level and all row keys are sorted in the ascending order. Phoenix flips the bits in row keys for DESC pks. This should be a concern at the Phoenix level but not here.

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/IndexRebuildRegionScanner.java
```diff
@@ -1398,7 +1410,64 @@ public boolean next(List<Cell> results) throws IOException {
...
+            try {
+                ScanRanges scanRanges = ScanRanges.createPointLookup(keys);
+                scanRanges.initializeScan(incrScan);
+                SkipScanFilter skipScanFilter = scanRanges.getSkipScanFilter();
+                incrScan.setFilter(new SkipScanFilter(skipScanFilter, true));
+                //putting back the min time to 0 for index and data reads
+                incrScan.setTimeRange(0, scan.getTimeRange().getMax());
+                scan.setTimeRange(0, scan.getTimeRange().getMax());
+                return region.getScanner(incrScan);
+            } catch (IndexOutOfBoundsException ex) {
```
kadirozde, line 1463: You can eliminate IndexOutOfBoundsException by checking if key.size() == 0 and if so, returning innerScanner.

PR #790, opened 2020-05-27
PHOENIX-5922 - IndexUpgradeTool should always re-enable tables on fai…

gjacoby126:
…lure

### DIFF
--- a/phoenix-core/src/main/java/org/apache/phoenix/mapreduce/index/IndexUpgradeTool.java
+++ b/phoenix-core/src/main/java/org/apache/phoenix/mapreduce/index/IndexUpgradeTool.java
@@ -48,10 +48,8 @@ import org.apache.phoenix.query.ConnectionQueryServices;
 
 import org.apache.phoenix.query.QueryServices;
 import org.apache.phoenix.query.QueryServicesOptions;
-import org.apache.phoenix.schema.PIndexState;
 import org.apache.phoenix.schema.PTable;
 import org.apache.phoenix.schema.PTableType;
-import org.apache.phoenix.util.EnvironmentEdge;
 import org.apache.phoenix.util.EnvironmentEdgeManager;
 import org.apache.phoenix.util.MetaDataUtil;
 import org.apache.phoenix.util.PhoenixRuntime;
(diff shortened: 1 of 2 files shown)


gjacoby126:
btw, also found and fixed a bug with the rollback logic after a failed upgrade/downgrade where it was modifying a list being iterated over, which was causing failures.

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/index/IndexUpgradeTool.java
```diff
@@ -463,14 +489,20 @@ private void handleFailure(ConnectionQueryServices queryServices,
             upgrade = !upgrade;
 
             tablesAndIndexes.remove(dataTableFullName); //removing from the map
-            tableList.remove(dataTableFullName); //removing from the list
+            failedTables.add(dataTableFullName); //removing from the list
```
abhishek-chouhan, line 492: Nit: comment

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/index/IndexUpgradeTool.java
```diff
@@ -398,28 +414,34 @@ private int executeTool(Connection conn,
...
+        List<String> failedTables = new ArrayList<String>();
         for (String dataTableFullName : immutableList) {
             try (Admin admin = queryServices.getAdmin()) {
                 HashSet<String> indexes = tablesAndIndexes.get(dataTableFullName);
                 LOGGER.info("Executing " + operation + " of " + dataTableFullName
                         + " (immutable)");
                 disableTable(admin, dataTableFullName, indexes);
                 modifyTable(admin, dataTableFullName, indexes);
-            } catch (IOException | SQLException e) {
+            } catch (IOException | SQLException | RuntimeException e) {
```
abhishek-chouhan, line 438: Should we just catch throwable here(and in the mutable case), so that we cover errors too? in all cases i guess we want the tables to be enabled back.

gjacoby126: Done

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/index/IndexUpgradeTool.java
gjacoby126: Fixed

PR #792, opened 2020-05-29
PHOENIX-5920 Skip SYSTEM TABLE checks while creating phoenix connection if client has set the DoNotUpgrade config

nsgupta1:
I've made following changes to honor "DoNotUpgrade" config:
1. This PR makes a strong assumption that if client has set "DoNotUpgrade" config server already has SYSCAT tables even for firstConnection where isAutoUpgrade prop is enabled by default.
2. If "DoNotUpgrade" is set we are only doing client-server compatibility check and namespace checks.
3. We are skipping SYSCAT existence and timestamp related checks if "DoNotUpgrade" config is set.
I can remove the test cases that I've marked as ignore once we finalize its a valid removal.
@ChinmaySKulkarni @dbwong @gjacoby126 @jpisaac can you please review this PR ?

### DIFF
--- a/phoenix-core/src/it/java/org/apache/phoenix/end2end/SystemCatalogCreationOnConnectionIT.java
+++ b/phoenix-core/src/it/java/org/apache/phoenix/end2end/SystemCatalogCreationOnConnectionIT.java
@@ -674,8 +674,7 @@ public class SystemCatalogCreationOnConnectionIT {
         startMiniClusterWithToggleNamespaceMapping(Boolean.FALSE.toString());
         Properties propsDoNotUpgradePropSet = new Properties();
         // Create a dummy connection to make sure we have all system tables in place
-        try (Connection con1 = DriverManager.getConnection(getJdbcUrl(), propsDoNotUpgradePropSet))
-        {
+        try (Connection con1 = DriverManager.getConnection(getJdbcUrl(), propsDoNotUpgradePropSet)) {
             String ddl = "CREATE TABLE " + tableName + " (PK1 VARCHAR not null, " +
                     "PK2 VARCHAR not null, COL1 varchar, COL2 varchar "
                     + "CONSTRAINT pk PRIMARY KEY(PK1,PK2))";
@@ -683,18 +682,19 @@ public class SystemCatalogCreationOnConnectionIT {
             con1.createStatement().execute(
                     "UPSERT INTO " + tableName + " values ('pk1','pk2','c1','c2')");
             con1.commit();
+            // Stop HMaster to check if we can create connection without active HMaster
+            testUtil.getMiniHBaseCluster().getMaster().stopMaster();
+            // Set doNotUpgradeProperty to true
+            UpgradeUtil.doNotUpgradeOnFirstConnection(propsDoNotUpgradePropSet);
+            try (Connection con2 = DriverManager.getConnection(getJdbcUrl(), propsDoNotUpgradePropSet)) {
+                ResultSet rs = con2.createStatement().executeQuery("select * from " + tableName);
+                assertTrue(rs.next());
+                assertEquals("pk1", rs.getString(1));
+                assertEquals("pk2", rs.getString(2));
+                assertFalse(rs.next());
+            }
+        } finally {
+            testUtil.getMiniHBaseCluster().startMaster();
         }
-        // Stop HMaster to check if we can create connection without active HMaster
-        testUtil.getMiniHBaseCluster().getMaster().stopMaster();
-        // Set doNotUpgradeProperty to true
-        UpgradeUtil.doNotUpgradeOnFirstConnection(propsDoNotUpgradePropSet);
-        try(Connection con2 = DriverManager.getConnection(getJdbcUrl(), propsDoNotUpgradePropSet)) {
-            ResultSet rs = con2.createStatement().executeQuery("select * from " + tableName);
-            assertTrue(rs.next());
-            assertEquals("pk1", rs.getString(1));
-            assertEquals("pk2", rs.getString(2));
-            assertFalse(rs.next());
-        }
-        testUtil.getMiniHBaseCluster().startMaster();
     }
 }
--- a/phoenix-core/src/it/java/org/apache/phoenix/end2end/SystemCatalogCreationOnConnectionIT.java
+++ b/phoenix-core/src/it/java/org/apache/phoenix/end2end/SystemCatalogCreationOnConnectionIT.java
@@ -674,9 +674,11 @@ public class SystemCatalogCreationOnConnectionIT {
         startMiniClusterWithToggleNamespaceMapping(Boolean.FALSE.toString());
         Properties propsDoNotUpgradePropSet = new Properties();
         // Create a dummy connection to make sure we have all system tables in place
-        try (Connection con1 = DriverManager.getConnection(getJdbcUrl(), propsDoNotUpgradePropSet)) {
-            String ddl = "CREATE TABLE " + tableName + " (PK1 VARCHAR not null, PK2 VARCHAR not null, "
-                    + "COL1 varchar, COL2 varchar " + "CONSTRAINT pk PRIMARY KEY(PK1,PK2))";
+        try (Connection con1 = DriverManager.getConnection(getJdbcUrl(), propsDoNotUpgradePropSet))
+        {
+            String ddl = "CREATE TABLE " + tableName + " (PK1 VARCHAR not null, " +
+                    "PK2 VARCHAR not null, COL1 varchar, COL2 varchar "
+                    + "CONSTRAINT pk PRIMARY KEY(PK1,PK2))";
             con1.createStatement().execute(ddl);
             con1.createStatement().execute(
                     "UPSERT INTO " + tableName + " values ('pk1','pk2','c1','c2')");
@@ -693,5 +695,6 @@ public class SystemCatalogCreationOnConnectionIT {
             assertEquals("pk2", rs.getString(2));
             assertFalse(rs.next());
         }
+        testUtil.getMiniHBaseCluster().startMaster();
     }
 }
--- a/phoenix-core/src/it/java/org/apache/phoenix/end2end/SystemCatalogCreationOnConnectionIT.java
+++ b/phoenix-core/src/it/java/org/apache/phoenix/end2end/SystemCatalogCreationOnConnectionIT.java
@@ -43,6 +43,7 @@ import org.apache.hadoop.hbase.HConstants;
 import org.apache.hadoop.hbase.HTableDescriptor;
 import org.apache.hadoop.hbase.NamespaceNotFoundException;
 import org.apache.hadoop.hbase.TableName;
+import org.apache.hadoop.hbase.TableNotFoundException;
 import org.apache.phoenix.coprocessor.MetaDataProtocol;
 import org.apache.phoenix.exception.SQLExceptionCode;
 import org.apache.phoenix.exception.UpgradeRequiredException;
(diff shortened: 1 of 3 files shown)


phoenix-core/src/main/java/org/apache/phoenix/query/ConnectionQueryServicesImpl.java
```diff
@@ -1307,131 +1307,136 @@ private TableDescriptor ensureTableCreated(byte[] physicalTableName, PTableType
...
-                existingDesc = admin.getDescriptor(TableName.valueOf(physicalTableName));
-            } catch (org.apache.hadoop.hbase.TableNotFoundException e) {
-                tableExist = false;
-                if (tableType == PTableType.VIEW) {
-                    String fullTableName = Bytes.toString(physicalTableName);
-                    throw new ReadOnlyTableException(
-                            "An HBase table for a VIEW must already exist",
-                            SchemaUtil.getSchemaNameFromFullName(fullTableName),
-                            SchemaUtil.getTableNameFromFullName(fullTableName));
+            // If DoNotUpgrade config is set only check namespace mapping and
```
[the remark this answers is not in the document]
nsgupta1: Yes, you are right, this check should only be for system tables. I'll modify this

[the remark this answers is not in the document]
dbwong: Can you explain why this check has to be in this path and not in init?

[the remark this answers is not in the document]
nsgupta1: The basic goal of this check is to skip the call to [admin.getDescriptor](https://github.com/nsgupta1/phoenix/blob/18b219d7d270d546d7be29164229161531bd9330/phoenix-core/src/main/java/org/apache/phoenix/query/ConnectionQueryServicesImpl.java#L1311) and creation of system tables if "DoNotUpgrade" config is set. I had to add the if condition here anyway so I added CheckClientServerCompatibility too. Also we still need to keep the check from CQSI.ensureTableCreated 1259-1308. Then we will have to move this check also to init.

phoenix-core/src/main/java/org/apache/phoenix/query/ConnectionQueryServicesImpl.java
```diff
@@ -1306,6 +1306,12 @@ private TableDescriptor ensureTableCreated(byte[] physicalTableName, PTableType
                       .build().buildException();
                 }
             }
+            // If DoNotUpgrade config is set only check namespace mapping and
+            // Client-server compatibility for system tables.
+            if(isDoNotUpgradePropSet && SchemaUtil.isSystemTable(physicalTableName)) {
+                checkClientServerCompatibility(SYSTEM_CATALOG_NAME_BYTES);
```
ChinmaySKulkarni, line 1312: If the client has `doNotUpgrade == true` and SYSTEM.CATALOG doesn't exist, `checkClientServerCompatibility` will throw a `TableNotFoundException`. If we were to catch this exception and rethrow as `UpgradeRequiredException`, won't it follow a similar path as expected today?

phoenix-core/src/main/java/org/apache/phoenix/query/ConnectionQueryServicesImpl.java
```diff
@@ -1546,17 +1538,14 @@ public GetVersionResponse call(MetaDataService instance)
...
                                 + "Phoenix servers. Client version: "
                                 + MetaDataProtocol.CURRENT_CLIENT_VERSION
                                 + "; Server version: "
                                 + getServerVersion(serverJarVersion)
                                 + " The following servers require an updated "
                                 + QueryConstants.DEFAULT_COPROCESS_JAR_NAME
-                                + " to be put in the classpath of HBase: ");
-                        errorMessage.append(name);
-                        errorMessage.append(';');
+                                + " to be put in the classpath of HBase; ");
```
ChinmaySKulkarni, line 1548: Change this error message log now that we aren't mentioning the region name. Instead of "The following servers .." we can say "Servers require an updated .. "

phoenix-core/src/main/java/org/apache/phoenix/query/ConnectionQueryServicesImpl.java
```diff
@@ -3199,12 +3188,9 @@ public Void call() throws Exception {
                                     }
                                     return null;
                                 } catch (UpgradeRequiredException e) {
-                                    // This will occur in 3 cases:
-                                    // 1. SYSTEM.CATALOG does not exist and we don't want to allow the user to create it i.e.
-                                    //    !isAutoUpgradeEnabled or isDoNotUpgradePropSet is set
-                                    // 2. SYSTEM.CATALOG exists and its timestamp < MIN_SYSTEM_TABLE_TIMESTAMP
-                                    // 3. SYSTEM.CATALOG exists, but client and server-side namespace mapping is enabled so
-                                    //    we need to migrate SYSTEM tables to the SYSTEM namespace
+                                    // This will occur when SYSTEM.CATALOG exists, but client and
```
ChinmaySKulkarni, line 3191: So if the client has `doNotUpgrade == true`, SYSTEM.CATALOG exists, server and client both have ns-mapping enabled -> this will continue throwing `UpgradeRequiredException`?
What about the other 2 cases where we used to fall into this block and later `setUpgradeRequired`?

For original case1 (syscat doesn't exist and client has doNotUpgrade == true) -> You changes will lead to a TableNotFoundException and as mentioned earlier, can we catch and rethrow as UpgradeRequiredException so that this follows the original code path?

What about case2?

phoenix-core/src/it/java/org/apache/phoenix/end2end/SystemCatalogCreationOnConnectionIT.java
```diff
@@ -171,6 +173,7 @@ public void tearDownMiniCluster() {
 
      // Conditions: isDoNotUpgradePropSet is true
      // Expected: We do not create SYSTEM.CATALOG even if this is the first connection to the server
+     // UpgradeRequiredException is also not thrown even if it is first connection to server
```
ChinmaySKulkarni, line 176: Based on my comments below, this test can probably stay unchanged if we catch TNFE and rethrow as UpgradeRequiredException.

phoenix-core/src/it/java/org/apache/phoenix/end2end/SystemCatalogCreationOnConnectionIT.java
```diff
@@ -658,4 +670,30 @@ private Properties getClientProperties(boolean nsMappingEnabled, boolean systemT
         assertFalse(isSystemNamespaceCreated());
         return driver;
     }
+
+    @Test
+    public void testDoNotUpgradePropSet() throws Exception {
```
ChinmaySKulkarni, line 675: nit: Add comments like in other test cases as a header for the expected behavior for ease of understanding.

phoenix-core/src/it/java/org/apache/phoenix/end2end/SystemCatalogCreationOnConnectionIT.java
```diff
@@ -658,4 +670,30 @@ private Properties getClientProperties(boolean nsMappingEnabled, boolean systemT
...
+        String tableName = "HBASE_SYNTH_TEST";
+        startMiniClusterWithToggleNamespaceMapping(Boolean.FALSE.toString());
+        Properties propsDoNotUpgradePropSet = new Properties();
+        // Create a dummy connection to make sure we have all system tables in place
+        Connection con1 = DriverManager.getConnection(getJdbcUrl(),propsDoNotUpgradePropSet);
```
ChinmaySKulkarni, line 680: nit: try-with-resources for conn and stmt

phoenix-core/src/it/java/org/apache/phoenix/end2end/SystemCatalogCreationOnConnectionIT.java
```diff
@@ -658,4 +670,30 @@ private Properties getClientProperties(boolean nsMappingEnabled, boolean systemT
...
+        String ddl = "CREATE TABLE "+ tableName +" (PK1 VARCHAR not null, PK2 VARCHAR not null, "
+                + "COL1 varchar, COL2 varchar " + "CONSTRAINT pk PRIMARY KEY(PK1,PK2))";
+        con1.createStatement().execute(ddl);
+        con1.createStatement().execute("UPSERT INTO " + tableName + " values ('pk1','pk2','c1','c2')");
+        con1.commit();
+        // Stop HMaster to check if we can create connection without active HMaster
+        testUtil.getMiniHBaseCluster().getMaster().stopMaster();
+        // Set doNotUpgradeProperty to true
+        UpgradeUtil.doNotUpgradeOnFirstConnection(propsDoNotUpgradePropSet);
+        Connection con2 = DriverManager.getConnection(getJdbcUrl(),propsDoNotUpgradePropSet);
```
ChinmaySKulkarni, line 690: Same here.

phoenix-core/src/it/java/org/apache/phoenix/end2end/SystemCatalogCreationOnConnectionIT.java
```diff
@@ -658,4 +670,30 @@ private Properties getClientProperties(boolean nsMappingEnabled, boolean systemT
...
+        Properties propsDoNotUpgradePropSet = new Properties();
+        // Create a dummy connection to make sure we have all system tables in place
+        Connection con1 = DriverManager.getConnection(getJdbcUrl(),propsDoNotUpgradePropSet);
+        String ddl = "CREATE TABLE "+ tableName +" (PK1 VARCHAR not null, PK2 VARCHAR not null, "
+                + "COL1 varchar, COL2 varchar " + "CONSTRAINT pk PRIMARY KEY(PK1,PK2))";
+        con1.createStatement().execute(ddl);
+        con1.createStatement().execute("UPSERT INTO " + tableName + " values ('pk1','pk2','c1','c2')");
+        con1.commit();
+        // Stop HMaster to check if we can create connection without active HMaster
+        testUtil.getMiniHBaseCluster().getMaster().stopMaster();
```
ChinmaySKulkarni, line 687: Restart hmaster later?

ChinmaySKulkarni:
A generic comment: We set upgradeNotRequired when getting a connectionOnServer. Can you please check that this continues working as expected after your changes?

phoenix-core/src/main/java/org/apache/phoenix/query/ConnectionQueryServicesImpl.java
nsgupta1: Changed the error message.

phoenix-core/src/main/java/org/apache/phoenix/query/ConnectionQueryServicesImpl.java
nsgupta1: We are not checking case 2 if DoNotUpgrade conf is set to true.

phoenix-core/src/it/java/org/apache/phoenix/end2end/SystemCatalogCreationOnConnectionIT.java
nsgupta1: I tried to start hmaster later but it didn't work. I'll check and put this one in next commit.

phoenix-core/src/it/java/org/apache/phoenix/end2end/SystemCatalogCreationOnConnectionIT.java
nsgupta1: Added comments here.

gjacoby126:
JIRA is resolved, closing PR

PR #794, opened 2020-06-01
PHOENIX-5928 Index rebuilds without replaying data table mutations

kadirozde:
This PR eliminates data tables in the rebuild write path by letting UngroupedAggregateRegionObserver directly write to index tables without going through IndexRegionObserver.  By doing so, this PR eliminates going through data table update path and its overhead (its interaction with flushes, row locking etc). This PR also applies the index table updates in the index table row key order. It is expected that this will reduce the number of RPC calls compared to the existing implementation which generates index table updates in the data table row key order. Finally, the PR makes index rebuild from a given table region multi-threaded. Thus, this PR improves the rebuild performance and efficiency.

### DIFF
--- a/phoenix-core/src/main/java/org/apache/phoenix/coprocessor/GlobalIndexRegionScanner.java
+++ b/phoenix-core/src/main/java/org/apache/phoenix/coprocessor/GlobalIndexRegionScanner.java
@@ -19,6 +19,7 @@ package org.apache.phoenix.coprocessor;
 
 import org.apache.hadoop.conf.Configuration;
 import org.apache.hadoop.hbase.Cell;
+import org.apache.hadoop.hbase.CellUtil;
 import org.apache.hadoop.hbase.HConstants;
 import org.apache.hadoop.hbase.client.Delete;
 import org.apache.hadoop.hbase.client.Mutation;
@@ -30,6 +31,7 @@ import org.apache.hadoop.hbase.io.ImmutableBytesWritable;
 import org.apache.hadoop.hbase.regionserver.Region;
 import org.apache.hadoop.hbase.regionserver.RegionScanner;
 import org.apache.hadoop.hbase.regionserver.ScanInfoUtil;
+import org.apache.hadoop.hbase.util.Bytes;
 import org.apache.hadoop.hbase.util.Pair;
 import org.apache.phoenix.hbase.index.ValueGetter;
 import org.apache.phoenix.hbase.index.covered.update.ColumnReference;
@@ -42,22 +44,27 @@ import org.apache.phoenix.hbase.index.table.HTableFactory;
 import org.apache.phoenix.hbase.index.util.ImmutableBytesPtr;
 import org.apache.phoenix.index.IndexMaintainer;
 import org.apache.phoenix.index.PhoenixIndexCodec;
+import org.apache.phoenix.mapreduce.index.IndexTool;
 import org.apache.phoenix.mapreduce.index.IndexVerificationResultRepository;
 import org.apache.phoenix.query.QueryServicesOptions;
+import org.apache.phoenix.util.ScanUtil;
 import org.apache.phoenix.util.ServerUtil;
 
 import java.io.IOException;
 import java.util.List;
+import java.util.Map;
+import java.util.NavigableSet;
 
 import static org.apache.phoenix.hbase.index.write.AbstractParallelWriterIndexCommitter.INDEX_WRITER_KEEP_ALIVE_TIME_CONF_KEY;
 import static org.apache.phoenix.query.QueryServices.INDEX_REBUILD_PAGE_SIZE_IN_ROWS;
 import static org.apache.phoenix.query.QueryServices.MUTATE_BATCH_SIZE_ATTRIB;
+import static org.apache.phoenix.query.QueryServices.MUTATE_BATCH_SIZE_BYTES_ATTRIB;
 
 public abstract class GlobalIndexRegionScanner extends BaseRegionScanner {
 
     public static final String NUM_CONCURRENT_INDEX_VERIFY_THREADS_CONF_KEY = "index.verify.threads.max";
-    public static final int DEFAULT_CONCURRENT_INDEX_VERIFY_THREADS = 17;
-    public static final String INDEX_VERIFY_ROW_COUNTS_PER_TASK_CONF_KEY = "index.verify.threads.max";
+    public static final int DEFAULT_CONCURRENT_INDEX_VERIFY_THREADS = 16;
+    public static final String INDEX_VERIFY_ROW_COUNTS_PER_TASK_CONF_KEY = "index.verify.row.count.per.task";
     public static final int DEFAULT_INDEX_VERIFY_ROW_COUNTS_PER_TASK = 2048;
     public static final String NO_EXPECTED_MUTATION = "No expected mutation";
     public static final String ACTUAL_MUTATION_IS_NULL_OR_EMPTY = "actualMutationList is null or empty";
@@ -68,6 +75,10 @@ public abstract class GlobalIndexRegionScanner extends BaseRegionScanner {
     protected int rowCountPerTask;
     protected boolean hasMore;
     protected int maxBatchSize;
+    protected final long maxBatchSizeBytes;
+    protected final long blockingMemstoreSize;
+    protected final byte[] clientVersionBytes;
+    protected boolean useProto = true;
     protected byte[] indexMetaData;
     protected Scan scan;
     protected RegionScanner innerScanner;
@@ -75,16 +86,18 @@ public abstract class GlobalIndexRegionScanner extends BaseRegionScanner {
     protected IndexMaintainer indexMaintainer;
     protected Table indexHTable;
     protected TaskRunner pool;
-    protected TaskBatch<Boolean> tasks;
     protected String exceptionMessage;
     protected HTableFactory hTableFactory;
     protected int indexTableTTL;
     protected long maxLookBackInMills;
-    protected IndexToolVerificationResult verificationResult;
-    protected IndexVerificationResultRepository verificationResultRepository;
+    protected IndexToolVerificationResult verificationResult = null;
+    protected IndexVerificationResultRepository verificationResultRepository = null;
+    protected Map<byte[], NavigableSet<byte[]>> familyMap;
+    protected IndexTool.IndexVerifyType verifyType = IndexTool.IndexVerifyType.NONE;
+    protected boolean verify = false;
 
     public GlobalIndexRegionScanner(RegionScanner innerScanner, final Region region, final Scan scan,
-            final RegionCoprocessorEnvironment env) throws IOException {
+                                    final RegionCoprocessorEnvironment env) throws IOException {
         super(innerScanner);
         final Configuration config = env.getConfiguration();
         if (scan.getAttribute(BaseScannerRegionObserver.INDEX_REBUILD_PAGING) != null) {
@@ -92,6 +105,14 @@ public abstract class GlobalIndexRegionScanner extends BaseRegionScanner {
                     QueryServicesOptions.DEFAULT_INDEX_REBUILD_PAGE_SIZE_IN_ROWS);
         }
         maxBatchSize = config.getInt(MUTATE_BATCH_SIZE_ATTRIB, QueryServicesOptions.DEFAULT_MUTATE_BATCH_SIZE);
+        maxBatchSizeBytes = config.getLong(MUTATE_BATCH_SIZE_BYTES_ATTRIB,
+                QueryServicesOptions.DEFAULT_MUTATE_BATCH_SIZE_BYTES);
+        blockingMemstoreSize = UngroupedAggregateRegionObserver.getBlockingMemstoreSize(region, config);
+        clientVersionBytes = scan.getAttribute(BaseScannerRegionObserver.CLIENT_VERSION);
+        familyMap = scan.getFamilyMap();
+        if (familyMap.isEmpty()) {
+            familyMap = null;
+        }
         indexMetaData = scan.getAttribute(PhoenixIndexCodec.INDEX_PROTO_MD);
         if (indexMetaData == null) {
             indexMetaData = scan.getAttribute(PhoenixIndexCodec.INDEX_MD);
@@ -101,6 +122,13 @@ public abstract class GlobalIndexRegionScanner extends BaseRegionScanner {
         this.scan = scan;
         this.innerScanner = innerScanner;
         this.region = region;
+        byte[] valueBytes = scan.getAttribute(BaseScannerRegionObserver.INDEX_REBUILD_VERIFY_TYPE);
+        if (valueBytes != null) {
+            verifyType = IndexTool.IndexVerifyType.fromValue(valueBytes);
+            if (verifyType != IndexTool.IndexVerifyType.NONE) {
+                verify = true;
+            }
+        }
         // Create the following objects only for rebuilds by IndexTool
         hTableFactory = ServerUtil.getDelegateHTableFactory(env, ServerUtil.ConnectionType.INDEX_WRITER_CONNECTION);
         indexHTable = hTableFactory.getTable(new ImmutableBytesPtr(indexMaintainer.getIndexTableName()));
@@ -196,5 +224,18 @@ public abstract class GlobalIndexRegionScanner extends BaseRegionScanner {
             ts1 = getMaxTimestamp(del);
         }
         return (ts1 > ts2) ? ts1 : ts2;
+    }
+
+    protected boolean isColumnIncluded(Cell cell) {
+        byte[] family = CellUtil.cloneFamily(cell);
+        if (!familyMap.containsKey(family)) {
+            return false;
+        }
+        NavigableSet<byte[]> set = familyMap.get(family);
+        if (set == null || set.isEmpty()) {
+            return true;
+        }
+        byte[] qualifier = CellUtil.cloneQualifier(cell);
+        return set.contains(qualifier);
     }
 }
--- a/phoenix-core/src/main/java/org/apache/phoenix/coprocessor/IndexRebuildRegionScanner.java
+++ b/phoenix-core/src/main/java/org/apache/phoenix/coprocessor/IndexRebuildRegionScanner.java
@@ -99,16 +99,15 @@ public class IndexRebuildRegionScanner extends GlobalIndexRegionScanner {
     private boolean skipped = false;
     private boolean shouldVerifyCheckDone = false;
     final private RegionCoprocessorEnvironment env;
-    byte[][] endKeys;
+    private byte[][] endKeys;
+    private org.apache.hadoop.hbase.client.Connection connection;
     @VisibleForTesting
     public IndexRebuildRegionScanner(final RegionScanner innerScanner, final Region region, final Scan scan,
                               final RegionCoprocessorEnvironment env) throws IOException {
         super(innerScanner, region, scan, env);
         this.env = env;
-        try (org.apache.hadoop.hbase.client.Connection connection =
-                     HBaseFactoryProvider.getHConnectionFactory().createConnection(env.getConfiguration())) {
-            endKeys = connection.getRegionLocator(indexHTable.getName()).getEndKeys();
-        }
+        connection = HBaseFactoryProvider.getHConnectionFactory().createConnection(env.getConfiguration());
+        endKeys = connection.getRegionLocator(indexHTable.getName()).getEndKeys();
         indexRowKey = scan.getAttribute(BaseScannerRegionObserver.INDEX_ROW_KEY);
         if (indexRowKey != null) {
             setReturnCodeForSingleRowRebuild();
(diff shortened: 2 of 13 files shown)


phoenix-core/src/main/java/org/apache/phoenix/coprocessor/GlobalIndexRegionScanner.java
```diff
@@ -82,16 +94,30 @@
     protected long maxLookBackInMills;
     protected IndexToolVerificationResult verificationResult;
     protected IndexVerificationResultRepository verificationResultRepository;
+    protected UngroupedAggregateRegionObserver ungroupedAggregateRegionObserver;
```
gjacoby126, line 97: This superclass does not appear to use the ungroupedAggregateRegionObserver pointer, so it should belong to appropriate subclasses. (Also, we shouldn't be passing coprocs around -- see comments below.)

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/GlobalIndexRegionScanner.java
```diff
@@ -82,16 +94,30 @@
...
         super(innerScanner);
         final Configuration config = env.getConfiguration();
         if (scan.getAttribute(BaseScannerRegionObserver.INDEX_REBUILD_PAGING) != null) {
             pageSizeInRows = config.getLong(INDEX_REBUILD_PAGE_SIZE_IN_ROWS,
                     QueryServicesOptions.DEFAULT_INDEX_REBUILD_PAGE_SIZE_IN_ROWS);
         }
         maxBatchSize = config.getInt(MUTATE_BATCH_SIZE_ATTRIB, QueryServicesOptions.DEFAULT_MUTATE_BATCH_SIZE);
+        maxBatchSizeBytes = config.getLong(MUTATE_BATCH_SIZE_BYTES_ATTRIB,
+                QueryServicesOptions.DEFAULT_MUTATE_BATCH_SIZE_BYTES);
+        blockingMemstoreSize = UngroupedAggregateRegionObserver.getBlockingMemstoreSize(region, config);
```
gjacoby126, line 114: If this method is needed outside UARO, perhaps it should be in a static Util class somewhere.

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/IndexerRegionScanner.java
```diff
@@ -305,42 +320,96 @@ private void verifyIndex() throws IOException {
...
+        m.setAttribute(BaseScannerRegionObserver.REPLAY_WRITES,
+                BaseScannerRegionObserver.REPLAY_INDEX_REBUILD_WRITES);
+        m.setAttribute(BaseScannerRegionObserver.CLIENT_VERSION, clientVersionBytes);
+        // Since we're replaying existing mutations, it makes no sense to write them to the wal
+        m.setDurability(Durability.SKIP_WAL);
+    }
+
+    private byte[] commitIfReady(byte[] uuidValue, UngroupedAggregateRegionObserver.MutationList mutationList) throws IOException {
+        if (ServerUtil.readyToCommit(mutationList.size(), mutationList.byteSize(), maxBatchSize, maxBatchSizeBytes)) {
+            ungroupedAggregateRegionObserver.checkForRegionClosing();
```
gjacoby126, line 335: Should this also go in a util class along with the get blocking store size method?

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/IndexerRegionScanner.java
```diff
@@ -305,42 +320,96 @@ private void verifyIndex() throws IOException {
...
+            ungroupedAggregateRegionObserver.commitBatchWithRetries(region, mutationList, blockingMemstoreSize);
```
gjacoby126, line 336: Seems like it would be pretty straightforward to factor out the UARO.commitWith* functions into a dedicated Committer class that takes in a Region in the constructor and can be called from either UARO or here, to avoid the need for passing the UARO itself around. More in keeping with Single Responsibility Principle. 

Over time, we ideally should whittle UARO down to where it's just a dispatcher, catching Scan events and then calling appropriate objects to serve them based on the Scan attribute. That's out of scope here, but we can get a good start.

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/IndexerRegionScanner.java
```diff
@@ -305,42 +320,96 @@ private void verifyIndex() throws IOException {
...
-                        indexKeyToDataPutMap
-                                .put(getIndexRowKey(indexMaintainer, put), put);
+
                     }
                 } while (hasMore && rowCount < pageSizeInRows);
-                verifyIndex();
+                if (verify) {
+                    verifyIndex();
+                } else if (!mutations.isEmpty()) {
+                    ungroupedAggregateRegionObserver.checkForRegionClosing();
```
gjacoby126, line 398: I'm a bit confused -- I thought IndexerRegionScanner was only used for old-style indexes when doing ONLY-mode index verifications so we got the new correct verify features on the old indexes -- why are we committing things?

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/UngroupedAggregateRegionObserver.java
```diff
@@ -241,6 +243,16 @@ public void start(CoprocessorEnvironment e) throws IOException {
...
         indexWriteProps = new ReadOnlyProps(indexWriteConfig.iterator());
+        final RegionCoprocessorEnvironment env = (RegionCoprocessorEnvironment) e;
+        String serverName = env.getRegionServerServices().getServerName().getServerName();
+        DelegateRegionCoprocessorEnvironment indexWriterEnv = new
+                DelegateRegionCoprocessorEnvironment(env, ConnectionType.INDEX_WRITER_CONNECTION);
+        // setup the global index writer
+        this.globalIndexRebuildWriter = new IndexWriter(indexWriterEnv, serverName + "-globalIndexRebuildWriter", false);
+    }
+
+    public IndexWriter getGlobalIndexRebuildWriter() {
```
gjacoby126, line 254: if other classes need access to this, better to pass it to them directly rather than pass an UARO and make them call this getter.

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/IndexerRegionScanner.java
gjacoby126: OK, I see below that we're now using IndexerRegionScanner more broadly. Is this necessary -- just curious if it's worth doing a lot of refactoring and requisite testing on an index framework we know we want to deprecate soon. If it's actually harder to avoid the refactoring while touching UARO and the new indexing classes, that's fine.

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/index/IndexVerificationResultRepository.java
```diff
@@ -83,7 +83,7 @@
             "BeforeRebuildUnknownIndexRowCount";
     public final static byte[] BEFORE_REBUILD_UNKNOWN_INDEX_ROW_COUNT_BYTES = Bytes.toBytes(BEFORE_REBUILD_UNKNOWN_INDEX_ROW_COUNT);
     public final static String AFTER_REBUILD_VALID_INDEX_ROW_COUNT =
-        "AfterValidExpiredIndexRowCount";
+        "AfterRebuildValidIndexRowCount";
```
gjacoby126, line 86: Oops, thanks for the fix.

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/UngroupedAggregateRegionObserver.java
```diff
@@ -1088,9 +1100,17 @@ private RegionScanner rebuildIndices(final RegionScanner innerScanner, final Reg
             }
             innerScanner.close();
             RegionScanner scanner = region.getScanner(rawScan);
-            return new IndexRebuildRegionScanner(scanner, region, scan, env, this);
+            if (oldCoproc) {
+                return new IndexerRegionScanner(scanner, region, scan, env, this);
```
swaroopak, line 1104: For old coproc, we wanted to do only one version verification. Moving this inside not-raw block, wouldn't it change the strategy?

swaroopak: Also, what happens to the specific check for ONLY?

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/GlobalIndexRegionScanner.java
```diff
@@ -197,4 +230,17 @@ protected static long getMaxTimestamp(Pair<Put, Delete> pair) {
         }
         return (ts1 > ts2) ? ts1 : ts2;
     }
+
+    protected boolean isColumnIncluded(Cell cell) {
```
gokceni, line 234: Perhaps move to a utility class? CellUtil or some other?

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/GlobalIndexRegionScanner.java
kadirozde: Not currently

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/UngroupedAggregateRegionObserver.java
kadirozde: For the old indexing design, the verification other than ONLY is not supported. This was a problem before this PR too. This should be fixed in IndexTool. I will add the fix.

swaroopak: Okay, thanks! I think I misread the change at line 1088.

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/IndexerRegionScanner.java
kadirozde: This is only used currently by the old index design. I would not want to invest on it. Also, because of ungroupedAggregateRegionObserver.checkForRegionClosing(), it cannot go into a util call.

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/IndexerRegionScanner.java
kadirozde: This PR moved the rebuild path from IndexRegionObserver into UARO, more specifically IndexRebuildRegionObserver. This means that we cannot use IndexRebuildRegionObserver for building old indexes anymore. So I needed to move the responsibility of building old indexes to IndexerRegionObserver in order to continue using the index rebuild code in Indexer for old indexes.

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/UngroupedAggregateRegionObserver.java
kadirozde: We have to already pass the UARO coproc object directly. Are suggesting passing the writer directly too?

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/IndexerRegionScanner.java
kadirozde: Please see my related comments.

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/GlobalIndexRegionScanner.java
kadirozde: It is using it

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/GlobalIndexRegionScanner.java
kadirozde: I agree with you on this. There are several other such methods that should be moved to utility classes. I suggest opening a Jira for this and doing it there as I want to control the size of the changes here. It is already getting large

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/GlobalIndexRegionScanner.java
gjacoby126: Could you please point out where? I just expanded the file and I don't see it.

kadirozde: You are right that it is not used anymore here in this abstract class but both the classes that expand this class access this attribute. So it is a common attribute.

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/GlobalIndexRegionScanner.java
```diff
@@ -68,30 +75,44 @@
...
                     QueryServicesOptions.DEFAULT_INDEX_REBUILD_PAGE_SIZE_IN_ROWS);
         }
         maxBatchSize = config.getInt(MUTATE_BATCH_SIZE_ATTRIB, QueryServicesOptions.DEFAULT_MUTATE_BATCH_SIZE);
+        maxBatchSizeBytes = config.getLong(MUTATE_BATCH_SIZE_BYTES_ATTRIB,
+                QueryServicesOptions.DEFAULT_MUTATE_BATCH_SIZE_BYTES);
+        blockingMemstoreSize = UngroupedAggregateRegionObserver.getBlockingMemstoreSize(region, config);
+        clientVersionBytes = scan.getAttribute(BaseScannerRegionObserver.CLIENT_VERSION);
+        familyMap = scan.getFamilyMap();
+        if (familyMap.isEmpty()) {
+            familyMap = null;
```
gokceni, line 114: Any reason why we set this to null? isColumnIncluded should work if it is not null but empty right?

kadirozde: It is just a little optimization for run-time checking by mapping the empty map case to the null map case so we do not check both. This check is done for every cell scanned. This code is not introduced in this PR is just moved to here.

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/IndexRebuildRegionScanner.java
```diff
@@ -952,111 +890,89 @@ private void parallelizeIndexVerify(IndexToolVerificationResult.PhaseResult veri
...
-                if (entry.getValue().getFirst() != null) {
-                    mutations.add(entry.getValue().getFirst());
-                }
-                if (entry.getValue().getSecond() != null) {
-                    mutations.add(entry.getValue().getSecond());
-                }
+        if (verifyType != IndexTool.IndexVerifyType.ONLY) {
+            // Since we are building the index table, the region boundaries may change while we are building. Thus,
+            // we should update endKeys
+            try (org.apache.hadoop.hbase.client.Connection connection =
```
gjacoby126, line 955: Creating HBase connections is heavy-weight. Can we re-use one from somewhere, such as a Phoenix connection?

Also, since this code is duplicated in the constructor, can we factor out into a small helper function?

gjacoby126: What are the implications of endKeys being out-of-date? Is it a correctness problem, or a perf problem (as tasks would be crossing region boundaries)

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/IndexRebuildRegionScanner.java
```diff
@@ -952,111 +890,89 @@ private void parallelizeIndexVerify(IndexToolVerificationResult.PhaseResult veri
...
+    public static List<Map<byte[], List<Mutation>>> getPerTaskIndexKeyToMutationMaps(
+            TreeMap<byte[], List<Mutation>> indexKeyToMutationMap, byte[][] endKeys, int maxMapSize) {
+        List<Map<byte[], List<Mutation>>> mapList = new ArrayList<>();
+        int regionCount = endKeys.length;
+        int regionIndex = 0;
+        byte[] indexKey = indexKeyToMutationMap.firstKey();
+        while (regionIndex < regionCount) {
+            Map<byte[], List<Mutation>> perTaskIndexKeyToMutationMap;
+            // Find the region including indexKey
+            while (regionIndex < regionCount - 1 && Bytes.compareTo(indexKey, endKeys[regionIndex]) != 1) {
```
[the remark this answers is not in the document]
gjacoby126: (I'm a little confused that you seem to be taking great care to keep endKeys up to date in other places but not here where it's being relied on.)

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/IndexRebuildRegionScanner.java
```diff
@@ -91,90 +83,51 @@
...
-        }
-        familyMap = scan.getFamilyMap();
-        if (familyMap.isEmpty()) {
-            familyMap = null;
-        }
-        this.ungroupedAggregateRegionObserver = ungroupedAggregateRegionObserver;
+        this.env = env;
+        try (org.apache.hadoop.hbase.client.Connection connection =
+                     HBaseFactoryProvider.getHConnectionFactory().createConnection(env.getConfiguration())) {
+            endKeys = connection.getRegionLocator(indexHTable.getName()).getEndKeys();
```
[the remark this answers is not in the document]
gjacoby126: (I've been assuming that we'd get a different region rebuild scanner per mapper task (and hence usually per region), but parts of the logic elsewhere seem like they're trying to do the whole table. Is that just to account for splits?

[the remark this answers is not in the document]
kadirozde: For ONLY verifications, we do this once a table region so it should not be a concern. For the other types of index tool runs, we do this per page per table region as we rebuild index tables. The default size of the page is 32K index rows. So for every 32K index rows, we do this. Assuming creating a connection is expensive as you pointed out, I will use one connection for the table region instead of creating one for every page.

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/IndexRebuildRegionScanner.java
```diff
@@ -836,74 +747,122 @@ private void verifyIndexRows(List<KeyRange> keys,
...
+                                           final Map<byte[], List<Mutation>> indexKeyToMutationMap,
+                                           final Set<byte[]> mostRecentIndexRowKeys,
+                                           final IndexToolVerificationResult verificationResult) {
         tasks.add(new Task<Boolean>() {
             @Override
             public Boolean call() throws Exception {
                 try {
-                    if (Thread.currentThread().isInterrupted()) {
-                        exceptionMessage = "Pool closed, not attempting to verify index rows! " + indexHTable.getName();
+                    if (Thread.currentThread().isInterrupted() || env.getRegionServerServices().isAborted() ||
```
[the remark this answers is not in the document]
kadirozde: Indexers do the same. We need to change all then later.

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/IndexRebuildRegionScanner.java
kadirozde: As I mentioned above, I will fix this, i.e., I  will use one connection per table region.  I will also look into sharing the same connection within a region server or within a cluster. I think there is a way to do this. @abhishek-chouhan, let me know if you know already there is a way to do this.

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/IndexRebuildRegionScanner.java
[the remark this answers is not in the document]
kadirozde: As I described before, we get the end keys at the beginning of a page (every 32K index row verification/rebuild).  Our scans misses small number of rows when they operate on multiple table regions occasionally. Let me see if I can improve this further

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/IndexRebuildRegionScanner.java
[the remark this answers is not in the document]
kadirozde: We have a different data table region per mapper.  Mappers initiate a separate IndexRebuildRegionScanner object on their data table regions. Then in parallel these objects scan their data table regions and rebuild index table rows.  A object may rebuild rows from any index table region.

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/GlobalIndexRegionScanner.java
kadirozde: The attribute ungroupedAggregateRegionObserver has been removed from this class

PR #797, opened 2020-06-04
PHOENIX-5932 View Index rebuild results in surplus rows from other vi…

abhishek-chouhan:
…ew indexes

### DIFF
--- a/phoenix-core/src/main/java/org/apache/phoenix/coprocessor/UngroupedAggregateRegionObserver.java
+++ b/phoenix-core/src/main/java/org/apache/phoenix/coprocessor/UngroupedAggregateRegionObserver.java
@@ -60,6 +60,7 @@ import org.apache.hadoop.hbase.client.Put;
 import org.apache.hadoop.hbase.client.Scan;
 import org.apache.hadoop.hbase.coprocessor.ObserverContext;
 import org.apache.hadoop.hbase.coprocessor.RegionCoprocessorEnvironment;
+import org.apache.hadoop.hbase.filter.FirstKeyOnlyFilter;
 import org.apache.hadoop.hbase.io.ImmutableBytesWritable;
 import org.apache.hadoop.hbase.ipc.RpcControllerFactory;
 import org.apache.hadoop.hbase.ipc.controller.InterRegionServerIndexRpcControllerFactory;
@@ -87,6 +88,7 @@ import org.apache.phoenix.expression.ExpressionType;
 import org.apache.phoenix.expression.aggregator.Aggregator;
 import org.apache.phoenix.expression.aggregator.Aggregators;
 import org.apache.phoenix.expression.aggregator.ServerAggregators;
+import org.apache.phoenix.filter.AllVersionsIndexRebuildFilter;
 import org.apache.phoenix.hbase.index.Indexer;
 import org.apache.phoenix.hbase.index.ValueGetter;
 import org.apache.phoenix.hbase.index.covered.update.ColumnReference;
@@ -1081,7 +1083,15 @@ public class UngroupedAggregateRegionObserver extends BaseScannerRegionObserver 
             rawScan.setRaw(true);
             rawScan.setMaxVersions();
             rawScan.getFamilyMap().clear();
-            rawScan.setFilter(null);
+            // For rebuilds we use count (*) as query for regular tables which ends up setting the FKOF on scan
+            // This filter doesn't give us all columns and skips to the next row as soon as it finds 1 col
+            // For rebuilds we need all columns and all versions
+            if (scan.getFilter() instanceof FirstKeyOnlyFilter) {
+                rawScan.setFilter(null);
+            } else if (scan.getFilter() != null) {
+                // Override the filter so that we get all versions
+                rawScan.setFilter(new AllVersionsIndexRebuildFilter(scan.getFilter()));
+            }
             rawScan.setCacheBlocks(false);
             for (byte[] family : scan.getFamilyMap().keySet()) {
                 rawScan.addFamily(family);
--- a/phoenix-core/src/main/java/org/apache/phoenix/filter/AllVersionsIndexRebuildFilter.java
+++ b/phoenix-core/src/main/java/org/apache/phoenix/filter/AllVersionsIndexRebuildFilter.java
@@ -1,0 +1,45 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.phoenix.filter;
+
+import java.io.IOException;
+
+import org.apache.hadoop.hbase.Cell;
+import org.apache.hadoop.hbase.filter.Filter;
+
+/**
+ * This filter overrides the behavior of delegate so that we do not jump to the next
+ * column as soon as we find a value for a column but rather include all versions which is
+ * needed for rebuilds.
+ */
+public class AllVersionsIndexRebuildFilter extends DelegateFilter {
+
+    public AllVersionsIndexRebuildFilter(Filter originalFilter) {
+        super(originalFilter);
+    }
+
+    @Override
+    public ReturnCode filterKeyValue(Cell v) throws IOException {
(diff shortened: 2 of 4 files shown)


phoenix-core/src/it/java/org/apache/phoenix/end2end/IndexToolIT.java
```diff
@@ -208,6 +209,195 @@ protected static void setEveryNthRowWithNull(int nrows, int nthRowNull, Prepared
...
+            conn.commit();
+            // Rebuild using index tool
+            runIndexTool(directApi, useSnapshot, schemaName, view1Name, indexTableName);
+            ResultSet rs =
+                    conn.createStatement()
+                            .executeQuery("SELECT COUNT(*) FROM " + indexTableFullName);
+            rs.next();
+            assertEquals(2, rs.getInt(1));
+            try (org.apache.hadoop.hbase.client.Connection hcon =
+                    ConnectionFactory.createConnection(config)) {
```
gjacoby126, line 378: You can factor out the cell counting into its own helper function to avoid duplication between the 2 tests. (TestUtil.getRawCellCount may also be useful if you can extend it to also keep track of what Cell types are scanned.)

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/UngroupedAggregateRegionObserver.java
```diff
@@ -1081,7 +1083,11 @@ private RegionScanner rebuildIndices(final RegionScanner innerScanner, final Reg
             rawScan.setRaw(true);
             rawScan.setMaxVersions();
             rawScan.getFamilyMap().clear();
-            rawScan.setFilter(null);
+            if (scan.getFilter() instanceof FirstKeyOnlyFilter) {
```
gjacoby126, line 1086: Please add a comment to explain why FirstKeyOnlyFilter is a special case. If the rebuild index scan is explicitly asking for only the first key, why do we avoid using the AllVersions filter which also only gives the first key? 

And if there is a reason not to use the FirstKeyOnlyFilter, are we still OK with using the AllVersionsIndexRebuildFilter if the Scan's filter it will delegate to is a composite filter which contains a FirstKeyOnlyFilter?

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/UngroupedAggregateRegionObserver.java
```diff
@@ -1081,7 +1083,11 @@ private RegionScanner rebuildIndices(final RegionScanner innerScanner, final Reg
...
+                rawScan.setFilter(null);
+            } else if (scan.getFilter() != null) {
+                rawScan.setFilter(new AllVersionsIndexRebuildFilter(scan.getFilter()));
+            }
```
gjacoby126, line 1090: Do we need to do any special filter logic down at line 1099 in the else block if the Scan was raw in the first place?

phoenix-core/src/main/java/org/apache/phoenix/filter/AllVersionsIndexRebuildFilter.java
```diff
@@ -0,0 +1,40 @@
...
+
+    public AllVersionsIndexRebuildFilter(Filter originalFilter) {
+        super(originalFilter);
+    }
+
+    @Override
+    public ReturnCode filterKeyValue(Cell v) throws IOException {
+        ReturnCode delegateCode = super.filterKeyValue(v);
+        if (delegateCode == ReturnCode.INCLUDE_AND_NEXT_COL) {
+            return ReturnCode.INCLUDE;
```
gjacoby126, line 35: This is simulating the effects of a FirstKeyOnlyFilter? Comment would be good

phoenix-core/src/it/java/org/apache/phoenix/end2end/IndexToolIT.java
```diff
@@ -208,6 +209,195 @@ protected static void setEveryNthRowWithNull(int nrows, int nthRowNull, Prepared
         }
     }
 
+    @Test
+    public void testUpdatableViewIndex() throws Exception {
```
swaroopak, line 213: Please move this to IndexToolForNonTxGlobalIndexIT

phoenix-core/src/it/java/org/apache/phoenix/end2end/IndexToolIT.java
```diff
@@ -208,6 +209,195 @@ protected static void setEveryNthRowWithNull(int nrows, int nthRowNull, Prepared
...
+                    }
+                }
+                assertEquals(4, numPuts);
+                assertEquals(2, numDeletes);
+            }
+        }
+    }
+
+    @Test
+    public void testUpdatableViewIndex2() throws Exception {
```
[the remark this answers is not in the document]
abhishek-chouhan: Tried a more descriptive name in the latest commit :)

phoenix-core/src/it/java/org/apache/phoenix/end2end/IndexToolIT.java
```diff
@@ -208,6 +209,195 @@ protected static void setEveryNthRowWithNull(int nrows, int nthRowNull, Prepared
...
+        String view1Name = generateUniqueName();
+        String view1FullName = SchemaUtil.getTableName(schemaName, view1Name);
+        String view2Name = generateUniqueName();
+        String view2FullName = SchemaUtil.getTableName(schemaName, view2Name);
+        String indexTableName = generateUniqueName();
+        String indexTableFullName = SchemaUtil.getTableName(schemaName, indexTableName);
+        Properties props = PropertiesUtil.deepCopy(TEST_PROPERTIES);
+
+        try (Connection conn = DriverManager.getConnection(getUrl(), props)) {
+            // Create Table and Views
```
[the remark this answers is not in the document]
abhishek-chouhan: Done. Yes, the point of 2 tests is to test out two different filters that end up being used. One test has view on a non-leading part of pk, other one has view on a non pk column.

phoenix-core/src/it/java/org/apache/phoenix/end2end/IndexToolIT.java
abhishek-chouhan: Done

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/UngroupedAggregateRegionObserver.java
abhishek-chouhan: Allversions filter does not only give the first key value, its purpose is to make sure all versions of a column are returned(when matched by underlying supplied filter), instead of just one. Usually the filters used in normal queries(which also end up being used for rebuild since we use select count(*)) returns only 1 version of a column, in rebuild we want to return all versions hence this.

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/UngroupedAggregateRegionObserver.java
abhishek-chouhan: AFAIK we get raw scan here in case of old design and partial rebuild (Correct me if i'm wrong here @kadirozde ). I didn't want to mess with the old design and hence only made the changes for new.

phoenix-core/src/it/java/org/apache/phoenix/end2end/IndexToolIT.java
abhishek-chouhan: Done

phoenix-core/src/main/java/org/apache/phoenix/filter/AllVersionsIndexRebuildFilter.java
```diff
@@ -0,0 +1,45 @@
...
+public class AllVersionsIndexRebuildFilter extends DelegateFilter {
+
+    public AllVersionsIndexRebuildFilter(Filter originalFilter) {
+        super(originalFilter);
+    }
+
+    @Override
+    public ReturnCode filterKeyValue(Cell v) throws IOException {
+        ReturnCode delegateCode = super.filterKeyValue(v);
+        if (delegateCode == ReturnCode.INCLUDE_AND_NEXT_COL) {
```
gokceni, line 39: Could you add a comment here why we convert it to INCLUDE? Why are we not happy with NEXT_COL?

abhishek-chouhan: @gokceni NEXT_COL skips this column and goes to the next one. What we want to do is when the underlying filter says yes to a column, we want to say yes too, but instead of jumping to the next col since we got a value, we want to get all versions.

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/UngroupedAggregateRegionObserver.java
kadirozde: Yes, we get raw scan only for the old design partial rebuilds (i.e., auto-rebuilds).

phoenix-core/src/main/java/org/apache/phoenix/filter/AllVersionsIndexRebuildFilter.java
swaroopak: Yes, I agree, comment in the code would be helpful here. @abhishek-chouhan

PR #798, opened 2020-06-06
PHOENIX-5942 - ParameterizedIndexUpgradeToolIT is too slow

gjacoby126:


gjacoby126:
First pass -- this gets the running time down from ~30min to ~16.5min by giving each table a unique name rather than creating/dropping all tables for each test. Will take a second look to see if I can get it down lower. Perhaps we could change parameter order so that the same minicluster can be reused more, or just use fewer tables for some of the corner case / error handling tests.

phoenix-core/src/it/java/org/apache/phoenix/end2end/ParameterizedIndexUpgradeToolIT.java
```diff
@@ -169,59 +167,144 @@ private void setClusterProperties() {
...
+            TABLE_LIST_NAMESPACE[1] = mockTableTwo.replace(".", ":");
+            TABLE_LIST_NAMESPACE[2] = mockTableThree.replace(".", ":");
+            TABLE_LIST_NAMESPACE[3] = multiTenantTable.replace(".", ":");
+            conn.createStatement().execute("CREATE SCHEMA IF NOT EXISTS TEST");
+            conn.createStatement().execute("CREATE SCHEMA IF NOT EXISTS TEST1");
         }
-        conn.createStatement().execute("CREATE TABLE TEST.MOCK1 (id bigint NOT NULL "
+
+
+
```
swaroopak, line 190: nit: extra lines

gjacoby126:
Pushed up my second pass. Now it's down to 12min 19s locally. Will do a HadoopQA run to make sure that this is good enough for Jenkins.

gjacoby126:
@swaroopak - yes, this passes locally. The presence of extra tables from past tests shouldn't affect the IUT, so long as you don't try to upgrade/downgrade them. That's why there's so much bookkeeping code added around swapping out the "active" table / index names the current test is using.

gjacoby126:
Thanks for the review, @swaroopak and @gokceni

PR #799, opened 2020-06-09
PHOENIX-5933 Rename VIEW_TTL property to be PHOENIX_TTL

jpisaac:
As there is more interest in extending the TTL feature to all Phoenix entities (tables and views)

More specifically separate the usage of Phoenix application-level TTL (row-based) and HBase level TTL (Store/Column Family-based).

This PR renames the VIEW_TTL property to be PHOENIX_TTL

### DIFF
--- a/phoenix-core/src/main/java/org/apache/phoenix/coprocessor/IndexRebuildRegionScanner.java
+++ b/phoenix-core/src/main/java/org/apache/phoenix/coprocessor/IndexRebuildRegionScanner.java
@@ -1373,8 +1373,9 @@ public class IndexRebuildRegionScanner extends GlobalIndexRegionScanner {
                     }
                 }
             }
-        } catch (IOException e) {
-            LOGGER.error("IOException during rebuilding: " + Throwables.getStackTraceAsString(e));
+        } catch (Throwable e) {
+            LOGGER.error("Exception in IndexRebuildRegionScanner for region "
+                    + region.getRegionInfo().getRegionNameAsString(), e);
             throw e;
         } finally {
             region.closeRegionOperation();
--- a/phoenix-core/src/main/java/org/apache/phoenix/index/PhoenixIndexFailurePolicy.java
+++ b/phoenix-core/src/main/java/org/apache/phoenix/index/PhoenixIndexFailurePolicy.java
@@ -493,7 +493,7 @@ public class PhoenixIndexFailurePolicy extends DelegateIndexFailurePolicy {
                 return;
             } catch (IOException e) {
                 SQLException inferredE = ServerUtil.parseLocalOrRemoteServerException(e);
-                if (inferredE == null || inferredE.getErrorCode() != SQLExceptionCode.INDEX_WRITE_FAILURE.getErrorCode()) {
+                if (inferredE != null && inferredE.getErrorCode() != SQLExceptionCode.INDEX_WRITE_FAILURE.getErrorCode()) {
                     // If this call is from phoenix client, we also need to check if SQLException
                     // error is INDEX_METADATA_NOT_FOUND or not
                     // if it's not an INDEX_METADATA_NOT_FOUND, throw exception,


jpisaac:
@ChinmaySKulkarni @yanxinyi @gjacoby126 Please review.

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/MetaDataEndpointImpl.java
```diff
@@ -19,63 +19,7 @@
...
-import static org.apache.phoenix.jdbc.PhoenixDatabaseMetaData.USE_STATS_FOR_PARALLELIZATION_BYTES;
-import static org.apache.phoenix.jdbc.PhoenixDatabaseMetaData.VIEW_CONSTANT_BYTES;
-import static org.apache.phoenix.jdbc.PhoenixDatabaseMetaData.VIEW_INDEX_ID_BYTES;
-import static org.apache.phoenix.jdbc.PhoenixDatabaseMetaData.VIEW_INDEX_ID_DATA_TYPE_BYTES;
-import static org.apache.phoenix.jdbc.PhoenixDatabaseMetaData.VIEW_STATEMENT_BYTES;
-import static org.apache.phoenix.jdbc.PhoenixDatabaseMetaData.VIEW_TTL_BYTES;
-import static org.apache.phoenix.jdbc.PhoenixDatabaseMetaData.VIEW_TTL_HWM_BYTES;
-import static org.apache.phoenix.jdbc.PhoenixDatabaseMetaData.VIEW_TTL_NOT_DEFINED;
-import static org.apache.phoenix.jdbc.PhoenixDatabaseMetaData.VIEW_TYPE_BYTES;
+import static org.apache.phoenix.jdbc.PhoenixDatabaseMetaData.*;
```
yanxinyi, line 22: please do not use import *

phoenix-protocol/src/main/PTable.proto
```diff
@@ -106,9 +106,9 @@ message PTable {
   optional int32 viewIndexIdType = 39 [default = 5];
   optional bool viewModifiedUpdateCacheFrequency = 40;
   optional bool viewModifiedUseStatsForParallelization = 41;
-  optional int64 viewTTL = 42;
-  optional int64 viewTTLHighWaterMark = 43;
-  optional bool viewModifiedViewTTL = 44;
+  optional int64 phoenixTTL = 42;
+  optional int64 phoenixTTLHighWaterMark = 43;
+  optional bool viewModifiedPhoenixTTL = 44;
```
yanxinyi, line 111: do we still want to keep the naming of viewModifiedPhoenixTTL? do view or general phoenix TTL matter here?

phoenix-core/src/main/java/org/apache/phoenix/util/ViewUtil.java
```diff
@@ -635,11 +639,52 @@ public static void addTagsToPutsForViewAlteredProperties(List<Mutation> tableMet
...
+
+        String viewTTLFilterClause = context.getConnection().getSCN() == null ?
+                String.format(
+                        "((TO_NUMBER(CURRENT_TIME()) - TO_NUMBER(phoenix_row_timestamp())) < %d)",
+                        table.getPhoenixTTL()) :
+                String.format(
+                "((%d - TO_NUMBER(phoenix_row_timestamp())) < %d)",
+                        context.getConnection().getSCN(), table.getPhoenixTTL());
+
+        if (table.getViewStatement() != null) {
```
yanxinyi, line 668: technically we are allowing the user to create a view with `SELECT * FROM BASE_TABLE` without anywhere clause. Do we want to count this case? 
Personally, I don't think anyone intentional to do this, but I will let @ChinmaySKulkarni to confirm we don't support this.

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/generated/PTableProtos.java
```diff
@@ -3725,35 +3725,35 @@ public Builder clearPGuidePosts() {
      */
```
yanxinyi, line 3725: If I remember it correctly, this is an autogenerated file?

phoenix-core/src/it/java/org/apache/phoenix/end2end/ViewUtilIT.java
```diff
@@ -333,4 +356,120 @@ public void testFindAllRelativesForTenantConnection() throws Exception {
...
+                        + "AS SELECT * FROM " + multiTenantTableName + " WHERE KP = '001' VIEW_TTL=30";
+
+        String viewDDL = "CREATE VIEW IF NOT EXISTS %s (V1 BIGINT, V2 BIGINT) AS SELECT * FROM %s VIEW_TTL=30";
+        String
+                viewIndexDDL =
+                "CREATE INDEX IF NOT EXISTS " + tenantViewIndex + " ON " + tenantViewOnGlobalView
+                        + "(NUM DESC) INCLUDE (ID)";
+
+        String
+                customObjectViewDDL =
```
yanxinyi, line 397: maybe rename here?

phoenix-core/src/it/java/org/apache/phoenix/end2end/ViewUtilIT.java
```diff
@@ -333,4 +356,120 @@ public void testFindAllRelativesForTenantConnection() throws Exception {
...
+                try (Statement stmt = tenantConn.createStatement()) {
+                    PhoenixStatement pstmt = stmt.unwrap(PhoenixStatement.class);
+                    QueryPlan plan = pstmt.compileQuery(sql);
+                    validatePhoenixRowTimestampParseNodeExists(plan);
+                }
+            }
+        }
+
+    }
+
```
yanxinyi, line 449: can we add one more test for view without ttl but calling `getViewWhereWithPhoenixTTL` method to validate the right behavior?

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/generated/PTableProtos.java
jpisaac: Yes.

phoenix-core/src/main/java/org/apache/phoenix/util/ViewUtil.java
jpisaac: The getViewStatement will be "not null" in the case when "SELECT * FROM BASE_TABLE" and the processing should proceed.

phoenix-protocol/src/main/PTable.proto
jpisaac: As explained offline this holds the info when view modified the PHOENIX_TTL attribute, so the naming is appropriate.

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/MetaDataEndpointImpl.java
jpisaac: Will remove/modify with appropriate imports.

jpisaac:
@yanxinyi @ChinmaySKulkarni Just housekeeping - squashed PHOENIX-5993 changes and removed the PHOENIX-5821 changes, will create a patch out this PR.

PR #801, opened 2020-06-11
PHOENIX-5951 - Index rebuild output logging for past-max-lookback row…

gjacoby126:
…s should be configurable

### DIFF
--- a/phoenix-core/src/it/java/org/apache/phoenix/end2end/IndexToolForNonTxGlobalIndexIT.java
+++ b/phoenix-core/src/it/java/org/apache/phoenix/end2end/IndexToolForNonTxGlobalIndexIT.java
@@ -41,6 +41,7 @@ import org.apache.hadoop.hbase.regionserver.RegionScanner;
 import org.apache.hadoop.hbase.regionserver.ScanInfoUtil;
 import org.apache.hadoop.hbase.util.Bytes;
 import org.apache.hadoop.hbase.util.Pair;
+import org.apache.hadoop.mapreduce.Counters;
 import org.apache.phoenix.coprocessor.IndexRebuildRegionScanner;
 import org.apache.phoenix.jdbc.PhoenixConnection;
 import org.apache.phoenix.mapreduce.index.IndexTool;
@@ -51,11 +52,9 @@ import org.apache.phoenix.query.ConnectionQueryServices;
 import org.apache.phoenix.query.QueryServices;
 import org.apache.phoenix.query.QueryServicesOptions;
 import org.apache.phoenix.schema.PIndexState;
-import org.apache.phoenix.schema.PTable;
 import org.apache.phoenix.util.EnvironmentEdgeManager;
 import org.apache.phoenix.util.IndexScrutiny;
 import org.apache.phoenix.util.ManualEnvironmentEdge;
-import org.apache.phoenix.util.PhoenixRuntime;
 import org.apache.phoenix.util.PropertiesUtil;
 import org.apache.phoenix.util.ReadOnlyProps;
 import org.apache.phoenix.util.SchemaUtil;
@@ -118,6 +117,7 @@ import static org.junit.Assert.assertTrue;
 @RunWith(Parameterized.class)
 public class IndexToolForNonTxGlobalIndexIT extends BaseUniqueNamesOwnClusterIT {
 
+    public static final int MAX_LOOKBACK_AGE = 3600;
     private final String tableDDLOptions;
     private boolean directApi = true;
     private boolean useSnapshot = false;
@@ -150,7 +150,7 @@ public class IndexToolForNonTxGlobalIndexIT extends BaseUniqueNamesOwnClusterIT 
         serverProps.put(QueryServices.EXTRA_JDBC_ARGUMENTS_ATTRIB,
                 QueryServicesOptions.DEFAULT_EXTRA_JDBC_ARGUMENTS);
         serverProps.put(QueryServices.INDEX_REBUILD_PAGE_SIZE_IN_ROWS, Long.toString(8));
-        serverProps.put(ScanInfoUtil.PHOENIX_MAX_LOOKBACK_AGE_CONF_KEY, Long.toString(3600));
+        serverProps.put(ScanInfoUtil.PHOENIX_MAX_LOOKBACK_AGE_CONF_KEY, Long.toString(MAX_LOOKBACK_AGE));
         serverProps.put(QueryServices.TASK_HANDLING_INTERVAL_MS_ATTRIB,
             Long.toString(Long.MAX_VALUE));
         serverProps.put(QueryServices.TASK_HANDLING_INITIAL_DELAY_MS_ATTRIB,
@@ -648,8 +648,12 @@ public class IndexToolForNonTxGlobalIndexIT extends BaseUniqueNamesOwnClusterIT 
             expectedStatus.add(RUN_STATUS_EXECUTED);
             expectedStatus.add(RUN_STATUS_EXECUTED);
             expectedStatus.add(RUN_STATUS_EXECUTED);
-
-            verifyRunStatusFromResultTable(conn, scn, indexTableFullName, 3, expectedStatus);
+            try {
+                verifyRunStatusFromResultTable(conn, scn, indexTableFullName, 3, expectedStatus);
+            } catch (AssertionError ae) {
+                TestUtil.dumpTable(conn, TableName.valueOf(RESULT_TABLE_NAME));
+                throw ae;
+            }
 
             deleteOneRowFromResultTable(conn, scn, indexTableFullName);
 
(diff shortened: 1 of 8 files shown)


phoenix-core/src/it/java/org/apache/phoenix/end2end/index/IndexVerificationOutputRepositoryIT.java
```diff
@@ -53,6 +54,7 @@
 import java.util.Map;
 
 import static org.apache.phoenix.coprocessor.MetaDataProtocol.DEFAULT_LOG_TTL;
+import static org.apache.phoenix.mapreduce.index.IndexVerificationOutputRepository.IndexVerificationErrorType.*;
```
gokceni, line 57: Let's remove the *

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/IndexRebuildRegionScanner.java
```diff
@@ -98,6 +103,9 @@
 
     private static final Logger LOGGER = LoggerFactory.getLogger(IndexRebuildRegionScanner.class);
 
+    public static final String PHOENIX_LOG_BEYOND_MAX_LOOKBACK_INVALID_ROWS_CONF_KEY =
+        "phoenix.log.beyond.max.lookback.age.invalid.rows";
+    public static final boolean DEFAULT_PHOENIX_LOG_BEYOND_MAX_LOOKBACK_INVALID_ROWS = true;
```
gokceni, line 108: Should this be false? I feel like "beyond max lookback" is kind of analogy to debug log. Why don't we make it false?

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/IndexRebuildRegionScanner.java
```diff
@@ -98,6 +103,9 @@
 
     private static final Logger LOGGER = LoggerFactory.getLogger(IndexRebuildRegionScanner.class);
 
+    public static final String PHOENIX_LOG_BEYOND_MAX_LOOKBACK_INVALID_ROWS_CONF_KEY =
```
swaroopak, line 106: invalid and missing, right?

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/index/IndexVerificationOutputRepository.java
```diff
@@ -85,6 +88,15 @@
     public static final byte[] PHASE_BEFORE_VALUE = Bytes.toBytes("BEFORE");
     public static final byte[] PHASE_AFTER_VALUE = Bytes.toBytes("AFTER");
 
+
+    public enum IndexVerificationErrorType {
+        INVALID_ROW,
```
swaroopak, line 93: shall we be specific with INVALID_INDEX_ROW and same for others?

gjacoby126: Since all of them refer to index rows, and Index is in the name of the enum, isn't that implicit?

phoenix-core/src/it/java/org/apache/phoenix/end2end/index/IndexVerificationOutputRepositoryIT.java
gjacoby126: Done

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/IndexRebuildRegionScanner.java
gjacoby126: Done

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/IndexRebuildRegionScanner.java
gjacoby126: Switched to "ERROR" to encompass both INVALID and MISSING.

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/IndexRebuildRegionScanner.java
```diff
@@ -152,6 +158,9 @@ public IndexRebuildRegionScanner(final RegionScanner innerScanner, final Region
             verifyType = IndexTool.IndexVerifyType.fromValue(valueBytes);
             if (verifyType != IndexTool.IndexVerifyType.NONE) {
                 verify = true;
+                boolean shouldLogBeyondMaxLookbackInvalidRows =
+                    config.getBoolean(PHOENIX_LOG_BEYOND_MAX_LOOKBACK_ERROR_ROWS_CONF_KEY,
```
swaroopak, line 162: Can this property be modified via -D option from the client-side? Good to have a test for that.

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/IndexRebuildRegionScanner.java
```diff
@@ -91,13 +95,15 @@
...
 
-import com.google.common.base.Throwables;
 import com.google.common.collect.Maps;
 
 public class IndexRebuildRegionScanner extends GlobalIndexRegionScanner {
 
     private static final Logger LOGGER = LoggerFactory.getLogger(IndexRebuildRegionScanner.class);
 
+    public static final String PHOENIX_LOG_BEYOND_MAX_LOOKBACK_ERROR_ROWS_CONF_KEY =
+        "phoenix.log.beyond.max.lookback.age.error.rows";
```
priyankporwal, line 105: Can we please have this config be named consistently like the previous one? 
"phoenix.mr.index.IndexDisableLoggingType=BEFORE" .. It would make it easier to find and put those together whenever they need non-default values.

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/IndexRebuildRegionScanner.java
gjacoby126: So far I have done nothing special to allow it be overridden from client-side. If the IndexTool is already taking all conf parameters from the client and using them on the server-side, then that's the case for this as well, and if it's not, it's not. (At a cursory look it appears that it's only certain flags which the InputFormat explicitly passes as scan attributes to the coproc?)

My original intention was that this would be a cluster-wide setting. I could change it to be overridable. 

@swaroopak - if this were overridable from the client-side, what kind of test would be sufficient? Given the long running time of our index tool tests already, I'm reluctant to add yet-another-end-to-end test just to check that a flag is carried through, but I'm not sure what other options there are to actually verify that each of the several layers passes the config through properly.

swaroopak: It is important that we have this property overridable for each IndexTool run. You could modify the existing/the one added in this PR run of IndexTool to make sure the property changes from the client-side are reflecting properly.

gjacoby126: @swaroopak - Added a test in latest commit.

gjacoby126:
@kadirozde - please take a look at the change to the IndexRebuildRegionScanner when you have a moment. 

When testing the max lookback counters and output logging, I found that ONLY (and I assume BEFORE, AFTER and BOTH) rebuilds were being treated as NONEs because the IndexRebuildRegionScanner wasn't setting verifyType. I'm thinking this got introduced in PHOENIX-5928, but I'm confused why the IndexToolIT.testSecondaryIndex:308 check for the valid counter in a BEFORE rebuild didn't catch this.

kadirozde:
> @kadirozde - please take a look at the change to the IndexRebuildRegionScanner when you have a moment.
> 
> When testing the max lookback counters and output logging, I found that ONLY (and I assume BEFORE, AFTER and BOTH) rebuilds were being treated as NONEs because the IndexRebuildRegionScanner wasn't setting verifyType. I'm thinking this got introduced in PHOENIX-5928, but I'm confused why the IndexToolIT.testSecondaryIndex:308 check for the valid counter in a BEFORE rebuild didn't catch this.

I see that verifyType is set in the constructor. Please make sure to check the constructor in both IndexRebuildRegionScanner and its super class GlobalIndexRegionScanner.

gjacoby126:
I see that verifyType is set in the superclass constructor, if the Scan has the right attribute, as you say. But without manually setting it in the IndexRebuildRegionScanner, I was getting job counters with scan=2, rebuilt=2, and all other counters 0, even though it was an ONLY rebuild of an never-built index of a data table with 2 rows. Setting verifyType let the MISSING or BEYOND_MAX_LOOKBACK_AGE_MISSING counter be 2 as they should, depending on the age of the rows. 

Could be a test bug on my part, or could be there's some code path where the Scan attribute isn't being set properly. I'll look further in the morning.

gjacoby126:
@kadirozde - figured out the problem. A previous rebase on this feature branch had re-introduced a private verifyType in IndexRebuildRegionScanner and that was masking the protected verifyType in the superclass. My test passes upon removing the extra verifyType. 

Will get a new commit up and also submit to HadoopQA

gjacoby126:
OK, @swaroopak , fixed the test bug I mentioned to you and the real bug that revealed. It also now uses Configuration to verify.

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/IndexRebuildRegionScanner.java
```diff
@@ -96,6 +102,11 @@
 
     private static final Logger LOGGER = LoggerFactory.getLogger(IndexRebuildRegionScanner.class);
 
+    public static final String PHOENIX_INDEX_MR_LOG_BEYOND_MAX_LOOKBACK_ERRORS =
+        "phoenix.index.mr.log.beyond.max.lookback.errors";
+    public static final boolean DEFAULT_PHOENIX_INDEX_MR_LOG_BEYOND_MAX_LOOKBACK_ERRORS = false;
+    private boolean useProto = true;
+    private byte[] indexRowKey;
```
swaroopak, line 109: nit: looks like this made its way in the rebase?

phoenix-core/src/it/java/org/apache/phoenix/end2end/IndexToolForNonTxGlobalIndexIT.java
```diff
@@ -1006,6 +1011,72 @@ public void testDisableOutputLogging() throws Exception {
...
+            String stmtString2 =
+                String.format(
+                    "CREATE INDEX %s ON %s (LPAD(UPPER(NAME, 'en_US'),8,'x')||'_xyz') ASYNC ",
+                    indexTableName, dataTableFullName);
+            conn.createStatement().execute(stmtString2);
+            conn.commit();
+            injectEdge.incrementValue(MAX_LOOKBACK_AGE * 1000);
+            deleteAllRows(conn, TableName.valueOf(IndexVerificationOutputRepository.OUTPUT_TABLE_NAME));
+            getUtility().getConfiguration().
+                set(IndexRebuildRegionScanner.INDEX_VERIFY_ROW_COUNTS_PER_TASK_CONF_KEY, "true");
```
kadirozde, line 1060: This is not the right conf key to set here

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/IndexRebuildRegionScanner.java
kadirozde: indexRowKey and useProto are not used

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/IndexRebuildRegionScanner.java
```diff
@@ -1223,7 +1254,7 @@ public int prepareIndexMutations(Put put, Delete del, Map<byte[], List<Mutation>
                                      Set<byte[]> mostRecentIndexRowKeys) throws IOException {
         List<Mutation> indexMutations = prepareIndexMutationsForRebuild(indexMaintainer, put, del);
         boolean mostRecentDone = false;
-        // Do not populate mostRecentIndexRowKeys when verifyType != IndexTool.IndexVerifyType.ONLY
+        // Do not populate mostRecentIndexRowKeys when verifyType is NONE or AFTER
```
kadirozde, line 1257: Thanks for correcting the comment

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/PhoenixServerBuildIndexInputFormat.java
```diff
@@ -108,6 +109,12 @@ protected QueryPlan getQueryPlan(final JobContext context, final Configuration c
                 scan.setAttribute(BaseScannerRegionObserver.INDEX_RETRY_VERIFY, Bytes.toBytes(lastVerifyTimeValue));
                 scan.setAttribute(BaseScannerRegionObserver.INDEX_REBUILD_DISABLE_LOGGING_VERIFY_TYPE,
                     getDisableLoggingVerifyType(configuration).toBytes());
+                String shouldLogMaxLookbackOutput =
+                    configuration.get(IndexRebuildRegionScanner.INDEX_VERIFY_ROW_COUNTS_PER_TASK_CONF_KEY);
```
kadirozde, line 113: PHOENIX_INDEX_MR_LOG_BEYOND_MAX_LOOKBACK_ERRORS rather than INDEX_VERIFY_ROW_COUNTS_PER_TASK_CONF_KEY

gjacoby126: Fixed

phoenix-core/src/it/java/org/apache/phoenix/end2end/IndexToolForNonTxGlobalIndexIT.java
gjacoby126: Fixed

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/IndexRebuildRegionScanner.java
gjacoby126: Fixed

gjacoby126:
@kadirozde - I believe I have addressed all your comments. (Thanks for pointing out the auto-complete errors where the wrong conf key was being used.)

PR #802, opened 2020-06-16
PHOENIX-5960 Creating a view on a non-existent table throws the wrong…

yanxinyi:
… exception

### DIFF
--- a/phoenix-core/src/main/java/org/apache/phoenix/compile/CreateTableCompiler.java
+++ b/phoenix-core/src/main/java/org/apache/phoenix/compile/CreateTableCompiler.java
@@ -60,6 +60,7 @@ import org.apache.phoenix.schema.PTable;
 import org.apache.phoenix.schema.PTable.ViewType;
 import org.apache.phoenix.schema.PTableType;
 import org.apache.phoenix.schema.SortOrder;
+import org.apache.phoenix.schema.TableNotFoundException;
 import org.apache.phoenix.schema.TableRef;
 import org.apache.phoenix.schema.types.PDataType;
 import org.apache.phoenix.schema.types.PVarbinary;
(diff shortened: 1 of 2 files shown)


yanxinyi:
@ChinmaySKulkarni please review this short changes.

phoenix-core/src/main/java/org/apache/phoenix/compile/CreateTableCompiler.java
```diff
@@ -178,8 +179,8 @@ public MutationPlan compile(CreateTableStatement create) throws SQLException {
                     }
                 }
                 if (isPKMissed) {
-                    throw new SQLExceptionInfo.Builder(SQLExceptionCode.PRIMARY_KEY_MISSING)
-                            .build().buildException();
+                    throw new TableNotFoundException(tableRef.getTable().getSchemaName().toString(),
```
gjacoby126, line 182: Are there cases where the original exception would be the correct one that we need to distinguish?

yanxinyi:
@gjacoby126 yes, we do need handle table is not available v.s pk is missing.

phoenix-core/src/main/java/org/apache/phoenix/compile/CreateTableCompiler.java
```diff
@@ -177,9 +179,22 @@ public MutationPlan compile(CreateTableStatement create) throws SQLException {
                         }
                     }
                 }
+
                 if (isPKMissed) {
-                    throw new SQLExceptionInfo.Builder(SQLExceptionCode.PRIMARY_KEY_MISSING)
-                            .build().buildException();
+                    boolean isTableAvailable = true;
+                    try (HBaseAdmin admin = connection.unwrap(PhoenixConnection.class).getQueryServices().getAdmin()) {
```
ChinmaySKulkarni, line 185: Instead of an extra admin call to check the base table existence here, wouldn't we be checking this already somewhere in the create view path already? I'm wondering if we can avoid this extra check and just let the exception bubble out from that original place. Maybe we can pass the `isPKMissed` flag so that we know later on whether to throw TNFE or Primary-key-missing SQLException

PR #806, opened 2020-06-19
PHOENIX-5946: Implement SchemaExtractionTool utility to get effective…

swaroopak:
… DDL from cluster (Co-authored by Tanuj Khurana)

This is an initial commit and has limited test coverage (have filed PHOENIX-5947 to add tests in the branch)


swaroopak:
Could you please review @tkhurana @qinrui.li

phoenix-core/src/main/java/org/apache/phoenix/schema/SchemaExtractionTool.java
```diff
@@ -0,0 +1,473 @@
...
+        ConnectionQueryServices cqsi = getCQSIObject();
+        HTableDescriptor htd = getHTableDescriptor(cqsi, table);
+        HColumnDescriptor hcd = htd.getFamily(SchemaUtil.getEmptyColumnFamily(table));
+
+        populateDefaultProperties(table);
+        setPTableProperties(table);
+        setHTableProperties(htd);
+        setHColumnFamilyProperties(hcd);
+
+        String columnInfoString = getColumnInfoString(table, null);
```
tkhurana, line 224: might be better if we have 2 different APIs getColumnInfoStringForTable and getColumnInfoStringForView rather than using null

phoenix-core/src/main/java/org/apache/phoenix/schema/SchemaExtractionTool.java
```diff
@@ -0,0 +1,473 @@
...
+        try(Connection conn = getConnection()) {
+            return conn.unwrap(PhoenixConnection.class).getQueryServices();
+        }
+    }
+
+    public static Connection getConnection() throws SQLException {
+        return ConnectionUtil.getInputConnection(conf);
+    }
+
+    private String getColumnInfoString(PTable table, PTable baseTable) {
```
tkhurana, line 322: As mentioned above also, it might be cleaner to refactor this API into two different APIs for table and view and move common functionality to a shared function.

phoenix-core/src/it/java/org/apache/phoenix/end2end/SchemaExtractionToolIT.java
```diff
@@ -0,0 +1,164 @@
...
+            conn.createStatement().execute("CREATE TABLE "+pTableFullName + "(k VARCHAR NOT NULL PRIMARY KEY, v1 VARCHAR, v2 VARCHAR)"
+                    + properties);
+            conn.commit();
+            String [] args = {"-tb", tableName, "-s", schemaName};
+
+            SchemaExtractionTool set = new SchemaExtractionTool();
+            set.setConf(conn.unwrap(PhoenixConnection.class).getQueryServices().getConfiguration());
+            set.run(args);
+            String actualProperties = set.output.substring(set.output.lastIndexOf(")")+1).replace(" ","");
+            Assert.assertEquals(5, actualProperties.split(",").length);
```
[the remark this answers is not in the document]
swaroopak: I am going to heavily rely on Rachel's PR for test coverage if you don't mind.

gjacoby126:
Also would be useful to separate the extraction logic from the SQL generation logic (if possible), so that different output formats are possible (e.g, an Avro schema, a protobuf definition, etc) with generator as an interface.

phoenix-core/src/main/java/org/apache/phoenix/schema/PTable.java
```diff
@@ -800,6 +801,8 @@ private static int getReservedQualifier(byte[] bytes, int offset, int length) {
     Boolean useStatsForParallelization();
     boolean hasViewModifiedUpdateCacheFrequency();
     boolean hasViewModifiedUseStatsForParallelization();
+    Map<String, String> getValues();
```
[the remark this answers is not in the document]
swaroopak: This was used after HTableDescriptor#getValues() but I renamed it for PTable

phoenix-core/src/main/java/org/apache/phoenix/schema/SchemaExtractionTool.java
```diff
@@ -0,0 +1,488 @@
...
+        for(Map.Entry<String, String> entry : map.entrySet()) {
+            String key = entry.getKey();
+            String value = entry.getValue();
+            if(value != null) {
+                definedProps.put(key, value);
+            }
+        }
+    }
+
+    private HTableDescriptor getHTableDescriptor(ConnectionQueryServices cqsi, PTable table)
```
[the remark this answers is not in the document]
swaroopak: That would be a merge conflict change I believe. Otherwise code won't compile on 1.x

phoenix-core/src/main/java/org/apache/phoenix/schema/SchemaExtractionTool.java
```diff
@@ -0,0 +1,488 @@
...
+        return outputBuilder.toString();
+    }
+
+    private void populateDefaultProperties(PTable table) {
+        Map<String, String> propsMap = HColumnDescriptor.getDefaultValues();
+        for (Map.Entry<String, String> entry : propsMap.entrySet()) {
+            String key = entry.getKey();
+            String value = entry.getValue();
+            defaultProps.put(key, value);
+            if(key.equalsIgnoreCase(BLOOMFILTER) || key.equalsIgnoreCase(COMPRESSION)) {
```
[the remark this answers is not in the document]
swaroopak: This is just to override the values that are specifically defaulted in Phoenix. Otherwise, we use HColumnDescriptor.getDefaultValues() to get all version-specific default values for hbase.

phoenix-core/src/main/java/org/apache/phoenix/schema/SchemaExtractionTool.java
```diff
@@ -0,0 +1,488 @@
...
+    protected String extractCreateIndexDDL(PTable indexPTable)
+            throws SQLException {
+        String pSchemaName = indexPTable.getSchemaName().getString();
+        String pTableName = indexPTable.getTableName().getString();
+
+        String baseTableName = indexPTable.getParentTableName().getString();
+        String baseTableFullName = SchemaUtil.getQualifiedTableName(indexPTable.getSchemaName().getString(), baseTableName);
+        PTable dataPTable = getPTable(baseTableFullName);
+
+        String defaultCF = SchemaUtil.getEmptyColumnFamilyAsString(indexPTable);
```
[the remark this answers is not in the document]
swaroopak: It would come as a Test and handle this particular case.

phoenix-core/src/main/java/org/apache/phoenix/schema/SchemaExtractionTool.java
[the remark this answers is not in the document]
swaroopak: The least I could do is change the function name as you suggested.

swaroopak:
@gjacoby126 except for refactoring to extend the support for different output formats, I fixed all the comments. Maybe we can take this up in the following jira?

phoenix-core/src/main/java/org/apache/phoenix/schema/PTableImpl.java
```diff
@@ -230,6 +254,7 @@
         private Boolean useStatsForParallelization;
         private long viewTTL;
         private long viewTTLHighWaterMark;
+        private Map<String, String> values = new HashMap<>();
```
priyankporwal, line 257: "values" seem like a bad name for maps that has both keys and values - how about "properties"? Looks like that's what are being put there.

phoenix-tools/src/main/java/org/apache/phoenix/schema/SchemaExtractionTool.java
```diff
@@ -0,0 +1,98 @@
...
+    private static final Option TABLE_OPTION = new Option("tb", "table", true,
+            "[Required] Table name ex. table1");
+    private static final Option SCHEMA_OPTION = new Option("s", "schema", true,
+            "[Optional] Schema name ex. schema");
+
+    private String pTableName;
+    private String pSchemaName;
+
+    public static Configuration conf;
+    public String output;
```
gjacoby126, line 33: output should have a getter since you're accessing it lots of times from tests and I assume it's read-only externally.

phoenix-tools/src/main/java/org/apache/phoenix/schema/SchemaExtractionProcessor.java
```diff
@@ -0,0 +1,410 @@
...
+            definedProps.put(Bytes.toString(key.get()), Bytes.toString(value.get()));
+        }
+    }
+
+    private void setHColumnFamilyProperties(HColumnDescriptor columnDescriptor) {
+        Map<ImmutableBytesWritable, ImmutableBytesWritable> propsMap = columnDescriptor.getValues();
+        for (Map.Entry<ImmutableBytesWritable, ImmutableBytesWritable> entry : propsMap.entrySet()) {
+            ImmutableBytesWritable key = entry.getKey();
+            ImmutableBytesWritable value = entry.getValue();
+            definedProps.put(Bytes.toString(key.get()), Bytes.toString(value.get()));
```
gjacoby126, line 246: This will be incorrect for multiple column families. (e.g two CFs with different values for the same property will overwrite each other)

phoenix-tools/src/main/java/org/apache/phoenix/schema/SchemaExtractionProcessor.java
```diff
@@ -0,0 +1,410 @@
...
+    public String process() throws Exception {
+        String ddl = null;
+        if(this.table.getType().equals(PTableType.TABLE)) {
+            ddl = extractCreateTableDDL(this.table);
+        } else if(this.table.getType().equals(PTableType.INDEX)) {
+            ddl = extractCreateIndexDDL(this.table);
+        } else if(this.table.getType().equals(PTableType.VIEW)) {
+            ddl = extractCreateViewDDL(this.table);
+        }
+        return ddl;
```
gjacoby126, line 54: Since the object's immutable, any reason not to cache the generated DDL and return the cached copy if process() is called more than once?

phoenix-tools/src/main/java/org/apache/phoenix/schema/SchemaExtractionProcessor.java
```diff
@@ -0,0 +1,410 @@
...
+        outputBuilder.append(")");
+        if(!coveredColumnString.equals("")) {
+            outputBuilder.append(" INCLUDE (");
+            outputBuilder.append(coveredColumnString);
+            outputBuilder.append(")");
+        }
+        return outputBuilder.toString();
+    }
+
+    PTable getPTable(String pTableFullName) throws SQLException {
```
gjacoby126, line 153: This won't work for tenant-owned schema objects -- if you're not going to tackle this now please file a subtask JIRA to do so before we merge into a real branch

phoenix-tools/src/main/java/org/apache/phoenix/schema/SchemaExtractionProcessor.java
swaroopak: Would like to get it fixed in subsequent commits.

phoenix-tools/src/main/java/org/apache/phoenix/schema/SchemaExtractionProcessor.java
swaroopak: Would like to get it fixed in subsequent commits. Yes, Let me file one.

swaroopak: I fixed it in the same commit.

phoenix-tools/src/main/java/org/apache/phoenix/schema/SchemaExtractionProcessor.java
swaroopak: Added a message to throw Exception with "Not Supported" message

PR #808, opened 2020-06-20
PHOENIX-5969 Read repair reduces the number of rows returned for LIMI…

kadirozde:
…T queries

### DIFF
--- a/phoenix-core/src/main/java/org/apache/phoenix/index/GlobalIndexChecker.java
+++ b/phoenix-core/src/main/java/org/apache/phoenix/index/GlobalIndexChecker.java
@@ -27,13 +27,10 @@ import static org.apache.phoenix.schema.types.PDataType.TRUE_BYTES;
 
 import static org.apache.phoenix.compat.hbase.CompatUtil.*;
 
-
 import java.io.IOException;
-import java.util.Collections;
 import java.util.Iterator;
 import java.util.List;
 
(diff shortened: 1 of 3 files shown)


phoenix-core/src/main/java/org/apache/phoenix/index/GlobalIndexChecker.java
```diff
@@ -49,12 +51,16 @@
...
+import org.apache.hadoop.hbase.filter.Filter;
+import org.apache.hadoop.hbase.filter.FilterList;
+import org.apache.hadoop.hbase.filter.PageFilter;
 import org.apache.hadoop.hbase.io.ImmutableBytesWritable;
 import org.apache.hadoop.hbase.regionserver.Region;
 import org.apache.hadoop.hbase.regionserver.RegionScanner;
 import org.apache.hadoop.hbase.regionserver.ScannerContext;
 import org.apache.hadoop.hbase.util.Bytes;
 import org.apache.phoenix.coprocessor.BaseScannerRegionObserver;
+import org.apache.phoenix.filter.EncodedQualifiersColumnProjectionFilter;
```
gjacoby126, line 63: nit: unused import?

abhishek-chouhan:
Discussed offline with @kadirozde . It would also be good (on both server and client side) to recursively look for filter lists inside of filter lists too

PR #818, opened 2020-07-02
PHOENIX-5984: Query timeout counter is not updated in all timeouts cases

ChinmaySKulkarni:


ChinmaySKulkarni:
Please review @gjacoby126 @yanxinyi @jpisaac

phoenix-core/src/it/java/org/apache/phoenix/monitoring/PhoenixMetricsIT.java
```diff
@@ -93,6 +97,23 @@
 
     private static final Logger LOGGER = LoggerFactory.getLogger(PhoenixMetricsIT.class);
 
+    private static class MyClock extends EnvironmentEdge {
```
dbwong, line 100: I'm pretty sure we have a similar type of clock before.  Maybe we put it into test utils?

ChinmaySKulkarni: We have custom EnvironmentEdge classes throughout our tests. I agree it is a good idea to consolidate them to one TestUtil, but I think that will be more diffs than the main diff of this Jira. Created this Jira: https://issues.apache.org/jira/browse/PHOENIX-5985

PR #820, opened 2020-07-02
PHOENIX-5861 Delete index data failed,due to pool closed

wangchao316:
PHOENIX-5861 Delete index data failed,due to pool closed

### DIFF
--- a/phoenix-core/src/main/java/org/apache/phoenix/hbase/index/Indexer.java
+++ b/phoenix-core/src/main/java/org/apache/phoenix/hbase/index/Indexer.java
@@ -652,14 +652,18 @@ public class Indexer extends BaseRegionObserver {
 
   @Override
   public void postOpen(final ObserverContext<RegionCoprocessorEnvironment> c) {
-    Multimap<HTableInterfaceReference, Mutation> updates = failedIndexEdits.getEdits(c.getEnvironment().getRegion());
-    
     if (this.disabled) {
         super.postOpen(c);
         return;
     }
 
+    // if stopped is true, and then init writer
+    if (this.stopped) {
+        initBuilderAndWriter(c.getEnvironment());
+    }
+
     long start = EnvironmentEdgeManager.currentTimeMillis();
+    Multimap<HTableInterfaceReference, Mutation> updates = failedIndexEdits.getEdits(c.getEnvironment().getRegion());
     try {
         //if we have no pending edits to complete, then we are done
         if (updates == null || updates.size() == 0) {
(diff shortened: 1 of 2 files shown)


phoenix-core/src/main/java/org/apache/phoenix/hbase/index/Indexer.java
```diff
@@ -800,5 +804,62 @@ public static void enableIndexing(HTableDescriptor desc, Class<? extends IndexBu
...
+     * this need init builder and writer,recoveryWriter.
+     *
+     * @param env
+     */
+  private void initBuilderAndWriter(RegionCoprocessorEnvironment env) {
+      // init index build manager
+      try {
+          this.builder = new IndexBuildManager(env);
+      } catch (IOException e) {
+          LOG.warn("failed to new index build manager", e);
```
stoty, line 819: Catching the exception here, and letting the function continue doesn't seem right, as we'll run into problems later.
Why not just let them bubble up, and handle them at the caller ?

phoenix-core/src/main/java/org/apache/phoenix/hbase/index/Indexer.java
```diff
@@ -800,5 +804,62 @@ public static void enableIndexing(HTableDescriptor desc, Class<? extends IndexBu
...
+              DEFAULT_INDEX_WRITER_RPC_RETRIES_NUMBER));
+      clonedConfig.setInt(HConstants.HBASE_CLIENT_PAUSE, env.getConfiguration()
+          .getInt(INDEX_WRITER_RPC_PAUSE, DEFAULT_INDEX_WRITER_RPC_PAUSE));
+      DelegateRegionCoprocessorEnvironment indexWriterEnv = new DelegateRegionCoprocessorEnvironment(clonedConfig, env);
+      // setup the actual index writer
+      String serverName = env.getRegionServerServices().getServerName().getServerName();
+      try {
+          this.writer = new IndexWriter(indexWriterEnv, serverName + "-index-writer");
+      } catch (IOException e) {
+          LOG.warn("Failed to init index writer ", e);
```
stoty, line 843: Again, why catch here ?

phoenix-core/src/main/java/org/apache/phoenix/hbase/index/Indexer.java
```diff
@@ -800,5 +804,62 @@ public static void enableIndexing(HTableDescriptor desc, Class<? extends IndexBu
...
+          Class<? extends IndexFailurePolicy> policyClass =
+              env.getConfiguration().getClass(INDEX_RECOVERY_FAILURE_POLICY_KEY,
+                  StoreFailuresInCachePolicy.class, IndexFailurePolicy.class);
+          IndexFailurePolicy policy =
+              policyClass.getConstructor(PerRegionIndexWriteCache.class).newInstance(failedIndexEdits);
+          LOG.debug("Setting up recovery writter with failure policy: " + policy.getClass());
+          recoveryWriter =
+              new RecoveryIndexWriter(policy, indexWriterEnv, serverName + "-recovery-writer");
+      } catch (Exception e) {
+          LOG.warn("Failed to init recovery writer ", e);
```
stoty, line 859: Again, why catch here ?

phoenix-core/src/main/java/org/apache/phoenix/hbase/index/Indexer.java
```diff
@@ -652,14 +652,18 @@ private void doPostWithExceptions(ObserverContext<RegionCoprocessorEnvironment>
...
-    Multimap<HTableInterfaceReference, Mutation> updates = failedIndexEdits.getEdits(c.getEnvironment().getRegion());
-    
     if (this.disabled) {
         super.postOpen(c);
         return;
     }
 
+    // if stopped is true, and then init writer
+    if (this.stopped) {
+        initBuilderAndWriter(c.getEnvironment());
```
stoty, line 662: This would mean that we always re-enable a stopped indexer.

I would guess that there are some cases when we don't want to this.

phoenix-core/src/main/java/org/apache/phoenix/hbase/index/Indexer.java
```diff
@@ -800,5 +804,62 @@ public static void enableIndexing(HTableDescriptor desc, Class<? extends IndexBu
...
     desc.addCoprocessor(Indexer.class.getName(), null, priority, properties);
   }
+
+    /**
+     * stop indexer when region split, but split failed, and then rollback, open region.
+     * this need init builder and writer,recoveryWriter.
+     *
+     * @param env
+     */
+  private void initBuilderAndWriter(RegionCoprocessorEnvironment env) {
```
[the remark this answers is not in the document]
wangchao316: this is a part of start(), extract init builder and writer, because region split, builder and writer are close. but split failed, and then reigon rollback,  reopen region, so need init builder and writer again.

phoenix-core/src/main/java/org/apache/phoenix/hbase/index/Indexer.java
wangchao316: only this.stopped is true when region close. when region open again, need to init builder and writer. when region split failed, and then rollback ,do not have init builder and writer.

phoenix-core/src/main/java/org/apache/phoenix/hbase/index/Indexer.java
wangchao316: > Catching the exception here, and letting the function continue doesn't seem right, as we'll run into problems later.
> Why not just let them bubble up, and handle them at the caller ?

the postOpen() of hbase coprocessor can not bubble up this catch, only print log in this.

phoenix-core/src/main/java/org/apache/phoenix/hbase/index/Indexer.java
wangchao316: > Again, why catch here ?

the postOpen() of hbase coprocessor can not bubble up this catch, only print log in this.

phoenix-core/src/main/java/org/apache/phoenix/hbase/index/Indexer.java
wangchao316: > Again, why catch here ?

the postOpen() of hbase coprocessor can not bubble up this catch, only print log in this.

PR #809, opened 2020-06-24
PHOENIX-5973 - Stabilize and speed up IndexToolForNonTxGlobalIndexIT

gjacoby126:

PR #821, opened 2020-07-02
PHOENIX-5973 - Stabilize and speed up IndexToolForNonTxGlobalIndexIT …

gjacoby126:
…(#809)

### DIFF
--- a/phoenix-core/src/it/java/org/apache/phoenix/end2end/IndexToolForNonTxGlobalIndexIT.java
+++ b/phoenix-core/src/it/java/org/apache/phoenix/end2end/IndexToolForNonTxGlobalIndexIT.java
@@ -20,16 +20,25 @@ package org.apache.phoenix.end2end;
 import com.google.common.collect.Maps;
 import org.apache.hadoop.hbase.Cell;
 import org.apache.hadoop.hbase.CellUtil;
+import org.apache.hadoop.hbase.DoNotRetryIOException;
 import org.apache.hadoop.hbase.TableName;
 import org.apache.hadoop.hbase.client.Admin;
 import org.apache.hadoop.hbase.client.ColumnFamilyDescriptor;
 import org.apache.hadoop.hbase.client.ColumnFamilyDescriptorBuilder;
 import org.apache.hadoop.hbase.client.ConnectionFactory;
+import org.apache.hadoop.hbase.client.CoprocessorDescriptor;
+import org.apache.hadoop.hbase.client.CoprocessorDescriptorBuilder;
 import org.apache.hadoop.hbase.client.Delete;
 import org.apache.hadoop.hbase.client.Result;
 import org.apache.hadoop.hbase.client.ResultScanner;
 import org.apache.hadoop.hbase.client.Scan;
 import org.apache.hadoop.hbase.client.Table;
+import org.apache.hadoop.hbase.client.TableDescriptor;
+import org.apache.hadoop.hbase.client.TableDescriptorBuilder;
+import org.apache.hadoop.hbase.coprocessor.BaseRegionObserver;
+import org.apache.hadoop.hbase.coprocessor.ObserverContext;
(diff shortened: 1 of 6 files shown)


phoenix-core/src/it/java/org/apache/phoenix/end2end/IndexToolForNonTxGlobalIndexIT.java
```diff
@@ -952,20 +998,22 @@ public void testIndexToolForIncrementalVerify_viewIndex() throws Exception {
...
                             "-et", String.valueOf(t3));
             verifyCounters(it, 2, 2);
-
+/*
             // job with update on only one row
             it =
                     IndexToolIT.runIndexTool(directApi, useSnapshot, schemaName, viewName,
                             viewIndexName, null, 0, IndexTool.IndexVerifyType.ONLY, "-st", String.valueOf(t3),
                             "-et", String.valueOf(t4));
             verifyCounters(it, 1, 1);
```
swaroopak, line 1007: This should return 0 scanned and 0 valid because the view has where clause on val6 which won't qualify for the delete marker that is there on HBase row. Thank you @abhishek-chouhan for clarifying this. 

That means in order to consider a deleted row for incremental verification on a view with where clause, its upsert should also fall into the same window. Otherwise, any such mismatches will be caught in full verification. 

FYI @kadirozde @priyankporwal

phoenix-core/src/it/java/org/apache/phoenix/end2end/IndexToolForNonTxGlobalIndexIT.java
```diff
@@ -791,95 +826,107 @@ public void testIndexToolForIncrementalVerify() throws Exception {
...
             it = IndexToolIT.runIndexTool(directApi, useSnapshot, schemaName, dataTableName, indexTableName,
                     null, 0, IndexTool.IndexVerifyType.ONLY, "-st", String.valueOf(t0),"-et", String.valueOf(t2));
             verifyCounters(it, 2, 2);
+            customEdge.incrementValue(waitForUpsert);
 
             // job with update on only one row
             it = IndexToolIT.runIndexTool(directApi, useSnapshot, schemaName, dataTableName, indexTableName,
                     null, 0, IndexTool.IndexVerifyType.ONLY, "-st", String.valueOf(t1),"-et", String.valueOf(t3));
-            verifyCounters(it, 1, 2);
+            verifyCounters(it, 1, 1);
```
swaroopak, line 908: There are 2 index rows in this time window. Wondering if it is returning 1 because of different skip scan behaviors in the master branch's HBase or something else? Do you know @abhishek-chouhan?

phoenix-core/src/it/java/org/apache/phoenix/end2end/IndexToolForNonTxGlobalIndexIT.java
gjacoby126: That seems like incorrect behavior though, because that delete statement _does_ delete a row that's part of the view.

phoenix-core/src/it/java/org/apache/phoenix/end2end/IndexToolForNonTxGlobalIndexIT.java
gjacoby126: There should be 2 versions of a single row, I believe. Different max versions behavior coming from lack of max lookback protection maybe?

phoenix-core/src/it/java/org/apache/phoenix/end2end/IndexToolForNonTxGlobalIndexIT.java
gjacoby126: Based on our previous conversation that led to the filing of PHOENIX-5989, I've just updated this with a reference to 5989 and kept it commented out.

gjacoby126:
thanks for the review, @swaroopak

PR #823, opened 2020-07-07
PHOENIX-5991 IndexRegionObserver should not overwrite mutation timest…

kadirozde:
…amps set by clients

### DIFF
--- a/phoenix-core/src/it/java/org/apache/phoenix/end2end/DeleteIT.java
+++ b/phoenix-core/src/it/java/org/apache/phoenix/end2end/DeleteIT.java
@@ -37,6 +37,8 @@ import java.util.List;
 import java.util.Properties;
 
 import org.apache.phoenix.query.QueryServices;
+import org.apache.phoenix.util.EnvironmentEdge;
+import org.apache.phoenix.util.EnvironmentEdgeManager;
 import org.apache.phoenix.util.PropertiesUtil;
 import org.apache.phoenix.util.QueryUtil;
 import org.junit.Test;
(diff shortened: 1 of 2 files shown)


phoenix-core/src/it/java/org/apache/phoenix/end2end/DeleteIT.java
```diff
@@ -817,7 +838,59 @@ private void testDeleteCount(boolean autoCommit, Integer limit) throws Exception
...
 
     }
-    
+
+    private void testDeleteCountWithIndex(boolean autoCommit, boolean scn) throws Exception {
+        String tableName = generateUniqueName();
+        String tableDDL = "CREATE TABLE IF NOT EXISTS " + tableName + " (pk1 DECIMAL NOT NULL, v1 VARCHAR CONSTRAINT PK PRIMARY KEY (pk1))";
+        String indexName = generateUniqueName();
+        String indexDDL = "CREATE INDEX IF NOT EXISTS " + indexName + " ON " + tableName + "(v1)";
+        int numRecords = 1010;
```
gjacoby126, line 847: do we need so many rows to test the timestamp behavior? How much time do these tests add to the test suite run? Seems like you could get the same effect in the scn case with 2 rows, using the edge manager to force a two millisecond time increment between the two upserts.

gjacoby126: If you do go that way, please remember to reset() the EnvironmentEdgeManager in a finally block

kadirozde: Each test takes about 5 to 10 seconds even with 1010 rows. I needed to see the same timestamp is used across multiple batches so I need to have several batches of rows. I have not injected an EnvironmentEdge and so no need to reset.

gjacoby126: OK, that's fine. Thanks!

gjacoby126:
@kadirozde - since HBase timestamps are ultimately derived from physical timestamps, rather than, say HLCs, we're already prone to weird effects from clock skew (such as different rs's having different clocks). This improves that in one aspect, since when the client-set timestamp "wins" it'll be consistent across region servers.

I'm curious about the implications of letting the server timestamp win when the server is behind the client. What if the client clock is slightly ahead of a server? Don't we go back to inconsistent timestamps then?

It seems like a preference for either the client or the server can cause skew-related problems.

When we honor the client timestamps, we get consistent timestamps across a logical operation no matter how many region servers it involves. However, we can get weird behavior where two clients with clock skew wrt each other can invert the order of operations.

When we honor the server timestamps, we can enforce an order between client operations regardless of the clients' clocks, but we can get inconsistent timestamps within each operation. 

I can think of use cases I've worked with in the past that have a strong need for either consistent timestamps across a long-running operations and others with lots of small operations where order of operations across clients _really_ matters , so I'm thinking this needs to be configurable.

gjacoby126:
@kadirozde - good point about the SCN. I think if this is configurable, there would need to be a validation check giving you an exception if you try to ask for both a client-set SCN and server side timestamps on mutations. 

I like your suggestion about a given operation deterministically using either the client or server timestamp. 

Seems to me the big question is how clients and servers decide which of those two to use. If it's the client's choice to either set or not set a timestamp, that still leaves the question of how the client chooses. Same if it's the server's choice -- is it a cluster setting, a table setting, or a hard-coded default? Dependent on mutable vs immutable? Lots of implications to work through for all possibilities, and I don't have a firm opinion yet. 

For example, it would be good for a particular table to be able to override the default to say whether client or server-side timestamps should govern. I can also see it being useful for clients to request it as well as a connection property, as they do for SCN, but I worry a little about what happens if different clients make different choices...so I'm undecided.

PR #824, opened 2020-07-07
PHOENIX-5988 Bump HBase and Hadoop versions to latest patch level on …

stoty:
…master

Hadoop: 3.0.0 -> 3.0.3
HBase 2.1: 2.1.8 -> 2.1.9
Hbase 2.2: 2.2.3 -> 2.2.4 (see ticket)

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -78,7 +78,7 @@
     <!-- Hadoop and Hbase-thirdparty version -->
     <!-- These are expected to be overridden to conform to cluster versions
     along with hbase.version (defined in profiles) -->
-    <hadoop.version>3.0.0</hadoop.version>
+    <hadoop.version>3.0.3</hadoop.version>
     <hbase.thirdparty.version>2.2.1</hbase.thirdparty.version>
 
     <!-- General Properties -->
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -78,7 +78,7 @@
     <!-- Hadoop and Hbase-thirdparty version -->
     <!-- These are expected to be overridden to conform to cluster versions
     along with hbase.version (defined in profiles) -->
-    <hadoop.version>3.0.0</hadoop.version>
+    <hadoop.version>3.0.3</hadoop.version>
```
joshelser, line 81: I don't see any Hadoop 3.0 releases listed https://hadoop.apache.org/releases.html anymore. According to https://cwiki.apache.org/confluence/display/HADOOP/EOL+%28End-of-life%29+Release+Branches 3.0 is EOL.

Maybe we make this 3.1.3?

stoty: 3.1.3 is right out until we do something about Guava, as it pulls in Guava 28, which breaks Phoenix+ Tephra spectacularly.

We could use 3.1.2, but that is known to be incompatible with the HBase 2.2.4 maven artifact, as documented in HBASE-22394 .

Hadoop 3.0.3  + Hbase 2.2.4 is the highest we can go until we solve PHOENIX-5993. (Really looking forward to hearing your ideas on that, in the ticket or in the discuss thread on the list)

joshelser: Ooof, ok. Let me go over there while I'm in a "trying to catch up" mode :)

PR #827, opened 2020-07-10
PHOENIX-5997 Phoenix Explain Plan for Deletes does not clearly differentiate between server side and client side paths.

dbwong:
https://issues.apache.org/jira/browse/PHOENIX-5997

### DIFF
--- a/phoenix-core/src/test/java/org/apache/phoenix/query/ExplainPlanTextTest.java
+++ b/phoenix-core/src/test/java/org/apache/phoenix/query/ExplainPlanTextTest.java
@@ -1,0 +1,50 @@
+package org.apache.phoenix.query;
+
+import org.apache.phoenix.jdbc.PhoenixConnection;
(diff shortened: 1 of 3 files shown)


dbwong:
Whoops just noticed the formating is wrong let me fix that real quick.

phoenix-core/src/test/java/org/apache/phoenix/query/ExplainPlanTextTest.java
```diff
@@ -0,0 +1,52 @@
+package org.apache.phoenix.query;
```
gjacoby126, line 1: Missing an Apache License

dbwong:
@gjacoby126 added header.

yanxinyi:
@gjacoby126 Hi Geoffrey, is there anything else other than adding Apache License?

gjacoby126:
JIRA already committed, closing PR

PR #834, opened 2020-07-18
PHOENIX-6026: Fix BackwardCompatibilityIT so it can run locally

ChinmaySKulkarni:


phoenix-core/src/it/java/org/apache/phoenix/end2end/BackwardCompatibilityIT.java
```diff
@@ -118,6 +125,7 @@ public synchronized void doSetup() throws Exception {
         zkQuorum = "localhost:" + hbaseTestUtil.getZkCluster().getClientPort();
         url = PhoenixRuntime.JDBC_PROTOCOL + PhoenixRuntime.JDBC_PROTOCOL_SEPARATOR + zkQuorum;
         DriverManager.registerDriver(PhoenixDriver.INSTANCE);
+        checkForPreConditions();
```
sguggilam, line 128: We need to have this before each test as it checks for the non existence of any SYSTEM tables before each test run to make sure that there is no existing metadata

sguggilam:
Just one comment, else looks good to me

phoenix-core/src/it/java/org/apache/phoenix/end2end/BackwardCompatibilityIT.java
ChinmaySKulkarni: Yes that's what is happening. This is now called in `doSetup()` which will be invoked before each test i.e. it is marked as `@Before`

sguggilam: Makes sense, thanks

PR #837, opened 2020-07-19
PHOENIX-6016 Support HBase 2.3.0

stoty:


stoty:
This still has a few IT flakies/failures, which seem more related to HBase 2.3 or Hadoop 3.1.2 than
 to Phoenix.
Those are easier to classify and fix later, when we figure out how to run tests from Jenkins with 2.3 (PHOENIX-5993 )

stoty:
It seems to have a problem with local indexes.
Investigating.

phoenix-core/src/it/java/org/apache/hadoop/hbase/regionserver/wal/WALReplayWithIndexWritesAndCompressedWALIT.java
```diff
@@ -176,11 +177,15 @@ private void deleteDir(final Path p) throws IOException {
...
 @Test
   public void testReplayEditsWrittenViaHRegion() throws Exception {
-    final String tableNameStr = "testReplayEditsWrittenViaHRegion";
-    final RegionInfo hri = RegionInfoBuilder.newBuilder(org.apache.hadoop.hbase.TableName.valueOf(tableNameStr)).setSplit(false).build();
-    final Path basedir = FSUtils.getTableDir(hbaseRootDir, org.apache.hadoop.hbase.TableName.valueOf(tableNameStr));
-    deleteDir(basedir);
-    final TableDescriptor htd = createBasic3FamilyHTD(tableNameStr);
+      final String tableNameStr = "testReplayEditsWrittenViaHRegion";
+      final RegionInfo hri =
+              RegionInfoBuilder.newBuilder(org.apache.hadoop.hbase.TableName.valueOf(tableNameStr))
```
joshelser, line 182: Why the fully-qualified TableName?

stoty: I have no I idea, I've only replaced the class, but didn't touch the arguments.
Will try to pretty this up.

stoty:
This (minus a rebase and some formatting fixups) has passed a full mvn verify run now, with only a known flaky test failing.

Please review!

joshelser:
> This (minus a rebase and some formatting fixups) has passed a full mvn verify run now, with only a known flaky test failing.

🎉 🎉 🎉

phoenix-core/pom.xml
```diff
@@ -493,6 +493,33 @@
...
+      <dnsjava.version>2.1.7</dnsjava.version>
+    </properties>
+    <dependencies>
+      <dependency>
+        <groupId>org.apache.zookeeper</groupId>
+        <artifactId>zookeeper-jute</artifactId>
+      </dependency>
+      <dependency>
+        <groupId>dnsjava</groupId>
+        <artifactId>dnsjava</artifactId>
```
joshelser, line 516: Why's this coming in?

YARN was using dnsjava, but I didn't think we had anything directly in hbase or phoenix.

stoty: No Idea.
Dependency plugin seems to get confused sometimes. (Maybe some unused return value ?)
I've removed the dependency, and added an exception to the dependency plugin config.

stoty:
BackwardCompatibilityIT was broken by the rebase.
I've fixed it.

PR #838, opened 2020-07-21
PHOENIX-6034 Optimize InListIT

ankitsinghal:
These changes have reduced the time of the test from 16mins to 50seconds.

### DIFF
--- a/phoenix-core/src/it/java/org/apache/phoenix/end2end/InListIT.java
+++ b/phoenix-core/src/it/java/org/apache/phoenix/end2end/InListIT.java
@@ -26,15 +26,18 @@ import static org.junit.Assert.assertTrue;
 
 import java.sql.Connection;
 import java.sql.DriverManager;
+import java.sql.PreparedStatement;
 import java.sql.ResultSet;
 import java.sql.SQLException;
 import java.sql.Statement;
-import java.sql.PreparedStatement;
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.List;
 import java.util.Properties;
 
+import com.google.common.base.Function;
+import com.google.common.base.Joiner;
+import com.google.common.collect.Lists;
 import org.apache.phoenix.compile.QueryPlan;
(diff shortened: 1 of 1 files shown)


stoty (approved):
LGTM

phoenix-core/src/it/java/org/apache/phoenix/end2end/InListIT.java
```diff
@@ -43,27 +46,39 @@
...
+    private static String tableName2 = generateUniqueName();
+    private static String descViewName = generateUniqueName();
+    private static String ascViewName = generateUniqueName();
+    private static String viewName1 = generateUniqueName();
+    private static String viewName2 = generateUniqueName();
+    private static String prefix = generateUniqueName();
 
     @Before
     public void setup() throws Exception {
-        tableName = generateUniqueName();
```
lhofhansl, line 62: Instead of the initialized variable, could we change this to the @BeforeClass annotation instead?
In fact would the whole test still work if we only change @Before to @BeforeClass?

lhofhansl: And sorry I missed this for so long. I was off end of July and then missed it.

ankitsinghal: > would the whole test still work if we only change @before to @BeforeClass?
Yeah , actually @BeforeClass requires a method to be static so all the data structures need to be updated as static. I didn't spend time doing that probably because of being lazy though could have used @BeforeAll and make the Junit lifecycle to be @TestInstance(Lifecycle.PER_CLASS) to avoid an initialized variable.

stoty: I looked up @TestInstance because it sounds interesting, but it's a Junit5 feature, and we're still on Junit 4. AFAICT The only solutions are your current one, or using @BeforeClass as you said.

PR #839, opened 2020-07-22
PHOENIX-5986 Fix flapping test

gokceni:


gokceni:
The issue was that the task was still running when we do syscat query in the test @kadirozde

kadirozde:
> The issue was that the task was still running when we do syscat query in the test @kadirozde
@gokceni, I think the issue here is that we call TaskRegionObserver.run() within the IT while TaskRegionObserver is scheduled to run periodically. In some cases, the tasks are picked up by the periodic execution and in other cases by the direct call from the IT. This causes a race between these two different execution mode. So, the right solution is to disable the periodic runs in this IT to eliminate this race.

phoenix-core/src/it/java/org/apache/phoenix/end2end/IndexRebuildTaskIT.java
```diff
@@ -178,11 +178,15 @@ public static void waitForTaskState(Connection conn, PTable.TaskType taskType, S
         String taskData = "";
         do {
             Thread.sleep(2000);
-            ResultSet rs = conn.createStatement().executeQuery("SELECT * " +
+            String stmt = "SELECT * " +
                     " FROM " + PhoenixDatabaseMetaData.SYSTEM_TASK_NAME +
-                    " WHERE " + PhoenixDatabaseMetaData.TABLE_NAME + "='" + expectedTableName + "' AND " +
```
kadirozde, line 183: Let's keep the table name in the query. When the task table is used for other use cases, there might be other entries from some other tables in the task table even during this IT

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/tasks/DropChildViewsTask.java
```diff
@@ -46,6 +46,7 @@
     public TaskRegionObserver.TaskResult run(Task.TaskRecord taskRecord) {
         PhoenixConnection pconn = null;
         Timestamp timestamp = taskRecord.getTimeStamp();
+        LOGGER.info("DropChildViewTask is running for task " + taskRecord.getTableName());
```
kadirozde, line 49: This was suggested for only debugging. Let's remove it or use LOGGER.debug() instead

kadirozde: And the same comment for the other instances of LOGGER.info

gokceni:
Merged

PR #841, opened 2020-07-23
PHOENIX-6028 Change default HBase profile to 2.1 in 5.x

stoty:
* change default HBase profile
* bump the default HBase version to 2.2.5 for the 2.2 profile
* update BUILDING.md and add instuctions for PHOENIX-5593

### DIFF
--- a/dev/rebuild_hbase.sh
+++ b/dev/rebuild_hbase.sh
@@ -1,0 +1,16 @@
+#!/bin/sh
+# Downloads the specified HBase version sources into PWD, extracts it, 
+# then rebuilds and installs the artifacts with -Dhadoop.profile=3.0
+
+if [ $# -ne 1 ]
+  then
+    echo "Supply the Hbase version as paramater i.e.: rebuild_hbase.sh 2.2.6 "
+fi
+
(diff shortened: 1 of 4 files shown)


dev/rebuild_hbase.sh
```diff
@@ -0,0 +1,16 @@
+#!/bin/sh
+# Downloads the specified HBase version sources into PWD, extracts it, 
+# then rebuilds and installs the artifacts with -Dhadoop.profile=3.0
+
+if [ $# -ne 1 ]
+  then
+    echo "Supply the Hbase version as paramater i.e.: rebuild_hbase.sh 2.2.6 "
+fi
+
+wget https://downloads.apache.org/hbase/$1/hbase-$1-src.tar.gz
```
joshelser, line 10: Infra requests that downloads.a.o/archive.a.o is only used for downloading KEYS, checksums, and signatures.

Can you adapt the solution I have in https://github.com/apache/phoenix-queryserver/blob/master/python-phoenixdb/dev-support/cache-apache-project-artifact.sh to this script as well, please?

stoty: Sure, I will do that before commit.

PR #843, opened 2020-07-23
PHOENIX-6037 Change default HBase profile to 1.3 in 4.x

stoty:


gjacoby126:
Defaulting to the latest patch of the earliest supported HBase minor release makes sense to me. That said, it might be worth having a conversation on the dev list about whether we should support 1.3 in the forthcoming 4.16 release, since the HBase community almost-but-not-quite EOLed it late last year and there's been no releases since.

stoty:
@gjacoby126 I don't have a strong opinion on 1.3.
One thing to keep in mind is that there's a lot of compatibility module code that is needed because of 1.3, so when drop 1.3, we should not simply remove the module, but revert the HBase-1.3 compatibility commit to simplify the compat modules.

stoty:
Hmm. Github didn't auto-close this.

PR #845, opened 2020-07-27
PHOENIX-6010 Create phoenix-thirdparty, and consume guava through it

stoty:


stoty:
It's a huge patch, but 99% of it is import renames.
There are some minor code changes to account for the the Guava differences between 13 and 29.
The rest is the POM change, where we switch to consuming the shaded Tephra artifacts, and change to the SNAPSHOT versions of Omid and Tephra.

pom.xml
```diff
@@ -82,6 +82,8 @@
     <hadoop.version>3.0.3</hadoop.version>
     <hbase.thirdparty.version>2.2.1</hbase.thirdparty.version>
 
+    <phoenix.thirdparty.version>1.0.0-SNAPSHOT</phoenix.thirdparty.version>
```
joshelser, line 85: Noting that these are pulling snapshots that someone deploys by hand. We need to prioritize a release of these repo's (thirdparty as well as the tephra and omid release) in some shape/form. Doesn't matter to me if we slap a "-beta" suffix onto any of them, just depending on SNAPSHOT releases long term is going to create problems for us.

joshelser (changes requested):
Running `fgrep -R 'com.google.' **/src/main/ | fgrep -v thirdparty | fgrep -v coprocessor/generated | fgrep -v com.google.protobuf | fgrep -v com.google.inject | fgrep -v '@see'` gives me a list of about 30 or so imports of non-shaded guava classes. Some you missed?

I assume you ran this through some basic tests locally (not just compilation)?

pom.xml
stoty: Omid and Tephra have the same SDNAPSHOT issue.

As soon as we get this patch in, and it has had a week or so to prove not having caused major regressions, I'll start the release discussion for all dependee subprojects (thirdparty, omid, tephra) in preparation of 5.1.

I did not want to release those, until we have had a chance to test their full integration with Phoenix.

stoty:
That's embarrassing. 
Yes, I couldn't find those references, and while dependency-check did warn of the problem, I excluded guava from it instead, which seemed like a good idea at the time.

I have only run the standard maven IT suite (i.e maven verify) on the code, but I can do a smoke test with a local HBase install as well.

pom.xml
joshelser: Sounds good! You have a plan and that's good enough for me.

stoty:
I've just run a smoke test with a local install of HBase 2.3.1 and this branch of Phoenix, and it passed 
(!tables, create table, upsert, select)

ankitsinghal:
Changes look good but are there any change in the public API( or utility or tools) which breaks a compatibility contract for minor releases?

stoty:
That's a good question.

Our main public API is JDBC, which doesn't have Guava in it.
org.apache.phoenix.expression.function.ScalarFunction for UDFs doesn't have Guava in it either.
Our tools are mainly the Index and bulk load related MR jobs, which are part of the release, and don't have programmatic APIs.

I am not sure what other interfaces would count as public API, as we don't use @InterfaceAudience

One change of note is the use of (Tephra's) shaded guava classes for the shaded Tephra API, as Tephra DOES have guava in its Phoenix-facing API. However, this change is introduced in a new maven Tephra artifact, and the old Tephra artifact with unshaded Guava is still available in Tephra.

stoty:
The connectors project builds without changes when using patched Phoenix.
The APIs used by connectors is the closest thing I can think of to LimitedPublic interfaces that we have.

ankitsinghal:
>org.apache.phoenix.expression.function.ScalarFunction for UDFs doesn't have Guava in it either.
> The connectors project builds without changes when using patched Phoenix.
> The APIs used by connectors is the closest thing I can think of to LimitedPublic interfaces that we have.

Thanks Istvan for checking the built-in functions and connectors, I think all these are good validations. I also checked all *IT.java and it seems guava migration changes in them is just for helpers like Lists, Maps, etc .. so It seems we are good here.

ankitsinghal (approved):
+1, great work in overall migration @stoty , thanks

PR #846, opened 2020-07-27
PHOENIX-6045 Delete that should qualify for index path does not use i…

kadirozde:
…ndex when multiple indexes are available.

### DIFF
--- a/phoenix-core/src/main/java/org/apache/phoenix/compile/DeleteCompiler.java
+++ b/phoenix-core/src/main/java/org/apache/phoenix/compile/DeleteCompiler.java
@@ -89,11 +89,13 @@ import org.apache.phoenix.schema.types.PLong;
 import org.apache.phoenix.transaction.PhoenixTransactionProvider.Feature;
 import org.apache.phoenix.util.ByteUtil;
 import org.apache.phoenix.util.IndexUtil;
+import org.apache.phoenix.util.PhoenixRuntime;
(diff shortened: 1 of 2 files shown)


dbwong:
FYI @lhofhansl

phoenix-core/src/main/java/org/apache/phoenix/compile/DeleteCompiler.java
```diff
@@ -486,19 +494,22 @@ else if (table.isTransactional() && connection.getSCN() != null) {
             projectedColumns.add(column);
             aliasedNodes.add(FACTORY.aliasedNode(null, FACTORY.column(null, '"' + column.getName().getString() + '"', null)));
         }
-        // Project all non PK indexed columns so that we can do the proper index maintenance
+        // Project all non PK indexed columns so that we can do the proper index maintenance on the indexes for which
+        // mutations are generated on the client side. Indexed columns are needed to identify index rows to be deleted
         for (PTable index : table.getIndexes()) {
```
comnetwork, line 499: Perhaps use `clientSideIndexes`  here is better?

kadirozde: Possibly but I have tested this code and do not want to go through the whole testing cycle which includes testing on a real cluster as this is a production issue for us now. Thanks for the suggestion

PR #848, opened 2020-07-28
Phoenix-5947 Add tests and extensions to Schema Extraction Tool utility

Qinrui98:
1. Added support for extracting ddl for salted tables, array column type and non-default column families
2. Added tests for extracting table, view and index ddls


phoenix-tools/src/main/java/org/apache/phoenix/schema/SchemaExtractionProcessor.java
```diff
@@ -54,9 +55,6 @@ public String process() throws Exception {
         if (ddl != null) {
             return ddl;
         }
-        if(this.table.getColumnFamilies().size() > 1 ) {
```
swaroopak, line 57: This PR does not completely support the multiple column families, especially if there are different properties for each column family as mentioned by Goeffrey on the previous review. Let's modify this if clause to be specific unless you want to include the change in this one and completely get rid of "Not Supported Exception".

phoenix-tools/src/main/java/org/apache/phoenix/schema/SchemaExtractionProcessor.java
```diff
@@ -1,35 +1,35 @@
...
-import java.util.Map;
-import java.util.Set;
+import java.util.*;
 
 public class SchemaExtractionProcessor {
 
-    public static final String
-            FEATURE_NOT_SUPPORTED_ON_EXTRACTION_TOOL =
-            "Multiple CF feature not supported on extraction tool";
+    public static final List<String> SYNCED_DATA_TABLE_AND_INDEX_COL_FAM_PROPERTIES = ImmutableList.of(
```
swaroopak, line 29: This should be available in other parts of the code. You can just import here.

phoenix-tools/src/main/java/org/apache/phoenix/schema/SchemaExtractionProcessor.java
```diff
@@ -1,35 +1,35 @@
...
 import java.io.IOException;
 import java.sql.Connection;
 import java.sql.SQLException;
-import java.util.ArrayList;
-import java.util.HashMap;
-import java.util.HashSet;
-import java.util.List;
-import java.util.Map;
-import java.util.Set;
+import java.util.*;
```
swaroopak, line 25: Avoid import *

phoenix-tools/src/main/java/org/apache/phoenix/schema/SchemaExtractionProcessor.java
```diff
@@ -250,12 +247,31 @@ private void setHTableProperties(HTableDescriptor htd) {
...
-        Map<ImmutableBytesWritable, ImmutableBytesWritable> propsMap = columnDescriptor.getValues();
-        for (Map.Entry<ImmutableBytesWritable, ImmutableBytesWritable> entry : propsMap.entrySet()) {
+    private void setHColumnFamilyProperties(HColumnDescriptor[] columnDescriptors) {
+        Map<ImmutableBytesWritable, ImmutableBytesWritable> propsMap = columnDescriptors[0].getValues();
+        for(Map.Entry<ImmutableBytesWritable, ImmutableBytesWritable> entry : propsMap.entrySet()){
             ImmutableBytesWritable key = entry.getKey();
-            ImmutableBytesWritable value = entry.getValue();
-            definedProps.put(Bytes.toString(key.get()), Bytes.toString(value.get()));
+            ImmutableBytesWritable defaultValue = entry.getValue();
+            Map<String, String> CFMap = new HashMap<String, String>();
```
swaroopak, line 255: nit: cfMap

phoenix-tools/src/main/java/org/apache/phoenix/schema/SchemaExtractionProcessor.java
```diff
@@ -250,12 +247,31 @@ private void setHTableProperties(HTableDescriptor htd) {
...
+            Set<ImmutableBytesWritable> set = new HashSet<ImmutableBytesWritable>();
+            for(HColumnDescriptor columnDescriptor: columnDescriptors){
```
swaroopak, line 257: space before `{`

phoenix-tools/src/main/java/org/apache/phoenix/schema/SchemaExtractionProcessor.java
```diff
@@ -250,12 +247,31 @@ private void setHTableProperties(HTableDescriptor htd) {
...
+                ImmutableBytesWritable value = columnDescriptor.getValues().get(key);
+                // check if it is universal properties
+                if (SYNCED_DATA_TABLE_AND_INDEX_COL_FAM_PROPERTIES.contains(Bytes.toString(key.get()))){
+                    definedProps.put(Bytes.toString(key.get()), Bytes.toString(value.get()));
+                    continue;
+                }
+                CFMap.put(columnFamilyName, Bytes.toString(value.get()));
+                set.add(value);
+            }
+            if (set.size() > 1){
```
swaroopak, line 268: space before {

phoenix-tools/src/main/java/org/apache/phoenix/schema/SchemaExtractionProcessor.java
```diff
@@ -250,12 +247,31 @@ private void setHTableProperties(HTableDescriptor htd) {
...
-    private void setHColumnFamilyProperties(HColumnDescriptor columnDescriptor) {
-        Map<ImmutableBytesWritable, ImmutableBytesWritable> propsMap = columnDescriptor.getValues();
-        for (Map.Entry<ImmutableBytesWritable, ImmutableBytesWritable> entry : propsMap.entrySet()) {
+    private void setHColumnFamilyProperties(HColumnDescriptor[] columnDescriptors) {
+        Map<ImmutableBytesWritable, ImmutableBytesWritable> propsMap = columnDescriptors[0].getValues();
+        for(Map.Entry<ImmutableBytesWritable, ImmutableBytesWritable> entry : propsMap.entrySet()){
             ImmutableBytesWritable key = entry.getKey();
-            ImmutableBytesWritable value = entry.getValue();
-            definedProps.put(Bytes.toString(key.get()), Bytes.toString(value.get()));
+            ImmutableBytesWritable defaultValue = entry.getValue();
```
swaroopak, line 254: let's call it globalValue

phoenix-tools/src/main/java/org/apache/phoenix/schema/SchemaExtractionProcessor.java
swaroopak, line 268: this would fail if two CF have same non-default value for `VERSIONS` or any other property

phoenix-tools/src/main/java/org/apache/phoenix/schema/SchemaExtractionProcessor.java
```diff
@@ -250,12 +247,31 @@ private void setHTableProperties(HTableDescriptor htd) {
...
+            Map<String, String> CFMap = new HashMap<String, String>();
+            Set<ImmutableBytesWritable> set = new HashSet<ImmutableBytesWritable>();
```
swaroopak, line 256: nit: cfPropertyValueSet

phoenix-tools/src/main/java/org/apache/phoenix/schema/SchemaExtractionProcessor.java
swaroopak, line 268: Let's say there are 3 CFs and only 2 of them have VERSIONS=2. Would this work? Let's add a test if we don't have one to cover this scenario.

phoenix-tools/src/main/java/org/apache/phoenix/schema/SchemaExtractionProcessor.java
```diff
@@ -250,12 +247,31 @@ private void setHTableProperties(HTableDescriptor htd) {
...
+            for(HColumnDescriptor columnDescriptor: columnDescriptors){
+                String columnFamilyName = Bytes.toString(columnDescriptor.getName());
+                ImmutableBytesWritable value = columnDescriptor.getValues().get(key);
+                // check if it is universal properties
+                if (SYNCED_DATA_TABLE_AND_INDEX_COL_FAM_PROPERTIES.contains(Bytes.toString(key.get()))){
+                    definedProps.put(Bytes.toString(key.get()), Bytes.toString(value.get()));
+                    continue;
+                }
+                CFMap.put(columnFamilyName, Bytes.toString(value.get()));
+                set.add(value);
```
swaroopak, line 266: I think we can maintain a map instead of [propertyValue and occurances] and execute line no. 273 only when exactly one entry from the map has occurrences equal to the no. of column families. 
Also, good to refactor this logic in another function.

phoenix-tools/src/main/java/org/apache/phoenix/schema/SchemaExtractionProcessor.java
```diff
@@ -374,20 +397,49 @@ private String getColumnInfoStringForView(PTable table, PTable baseTable) {
...
-            buf.append(maxLength);
-            Integer scale = column.getScale();
-            if (scale != null) {
-                buf.append(',');
-                buf.append(scale); // has both max length and scale. For ex- decimal(10,2)
+
+        if (isArrayType) {
+            String arrayPrefix = type.split("\\s+")[0];
+            buf.append(arrayPrefix);
+            if (maxLength != null) {
```
swaroopak, line 416: nit: this if condition is duplicated in the else part. Please refactor it by extracting common parts and executing specific Array related logic when if clause passes.

phoenix-tools/src/it/java/org/apache/phoenix/schema/SchemaExtractionToolIT.java
```diff
@@ -210,4 +206,188 @@ public void testCreateViewIndexStatement() throws Exception {
...
+
+            SchemaExtractionTool set = new SchemaExtractionTool();
+            set.setConf(conn.unwrap(PhoenixConnection.class).getQueryServices().getConfiguration());
+            String [] args2 = {"-tb", indexName, "-s", schemaName};
+            set.run(args2);
+            Assert.assertEquals(createIndexStatement.toUpperCase(), set.getOutput().toUpperCase());
+        }
+    }
+
+    private boolean compareProperties(String prop1, String prop2){
```
swaroopak, line 384: Thanks for adding this.

phoenix-tools/src/it/java/org/apache/phoenix/schema/SchemaExtractionToolIT.java
```diff
@@ -33,22 +30,21 @@ public void testCreateTableStatement() throws Exception {
...
+            conn.createStatement().execute(createTable);
             conn.commit();
             String [] args = {"-tb", tableName, "-s", schemaName};
 
             SchemaExtractionTool set = new SchemaExtractionTool();
             set.setConf(conn.unwrap(PhoenixConnection.class).getQueryServices().getConfiguration());
             set.run(args);
-            String actualProperties = set.getOutput().substring(set.getOutput().lastIndexOf(")")+1).replace(" ","");
-            Assert.assertEquals(3, actualProperties.split(",").length);
+            //String actualProperties = set.getOutput().substring(set.getOutput().lastIndexOf(")")+1).replace(" ","");
```
swaroopak, line 46: nit: remove the commented code

phoenix-tools/src/it/java/org/apache/phoenix/schema/SchemaExtractionToolIT.java
```diff
@@ -210,4 +206,188 @@ public void testCreateViewIndexStatement() throws Exception {
...
+                    "\"cv\".\"_\" CHAR(1), " +
+                    "\"dv\".\"_\" CHAR(1) CONSTRAINT PK PRIMARY KEY (a_char, b_char)) " + properties;
+            conn.createStatement().execute(query);
+            conn.commit();
+            String[] args = {"-tb", tableName, "-s", schemaName};
+            SchemaExtractionTool set = new SchemaExtractionTool();
+            set.setConf(conn.unwrap(PhoenixConnection.class).getQueryServices().getConfiguration());
+            set.run(args);
+
+            Assert.assertEquals(true, compareProperties(properties, set.getOutput().substring(set.getOutput().lastIndexOf(")")+1)));
```
swaroopak, line 359: can use assertTrue directly.

phoenix-tools/src/it/java/org/apache/phoenix/schema/SchemaExtractionToolIT.java
```diff
@@ -33,22 +30,21 @@ public void testCreateTableStatement() throws Exception {
         String tableName = generateUniqueName();
         String schemaName = generateUniqueName();
         Properties props = PropertiesUtil.deepCopy(TEST_PROPERTIES);
+        String pTableFullName = SchemaUtil.getQualifiedTableName(schemaName, tableName);
         String properties = "TTL=2592000,IMMUTABLE_ROWS=true,DISABLE_WAL=true";
```
swaroopak, line 34: nit: could you please remove this variable? It is not used anymore.

phoenix-tools/src/main/java/org/apache/phoenix/schema/SchemaExtractionProcessor.java
Qinrui98: Yes it's working with only 2 of them having VERSIONS=2. The size of the set will be 2 in this case. I'll add a test to cover this scenario.

phoenix-tools/src/main/java/org/apache/phoenix/schema/SchemaExtractionProcessor.java
```diff
@@ -250,12 +247,31 @@ private void setHTableProperties(HTableDescriptor htd) {
...
+                if (SYNCED_DATA_TABLE_AND_INDEX_COL_FAM_PROPERTIES.contains(Bytes.toString(key.get()))) {
+                    definedProps.put(Bytes.toString(key.get()), Bytes.toString(value.get()));
+                    break;
+                }
+                cfToPropertyValueMap.put(columnFamilyName, Bytes.toString(value.get()));
+                cfPropertyValueSet.add(value);
+            }
+            if (cfPropertyValueSet.size() > 1) {
+                for(Map.Entry<String, String> mapEntry: cfToPropertyValueMap.entrySet()) {
+                    definedProps.put(String.format("%s.%s",  mapEntry.getKey(), Bytes.toString(key.get())), mapEntry.getValue());
```
gjacoby126, line 270: I'm curious why we're qualifying the CF properties in one flat HashSet instead of a HashSet<String, HashSet<String, String>>? Can we guarantee that no CF properties will ever have a period in them?

phoenix-tools/src/main/java/org/apache/phoenix/schema/SchemaExtractionProcessor.java
```diff
@@ -312,9 +336,8 @@ public Connection getConnection() throws SQLException {
 
     private String getColumnInfoStringForTable(PTable table) {
         StringBuilder colInfo = new StringBuilder();
-
-        List<PColumn> columns = table.getColumns();
-        List<PColumn> pkColumns = table.getPKColumns();
+        List<PColumn> columns = table.getBucketNum() == null ? table.getColumns() : table.getColumns().subList(1, table.getColumns().size());
```
gjacoby126, line 339: Good catch for salted tables, those are easy to miss.

phoenix-tools/src/main/java/org/apache/phoenix/schema/SchemaExtractionProcessor.java
Qinrui98: @gjacoby126 We are trying to find common CF property so that we don't need to specify it for every column family in the output string. Are you suggesting that we could use hashmap instead?

phoenix-tools/src/it/java/org/apache/phoenix/schema/SchemaExtractionToolIT.java
```diff
@@ -210,4 +204,242 @@ public void testCreateViewIndexStatement() throws Exception {
...
+            String actualProperties = set.getOutput().substring(set.getOutput().lastIndexOf(")")+1);
+            Assert.assertEquals(true, actualProperties.contains("SALT_BUCKETS=16"));
+        }
+    }
+
+    @Test
+    public void testCreateTableWithPKConstraint() throws Exception {
+        String tableName = generateUniqueName();
+        String schemaName = generateUniqueName();
+        Properties props = PropertiesUtil.deepCopy(TEST_PROPERTIES);
```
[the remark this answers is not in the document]
Qinrui98: Yes I'll extract the common part into a helper function and avoid duplicated code here.

phoenix-tools/src/it/java/org/apache/phoenix/schema/SchemaExtractionToolIT.java
```diff
@@ -58,156 +130,251 @@ public void testCreateTableStatement_tenant() throws Exception {
...
+    }
 
-            String pTableFullName = SchemaUtil.getQualifiedTableName(schemaName, tableName);
-            conn.createStatement().execute("CREATE TABLE "+pTableFullName + "(k BIGINT NOT NULL PRIMARY KEY, "
-                    + "v1 VARCHAR, v2 VARCHAR)"
-                    + properties);
-            String viewFullName = SchemaUtil.getQualifiedTableName(schemaName, viewName);
-            String viewFullName1 = SchemaUtil.getQualifiedTableName(schemaName, viewName+"1");
+    @Test
+    public void testCreateTableWithUniversalCFProperties() throws Exception {
```
swaroopak, line 215: I feel this and testCreateTableWithNonDefaultColumnFamily can be clubbed into one test?

phoenix-tools/src/it/java/org/apache/phoenix/schema/SchemaExtractionToolIT.java
```diff
@@ -58,156 +130,251 @@ public void testCreateTableStatement_tenant() throws Exception {
...
+                "\"cv\".\"_\" CHAR(1), " +
+                "\"dv\".\"_\" CHAR(1) CONSTRAINT PK PRIMARY KEY (a_char, b_char)) " + properties;
+        List<String> queries = new ArrayList<String>(){};
+        queries.add(query);
+        String result = runSchemaExtractionTool(schemaName, tableName, null, queries);
+        Assert.assertEquals(query.toUpperCase(), result.toUpperCase());
+    }
 
+    @Test
+    public void testCreateTableWithDefaultCFProperties() throws Exception {
```
swaroopak, line 234: even this can be clubbed with above one as there is only one extra property here.

PR #857, opened 2020-08-14
PHOENIX-5881 - Port PHOENIX-5645 (MaxLookbackAge) to 5.x

gjacoby126:


gjacoby126:
Note that since HBase 2.1 and HBase 2.2 lack HBASE-24321, a lot of the logic is behind a 2.3 compatibility shim.

gjacoby126:
@stoty - I'm curious about your opinion on how I'm using your compatibility projects to get version-specific coproc logic, plus version-specific feature flags.

stoty:
In 4.x we use org.apache.phoenix.compat.hbase.HbaseCompatCapabilities for the feature flags, It would be more consistent to move the feature flags there.

I did not go through the code in detail, but it seems that extending BaseScannerRegionObserver from CompatBaseScannerRegionObserver would be a more elegant , and perhaps marginally faster solution than calling it explicitly.

gjacoby126:
@stoty - introduced HbaseCompatCapabilities and subclassed from ComaptBaseScannerRegionObserver, as you suggested.

phoenix-core/src/main/java/org/apache/hadoop/hbase/regionserver/ScanInfoUtil.java
```diff
@@ -0,0 +1,24 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.hadoop.hbase.regionserver;
+
+import org.apache.hadoop.conf.Configuration;
+
+public class ScanInfoUtil {
```
stoty, line 22: This doesn't seem to reference anywhere.

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/BaseScannerRegionObserver.java
```diff
@@ -367,4 +374,16 @@ RegionScanner getWrappedScanner(final ObserverContext<RegionCoprocessorEnvironme
                 dataRegion, indexMaintainer, null, viewConstants, null, null, projector, ptr, useQualiferAsListIndex);
     }
 
+    @Override
+    public void preStoreScannerOpen(ObserverContext<RegionCoprocessorEnvironment> ctx, Store store,
```
stoty, line 378: This doesn't seem to do anything since the subclassing.

phoenix-hbase-compat-2.2.1/src/main/java/org/apache/phoenix/compat/hbase/HbaseCompatCapabilities.java
```diff
@@ -0,0 +1,32 @@
...
+import org.apache.hadoop.conf.Configuration;
+
+public class HbaseCompatCapabilities {
+
+    public static boolean isMaxLookbackTimeSupported() {
+        return false;
+    }
+
+    //In HBase 2.1 and 2.2, a lookback query won't return any results if covered by a future delete
+    public static boolean isLookbackBeyondDeletesSupported() { return false; }
```
stoty, line 30: Nit: indent like above in both files?

stoty (approved):
+1 LGTM

phoenix-core/src/main/java/org/apache/phoenix/compile/QueryCompiler.java
```diff
@@ -172,6 +178,28 @@ public QueryPlan compile() throws SQLException{
...
+        if (scn == null) {
+            return;
+        }
+        ColumnResolver resolver =
+            FromCompiler.getResolverForQuery(select, conn);
+        int maxLookBackAge = conn.getQueryServices().
+            getConfiguration().getInt(CompatBaseScannerRegionObserver.PHOENIX_MAX_LOOKBACK_AGE_CONF_KEY,
+            CompatBaseScannerRegionObserver.DEFAULT_PHOENIX_MAX_LOOKBACK_AGE);
+        long now = EnvironmentEdgeManager.currentTimeMillis();
+        if (maxLookBackAge > 0 && now - maxLookBackAge * 1000L > scn){
```
swaroopak, line 196: nit: would be good in understanding if s -> ms conversion happens outside of if condition.

gjacoby126: Not sure I follow? The configuration property is in seconds (to correspond with HBase TTL, which is also in seconds), but we convert to ms in the if-block to compare with the current timestamp.

swaroopak: maxLookBackAge*1000 can be stored into a variable like done for ttl in other function.

gjacoby126: @swaroopak Switched to using an existing helper function to get the millisecond version of the max lookback config.

gjacoby126:
Thanks for the reviews, @stoty and @swaroopak .

PR #859, opened 2020-08-17
PHOENIX-6073 - IndexTool IndexDisableLoggingType can't be set to NONE

gjacoby126:


gjacoby126:
Thanks for the reviews, @kadirozde and @swaroopak

PR #869, opened 2020-08-31
PHOENIX-6102 Better isolation for CI jobs an ASF Jenkins

stoty:
run post-commit tests in Docker
also don't mask Hadoop exception when tests break on anonymous unix user


stoty:
Tested on https://ci-hadoop.apache.org/job/Phoenix/job/Phoenix-experiment/
We lose 5-10 minutes on not having a maven cache, but we are guaranteed not to have conflicts with installed modules either.

PR #875, opened 2020-09-10
PHOENIX-6120 Covered column handling for SINGLE_CELL_ARRAY storage sc…

gokceni:
…heme


phoenix-core/src/main/java/org/apache/phoenix/index/IndexMaintainer.java
```diff
@@ -1030,35 +1032,40 @@ public Put buildUpdateMutation(KeyValueBuilder kvBuilder, ValueGetter valueGette
...
-                        public Integer getMaxLength() {
-                            return null;
-                        }
-                        
-                        @Override
-                        public PDataType getDataType() {
-                            return null;
+
+                    byte[] value = null;
+                    if (this.dataTableImmutableStorageScheme == this.immutableStorageScheme) {
```
kadirozde, line 1037: It is not clear to me how this extra check is useful here and how it fixes the issue. Why is this check necessary, given that the index is supposed to  inherit the storage scheme from the data table currently? Is it possible that they can be different.

kadirozde (approved):
+1, Thanks

gjacoby126:
@gokceni , @kadirozde , I'm curious what the status of PHOENIX-6120 is? Looks like it was approved but never merged.

gokceni:
@gjacoby126 I will close this PR and open another one.

PR #883, opened 2020-09-15
PHOENIX-6072: SYSTEM.MUTEX not created with a TTL on a fresh cluster connected to by a 4.15+ client

ChinmaySKulkarni:


ChinmaySKulkarni:
Please review @yanxinyi @jpisaac @gjacoby126 @twdsilva

phoenix-core/src/it/java/org/apache/phoenix/end2end/SystemTablesCreationOnConnectionIT.java
```diff
@@ -18,6 +18,9 @@
 package org.apache.phoenix.end2end;
```
ChinmaySKulkarni, line 18: A lot of these diffs are auto-formatting by my IDE to be according to Phoenix coding standards for whitespace, line length, etc.

ChinmaySKulkarni: Along with some changes to close statements, conns, etc. which were previously unclosed.

phoenix-core/src/main/java/org/apache/phoenix/query/QueryConstants.java
```diff
@@ -40,6 +37,125 @@
 import org.apache.phoenix.schema.SystemStatsSplitPolicy;
 import org.apache.phoenix.schema.TableProperty;
 
+import static org.apache.phoenix.jdbc.PhoenixDatabaseMetaData.APPEND_ONLY_SCHEMA;
```
ChinmaySKulkarni, line 40: Most of these changes are also after passing through a linter which pointed out that `public static final` is unnecessary since this is an interface.

phoenix-core/src/main/java/org/apache/phoenix/query/QueryConstants.java
```diff
@@ -338,48 +448,53 @@
...
+            TABLE_SCHEM + " VARCHAR NULL," +
+            TABLE_NAME + " VARCHAR NOT NULL," +
+            COLUMN_NAME + " VARCHAR NULL," + // null for table row
+            COLUMN_FAMILY + " VARCHAR NULL " + // using for CF to uniqueness for columns
+            "CONSTRAINT " + SYSTEM_TABLE_PK_NAME + " PRIMARY KEY (" + TENANT_ID + "," +
+            TABLE_SCHEM + "," + TABLE_NAME + "," + COLUMN_NAME + "," + COLUMN_FAMILY + "))\n" +
+            HConstants.VERSIONS + "=%s,\n" +
+            HColumnDescriptor.KEEP_DELETED_CELLS + "=%s,\n" +
+            TRANSACTIONAL + "=" + Boolean.FALSE + ",\n" +
+            HColumnDescriptor.TTL + "=" + TTL_FOR_MUTEX;
```
ChinmaySKulkarni, line 494: main change - part 2 (for fresh clusters starting with a 4.16 client)

phoenix-core/src/main/java/org/apache/phoenix/query/ConnectionQueryServicesImpl.java
```diff
@@ -3286,6 +3288,43 @@ void createSysMutexTableIfNotExists(HBaseAdmin admin) throws IOException, SQLExc
...
 
+    /**
+     * Check if the SYSTEM MUTEX table exists. If it does, ensure that its TTL is correct and if
+     * not, modify its table descriptor
+     * @param admin HBase admin
+     * @return true if SYSTEM MUTEX exists already and false if it needs to be created
+     * @throws IOException thrown if there is an error getting the table descriptor
+     */
+    @VisibleForTesting
+    boolean checkIfSysMutexExistsAndModifyTTLIfRequired(HBaseAdmin admin) throws IOException {
```
ChinmaySKulkarni, line 3299: main change - part 1 (for the upgrade path used for clusters that have only seen 4.15 clients and will have a "FOREVER" TTL on SYSTEM.MUTEX)

phoenix-core/src/main/java/org/apache/phoenix/jdbc/PhoenixDatabaseMetaData.java
```diff
@@ -224,6 +224,7 @@
     public static final String TASK_PRIORITY = "TASK_PRIORITY";
     public static final String TASK_DATA = "TASK_DATA";
     public static final String TASK_TABLE_TTL = "864000";
+    public static final int TTL_FOR_MUTEX = 15 * 60; // 15min
```
yanxinyi, line 227: just wondering, why 15 mins instead of 10 mins or 20 mins.

yanxinyi:
overall looks good to me, but please let other ppl review as well.

phoenix-core/src/main/java/org/apache/phoenix/jdbc/PhoenixDatabaseMetaData.java
ChinmaySKulkarni: I haven't changed the value of the TTL. It was 15 min originally

phoenix-core/src/main/java/org/apache/phoenix/query/ConnectionQueryServicesImpl.java
```diff
@@ -295,7 +300,6 @@
             LoggerFactory.getLogger(ConnectionQueryServicesImpl.class);
     private static final int INITIAL_CHILD_SERVICES_CAPACITY = 100;
     private static final int DEFAULT_OUT_OF_ORDER_MUTATIONS_WAIT_TIME_MS = 1000;
-    private static final int TTL_FOR_MUTEX = 15 * 60; // 15min
```
ChinmaySKulkarni, line 298: @yanxinyi here is where it was defined earlier.

phoenix-core/src/main/java/org/apache/phoenix/jdbc/PhoenixDatabaseMetaData.java
twdsilva: We just picked a long enough time so that the client that created the mutex would be able to complete the metadata operation.

phoenix-core/src/main/java/org/apache/phoenix/query/ConnectionQueryServicesImpl.java
```diff
@@ -3286,6 +3288,43 @@ void createSysMutexTableIfNotExists(HBaseAdmin admin) throws IOException, SQLExc
...
+                // Try with the namespace mapping name
+                htd = admin.getTableDescriptor(TableName.valueOf(SYSTEM_SCHEMA_NAME,
+                        SYSTEM_MUTEX_TABLE_NAME));
+            } catch (org.apache.hadoop.hbase.TableNotFoundException ignored2) {
+                return false;
+            }
+        }
+
+        // The SYSTEM MUTEX table already exists so check its TTL
+        if (htd.getFamily(SYSTEM_MUTEX_FAMILY_NAME_BYTES).getTimeToLive() != TTL_FOR_MUTEX) {
```
twdsilva, line 3314: I assume the ``SYSTEM_MUTEX_FAMILY_NAME_BYTES`` column family should always exist.

twdsilva (approved):
LGTM

phoenix-core/src/main/java/org/apache/phoenix/query/ConnectionQueryServicesImpl.java
ChinmaySKulkarni: @twdsilva yes, it is basically `DEFAULT_COLUMN_FAMILY` i.e. "0"

phoenix-core/src/main/java/org/apache/phoenix/query/ConnectionQueryServicesImpl.java
ChinmaySKulkarni: FYI @jpisaac

phoenix-core/src/main/java/org/apache/phoenix/query/QueryConstants.java
jpisaac: @ChinmaySKulkarni This will require a restart of the region server/region/store that holds the data for the new TTL to get in effect. Since the TTL value is set when the Store/Region is initialized.

phoenix-core/src/main/java/org/apache/phoenix/query/ConnectionQueryServicesImpl.java
```diff
@@ -3286,6 +3288,43 @@ void createSysMutexTableIfNotExists(HBaseAdmin admin) throws IOException, SQLExc
...
+            LOGGER.debug("SYSTEM MUTEX already appears to exist, but has the wrong TTL. " +
+                    "Will modify the TTL");
+            HColumnDescriptor hColFamDesc = htd.removeFamily(SYSTEM_MUTEX_FAMILY_NAME_BYTES);
+            hColFamDesc.setTimeToLive(TTL_FOR_MUTEX);
+            htd.addFamily(hColFamDesc);
+            admin.modifyTable(htd.getTableName(), htd);
```
jpisaac, line 3320: Does this reload the region/store, not sure offhand, but good to check, if yes then ignore my previous on comment about region restart/reload.

jpisaac:
LGTM, other than the modifyTable check.

ChinmaySKulkarni:
Thanks for the reviews @twdsilva @yanxinyi @jpisaac.

ChinmaySKulkarni:
@jpisaac verified that the TTL is honored when modified via admin API

PR #887, opened 2020-09-17
PHOENIX-6143: Get Phoenix Tracing Webapp work

tamaashu:
Tracing webapp was broken for a while, these changes get it work.

Change-Id: I81b20d9abff220b01a984af3722ee57fef316621


phoenix-tracing-webapp/pom.xml
```diff
@@ -155,6 +131,7 @@
         <plugin>
           <groupId>com.github.searls</groupId>
           <artifactId>jasmine-maven-plugin</artifactId>
+          <version>2.2</version>
```
stoty, line 134: You are overriding the version set in the master pom pluginManagement section.
I suggest that you override the plugin version there instead.

phoenix-tracing-webapp/src/build/trace-server-runnable.xml
```diff
@@ -43,7 +43,19 @@
       <outputDirectory>/</outputDirectory>
       <includes>
         <include>org.apache.phoenix:phoenix-tracing-webapp</include>
```
stoty, line 45: Ideally, this should be done by the shade plugin instead. (Not required for this patch, just a remark)

phoenix-tracing-webapp/README.md
```diff
@@ -3,10 +3,10 @@
  *mvn clean install*
```
stoty, line 3: Nit: We use backticks to format command lines elsewhere, might as well switch to that here, as well.

stoty (changes requested):

stoty (approved):

PR #889, opened 2020-09-22
PHOENIX-6148 & PHOENIX-6149 : [SchemaExtractionTool] DDL parsing exception in Phoenix…

swaroopak:
… in view name

### DIFF
--- a/phoenix-tools/src/main/java/org/apache/phoenix/schema/SchemaExtractionProcessor.java
+++ b/phoenix-tools/src/main/java/org/apache/phoenix/schema/SchemaExtractionProcessor.java
@@ -17,8 +17,10 @@
  */
 package org.apache.phoenix.schema;
 
+import com.google.common.base.Strings;
 import com.google.common.collect.Sets;
 import org.apache.commons.lang.StringUtils;
+import org.apache.commons.lang.math.NumberUtils;
 import org.apache.hadoop.conf.Configuration;
 import org.apache.hadoop.hbase.HColumnDescriptor;
 import org.apache.hadoop.hbase.HTableDescriptor;
@@ -202,7 +204,7 @@ public class SchemaExtractionProcessor {
 
     private String generateCreateViewDDL(String columnInfoString, String baseTableFullName,
             String whereClause, String pSchemaName, String pTableName) {
-        String viewFullName = SchemaUtil.getQualifiedTableName(pSchemaName, pTableName);
+        String viewFullName = SchemaUtil.getPTableFullNameWithQuotes(pSchemaName, pTableName);
         StringBuilder outputBuilder = new StringBuilder(String.format(CREATE_VIEW, viewFullName,
                 columnInfoString, baseTableFullName, whereClause));
         return outputBuilder.toString();
@@ -225,9 +227,10 @@ public class SchemaExtractionProcessor {
 
         return generateTableDDLString(columnInfoString, propertiesString, pSchemaName, pTableName);
     }
+
     private String generateTableDDLString(String columnInfoString, String propertiesString,
             String pSchemaName, String pTableName) {
-        String pTableFullName = SchemaUtil.getQualifiedTableName(pSchemaName, pTableName);
+        String pTableFullName = SchemaUtil.getPTableFullNameWithQuotes(pSchemaName, pTableName);
         StringBuilder outputBuilder = new StringBuilder(String.format(CREATE_TABLE, pTableFullName));
         outputBuilder.append(columnInfoString).append(" ").append(propertiesString);
         return outputBuilder.toString();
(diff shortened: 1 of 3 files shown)


phoenix-tools/src/main/java/org/apache/phoenix/schema/SchemaExtractionProcessor.java
```diff
@@ -203,6 +203,9 @@ protected String extractCreateViewDDL(PTable table) throws SQLException {
     private String generateCreateViewDDL(String columnInfoString, String baseTableFullName,
             String whereClause, String pSchemaName, String pTableName) {
         String viewFullName = SchemaUtil.getQualifiedTableName(pSchemaName, pTableName);
+        if(Character.isDigit(pTableName.charAt(0))) {
```
gjacoby126, line 206: Is this a bug in SchemaUtil.getQualfiedTableName? This seems like something we shouldn't have to repeat everywhere.

swaroopak: I think not, this is mainly for parsing. For Internals of phoenix, function should ignore quotes around the name.

tkhurana: If you look at the phoenix grammar for create table or create view the table name is of the format below: https://phoenix.apache.org/language/index.html#table_ref
https://phoenix.apache.org/language/index.html#name

So table names can begin with [A-Z] or _ so any table name that doesn't begin with this has to begin with a quote. So just checking for `isDigit` is wrong as we might miss other cases where the first letter is something other than a digit for example like a `$` or other character. The fix should be if the first character is not a letter or  underscore then quote.

swaroopak: Makes sense! Let me correct it.

phoenix-tools/src/main/java/org/apache/phoenix/schema/SchemaExtractionProcessor.java
```diff
@@ -202,12 +202,20 @@ protected String extractCreateViewDDL(PTable table) throws SQLException {
...
     private String generateCreateViewDDL(String columnInfoString, String baseTableFullName,
             String whereClause, String pSchemaName, String pTableName) {
-        String viewFullName = SchemaUtil.getQualifiedTableName(pSchemaName, pTableName);
+        String viewFullName = getPTableFullName(pSchemaName, pTableName);
         StringBuilder outputBuilder = new StringBuilder(String.format(CREATE_VIEW, viewFullName,
                 columnInfoString, baseTableFullName, whereClause));
         return outputBuilder.toString();
     }
 
+    private String getPTableFullName(String pSchemaName, String pTableName) {
```
priyankporwal, line 211: Why not add this function to SchemaUtil class itself?

swaroopak: This is not needed in for internals for phoenix.

swaroopak: This is not needed in internals of phoenix.

priyankporwal: Maybe so today, but we never know when it might be needed later. SchemaUtil seems like the right place for it instead of fragmented/duplicated in various tools.

phoenix-tools/src/main/java/org/apache/phoenix/schema/SchemaExtractionProcessor.java
```diff
@@ -326,8 +334,13 @@ private String convertPropertiesToString() {
                 if (optionBuilder.length() != 0) {
                     optionBuilder.append(", ");
                 }
-                key = columnFamilyName.equals(QueryConstants.DEFAULT_COLUMN_FAMILY)? key : String.format("\"%s\".%s", columnFamilyName, key);
-                optionBuilder.append(key+"="+value);
+                key = columnFamilyName.equals(QueryConstants.DEFAULT_COLUMN_FAMILY)?
+                        key : String.format("\"%s\".%s", columnFamilyName, key);
+                if(!Character.isDigit(value.charAt(0)) &&
```
priyankporwal, line 339: what does digit in the first position signify? Perhaps add a comment as well.

swaroopak: added a comment.

phoenix-tools/src/main/java/org/apache/phoenix/schema/SchemaExtractionProcessor.java
```diff
@@ -326,8 +327,15 @@ private String convertPropertiesToString() {
                 if (optionBuilder.length() != 0) {
                     optionBuilder.append(", ");
                 }
-                key = columnFamilyName.equals(QueryConstants.DEFAULT_COLUMN_FAMILY)? key : String.format("\"%s\".%s", columnFamilyName, key);
-                optionBuilder.append(key+"="+value);
+                key = columnFamilyName.equals(QueryConstants.DEFAULT_COLUMN_FAMILY)?
+                        key : String.format("\"%s\".%s", columnFamilyName, key);
+                // properties value that corresponds to a number will not need single quotes around it
```
priyankporwal, line 332: Nit: given this comment, shouldn't we check that the string beginning with a digit is actually a number and not something like '72HOURS'? :)

swaroopak: That makes sense, I changed it.

priyankporwal (approved):

tkhurana (approved):

gokceni (approved):
+ 1

PR #891, opened 2020-09-22
PHOENIX-6055: Not matching index mutation error needs to report more …

gokceni:
…information rather than first expected mutation

### DIFF
--- a/phoenix-core/src/main/java/org/apache/phoenix/coprocessor/IndexRebuildRegionScanner.java
+++ b/phoenix-core/src/main/java/org/apache/phoenix/coprocessor/IndexRebuildRegionScanner.java
@@ -785,9 +785,10 @@ public class IndexRebuildRegionScanner extends GlobalIndexRegionScanner {
             }
             else {
                 byte[] dataKey = indexMaintainer.buildDataRowKey(new ImmutableBytesWritable(indexRow.getRow()), viewConstants);
-                String errorMsg = "Not matching index row";
+                String errorMsg = String.format("Not matching index row. expectedIndex=%d. expectedMutationSize=%d. actualIndex=%d. actualMutationSize=%d. expectedType=%s. actualType=%s",
+                        expectedIndex, expectedSize, actualIndex, actualSize, expected.getClass().getName(), (actualIndex < actualSize ? actual.getClass().getName() : "null"));
                 logToIndexToolOutputTable(dataKey, indexRow.getRow(),
-                        getTimestamp(expectedMutationList.get(0)), 0L, errorMsg, isBeforeRebuild,
+                        getTimestamp(expected), (actualIndex < actualSize ? getTimestamp(actual): 0L), errorMsg, isBeforeRebuild,
                     INVALID_ROW);
             }
             verificationPhaseResult.setInvalidIndexRowCount(verificationPhaseResult.getInvalidIndexRowCount() + 1);


gokceni:
@kadirozde

gjacoby126 (approved):
+1

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/IndexRebuildRegionScanner.java
```diff
@@ -768,7 +768,8 @@ public boolean verifySingleIndexRow(Result indexRow, IndexToolVerificationResult
                 }
             } else {
                 byte[] dataKey = indexMaintainer.buildDataRowKey(new ImmutableBytesWritable(indexRow.getRow()), viewConstants);
-                String errorMsg = "Not matching index row";
+                String errorMsg = String.format("Not matching index row. matchingCount=0. expectedIndex=%d. expectedMutationSize=%d. actualIndex=%d. actualMutationSize=%d. expectedTs=%d. actualTs=%d",
```
kadirozde, line 771: The version of this file is not the most recent as I noticed that it does not have the MaxLookBack changes. The patch with the latest version should be different, I think. Also, there is no need to include expected and actual timestamps in the error message since they are reported as separate columns in the output table. We should fix those reported values instead. Please use the most recent version of IndexRebuildRegionScanner.java and resubmit your patch.

gokceni: @kadirozde Rebased. I don't think we put the actualTs since in the next line we always put 0L. For the expected, since we put expectedMutations.get(0) I don't think it hurts to put what it is here in case we change the above loop and since we put the actualTs. what do you think?

swaroopak (approved):

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/IndexRebuildRegionScanner.java
kadirozde: @gokceni, this fix is not correct and will not be very useful for troubleshooting. First, the check actual == null is not meaningful here. You need to check if actualIndex < actualSIze then actual is useful and the we should report the type of actual (put or delete) and the timestamp of the actual.  I still think we absolutely should not include the timestamps in the error message as we have separate columns for them. In the following line expectedMutationList.get(0)) should be replaced by getTimestamp(expected)). expected cannot be null here. If actualIndex < actualSIze then getTimestamp(actual)) should be reported instead of 0L.

kadirozde (approved):
+1

stoty:
:broken_heart: **-1 overall**






| Vote | Subsystem | Runtime | Comment |
|:----:|----------:|--------:|:--------|
| +0 :ok: |  reexec  |   0m  0s |  Docker mode activated.  |
| -1 :x: |  patch  |   0m  4s |  https://github.com/apache/phoenix/pull/891 does not apply to master. Rebase required? Wrong Branch? See https://yetus.apache.org/documentation/in-progress/precommit-patchnames for help.  |


| Subsystem | Report/Notes |
|----------:|:-------------|
| GITHUB PR | https://github.com/apache/phoenix/pull/891 |
| JIRA Issue | PHOENIX-6055 |
| Console output | https://ci-hadoop.apache.org/job/Phoenix/job/Phoenix-PreCommit-GitHub-PR/job/PR-891/1/console |
| versions | git=2.17.1 |
| Powered by | Apache Yetus 0.12.0 https://yetus.apache.org |


This message was automatically generated.

gjacoby126:
JIRA is resolved, closing PR

PR #892, opened 2020-09-22
PHOENIX-5909 - Table and index-level metrics for indexing coprocs

gjacoby126:


gjacoby126:
This is a copy of #786 which is already approved, ported over to the master branch. Tests passed, aside from one which crashed and which passed for me locally. Merging.

PR #894, opened 2020-09-24
PHOENIX-6152 Add phoenix-shaded-protobuf to phoenix-thirdparty.

richardantal:


stoty:
I don't think the current patch switches protobuf over to the shaded version.
You are importing the new package, but not using it anywhere, but rely on the protobuf from the Hadoop dependency instead.

You would need to replace every org.apache.protobuf reference with the shaded package, and probably also regenerate the /phoenix/phoenix-core/src/main/java/org/apache/phoenix/coprocessor/generated classes.

This probably won't work, as HBase won't be able to communicate with the coprocessors, but even if it does, I'm pretty sure it would break both the HBase API contract, and Phoenix backwards compatibility, which is no-go for a minor release.

Feel free to experiment with this further, but ATM I am sceptical that this can be made to work like this.

PR #901, opened 2020-09-29
PHOENIX-6058 When maxLookback is not enabled, IndexTool should not do…

kadirozde:
… deep verification

### DIFF
--- a/phoenix-core/src/it/java/org/apache/phoenix/end2end/IndexExtendedIT.java
+++ b/phoenix-core/src/it/java/org/apache/phoenix/end2end/IndexExtendedIT.java
@@ -35,6 +35,7 @@ import java.util.Map;
 import java.util.Properties;
 
 import org.apache.hadoop.hbase.client.Table;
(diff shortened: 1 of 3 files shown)


gjacoby126:
I can see two ways to handle the case where max lookback is disabled / not supported :
1. As @kadirozde has done here, treat everything as beyond max lookback. This will let the jobs pass, but all missing / invalid rows will go into the beyond max lookback counters rather than the normal ones. 
2. Treat nothing as beyond max lookback, because the concept is disabled / not supported, but change the reducer to pass the job in this case. This will put all missing / invalid rows in the normal counters. 

Either of these could cause user confusion. In the first case, users might wonder why everything's in the max lookback counters, and not know to look at the data more closely in case there's a real failure. In the second case, users might wonder why the job passed even though there's failed counters, and not know that some of them might be compaction artifacts from the lack of max lookback. (This might be alleviated with logging, assuming users read the logs.) 

Curious on what others' thoughts are, or if there's another approach I'm missing.

gjacoby126 (approved):
+1

PR #904, opened 2020-09-30
PHOENIX-6167 Adding maxMutationCellSizeBytes config and exception

yanxinyi:


phoenix-core/src/main/java/org/apache/phoenix/compile/UpsertCompiler.java
```diff
@@ -139,6 +140,12 @@ private static void setValues(byte[][] values, int[] pkSlotIndex, int[] columnIn
         for (int i = 0, j = numSplColumns; j < values.length; j++, i++) {
             byte[] value = values[j];
             PColumn column = table.getColumns().get(columnIndexes[i]);
+            if (value.length >= maxCellSizeBytes) {
```
ChinmaySKulkarni, line 143: What about use-cases that use SINGLE_CELL_ARRAY_WITH_OFFSETS storage scheme wherein all the column keyvalues are stored inside a single cell? In that case, the hbase cell limit would apply to the sum of all the columns (or something along those lines). Either we should expand this to apply to all storage schemes, or restrict scope to only apply to ONE_CELL_PER_COLUMN

phoenix-core/src/main/java/org/apache/phoenix/compile/UpsertCompiler.java
```diff
@@ -187,6 +213,9 @@ public static MutationState upsertSelect(StatementContext childContext, TableRef
         int maxSizeBytes =
                 services.getProps().getInt(QueryServices.MAX_MUTATION_SIZE_BYTES_ATTRIB,
                     QueryServicesOptions.DEFAULT_MAX_MUTATION_SIZE_BYTES);
+        int maxCellSizeBytes =
```
ChinmaySKulkarni, line 216: I don't understand why we need another config for this. Introducing a Phoenix level config to mean the same thing as an existing hbase-level config will lead to confusion and now we have to make sure that this config is updated in case the main hbase config is updated. Why not just use 
`hbase.client.keyvalue.maxsize` or `hbase.server.keyvalue.maxsize` ?

ChinmaySKulkarni (changes requested):
Let's avoid a new config and just use hbase.client/server.keyvalue.maxsize

ChinmaySKulkarni (approved):
+1 thanks @yanxinyi

PR #934, opened 2020-10-21
PHOENIX-6202: New column in index gets added as PK with CASCADE INDEX

swaroopak:


gjacoby126 (approved):
+1

gokceni (approved):
+1 I tested your change and observed that it is indeed a covered column for indexes. 
Thanks for the quick fix!

PR #1018, opened 2020-12-15
Phoenix 6243

fengchen8086:
Fix bug where ValueBitSet can be "true" for incorrect columns.


fengchen8086:
incorrect branch to master

PR #1064, opened 2021-01-06
PHOENIX-6050 Set properties is invalid in client

wangchao316:


wangchao316:
@virajjasani @gjacoby126 could you please review this ? 
Thanks.

phoenix-core/src/main/java/org/apache/phoenix/jdbc/PhoenixDriver.java
```diff
@@ -148,21 +160,23 @@ public PhoenixDriver() { // for Squirrel
...
         int maxCacheDuration = config.getInt(QueryServices.CLIENT_CONNECTION_CACHE_MAX_DURATION_MILLISECONDS,
             QueryServicesOptions.DEFAULT_CLIENT_CONNECTION_CACHE_MAX_DURATION);
         RemovalListener<ConnectionInfo, ConnectionQueryServices> cacheRemovalListener =
             new RemovalListener<ConnectionInfo, ConnectionQueryServices>() {
                 @Override
                 public void onRemoval(RemovalNotification<ConnectionInfo, ConnectionQueryServices> notification) {
                     String connInfoIdentifier = notification.getKey().toString();
-                    LOGGER.debug("Expiring " + connInfoIdentifier + " because of "
-                        + notification.getCause().name());
+                    if (LOGGER.isDebugEnabled()) {
```
yanxinyi, line 172: Do we need if check here?

phoenix-core/src/main/java/org/apache/phoenix/jdbc/PhoenixDriver.java
```diff
@@ -182,7 +196,7 @@ public void onRemoval(RemovalNotification<ConnectionInfo, ConnectionQueryService
     
 
     @Override
-    public QueryServices getQueryServices() throws SQLException {
+    public QueryServices getQueryServices(final Properties info) throws SQLException {
```
yanxinyi, line 199: Adding extra argument won’t override the method

phoenix-core/src/main/java/org/apache/phoenix/jdbc/PhoenixDriver.java
wangchao316: thanks review, pheonix checkstyle do have, but logger use by this in industry.

phoenix-core/src/test/java/org/apache/phoenix/query/BaseTest.java
```diff
@@ -184,22 +184,22 @@
...
     public static final String DRIVER_CLASS_NAME_ATTRIB = "phoenix.driver.class.name";
     private static final double ZERO = 1e-9;
-    private static final Map<String,String> tableDDLMap;
+    private static final Map<String, String> tableDDLMap;
     private static final Logger LOGGER = LoggerFactory.getLogger(BaseTest.class);
     @ClassRule
     public static TemporaryFolder tmpFolder = new TemporaryFolder();
-    private static final int dropTableTimeout = 300; // 5 mins should be long enough.
-    private static final ThreadFactory factory = new ThreadFactoryBuilder().setDaemon(true)
+    private static final int DROP_TABLE_TIMEOUT = 300; // 5 mins should be long enough.
```
yanxinyi, line 191: Nice coding style changing!

phoenix-core/src/main/java/org/apache/phoenix/jdbc/PhoenixEmbeddedDriver.java
```diff
@@ -91,8 +89,15 @@
...
-    
-    abstract public QueryServices getQueryServices() throws SQLException;
+
+    /**
+     *  get query services
+     * @param info properties info
+     * @return query services
+     * @throws SQLException if failed to get query service.
+     */
+    public abstract QueryServices getQueryServices(Properties info)
```
yanxinyi, line 99: I see. You made an interface changes. I would suggest keep the original method with depreciation tag and remove it after two release.

phoenix-core/src/main/java/org/apache/phoenix/jdbc/PhoenixDriver.java
wangchao316: > Adding extra argument won’t override the method

I find this getQueryServices only use in inner the project， only test and PhoenixDriver use.   client use by PhoenixDriver, do not oveeride this interface.

phoenix-core/src/main/java/org/apache/phoenix/jdbc/PhoenixEmbeddedDriver.java
wangchao316: > I see. You made an interface changes. I would suggest keep the original method with depreciation tag and remove it after two release.
thanks , this is a good idea, I will do..

stoty:
I am generally not convinced that overriding poolsize from the client side is a good idea.
It is very easy to overload the cluster by specifying a too large poolsize.
IMO it's better to leave this knob in the cluster administrator's hand.

stoty:
Also see my objection from the ticket that this will cause discrepancy between the poolsize used by the client, and by the Index rebuilding initiated by MetaDataObserver.

wangchao316:
> Also see my objection from the ticket that this will cause discrepancy between the poolsize used by the client, and by the Index rebuilding initiated by MetaDataObserver.

Thanks @stoty  ,  I deal this ，have a big reason.
when have not use phoenix-server,  compiler and hbase scan operate in client-side,   and scan by poolsize in parallelIterator. I test in our prodect environment, this performance improvement more.

wangchao316:
and I find  from 4.x to 5.x,   reduce performance in some changed, which consumer use in our product environment  by upgrading.  I will always deal this question later.

stoty:
I agree @wangchao316 that tuning poolsize is crucial for performance optimization.
It is just my opinion that the tuning should be done on the cluster size, in hbase-site.xml, rather than on the client side.

wangchao316:
> I agree @wangchao316 that tuning poolsize is crucial for performance optimization.
> It is just my opinion that the tuning should be done on the cluster size, in hbase-site.xml, rather than on the client side.

yes, I feel we provide this parameter, should be altered in effect performance place. 
phoenix deploy in phoenix-server mode and no phoenix-server mode.
1. if be no phoenix-server mode, client will be  compiler, optimization，scan to hbase-regionserver side. so client will use this parameter in ParallelIterator.java， which effect scan.
2. if be phoenix-server mode, we will alter parameter in phoenix-server     .

PR #1086, opened 2021-01-13
PHOENIX-6307 Build and release official binary distributions with eac…

stoty:
…h HBase profile

also change default HBase profile to 1.6


yanxinyi:
Just tested it with your patch locally with the hbase1.3 profile, looks good.

yanxinyi:
also verified without hbase profile param, it built the hbase 1.6 as the default profile.

Thanks for the patch @stoty, and I don't have anything to add. Do you think we still need @gjacoby126 or @joshelser to review the patch?

stoty:
Thank you @yanxinyi.
The big issue here is not the maven mechanical changes, but new artifact/file naming scheme, and you and @joshelser have already approved that on the list.
Having approval from both sides is enough for me.

PR #1112, opened 2021-01-26
PHOENIX-6340 Infinitely looping ResultSet.next()

stoty:


dbwong:
This should be integration testable; however, I saw in the JIRA that you were having issues recreating this.  What area according to was giving you issues making an IT for this?

stoty:
There are two issues here.

The one that I am fixing is that RoundRobinResultIterator doesn't close itself when every iterator 's last element is the last cached one, and will re-initialize its iterator list on the next next() call.

The other, deeper problem is that even if those iterators are re-initialized, they should still be at the their last element, and not start returning their contents again.

Most of the the time this doesn't happen, I couldn't repro this in an IT, and even sqlline calls almost the same methods in both modes, yet the bug triggers only with BufferedRows.

So the second issue seems to be a timing/race  issue.

The bug is old, but it was masked by ResultSet close() -ing itself at last row. This was removed as this caused problem with many apps. This patch does the same thing that ResultSet used to do, only one level lower.

phoenix-core/src/main/java/org/apache/phoenix/iterate/RoundRobinResultIterator.java
```diff
@@ -300,7 +305,11 @@ public Tuple call() throws Exception {
         private RoundRobinIterator(PeekingResultIterator itr, Tuple tuple) {
             this.delegate = itr;
             this.tuple = tuple;
-            this.numRecordsRead = 0;
+            if (tuple != null) {
```
stoty, line 308: This is meant to be a performance fix.
The tuple was peek() -ed from the delegate, so that element is actually the first of the cached rows.
Fixing this SHOULD enable processing the 2nd and later batches parallelly, but I didn't thest this.

stoty:
Closing this.
I am still not convinced that the threshold logic is correct, but that's only a performance issue.
I've opened PHOENIX-6342 for a focused fix on the close() issue.

PR #1162, opened 2021-03-10
PHOENIX-6399 Updating BackwardCompatibilityIT supported versions

yanxinyi:


stoty:
I have just uploaded an alternative solution for this in #1163 , which lets you specify the artifact coordinates directly in the config file.

Either one gets the job done, but mine is perhaps a bit easier to work with in the future.

PR #1321, opened 2021-09-22
PHOENIX-5865 column that has default value can not be correctly indexed

richardantal:


phoenix-core/src/main/java/org/apache/phoenix/schema/MetaDataClient.java
```diff
@@ -1641,9 +1641,6 @@ public MutationState createIndex(CreateIndexStatement statement, byte[][] splits
                     String columnFamilyName = column.getFamilyName()!=null ? column.getFamilyName().getString() : null;
                     colName = ColumnName.caseSensitiveColumnName(IndexUtil.getIndexColumnName(columnFamilyName, column.getName().getString()));
                     isRowTimestamp = column.isRowTimestamp();
-                    if (colRef.getColumn().getExpressionStr() != null) {
-                        expressionStr = colRef.getColumn().getExpressionStr();
```
richardantal, line 1645: getExpressionStr() != null means we have a default value set for the column.
We use this expressionStr to generate the indexExpressions later. https://github.com/apache/phoenix/blob/8c68fe435d0cc7d8bc69ccb05bc6536ba5647113/phoenix-core/src/main/java/org/apache/phoenix/index/IndexMaintainer.java#L504-L506
In case of a column that has default value, lets say 0 a LiteralExpression was created and it always returned 0. (instead of DefaultValueExpression)
Because of this, in the index we wrote the default value always.

phoenix-core/src/main/java/org/apache/phoenix/expression/function/DefaultValueExpression.java
```diff
@@ -45,7 +45,7 @@ public DefaultValueExpression(List<Expression> children) throws SQLException {
     @Override
     public boolean evaluate(Tuple tuple, ImmutableBytesWritable ptr) {
         boolean evaluated = children.get(0).evaluate(tuple, ptr);
-        if (evaluated) {
+        if (evaluated && ptr.getLength() > 0) {
```
richardantal, line 48: I think this DefaultValueExpression was not used before in case of indexes.
This evaluate function is called here: https://github.com/apache/phoenix/blob/8c68fe435d0cc7d8bc69ccb05bc6536ba5647113/phoenix-core/src/main/java/org/apache/phoenix/index/IndexMaintainer.java#L714
Here we try to evaluate the first child, something like a KeyValueColumnExpression and if it was not successful we fall back to the defaultExpression.
When we ran this on a column that had no value in it (it should be interpreted as the default value) this evaluate returned true.
Examples of reading the score column:
Row1: (id, no, total, score) values ('1111','1112', 1113, 1114)
```
ptr.get(): [0, 0, 0, 19, 0, 0, 0, 4, 0, 4, 49, 49, 49, 49, 1, 48, -128, 13, 0, 0, 1, 124, 18, 99, 16, -119, 4, -128, 0, 4, 90]
ptr.getoffset() 27
ptr.getLength(): 4
```
It meanse the -128, 0, 4, 90 part will be in the built index Row.
Row2 (id, no, total) values ('1121', '1122', 1123) - score should be default
```
ptr.get():  [0, 0, 0, 19, 0, 0, 0, 0, 0, 4, 49, 49, 50, 49, 1, 48, -128, 13, 0, 0, 1, 124, 18, 99, 16, -119, 4]
ptr.getOffset() 27
ptr.getLength() 0
```
In this case we ended up writing this, with 0 Length
https://github.com/apache/phoenix/blob/8c68fe435d0cc7d8bc69ccb05bc6536ba5647113/phoenix-core/src/main/java/org/apache/phoenix/index/IndexMaintainer.java#L728
With my change we will fall back to the default value.

stoty (dismissed):
The DefaultColumnValueIT test failures are relevant.
It looks like the your changes break the semantics, which seems to be that the default is applied on write, not on read.

ss77892:
LGTM +1 Do we have test passed for this PR?

richardantal:
I rebased the commit to trigger the test run.

richardantal:
Should I edit the lines which are not relevant to this change but checkstyle mentions as new errors?
The 2  failing tests are known as flaky.

stoty:
If you haven't touched those lines, then fixing those checkstyle errors is optional.
If you fix it, then the code will be in a better shape, but the patch will be noisier.
If you don't fix it, then the patch will be easier to read, but the code will stay ugly.

stoty (commented):
Removing my Request changes review

PR #1328, opened 2021-10-06
PHOENIX-6548: Throw IOException instead of IllegalArgumentException when RS crashes during index rebuilds.

ankitjain64:
JIRA: https://issues.apache.org/jira/browse/PHOENIX-6548

In our pipelines, we saw that after the ClusterConnection is received from [ConnectionFactory](https://github.com/apache/phoenix/blob/4.x/phoenix-core/src/main/java/org/apache/phoenix/util/ServerUtil.java#L362) the region server crashes and the underlying connection is closed which causes IllegalArgumentException from hbase in [getTable](https://github.com/apache/phoenix/blob/4.x/phoenix-core/src/main/java/org/apache/phoenix/util/ServerUtil.java#L317) and DoNotRetryIOException is sent back to client. 

Instead, we want the client exception to be a normal retriable IOException, because if they try again after the region comes up again somewhere else, the Scan will likely succeed.

### DIFF
--- a/phoenix-core/src/main/java/org/apache/phoenix/util/ServerUtil.java
+++ b/phoenix-core/src/main/java/org/apache/phoenix/util/ServerUtil.java
@@ -308,13 +308,13 @@ public class ServerUtil {
             this.connectionType = connectionType;
         }
 
-        private ClusterConnection getConnection() throws IOException {
+        public ClusterConnection getConnection() throws IOException {
             return ConnectionFactory.getConnection(connectionType, conf, server);
         }
 
         @Override
         public HTableInterface getTable(ImmutableBytesPtr tablename) throws IOException {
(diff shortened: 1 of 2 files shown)


ankitjain64:
@gokceni @gjacoby126 @tkhurana. Please have a look when you get a chance. Thanks.

ankitjain64:
> @ankitjain64 - looks like there was a test failure of PhoenixTableLevelMetricsIT when running against your patch. I just ran that test without your patch locally and it passed. Could be a flapper, but could you please check if you can reproduce the failure?

It passes locally for me too.

ankitjain64:
> Thank you @ankitjain64 for fixing this. Can you please also create PR against master branch?

Created here: https://github.com/apache/phoenix/pull/1329

PR #1340, opened 2021-10-27
PHOENIX-6579 ACL check doesn't honor the namespace mapping for mapped…

ss77892:
… views.

### DIFF
--- a/phoenix-core/src/it/java/org/apache/phoenix/end2end/PermissionNSEnabledIT.java
+++ b/phoenix-core/src/it/java/org/apache/phoenix/end2end/PermissionNSEnabledIT.java
@@ -18,8 +18,14 @@
 package org.apache.phoenix.end2end;
 
 import org.apache.hadoop.hbase.TableName;
+import org.apache.hadoop.hbase.client.Admin;
+import org.apache.hadoop.hbase.client.ColumnFamilyDescriptor;
+import org.apache.hadoop.hbase.client.ColumnFamilyDescriptorBuilder;
+import org.apache.hadoop.hbase.client.TableDescriptor;
+import org.apache.hadoop.hbase.client.TableDescriptorBuilder;
 import org.apache.hadoop.hbase.security.AccessDeniedException;
 import org.apache.hadoop.hbase.security.access.Permission;
+import org.apache.hadoop.hbase.util.Bytes;
 import org.apache.phoenix.exception.SQLExceptionCode;
 import org.apache.phoenix.util.SchemaUtil;
 import org.junit.BeforeClass;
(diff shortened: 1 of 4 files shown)


stoty:
This got me thinking that we should in fact 
use `"^"+tableName.getNameWithNamespaceInclAsString()+"$"`

to avoid a similar stituation, where we have permission for `ns:prefixTable,` and we do not add the required new permission for `ns:prefixTableFoo` because we THINK we already have it.

I think I'm going to open am HBase ticket about the horrific API design (if one doesn't exist yet).

phoenix-core/src/it/java/org/apache/phoenix/end2end/PermissionNSEnabledIT.java
```diff
@@ -48,6 +56,43 @@ public PermissionNSEnabledIT() throws Exception {
     public static synchronized void doSetup() throws Exception {
         BasePermissionsIT.initCluster(true);
     }
+    private AccessTestAction createMappedView(final String schemaName, final String tableName) throws SQLException {
```
stoty, line 59: We should run this test for the NS Disabled case as well. 
I.e. I think this should be in the parent class.

ss77892: Well, didn't realize it from the beginning, but with NS disabled this test case doesn't make sense. We have a single namespace 'default' and all system and user tables are in there.

stoty: Are you sure ?
While we are not mapping the Phoenix schema to a the HBase namespaces, we prepend it to the Hbase table name directly, I think that the regex confustion can still happens the same.

i.e. the `tablename` regex will still match `anyschema.tablename,` and even `anyschema.tablenamefoo` , only maybe on a different codepath in HBase.

ss77892: Actually, the problem with the regexp is a bit different. It's the dot symbol. When we check the user permission for the table 'TEST.TEST' we accidentally get the permission for 'TEST:TEST' and would not check whether the table names match or not. So in the fix, we would remove the namespace problem by making the regex string have the namespace (so it would be 'default:TEST.TEST'), but I just realized that in theory if we don't have namespace mapping enabled and there are tables 'TEST.TEST' and 'TESTATEST' with a user that has access to TESTATEST than he/she would have access the 'TEST.TEST'. So to avoid that we also must escape the dot in the table name. Would write a test case and if it fails will file a separate bug.

stoty:
Closing as this has been merged.

PR #1343, opened 2021-11-04
PHOENIX-6586 Set NORMALIZATION_ENABLED to false on salted tables

stoty:


phoenix-core/src/main/java/org/apache/phoenix/query/ConnectionQueryServicesImpl.java
```diff
@@ -1438,6 +1438,11 @@ private TableDescriptor ensureTableCreated(byte[] physicalTableName, byte[] pare
                         PBoolean.INSTANCE.toObject(newDesc.build().getValue(MetaDataUtil.IS_LOCAL_INDEX_TABLE_PROP_BYTES)))) {
                     newDesc.setRegionSplitPolicyClassName(IndexRegionSplitPolicy.class.getName());
                 }
+                if (props.get(PhoenixDatabaseMetaData.SALT_BUCKETS) != null
+                        && (Integer) (props.get(PhoenixDatabaseMetaData.SALT_BUCKETS)) > 0) {
+                    //Silently overwrites property if it was specified explicitly
```
gjacoby126, line 1443: probably want at least a LOG.WARN() here if we're overwriting a property the client explicitly asked for, or alternatively throw an exception

stoty: I have added an exception.
Please take a look again @gjacoby126.

ankitsinghal (approved):

stoty:
Sorry for the additional review requests, I missed that @ankitsinghal  has already approved. (Thanks!)

PR #1345, opened 2021-11-08
PHOENIX-6227 - Option for DDL changes to export to external schema re…

gjacoby126:
…pository


gjacoby126:
Looks like test failures around the ViewUtil changes because the existing logic seems slightly different than on the 4.x side. Will investigate and fix. I've also fixed the new javadoc warnings

gjacoby126:
As this is a port of #1341 which has been approved, and tests are now passing, I'll go ahead and merge.

PR #1347, opened 2021-11-11
PHOENIX-6589 - Add metrics for schema registry export

gjacoby126:


phoenix-pherf/src/main/java/org/apache/phoenix/pherf/util/PhoenixUtil.java
```diff
@@ -51,6 +51,7 @@
 import java.text.SimpleDateFormat;
 import java.util.ArrayList;
 import java.util.Collections;
+import java.util.HashMap;
```
tkhurana, line 54: nit: redundant import

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/MetaDataEndpointImpl.java
```diff
@@ -571,6 +573,8 @@ public static PName newPName(byte[] keyBuffer, int keyOffset, int keyLength) {
     // before 4.15, so that we can rollback the upgrade to 4.15 if required
     private boolean allowSplittableSystemCatalogRollback;
 
+    private MetricsMetadataSource metricsSource;
```
tkhurana, line 576: Should this also be declared volatile ? I noticed you declared another similar variable volatile below.

phoenix-pherf/src/main/java/org/apache/phoenix/pherf/util/PhoenixUtil.java
gjacoby126: Where's the redundancy? This was fixing a build failure from a missing import from a previous JIRA someone else pushed. (I've since fixed it as an addendum so this will go away when I rebase, I think.)

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/MetaDataEndpointImpl.java
gjacoby126: Is one necessary? In the other cases I was populating the variable in a synchronized block with a null check, which I believe needs a volatile, but in this case I'm just setting it from start() and not treating it as an explicit singleton. (Do I need to?)

tkhurana: I see. It is not needed then.

tkhurana (approved):

gjacoby126:
The failing test suite (TenantSpecificTablesDDLIT) passed locally for me for this branch, and passed in the 5.x version of the PR as well. I'll go ahead with the merge. Thanks for the reviews, @swaroopak and @tkhurana !

PR #1348, opened 2021-11-12
PHOENIX-6589 - Add metrics for schema registry export

gjacoby126:


gjacoby126:
As this is an identical port of #1347 which has been approved, I'll merge this too.

PR #1351, opened 2021-11-15
PHOENIX-6592 PhoenixStatsCacheLoader uses non-deamon threads

stoty:


joshelser (approved):
Thanks for fixing, Istvan.

For everyone else, we got this bug report from a user whose Spark application would not terminate normally when talking to Phoenix. After obtaining a heapdump of the "stuck" application, we could see that the threads launched by this threadpool were the reason the JVM was not terminating.

phoenix-core/src/main/java/org/apache/phoenix/query/PhoenixStatsCacheLoader.java
```diff
@@ -43,15 +46,20 @@
...
                     // The size of the thread pool used for refreshing cached table stats
                     final int statsCacheThreadPoolSize = config.getInt(
                             QueryServices.STATS_CACHE_THREAD_POOL_SIZE,
                             QueryServicesOptions.DEFAULT_STATS_CACHE_THREAD_POOL_SIZE);
-
-                    this.executor = Executors.newFixedThreadPool(statsCacheThreadPoolSize);
+                    final ThreadFactory threadFactory =
+                            new ThreadFactoryBuilder()
+                                    .setDaemon(true)
+                                    .setNameFormat("PHOENIX-STATS-CACHE-LOADER" + "-thread-%s")
```
virajjasani, line 59: nit: why not `"PHOENIX-STATS-CACHE-LOADER-thread-%s"` directly?

phoenix-core/src/main/java/org/apache/phoenix/query/PhoenixStatsCacheLoader.java
```diff
@@ -43,15 +46,20 @@
...
+                if (executor == null) {
                     // The size of the thread pool used for refreshing cached table stats
                     final int statsCacheThreadPoolSize = config.getInt(
                             QueryServices.STATS_CACHE_THREAD_POOL_SIZE,
                             QueryServicesOptions.DEFAULT_STATS_CACHE_THREAD_POOL_SIZE);
-
-                    this.executor = Executors.newFixedThreadPool(statsCacheThreadPoolSize);
+                    final ThreadFactory threadFactory =
+                            new ThreadFactoryBuilder()
+                                    .setDaemon(true)
```
virajjasani, line 58: This is the main fix right? Looks good.

virajjasani (approved):
+1

phoenix-core/src/main/java/org/apache/phoenix/query/PhoenixStatsCacheLoader.java
stoty: Because that's how it was in the method that I copy-pasted this from :)
I'm going to fix it before commit.

PR #1352, opened 2021-11-17
PHOENIX-6594 Clean up vararg warnings flagged as errors by Eclipse

stoty:


phoenix-core/src/main/java/org/apache/phoenix/mapreduce/index/IndexToolUtil.java
```diff
@@ -74,7 +74,7 @@ public static void updateIndexState(Connection connection, final String masterTa
         Preconditions.checkNotNull(connection);
         final String alterQuery = String.format(ALTER_INDEX_QUERY_TEMPLATE,indexTable,masterTable,state.name());
         connection.createStatement().execute(alterQuery);
-        LOGGER.info(" Updated the status of the index {} on {} to {} ", indexTable , masterTable, state.name());
+        LOGGER.info(" Updated the status of the index {} on {} to {} ", indexTable ,masterTable, state.name());
```
richardantal, line 77: nit: whitespace should be after the comma

stoty: Thanks, I'll fix that on commit.

stoty: Actually, I've just pushed the fix here.

richardantal (approved):
LGTM

PR #1353, opened 2021-11-17
PHOENIX-6587 Disallow specifying explicit pre-split on salted tables

stoty:


dbwong:
Do we already prevent create table on preexisting table with salt set?

stoty:
> Do we already prevent create table on preexisting table with salt set?

I wouldn't want to prevent that.
Copying/Loading a Phoenix data table with HBase tools into a cluster, and then running _create table_ to re-create the Phoenix metadata is a supported and standard procedure, (even if I have seen people messing that up.)

I haven't tested it, but as we're not creating a new HBase table, we probably either error out, or just ignore the pre-split points in this case. I'm going to try out what happens in this case.

dbwong:
> > Do we already prevent create table on preexisting table with salt set?
> 
> I wouldn't want to prevent that. Copying/Loading a Phoenix data table with HBase tools into a cluster, and then running _create table_ to re-create the Phoenix metadata is a supported and standard procedure, (even if I have seen people messing that up.)
> 
> I haven't tested it, but as we're not creating a new HBase table, we probably either error out, or just ignore the pre-split points in this case. I'm going to try out what happens in this case.

I know we had a case in pre-production where we got region boundaries that overlapped the salt boundaries via restore or manual create table ontop of existing table.  This led to wrong results due to phoenix's explicit assumption on the salted boundaries.

dbwong (approved):

stoty:
thank you @dbwong 

> Maybe a better approach for this worry is to check during create table that the table is split sanely if its salted etc.

I have started working on this, but got sidetracked by a deadline on another project.
I'm about 50% done with the split verification solution, and hope to get that ready and tested by early next year.

stoty:
Note that new patch removes the SchemaUtil.processSplit() logic, which attempts to expend the split point to cover the whole PK.

My reasons removing it:

* It doesn't really work and uses arbitrary lengths when working with PKs with variable length fields.
* It was only ever enforced on table creation with pre-split tables, and not durint any other splits, and we haven't encountered any bugs from those later splits.
* I could not figure why it was there in the first place.
* It made adding the validation logic awkward

stoty:
Some more reasoning to remove SchemaUtil.processSplit() :

AFAICT it tries to emulate how HBase would automatically split a region.
However, I could not find any reason why we would want to do that.
* It is not necessary for correctness
* It always adds an extra region to scan, as a salt key will always extend to an extra region (but this extra region is  guaranteed to not have any rows)

stoty:
There two outstanding problems:
* Upgrade tests fail, because the split points on the System tables use the old longer split points, and the update code errors out.
* org.apache.phoenix.end2end.LocalIndexSplitMergeIT.testLocalIndexScanWithMergeSpecialCase() fails, but I cannot figure out why. Maybe the local index code dependes on the longer split points feature that I have removed ?

stoty:
Both could probably be fixed by re-adding the split point extesion code, but I'd like to understand it more first.

stoty:
org.apache.phoenix.end2end.LocalIndexSplitMergeIT.testLocalIndexScanWithMergeSpecialCase() doesn't even use salted tables.
I must have broken something else in the patch.

stoty:
Turns out that split processing is not necessary for salted tables, but rather for all tables that have local indexes, which LocalIndexSplitMergeIT.testLocalIndexScanWithMergeSpecialCase() has luckily caught. 

This of course means that we must keep the split processing.

This also means that we are already silently changing the manually specified split points.
I think that we should just fix the split points specified for salted tables, as there is precedence for similar behaviour, and I think that this is the closet thing to doing what the user wanted.

stoty:
Could you please review the new version @dbwong ?

phoenix-core/src/main/java/org/apache/phoenix/schema/SaltingUtil.java
```diff
@@ -123,4 +130,29 @@ public static void addRegionStartKeyToScanStartAndStopRows(byte[] startKey, byte
...
+        byte[] maxStartKey = new byte[] {0} ;
+        List<RegionInfo> regionInfos =
+                admin.getRegions(tableName);
+        for (RegionInfo regionInfo : regionInfos) {
+            if (Bytes.compareTo(regionInfo.getStartKey(), maxStartKey) > 0) {
+                maxStartKey = regionInfo.getStartKey();
+            }
+        }
+        byte lastRegionStartSaltKey = maxStartKey[0];
+        if (lastRegionStartSaltKey == (byte) (buckets - 1)) {
```
joshelser, line 147: I like the vision of what you are doing here. It's 100% OK to me to try to catch the obvious case(s) that we can and not try to catch all possibilities for messed up region boundaries + salt.

Is it possible to add an "escape hatch" to this method for an experienced user to give a "I know what I'm doing" override? I can't think of a case where your logic is insufficient, but could envision a scenario where something unexpected happens ;)

phoenix-core/src/main/java/org/apache/phoenix/util/SchemaUtil.java
```diff
@@ -595,25 +594,35 @@ public static boolean isSystemTable(byte[] fullTableName) {
...
-    
-    // Given the splits and the rowKeySchema, find out the keys that 
-    public static byte[][] processSplits(byte[][] splits, LinkedHashSet<PColumn> pkColumns, Integer saltBucketNum, boolean defaultRowKeyOrder) throws SQLException {
-        // FIXME: shouldn't this return if splits.length == 0?
-        if (splits == null) return null;
-        // We do not accept user specified splits if the table is salted and we specify defaultRowKeyOrder. In this case,
-        // throw an exception.
-        if (splits.length > 0 && saltBucketNum != null && defaultRowKeyOrder) {
-            throw new SQLExceptionInfo.Builder(SQLExceptionCode.NO_SPLITS_ON_SALTED_TABLE).build().buildException();
-        }
```
joshelser, line 607: I'm trying to double-check that it's intentional that you dropped throwing this exception

> found that merging the regions will break Phoenix, irrespective of the value of phoenix.query.force.rowkeyorder

Did you mean merging salt split points with user-provided split points (like I think this method was doing) or did you just mean general region merging?

joshelser (commented):
A couple of questions. I think this seems fine to me, but it's stretching my level of comfort to drop an approval on it :)

phoenix-core/src/main/java/org/apache/phoenix/schema/SaltingUtil.java
stoty: Technically it is possible, but it would be a lot of work as we'd need to start from the grammar to add a force option.

I've handled a few such cases, and in every case it was a user error.
I have given this some thought, and I cannot imagine such a scenario.

If the user has merged his last regions on his table that otherwise matches the salt bucket count, and wants to create the table, it still wouldn't work correctly, and they are better off fixing their splits first, and re-creating the Phoenix table second.

stoty:
However, I need to update the ticket and patch description, as the patch does kind of the opposite of what it says.

stoty:
Squashed and renamed the patch.

phoenix-core/src/main/java/org/apache/phoenix/schema/SaltingUtil.java
joshelser: > Technically it is possible, but it would be a lot of work as we'd need to start from the grammar to add a force option.

I see that sqlline has some functionality to pass down generic properties, but we don't have that hooked up through sqlline.py

Can defer this until it bites us, I guess.

joshelser (approved):
I'll give approval but it would be great for someone else to also take a look. I trust that you've thought this through quite a bit already, Istvan.

My final thoughts are about a Phoenix user who has an HBase table (or snapshot) which has Phoenix data in it but the Phoenix table doesn't yet exist. The path forward for them to re-create the table in Phoenix is "ensure that split points exist for each salt bucket", but we don't have a tool to dynamically do that, do we?

phoenix-core/src/it/java/org/apache/phoenix/end2end/salted/SaltedTableIT.java
```diff
@@ -63,14 +65,43 @@ public void testTableWithInvalidBucketNumber() throws Exception {
...
+        createTestTable(getUrl(), "create table " + tableName + " (a_integer integer not null primary key) SALT_BUCKETS = 4",
+                new byte[][] {{1}, {2,3}, {2,5}, {3}}, null);
+        Admin admin = utility.getAdmin();
+        ArrayList<byte[]> startKeys = new ArrayList<>();
+        List<RegionInfo> regionInfos =
+                admin.getRegions(TableName.valueOf(tableName));
+        for (RegionInfo regionInfo : regionInfos) {
+            startKeys.add(regionInfo.getStartKey());
+        }
+        assertTrue(Arrays.deepEquals(expectedStartKeys, startKeys.toArray()));
```
ankitsinghal, line 86: This makes sense, we are ensuring that we never overlap two buckets.

phoenix-core/src/main/java/org/apache/phoenix/query/ConnectionlessQueryServicesImpl.java
```diff
@@ -289,8 +289,11 @@ public MetaDataMutationResult getTable(PName tenantId, byte[] schemaBytes, byte[
     }
     
     private static List<HRegionLocation> generateRegionLocations(byte[] physicalName, byte[][] splits) {
-        byte[] startKey = HConstants.EMPTY_START_ROW;
         List<HRegionLocation> regions = Lists.newArrayListWithExpectedSize(splits.length);
+        if (splits.length == 0) {
+            return regions;
+        }
+        byte[] startKey = HConstants.EMPTY_START_ROW;
```
ankitsinghal, line 296: nit:
```suggestion
        if (splits.length == 0) {
            return regions;
        }
        List<HRegionLocation> regions = Lists.newArrayListWithExpectedSize(splits.length);
        byte[] startKey = HConstants.EMPTY_START_ROW;
```

phoenix-core/src/main/java/org/apache/phoenix/util/SchemaUtil.java
```diff
@@ -595,25 +594,35 @@ public static boolean isSystemTable(byte[] fullTableName) {
...
+            byte[][] saltSplits = SaltingUtil.getSaltedByteSplitPoints(saltBucketNum);
+            for (int i=0; i<saltSplits.length; i++) {
+                allSplits.add(processSplit(saltSplits[i], pkColumns));
+            }
+        }
+
+        //Add the specified split points
         for (int i=0; i<splits.length; i++) {
-            newSplits[i] = processSplit(splits[i], pkColumns); 
+            allSplits.add(processSplit(splits[i], pkColumns));
```
ankitsinghal, line 617: How we are ensuring that the explicit split will always start with the SALT < buckets?

phoenix-core/src/main/java/org/apache/phoenix/query/ConnectionlessQueryServicesImpl.java
stoty: This won't compile, we need an empty List to return.

phoenix-core/src/main/java/org/apache/phoenix/util/SchemaUtil.java
stoty: Added check

stoty:
PHOENIX-6910 makes this moot.

PR #1357, opened 2021-11-22
PHOENIX-6601 Fix IndexTools bugs with namespace mapping

stoty:


gjacoby126:
@stoty - code looks fine, but it looks like there's a build error in IndexBuildTimestampIT

ss77892:
The way how the problem was addressed LGTM. As for the tests. There are a number of tests that are using IndexToolIT directly (usually runIndexTool method). Only one of those (IndexBuildTimestampIT) is calling setup for some purposes and that caused the compilation failure, but we might expect that others can be affected in some way as well. Not related to this particular patch, but some cleanup with the index-related IT tests needs to be done sometimes.

stoty:
Sorry for the build error, I was too focused on IndexToolIT.java.
I've fixed the build now.

A lot of tests need cleanup, unfortunately.
We still have a lot of flakey tests as well.

gjacoby126 (approved):
+1 assuming the test run is clean. Thanks @stoty

stoty:
ASF Jenkins is down.
Gonna run an off-line mvn verify check.

stoty:
Offline test has completed successfully (well, actually it failed on a known flakey test, but that's expected)

PR #1364, opened 2021-12-09
PHOENIX-6596 Schema extraction double quotes expressions, resulting i…

richardantal:
…n un-executabe create statements

Co-authored-by: Istvan Toth <stoty@apache.org>

### DIFF
--- a/phoenix-core/src/main/java/org/apache/phoenix/schema/tool/SchemaExtractionProcessor.java
+++ b/phoenix-core/src/main/java/org/apache/phoenix/schema/tool/SchemaExtractionProcessor.java
@@ -135,8 +135,8 @@ public class SchemaExtractionProcessor implements SchemaProcessor {
         StringBuilder indexedColumnsBuilder = new StringBuilder();
 
         for (PColumn indexedColumn : indexPK) {
-            String indexColumn = extractIndexColumn(indexedColumn.getName().getString(), defaultCF);
-            if(indexColumn.equalsIgnoreCase(MetaDataUtil.VIEW_INDEX_ID_COLUMN_NAME)) {
+            String indexColumn = extractIndexColumn(indexedColumn.getExpressionStr(), defaultCF);
+            if (indexColumn == null) {
                 continue;
             }
             indexPKName.add(indexColumn);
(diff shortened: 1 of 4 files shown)


phoenix-core/src/main/java/org/apache/phoenix/util/SchemaUtil.java
```diff
@@ -1301,6 +1301,29 @@ public static String formatColumnName(String name) {
         return name;
     }
 
+    public static String formatIndexColumnName(String name) {
+        String[] splittedName = name.split("\\.");
```
stoty, line 1305: nit: split is an irregular verb, so this should be splitName

phoenix-core/src/main/java/org/apache/phoenix/schema/tool/SchemaExtractionProcessor.java
```diff
@@ -135,8 +135,8 @@ private String getIndexedColumnsString(PTable indexPTable, PTable dataPTable, St
         StringBuilder indexedColumnsBuilder = new StringBuilder();
 
         for (PColumn indexedColumn : indexPK) {
-            String indexColumn = extractIndexColumn(indexedColumn.getName().getString(), defaultCF);
-            if(indexColumn.equalsIgnoreCase(MetaDataUtil.VIEW_INDEX_ID_COLUMN_NAME)) {
+            String indexColumn = extractIndexColumn(indexedColumn.getExpressionStr(), defaultCF);
```
stoty, line 138: so indexedColumn.getExpressionStr() will return the colum name for "simple" indexes ?

phoenix-core/src/main/java/org/apache/phoenix/schema/tool/SchemaExtractionProcessor.java
```diff
@@ -135,8 +135,8 @@ private String getIndexedColumnsString(PTable indexPTable, PTable dataPTable, St
...
+            if(indexColumn == null || indexColumn.equalsIgnoreCase(MetaDataUtil.VIEW_INDEX_ID_COLUMN_NAME)) {
```
stoty, line 139: In what case do we get a null  here ?

phoenix-core/src/main/java/org/apache/phoenix/schema/tool/SchemaExtractionProcessor.java
richardantal: We still parse it but indexedColumn.getExpressionStr() is better to parse.
indexedColumn.getName().getString() returnes 'QUOTED' for quoted string and ('number' * 'currency') for lowercase columns so we would not be able to differentiate them.
On the other hand  indexedColumn.getExpressionStr() returns these columns as 'QUOTED' and ("number" * "currency")

phoenix-core/src/main/java/org/apache/phoenix/schema/tool/SchemaExtractionProcessor.java
richardantal: If indexedColumn.getExpressionStr() is null I return null in extractIndexColumn.

richardantal: And indexedColumn.getExpressionStr() is null when indexedColumn.getName().getString() was equal to MetaDataUtil.VIEW_INDEX_ID_COLUMN_NAME

stoty (approved):
+1 LGTM

phoenix-core/src/main/java/org/apache/phoenix/schema/tool/SchemaExtractionProcessor.java
```diff
@@ -184,21 +184,39 @@ private String getIndexedColumnsString(PTable indexPTable, PTable dataPTable, St
...
                 return String.format("%s.%s", schema, name);
             } else {
-                return SchemaUtil.formatColumnName(columnNameSplit[0]);
+                return formatColumnOrExpression(columnNameSplit[0]);
             }
         }
     }
 
+    private String formatColumnOrExpression(String columnOrExpression) {
+        System.out.println("columnOrExpressioncolumnOrExpressioncolumnOrExpression");
```
gjacoby126, line 206: Please remove the System.out.printlns

gjacoby126 (changes requested):

gjacoby126 (approved):
+1, thanks @richardantal

PR #1365, opened 2021-12-14
PHOENIX-6611 Multiple IndexTool bugs

stoty:
Copying from the ticket:

There is a lot of noise, due to the removal of the directApi option, and movinf the VERIFIED/UNVERIFIED constants to QueryConstants.
The interesting stuff is in:

IndexTool
FormatToBytesWritableMapper
PhoenixIndexImportDirectMapper
PhoenixTxIndexMutationGenerator
IndexUtil
IndexMaintainer
IndexToolIT
IndexBuildTimestampIT

### DIFF
--- a/phoenix-core/src/it/java/org/apache/phoenix/end2end/IndexToolIT.java
+++ b/phoenix-core/src/it/java/org/apache/phoenix/end2end/IndexToolIT.java
@@ -17,11 +17,39 @@
  */
 package org.apache.phoenix.end2end;
 
-import org.apache.phoenix.compile.ExplainPlan;
-import org.apache.phoenix.compile.ExplainPlanAttributes;
-import org.apache.phoenix.jdbc.PhoenixPreparedStatement;
-import org.apache.phoenix.thirdparty.com.google.common.collect.Lists;
-import org.apache.phoenix.thirdparty.com.google.common.collect.Maps;
+import static org.apache.phoenix.mapreduce.PhoenixJobCounters.INPUT_RECORDS;
+import static org.apache.phoenix.mapreduce.index.PhoenixIndexToolJobCounters.BEFORE_REBUILD_BEYOND_MAXLOOKBACK_INVALID_INDEX_ROW_COUNT;
+import static org.apache.phoenix.mapreduce.index.PhoenixIndexToolJobCounters.BEFORE_REBUILD_BEYOND_MAXLOOKBACK_MISSING_INDEX_ROW_COUNT;
+import static org.apache.phoenix.mapreduce.index.PhoenixIndexToolJobCounters.BEFORE_REBUILD_EXPIRED_INDEX_ROW_COUNT;
+import static org.apache.phoenix.mapreduce.index.PhoenixIndexToolJobCounters.BEFORE_REBUILD_INVALID_INDEX_ROW_COUNT;
+import static org.apache.phoenix.mapreduce.index.PhoenixIndexToolJobCounters.BEFORE_REBUILD_INVALID_INDEX_ROW_COUNT_COZ_EXTRA_CELLS;
+import static org.apache.phoenix.mapreduce.index.PhoenixIndexToolJobCounters.BEFORE_REBUILD_INVALID_INDEX_ROW_COUNT_COZ_MISSING_CELLS;
+import static org.apache.phoenix.mapreduce.index.PhoenixIndexToolJobCounters.BEFORE_REBUILD_MISSING_INDEX_ROW_COUNT;
+import static org.apache.phoenix.mapreduce.index.PhoenixIndexToolJobCounters.BEFORE_REBUILD_VALID_INDEX_ROW_COUNT;
+import static org.apache.phoenix.mapreduce.index.PhoenixIndexToolJobCounters.REBUILT_INDEX_ROW_COUNT;
+import static org.apache.phoenix.mapreduce.index.PhoenixIndexToolJobCounters.SCANNED_DATA_ROW_COUNT;
+import static org.apache.phoenix.util.TestUtil.TEST_PROPERTIES;
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertTrue;
+import static org.junit.Assert.fail;
+
+import java.io.IOException;
+import java.sql.Connection;
+import java.sql.DriverManager;
+import java.sql.PreparedStatement;
+import java.sql.ResultSet;
+import java.sql.SQLException;
+import java.sql.Types;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Collection;
+import java.util.List;
+import java.util.Map;
+import java.util.Properties;
+import java.util.UUID;
+import java.util.concurrent.atomic.AtomicInteger;
+
 import org.apache.hadoop.conf.Configuration;
 import org.apache.hadoop.hbase.Cell;
 import org.apache.hadoop.hbase.CellUtil;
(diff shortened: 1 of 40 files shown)


stoty:
Sorry for asking for reviews for a half-done patch.

It is really ready for review now. 
(The last patch is just just formatting fixes, and does not change the behaviour.)

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/index/IndexTool.java
```diff
@@ -493,6 +493,10 @@ public Job getJob() throws Exception {
                 }
                 if (useSnapshot || (!isLocalIndexBuild && pDataTable.isTransactional())) {
                     PhoenixConfigurationUtil.setCurrentScnValue(configuration, maxTimeRange);
+                    if (indexVerifyType != IndexVerifyType.NONE) {
+                        LOGGER.warn("Verification is not supported for client index rebuilds,"
```
gjacoby126, line 497: The error message seems wrong -- verification is supported for some client-initiated index rebuilds, just not ones using a mapper other than the server build one. If you mean that the direct mapper creates the mutations and writes them as a client, that's probably not clear to the end user.

gjacoby126 (changes requested):
@stoty - Very elegant solution to fixing the direct mapper. Just a small nit.

phoenix-core/src/main/java/org/apache/phoenix/execute/PhoenixTxIndexMutationGenerator.java
```diff
@@ -364,7 +364,8 @@ private boolean generatePuts(
                 throw new IOException(e);
             }
         }
-        Iterable<IndexUpdate> puts = codec.getIndexUpserts(state, indexMetaData, regionStartKey, regionEndKey);
+        Iterable<IndexUpdate> puts =
+                codec.getIndexUpserts(state, indexMetaData, regionStartKey, regionEndKey, true);
```
gokceni, line 368: Shouldn't verified be false here? Are we generating index upsert mutations in verified=true even though the data table write fails? A code comment would be nice.

phoenix-core/src/it/java/org/apache/phoenix/end2end/IndexToolIT.java
```diff
@@ -378,6 +386,20 @@ public void testSecondaryIndex() throws Exception {
         }
     }
 
+    public static void checkVerifiedStatus(Connection conn, String indexFullName, byte[] rowKey) throws SQLException, IOException {
```
gokceni, line 389: ImmutableIndexIT has a similar method (verifyRowsForEmptyColValue). Could we reuse it here?

gokceni (commented):
Left some comments..

phoenix-core/src/main/java/org/apache/phoenix/execute/PhoenixTxIndexMutationGenerator.java
gjacoby126: @gokceni - I believe the purpose is to allow the old "direct" index tool mapper to be able to build new-style indexes, since snapshot-based and transactional index builds still need to use the direct mapper. Since it's reading existing base table data, we can assume the new index write is verified.

stoty: This is in the normal write path via PhoenixTransactionalIndexer, not the in the async index generation.

Three-phase commit is not used in the normal TX write path, because the transactionality is taking care of the
consistency, and accordingly it never updates the status to VERIFIED, and leaves the status as "x".

This method is also used by the old indexing implementation, and I didn't want old-style tables to set the status to verified, hence the new parameter.

The same basic problem pattern that BulkLoad and Async Indextool has, but this one is in the normal write path.

I should have called this change out explicitly, as admittedly this has the bigggest impact in the patch, but forgot, sorry.

This really should have been four patches (snapshot, Async VERIFIED, TX index status, and the directApi purge), but I got carried away as I kept finding the issues by the expanded test case.

phoenix-core/src/it/java/org/apache/phoenix/end2end/IndexToolIT.java
stoty: I've merged all three similar methods that I could find.

stoty:
Thank you @gjacoby126  and @gokceni.

I have implemented your suggestions, and explained the TX index change.

gokceni:
@stoty Thank you for implementing suggestions. There seems to be unit test failures. Could you please check if they are related to this patch? It LGTM but please make sure that tests are not broken.

gjacoby126 (approved):
+1, thanks @stoty (I'm assuming the permissions failures are due to them being flaky, but would be good to verify they pass locally before merging.)

stoty:
@gokceni 
I had to rebase this on top your TransformTool work.
For now, I have added logic in TransformMaintainer to set the VERIFIED status according to the old table indexing style.

Please check if that is acceptable solution (at least until PHOENIX-6617)

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/index/IndexTool.java
```diff
@@ -667,7 +672,12 @@ private Job configureJobForAsyncIndex() throws Exception {
                 try {
                     admin = pConnection.getQueryServices().getAdmin();
                     TableName hDdataTableName = TableName.valueOf(pDataTable.getPhysicalName().getBytes());
-                    snapshotName = new StringBuilder(hDdataTableName.toString()).append("-Snapshot").toString();
+                    snapshotName = new StringBuilder("INDEXTOOL-")
+                            .append(pDataTable.getName().getString())
+                            .append("-Snapshot-")
+                            .append(System.currentTimeMillis())
+                            .toString();
+                    //FIXME Drop this snapshot after we're done ?
```
joshelser, line 680: 👍🏼

phoenix-core/src/main/java/org/apache/phoenix/schema/transform/TransformMaintainer.java
```diff
@@ -133,6 +140,30 @@ private TransformMaintainer(final PTable oldTable, final PTable newTable, Phoeni
...
+            if (hasGlobalIndex) {
+                ConnectionQueryServices queryServices = connection.unwrap(PhoenixConnection.class).getQueryServices();
+                HTable oldHBaseTable = (HTable) queryServices.getTable(oldTable.getPhysicalName().getBytes());
+                if (!oldHBaseTable.getDescriptor().hasCoprocessor(org.apache.phoenix.hbase.index.Indexer.class.getName())) {
+                    newTableIndexVerified = true;
+                }
+            }
+        } catch (IOException e) {
+            throw new IllegalArgumentException("Could not determine empty column value", e);
+        } catch (SQLException e) {
```
joshelser, line 163: nit: mulit-catch `catch (IOException | SQLException e) {...`

joshelser (approved):
I'm a bit out of my depth (obviously, given the review feedback I can give) but looks fine to me :)

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/index/IndexTool.java
stoty: I think I'm going to open a follow-up ticket for that, this patch already has more than enough changes.

phoenix-core/src/main/java/org/apache/phoenix/schema/transform/TransformMaintainer.java
stoty: Thanks, done.

stoty:
I have removed the VERIFIED I added to TransformMaintainer, as it is not necessary after PHOENIX-6617.

I think that it would be a good idea to check if Transform handles the VERIFED status for transactional tables correctly, to avoid issues like IndexTool and Bulk Load had.

gokceni (approved):

PR #1371, opened 2022-01-05
PHOENIX-6620 TransformTool to fix unverified rows and do validation

gokceni:


gokceni:
@gjacoby126 @virajjasani

stoty:
I haven't reviewed the code in detail, but could you add some tests to make sure that VERIFIED status is preserved when doing the transform, especially for transactional tables that do not have the index maintainer coprocessor ?
We've had problems like that in IndexTool and BulkLoad before.

gokceni:
@stoty will do. Do you have a sample test that I can tweak for TransformTool? If the table is transactional and going trough transform and if an update is made, you want to see that the verified flag is set correctly?
Also could you point me to the fix for IndexTool?

stoty:
The indextool fix is PHOENIX-6611 .
The fix for setting the verified flag for transactional tables in the normal path was also included in that commit.

I am more worried abot the scenario where whe have a (transactional) table with global indexes, and where the rows are in verified state, losing the VERFIED status after transforming.

IndexToolIT.testSecondaryIndex() checks this here:
https://github.com/apache/phoenix/blob/01e1d2b4ef1758afef6b967f0959c8874e094596/phoenix-core/src/it/java/org/apache/phoenix/end2end/IndexToolIT.java#L332

The above test case checks for mutable, immutable, and transactional tables, doing the same for Transforms may also be useful.

I can see you are adding the option to check all rows in this patch, but not sure if we test that we don't break the already VERFIED rows' status during transform, where we don't need to touch the index rows.

You may have added check for this for the other cases, I haven't read the patch fully.

phoenix-core/src/it/java/org/apache/phoenix/end2end/TransformToolIT.java
```diff
@@ -413,59 +432,314 @@ public void testTransformMutationFailureRepair() throws Exception {
...
+
+    @Test
+    public void testTransformVerifiedForTransactionalTable() throws Exception {
+        // TODO: Column encoding is not supported for OMID. Have omid test for other type of transforms.
+        // For now, we don't support transforms other than storage and column encoding type change.
+        //testVerifiedForTransactionalTable("OMID");
+        testVerifiedForTransactionalTable("TEPHRA");
+    }
+
+    private void testVerifiedForTransactionalTable(String provider) throws Exception{
```
gokceni, line 519: @stoty added this test for TEPHRA. For OMID, we cannot have transform for now because the only type we support is changing column encoding and storage format encoding which is not compatible with OMID. When we have different type of transforms, we can add a test for OMID too.

stoty:
Thank you @gokceni.

gjacoby126:
@gokceni - looks like there's a compilation error (I think because of the moving of VERIFIED_BYTES in Istvan's recent patch)

gokceni:
@gjacoby126 fixed. Thanks

phoenix-core/src/it/java/org/apache/phoenix/end2end/TransformToolIT.java
gjacoby126: @gokceni -  I see existing error messages UNSUPPORTED_COLUMN_ENCODING_FOR_TXN_PROVIDER and UNSUPPORTED_STORAGE_FORMAT_FOR_TXN_PROVIDER -- do those get thrown on an ALTER of an OMID table or just a CREATE? If just a CREATE, it should probably do so for ALTER too. (Can be a separate JIRA.)

gjacoby126 (approved):
+1 @gokceni (but see a question below)

phoenix-core/src/it/java/org/apache/phoenix/end2end/TransformToolIT.java
```diff
@@ -413,59 +431,314 @@ public void testTransformMutationFailureRepair() throws Exception {
...
+            runTransformTool(args.toArray(new String[0]), 0);
+            assertEquals(1, getRowCountForEmptyColValue(conn, record.getNewPhysicalTableName(), VERIFIED_BYTES));
+
+
+            String upsertQuery = String.format("UPSERT INTO %s VALUES(?, ?, ?)", dataTableFullName);
+            PreparedStatement stmt1 = conn.prepareStatement(upsertQuery);
+
+            // Run again to check that VERIFIED row still remains verified
+            runTransformTool(args.toArray(new String[0]), 0);
+            assertEquals(1, getRowCountForEmptyColValue(conn, record.getNewPhysicalTableName(), VERIFIED_BYTES));
```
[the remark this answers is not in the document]
gokceni: @gjacoby126 for the UNSUPPORTED_STORAGE_FORMAT_FOR_TXN_PROVIDER question, it is thrown in createTableInternal function. But it is also thrown for ALTER. Here is the explanation: 1. The table cannot be created due to this check. 2. If you ALTER and change encoding, we use createTableInternal to create a new transforming table and it is thrown again during ALTER. 

For the test, I added it because of a previous issue Istvan fixed to and asked to make sure we are not breaking it. I wouldn't worry too much about Tephra.

gokceni:
merged

PR #1374, opened 2022-01-27
PHOENIX-6441 Remove TSOMockModule reference from OmidTransactionProvider

stoty:


phoenix-core/src/main/java/org/apache/phoenix/transaction/OmidTransactionProvider.java
```diff
@@ -120,84 +101,20 @@ public void close() throws IOException {}
...
-        }
+    // For testing only
+    public void injectTestService(HBaseTransactionManager transactionManager, CommitTable.Client commitTableClient) {
+        this.transactionManager = transactionManager;
+        this.commitTableClient = commitTableClient;
     }
 
     @Override
-    public Class<? extends RegionObserver> getCoprocessor() {
-        return OmidTransactionalProcessor.class;
```
chrajeshbabu, line 195: Any reason to use static string to specify the coprocessor class name?

chrajeshbabu (approved):

phoenix-core/src/main/java/org/apache/phoenix/transaction/OmidTransactionProvider.java
stoty: Originally I did this to handle missing classes better, but it's not really needed.
However, returning strings here simplifies the resto fo the code, as all HBase coprocessor admin methods expect and return class names.
I'm going to replace these with class.getName()

stoty (commented):
Thanks for the review @chrajeshbabu.
I'm going to make the change that you hinted at, and let another CI
run finish before committing.

PR #1385, opened 2022-02-01
PHOENIX-6638 Test suite fails with -Dwithout.tephra

stoty:


gjacoby126 (approved):
+1

stoty:
Thank you @gjacoby126 . I still had some weird failures with the patch when running with -Dwithout.tephra, so I moved the tephra class detection logic to TransactionFactory.

PR #1394, opened 2022-02-09
PHOENIX-6646 System tables are not upgraded after namespace migration

stoty:


stoty:
Tests hang on MigrateSystemTablesToSystemNamespaceIT

stoty:
The patch is a bit noisy because I did some reformatting to keep Yetus happy, but there the two main changes are the refactor of the upgrade logic to always run the full version upgrade code after the namespace migration, and handling missing system tables with a warning, instead of aborting the upgrade process.

phoenix-core/src/it/java/org/apache/phoenix/end2end/ConnectionQueryServicesTestImpl.java
```diff
@@ -53,23 +55,25 @@
 public class ConnectionQueryServicesTestImpl extends ConnectionQueryServicesImpl {
     private static final Logger LOGGER =
             LoggerFactory.getLogger(ConnectionQueryServicesTestImpl.class);
-    protected int NUM_SLAVES_BASE = 1; // number of slaves for the cluster
     // Track open connections to free them on close as unit tests don't always do this.
-    private Set<PhoenixConnection> connections = Sets.newHashSet();
+    private Set<PhoenixConnection> connections =
+            Collections.newSetFromMap(new ConcurrentHashMap<PhoenixConnection, Boolean>());
```
joshelser, line 60: TIL we don't have a ConcurrentHashSet in the JDK :)

phoenix-core/src/it/java/org/apache/phoenix/end2end/SnapshotTestTemplateIT.java
```diff
@@ -0,0 +1,119 @@
...
+        assertTrue(FileSystem.getLocal(config).exists(new Path(localPath, ".hbase-snapshot")));
+        FileSystem hdfsFs = FileSystem.get(config);
+        Path hdfsImportPath = new Path(hdfsFs.getHomeDirectory(), "snapshot-import" + "/" + key + "/");
+        assertTrue(hdfsFs.mkdirs(hdfsImportPath));
+        hdfsFs.copyFromLocalFile(localPath, hdfsImportPath);
+        hdfsImportPath = new Path(hdfsImportPath, localPath.getName());
+        assertTrue(hdfsFs.exists(new Path(hdfsImportPath, ".hbase-snapshot")));
+
+        //import the snapshot
+        ExportSnapshot exportTool = new ExportSnapshot();
```
joshelser, line 101: Does this automatically run with the `local` mapreduce runner? Not sure where to look to figure out if we already have a MiniMRCluster launched. Either should give you the same results, but debugging embedded MR jobs is painful.

phoenix-core/src/main/java/org/apache/phoenix/query/ConnectionQueryServicesImpl.java
```diff
@@ -4142,12 +4124,12 @@ private PhoenixConnection upgradeOtherSystemTablesIfRequired(
         // by this point we would have already taken mutex lock, hence
         // we can proceed with creation of snapshots and add table to
         // snapshot entries in systemTableToSnapshotMap
+        metaConnection = upgradeSystemStats(metaConnection,
+            systemTableToSnapshotMap);
```
joshelser, line 4128: Is this reordering significant?

phoenix-core/src/main/java/org/apache/phoenix/util/MetaDataUtil.java
```diff
@@ -275,7 +277,9 @@ public static long encodeVersion(String hbaseVersionStr, Configuration config) {
         long version =
         // Encode HBase major, minor, patch version
         (hbaseVersion << (Byte.SIZE * 5))
-                // Encode if systemMappingEnabled are enabled on the server side
+                // Encode if table namespace mapping is enabled on the server side
+                // Note that we DO NOT return information on whether system tables are mapped
+                // on the server side
```
joshelser, line 282: Should we?

joshelser (approved):
I'm no expert here, but this looks pretty reasonable to me.

stoty (commented):
Thanks for the review, Josh.

PR #1395, opened 2022-02-10
PHOENIX-5894 Table versus Table Full Outer join on Salted tables not …

chrajeshbabu:
…working


chrajeshbabu:
The test case failure because the salt column getting added twice. Looking into it.

chrajeshbabu:
The test case failure is not related to the fix.
@stoty @SergeySoldatov  could you please take a look at it.

ss77892 (approved):
LGTM.

chrajeshbabu:
Thanks @SergeySoldatov  for the reviews.

PR #1397, opened 2022-02-25
PHOENIX-6649 TransformTool to support views and tenant views

gokceni:


gokceni:
@gjacoby126

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/transform/PhoenixTransformWithViewsInputFormat.java
```diff
@@ -0,0 +1,116 @@
...
+                    columnMap.put(column.getName().getString(), column);
+                }
+
+                for (PTable decendant : legitimateDecendants) {
+                    if (decendant.getViewType() == PTable.ViewType.READ_ONLY) {
+                        continue;
+                    }
+                    PTable newView = Transform.getTransformedView(decendant, newDataTable, columnMap, true);
+                    QueryPlan queryPlan = getQueryPlan(newView, decendant, connection);
+                    inputSplits.addAll(generateSplits(queryPlan, configuration));
```
gjacoby126, line 77: Does the use of the setMultiInputMapperSplitSize in the TransformTool mean that multiple splits can be worked by the same mapper? (If so that's very good, because we can have lots of small views.) Either way, there's probably lots of future optimization work based on the view size, but we need better stats to be able to do that sort of thing so we can defer for now.

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/transform/PhoenixTransformWithViewsInputFormat.java
```diff
@@ -0,0 +1,116 @@
...
+                }
```
gjacoby126, line 78: What happens if the views are not disjoint? Do we just harmlessly transform the same data multiple times into the same shape? Seems like we should have a test for that if we don't already.

phoenix-core/src/it/java/org/apache/phoenix/end2end/transform/TransformToolIT.java
```diff
@@ -948,6 +959,175 @@ public void testTransformVerify_ForceCutover() throws Exception {
...
+            String sql = "SELECT VIEW_COL21, VIEW_COL22 FROM %s ";
+            ResultSet rs1 = tenantConn2.createStatement().executeQuery(String.format(sql, view2Name));
+            assertTrue(rs1.next());
+            assertEquals(111, rs1.getInt(1));
+            assertEquals("viewCol22", rs1.getString(2));
+        }
+    }
+
+
+    public static Connection getTenantConnection(String tenant) throws SQLException {
```
gjacoby126, line 1125: Could we please get a test for overlapping but still updatable views? (For example, one view is WHERE Foo = 'a' AND Bar = 'b', and another is just WHERE Foo = 'a'.)

gjacoby126: In TransformMonitorExtendedIT I believe you have an updatable and non-updatable tenant view that overlap, but since we skip transforming read-only views, that's not quite the same thing.

gjacoby126 (changes requested):
Looks good, just a request for one extra test please

phoenix-core/src/it/java/org/apache/phoenix/end2end/transform/TransformToolIT.java
gokceni: Sure. Will add one.

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/transform/PhoenixTransformWithViewsInputFormat.java
gokceni: Let me add a parameter to transform tool since it is 10 now (default).

phoenix-core/src/it/java/org/apache/phoenix/end2end/transform/TransformMonitorExtendedIT.java
```diff
@@ -156,16 +162,138 @@ public void testTransformTableWithNamespaceEnabled() throws Exception {
...
+            stmt1.setInt(1, 2);
+            stmt1.setString(2, "uname2");
+            stmt1.setInt(3, 95053);
+            stmt1.setString(4, "GLOBAL_VIEW");
+            stmt1.setInt(5, 111);
+            stmt1.setString(6, "viewcol2");
+            stmt1.executeUpdate();
+
+            createViewSql = "CREATE VIEW " + view2Name + " ( VIEW_COL1 INTEGER, VIEW_COL2 VARCHAR ) AS SELECT * FROM "
+                    + dataTableFullName1 + " where DATA='GLOBAL_VIEW' AND ZIP=95053";
```
gokceni, line 210: @gjacoby126 for overlapping view

gjacoby126 (changes requested):
Never mind, PhoenixTransformWithViewsInputFormat looks like it's missing an Apache License. Could you please fix before committing?

PR #1398, opened 2022-02-28
PHOENIX-6658 Replace HRegion.get() calls

stoty:


joshelser (changes requested):
The fix makes sense overall (switch over to using RegionScanner to take advantage of the off-heaping work in HBase as already written).

Istvan and I also talking about the value of looking at Phoenix now and trying to see how often we're doing unnecessary on-heap copies. He made https://issues.apache.org/jira/browse/PHOENIX-6621 for this. Made me wonder how we could measure the possible impact from off-heaping everything.

gjacoby126 (approved):
+1

PR #1399, opened 2022-03-02
PHOENIX-6501 Use batching when joining data table rows with uncovered global index rows

kadirozde:


lhofhansl:
Can we do this for local indexes as well? There is also a significant cost to seeking (even when done locally)

kadirozde:
> Can we do this for local indexes as well? There is also a significant cost to seeking (even when done locally)
> 
> In fact the only difference might be how we get the a table reference.

@lhofhansl, we can definitely do this for local indexes. There are some differences. The first is as you pointed out table vs region. For local indexes, both data and index are accessed via the local region while the global index is accessed remotely so we need to get the table for the global index and then get connections to access the table. The concern of which connection pool to use is not applicable to the local index.  The second difference is that there is one table region to retrieve data rows for the local index. However, for the global index, there can be many. So we discover the table region boundaries and access them in parallel for the global indexes using a thread pool, which is not necessary for the local index. The last difference is handling the row key offset for local indexes, which is not necessary for the global indexes. So, I thought instead of lumping local and global index batching together, we should handle them separately. I suggest having a separate Jira and PR for the local index.

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/UncoveredGlobalIndexRegionScanner.java
```diff
@@ -0,0 +1,387 @@
...
+                }
+                if (state == State.INITIAL) {
+                    indexRows = new ArrayList();
+                    dataRows = Maps.newConcurrentMap();
+                    indexToDataRowKeyMap = Maps.newTreeMap(Bytes.BYTES_COMPARATOR);
+                    state = State.SCANNING_INDEX;
+                }
+                if (state == State.SCANNING_INDEX) {
+                    do {
+                        List<Cell> row = new ArrayList<Cell>();
```
tkhurana, line 332: We are missing check for region closing here.

kadirozde: @tkhurana, the check for region closing is currently implemented only by the UngroupedAggregateRegionObserver coproc. When the preClose coproc hook is called on UngroupedAggregateRegionObserver by the region server, the coproc sets a flag and the region scanners on this coproc can check this flag to make sure that they do not prolong the region close operation.  This was implemented especially for long running scanners for index rebuild operations. UncoveredGlobalIndexRegionScanner is coproc agnostic and can be initiated from any coproc. So the region closing check can be executed only if all corpoc implements it. Please note that the region closing checks were implemented when the entire index region is scanned and built in one RPC call. This behavior has changed after we introduced server side paging. Please also note that in this PR, we have two levels of paging. One is based on the number of rows and the other based on the time spent within a single RPC call. So, I agree that region closing check is good in general but the need for it much less after server side paging. We need to restructure the server side code in order to implement region closing everywhere. I think that should be a separate improvement.

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/UncoveredGlobalIndexRegionScanner.java
```diff
@@ -0,0 +1,387 @@
...
+        try (ResultScanner resultScanner = dataHTable.getScanner(dataScan)) {
+            for (Result result = resultScanner.next(); (result != null);
+                 result = resultScanner.next()) {
+                dataRows.put(new ImmutableBytesPtr(result.getRow()), result);
+                if ((EnvironmentEdgeManager.currentTimeMillis() - startTime) >= pageSizeMs) {
+                    LOGGER.info("One of the scan tasks in UncoveredGlobalIndexRegionScanner"
+                            + " for region " + region.getRegionInfo().getRegionNameAsString()
+                            + " could not complete on time (in " + pageSizeMs+ " ms) and"
+                            + " will be resubmitted");
+                    state = State.SCANNING_DATA_INTERRUPTED;
```
tkhurana, line 220: state is a shared variable across all tasks so even if one task updates it it could be overwritten by other scan tasks before it is checked.

kadirozde: That is intention. The state is updated by these tasks only to signal that at least one of them needed to interrupt its scan. In this case, the scan operations for all tasks will be retried later.

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/UncoveredGlobalIndexRegionScanner.java
```diff
@@ -0,0 +1,387 @@
...
+                    break;
+                }
+            }
+        } catch (Throwable t) {
+            ServerUtil.throwIOException(dataHTable.getName().toString(), t);
```
tkhurana, line 225: I think the exceptionMessage object needs to be set here

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/UncoveredGlobalIndexRegionScanner.java
```diff
@@ -0,0 +1,387 @@
...
+        } catch (EarlyExitFailure e) {
+            throw new RuntimeException("Stopped while waiting for batch, quitting!", e);
+        }
+        int index = 0;
+        for (Boolean result : resultsAndFutures.getFirst()) {
+            if (result == null) {
+                Throwable cause = ServerUtil.getExceptionFromFailedFuture(
+                        resultsAndFutures.getSecond().get(index));
+                // there was a failure
+                throw new IOException(exceptionMessage, cause);
```
tkhurana, line 270: exceptionMessage might not be set.

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/UncoveredGlobalIndexRegionScanner.java
tkhurana: I see. All the tasks are updating the state to INTERRUPTED state. It doesn't matter which one sets it.

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/UncoveredGlobalIndexRegionScanner.java
kadirozde: I will do that

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/UncoveredGlobalIndexRegionScanner.java
```diff
@@ -0,0 +1,387 @@
...
+                if (state == State.SCANNING_DATA) {
+                    scanDataTableRows(startTime);
+                    indexRowIterator = indexRows.iterator();
+                }
+                if (state == State.READY) {
+                    if (indexRowIterator.hasNext()) {
+                        List<Cell> indexRow = indexRowIterator.next();
+                        result.addAll(indexRow);
+                        try {
+                            IndexUtil.addTupleAsOneCell(result, new ResultTuple(dataRows.get(new ImmutableBytesPtr(
```
tkhurana, line 362: Not very critical but we are assuming here that we will always find the data row corresponding to the index row. We can make the code more robust by handling the scenario where there is an extra verified row in the index so dataRows.get() will return null.

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/UncoveredGlobalIndexRegionScanner.java
```diff
@@ -0,0 +1,387 @@
...
+        try {
+            synchronized (innerScanner) {
+                if (state == State.READY && !indexRowIterator.hasNext()) {
+                    state = State.INITIAL;
+                }
+                if (state == State.INITIAL) {
+                    indexRows = new ArrayList();
+                    dataRows = Maps.newConcurrentMap();
+                    indexToDataRowKeyMap = Maps.newTreeMap(Bytes.BYTES_COMPARATOR);
+                    state = State.SCANNING_INDEX;
```
tkhurana, line 328: indexRowCount should also be initialized here for the next page

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/UncoveredGlobalIndexRegionScanner.java
kadirozde: Please note that the index table scan is not raw here. So we get only one valid version for a given index row. The corresponding data table row should exist. The only time it may not exist if the index table row is replicated but the data table row is not. So, if the data table row does not exist we should skip the index row. I will update the code.

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/UncoveredGlobalIndexRegionScanner.java
kadirozde: I will check it

tkhurana (approved):

phoenix-core/src/it/java/org/apache/phoenix/end2end/index/UncoveredGlobalIndexRegionScannerIT.java
```diff
@@ -0,0 +1,393 @@
...
+            assertEquals("efg", rs.getString(2));
+            assertFalse(rs.next());
+        }
+    }
+
+    private void assertIndexTableNotSelected(Connection conn, String dataTableName, String indexTableName, String sql)
+            throws Exception {
+        try {
+            assertExplainPlan(conn, sql, dataTableName, indexTableName);
+            throw new AssertionError("The index table should not be selected without an index hint");
```
gokceni, line 216: Isn't this caught below silently? Not sure if any exception is thrown from here

phoenix-core/src/main/java/org/apache/phoenix/iterate/RegionScannerFactory.java
```diff
@@ -20,11 +20,11 @@
 
 import static org.apache.phoenix.coprocessor.ScanRegionObserver.WILDCARD_SCAN_INCLUDES_DYNAMIC_COLUMNS;
 import static org.apache.phoenix.schema.types.PDataType.TRUE_BYTES;
-import static org.apache.phoenix.util.ScanUtil.getDummyResult;
-import static org.apache.phoenix.util.ScanUtil.getPageSizeMsForRegionScanner;
-import static org.apache.phoenix.util.ScanUtil.isDummy;
+import static org.apache.phoenix.util.ScanUtil.*;
```
gokceni, line 23: nit: star include also on line 27

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/UncoveredGlobalIndexRegionScanner.java
```diff
@@ -0,0 +1,396 @@
...
+                            }
+                            lastIndexRowKey = CellUtil.cloneRow(row.get(0));
+                            indexToDataRowKeyMap.put(lastIndexRowKey, indexMaintainer.buildDataRowKey(
+                                    new ImmutableBytesWritable(lastIndexRowKey), viewConstants));
+                            indexRows.add(row);
+                            indexRowCount++;
+                            if (hasMore && (EnvironmentEdgeManager.currentTimeMillis() - startTime)
+                                    >= pageSizeMs) {
+                                getDummyResult(lastIndexRowKey, result);
+                                return true;
```
gokceni, line 349: Is it safe to return without changing the state? Someone has to update the state. What if all tasks return?

phoenix-core/src/it/java/org/apache/phoenix/end2end/index/UncoveredGlobalIndexRegionScannerIT.java
```diff
@@ -0,0 +1,393 @@
...
+import org.apache.hadoop.hbase.coprocessor.ObserverContext;
+import org.apache.hadoop.hbase.coprocessor.RegionCoprocessorEnvironment;
+import org.apache.hadoop.hbase.coprocessor.SimpleRegionObserver;
+import org.apache.phoenix.filter.SkipScanFilter;
+import org.apache.phoenix.query.KeyRange;
+import org.apache.phoenix.thirdparty.com.google.common.collect.Maps;
+import org.apache.phoenix.end2end.NeedsOwnMiniClusterTest;
+import org.apache.phoenix.query.BaseTest;
+import org.apache.phoenix.query.QueryServices;
+import org.apache.phoenix.util.*;
```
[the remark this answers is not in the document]
kadirozde: This must be due to my IntelliJ config. I will fix it.

phoenix-core/src/it/java/org/apache/phoenix/end2end/index/UncoveredGlobalIndexRegionScannerIT.java
kadirozde: Good catch. I should throw a different exception here.

phoenix-core/src/main/java/org/apache/phoenix/iterate/RegionScannerFactory.java
kadirozde: Will fix it

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/UncoveredGlobalIndexRegionScanner.java
kadirozde: I will add a comment for this as follows : We do not need to change the state, State.SCANNING_INDEX since we will continue scanning the index table after the client drops the dummy requests and then calls the next method on its ResultScanner within ScanningResultIterator

gokceni (approved):
+1 Thanks @kadirozde

lhofhansl (approved):
I tried it.
It works great. With the caveat of performance when a lot of rows match the index condition (which to some extent is expected here).
With local indexes we can speed things up in this case by using ROW_INDEX_V1 as block encoding, I did not observe the same improvement for global indexes.
So we can use these only when the index is *very* selective - as in 0.1%-1% or so.

Good to have in the arsenal.

PR #1402, opened 2022-03-07
PHOENIX-6662 Failed to delete when PK with DESC with IN clause

gokceni:


phoenix-core/src/main/java/org/apache/phoenix/compile/WhereOptimizer.java
```diff
@@ -2197,10 +2203,11 @@ public SortOrder getSortOrder() {
                     // for the non-equality cases return actual sort order
                     //This work around should work
                     // but a more general approach can be taken.
-                    if(rvcElementOp == CompareOp.EQUAL ||
-                            rvcElementOp == CompareOp.NOT_EQUAL){
-                        return SortOrder.ASC;
-                    }
+                    //This optimization causes PHOENIX-6662 (when desc pk used with in clause)
```
yanxinyi, line 2206: nit: maybe remove instead of keeping it?

gokceni:
Merged

PR #1403, opened 2022-03-17
PHOENIX-6663 Use batching when joining data table rows with uncovered…

kadirozde:
… local index rows

### DIFF
--- a/phoenix-core/src/main/java/org/apache/phoenix/coprocessor/UncoveredIndexRegionScanner.java
+++ b/phoenix-core/src/main/java/org/apache/phoenix/coprocessor/UncoveredIndexRegionScanner.java
@@ -1,0 +1,329 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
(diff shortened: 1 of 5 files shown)


lhofhansl (approved):
Code looks great. I love the simplification and the unification between global and local uncovered indexes.

Also did a totally unscientific quick test on a full (complete with HDFS, etc) single node cluster. Simple table with 6m rows, and indexes that are about 10% selective, so 600k rows need to have main table columns merged in.

Before patch:
FAST_DIFF cold: 19.5s, FAST_DIFF hot: 17.8s, ROW_IDX cold: 9.3s, ROW_IDX hot: 5.6s

After Patch:
FAST_DIFF cold: 7.1s, FAST_DIFF hot: 3.0s, ROW_IDX cold: 4.7s, ROW_IDX hot: 2.6s

So this makes uncovered local indexes far more useful.

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/UncoveredIndexRegionScanner.java
```diff
@@ -0,0 +1,325 @@
...
+        boolean hasMore;
+        region.startRegionOperation();
+        try {
+            synchronized (innerScanner) {
+                if (state == State.READY && !indexRowIterator.hasNext()) {
+                    state = State.INITIAL;
+                }
+                if (state == State.INITIAL) {
+                    indexRowCount = 0;
+                    indexRows = new ArrayList();
```
lhofhansl, line 294: Nit: Should be `new ArrayList<>()`

kadirozde: Will fix it

PR #1406, opened 2022-03-21
PHOENIX-6665 PreparedStatement#getMetaData() fails on parametrized "s…

stoty:
…elect next ? values for SEQ"


stoty:
SequenceBulkAllocationIT.testMultipleDifferentExpressionsForSameSequence failures seem relevant.

joshelser (approved):
Would be nice to document the `NOOP` enum value you added, but I'm not super worried about that since none of the other field were documented.

It looks like we have test coverage already in `SequenceBulkAllocationIT#testNextValuesForSequenceUsingBinds`, so +1 from me.

PR #1409, opened 2022-03-24
CDPD-34599 Add support for TRUNCATE TABLE

ragarkar:
1. Added support for TRUNCATE TABLE <table> statement in phoenix.
2. Added an integration test to test the feature.

### DIFF
--- a/phoenix-core/src/it/java/org/apache/phoenix/end2end/TruncateTableIT.java
+++ b/phoenix-core/src/it/java/org/apache/phoenix/end2end/TruncateTableIT.java
@@ -1,0 +1,186 @@
+package org.apache.phoenix.end2end;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertTrue;
+import static org.junit.Assert.fail;
+
+import java.sql.Connection;
+import java.sql.DriverManager;
+import java.sql.PreparedStatement;
+import java.sql.ResultSet;
+import java.sql.SQLException;
+import java.util.Properties;
+import org.apache.phoenix.query.QueryServices;
+import org.apache.phoenix.schema.TableNotFoundException;
+import org.junit.Test;
+import org.junit.experimental.categories.Category;
+
+@Category(ParallelStatsDisabledTest.class)
+public class TruncateTableIT extends ParallelStatsDisabledIT {
+
+    @Test
+    public void testTruncateTable() throws SQLException {
+        Properties props = new Properties();
+        Connection conn = DriverManager.getConnection(getUrl(), props);
+        String tableName = generateUniqueName();
+        try {
+            String createTableDDL =
+                "CREATE TABLE " + tableName + " (pk char(2) not null primary key)";
+            conn.createStatement().execute(createTableDDL);
+            conn.close();
+
+            conn = DriverManager.getConnection(getUrl(), props);
+            conn.setAutoCommit(true);
+            PreparedStatement stmt = conn.prepareStatement(
+                "UPSERT INTO " + tableName + " VALUES('a')");
+            stmt.execute();
+            conn.close();
+
+            conn = DriverManager.getConnection(getUrl(), props);
+            ResultSet rs = conn.createStatement().executeQuery("SELECT COUNT(*) FROM " + tableName);
+            assertTrue(rs.next());
+            assertEquals(1, rs.getInt(1));
+            conn.close();
+
+            conn = DriverManager.getConnection(getUrl(), props);
+            conn.createStatement().execute("TRUNCATE TABLE " + tableName);
+            conn.close();
+
+            conn = DriverManager.getConnection(getUrl(), props);
+            rs = conn.createStatement().executeQuery("SELECT COUNT(*) FROM " + tableName);
+            assertTrue(rs.next());
+            assertEquals(0, rs.getInt(1));
+            conn.createStatement().execute("DROP TABLE " + tableName);
+            conn.close();
+        } catch (SQLException e) {
+            fail();
+        }
+    }
+
+    @Test
+    public void testTruncateTableNotExist() throws Exception {
+        Properties props = new Properties();
+        Connection conn = DriverManager.getConnection(getUrl(), props);
+        String tableName = "nonExistentTable";
+        try {
+            conn.createStatement().execute("TRUNCATE TABLE " + tableName);
+            fail();
+        } catch (TableNotFoundException e) {
+            return;
+        }
+    }
+
+    @Test
+    public void testTruncateTableNonExistentSchema() throws SQLException {
+        Properties props = new Properties();
+        Connection conn = DriverManager.getConnection(getUrl(), props);
+        String schemaName = "nonExistentSchema";
+        String tableName = generateUniqueName();
+
+        try {
+            conn.createStatement()
+                .execute("CREATE TABLE " + tableName + " (C1 INTEGER PRIMARY KEY)");
+            conn.setAutoCommit(true);
+            PreparedStatement stmt = conn.prepareStatement(
+                "UPSERT INTO " + tableName + " VALUES(1)");
+            stmt.execute();
+            ResultSet rs = conn.createStatement().executeQuery("SELECT COUNT(*) FROM " + tableName);
+            assertTrue(rs.next());
+            assertEquals(1, rs.getInt(1));
+            conn.createStatement().execute("TRUNCATE TABLE " + schemaName + "." + tableName);
+            fail();
+        } catch (SQLException e) {
+            conn.createStatement().execute("DROP TABLE " + tableName);
+        }
+    }
+
+    @Test
+    public void testTruncateTableWithImplicitSchema() throws SQLException {
+        Properties props = new Properties();
+        Connection conn = DriverManager.getConnection(getUrl(), props);
+        String schemaName = generateUniqueName();
+        String tableName = generateUniqueName();
+        String createTableWithSchema = "CREATE TABLE " + schemaName + "." + tableName + " (C1 char(2) NOT NULL PRIMARY KEY)";
+
+        try {
+            conn.createStatement().execute(createTableWithSchema);
+            conn.setAutoCommit(true);
+            PreparedStatement stmt = conn.prepareStatement("UPSERT INTO " + schemaName + "." + tableName + " values('a')");
+            stmt.execute();
+            ResultSet rs = conn.createStatement().executeQuery("SELECT COUNT(*) FROM " + schemaName + "." + tableName);
+            assertTrue(rs.next());
+            assertEquals(1, rs.getInt(1));
+            stmt = conn.prepareStatement("UPSERT INTO " + schemaName + "." + tableName + " values('b')");
+            stmt.execute();
+            rs = conn.createStatement().executeQuery("SELECT COUNT(*) FROM " + schemaName + "." + tableName);
+            assertTrue(rs.next());
+            assertEquals(2, rs.getInt(1));
+            conn.close();
+
+            conn = DriverManager.getConnection(getUrl(), props);
+            conn.createStatement().execute("TRUNCATE TABLE " + schemaName + "." + tableName);
+            conn.close();
+
+            conn = DriverManager.getConnection(getUrl(), props);
+            rs = conn.createStatement().executeQuery("SELECT COUNT(*) FROM " + schemaName + "." + tableName);
+            assertTrue(rs.next());
+            assertEquals(0, rs.getInt(1));
+            conn.close();
+
+            conn = DriverManager.getConnection(getUrl(), props);
+            conn.createStatement().execute("DROP TABLE " + schemaName + "." + tableName);
+            conn.close();
+        } catch (SQLException e) {
+            fail();
+        }
+    }
+
+    @Test
+    public void testTruncateTableWithExplicitSchema() throws SQLException {
+        Properties props = new Properties();
+        props.setProperty(QueryServices.IS_NAMESPACE_MAPPING_ENABLED, Boolean.toString(true));
+
+        String schemaName = generateUniqueName();
+        String tableName = generateUniqueName();
+
+        String schemaCreateStmt = "CREATE SCHEMA IF NOT EXISTS " + schemaName;
+        String tableCreateStmt = "CREATE TABLE IF NOT EXISTS " + tableName + " (C1 char(2) NOT NULL PRIMARY KEY) SALT_BUCKETS=100";
+
+        try {
+            Connection conn = DriverManager.getConnection(getUrl(), props);
+            conn.setAutoCommit(true);
+            conn.createStatement().execute(schemaCreateStmt);
+            conn.createStatement().execute("USE " + schemaName);
+            conn.createStatement().execute(tableCreateStmt);
+
+            PreparedStatement stmt = conn.prepareStatement("UPSERT INTO " + tableName + " VALUES ('a')");
+            stmt.execute();
+
+            ResultSet rs = conn.createStatement().executeQuery("SELECT COUNT(*) FROM " + tableName);
+            assertTrue(rs.next());
+            assertEquals(1, rs.getInt(1));
+            conn.close();
+
+            conn = DriverManager.getConnection(getUrl(), props);
+            conn.setAutoCommit(true);
+            conn.createStatement().execute("USE " + schemaName);
+            stmt = conn.prepareStatement("UPSERT INTO " + tableName + " VALUES ('b')");
+            stmt.execute();
+            rs = conn.createStatement().executeQuery("SELECT COUNT(*) FROM " + tableName);
+            assertTrue(rs.next());
+            assertEquals(2, rs.getInt(1));
+            conn.close();
+
+            conn = DriverManager.getConnection(getUrl(), props);
+            conn.createStatement().execute("USE " + schemaName);
+            conn.createStatement().execute("TRUNCATE TABLE " + tableName);
+            rs = conn.createStatement().executeQuery("SELECT COUNT(*) FROM " + tableName);
+            assertTrue(rs.next());
+            assertEquals(0, rs.getInt(1));
+            conn.createStatement().execute("DROP TABLE " + schemaName + "." + tableName);
+            conn.close();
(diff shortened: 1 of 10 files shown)


stoty:
Please do not use private JIRA IDs in the Apache project.
Use the PHOENIX-300 ticket ID.

stoty:
There are a few issues to be addressed:

- child tables (global secondary index tables and and global secondary indexes on views) also need to be handled.
- This doesn't work for multitenant tables from tenant-specific connections (i think simply erroring out in that case is fine)
- Some sanity checks for critical system tables would also be needed.
- Need to think about what happens when trying to access the tables while truncating.

stoty:
You are preserving the splits during truncate, which is good, but you may also want to consider exposing an option that lets users remove the existing split points. (not for salted tables, of course.)

ragarkar:
> 

Thanks for looking into these changes. How about we provide an extension to the TRUNCATE TABLE statement to preserve the splits? We can have an optional clause something like:

TRUNCATE TABLE <tablename> PRESERVE SPLITS;

stoty:
That sounds good, @ragarkar .

stoty:
Or rather make preserving splits the default, and add an option to NOT preserver them.

phoenix-core/src/main/java/org/apache/phoenix/query/ConnectionQueryServicesImpl.java
```diff
@@ -2223,6 +2223,24 @@ private void dropTable(byte[] tableNameToDelete) throws SQLException {
         dropTables(Collections.<byte[]>singletonList(tableNameToDelete));
     }
 
+    @Override
+    public void truncateTable(String schemaName, String tableName, boolean isNamespaceMapped) throws SQLException {
+        SQLException sqlE = null;
+        TableName hbaseTableName = SchemaUtil.getPhysicalTableName(SchemaUtil.getTableName(schemaName, tableName).getBytes(), isNamespaceMapped);
+        try {
+            Admin admin = getAdmin();
+            admin.disableTable(hbaseTableName);
```
chrajeshbabu, line 2232: Truncate table automatically disables the table we need not do it explicitly.

phoenix-core/src/main/java/org/apache/phoenix/query/ConnectionQueryServicesImpl.java
```diff
@@ -2223,6 +2223,24 @@ private void dropTable(byte[] tableNameToDelete) throws SQLException {
...
+            admin.truncateTable(hbaseTableName, true);
```
chrajeshbabu, line 2233: Have you checked the truncate table always waiting for the table to get created and enabled? If not we might need to add retries in case if required.

chrajeshbabu, line 2233: After the truncate table the regions cache in hbase is not longer valid so would be better to call
ConnectionQueryServicesImpl#clearTableRegionCache after that so that further queries fetch new locations of table regions.

phoenix-core/src/main/java/org/apache/phoenix/schema/MetaDataClient.java
```diff
@@ -3303,6 +3304,30 @@ private static void throwIfInsufficientColumns(String schemaName, String tableNa
         }
     }
 
+    public MutationState truncateTable(TruncateTableStatement statement) throws SQLException {
+        String schemaName = connection.getSchema() != null && statement.getTableName().getSchemaName() == null
+            ? connection.getSchema() : statement.getTableName().getSchemaName();
+        String tableName = statement.getTableName().getTableName();
+        boolean isNamespaceMapped = SchemaUtil.isNamespaceMappingEnabled(statement.getTableType(), connection.getQueryServices().getProps());
+        boolean wasAutoCommit = connection.getAutoCommit();
```
chrajeshbabu, line 3312: Since we are not doing any meta update changes here changing getting autocommit status and setting back is not much impact so better to remove it.

phoenix-core/src/it/java/org/apache/phoenix/end2end/TruncateTableIT.java
```diff
@@ -0,0 +1,186 @@
...
+import java.sql.ResultSet;
+import java.sql.SQLException;
+import java.util.Properties;
+import org.apache.phoenix.query.QueryServices;
+import org.apache.phoenix.schema.TableNotFoundException;
+import org.junit.Test;
+import org.junit.experimental.categories.Category;
+
+@Category(ParallelStatsDisabledTest.class)
+public class TruncateTableIT extends ParallelStatsDisabledIT {
```
chrajeshbabu, line 19: You can add a test case creating tables with split keys which is missing.

phoenix-core/src/it/java/org/apache/phoenix/end2end/TruncateTableIT.java
```diff
@@ -0,0 +1,186 @@
...
+        Connection conn = DriverManager.getConnection(getUrl(), props);
+        String tableName = generateUniqueName();
+        try {
+            String createTableDDL =
+                "CREATE TABLE " + tableName + " (pk char(2) not null primary key)";
+            conn.createStatement().execute(createTableDDL);
+            conn.close();
+
+            conn = DriverManager.getConnection(getUrl(), props);
+            conn.setAutoCommit(true);
```
chrajeshbabu, line 33: This is by default not required.

phoenix-core/src/it/java/org/apache/phoenix/end2end/TruncateTableIT.java
```diff
@@ -0,0 +1,186 @@
...
+            PreparedStatement stmt = conn.prepareStatement(
+                "UPSERT INTO " + tableName + " VALUES('a')");
+            stmt.execute();
+            conn.close();
+
+            conn = DriverManager.getConnection(getUrl(), props);
```
chrajeshbabu, line 39: Need not recreate the connection every time. Can make use the same connection and close it at the end of test case.

phoenix-core/src/it/java/org/apache/phoenix/end2end/TruncateTableIT.java
chrajeshbabu: You can add the code to check the number of regions after truncate also.

phoenix-core/src/it/java/org/apache/phoenix/end2end/TruncateTableIT.java
```diff
@@ -0,0 +1,186 @@
...
+            conn.createStatement().execute("DROP TABLE " + tableName);
+            conn.close();
+        } catch (SQLException e) {
+            fail();
+        }
+    }
+
+    @Test
+    public void testTruncateTableNotExist() throws Exception {
+        Properties props = new Properties();
```
chrajeshbabu, line 62: Need not pass empty properties, can call directly DriverManager.getConnection(getUrl())

phoenix-core/src/it/java/org/apache/phoenix/end2end/TruncateTableIT.java
```diff
@@ -0,0 +1,186 @@
...
+        Connection conn = DriverManager.getConnection(getUrl(), props);
```
gjacoby126, line 63: Please create the connection using try-with-resources to make sure that it gets closed at the end of the test.

phoenix-core/src/it/java/org/apache/phoenix/end2end/TruncateTableIT.java
```diff
@@ -0,0 +1,186 @@
...
+import org.junit.Test;
+import org.junit.experimental.categories.Category;
+
+@Category(ParallelStatsDisabledTest.class)
+public class TruncateTableIT extends ParallelStatsDisabledIT {
+
+    @Test
+    public void testTruncateTable() throws SQLException {
+        Properties props = new Properties();
+        Connection conn = DriverManager.getConnection(getUrl(), props);
```
gjacoby126, line 24: Best to create the connection using try-with-resources

phoenix-core/src/it/java/org/apache/phoenix/end2end/TruncateTableIT.java
```diff
@@ -0,0 +1,186 @@
...
+
+            conn = DriverManager.getConnection(getUrl(), props);
+            conn.createStatement().execute("TRUNCATE TABLE " + tableName);
+            conn.close();
+
+            conn = DriverManager.getConnection(getUrl(), props);
+            rs = conn.createStatement().executeQuery("SELECT COUNT(*) FROM " + tableName);
+            assertTrue(rs.next());
+            assertEquals(0, rs.getInt(1));
+            conn.createStatement().execute("DROP TABLE " + tableName);
```
gjacoby126, line 53: Consider dropping the table in a finally block

phoenix-core/src/it/java/org/apache/phoenix/end2end/TruncateTableIT.java
```diff
@@ -0,0 +1,186 @@
...
+            PreparedStatement stmt = conn.prepareStatement(
+                "UPSERT INTO " + tableName + " VALUES(1)");
+            stmt.execute();
+            ResultSet rs = conn.createStatement().executeQuery("SELECT COUNT(*) FROM " + tableName);
+            assertTrue(rs.next());
+            assertEquals(1, rs.getInt(1));
+            conn.createStatement().execute("TRUNCATE TABLE " + schemaName + "." + tableName);
+            fail();
+        } catch (SQLException e) {
+            conn.createStatement().execute("DROP TABLE " + tableName);
```
gjacoby126, line 93: Good to assert on the actual SQLException you're expecting.

phoenix-core/src/it/java/org/apache/phoenix/end2end/TruncateTableIT.java
```diff
@@ -0,0 +1,186 @@
...
+            conn.close();
+        } catch (SQLException e) {
+            fail();
+        }
+    }
+
+    @Test
+    public void testTruncateTableWithExplicitSchema() throws SQLException {
+        Properties props = new Properties();
+        props.setProperty(QueryServices.IS_NAMESPACE_MAPPING_ENABLED, Boolean.toString(true));
```
gjacoby126, line 141: A given Phoenix cluster can only have Namespace mapping enabled or disabled; it can't be enabled for some tables and disabled for others. This usually means either having to shut down the minicluster and spin it back up with namespace enabled, or putting this is in a separate IT test.

phoenix-core/src/main/antlr3/PhoenixSQL.g
```diff
@@ -460,6 +462,12 @@ create_table_node returns [CreateTableStatement ret]
         (SPLIT ON s=value_expression_list)?
         {ret = factory.createTable(t, p, c, pk, s, PTableType.TABLE, ex!=null, null, null, getBindCount(), im!=null ? true : null); }
     ;
+
+// Parse a truncate table statement.
+truncate_table_node returns [TruncateTableStatement ret]
+    :   TRUNCATE TABLE t=from_table_name
+        {ret = factory.truncateTable(t, PTableType.TABLE);}
```
gjacoby126, line 469: What happens if someone tries to truncate a view? Looks like we're assuming that from_table_name is a table, but there's nothing in the grammar to enforce that, is there?

phoenix-core/src/main/java/org/apache/phoenix/schema/MetaDataClient.java
```diff
@@ -3303,6 +3304,30 @@ private static void throwIfInsufficientColumns(String schemaName, String tableNa
...
+        try {
+            PTable ptable = connection.getTable(
+                new PTableKey(connection.getTenantId(), SchemaUtil.getTableName(schemaName, tableName)));
+        } catch (TableNotFoundException e) {
+            throw e;
+        }
+        try {
+            connection.getQueryServices().truncateTable(schemaName, tableName, isNamespaceMapped);
```
gjacoby126, line 3320: There should be a check in here to make sure that we don't try to truncate a SYSTEM table, and if we do, throw a SQLException with a new SQLExceptionCode explaining we don't allow that.

gjacoby126: Likewise, I don't think you should be able to truncate an index directly (but an index should get truncated if its base table gets truncated, as I think @chrajeshbabu  pointed out.)

gjacoby126 (changes requested):
Nice feature @ragarkar . Thanks for the patch, this has been wanted by the community for a long time. Some thoughts below on polishing it and adding some extra safety

phoenix-core/src/main/java/org/apache/phoenix/query/ConnectionQueryServicesImpl.java
kabhishek4: I had removed the call to disable the table. However, all the tests started failing with org.apache.hadoop.hbase.TableNotDisabledException: Hence, I put that back. Looks like explicit call to disable is needed.

phoenix-core/src/main/java/org/apache/phoenix/query/ConnectionQueryServicesImpl.java
kabhishek4: Yes, it waits for the table to get created and enabled. I have added assert for the same in the code.

kabhishek4: done.

phoenix-core/src/main/java/org/apache/phoenix/schema/MetaDataClient.java
kabhishek4: done

phoenix-core/src/it/java/org/apache/phoenix/end2end/TruncateTableIT.java
kabhishek4: done

phoenix-core/src/it/java/org/apache/phoenix/end2end/TruncateTableIT.java
kabhishek4: removed

phoenix-core/src/it/java/org/apache/phoenix/end2end/TruncateTableIT.java
kabhishek4: addressed

phoenix-core/src/it/java/org/apache/phoenix/end2end/TruncateTableIT.java
kabhishek4: addressed

phoenix-core/src/it/java/org/apache/phoenix/end2end/TruncateTableIT.java
kabhishek4: addressed

phoenix-core/src/it/java/org/apache/phoenix/end2end/TruncateTableIT.java
kabhishek4: done

phoenix-core/src/it/java/org/apache/phoenix/end2end/TruncateTableIT.java
kabhishek4: done

phoenix-core/src/it/java/org/apache/phoenix/end2end/TruncateTableIT.java
kabhishek4: done

phoenix-core-client/src/main/antlr3/PhoenixSQL.g
```diff
@@ -460,6 +462,12 @@ create_table_node returns [CreateTableStatement ret]
         (SPLIT ON s=value_expression_list)?
         {ret = factory.createTable(t, p, c, pk, s, PTableType.TABLE, ex!=null, null, null, getBindCount(), im!=null ? true : null); }
     ;
+
+// Parse a truncate table statement.
+truncate_table_node returns [TruncateTableStatement ret]
+    :   TRUNCATE TABLE t=from_table_name
+        {ret = factory.truncateTable(t, PTableType.TABLE);}
```
kabhishek4: Added validation check for this case in MetaDataClient.java

phoenix-core/src/main/java/org/apache/phoenix/schema/MetaDataClient.java
kabhishek4: done

phoenix-core/src/it/java/org/apache/phoenix/end2end/TruncateTableIT.java
```diff
@@ -0,0 +1,533 @@
...
+      }
+    } catch (SQLException e) {
+      fail(e.getMessage());
+    }
+  }
+
+  @Test
+  public void testTruncateTableWithExplicitSchema() throws SQLException {
+    Properties props = new Properties();
+    props.setProperty(QueryServices.IS_NAMESPACE_MAPPING_ENABLED, Boolean.toString(true));
```
stoty, line 182: You should either
1. split this into two test files, one with namespace mapping enabled andone with disabled, and classify those are @NeedOwnCluster tests.
2. Or you should set phoenix.schema.mapSystemTablesToNamespace to false to avoid migrating the system tables on establishing the connection.

phoenix-core/src/it/java/org/apache/phoenix/end2end/TruncateTableIT.java
kabhishek4: Done

phoenix-core/src/it/java/org/apache/phoenix/end2end/TruncateTableIT.java
kabhishek4: addressed this comment and uploaded new patch.

tkhurana (approved):

PR #1412, opened 2022-04-04
PHOENIX-1291 Add ILIKE optimization for initial literal part

Rodolffo:
For optimizing the ILIKE queries, during the optimization phase, a set of ranges (2^n to be precise) are created for limiting results, if the pattern starts with a string literal, as such:
'abc%' will generate 'abc', 'Abc', 'aBc', 'abC', 'ABc', 'AbC', 'aBC', and 'ABC'.

I'm not completely sure if this solution is valid, and I'd like to do a few performance tests on how it performs which I'll share here, but in the meanwhile I'd appreaciate any feedback and suggestions. I mostly followed James Taylor's comments and the discussion in https://issues.apache.org/jira/browse/PHOENIX-1273, which the original ticket refers to.

### DIFF
--- a/phoenix-core/src/main/java/org/apache/phoenix/compile/WhereOptimizer.java
+++ b/phoenix-core/src/main/java/org/apache/phoenix/compile/WhereOptimizer.java
@@ -29,8 +29,6 @@ import java.util.Map;
 import java.util.NoSuchElementException;
 import java.util.Set;
 
-import org.apache.phoenix.expression.DelegateExpression;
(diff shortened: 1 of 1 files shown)


rudolfzundel-seon:
Also the number of ranges generated grows exponentially as the prefix size is larger, so it may make sense to introduce a reasonable limit on prefix size generated.

Rodolffo:
Also the number of ranges generated grows exponentially as the prefix size is larger, so it may make sense to introduce a reasonable limit on prefix size generated.

PR #1415, opened 2022-04-14
PHOENIX-6686 Update jackson to 2.12.6.1

stoty:
also switch to using jackson-bom for version version management

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -97,7 +97,7 @@
     <top.dir>${project.basedir}/..</top.dir>
 
     <!-- Dependency versions -->
-    <jackson.version>2.12.4</jackson.version>
+    <jackson-bom.version>2.12.6.20220326</jackson-bom.version>
     <antlr.version>3.5.2</antlr.version>
     <reload4j.version>1.2.19</reload4j.version>
     <disruptor.version>3.3.6</disruptor.version>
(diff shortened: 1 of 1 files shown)


gjacoby126:
@stoty - On the one hand, just taking in the bom simplifies the pom and guarantees that the different jackson subprojects will always be compatible with each other. On the other hand, that adds quite a few transitive dependencies we didn't formerly have by taking everything jackson does, correct?

stoty:
No @gjacoby126 , it only imports the dependecyManagement entries for the jackson components.
The
`<scope>import</scope>
<type>pom</type>`
does the trick. 
(It's the first time see this maven feature used)

I'm attaching the old and new dependency tree to the ticket.

stoty:
The only side effect is that importing the bom overrides the versions for

the 
_com.fasterxml.jackson.module:jackson-module-jaxb-annotations_ and
_com.fasterxml.jackson.jaxrs:jackson-jaxrs-json-provider_ 

artifacts pulled in by in _org.apache.hadoop:hadoop-mapreduce-client-core_,
and those new versions in turn pull in the new _jakarta.xml.bind:jakarta.xml.bind-api_ and jakarta.activation:jakarta.activation-api artifacts.

I don't expect that overriding the artifacts in Hadoop would be a problem, we're already overriding plenty of Hadoop/HBase dependency versions, and the test suite was successful
(even though the postback to PR didn't work )

https://ci-hadoop.apache.org/job/Phoenix/job/Phoenix-PreCommit-GitHub-PR/view/change-requests/job/PR-1415/2/consoleFull

gjacoby126 (approved):
+1, thanks for the explanation, @stoty

PR #1416, opened 2022-04-14
PHOENIX-6682 Jenkins tests are failing for Java 11.0.14.1

stoty:
pin Java11 to the Ubuntu 20.04 release version


gjacoby126 (approved):
+1

stoty:
Last minute testing shows that apt-get also needs _--allow-downgrades_, otherwise the downgrade does not happen.

Going to commit with that change.

PR #1425, opened 2022-04-26
PHOENIX-6693 Remove Hbase 2.1 and Hbase 2.2 support from Phoenix

stoty:


phoenix-core/src/main/java/org/apache/phoenix/coprocessor/BaseScannerRegionObserver.java
```diff
@@ -403,14 +419,165 @@ RegionScanner getWrappedScanner(final ObserverContext<RegionCoprocessorEnvironme
...
+        boolean timeRangeIsLatest = scan.getTimeRange().getMax() == HConstants.LATEST_TIMESTAMP;
+        boolean timestampIsTransactional =
+            isTransactionalTimestamp(scan.getTimeRange().getMax());
+        return isRaw
+            || keepDeletedCells
+            || timeRangeIsLatest
+            || timestampIsTransactional;
+    }
+
+    private boolean isTransactionalTimestamp(long ts) {
```
gjacoby126, line 498: @stoty - since we've deprecated Tephra this isTransactionalTimestamp method can probably go away. It's a hack to detect when Tephra is adjusting the HBase timestamps to be nanos, but iirc Omid doesn't change timestamps in the same way.

gjacoby126: If you'd rather wait until a more general "get rid of Tephra references in the code" PR, that's fine too.

gjacoby126:
@stoty - looks like a bunch of test failures around the refCount leak check in tearDown, particularly in DateTimeIT and SequenceIT, plus a couple misc. test failures.

phoenix-core/src/main/java/org/apache/hadoop/hbase/regionserver/IndexHalfStoreFileReader.java
```diff
@@ -89,7 +92,12 @@ public IndexHalfStoreFileReader(final FileSystem fs, final Path p, final CacheCo
             final byte[][] viewConstants, final RegionInfo regionInfo,
             byte[] regionStartKeyInHFile, byte[] splitKey, boolean primaryReplicaStoreFile,
             AtomicInteger refCount, RegionInfo currentRegion) throws IOException {
-        super(fs, p, in, size, cacheConf, primaryReplicaStoreFile, refCount, conf);
+        super(new ReaderContext(p, in, size, new HFileSystem(fs), primaryReplicaStoreFile,
+            ReaderType.STREAM),
+            new HFileInfo(new ReaderContext(p, in, size, new HFileSystem(fs),
```
joshelser, line 97: Looking at HBase code, it looks like we shouldn't leak any copied filesystem objects here.

We had a lot of perf issues with STREAM reads where they ended up being slower than PREAD. It might be safer to start with PREAD and return to this specifically when we're doing perf testing on this code.

phoenix-core/src/main/java/org/apache/phoenix/transaction/OmidTransactionTable.java
```diff
@@ -372,4 +373,8 @@ public long getOperationTimeout(TimeUnit unit) {
         throw new UnsupportedOperationException();
     }
 
+    @Override
+    public RegionLocator getRegionLocator() throws IOException {
+        throw new UnsupportedOperationException();
```
joshelser, line 378: Phoenix doesn't end up calling this at all, I"m assuming?

joshelser (approved):
Generally looks ok. A little about the index half reader, but if you're just copying from something done elsewhere in Phoenix, I assume it's fine.

If it's net-new, let's think about it carefully.

stoty:
@gjacoby126 The refCount tests have been randomly failing for a long time.

ConcurrentMutationsExtendedIT and PartialIndexRebuilderIT are respectively flakey and always fail even without this patch.

I've investigated ConcurrentMutationsExtendedIT and PartialIndexRebuilderIT, and those failures seem to be tied to Hbase 2.4. PartialIndexRebuilderIT always fails with HBase 2.4.11, while ConcurrentMutationsExtendedIT is flakey with 2.4.11.
I've run these two tests with HBase 2.3.7, and both passed 3 out of 3 times.

Opened PHOENIX-6702 for the test failures.

@elserj Yes, I've aimed to keep the behaviour exactly the same as before, just moving some code around classes.

Opened PHOENIX-6703 for the STREAM vs PREAD issue.

PR #1427, opened 2022-04-27
PHOENIX-6696 Drop legacy phoenix-client jar

stoty:


virajjasani:
Doesn't seem relevant to this PR but I see lots of `refCount leaked` errors in `DateTimeIT`

virajjasani:
Otherwise +1 for dropping phoenix-client and making only embedded jar available (target version only 5.2.x)

phoenix-assembly/src/build/components/all-common-jars.xml
```diff
@@ -55,13 +47,5 @@
         <include>phoenix-pherf.jar</include>
       </includes>
     </fileSet>
-    <fileSet>
-      <directory>${project.basedir}/../phoenix-tools/target</directory>
```
gjacoby126, line 59: I'm curious -- why does this affect phoenix-tools?

gjacoby126 (commented):
lgtm but one question about phoenix-tools

pom.xml
```diff
@@ -1519,7 +1494,6 @@
       </activation>
       <modules>
         <module>phoenix-client-parent</module>
-        <module>phoenix-client-parent/phoenix-client</module>
```
joshelser, line 1522: Drop `phoenix-client-parent/phoenix-client` directory and pom.xml too?

phoenix-assembly/src/build/components/all-common-jars.xml
stoty: We've dropped the phoenix-tools module/artifact some time ago, this is just an (unrelated) cleanup.

pom.xml
stoty: We could. 
However, I plan to have at least one more variant that that can coexist  hbase-shaded-client (PHOENIX-6053), and it's easier to keep the structure than to remove and re-add it later.
If you feel strongly about it, then we certainly can remove this.

stoty:
> Doesn't seem relevant to this PR but I see lots of `refCount leaked` errors in `DateTimeIT`

Yes, we have a lot of those lately (we get that in our upstream tests).
I haven't been able to look into that, but it would be great to solve those sooner than later.

pom.xml
stoty: Oh, you mean the directory.
Sure.

stoty: Done.

PR #1432, opened 2022-05-03
PHOENIX-6703 Exclude Jetty and servlet-api from phoenix-client

stoty:


phoenix-client-parent/phoenix-client-embedded/pom.xml
```diff
@@ -94,5 +94,42 @@
...
+      <groupId>org.eclipse.jetty</groupId>
+      <artifactId>jetty-webapp</artifactId>
+      <scope>provided</scope>
+      <version>${jetty.version}</version>
+    </dependency>
+    <dependency>
+      <groupId>javax.servlet</groupId>
+      <artifactId>javax.servlet-api</artifactId>
+      <scope>provided</scope>
+      <version>3.1.0</version>
```
gjacoby126, line 132: nit: shouldn't this be java.servlet-api.version?

gjacoby126 (changes requested):

phoenix-client-parent/phoenix-client-embedded/pom.xml
stoty: You're right, fixed.

gjacoby126 (approved):
+1 thanks @stoty

PR #1434, opened 2022-05-06
PHOENIX-6723 Update client to use Zookeeper 3.5.7 and Curator 4.2.0

wendigo:
Previous Zookeeper version was broken on JDK 14+ due to the incorrect usage of
InetSocketAddress.toString. For reference see: https://bugs.java.com/bugdatabase/view_bug.do?bug_id=8232369

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -121,8 +121,8 @@
     <stream.version>2.9.5</stream.version>
     <i18n-util.version>1.0.4</i18n-util.version>
     <guice.version>4.0</guice.version>
-    <zookeeper.version>3.4.14</zookeeper.version>
-    <curator.version>4.0.0</curator.version>
+    <zookeeper.version>3.5.7</zookeeper.version>
+    <curator.version>4.2.0</curator.version>
     <jcodings.version>1.0.55</jcodings.version>
     <thrift.version>0.9.3-1</thrift.version>
     <!-- Test Dependencies -->
@@ -1532,7 +1532,6 @@
         <hbase.compat.version>2.4.1</hbase.compat.version>
         <hbase.version>${hbase-2.4.runtime.version}</hbase.version>
         <hbase.thirdparty.version>3.4.1</hbase.thirdparty.version>
-        <zookeeper.version>3.5.7</zookeeper.version>
         <tephra.hbase.compat.version>2.4</tephra.hbase.compat.version>
       </properties>
     </profile>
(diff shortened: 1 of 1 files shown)


wendigo:
cc @stoty 

This is relevant for the https://github.com/trinodb/trino project as we want to move to JDK17 (which is the LTS)

vincentpoon (approved):
+1 assuming tests pass

stoty:
Please open a JIRA, and add JIRA-XXXXX to the commit description, @wendigo.

stoty:
The ZK version doesn't do anything for ZK on master, we're already using 3.5.7 for both 2.3 and 2.4. (set in the hbase profiles)

Please remove the ZK version overrides from the hbase profiles in the pom, as they are not needed with this change.

As for Curator, HBase uses 4.2, not 4.3 on branch 2.3, 2.4, and 2.5. Is there a particular reason for using 4.3, or did you just pick the latest 4.3 release ?

In case you also want this merged into 5.1, we'd also have to check if HBase 2.1 and 2.2 works with ZK 3.5.7 (IIRC it does)

pom.xml
```diff
@@ -121,8 +121,8 @@
     <stream.version>2.9.5</stream.version>
     <i18n-util.version>1.0.4</i18n-util.version>
     <guice.version>4.0</guice.version>
-    <zookeeper.version>3.4.14</zookeeper.version>
-    <curator.version>4.0.0</curator.version>
+    <zookeeper.version>3.5.7</zookeeper.version>
```
stoty, line 124: Please remove the corresponding overrides from the profiles.

pom.xml
```diff
@@ -121,8 +121,8 @@
...
+    <curator.version>4.3.0</curator.version>
```
stoty, line 125: HBase uses 4.2.
While it's probably OK to use a newer version, if there is no particular reason to use 4.3, it's better to stick with the version HBase uses.

stoty (changes requested):

gjacoby126:
@vincentpoon @wendigo - I think this PR is currently waiting for changes that have been requested by @stoty and then I think would be ready for commit. Do you have bandwidth to get back to this in time for the Phoenix 5.2 release (probably in the next month or so?)

wendigo:
@gjacoby126 I will update this PR soon

wendigo:
@stoty @gjacoby126 PR updated. I tested it with patched Trino version, and now it works correctly on JDK 17 with HBase 2.3.0

gjacoby126:
@wendigo - the changes to the pom look good, but could you please change the commit message to refer to a JIRA number. (And if there isn't a JIRA for this yet, could you please create on on issues.apache.org.) Thanks!

wendigo:
@gjacoby126 yes, will do, I need to create jira account first.

wendigo:
@gjacoby126 done, JIRA handle is the same as in GitHub (@wendigo). I've updated the commit message too.

wendigo:
https://issues.apache.org/jira/browse/PHOENIX-6723

gjacoby126 (approved):
+1, thanks @wendigo . @stoty , did you want any other changes to this patch before we commit it?

stoty (approved):
+1 LGTM.
Thank you.

PR #1440, opened 2022-05-11
PHOENIX-6687 The region server hosting the SYSTEM.CATALOG fails to serve any metadata requests as default handler pool threads are exhausted

jpisaac:


phoenix-core/src/main/java/org/apache/hadoop/hbase/ipc/controller/ServerSideRPCControllerFactory.java
```diff
@@ -0,0 +1,41 @@
...
+package org.apache.hadoop.hbase.ipc.controller;
+
+import org.apache.hadoop.conf.Configuration;
+import org.apache.hadoop.hbase.ipc.RpcControllerFactory;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * {@link RpcControllerFactory} that should only be used when
+ * making server-server remote RPCs to the region servers hosting Phoenix SYSTEM tables.
```
gjacoby126, line 27: @jpisaac - if this is only for server-to-server RPCs to system tables, how is this different from the existing metadata pool? I thought that's what that was already for?

phoenix-core/src/main/java/org/apache/phoenix/query/QueryServicesOptions.java
```diff
@@ -223,11 +223,13 @@ public class QueryServicesOptions {
      * HConstants#HIGH_QOS is the max we will see to a standard table. We go higher to differentiate
      * and give some room for things in the middle
      */
+    public static final int DEFAULT_SERVER_SIDE_PRIORITY = 500;
```
gjacoby126, line 226: @jpisaac - I'm curious how you determined this priority (much less than metadata, which is very similar, and less than indexes)

phoenix-core/src/main/java/org/apache/phoenix/query/QueryServicesOptions.java
```diff
@@ -223,11 +223,13 @@ public class QueryServicesOptions {
...
     public static final int DEFAULT_INDEX_PRIORITY = 1000;
     public static final int DEFAULT_METADATA_PRIORITY = 2000;
     public static final boolean DEFAULT_ALLOW_LOCAL_INDEX = true;
     public static final int DEFAULT_INDEX_HANDLER_COUNT = 30;
     public static final int DEFAULT_METADATA_HANDLER_COUNT = 30;
+    public static final int DEFAULT_SERVERSIDE_HANDLER_COUNT = 30;
```
gjacoby126, line 232: is this the right default? (Are the current defaults the right defaults?) Will we expect a cluster to have approximately the same load of RS->index, metadata, and RS->Syscat/Link calls?

gjacoby126 (commented):
Some questions about the design

phoenix-core/src/main/java/org/apache/hadoop/hbase/ipc/controller/ServerSideRPCControllerFactory.java
jpisaac: @gjacoby126 The MetadataRpcController implements the org.apache.hbase.thirdparty.com.google.protobuf.RpcController which is not compatible with ServerRpcController which implements com.google.protobuf.RpcController. Hence had to use a separate executor for server-server RPC to Metadata tables.

phoenix-core/src/main/java/org/apache/phoenix/query/QueryServicesOptions.java
jpisaac: The priority is only used to dispatch to the right RPCExecutor. In this case, since the index, metadata and server-server use separate Executors not sure whether the actual value matters?

phoenix-core/src/main/java/org/apache/phoenix/query/ConnectionQueryServicesImpl.java
```diff
@@ -1788,6 +1799,50 @@ private MetaDataMutationResult childLinkMetaDataCoprocessorExec(byte[] parentTab
...
     }
 
+    @VisibleForTesting
+    protected RpcController getController() {
+        return getController(SYSTEM_CATALOG_HBASE_TABLE_NAME);
+    }
+
+        /**
+         * If configured to use the metadata handler pool for server side connections,
+         * use the {@link org.apache.hadoop.hbase.ipc.controller.MetadataRpcController}
```
gjacoby126, line 1809: This comment appears to be incorrect (it says it's using the MetadataRpcController)

phoenix-core/src/main/java/org/apache/hadoop/hbase/ipc/controller/ServerSideRPCControllerFactory.java
gjacoby126: OK, so let me make sure I understnad properly. 

Direct operations on a system table (UPSERTs or queries) would use the MetadataRpcController and coproc calls (such as CREATE and ALTER operations where MetadataClient calls MetadataEndpointImpl operations) would use ServerSideRPCController? And there's no way to consolidate the two in either direction because of the mismatch between shaded and unshaded protobufs disallows casting?

jpisaac:
@gjacoby126 
Yes, you are right in that the getController method cannot be refactored to use a common interface since MetadataRpcController and ServerRpcController use shaded and unshaded interfaces respectively, and casting leads to compile time errors.

As I understand puts and queries on the SYSTEM CATALOG table use the default handler pool. MetadataRpcController is used only for the MetdataService methods.

gjacoby126 (approved):
+1

gjacoby126:
@jpisaac - I added in two missing Apache license headers. Since the code is owned by our mutual employer and being contributed under a corporate CLA, I believe I have authority to do this. However, to remove all doubt, could you please reply to this and confirm that you want to Apache license all the source files in this PR. :-)

PR #1442, opened 2022-05-16
PHOENIX-6707 Use Configuration.getLongBytes where applicable

anmolnar:
I tried to find all the applicable places for the replacement. Please verify.


stoty:
Most preoperties are defined in QueryServicesOptions , I can see a bunch of further candidates there.

anmolnar:
@stoty I'm not sure what do you mean by "in QueryServices". That's an interface which doesn't have logic.
In most places I see something like the following:

```
                    connection
                            .getQueryServices()
                            .getProps()
                            .getLong(QueryServices.SEQUENCE_CACHE_SIZE_ATTRIB,
                                QueryServicesOptions.DEFAULT_SEQUENCE_CACHE_SIZE);
```

getProps() gives you `ReadOnlyProps` which doesn't have getLongBytes alternative.

anmolnar:
@stoty Thanks for the pointer. I've found a couple of more.

stoty:
Come to think of it, implementing getLongBytes on ReadOnlyProps would probably be a good idea.

anmolnar:
@stoty I've implemented getLongBytes() for `ReadOnlyProps` and rebased the patch. PTAL.

stoty:
Mostly looks good.
Can you add a small UT for getLongBytes()  in ReadOnlyProps to improve test coverage ?

stoty (approved):
+1 LGTM

PR #1443, opened 2022-05-17
PHOENIX-6715 Update Omid to 1.1.0

stoty:


virajjasani:
@stoty you have this tested against local m2 cache of omid 1.1.0?

stoty:
Yes, I have, at the time.
But there were changes in Omid since then, and we're gonna have to test again, once the Omid HBase2 API migration has landed in Omid.

pom.xml
```diff
@@ -121,7 +121,7 @@
     <collections.version>3.2.2</collections.version>
     <jodatime.version>2.10.5</jodatime.version>
     <joni.version>2.1.31</joni.version>
-    <omid.version>1.0.2</omid.version>
+    <omid.version>1.1.0-SNAPSHOT</omid.version>
```
richardantal, line 124: I think this version should be updated to 1.1.0 after we have the omid released.

stoty: That's correct, Richard.

stoty:
The failing tests are known flakeys.

virajjasani:
> The failing tests are known flakeys.

Correct, I just saw that the latest build test failures are limited to that build only. The previous build had the passing.

virajjasani (approved):
+1

PR #1446, opened 2022-05-27
PHOENIX-6689 Support multiple-characters delimiter for CsvBulkLoadTool

chaijunjie0101:


stoty:
You should introduce a new Configuration key for the string value, and fall back to the character value if that is not present, for backwards compatibility.

chaijunjie0101:
>


> You should introduce a new Configuration key for the string value, and fall back to the character value if that is not present, for backwards compatibility.

ok，i will change this MR for backwards compatibility, it is better, thank you for review and suggest.

chaijunjie0101:
the licenses warning is not related to this MR
2 Unknown Licenses

*****************************************************

Files with unapproved licenses:

  dev/create-release/cache-apache-project-artifact.sh
  dev/create-release/rebuild_hbase.sh

*****************************************************

chaijunjie0101:
Add a conf in the CSVBulkLoadTool, we can use "multiple-delimiter"("md") to set delimiter in shell command to support it.

The unit tests failed seems also not releate to this MR, the CsvBulkLoadToolIT test is OK, pls review this MR, thnak you.
image

PR #1448, opened 2022-05-27
PHOENIX-6719 - Duplicate Salt Columns in Schema Registry after ALTER

gjacoby126:


gjacoby126:
Note: the test that detected this is in a draft of phoenix-connectors' PHOENIX-5517 / PHOENIX-5448 (CDC message producer and Avro serialization). The test will be contributed when those JIRAs are contributed.

gjacoby126:
Test failures are known flappers -- a bunch of refCount leaked errors and 

```
[ERROR]   ConcurrentMutationsExtendedIT.testConcurrentUpserts:329->verifyIndexTable:140 Expected equality for V1, but 16!=-16
[ERROR]   PartialIndexRebuilderIT.testConcurrentUpsertsWithRebuild:240 Expected equality for V1, but 0!=-10
```

They don't seem relevant to the patch

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/MetaDataEndpointImpl.java
```diff
@@ -1532,23 +1538,26 @@ private PTable getTableFromCells(List<Cell> tableCellList, List<List<Cell>> allC
                 long columnTimestamp =
                     columnCellList.get(0).getTimestamp() != HConstants.LATEST_TIMESTAMP ?
                         columnCellList.get(0).getTimestamp() : timeStamp;
+                boolean isSalted = saltBucketNum != null
```
gjacoby126, line 1541: If we're doing an ALTER we need to check both the original version of the table AND the Cells from the ALTER to see if the table's salted

phoenix-core/src/it/java/org/apache/phoenix/end2end/AlterTableIT.java
```diff
@@ -325,8 +328,7 @@ public static void verifySchemaExport(PTable newTable, Configuration conf) throw
         //needed in an external schema registry. TODO: fix the base column count anyway
         String baseColumnCountPattern = "(?i)\\s*baseColumnCount:\\s\".*\"";
         expectedSchemaText = expectedSchemaText.replaceAll(baseColumnCountPattern, "");
-        actualSchemaText = expectedSchemaText.replaceAll(baseColumnCountPattern, "");
-
+        actualSchemaText = actualSchemaText.replaceAll(baseColumnCountPattern, "");
```
gjacoby126, line 331: This was a bug in the test -- it was checking the wrong schema, and so missed that the physical table names (and salted behavior, after I added the SALT_BUCKETS above) were wrong

gjacoby126:
After the latest test run, only the known flappers are failing

PR #1455, opened 2022-06-20
PHOENIX-6627 remove tephra references - WIP

zsgyulavari:


gjacoby126:
Thanks for taking this up, @zsgyulavari ! Please let me know if there's anything I can do to help -- it would be great if we could get this into 5.2 which is coming soon. 

On the first pass it looks like there are 151 test failures, but most of them seem to be from the same problem in TransactionFactory$Provider's fromCode method. On line 69 it's making an assumption that the length of the Provider enum is equal to the largest Provider enum value (that is, there are 2 providers, and the highest provider is enum value 2.) That's not true anymore now that the Tephra provider is gone, so the logic has to change.

gjacoby126:
Also, what will happen to user tables who upgrade from Phoenix 5.1 -> 5.2 and have the Provider set to Tephra? Would that cause the region to crash from trying to load an enum value that doesn't exist? If so, what sort of upgrade logic should we include? I can think of three options:
 1. Switch tables to non-transactional 
 2. Switch tables to Omid 
 3. Keep the Tephra enum and just WARN or ERROR on region load that it doesn't do anything 

1 and 2 are both tricky to coordinate on a rolling restart, because the normal upgrade logic runs when SYSTEM.CATALOG is upgraded, but the crashing would probably come before that, when a region hosting a TEPHRA-enabled table has its region server upgraded. But I worry that some operators would miss the warning in their logs. 
 
@stoty @jpisaac @kadirozde @virajjasani , curious on your opinions.

phoenix-core/src/main/java/org/apache/phoenix/transaction/TransactionFactory.java
```diff
@@ -22,43 +22,19 @@
...
-        } else {
-            tephraTransactionProvider = NotAvailableTransactionProvider.getInstance();
-        }
-    }
-
     public enum Provider {
-        TEPHRA((byte)1, tephraTransactionProvider,
-            tephraTransactionProvider instanceof TephraTransactionProvider),
-        OMID((byte)2, OmidTransactionProvider.getInstance(), true);
+        OMID((byte)2, OmidTransactionProvider.getInstance());
```
stoty, line 26: This changes the ordinal, which SHOULD not be a real problem as we don't use that for serialization, but keeping TEPHRA, and just renaming it to something like TEPHRA_REMOVED would possibly also make it easier to handle the hairy cases where tephra defined for old tables.

phoenix-core/src/main/java/org/apache/phoenix/transaction/TransactionFactory.java
```diff
@@ -94,7 +66,7 @@ public static PhoenixTransactionProvider getTransactionProvider(byte[] txState,
             return null;
         }
         Provider provider = (clientVersion < MetaDataProtocol.MIN_SYSTEM_TABLE_TIMESTAMP_4_14_0) 
-                ? Provider.TEPHRA
+                ? Provider.OMID
```
stoty, line 69: Again, should we just error out ?

phoenix-core/src/test/java/org/apache/phoenix/util/TestUtil.java
```diff
@@ -1268,27 +1268,8 @@ public static void assertResultSet(ResultSet rs, Object[][] rows) throws Excepti
...
-        List<Object[]> filteredData = Lists.newArrayListWithExpectedSize(data.size());
-        for (Object[] params : data) {
-            String provider = (String) params[index];
-            if (provider == null || TransactionFactory.Provider.valueOf(provider).runTests()) {
-                filteredData.add(params);
-            }
-        }
-        return filteredData;
+    	return data;
+    	//TODO: remove this method altogether
```
stoty, line 1272: Yes, it shouldn't be too difficult.

phoenix-core/src/main/java/org/apache/phoenix/transaction/NotAvailableTransactionProvider.java
```diff
@@ -60,7 +60,7 @@ public PhoenixTransactionClient getTransactionClient(Configuration config, Conne
 
     @Override
     public Provider getProvider() {
-        return TransactionFactory.Provider.TEPHRA;
+        return TransactionFactory.Provider.OMID;
```
stoty, line 63: This is not right.
We should either return null, or the UnavailableProvider's type, if we keep that,

stoty (changes requested):

phoenix-core/src/it/java/org/apache/phoenix/end2end/index/ImmutableIndexIT.java
```diff
@@ -137,9 +137,7 @@ public static synchronized Collection<Object[]> data() {
 		        Arrays.asList(new Object[][] {
     				{ false, false, null, false }, { false, false, null, true },
     				{ false, true, "OMID", false },
-                    { false, true, "TEPHRA", false }, { false, true, "TEPHRA", true },
     				{ true, false, null, false }, { true, false, null, true },
```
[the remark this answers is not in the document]
stoty: Correction:
Those seem to be unsupported operations for Omid, so this is fine.

apurtell:
I'm going to close this PR but we can reopen it if the alternative PR I am about to open is not acceptable for some reason.

PR #1456, opened 2022-06-20
PHOENIX-5274 - ConnectionQueryServiceImpl should use HBase Namespace …

gjacoby126:
…APIs that do not require ADMIN permissions for existence checks

### DIFF
--- a/phoenix-core/src/main/java/org/apache/phoenix/util/ServerUtil.java
+++ b/phoenix-core/src/main/java/org/apache/phoenix/util/ServerUtil.java
@@ -38,6 +38,7 @@ import org.apache.hadoop.hbase.DoNotRetryIOException;
 import org.apache.hadoop.hbase.HConstants;
 import org.apache.hadoop.hbase.NotServingRegionException;
 import org.apache.hadoop.hbase.TableName;
+import org.apache.hadoop.hbase.client.Admin;
 import org.apache.hadoop.hbase.client.Connection;
 import org.apache.hadoop.hbase.client.RetriesExhaustedWithDetailsException;
 import org.apache.hadoop.hbase.client.Table;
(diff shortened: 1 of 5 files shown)


phoenix-core/src/main/java/org/apache/phoenix/util/ServerUtil.java
```diff
@@ -417,4 +418,27 @@ public static <T> Throwable getExceptionFromFailedFuture(Future<T> f) {
...
+     * @param admin HbaseAdmin Object
+     * @param schemaName Phoenix schema name for which we check existence of the HBase namespace
+     * @return true if the HBase namespace exists, else returns false
+     * @throws SQLException If there is an exception checking the HBase namespace
+     */
+    public static boolean isHBaseNamespaceAvailable(Admin admin, String schemaName) throws IOException{
+        boolean namespaceExists = false;
+        try{
+            String[] hbaseNamespaces = admin.listNamespaces();
+            for(String namespace : hbaseNamespaces){
```
dbwong, line 433: nit: can we replace this with a utility method?

dbwong (approved):
Minor nit otherwise lgtm, address nit and feel free to push

phoenix-core/src/main/java/org/apache/phoenix/util/ServerUtil.java
gjacoby126: Done

gjacoby126:
The PermissionNSEnabledIT failure seems genuine -- it passes on master branch and fails locally for me on the feature branch. I'll give it a closer look before committing

gjacoby126:
Looks like the existing 5.x code wasn't calling the preCreateSchema hook properly, so the permission validation check wasn't happening. This behavior was hidden because the restrictive HBase API the code used to rely effectively ran an even more restrictive check. Removing the HBase check revealed the existing security hole. I've fixed the createSchema logic so that it matches the dropSchema logic and the test passes now.

PR #1462, opened 2022-07-19
PHOENIX-6746 Test suite executions do not provide usable logs

stoty:
rename log4j2 config fils to the correct names
use log4j for tests for every HBase profile


phoenix-core/pom.xml
```diff
@@ -568,6 +568,26 @@
...
 
     <!-- Other test dependencies -->
+    <dependency>
+        <groupId>org.apache.logging.log4j</groupId>
+        <artifactId>log4j-api</artifactId>
+        <scope>test</scope>
+    </dependency>
+    <dependency>
+        <groupId>org.apache.logging.log4j</groupId>
+        <artifactId>log4j-core</artifactId>
```
gjacoby126, line 578: By including log4j-core as a test dependency, do we create a problem if someone tries to use phoenix-core as a library in an application that wants to use logback / another slf4j implementation and then writes a test using Phoenix?

stoty: No, because test dependencies are not transitive in maven.

stoty: The real change here (not in this specific fix, but in the original log4j2 commit) is that previously, phoenix-core transitively pulled in log4j1 via hadoop-common (compile scope), but we're excluding that now, so for 5.2 the user will need to add some kind of logging backend explicitly if he depends on phoenix-core.

stoty: Also, currently in 5.1 we replace log4j with reload4j, but only in the assembly, and the phoenix-client JAR.
If one depends directly on phoenix-core, it will still pull log4j transitively via hadoop.
(Do you think we should replace the )

gjacoby126 (approved):
+1

PR #1465, opened 2022-07-20
PHOENIX-6751 Force using range scan vs skip scan when using large IN clause

jpisaac:


gjacoby126 (approved):
+1

gjacoby126:
Thanks for the patch and the very thorough tests, @jpisaac

PR #1467, opened 2022-07-20
PHOENIX-6752 Duplicate expression nodes in extract nodes during WHERE compilation phase leads to poor performance

jpisaac:

PR #1469, opened 2022-07-21
PHOENIX-6752 Duplicate expression nodes in extract nodes during WHERE…

gjacoby126:
… compilation phase leads to poor performance


gjacoby126:
This is a rebase of #1467 since merging #1465 caused GitHub to incorrectly think there was a merge conflict there. (This is still @jpisaac 's code, I'm just doing it since he's out of office.) @dbwong and @tkhurana , you know the query parsing code better than me, so could you please take a look?

gjacoby126:
This PR has caused InListIT to crash twice, and InListIT is a test that it modifies. More investigation is needed before it can be merged. @jpisaac , @dbwong , @tkhurana , fyi.

gjacoby126:
@jpisaac - the new test in InListIT was causing OOMs during the iteration where it ran with 50,000 ORs (both in CI and locally). I tweaked it so it runs with at most 5000 ORs, and now it passes for me locally. If 50,000 ORs needs to be supported, either the test heap size needs to be bigger, or we need further optimization.

gjacoby126:
I'll let @jpisaac re-base his original PR now that he's back and close this one.

PR #1474, opened 2022-07-26
PHOENIX-6754 Upgrades from pre 4.10 versions are broken again

richardantal:


stoty (changes requested):
This looks like only haf of the fix.
You also need to restore the code that uses the ALTER_SYSCATALOG_TABLE_UPGRADE statement for upgrades.

stoty:
It has been removed from the addColumnMutation() setup, and the addColumnMutation() method.

richardantal:
Thank stoty, Let me update the PR

stoty:
As we have no IT for it, the real test is to manully take a dump of a skeleton Phonix 4.7 install, load it, and test if the upgrade is successful.

stoty:
I like that you put the logic inside the addColumnMutationInternal() method.
LGTM, but let's add the test that we've discussed offline.

stoty (approved):
As discussed, the test is going to be added in a different commit.

richardantal:
Thank you for the review.

PR #1475, opened 2022-07-29
PHOENIX-6756 switch to using log4j2.properties instead of xml

stoty:
also fix serveral problems introduced in the original log4j2 patch
change log format to match HBase


gjacoby126 (approved):
+1

PR #1480, opened 2022-08-08
Phoenix 6653 Add upgrade tests based on HBase snapshots

richardantal:
Work is still in progress


richardantal:
There is a separate PR for pre 4.10 upgrade https://github.com/apache/phoenix/pull/1474 
But I'd like to trigger a test run here with these 2. I will rebase Phoenix 6653 to the master.

stoty (changes requested):
You should either duplicate the template class, and put the test there, or rename it.
Also, I think that instead of committing the snapshot directories directly, storing them as a single zip file, or in a zip file per snapshot, and extracting in the test would be better. 
The individial files are not human reabable anyway.

stoty:
(We've had offline discussions about this with Richard, so some context may not be immediately apparent from the patch)

You have uncovered two different additional problems here.

One is that moveChildLinkConnection doesn't handle namespace mapping.
Your current solution here is incorrect, because it manipulates the pre-upgrade default namespace system tables, even though by this point it has already been moved to the NS (and upgraded).
The correct solution would be to replace both  _TableName.valueOf(SYSTEM_CATALOG_NAME)_ and TableName.valueOf(SYSTEM_CHILD_LINK_NAME) with the right namespace aware method from SchmaUtil which preforms the '.' -> ':' replacement as needed.  (it is also much simpler)

This bugfix may even be split into a separate dependent ticket, if you prefer.
We should also check if we have other similar bugs in the upgrade code.

The other problem is that moveChildLinkConnection creates a the HBase Connection from a vanilla Configuration object (reading from hbase-site.xml), which doesn't get necessary Config changes that were set by BaseTest/minicluster.

Simply copying the ZK quorum from the CQSI connection may get the test to pass, but is not the right way.
We should clone the Configuration object from CQSI,  apply the necessary  the changes to the clone, and use that for creating the HBaseConnection.

phoenix-core/src/main/java/org/apache/phoenix/util/UpgradeUtil.java
```diff
@@ -376,10 +377,14 @@ private static void preSplitSequenceTable(PhoenixConnection conn, int nSaltBucke
      * @return
      * @throws IOException
      */
-    private static org.apache.hadoop.hbase.client.Connection getHBaseConnection(Map<String, String> options)
+    private static org.apache.hadoop.hbase.client.Connection getHBaseConnection(Configuration config, Map<String, String> options)
             throws IOException {
 
         Configuration conf = HBaseConfiguration.create();
+
+        Configuration oldconf = HBaseConfiguration.create(config);
```
stoty, line 385: Simply use 
`Configuration conf = HBaseConfiguration.create(config);`
and you'll automatically get all properties, including ZK.

phoenix-core/src/main/java/org/apache/phoenix/util/UpgradeUtil.java
```diff
@@ -1338,9 +1347,30 @@ public static void moveOrCopyChildLinks(PhoenixConnection oldMetaConnection, Map
...
         }
+    }
+
+    /**
+     * Wrapper around moveOrCopyChildLinksInner, we try move or copy child links from SYSTEM.CATALOG and SYSTEM:CATALOG
+     * @param oldMetaConnection caller should take care of closing the passed connection appropriately
+     * @throws SQLException
+     */
+
+    public static void moveOrCopyChildLinks(PhoenixConnection oldMetaConnection, Map<String, String> options) throws IOException {
```
stoty, line 1361: This is the wrong workaround.

stoty (changes requested):

phoenix-core/src/main/java/org/apache/phoenix/util/UpgradeUtil.java
```diff
@@ -1244,8 +1244,18 @@ public static void moveOrCopyChildLinks(PhoenixConnection oldMetaConnection, Map
...
 
-        try (org.apache.hadoop.hbase.client.Connection moveChildLinkConnection =  getHBaseConnection(options);
-            Table sysCatalogTable = moveChildLinkConnection.getTable(TableName.valueOf(SYSTEM_CATALOG_NAME))) {
+        Configuration conf = oldMetaConnection.getQueryServices().getConfiguration();
+
+        ReadOnlyProps readOnlyProps = oldMetaConnection.getQueryServices().getProps();
+        TableName sysCat = SchemaUtil.getPhysicalTableName(SYSTEM_CATALOG_NAME, readOnlyProps);
+        TableName sysChildLink = SchemaUtil.getPhysicalTableName(SYSTEM_CHILD_LINK_NAME, readOnlyProps);
+
+        LOGGER.info(String.format("SYSTEM CATALOG tabled use for copying child links: %s", sysCat.toString()));
```
stoty, line 1253: Once the testing is done, you may want remove this, or change it to DEBUG level.

stoty (changes requested):
Please also mention the addParentToChildLinks namespace fix in the commit message.

phoenix-core/src/it/java/org/apache/phoenix/end2end/LoadSystemTableSnapshotIT.java
```diff
@@ -112,7 +161,16 @@ private static void importSnapshot(String key, String value) throws IOException
...
+        Map<String, String> serverProps = Maps.newHashMapWithExpectedSize(2);
+        serverProps.put(QueryServices.EXTRA_JDBC_ARGUMENTS_ATTRIB, QueryServicesOptions.DEFAULT_EXTRA_JDBC_ARGUMENTS);
+        serverProps.put(QueryServices.IS_NAMESPACE_MAPPING_ENABLED, "true");
+        Map<String, String> clientProps = Maps.newHashMapWithExpectedSize(2);
+        clientProps.put(QueryServices.IS_NAMESPACE_MAPPING_ENABLED, "true");
+
+        //Now we can start Phoenix
+        setUpTestDriver(new ReadOnlyProps(serverProps.entrySet().iterator()), new ReadOnlyProps(clientProps.entrySet()
+                .iterator()));
         assertTrue(true);
```
stoty, line 174: Is this assertion enough ?
Will the driver throw an exception if there is an upgrade problem ?

richardantal: I think, yes
If you see the tests results in my [other PR](https://github.com/apache/phoenix/pull/1485), without the PHOENIX-6754 fix, that fails with 
`org.apache.phoenix.schema.ColumnNotFoundException: ERROR 504 (42703): Undefined column. columnName=SYSTEM.CATALOG.COLUMN_QUALIFIER`
Test run on this one is successful.

Will update this and the other PR, based on the things you suggested but it would be better to merge PHOENIX-6754 before

stoty: That's fine, ping me when the PHOENIX-6754 are finished.

richardantal: PHOENIX-6754 is finished, but we wanted to run these tests on the change
https://github.com/apache/phoenix/pull/1474

richardantal:
Thank you @stoty for the review and the suggestions, I've updated the log msgs, and added more detail to the commit message.

phoenix-core/pom.xml
```diff
@@ -517,6 +517,10 @@
       <groupId>org.apache.commons</groupId>
       <artifactId>commons-csv</artifactId>
     </dependency>
+    <dependency>
```
stoty, line 520: This should be test scope.

stoty (approved):
+1 LGTM 
but wait for the tests.

richardantal:
Thank you @stoty for the reviews,
I close this PR and merge the other 2 separately.

PR #1486, opened 2022-08-11
PHOENIX-6627 Remove all references to Tephra from 4.x and master

apurtell:
Retire Tephra.

Considerations:

Ordinals of `TransactionFactory.Provider` are maintained to preserve compatibility with deployed system catalogs. The enum label for the former Tephra provider is deliberately renamed to `NOTAVAILABLE` so any downstreams with ill-advised direct dependencies on it will fail to compile.

A coprocessor named `TephraTransactionalProcessor` is retained, with a no-op implementation, to prevent regionserver aborts in the worst case that a Tephra table remains deployed and the table coprocessor list has not changed. This should not happen. Users should be provided a migration runbook.

The `commitDDLFence` phase of `MutationState` is retained although it may no longer be necessary. Recommend a follow up issue.

`PhoenixTransactionContext.PROPERTY_TTL` is retained. It is possible a future txn engine option will have a similar design feature and will need to alter cell timestamps. Recommend a follow up issue.

`storeNulls` for transactional tables remains enabled for compatibility.  It is possible a future txn engine option will have a similar design feature and will need to manage tombstones in a non-default way. Recommend a follow up issue.

Requests to ALTER a table from transactional to non transactional are still rejected. It is very likely OMID has similar complex state cleanup and removal requirements, and txn engines in general are likely to have this class of problem.


kadirozde:
The PR looks good to me. Can we trigger the pre-checkin tests for this PR? Not sure why it was not triggered automatically yet.

virajjasani:
As of this comment, it looks like the first build run is still in progress, not completed yet: https://ci-hadoop.apache.org/job/Phoenix/job/Phoenix-PreCommit-GitHub-PR/job/PR-1486/

apurtell:
I did not submit until the UTs and ITs in phoenix-core all passed for me. I know that is not good enough on its own, but I submit it as a data point.

phoenix-core/src/it/java/org/apache/phoenix/end2end/NoLookbackMutableIndexExtendedIT.java
```diff
@@ -93,11 +93,11 @@ protected static Connection getConnection() throws SQLException {
...
     public static Collection<Object[]> data() {
-        return TestUtil.filterTxParamData(Arrays.asList(
-                new Object[][] { { false, null, false }, { false, null, true },
-                        { false, "TEPHRA", false }, { false, "TEPHRA", true },
-                        { false, "OMID", false }, { true, null, false }, { true, null, true },
-                        { true, "TEPHRA", false }, { true, "TEPHRA", true }, }), 1);
+        return Arrays.asList(
+            // OMID does not support local indexes
+            new Object[][] { { false, null, false }, { false, null, true },
+                { false, "OMID", false }, { false, "OMID", true },
```
stoty, line 99: OMID doesn't support column encoded tables, { false, "OMID", true } is not a valid combination, as you've already noted elsewhere.

phoenix-core/src/it/java/org/apache/phoenix/end2end/index/MutableIndexFailureIT.java
```diff
@@ -175,27 +175,25 @@ protected static Map<String,String> getServerProps(){
...
-                    { null, false, false, false, true, false},
-                    }), 0);
-        
+        return Arrays.asList(new Object[][] {
+            // Note: Can't disableIndexOnWriteFailure without throwIndexWriteFailure, PHOENIX-4130
+            { null, false, false, false, false, false},
+            { null, false, false, true, false, null},
+            // Note: OMID does not support local indexes
+            { "OMID", false, false, true, false, null},
+            { null, true, false, null, false, null},
```
stoty, line 187: We could a namespace mapped OMID combination here.
(Though I'm not sure if this test makes a lot of sense anymore with the new indexing.)

phoenix-core/src/main/java/org/apache/phoenix/transaction/NotAvailableTransactionProvider.java
```diff
@@ -29,7 +29,7 @@
 public class NotAvailableTransactionProvider implements PhoenixTransactionProvider {
     private static final NotAvailableTransactionProvider INSTANCE = new NotAvailableTransactionProvider();
 
-    private static final String message="This Phoenix has been built without Tephra support";
+    private static final String message = "This Phoenix has been built without Tephra support";
```
stoty, line 32: Nit: Change the string ?

stoty (approved):
+1 LGTM

I've noted some nits.

phoenix-core/src/it/java/org/apache/phoenix/end2end/index/MutableIndexFailureIT.java
gjacoby126: We still technically support the old indexes I think (and probably have to so long as we support upgrading from pre-4.15 versions, since you can't upgrade your indexes until you get that far.) so it makes sense to have tests for the failure-disable-rebuild paths.

gjacoby126 (approved):
+1

gjacoby126:
@stoty  - @apurtell is going to be unavailable for a couple of weeks, so I went ahead and fixed the NoLookbackMutableIndexExtendedIT and NoTransactionAvailableProvider nits.

gjacoby126:
I can't reproduce the test failure in TenantViewOperationsWorkloadIT either in the master branch or Andrew's feature branch. I've kicked off a new CI build to see if it recurs. If not I'll go ahead and merge.

stoty:
That failure sounds like PHOENIX-6552, @gjacoby126 , which is a simple infra overload issus (though the test is also overly optimistic wrt GC pauses, etc.)

PR #1488, opened 2022-08-11
 phoenix.schema.isNamespaceMappingEnabled set true

gurpreet0909:
we have  phoenix.schema.isNamespaceMappingEnabled enable, while running Phoenix MR jar this is the problem we are getting and there is no way to solve it.

### DIFF
--- a/phoenix-core/src/main/java/org/apache/phoenix/mapreduce/PhoenixInputFormat.java
+++ b/phoenix-core/src/main/java/org/apache/phoenix/mapreduce/PhoenixInputFormat.java
@@ -169,6 +169,9 @@ public class PhoenixInputFormat<T extends DBWritable> extends InputFormat<NullWr
             final String currentScnValue = configuration.get(PhoenixConfigurationUtil.CURRENT_SCN_VALUE);
             final String tenantId = configuration.get(PhoenixConfigurationUtil.MAPREDUCE_TENANT_ID);
             final Properties overridingProps = new Properties();
+                        
+            overridingProps.setProperty("phoenix.schema.isNamespaceMappingEnabled","true");
+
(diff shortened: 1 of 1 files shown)


gjacoby126:
@gurpreet0909 - I believe this patch would break MapReduce jobs that do not have isNamespaceMappingEnabled set to true. (You can see, for example, that in the Jenkins test run that TableSnapshotReadsMapReduceIT has several failures for this reason.) 

Shouldn't we only set the MapReduce client to isNamespaceMappingEnabled = true if the cluster is in fact namespace enabled?

gurpreet0909:
> @gurpreet0909 - I believe this patch would break MapReduce jobs that do not have isNamespaceMappingEnabled set to true. (You can see, for example, that in the Jenkins test run that TableSnapshotReadsMapReduceIT has several failures for this reason.)
> 
> Shouldn't we only set the MapReduce client to isNamespaceMappingEnabled = true if the cluster is in fact namespace enabled?

Hi, i am using these jars in client side only in my Mapreduce jar which i am running on hdfs, but the problem is my MapReduce jar is going the hit this jar class for connection at the same place where i have added the property in the commit. For me there is not way to inject this property inside these jars,

i have tried to create custom jar but since its maven project it needs to download dependencies from repository for which i am not authenticated. 

How do i solve it then?

stoty:
Having matching hbase-site.xml s on the server side and client (MR) side is non negotiatieble.
You can either:
- Add the hbase-site.xml to the JAR that contains jour MR job
- Package hbase-site.xml to a separate JAR, and add that to you job
- Distribute the hbase-side the hbase-site.xml everywhere and the directory

stoty (changes requested):
-2

PR #1490, opened 2022-08-19
PHOENIX-6764 Implement Binary and Hexadecimal literals

stoty:
also use VarBinaryFormatter for BINARY type

### DIFF
--- a/phoenix-core/src/main/java/org/apache/phoenix/util/ByteUtil.java
+++ b/phoenix-core/src/main/java/org/apache/phoenix/util/ByteUtil.java
@@ -59,7 +59,7 @@ public class ByteUtil {
 
 
     /** Mask for bit 0 of a byte. */
-    private static final int BIT_0 = 1;
+    private static final int BIT_0 = 0x01;
 
     /** Mask for bit 1 of a byte. */
     private static final int BIT_1 = 0x02;
--- a/phoenix-core/src/main/java/org/apache/phoenix/util/ByteUtil.java
+++ b/phoenix-core/src/main/java/org/apache/phoenix/util/ByteUtil.java
@@ -680,7 +680,8 @@ public class ByteUtil {
         return dst;
     }
 
-    // Adapted from Commons-codec BinaryCodec but without reverting the bytes
+    // Adapted from the Commons Codec BinaryCodec, but treat the input as a byte sequence, without
+    // the endinanness reversion in the original code
     public static byte[] fromAscii(final char[] ascii) {
         if (ascii == null || ascii.length == 0) {
             return EMPTY_BYTE_ARRAY;
@@ -688,10 +689,8 @@ public class ByteUtil {
         final int asciiLength = ascii.length;
         // get length/8 times bytes with 3 bit shifts to the right of the length
         final byte[] l_raw = new byte[asciiLength >> 3];
-        /*
-         * We incr index jj by 8 as we go along to not recompute indices using multiplication every time inside the
-         * loop.
-         */
+        // We incr index jj by 8 as we go along to not recompute indices using multiplication every
+        // time inside the loop.
         for (int ii = 0, jj = 0; ii < l_raw.length; ii++, jj += 8) {
             for (int bits = 0; bits < BITS.length; ++bits) {
                 if (ascii[jj + bits] == '1') {
--- a/phoenix-core/src/main/java/org/apache/phoenix/util/ByteUtil.java
+++ b/phoenix-core/src/main/java/org/apache/phoenix/util/ByteUtil.java
@@ -56,7 +56,34 @@ public class ByteUtil {
             EMPTY_BYTE_ARRAY);
     public static final ImmutableBytesWritable EMPTY_IMMUTABLE_BYTE_ARRAY = new ImmutableBytesWritable(
             EMPTY_BYTE_ARRAY);
-    
+
+
+    /** Mask for bit 0 of a byte. */
+    private static final int BIT_0 = 1;
+
+    /** Mask for bit 1 of a byte. */
+    private static final int BIT_1 = 0x02;
+
+    /** Mask for bit 2 of a byte. */
+    private static final int BIT_2 = 0x04;
+
+    /** Mask for bit 3 of a byte. */
+    private static final int BIT_3 = 0x08;
+
+    /** Mask for bit 4 of a byte. */
+    private static final int BIT_4 = 0x10;
+
+    /** Mask for bit 5 of a byte. */
+    private static final int BIT_5 = 0x20;
+
+    /** Mask for bit 6 of a byte. */
+    private static final int BIT_6 = 0x40;
+
+    /** Mask for bit 7 of a byte. */
+    private static final int BIT_7 = 0x80;
+
+    private static final int[] BITS = {BIT_7, BIT_6, BIT_5, BIT_4, BIT_3, BIT_2, BIT_1, BIT_0};
+
     public static final Comparator<ImmutableBytesPtr> BYTES_PTR_COMPARATOR = new Comparator<ImmutableBytesPtr>() {
 
         @Override
@@ -651,5 +678,27 @@ public class ByteUtil {
                 previousSepartor +1, src.length - previousSepartor -1);
         }
         return dst;
+    }
+
+    // Adapted from Commons-codec BinaryCodec but without reverting the bytes
+    public static byte[] fromAscii(final char[] ascii) {
+        if (ascii == null || ascii.length == 0) {
+            return EMPTY_BYTE_ARRAY;
+        }
+        final int asciiLength = ascii.length;
+        // get length/8 times bytes with 3 bit shifts to the right of the length
+        final byte[] l_raw = new byte[asciiLength >> 3];
+        /*
+         * We incr index jj by 8 as we go along to not recompute indices using multiplication every time inside the
+         * loop.
+         */
+        for (int ii = 0, jj = 0; ii < l_raw.length; ii++, jj += 8) {
+            for (int bits = 0; bits < BITS.length; ++bits) {
+                if (ascii[jj + bits] == '1') {
+                    l_raw[ii] |= BITS[bits];
+                }
+            }
+        }
+        return l_raw;
     }
 }
--- a/phoenix-core/src/test/java/org/apache/phoenix/parse/QueryParserTest.java
+++ b/phoenix-core/src/test/java/org/apache/phoenix/parse/QueryParserTest.java
@@ -66,6 +66,15 @@ public class QueryParserTest {
         }
     }
 
+    private void parseQueryThatShouldFailWithSQLException(String sql) throws Exception {
+        try {
+            parseQuery(sql);
+            fail("Query should throw a PhoenixParserException \n " + sql);
+        }
+        catch (SQLException e){
+        }
+    }
+
     @Test
     public void testParseGrantQuery() throws Exception {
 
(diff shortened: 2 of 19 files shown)


gjacoby126 (changes requested):
A few questions:
1. This doesn't seem to have a test covering the new feature
2. For numeric literal types, Phoenix does some custom serialization so that they sort lexicographically within HBase so that we can do range scans and order by clauses properly. Do we need to do something like this here, or will PBinary take care of that for us?
3. What if a user wants to treat a hexadecimal value as a numeric type rather than a fixed byte array?

stoty:
1. Forgot to add it to the commit, I've pushed it now.

2. Yes, PBinary should take care of that. This only affects the parser, otherwise everything should behave as if you specified a query parameter and set it with setObject(byte[]).

3. The standard BINMARY->NUMERIC coercion mechanism should apply. I will add a few tests for that.

stoty:
I've checked the SQL2011 standard

On page 188 it says

> 20) It is implementation-defined whether the declared type of a <binary string literal> is a fixed-length binary
> string type, a variable-length binary string type, or a binary large object string type.

So using the BINARY is OK.

Phoenix won't coerce binary types to numeric types, so you cannot treat a binary string literal as a number.

stoty:
I have also found some more requirements regarding the literal handling (multi-line literals and whitespace handling)

I am currently fighting ANTLR to implement those.

stoty:
I've updated the patch with a more flexible implementation.
As we don't enable backtracking, implementing the multi-line literals was a major pain, ended up quite hacky.

The query dumping BINARY format change is probably going to brake a lot of tests, I'll have to do another pass on those.

stoty:
There two open issues:

1. The new litarals don't work with BINARY ARRAYS. I suspect that this is not a new problem, so I'm not terribly concerned about that.

2. The way we convert binaries to strings is inconsitent:
On one hand, we have the Formatters that are used for VARBINARY (and now BINARY) results, on the other we have the toStringLiteral() method in the P*BBinary classes, which are used for Statement.toString(), and for BINARY_ARRAY (which doesn't have an explicit Formatter set)

Before the patch, the formatter generated  _'AABB'_, which the toStringLiteral() methods returned _'[170, 187]'_

With my patch, the formatter is the same, only it's now also set for BINARY, while previously it was only set for VARBINARY.
I've changed the toStringLiteral() formatter to return a valid binary string literal, _X'AABB'_ , so now stmt.toString() generates valid SQL.

The complicate things, it seems that Oracle and Mysql simply returns something like new String(byte[]) for getString(), so the Phoenix behaviour is unique.

I don't think that this is show-stopper either, the patch doesn't make any of those problems any worse.

phoenix-core/src/main/java/org/apache/phoenix/util/ByteUtil.java
```diff
@@ -56,7 +56,34 @@ public class ByteUtil {
             EMPTY_BYTE_ARRAY);
     public static final ImmutableBytesWritable EMPTY_IMMUTABLE_BYTE_ARRAY = new ImmutableBytesWritable(
             EMPTY_BYTE_ARRAY);
-    
+
+
+    /** Mask for bit 0 of a byte. */
+    private static final int BIT_0 = 1;
```
richardantal, line 62: Why don't we have 0x01 here?

stoty: Becuse that's how it's written in commons-codec where I copied this from.

But you're right, it's more consistent to use 0x01 here.

stoty: Fixed.

stoty:
> 1. The new litarals don't work with BINARY ARRAYS. I suspect that this is not a new problem, so I'm not terribly concerned about that. This may be related to the fact the parser generates an ARRAY of VARBINARY nodes, which is not a supported

I have changed the literals to be interpreted as BINARY instead of VARBINARY.
We can still coerce them to varbinary, but this allows using them for BINARY ARRAYs.

I also had to change the size estimation code.
While the BINARY type is fixed length, we don't know / pass its length when trying to estimate its size.

phoenix-core/src/test/java/org/apache/phoenix/parse/QueryParserTest.java
```diff
@@ -925,6 +925,31 @@ public void testShowCreateTable() throws Exception {
...
+    @Test
+    public void testBinaryLiteral() throws Exception {
+        // Happy paths
+        parseQuery("SELECT b, x from x WHERE x = x'00'");
+        parseQuery("SELECT b, x from x WHERE x = "
+                + "x  '0 12 ' --comment \n /* comment */ '34 567' \n \n 'aA'");
+        parseQuery("SELECT b, x from x WHERE x = "
+                + "b  '0 10 ' --comment \n /* comment */ '10 101' \n \n '00000000'");
 
+        // Expected failures.
```
gjacoby126, line 939: Some comments here would be welcome. In particular I'm having trouble seeing why 934 should pass and 940 should fail. (I also find it surprising that 934 and 936 should pass given the use of whitespace in the WHERE clause.)

stoty: Thanks.
I've misread the grammar in the standard, space is allowed before and after digits, but not between X and ' , as you would expect.

stoty:
I have fixed the space handling, and also adopted the ITs to expect the new literal format.

gjacoby126 (approved):
+1, thanks @stoty

gjacoby126 (changes requested):
Just noticed a test failure in the new BinaryLiteralStringIT:

`[ERROR] org.apache.phoenix.end2end.BinaryStringLiteralIT.testBinary  Time elapsed: 2.254 s  <<< ERROR!
org.apache.phoenix.exception.PhoenixParserException: ERROR 602 (42P00): Syntax error. Missing "LPAREN" at line 1, column 32.
	at org.apache.phoenix.exception.PhoenixParserException.newException(PhoenixParserException.java:33)
	at org.apache.phoenix.parse.SQLParser.parseStatement(SQLParser.java:111)
	at org.apache.phoenix.jdbc.PhoenixStatement$PhoenixStatementParser.parseStatement(PhoenixStatement.java:1961)
	at org.apache.phoenix.jdbc.PhoenixStatement.parseStatement(PhoenixStatement.java:2044)
	at org.apache.phoenix.jdbc.PhoenixStatement.executeUpdate(PhoenixStatement.java:2142)
	at org.apache.phoenix.end2end.BinaryStringLiteralIT.insertRow(BinaryStringLiteralIT.java:53)
	at org.apache.phoenix.end2end.BinaryStringLiteralIT.testBinary(BinaryStringLiteralIT.java:71)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:59)
	at org.junit.internal.runners.model.ReflectiveCallable.run(ReflectiveCallable.java:12)
	at org.junit.runners.model.FrameworkMethod.invokeExplosively(FrameworkMethod.java:56)
	at org.junit.internal.runners.statements.InvokeMethod.evaluate(InvokeMethod.java:17)
	at org.junit.runners.ParentRunner$3.evaluate(ParentRunner.java:306)
	at org.junit.runners.BlockJUnit4ClassRunner$1.evaluate(BlockJUnit4ClassRunner.java:100)
	at org.junit.runners.ParentRunner.runLeaf(ParentRunner.java:366)
	at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:103)
	at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:63)
	at org.junit.runners.ParentRunner$4.run(ParentRunner.java:331)
	at org.junit.runners.ParentRunner$1.schedule(ParentRunner.java:79)
	at org.junit.runners.ParentRunner.runChildren(ParentRunner.java:329)
	at org.junit.runners.ParentRunner.access$100(ParentRunner.java:66)
	at org.junit.runners.ParentRunner$2.evaluate(ParentRunner.java:293)
	at org.junit.internal.runners.statements.RunBefores.evaluate(RunBefores.java:26)
	at org.junit.internal.runners.statements.RunAfters.evaluate(RunAfters.java:27)
	at org.apache.phoenix.SystemExitRule$1.evaluate(SystemExitRule.java:40)
	at org.junit.rules.ExternalResource$1.evaluate(ExternalResource.java:54)
	at org.junit.rules.RunRules.evaluate(RunRules.java:20)
	at org.junit.runners.ParentRunner$3.evaluate(ParentRunner.java:306)
	at org.junit.runners.ParentRunner.run(ParentRunner.java:413)
	at org.junit.runners.Suite.runChild(Suite.java:128)
	at org.junit.runners.Suite.runChild(Suite.java:27)
	at org.junit.runners.ParentRunner$4.run(ParentRunner.java:331)
	at org.junit.runners.ParentRunner$1.schedule(ParentRunner.java:79)
	at org.junit.runners.ParentRunner.runChildren(ParentRunner.java:329)
	at org.junit.runners.ParentRunner.access$100(ParentRunner.java:66)
	at org.junit.runners.ParentRunner$2.evaluate(ParentRunner.java:293)
	at org.junit.runners.ParentRunner$3.evaluate(ParentRunner.java:306)
	at org.junit.runners.ParentRunner.run(ParentRunner.java:413)
	at org.apache.maven.surefire.junitcore.JUnitCore.run(JUnitCore.java:55)
	at org.apache.maven.surefire.junitcore.JUnitCoreWrapper.createRequestAndRun(JUnitCoreWrapper.java:137)
	at org.apache.maven.surefire.junitcore.JUnitCoreWrapper.executeLazy(JUnitCoreWrapper.java:119)
	at org.apache.maven.surefire.junitcore.JUnitCoreWrapper.execute(JUnitCoreWrapper.java:87)
	at org.apache.maven.surefire.junitcore.JUnitCoreWrapper.execute(JUnitCoreWrapper.java:75)
	at org.apache.maven.surefire.junitcore.JUnitCoreProvider.invoke(JUnitCoreProvider.java:158)
	at org.apache.maven.surefire.booter.ForkedBooter.runSuitesInProcess(ForkedBooter.java:456)
	at org.apache.maven.surefire.booter.ForkedBooter.execute(ForkedBooter.java:169)
	at org.apache.maven.surefire.booter.ForkedBooter.run(ForkedBooter.java:595)
	at org.apache.maven.surefire.booter.ForkedBooter.main(ForkedBooter.java:581)
Caused by: MissingTokenException(inserted [@-1,0:0='<missing LPAREN>',<105>,1:31] at x)
	at org.apache.phoenix.parse.PhoenixSQLParser.recoverFromMismatchedToken(PhoenixSQLParser.java:382)
	at org.antlr.runtime.BaseRecognizer.match(BaseRecognizer.java:115)
	at org.apache.phoenix.parse.PhoenixSQLParser.not_expression(PhoenixSQLParser.java:8328)
	at org.apache.phoenix.parse.PhoenixSQLParser.and_expression(PhoenixSQLParser.java:8142)
	at org.apache.phoenix.parse.PhoenixSQLParser.or_expression(PhoenixSQLParser.java:8079)
	at org.apache.phoenix.parse.PhoenixSQLParser.expression(PhoenixSQLParser.java:8044)
	at org.apache.phoenix.parse.PhoenixSQLParser.one_or_more_expressions(PhoenixSQLParser.java:10358)
	at org.apache.phoenix.parse.PhoenixSQLParser.upsert_node(PhoenixSQLParser.java:5987)
	at org.apache.phoenix.parse.PhoenixSQLParser.oneStatement(PhoenixSQLParser.java:883)
	at org.apache.phoenix.parse.PhoenixSQLParser.statement(PhoenixSQLParser.java:532)
	at org.apache.phoenix.parse.SQLParser.parseStatement(SQLParser.java:108)
	... 49 more`

stoty:
Thanks, I haven't fix the extra space in the other test method.
Pushed the fix now.

gjacoby126 (approved):
+1, thanks @stoty

PR #1495, opened 2022-08-29
PHOENIX-6751 Force using range scan vs skip scan when using large IN clause

jpisaac:


jpisaac:
@tkhurana Please review

jpisaac:
@tkhurana Please review

tkhurana (approved):

phoenix-core/src/main/java/org/apache/phoenix/compile/WhereOptimizer.java
```diff
@@ -308,6 +322,30 @@ public static Expression pushKeyExpressionsToScan(StatementContext context, Set<
...
+
+            // Do not use the skipScanFilter when there is a large IN clause (for e.g > 50k elements)
+            // Since the generation of point keys for skip scan filter will blow up the memory usage.
+            // See ScanRanges.getPointKeys(...) where using the various slot key ranges
+            // to generate point keys will lead to combinatorial explosion.
+            // The following check will ensure the cardinality of generated point keys
+            // is below the configured max (maxInListSkipScanSize).
+            // We shall force a range scan if the configured max is exceeded.
+            // cnfStartPos => is the start slot of this IN list
+            if (checkMaxSkipScanCardinality) {
```
chrajeshbabu, line 334: Can this conversion of skip scan to range scan configurable because range scan on bigger data sets is slow?

gjacoby126: There's already a config param to control how many elements triggers the conversion, which could be used to turn it off by setting it very high. However, when we get in this state (high cardinality skip scans) we find that we get OOM exceptions even with large client-side heaps, which is worse than a slow query.

gjacoby126:
@jpisaac @tkhurana - could you please explain what's different between this version of 6751 and the one that was previously committed and then reverted?

tkhurana:
> @jpisaac @tkhurana - could you please explain what's different between this version of 6751 and the one that was previously committed and then reverted?

@gjacoby126 The difference is the use of biginteger to avoid overflow issues when determining whether to use skip scan or range scan. 
https://github.com/apache/phoenix/pull/1495/files#diff-9494157265dca11f05041f6a7238b64857cc22e08c3c23e92412e1a0da1d12feR335-R343

gjacoby126 (approved):
+1

phoenix-core/src/main/java/org/apache/phoenix/compile/WhereOptimizer.java
jpisaac: @chrajeshbabu For some queries especially with RVC expression and mixed sort orders, the cost of optimization results in huge memory allocations and sometimes even exceeds the number of KEY_RANGES [allowed](https://github.com/apache/phoenix/blob/8612c81dfec001c5b88b13d7a5c7ccc9038f41f7/phoenix-core/src/main/java/org/apache/phoenix/filter/SkipScanFilter.java#L580) This JIRA provides a framework for opting out the optimization path when a certain threshold is reached. We will be working towards an algorithm that is more linear in nature than combinatorial as is the case today.

PR #1501, opened 2022-09-14
PHOENIX-5422 Use Java8 DateTime APIs instead of joda-time APIs

richardantal:
Change-Id: Icb0e89060528930282c139af1c7254cc8d36e054


phoenix-core/src/main/java/org/apache/phoenix/expression/function/FloorWeekExpression.java
```diff
@@ -36,8 +37,7 @@ public FloorWeekExpression(List<Expression> children) {
     }
 
     @Override
-    public long roundDateTime(DateTime datetime) {
-        return datetime.weekOfWeekyear().roundFloorCopy().getMillis();
+    public long roundDateTime(Date dateTime) {
+        return dateTime.getTime() - dateTime.getTime() % week - correction;
```
gjacoby126, line 41: This can't be done using the floor function in DateTimeUtils and Calendar.WEEK_OF_YEAR? 

As a side note, I see in the docs in https://phoenix.apache.org/language/functions.html#floor that we only support DAY, HOUR, MINUTE, SECOND and MILLISECOND, so I'm surprised this was here...

phoenix-core/src/main/java/org/apache/phoenix/expression/function/RoundWeekExpression.java
```diff
@@ -17,25 +17,36 @@
...
     public RoundWeekExpression(List<Expression> children) {
        super(children);
     }
 
     @Override
-    public long roundDateTime(DateTime dateTime) {
-       return dateTime.weekOfWeekyear().roundHalfEvenCopy().getMillis();
+    public long roundDateTime(Date dateTime) {
+        long nextMonday = new CeilWeekExpression().roundDateTime(dateTime);
+        long prevMonday = new FloorWeekExpression().roundDateTime(dateTime);
```
gjacoby126, line 43: Likewise, we can't use DateUtils.round here?

phoenix-core/src/test/java/org/apache/phoenix/util/DateUtilTest.java
```diff
@@ -183,10 +183,11 @@ public void testParseTimestamp_tooLargeNanos() {
         DateUtil.parseTimestamp("1970-01-01 00:00:10.9999999999");
     }
 
-    @Test(expected=IllegalDataException.class)
-    public void testParseTimestamp_missingNanos() {
-        DateUtil.parseTimestamp("1970-01-01 00:00:10.");
-    }
+//    @Test(expected=IllegalDataException.class)
+//    public void testParseTimestamp_missingNanos() {
```
gjacoby126, line 187: If the test is no longer valid then we can just delete it.

phoenix-core/src/main/java/org/apache/phoenix/util/DateUtil.java
```diff
@@ -72,13 +74,26 @@ public class DateUtil {
...
     public static final Format DEFAULT_TIMESTAMP_FORMATTER = DEFAULT_MS_DATE_FORMATTER;
 
-    private static final DateTimeFormatter JULIAN_DATE_TIME_FORMATTER = new DateTimeFormatterBuilder()
-        .append(ISODateTimeFormat.dateParser())
-        .appendOptional(new DateTimeFormatterBuilder()
-                .appendLiteral(' ').toParser())
-        .appendOptional(new DateTimeFormatterBuilder()
-                .append(ISODateTimeFormat.timeParser()).toParser())
-        .toFormatter().withChronology(GJChronology.getInstanceUTC());
+    private static final DateTimeFormatter ISO_DATE_TIME_FORMATTER = new DateTimeFormatterBuilder()
```
gjacoby126, line 77: Shouldn't ISO_DATE_TIME_FORMATTER and ISO_LOCAL_DATE_TIME_FORMATTER use IsoChronology?

Note that using BritishCutoverChronology is _not_ the same as Joda's GJChronology, because GJChronology cuts over between Julian and Gregorian time in 1582 and BritishCutoverChronology cuts over in 1752.

phoenix-core/src/main/java/org/apache/phoenix/expression/function/CeilWeekExpression.java
```diff
@@ -37,8 +40,7 @@ public CeilWeekExpression(List<Expression> children) {
     }
 
     @Override
-    public long roundDateTime(DateTime dateTime) {
-        return dateTime.weekOfWeekyear().roundCeilingCopy().getMillis();
+    public long roundDateTime(Date dateTime) {
```
gjacoby126, line 43: Nothing in DateUtils will do this?

phoenix-core/src/it/java/org/apache/phoenix/end2end/DateTimeIT.java
```diff
@@ -577,7 +577,7 @@ public void testYearFunctionDate() throws SQLException {
 
         assertEquals(2006, callYearFunction("\"YEAR\"(TO_DATE('2006-12-13'))"));
 
-        assertEquals(2015, callYearFunction("\"YEAR\"(TO_DATE('2015-W05'))"));
+        assertEquals(2015, callYearFunction("\"YEAR\"(TO_DATE('2015-W05-1'))"));
```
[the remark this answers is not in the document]
richardantal: Yes, Joda time supports weird date formats as well.
2015-W05-1 means the fifth week of 2015 and One digit for the day-of-week. The value run from Monday (1) to Sunday (7). I guess joda time defaulted the day value to 1.
I added a separate ISO_WEEK_DATE_TIME_FORMATTER to parse dates like this.

phoenix-core/src/it/java/org/apache/phoenix/end2end/DefaultColumnValueIT.java
```diff
@@ -593,7 +593,7 @@ public void testDefaultExpression() throws Exception {
                 + "c3 DECIMAL DEFAULT TO_NUMBER('" + DEFAULT_CURRENCY_SYMBOL + "123.33', '\u00A4###.##'),"
                 + "c4 VARCHAR DEFAULT 'AB' || 'CD',"
                 + "c5 CHAR(5) DEFAULT 'E' || 'F',"
-                + "c6 INTEGER DEFAULT \"MONTH\"(TO_TIMESTAMP('2015-6-05'))"
+                + "c6 INTEGER DEFAULT \"MONTH\"(TO_TIMESTAMP('2015-06-05'))"
```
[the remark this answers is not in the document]
richardantal: Yes, you are right.
ISO_LOCAL_DATE in Java time requires the month of year and the day of month to be 2 digits, but for joda 1 digit was fine.

phoenix-core/src/main/java/org/apache/phoenix/expression/function/FloorWeekExpression.java
richardantal: I have to check this, to be honest I've done this patch more than a year ago but we because of 4.x branch we delayed it until that is dropped and we can use java8.

https://phoenix.apache.org/language/functions.html#floor page is generated with an old phoenix, that is not really up to date as far as I know.

phoenix-core/src/main/java/org/apache/phoenix/expression/function/RoundWeekExpression.java
richardantal: I will check it again.

phoenix-core/src/main/java/org/apache/phoenix/util/DateUtil.java
richardantal: This is good question. 
If we use IsoChronology we would face PHOENIX-6486 again.
I found that BritishCutoverChronology was the closest to the gregorian julian chronology that we have and all the tests passed. I have to check the dates between 1582 and 1752, and maybe implement a Gregorian chronology for java time.

richardantal:
Thank you @gjacoby126 for the review and for the insightful questions.

I will be on holiday for a week and a half but after that I'll give further updates.

phoenix-core/src/main/java/org/apache/phoenix/expression/function/RoundWeekExpression.java
richardantal: DateUtils, doesn't support rounding to weeks.

phoenix-core/src/main/java/org/apache/phoenix/util/DateUtil.java
richardantal: added implementation for gregorianJulian Chronology

phoenix-core/src/main/java/org/apache/phoenix/util/chrono/AbstractDateCopy.java
```diff
@@ -0,0 +1,398 @@
...
+import java.time.temporal.UnsupportedTemporalTypeException;
+import java.time.temporal.ValueRange;
+
+/**
+ * An abstract date based on a year, month and day.
+ *
+ * <h3>Implementation Requirements</h3>
+ * Implementations must be immutable and thread-safe.
+ */
+abstract class AbstractDateCopy
```
richardantal, line 56: This class has been copied from threeten-extra

phoenix-core/src/main/java/org/apache/phoenix/util/chrono/JulianChronologyCopy.java
```diff
@@ -0,0 +1,395 @@
...
+ *  Month 2 has 28 days, or 29 in a leap year.
+ * <li>day-of-year - There are 365 days in a standard Julian year and 366 in a leap year.
+ *  The days are numbered from 1 to 365 or 1 to 366.
+ * <li>leap-year - Leap years occur every 4 years.
+ * </ul>
+ *
+ * <h3>Implementation Requirements</h3>
+ * This class is immutable and thread-safe.
+ */
+public final class JulianChronologyCopy extends AbstractChronology implements Serializable {
```
richardantal, line 88: This class has been copied from threeten-extra

phoenix-core/src/main/java/org/apache/phoenix/util/chrono/JulianDateCopy.java
```diff
@@ -0,0 +1,467 @@
...
+ * The Julian differs from the Gregorian only in terms of the leap year rule.
+ * Dates are aligned such that {@code 0001-01-01 (Julian)} is {@code 0000-12-30 (ISO)}.
+ *
+ * <h3>Implementation Requirements</h3>
+ * This class is immutable and thread-safe.
+ * <p>
+ * This class must be treated as a value type. Do not synchronize, rely on the
+ * identity hash code or use the distinction between equals() and ==.
+ */
+public final class JulianDateCopy
```
richardantal, line 76: This class has been copied from threeten-extra

phoenix-core/src/main/java/org/apache/phoenix/util/chrono/GJChronology.java
```diff
@@ -0,0 +1,467 @@
...
+ *  The cutover October 4, 1582 (Julian) is followed by October 15, 1582 (Gregorian)
+ * <li>day-of-year - There are 365 days in a standard year and 366 in a leap year.
+ *  The days are numbered from 1 to 365 or 1 to 366.
+ *  The cutover year 1582 has values from 1 to 356 and a length of 356 days.
+ * </ul>
+ *
+ * <h3>Implementation Requirements</h3>
+ * This class is immutable and thread-safe.
+ */
+public final class GJChronology
```
richardantal, line 100: This class was created based on BritishCutoverChronology in threeten-extra
There are minor changes in the CUTOVER,  CUTOVER_DAYS, comments.
Added test for the cutover dates.

phoenix-core/src/main/java/org/apache/phoenix/util/chrono/GJDate.java
```diff
@@ -0,0 +1,553 @@
...
+ * <p>
+ * This date operates using the {@linkplain GJChronology GregorianJulian calendar}.
+ *
+ * <h3>Implementation Requirements</h3>
+ * This class is immutable and thread-safe.
+ * <p>
+ * This class must be treated as a value type. Do not synchronize, rely on the
+ * identity hash code or use the distinction between equals() and ==.
+ */
+public final class GJDate
```
richardantal, line 72: This class was created based on BritishCutoverDate in threeten-extra
There are minor changes in the lengthOfMonth,  lengthOfYear, comments.
Added test for the cutover dates.

richardantal (commented):
Adding the Gregorian Julian chronology to threeten-extra would be the best solution on the long term. That way we could eliminate duplicating classes because of private / default access modifiers.

stoty:
I agree with you @gjacoby126, but let me provide some context:

In my experience the Chronology issue has less to with real historical use cases than with being able to handle synthetic "minus infinity" constants like '01-01-01' or '1000-01-01' consistently.

The reason we stuck with GJChornology at the last change was because that's what Phoenix has effectively been using already, based on the default behaviour of java.util.Calendar, and this was the easiest way to be consistent on the read and write paths (and date functions) and between Joda and j.u.Calendar, which is still extensively used in Phoenix code.

GJ is also the only Chronology that works without surprises with naive java Date code, i.e. when you do 

```
// assume we're in UTC
j.u.t.Date d = new j.u.d(1000,1,1,1,1,1);
// write d to Phoenix with a PreparedStatement
// read d back
assertEquals("1000-01-01 01:01:01", rs.getString("d"))
```

It is only going to succeed with GJChronology (or similar, like BritishCutoverChronology).

Consequently, as long as Phoenix uses j.u.Date / Calender with the default Chronology, it is the only Calendar that can guarantee internal consistency in Phoenix.
The above is also true for views on HBase tables where custom timestamps are naively generated.

In a nutshell, GJChronology is the default chronology for j.u.Calendar/Date, while ISOChronology is the default for Joda and java.time, and we need to handle this discrepancy somehow.

At the same time, using a cutover chronology is both against the standard, and has a performance cost, so I support using ISOChronology as the default, and basically pruging all j.u.Date/Calendar code from Phoenix.

However, for backwards compatibility both with existing code and existing data, I believe that we also ought to make Chronology configurable.

phoenix-core/pom.xml
```diff
@@ -462,6 +462,11 @@
     </dependency>
 
     <!-- Other dependencies -->
+    <dependency>
```
gjacoby126, line 465: We're copying code from it but still including it?

phoenix-core/src/main/java/org/apache/phoenix/util/chrono/AbstractDateCopy.java
gjacoby126, line 56: Why use "Copy" in the name? If we need to differentiate it from the original threeten class, perhaps we could prefix it with "Phoenix"?

phoenix-core/src/main/java/org/apache/phoenix/util/chrono/JulianChronologyCopy.java
gjacoby126: Similar question about the use of "Copy" vs some other distinguishing prefix

gjacoby126:
Thanks @stoty , I hadn't realized that the Java 7 (and before) GregorianCalendar had an internal cutover similar to GJChronology. 

So for me then the question is what version is this PR targetted for? If this is for 5.2, I can see it making sense for a short-term change that keeps the old behavior but removes the need for joda. (But would there be any problem to keeping joda for the 5.2 cycle and replacing it in 5.3 as part of a more comprehensive rework? If the answer is yes, that's fine, it's just a pretty big patch if a lot of this is temporary.) 

If this is for 5.3, it makes sense to do the larger work to get the date and time logic to what you suggest -- no j.u.Date / Calendar, switching to ISOChronology with the ability for users to configure an alternative.

PR #1503, opened 2022-09-21
PHOENIX-6773 PhoenixDatabaseMetadata.getColumns() always returns null COLUMN_DEF

Aarchy:


phoenix-core/src/main/java/org/apache/phoenix/jdbc/PhoenixDatabaseMetaData.java
```diff
@@ -852,7 +852,7 @@ public ResultSet getColumns(String catalog, String schemaPattern, String tableNa
                 cells.add(
                     PhoenixKeyValueUtil.newKeyValue(rowKey, TABLE_FAMILY_BYTES,
                         Bytes.toBytes(COLUMN_DEF),
-                        MetaDataProtocol.MIN_TABLE_TIMESTAMP, ByteUtil.EMPTY_BYTE_ARRAY));
+                        MetaDataProtocol.MIN_TABLE_TIMESTAMP, PVarchar.INSTANCE.toBytes(column.getExpressionStr())));
```
stoty, line 855: This is too long, break the line somewhere.

stoty (changes requested):

phoenix-core/src/main/java/org/apache/phoenix/jdbc/PhoenixDatabaseMetaData.java
Aarchy: Thanks, fixed.

PR #1504, opened 2022-09-27
PHOENIX-5066 The TimeZone is incorrectly used during writing or readi…

stoty:
…ng data


dbwong:
I had a high level question on your approach.  Your email used thread local variable for somewhat global access, i worry if our execution engine becomes more multi threaded you will have to copy this to all the threads and become more and more cumbersome.  I'm hesitant of such patterns for things that are not semaphore like.  I assume the resistance in the code is from our lack of structure and not wanting to pass a parameter everywhere.  I think another approach is possible with a provider pattern given the connection and some jvm level singleton map.  Did you consider this approach?

stoty:
Abandoning in favor of the WITHOUT TIMEZONE solution.

PR #1506, opened 2022-09-27
PHOENIX-6761 : Phoenix Client Side Metadata Caching Improvement

palashc:


gjacoby126 (commented):
Overall this is good -- it both simplifies the code and looks likely to improve our memory footprint, as the JIRA says. 

My main questions are around thread safety. Connections aren't thread-safe, but CQSI is supposed to be, so by exposing CQSI state off of Connection we have to verify that two Connections owned by different threads but using the same CQSI can't corrupt each other's state. 

CQSI seems to sometimes use a mutex (in metaDataMutated) to serialize access to the PMetaData, but that seems to just be used when dropping columns. There are plenty of times when it doesn't (such as when adding tables and columns).

Most of PMetaDataImpl's state is in PMetadataCache, which seems to use thread-safe constructs like the Guava cache and ConcurrentHashMaps, but I haven't worked through all the implications to make sure we're always OK. 

I'm curious to hear @palashc and @kadirozde 's thoughts on this. Have these issues been worked through?

phoenix-core/src/main/java/org/apache/phoenix/schema/PMetaDataCache.java
```diff
@@ -17,176 +17,85 @@
...
 
-import java.util.Comparator;
 import java.util.Iterator;
 import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
 
+import org.apache.phoenix.jdbc.PhoenixDatabaseMetaData;
 import org.apache.phoenix.parse.PFunction;
 import org.apache.phoenix.parse.PSchema;
+import org.apache.phoenix.thirdparty.com.google.common.cache.*;
```
tkhurana, line 27: nit: Don't use wildcard in import

tkhurana:
@palashc What kind of metrics do we have around the metadata cache ? We should have some metrics around how many requests we get, our hit rate etc..

palashc:
> @palashc What kind of metrics do we have around the metadata cache ? We should have some metrics around how many requests we get, our hit rate etc..

@tkhurana that's a good idea, right now we do not have any metrics, I can definitely add some.

phoenix-core/src/it/java/org/apache/phoenix/rpc/UpdateCacheIT.java
```diff
@@ -87,7 +87,7 @@ public void testUpdateCacheForTxnTable() throws Exception {
             String fullTableName = INDEX_DATA_SCHEMA + QueryConstants.NAME_SEPARATOR + tableName;
             Connection conn = DriverManager.getConnection(getUrl(), PropertiesUtil.deepCopy(TEST_PROPERTIES));
             conn.createStatement().execute("create table " + fullTableName + TestUtil.TEST_TABLE_SCHEMA + "TRANSACTIONAL=true,TRANSACTION_PROVIDER='" + provider + "'");
-            helpTestUpdateCache(fullTableName, new int[] {1, 1}, false);
+            helpTestUpdateCache(fullTableName, new int[] {1, 3}, false);
```
kadirozde, line 90: What is the reason for this increase from 1 to 3 in expected RPCs?

phoenix-core/src/main/java/org/apache/phoenix/jdbc/PhoenixConnection.java
```diff
@@ -367,7 +368,7 @@ public ReadOnlyProps getProps() {
             // We do not limit the metaData on a connection less than the global
             // one,
             // as there's not much that will be cached here.
-            Pruner pruner = new Pruner() {
+            PMetaData.Pruner pruner = new PMetaData.Pruner() {
```
kadirozde, line 371: Is this pruner used anymore?

phoenix-core/src/it/java/org/apache/phoenix/end2end/AppendOnlySchemaIT.java
```diff
@@ -112,9 +114,9 @@ private void testTableWithSameSchema(boolean notExists, boolean sameClient) thro
             }
             
             // verify getTable rpcs
-            verify(connectionQueryServices, sameClient ? never() : times(1))
-                    .getTable((PName) isNull(), eq(new byte[0]),
-                            eq(Bytes.toBytes(viewName)), anyLong(), anyLong());
+            verify(connectionQueryServices, never())
```
kadirozde, line 117: Is this change required?

phoenix-core/src/main/java/org/apache/phoenix/query/ConnectionlessQueryServicesImpl.java
```diff
@@ -159,7 +159,14 @@ public ConnectionlessQueryServicesImpl(QueryServices services, ConnectionInfo co
     }
 
     private PMetaData newEmptyMetaData() {
-        return new PMetaDataImpl(INITIAL_META_DATA_TABLE_CAPACITY, getProps());
+        long updateCacheFrequency = (Long) ConnectionProperty.UPDATE_CACHE_FREQUENCY.getValue(
+                getProps().get(QueryServices.DEFAULT_UPDATE_CACHE_FREQUENCY_ATRRIB));
+        // We can have zero update cache frequency for connectionless query services as we need to keep the
```
kadirozde, line 164: Did we mean "We cannot" here?

phoenix-core/src/it/java/org/apache/phoenix/end2end/AppendOnlySchemaIT.java
palashc: This test was checking if there was a `CQSI.getTable` call when a different connection is used to execute the same create ddl. With the new design, there will be no call to `CQSI.getTable` since the other connection will find the table metadata in the CQSI cache.

phoenix-core/src/main/java/org/apache/phoenix/jdbc/PhoenixConnection.java
palashc: It seems that it is not. I will remove it.

phoenix-core/src/it/java/org/apache/phoenix/end2end/AppendOnlySchemaIT.java
kadirozde: Got it

phoenix-core/src/main/java/org/apache/phoenix/schema/PMetaDataImpl.java
```diff
@@ -46,36 +48,52 @@ public class PMetaDataImpl implements PMetaData {
...
         this.timeKeeper = timeKeeper;
         this.metaData = metaData;
         this.tableRefFactory = tableRefFactory;
+        this.updateCacheFrequency = updateCacheFrequency;
     }
 
-    @Override
-    public PMetaDataImpl clone() {
-        return new PMetaDataImpl(new PMetaDataCache(this.metaData), this.timeKeeper, this.tableRefFactory);
+    private void updateGlobalMetric(PTableRef pTableRef) {
```
tkhurana, line 76: I don't think you need a lock to just increment counters

palashc: If multiple threads share the same CQSI object, will incrementing counters for cache access without synchronizing be correct?

tkhurana: Double check but I think these counters are atomic

palashc: You're right, they are of type `AtomicMetric`. I will remove the lock I added.

phoenix-core/src/it/java/org/apache/phoenix/rpc/UpdateCacheIT.java
kadirozde: I see that the same test for non TX table also needs 3 RPCs. I assume our change eliminated this discrepancy.

kadirozde (approved):
+1

PR #1508, opened 2022-09-28
PHOENIX-6752 Duplicate expression nodes in extract nodes during WHERE compilation phase leads to poor performance

jpisaac:


phoenix-core/src/test/java/org/apache/phoenix/query/BaseTest.java
```diff
@@ -1692,7 +1697,57 @@ public Void call() throws Exception {
 
         LOGGER.info("Disabled and dropped {} tables in {} ms", tableCount, endTime-startTime);
     }
-    
+
+    protected int setBindVariables(PhoenixPreparedStatement stmt, int startBindIndex, int numBinds, PDataType[] testPKTypes)
+            throws SQLException {
+
```
gjacoby126, line 1703: Should this be in BaseTest? It seems to make some test-specific assumptions (like defaulting a VARCHAR(25), and the random limits.

gjacoby126 (changes requested):
Just nits so far, let's see what the test results are

comnetwork (changes requested):

phoenix-core/src/it/java/org/apache/phoenix/end2end/QueryIT.java
```diff
@@ -169,4 +174,41 @@ public void testDistinctLimitScan() throws Exception {
...
+            stmtForTimingCheck.setTimestamp(1, new Timestamp(System.currentTimeMillis() + rnd.nextInt(50000)));
+            stmtForTimingCheck.setString(2, RandomStringUtils.randomAlphanumeric(30));
+
+            for (int i = 0; i<numORs+1; i++) {
+                stmtForTimingCheck.setInt(i + 3, rnd.nextInt() );
+            }
+
+            long startResultSetTime = System.currentTimeMillis();
+            ResultSet rs = stmtForTimingCheck.executeQuery(query.toString());
+            assertTrue("Query execution time exceeded limit exceeded 10s", (System.currentTimeMillis() - startResultSetTime) < 10000);
```
comnetwork, line 211: I suggest we would better not assert the query time in test,  which may be cause the test flaky. In my opinion, for test, we could assert the internal state which your PR modified to cause the query not timeout , and is same as above, I think it is would be a UT, not a IT test.

phoenix-core/src/main/java/org/apache/phoenix/compile/WhereOptimizer.java
```diff
@@ -926,7 +934,11 @@ private KeySlots andKeySlots(AndExpression andExpression, List<KeySlots> childSl
                             if (slot.getOrderPreserving() != null) {
                                 orderPreserving = slot.getOrderPreserving().combine(orderPreserving);
                             }
-                            if (slot.getKeyPart().getExtractNodes() != null) {
+                            // Extract once per iteration, when there are large number
+                            // of OR clauses (for e.g N > 100k).
+                            // The extractNodes.addAll method can get called N times.
+                            if (!visitedKeyParts.contains(slot.getKeyPart()) && slot.getKeyPart().getExtractNodes() != null) {
+                                visitedKeyParts.add(slot.getKeyPart());
                                 extractNodes.addAll(slot.getKeyPart().getExtractNodes());
```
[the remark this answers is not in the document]
gjacoby126: @comnetwork - I believe this JIRA is a mitigation for 5.2, with a more comprehensive rewrite to be done as part of PHOENIX-6791 in 5.3.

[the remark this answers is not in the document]
comnetwork: @gjacoby126 , ok,  thank you for reply.

gjacoby126:
@jpisaac - Checking back in on this patch -- it looks like there are still some review comments for the tests that haven't been addressed yet. Do you have bandwidth to help get this patch in? Alternately, what are the implications if it slips out of 5.2 into a future 5.2.1 or 5.3?

phoenix-core/src/it/java/org/apache/phoenix/end2end/QueryIT.java
stoty: I agree with not checking for timeouts. anything can happen on the test infra, from GC pauses to swapping.

jpisaac:
@gjacoby126 @stoty @tkhurana Can u take a look again? I have addressed all of the review comments. Let me know if there is more.

pom.xml
```diff
@@ -85,7 +85,7 @@
     <!-- The default hbase versions to build with (override with hbase.version) -->
     <hbase-2.3.runtime.version>2.3.7</hbase-2.3.runtime.version>
     <hbase-2.4.0.runtime.version>2.4.0</hbase-2.4.0.runtime.version>
-    <hbase-2.4.runtime.version>2.4.13</hbase-2.4.runtime.version>
```
stoty, line 88: This looks like a typo.

stoty:
I was able to repro the LogicalTableNameIT failures locally.

stoty (approved):
+1 LGTM

The LogicalTableNameIT failures are not caused by this patch.

phoenix-core/src/test/java/org/apache/phoenix/query/BaseTest.java
```diff
@@ -146,13 +149,15 @@
 import org.apache.phoenix.jdbc.PhoenixConnection;
 import org.apache.phoenix.jdbc.PhoenixDatabaseMetaData;
 import org.apache.phoenix.jdbc.PhoenixEmbeddedDriver;
+import org.apache.phoenix.jdbc.PhoenixPreparedStatement;
```
gjacoby126, line 152: nit: added imports with no added code anymore, which probably means these are unused imports.

gjacoby126 (changes requested):
@jpisaac - just a small nit left

jpisaac:
@gjacoby126 Rebased to latest and removed unused imports

jpisaac:
@gjacoby126 can u approve, if there are no concerns?

gjacoby126 (approved):
+1

PR #1516, opened 2022-10-07
PHOENIX-6807 Change return type of PHOENIX_ROW_TIMESTAMP() function

tkhurana:
Return timestamp instead of date.


tkhurana:
@jpisaac @gjacoby126

gjacoby126:
Makes sense. 

@tkhurana  Are there any backwards-compatibility concerns from either:
1. Row timestamps previously written under the old logic
2. A mixed client/server with a client having the old logic and a server having the new?

stoty (commented):
-1

The HBase timestamps are 8 byte values, while Phoenix timestamps are 12 bytes.
This would be an unneccessary perf hit, and additional code complexity for no gain.

As an alternative, I'd suggest adding an 8 bit DATETIME  type, which would basically be just another alias for DATE/TIME , and using that.

Mysql for example has a similar DATETIME type.

PR #1517, opened 2022-10-07
PHOENIX-6776 :- Abort scans of closed connections at ScanningResultIterator

lokiore:
Added isClosing property in PhoenixConnection as well, as we change isClose status after closing all the underline objects (Iterators, scanners, statements).

### DIFF
--- a/phoenix-core/src/it/java/org/apache/phoenix/end2end/PreMatureTimelyAbortScanIt.java
+++ b/phoenix-core/src/it/java/org/apache/phoenix/end2end/PreMatureTimelyAbortScanIt.java
@@ -19,12 +19,15 @@ package org.apache.phoenix.end2end;
 
 import org.apache.phoenix.coprocessor.BaseScannerRegionObserver;
 import org.apache.phoenix.exception.SQLExceptionCode;
+import org.apache.phoenix.iterate.ScanningResultIterator;
+import org.apache.phoenix.jdbc.PhoenixConnection;
 import org.apache.phoenix.query.QueryServices;
 import org.apache.phoenix.thirdparty.com.google.common.collect.Maps;
 import org.apache.phoenix.util.ReadOnlyProps;
 import org.junit.BeforeClass;
 import org.junit.Test;
 import org.junit.experimental.categories.Category;
+import org.mockito.Mockito;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -33,7 +36,6 @@ import java.sql.DriverManager;
 import java.sql.ResultSet;
 import java.sql.SQLException;
 import java.util.Map;
-import java.util.concurrent.locks.Lock;
 
 import static org.junit.Assert.*;
 @Category(ParallelStatsDisabledTest.class)
@@ -56,8 +58,12 @@ public class PreMatureTimelyAbortScanIt extends ParallelStatsDisabledIT {
     @Test
     public void testPreMatureScannerAbortForCount() throws Exception {
 
+        /* Not using Salt Buckets as when we close connection ConcatResultIterator closes Futures used in
+           ParallelResultIterators before we can get a dummy result for ScanningResultIterator to close Scanners in
+           between conn close call and actual closing of Scanners.
+        */
         try (Connection conn = DriverManager.getConnection(getUniqueUrl())) {
-            conn.createStatement().execute("CREATE TABLE LONG_BUG (ID INTEGER PRIMARY KEY, AMOUNT DECIMAL) SALT_BUCKETS = 16");
+            conn.createStatement().execute("CREATE TABLE LONG_BUG (ID INTEGER PRIMARY KEY, AMOUNT DECIMAL) ");
         }
         try (Connection conn = DriverManager.getConnection(getUniqueUrl())) {
             for (int i = 0; i<500000 ; i++) {
@@ -77,21 +83,19 @@ public class PreMatureTimelyAbortScanIt extends ParallelStatsDisabledIT {
                         synchronized (conn) {
                             conn.wait();
                         }
+                        ScanningResultIterator.setIsScannerClosedForceFully(true);
                         ResultSet resultSet = conn.createStatement().executeQuery(
                                     "SELECT COUNT(*) FROM LONG_BUG WHERE ID % 2 = 0");
-                        boolean result = resultSet.next();
-                        assertTrue(result);
+                        resultSet.next();
                         LOG.info("Count of modulus 2 for LONG_BUG :- " + resultSet.getInt(1));
-                        fail();
+                        fail("ResultSet should have been closed");
                     } catch (SQLException sqe) {
-                        //RESULTSET_CLOSED exception expected
-                        if (sqe.getErrorCode() != SQLExceptionCode.RESULTSET_CLOSED.getErrorCode()) {
-                            LOG.error("Error ", sqe);
-                            fail();
-                        }
+                        assertEquals(SQLExceptionCode.FAILED_KNOWINGLY_FOR_TEST.getErrorCode(), sqe.getErrorCode());
                     } catch (Exception e) {
                         LOG.error("Error", e);
                         fail();
+                    } finally {
+                        ScanningResultIterator.setIsScannerClosedForceFully(false);
                     }
                 }
             };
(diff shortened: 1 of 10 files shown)


phoenix-core/src/it/java/org/apache/phoenix/end2end/PreMatureTimelyAbortScanIt.java
```diff
@@ -0,0 +1,98 @@
+package org.apache.phoenix.end2end;
```
gjacoby126, line 1: @lokiore - could you please add an Apache license to this file?

gjacoby126 (changes requested):

phoenix-core/src/it/java/org/apache/phoenix/end2end/PreMatureTimelyAbortScanIt.java
```diff
@@ -0,0 +1,98 @@
...
+            }
+        }
+
+        try {
+            Connection conn = DriverManager.getConnection(getUniqueUrl());
+            Runnable runnable = new Runnable() {
+                @Override
+                public void run() {
+                    try {
+                        ResultSet resultSet = conn.createStatement().executeQuery(
```
gjacoby126, line 59: Doesn't this thread need to loop or wait on the other thread to close the connection first? What if Thread t gets scheduled to run to completion (hence failure) before t1 does anything?

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/BaseScannerRegionObserver.java
```diff
@@ -262,7 +262,7 @@ public void preScannerOpen(org.apache.hadoop.hbase.coprocessor.ObserverContext<R
             // last possible moment. You need to swap the start/stop and make the
             // start exclusive and the stop inclusive.
             ScanUtil.setupReverseScan(scan);
-            if (scan.getFilter() != null && !(scan.getFilter() instanceof PagedFilter)) {
+            if (!(scan.getFilter() instanceof PagedFilter)) {
```
gjacoby126, line 265: Could you explain more about this change? Was there a previous bug where Scans without a filter weren't being paged correctly?

gjacoby126: Is there any problem with a PagedFilter having an internal Filter which is null?

phoenix-core/src/it/java/org/apache/phoenix/end2end/PreMatureTimelyAbortScanIt.java
lokiore: I checked the time for completing query was more that 1500ms, but true that was on local, will add wait for proper conn closing thanks!

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/BaseScannerRegionObserver.java
lokiore: Previously, scans without filter were not being paged as we enabled them for filters only which were having large scans, but now after discussing with @kadirozde we are enabling them for all scans!

phoenix-core/src/it/java/org/apache/phoenix/end2end/PreMatureTimelyAbortScanIt.java
gjacoby126: Another option to consider is whether you can do a synchronous test by, for example, mocking a Connection to always return isClosing = true, isClosed = false, and verifying that ScanningResultIterator does what you expect (runs its close logic)

Fixing the asynchronous logic to be deterministic through locking as you suggest also works.

phoenix-core/src/main/java/org/apache/phoenix/jdbc/PhoenixConnection.java
```diff
@@ -757,6 +759,7 @@ synchronized public void close() throws SQLException {
             }
             
         } finally {
+            isClosing = false;
```
mnpoonia, line 762: It looks like we are bitten by some multi threading issue and we are trying to reduce the probability of that. @lokiore Is there something that i am missing?

lokiore: @mnpoonia this is being added as we don't have any flag unless all the layers below PhoenixConnection is already close and we needed something to know that Connection is being in process of closing.

phoenix-core/src/it/java/org/apache/phoenix/end2end/PreMatureTimelyAbortScanIt.java
gjacoby126: @lokiore - The Thread.sleep calls still make this non-deterministic, don't they?

kadirozde: @lokiore, in general we do not want to use thread sleep as we can only guess the required sleep time and it changes from one system to another. This can make the test flapping. You can implement this test without having threads. In PhoenixConnection, you can define a testing only method to set the connection state to closing. After getting the connection, set the connection state closing using this method. Then you run the test code without creating threads and without closing the connection.

lokiore: Thanks @gjacoby126 @kadirozde for suggesting, I have made changes Please take a look, Thanks

phoenix-core/src/main/java/org/apache/phoenix/iterate/ScanningResultIterator.java
```diff
@@ -190,4 +209,9 @@ public String toString() {
     public ResultScanner getScanner() {
         return scanner;
     }
+
+    @VisibleForTesting
+    public static void setIsScannerClosedForceFully(boolean throwException) {
```
gjacoby126, line 214: tiny nit: "forcefully" is one word so only the first F should be capitalized

gjacoby126 (commented):
Thanks for the test fix. Just one tiny nit left and then I think this is ready to go.

lokiore:
I am trying to re-request review from both @gjacoby126 and @kadirozde but somehow it is removing one when I add other 😓, please take a look, Thanks!!

gjacoby126 (approved):
+1, thanks @lokiore

phoenix-core/src/it/java/org/apache/phoenix/end2end/PreMatureTimelyAbortScanIt.java
```diff
@@ -0,0 +1,89 @@
...
+
+        try (Connection conn = DriverManager.getConnection(getUniqueUrl())) {
+            conn.createStatement().execute("CREATE TABLE LONG_BUG (ID INTEGER PRIMARY KEY, AMOUNT DECIMAL) SALT_BUCKETS = 16 ");
+        }
+        try (Connection conn = DriverManager.getConnection(getUniqueUrl())) {
+            for (int i = 0; i<5000 ; i++) {
+                int amount = -50000 + i;
+                String s = "UPSERT INTO LONG_BUG (ID, AMOUNT) VALUES( " + i + ", " + amount + ")";
+                conn.createStatement().execute(s);
+                conn.commit();
```
tkhurana, line 68: It will be faster if you move this commit out of the loop

kadirozde: Also we should not need to add 5000 rows for this test after the recent changes. Let's reduce it to say 100.

lokiore: Pushed the change Thanks for pointing.

phoenix-core/src/main/java/org/apache/phoenix/jdbc/PhoenixConnection.java
```diff
@@ -1374,6 +1381,11 @@ public void setTableResultIteratorFactory(TableResultIteratorFactory factory) {
         this.tableResultIteratorFactory = factory;
     }
 
+    @VisibleForTesting
```
kadirozde, line 1384: Can you add a comment for this method is that it is only used for testing?

kadirozde (approved):
+1, Thank you!

mnpoonia:
@lokiore This change is causing regression in IndexScrutiny tool which in turn is causing tests to fail
See https://issues.apache.org/jira/browse/PHOENIX-6828

PR #1522, opened 2022-10-31
PHOENIX-6826 Don't invalidate meta cache if CQSI#getTableRegionLocation and CQSI#getTableRegionLocation encounters IOException

shahrs87:


shahrs87:
Following tests failed:
```
[ERROR] Failures: 
[ERROR]   LogicalTableNameIT.testUpdatePhysicalIndexTableName_runScrutiny:229 expected:<2> but was:<1>
[ERROR]   LogicalTableNameIT.testUpdatePhysicalIndexTableName_runScrutiny:229 expected:<2> but was:<1>
[ERROR]   LogicalTableNameIT.testUpdatePhysicalIndexTableName_runScrutiny:229 expected:<2> but was:<1>
[ERROR]   LogicalTableNameIT.testUpdatePhysicalIndexTableName_runScrutiny:229 expected:<2> but was:<1>
[ERROR]   LogicalTableNameIT.testUpdatePhysicalTableNameWithIndex_runScrutiny:169 expected:<2> but was:<1>
[ERROR]   LogicalTableNameIT.testUpdatePhysicalTableNameWithIndex_runScrutiny:169 expected:<2> but was:<1>
[ERROR]   LogicalTableNameIT.testUpdatePhysicalTableNameWithIndex_runScrutiny:165 expected:<3> but was:<1>
[ERROR]   LogicalTableNameIT.testUpdatePhysicalTableNameWithIndex_runScrutiny:165 expected:<3> but was:<1>
[ERROR]   LogicalTableNameIT.testUpdatePhysicalTableNameWithViews_runScrutiny:353 expected:<2> but was:<0>
[ERROR]   LogicalTableNameIT.testUpdatePhysicalTableNameWithViews_runScrutiny:353 expected:<2> but was:<0>
```
These tests fails without this patch also. Created [PHOENIX-6828](https://issues.apache.org/jira/browse/PHOENIX-6828) to track them.
@virajjasani  @tkhurana  Can you please review the PR? Thank you!

phoenix-core/src/main/java/org/apache/phoenix/query/ConnectionQueryServicesImpl.java
```diff
@@ -725,17 +725,17 @@ public List<HRegionLocation> getAllTableRegions(byte[] tableName) throws SQLExce
...
                     locations.add(regionLocation);
                 } while (!Bytes.equals(currentKey, HConstants.EMPTY_END_ROW));
                 return locations;
             } catch (org.apache.hadoop.hbase.TableNotFoundException e) {
-                String fullName = Bytes.toString(tableName);
-                throw new TableNotFoundException(fullName);
+                throw new TableNotFoundException(table.getNameAsString());
             } catch (IOException e) {
+                LOGGER.error("Exception encountered in getAllTableRegions for table: "
+                        + table.getNameAsString(), e);
```
virajjasani, line 737: How about we keep this just above throwing `SQLExceptionInfo.Builder(SQLExceptionCode.GET_TABLE_REGIONS_FAIL)` ? That way we will not log this twice but rather only once when we actually throw the Exception `GET_TABLE_REGIONS_FAIL`

virajjasani (commented):
Left one nit, looks good otherwise

phoenix-core/src/main/java/org/apache/phoenix/query/ConnectionQueryServicesImpl.java
shahrs87: Both the exceptions could be different. I think logging both the exceptions should be fine given that this will be in error condition path and not regular code path. @virajjasani  WDYT?

virajjasani: Yeah sounds good but from debugging viewpoint, I was thinking maybe in some way if we could differentiate b/ first try and then eventual error after retry? Maybe let's include retry count in log message? I think that would be sufficient to differentiate the first and second attempts (and we can clearly see different root cause for both, if this happens).
Sounds good?

phoenix-core/src/main/java/org/apache/phoenix/query/ConnectionQueryServicesImpl.java
```diff
@@ -725,17 +725,17 @@ public List<HRegionLocation> getAllTableRegions(byte[] tableName) throws SQLExce
...
                     locations.add(regionLocation);
                 } while (!Bytes.equals(currentKey, HConstants.EMPTY_END_ROW));
                 return locations;
             } catch (org.apache.hadoop.hbase.TableNotFoundException e) {
-                String fullName = Bytes.toString(tableName);
-                throw new TableNotFoundException(fullName);
+                throw new TableNotFoundException(table.getNameAsString());
             } catch (IOException e) {
+                LOGGER.error(String.format("Exception encountered in getAllTableRegions for " +
+                        "table: {}, retryCount: {}", table.getNameAsString(), retryCount), e);
```
virajjasani: nit: `String.format` is redundant
```
LOGGER.error("Exception encountered in getAllTableRegions for " +
                        "table: {}, retryCount: {}", table.getNameAsString(), retryCount, e);
```

phoenix-core/src/main/java/org/apache/phoenix/query/ConnectionQueryServicesImpl.java
```diff
@@ -5966,16 +5966,17 @@ public HRegionLocation getTableRegionLocation(byte[] tableName, byte[] row) thro
...
+            TableName table = TableName.valueOf(tableName);
             try {
-                return connection.getRegionLocator(TableName.valueOf(tableName)).getRegionLocation(row, reload);
+                return connection.getRegionLocator(table).getRegionLocation(row, false);
             } catch (org.apache.hadoop.hbase.TableNotFoundException e) {
                 String fullName = Bytes.toString(tableName);
                 throw new TableNotFoundException(SchemaUtil.getSchemaNameFromFullName(fullName), SchemaUtil.getTableNameFromFullName(fullName));
             } catch (IOException e) {
+                LOGGER.error(String.format("Exception encountered in getTableRegionLocation for " +
+                        "table: {}, retryCount: {}", table.getNameAsString(), retryCount), e);
```
virajjasani: same as above

shahrs87:
`PermissionNSDisabledWithCustomAccessControllerIT#testDeletingStatsShouldNotFailWithADEWhenTableDropped` passed locally.
@virajjasani  Can you please merge the PR? Thank you!

virajjasani:
Sounds good, this seems good to go. 
@tkhurana any comments from your side?

tkhurana (approved):

PR #1523, opened 2022-11-01
PHOENIX-6827 Update hbase-version to 2.4.15 in phoenix master branch

shahrs87:


virajjasani:
Looks like end2end tests are not run https://ci-hadoop.apache.org/job/Phoenix/job/Phoenix-PreCommit-GitHub-PR/job/PR-1523/1/testReport/

Perhaps we can make one liner change to phoenix-core pom and that would trigger full test suite. Once test results are available, we can remove the one liner change in phoenix-core pom.

virajjasani:
This is weird, we already use hadoop 3 built hbase 2 artifacts so this should not have happened. @stoty @richardantal any idea what is wrong here?

phoenix-core/src/test/java/org/apache/phoenix/mapreduce/index/BaseIndexTest.java
```diff
@@ -36,7 +36,7 @@
  * Creates a simple data table and index table
  *
  */
-public class BaseIndexTest extends BaseConnectionlessQueryTest {
+public class  BaseIndexTest extends BaseConnectionlessQueryTest {
```
virajjasani, line 39: Btw somehow this change also didn't trigger end2end tests. Perhaps adding new line in phoenix-core/pom.xml would do.

stoty:
The Yetus script doesn't handle updating the HBase version in a patch.
The code I added to avoid recompiling HBase multiple times doesn't check if the version has changed.

You can ignore the test error, @virajjasani .

stoty:
I'm not sure why test weren't triggered, maybe only the non-test code is checked by Yetus ?

virajjasani:
> I'm not sure why test weren't triggered, maybe only the non-test code is checked by Yetus ?

Exactly, I feel the same. Hence, i usually just add one line to phoenix-core pom

shahrs87:
Added one line change to phoenix-core/pom.xml. Lets see if this triggers end2end tests.

virajjasani (approved):

virajjasani:
Unfortunately, we still don't have end2end tests

chrajeshbabu (commented):
Hi

virajjasani:
@stoty shall we merge this PR and let pre-commit run builds with 2.4 profile to verify results later?

stoty:
Yes, @virajjasani.

PR #1528, opened 2022-11-11
PHOENIX-6828 LogicalTableNameIT.testUpdatePhysicalIndexTableName failing

mnpoonia:
This test is failing on indexscrutiny tool. I am assuming that index scrutiny tool is not modified after the new index design. For now we can use the index tool instead of index scrutiny as it has all the functionality that scrutiny tool has. File another JIRA to fix index scrutiny


gjacoby126:
@mnpoonia - as far as I know, IndexScrutinyTool should still work even with the consistent secondary index framework that we added a couple of years ago. It's just slow.

mnpoonia:
@gjacoby126 sorry for the confusion. You are right. It was a issue with a commit which caused regression. More context and discussion on [jira. ](https://issues.apache.org/jira/browse/PHOENIX-6828?page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel&focusedCommentId=17635566#comment-17635566)

PR #1531, opened 2022-11-23
PHOENIX-6832 Uncovered Global Secondary Indexes

kadirozde:


tkhurana:
@kadirozde started going through the changes. Can you add some uncovered index tests in OnDuplicateKeyIT

kadirozde:
> @kadirozde started going through the changes. Can you add some uncovered index tests in OnDuplicateKeyIT

@tkhurana, it is a great suggestion. I will add uncovered index tests to OnDuplicateKeyIT.

phoenix-core/src/main/java/org/apache/phoenix/hbase/index/IndexRegionObserver.java
```diff
@@ -815,6 +818,10 @@ private void getCurrentRowStates(ObserverContext<RegionCoprocessorEnvironment> c
                                      BatchMutateContext context) throws IOException {
         Set<KeyRange> keys = new HashSet<KeyRange>(context.rowsToLock.size());
         context.dataRowStates = new HashMap<ImmutableBytesPtr, Pair<Put, Put>>(context.rowsToLock.size());
+        if (context.hasUncoveredIndex && !context.hasGlobalIndex && !context.hasTransform
```
tkhurana, line 821: @kadirozde If there are multiple columns part of an uncovered index row key and we get a partial update which includes only a subset of those columns then in this case we still need to read the data table row to correctly generate the index row key.

kadirozde: @tkhurana, You are right. I made the necessary change and included a test case for it. Currently, if a mutation in a mini batch is a partial update, the data table rows will read for all mutations in the batch. We can further optimize this in future by just reading rows only for partial mutations instead of all mutations.

phoenix-core/src/main/java/org/apache/phoenix/hbase/index/IndexRegionObserver.java
```diff
@@ -807,14 +810,39 @@ private void handleLocalIndexUpdates(TableName table,
...
+            if (miniBatchOp.getOperationStatus(i) == IGNORE) {
+                continue;
+            }
+            Mutation m = miniBatchOp.getOperation(i);
+            if (!this.builder.isEnabled(m)) {
+                continue;
+            }
+            for (ColumnReference indexedColumn : indexedColumns) {
+                if (m.get(indexedColumn.getFamily(), indexedColumn.getQualifier()).isEmpty()) {
+                    return true;
```
gjacoby126, line 834: nit: a comment here explaining how an empty column qualifier means the index update is partial

phoenix-core/src/main/java/org/apache/phoenix/index/IndexMaintainer.java
```diff
@@ -105,16 +107,7 @@
...
-import org.apache.phoenix.util.ByteUtil;
-import org.apache.phoenix.util.EncodedColumnsUtil;
-import org.apache.phoenix.util.ExpressionUtil;
-import org.apache.phoenix.util.IndexUtil;
-import org.apache.phoenix.util.MetaDataUtil;
-import org.apache.phoenix.util.PhoenixRuntime;
-import org.apache.phoenix.util.SchemaUtil;
-import org.apache.phoenix.util.TransactionUtil;
-import org.apache.phoenix.util.TrustedByteArrayOutputStream;
+import org.apache.phoenix.util.*;
```
gjacoby126, line 110: nit: avoid star imports

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/index/IndexUpgradeTool.java
```diff
@@ -52,9 +52,7 @@
 import org.apache.phoenix.schema.PIndexState;
 import org.apache.phoenix.schema.PTable;
 import org.apache.phoenix.schema.PTableType;
-import org.apache.phoenix.util.EnvironmentEdgeManager;
-import org.apache.phoenix.util.MetaDataUtil;
-import org.apache.phoenix.util.PhoenixRuntime;
+import org.apache.phoenix.util.*;
```
gjacoby126, line 55: nit: avoid star imports

phoenix-core/src/main/java/org/apache/phoenix/schema/PTable.java
```diff
@@ -104,8 +104,9 @@ public ViewType combine(ViewType otherType) {
     }
 
     public enum IndexType {
-        GLOBAL((byte)1),
-        LOCAL((byte)2);
+        GLOBAL((byte)1), // Covered Global
+        LOCAL((byte)2), // Covered Local
+        UNCOVERED((byte)3); // Uncovered Global
```
gjacoby126, line 109: Since we can do uncovered queries on local indexes, would it be more clear if the enum were UNCOVERED_GLOBAL (or GLOBAL_UNCOVERED)?

phoenix-core/src/it/java/org/apache/phoenix/end2end/index/UncoveredGlobalIndexRegionScannerIT.java
```diff
@@ -80,7 +93,39 @@ private void populateTable(String tableName) throws Exception {
...
+                try {
+                    conn.createStatement().execute("CREATE UNCOVERED INDEX " + indexTableName
+                            + " on " + dataTableName + " (val1) INCLUDE (val2)");
+                    Assert.fail();
+                } catch (Exception e) {
+                    // Expected
+                }
+                // The LOCAL keyword should not be allowed with UNCOVERED
+                try {
+                    conn.createStatement().execute("CREATE UNCOVERED LOCAL INDEX " + indexTableName
```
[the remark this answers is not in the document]
gjacoby126: This makes more sense when you know from the PTable.IndexType enum comments that UNCOVERED really means "UNCOVERED GLOBAL", but that's not obvious from the syntax.

phoenix-core/src/it/java/org/apache/phoenix/end2end/index/UncoveredGlobalIndexRegionScannerIT.java
```diff
@@ -369,18 +523,97 @@ public void testCount() throws Exception {
...
+                    + indexTableName + " on " + dataTableName + " (val1)");
+            String selectSql = "SELECT id from " + dataTableName + " WHERE val1  = 'ab'";
+
+            // Verify that we will read from the index table
+            assertExplainPlan(conn, selectSql, dataTableName, indexTableName);
+            ResultSet rs = conn.createStatement().executeQuery(selectSql);
+            assertTrue(rs.next());
+            assertEquals("a", rs.getString(1));
+            assertFalse(rs.next());
+            TestUtil.dumpTable(conn, TableName.valueOf(indexTableName));
```
gjacoby126, line 586: nit: are these TestUtil.dumpTable calls still necessary?

gjacoby126 (commented):
First pass. Mostly nits

phoenix-core/src/it/java/org/apache/phoenix/end2end/index/UncoveredGlobalIndexRegionScannerIT.java
```diff
@@ -80,7 +93,39 @@ private void populateTable(String tableName) throws Exception {
...
+            String indexTableName = generateUniqueName();
+            conn.createStatement().execute("create table " + dataTableName +
+                    " (id varchar(10) not null primary key, val1 varchar(10), val2 varchar(10), val3 varchar(10))");
+            if (uncovered) {
+                // The INCLUDE clause should not be allowed
+                try {
+                    conn.createStatement().execute("CREATE UNCOVERED INDEX " + indexTableName
+                            + " on " + dataTableName + " (val1) INCLUDE (val2)");
+                    Assert.fail();
+                } catch (Exception e) {
```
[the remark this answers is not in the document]
kadirozde: Yes

phoenix-core/src/it/java/org/apache/phoenix/end2end/index/UncoveredGlobalIndexRegionScannerIT.java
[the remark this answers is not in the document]
kadirozde: I defined an uncovered index as the index never covers a query. That means the index rows are always joined with the data table rows (inner join). This restriction allows optimizing the write path for index updates such that we can update an uncovered index in one phase (instead of using two phase commit) and do not need to retrieve data table rows when the data table mutations also includes the index columns. Indexes before this feature are called (defined as) covered indexes. A covered index can cover a query if all the referenced columns are the columns of (i.e., covered by) the index table. A covered index can be also used for queries that include uncovered columns, using index hints whereas uncovered indexes do not require hints. The write path for local indexes uses one phase update currently but the data table rows are read during updates. In theory, we can introduce the uncovered local indexes to eliminate the read operations during write transactions when mutations include all index columns but we need to introduce index repair during read. I suggest that we can reconsider this in future when the need arises. Please note when we redesigned indexes we preserved the implementation of local indexes as is since the old code was very difficult to crack. Adding changes for uncovered local indexes would increase the complexity of this jira which is already complex enough. Please let me know what you think about my reasoning.

phoenix-core/src/it/java/org/apache/phoenix/end2end/index/UncoveredGlobalIndexRegionScannerIT.java
kadirozde: Good catch. Those were added for debugging. I will remove them.

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/UncoveredIndexRegionScanner.java
```diff
@@ -162,7 +174,7 @@ protected Scan prepareDataTableScan(Collection<byte[]> dataRowKeys) throws IOExc
         scanRanges.initializeScan(dataScan);
         SkipScanFilter skipScanFilter = scanRanges.getSkipScanFilter();
         dataScan.setFilter(new SkipScanFilter(skipScanFilter, false));
-        scan.setAttribute(BaseScannerRegionObserver.SERVER_PAGE_SIZE_MS,
+        dataScan.setAttribute(BaseScannerRegionObserver.SERVER_PAGE_SIZE_MS,
```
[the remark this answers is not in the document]
kadirozde: Yes, it was a bug.

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/UncoveredIndexRegionScanner.java
```diff
@@ -244,20 +256,62 @@ protected boolean scanIndexTableRows(List<Cell> result,
...
+        for (Cell cell : dataRow.rawCells()) {
+            put.add(cell);
+        }
+        if (indexMaintainer.checkIndexRow(indexRowKey, put)) {
+            if (IndexUtil.getMaxTimestamp(put) != ts) {
+                Mutation[] mutations;
+                Put indexPut = new Put(indexRowKey);
+                indexPut.addColumn(emptyCF, emptyCQ, ts, QueryConstants.VERIFIED_BYTES);
+                if ((EnvironmentEdgeManager.currentTimeMillis() - ts) > ageThreshold) {
+                    Delete indexDelete = indexMaintainer.buildRowDeleteMutation(indexRowKey,
```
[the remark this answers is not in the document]
kadirozde: We need make sure that index and data table rows have the same timestamp for time range queries. That is way we need to delete the index row version with incorrect timestamps. Please note as part of the repair operation, we also insert the correct version. During the write path, we update the index row before the data row. If the data row update fails, the index row timestamp will be higher than the data row timestamp. If compaction happens before the index row is read, then the correct version would be removed by compaction. That is the reason we add the correct version during repair.

phoenix-core/src/main/java/org/apache/phoenix/index/IndexMaintainer.java
kadirozde: okay

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/index/IndexUpgradeTool.java
kadirozde: Okay

phoenix-core/src/main/java/org/apache/phoenix/schema/PTable.java
kadirozde: I will change the enum to UNCOVERED_GLOBAL as you suggested.

kadirozde:
@tkhurana, @gjacoby126, please let me know if you have further review comments. I think I addressed or replied to all your comments. Thanks

gjacoby126 (approved):
Thanks for the changes, @kadirozde

gjacoby126:
Kicked off a new PR build to check the Yetus results

tkhurana (approved):

PR #1538, opened 2022-12-01
PHOENIX-6841 Depend on omid-codahale-metrics

stoty:


richardantal:
I think we should add omid-codahale-metrics to the list of ignoredUnusedDeclaredDependencies too

richardantal:
Just realised that we have a line this line to cover everything under org.apache.omid:
https://github.com/apache/phoenix/blob/master/pom.xml#L369

stoty:
Not needed for master, but we don't have the wildcard ignore on 5.1.
I have updated the 5.1 pull request, thank you.

stoty:
Thanks, I'll let the 5.1 PR with the changes complete, then push it.

PR #1541, opened 2022-12-05
PHOENIX-5980 MUTATION_BATCH_FAILED_SIZE metric is incorrectly updated for failing delete mutations

mnpoonia:


mnpoonia:
@shahrs87 Please review when you have time.

mnpoonia:
Thanks @shahrs87 . Finally i was able to reproduce the bug for transactional table. Have not looked into the solution as of now but updated the test to repro the issue. 😄

mnpoonia:
@shahrs87 Looks like there a reason that we only return 1 for each batch for transactional tables. When it is transactional tables they consider it as one mutation as a whole. But if we really wanted to get number of deletes that were part of batch we need to fetch it using 
`DELETE_BATCH_FAILED_SIZE` 
instead of 
`MUTATION_BATCH_FAILED_SIZE`

mnpoonia:
On another hand it looks like it is working ine because in transaction table one batch is one transaction so one transaction(think mutation) failed but if we want to know number of delete operations within that mutatiion we can check through 
DELETE_BATCH_FAILED_SIZE

phoenix-core/src/it/java/org/apache/phoenix/end2end/MutationBatchFailedStateMetricIT.java
```diff
@@ -0,0 +1,146 @@
...
+
+            Connection finalConn = conn;
+            boolean autoCommit = conn.getAutoCommit();
+            System.out.println("Auto commit : " + autoCommit);
+            Statement stmt = conn.createStatement();
+
+            TestUtil.addCoprocessor(conn, deleteTableName,
+                ImmutableIndexIT.DeleteFailingRegionObserver.class);
+
+            String dml = String.format("DELETE FROM %s where val1 >  1", deleteTableName);
```
shahrs87, line 99: Adding a comment in the code helps. Something like this will delete 4 out of 5 rows from the table.

phoenix-core/src/it/java/org/apache/phoenix/end2end/MutationBatchFailedStateMetricIT.java
```diff
@@ -0,0 +1,146 @@
...
+                ImmutableIndexIT.DeleteFailingRegionObserver.class);
+
+            rowCount = TestUtil.getRowCount(conn, deleteTableName);
+            System.out.println("Row Count:" + rowCount);
+        } catch (SQLException e) {
+            Map<String, Map<MetricType, Long>> mutationMetrics =
+                    conn.unwrap(PhoenixConnection.class).getMutationMetrics();
+            int i = mutationMetrics.get(deleteTableName).get(MUTATION_BATCH_FAILED_SIZE).intValue();
+            System.out.println("mfs = " + i);
+            Assert.assertEquals(4, i);
```
shahrs87, line 113: Shouldn't this value be 1 if the metric is incorrectly updated?

phoenix-core/src/it/java/org/apache/phoenix/end2end/MutationBatchFailedStateMetricIT.java
```diff
@@ -0,0 +1,146 @@
...
+            conn = DriverManager.getConnection(getUrl(), props);
+            long rowCount = TestUtil.getRowCount(conn, deleteTableName);
+            System.out.println("Row Count:" + rowCount);
+
+            Connection finalConn = conn;
+            boolean autoCommit = conn.getAutoCommit();
+            System.out.println("Auto commit : " + autoCommit);
+            Statement stmt = conn.createStatement();
+
+            TestUtil.addCoprocessor(conn, deleteTableName,
```
shahrs87, line 96: How does this co proc work? I don't see `preBatchMutate` method being invoked anywhere in `MutationState#sendMutations` method. Maybe I am missing something.

phoenix-core/src/it/java/org/apache/phoenix/end2end/MutationBatchFailedStateMetricIT.java
mnpoonia: Yes it should be but only for transactional tables. Updated  the test with right assert

phoenix-core/src/it/java/org/apache/phoenix/end2end/MutationBatchFailedStateMetricIT.java
mnpoonia: This is not called  directly from sendMutation. It is part of coprocessor. It is called on server side before applying mutation at a region level
[dev docs](https://docs.cloudera.com/HDPDocuments/HDP2/HDP-2.5.0/bk_hbase-java-api-reference/org/apache/hadoop/hbase/coprocessor/BaseRegionObserver.html#preBatchMutate-org.apache.hadoop.hbase.coprocessor.ObserverContext-org.apache.hadoop.hbase.regionserver.MiniBatchOperationInProgress-)

phoenix-core/src/it/java/org/apache/phoenix/end2end/MutationBatchFailedStateMetricIT.java
```diff
@@ -0,0 +1,162 @@
...
+    @Test
+    public void testFailedDelete() throws SQLException {
+        Properties props = PropertiesUtil.deepCopy(TEST_PROPERTIES);
+        Connection conn = null;
+        try {
+            conn = DriverManager.getConnection(getUrl(), props);
+            long rowCount = TestUtil.getRowCount(conn, deleteTableName);
+            System.out.println("Row Count:" + rowCount);
+
+            Connection finalConn = conn;
```
virajjasani, line 98: nit: looks like `finalConn` not required?

virajjasani (commented):
Left few minor comments

phoenix-core/src/it/java/org/apache/phoenix/end2end/MutationBatchFailedStateMetricIT.java
```diff
@@ -0,0 +1,162 @@
...
+            boolean autoCommit = conn.getAutoCommit();
+            System.out.println("Auto commit : " + autoCommit);
```
virajjasani, line 100: Good to replace with slf4j logger#info

phoenix-core/src/it/java/org/apache/phoenix/end2end/MutationBatchFailedStateMetricIT.java
```diff
@@ -0,0 +1,162 @@
...
+                                    + "'")
+                            : "");
+            System.out.println(create_table);
+        }
+        createTables();
+        populateTables();
+    }
+
+    @BeforeClass
+    public static synchronized void doSetup() throws Exception {
```
tkhurana, line 66: Since you are modifying server props you should use the `NeedsOwnMiniClusterTest` category because the test driver object is defined at the jvm level and the ParallelStatsDisabledTest allows reusing the jvm for multiple tests so if some earlier test initializes the driver object it won't be set again. [see here](https://github.com/apache/phoenix/blob/master/phoenix-core/src/test/java/org/apache/phoenix/query/BaseTest.java#L522)

phoenix-core/src/it/java/org/apache/phoenix/end2end/MutationBatchFailedStateMetricIT.java
```diff
@@ -0,0 +1,162 @@
...
+ * BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License
+ * for the specific language governing permissions and limitations under the License.
+ */
+package org.apache.phoenix.end2end;
+
+import static org.apache.phoenix.monitoring.MetricType.DELETE_BATCH_FAILED_SIZE;
+import static org.apache.phoenix.monitoring.MetricType.MUTATION_BATCH_FAILED_SIZE;
+import static org.apache.phoenix.util.TestUtil.TEST_PROPERTIES;
+
+import java.sql.*;
```
tkhurana, line 17: nit: Don't use wildcard imports

phoenix-core/src/it/java/org/apache/phoenix/end2end/MutationBatchFailedStateMetricIT.java
tkhurana: Another option is to add your test to PartialCommitIT which seems to define similar properties as your setup function is doing.

phoenix-core/src/it/java/org/apache/phoenix/end2end/MutationBatchFailedStateMetricIT.java
mnpoonia: forgot to clean it. Done now.

phoenix-core/src/it/java/org/apache/phoenix/end2end/MutationBatchFailedStateMetricIT.java
mnpoonia: I think i would remove these now

mnpoonia:
Thanks for the review @tkhurana and @virajjasani . Have updated the PR with your suggestions where applicable.

phoenix-core/src/it/java/org/apache/phoenix/end2end/MutationBatchFailedStateMetricIT.java
virajjasani: Looks like final commit push is pending :)

virajjasani (approved):

virajjasani:
Looks good to go. @tkhurana anything from your side?

phoenix-core/src/it/java/org/apache/phoenix/end2end/MutationBatchFailedStateMetricIT.java
```diff
@@ -0,0 +1,159 @@
...
+
+    public MutationBatchFailedStateMetricIT(String transactionProvider) {
+        this.transactional = transactionProvider != null;
+        if (this.transactional) {
+            create_table =
+                    create_table + (this.transactional
+                            ? (" TRANSACTIONAL=true,TRANSACTION_PROVIDER='" + transactionProvider
+                                    + "'")
+                            : "");
+            System.out.println(create_table);
```
tkhurana, line 63: remove this system.out.println

tkhurana:
@mnpoonia one minor comment. Other than that looks good

mnpoonia:
@tkhurana Done

tkhurana (approved):

tkhurana:
@mnpoonia Just to make sure my understanding is right, the current behavior is correct ?

shahrs87:
> When it is transactional tables they consider it as one mutation as a whole. But if we really wanted to get number of deletes that were part of batch we need to fetch it using
DELETE_BATCH_FAILED_SIZE
instead of
MUTATION_BATCH_FAILED_SIZE

I think we need to use DELETE_BATCH_FAILED_SIZE, right? That was the reason this JIRA was created in the first place. Am I understanding something wrong? Table being transactional doesn't have to do anything with how many deletes failed in that batch. Both are orthogonal. We tried to delete 4 rows and that failed. So the metric should reflect how many rows we tried to delete and failed. 

> We should fix the logic for deletes/mixed delete + upsert mutation batch failures.

The above line from the JIRA description. I don't understand what the author meant by mixed delete + upsert mutation batch failures. Does he mean if we perform delete and upsert mutation in 1 transaction, it will update the metric just by 1 also? Also what is the behavior if UPSERT SELECT queries are run? 
@mnpoonia  @virajjasani @tkhurana

virajjasani:
> I think we need to use DELETE_BATCH_FAILED_SIZE, right? That was the reason this JIRA was created in the first place.

I am bit confused. Chinmay's description on the jira did not have any mention of transactional usecase.

phoenix-core/src/it/java/org/apache/phoenix/end2end/MutationBatchFailedStateMetricIT.java
```diff
@@ -0,0 +1,158 @@
...
+                    mutationMetrics.get(deleteTableName).get(MUTATION_BATCH_FAILED_SIZE).intValue();
+            int dbfs =
+                    mutationMetrics.get(deleteTableName).get(DELETE_BATCH_FAILED_SIZE).intValue();
+            if (transactional) {
+                Assert.assertEquals(4, dbfs);
+                Assert.assertEquals(1, mfs);
+            } else {
+                Assert.assertEquals(4, dbfs);
+                Assert.assertEquals(4, mfs);
+            }
```
virajjasani, line 127: Also, `DELETE_BATCH_FAILED_SIZE` already seems correct.
Here, in both if and else blocks, we have same value (4) being asserted for `DELETE_BATCH_FAILED_SIZE`.

virajjasani:
PR title to be updated based on the final patch.

virajjasani:
>> We should fix the logic for deletes/mixed delete + upsert mutation batch failures.

> The above line from the JIRA description. I don't understand what the author meant by mixed delete + upsert mutation batch failures. Does he mean if we perform delete and upsert mutation in 1 transaction, it will update the metric just by 1 also? Also what is the behavior if UPSERT SELECT queries are run?

Yeah these 2 points are worth testing.

mnpoonia:
>I think we need to use DELETE_BATCH_FAILED_SIZE, right? That was the reason this JIRA was created in the first place. Am I understanding something wrong? Table being transactional doesn't have to do anything with how many deletes failed in that batch. Both are orthogonal. We tried to delete 4 rows and that failed. So the metric should reflect how many rows we tried to delete and failed.

@shahrs87 You are right. table being transactional or not doesn't change the value of DELETE_BATCH_FAILED_SIZE. But when it  comes to  mutation  while we are in  transactional table we consider a set of command as  single mutation and that can be only reason i consider that MUTATION_BATCH_FAILED_SIZE differs.

So in that sense it looks fine. What do you think we are missing here?

shahrs87:
> I am bit confused. Chinmay's description on the jira did not have any mention of transactional usecase.
> In the conn.commit() path, we get the number of mutations that failed to be committed in the catch block of MutationState.sendMutations() (see [here](https://github.com/apache/phoenix/blob/dcc88af8acc2ba8df10d2e9d498ab3646fdf0a78/phoenix-core/src/main/java/org/apache/phoenix/execute/MutationState.java#L1195-L1198)).


In the jira description, he added link to the above code.He referred to `getUncommittedStatementIndexes` method which internally updates `uncommittedStatementIndexes` array. The only place where uncommittedStatementIndexes is being updated is `addUncommittedStatementIndexes` which is called from `MutationState#send` method iff table is transactional. See code [here](https://github.com/apache/phoenix/blob/dcc88af8acc2ba8df10d2e9d498ab3646fdf0a78/phoenix-core/src/main/java/org/apache/phoenix/execute/MutationState.java#L972). So I assumed that this jira is for transactional table only. @virajjasani

shahrs87:
> But when it comes to mutation while we are in transactional table we consider a set of command as single mutation and that can be only reason i consider that MUTATION_BATCH_FAILED_SIZE differs.

When you look at the metrics independently, it will look like only 1 mutation failed in the batch even if 100 mutations have failed. We don't have different metrics for transactional table and non-transactional table and we shouldn't also. If a table is transactional we want to revert the state to original state if any of the mutations failed but why we want to treat metrics differently? @mnpoonia

mnpoonia:
@shahrs87 I think now i understand what you are retrying to say. Now that i look at the metric name again i think we should not consider anything special about this metric if table is tx or non tx. If we ever need special cases we can create special metric with some prefix or suffix indicating it is for tx tables. 

Also  if  we look at the code again  ideally a delete statement should not be part of a batch else(we can talk about it some other day). The problem with deletes combing with upsert of having multiple  delete in same batch is as below
- If we have  a batch with two delete statements and if one statement failed we will get a number saying that x num of deletes failed. But we don't know if it was first delete stmt or second or both failed. Maybe there is a smart way to figure out which i will look in code.

mnpoonia:
@shahrs87 i have tried solving the discrepancy. need your review to see if this makes sense
@tkhurana @virajjasani Nice if you could also skim through and provide feedback.

phoenix-core/src/it/java/org/apache/phoenix/end2end/MutationBatchFailedStateMetricIT.java
```diff
@@ -0,0 +1,212 @@
...
+                    mutationMetrics.get(deleteTableName).get(MUTATION_BATCH_FAILED_SIZE).intValue();
+            int dbfs =
+                    mutationMetrics.get(deleteTableName).get(DELETE_BATCH_FAILED_SIZE).intValue();
+            if (transactional) {
+                Assert.assertEquals(4, dbfs);
+                Assert.assertEquals(5, mfs);
+            } else {
+                Assert.assertEquals(4, dbfs);
+                Assert.assertEquals(5, mfs);
+            }
```
virajjasani, line 181: We no longer need if/else

phoenix-core/src/it/java/org/apache/phoenix/end2end/MutationBatchFailedStateMetricIT.java
```diff
@@ -0,0 +1,212 @@
...
+                    mutationMetrics.get(deleteTableName).get(MUTATION_BATCH_FAILED_SIZE).intValue();
+            int dbfs =
+                    mutationMetrics.get(deleteTableName).get(DELETE_BATCH_FAILED_SIZE).intValue();
+            if (transactional) {
+                Assert.assertEquals(4, dbfs);
+                Assert.assertEquals(4, mfs);
+            } else {
+                Assert.assertEquals(4, dbfs);
+                Assert.assertEquals(4, mfs);
+            }
```
virajjasani, line 134: Same as above

phoenix-core/src/it/java/org/apache/phoenix/end2end/MutationBatchFailedStateMetricIT.java
```diff
@@ -0,0 +1,212 @@
...
+            // it is called before applying mutation to region
+            TestUtil.addCoprocessor(conn, deleteTableName,
+                ImmutableIndexIT.DeleteFailingRegionObserver.class);
+            // trying to delete 4 rows with this single delete statement
+            String dml = String.format("DELETE FROM %s where val1 >  1", deleteTableName);
+            stmt.execute(dml);
+            conn.commit();
+
+            TestUtil.removeCoprocessor(conn, deleteTableName,
+                ImmutableIndexIT.DeleteFailingRegionObserver.class);
```
virajjasani, line 118: We can remove this as it will not likely be executed for successful test scenario.

phoenix-core/src/it/java/org/apache/phoenix/end2end/MutationBatchFailedStateMetricIT.java
```diff
@@ -0,0 +1,212 @@
...
+            PreparedStatement preparedStatement = conn.prepareStatement(upsertSQL);
+            preparedStatement.setString(1, "ROW_6");
+            preparedStatement.setInt(2, 6);
+            preparedStatement.setInt(3, 12);
+            preparedStatement.execute();
+
+            conn.commit();
+
+            TestUtil.removeCoprocessor(conn, deleteTableName,
+                    ImmutableIndexIT.DeleteFailingRegionObserver.class);
```
virajjasani, line 165: Same here

phoenix-core/src/main/java/org/apache/phoenix/execute/MutationState.java
```diff
@@ -1572,12 +1572,19 @@ public static MutationMetricQueue.MutationMetric updateMutationBatchFailureMetri
                 numDeleteMutationsInBatch++;
             }
         }
+
+        long totalfailedMutation = numUpsertMutationsInBatch + numDeleteMutationsInBatch;
+        //TODO this case is something that should not happen
+        // but still if connditioon makes sese if this ever happens
+        if (totalfailedMutation < numFailedMutations) {
+            totalfailedMutation = numFailedMutations;
+        }
```
virajjasani, line 1581: As per this comment, we will likely not hit the case where `totalfailedMutation < numFailedMutations`?

mnpoonia: That is right. That is the understanding.

mnpoonia: Added warning log there.

mnpoonia:
@virajjasani FYI.  Updated code with suggestions

virajjasani (approved):

virajjasani:
@shahrs87 @tkhurana do you have bandwidth to take a look again?

mnpoonia:
Thanks @virajjasani

PR #1545, opened 2022-12-13
DOCS - grant and revoke command rule definition fix

mlorek:
The current rule is wrong as it permits statements such as:

GRANT 'R' ON SCHEMA schema table TO 'user'

also it does not mention tableRef rule which is listed on one of the examples

### DIFF
--- a/docs/phoenix.csv
+++ b/docs/phoenix.csv
@@ -372,7 +372,7 @@ DROP SCHEMA my_schema
 "
 
 "Commands","GRANT","
-GRANT {permissionString} [ON [[SCHEMA schemaName] | tableRef] ] TO [GROUP] userString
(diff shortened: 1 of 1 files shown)


stoty:
This is still not correct.
I've run some qucik test, and you need specify EITHER
SCHEMA schemaName
or
tableRef

SCHEMA schemaName tableRef throws an error.

Can you also fix this in the docs ?

stoty (changes requested):
See my comment above

stoty:
Thank you.
I will commit this, but unfortunately this won't appear on the website until we fix our site build process

stoty:
Thank you @mlorek .

stoty:
This was tracked and commit with the following JIRA:

https://issues.apache.org/jira/browse/PHOENIX-6848

PR #1546, opened 2022-12-20
PHOENIX-6052 Include syscat time in mutation time

mnpoonia:


mnpoonia:
@virajjasani @tkhurana a draft PR for  the fix.

tkhurana:
@mnpoonia would it make sense to introduce a separate metric to track the syscat time ? By clubbing the two metrics in one we will not know where we are spending time.

mnpoonia:
@tkhurana i think it  is upto us. Ultimately we care about how much time one mutation took. Now if we want  to divide it in syscat and then htable.batch it is perfectly fine but not sure if we want to introduce a new metric.
As per the discussion in JIRA https://issues.apache.org/jira/browse/PHOENIX-6052 It looked like we are not in favor  of new metric so i went ahead with this approach. 
@virajjasani and @shahrs87 what do you guys think?

mnpoonia:
@tkhurana do you mind if we include syscat time in the mutation commit time in this jira and create a seperate jira to track syscat time seperately as well.

virajjasani:
> create a seperate jira to track syscat time seperately as well

I think this should be good. Thoughts @gjacoby126 @tkhurana?

phoenix-core/src/main/java/org/apache/phoenix/execute/MutationState.java
```diff
@@ -1297,12 +1302,11 @@ private void sendMutations(Iterator<Entry<TableInfo, List<Mutation>>> mutationsI
             int retryCount = 0;
             boolean shouldRetry = false;
             long numMutations = 0;
-            long mutationSizeBytes = 0;
+            //long mutationSizeBytes = 0;
```
tkhurana, line 1305: nit: remove the comment

tkhurana:
@mnpoonia You have changed the scope of this metric. Let us say you are batching 1000 upserts. Phoenix internally in the `sendMutations` function groups them into sub batches of by default 100 and then calls `htable.batch()` on that sub batch. Earlier the metric was capturing the time taken by each individual sub-batch. Now the metric is capturing the time across all sub-batches.  I am not sure if this is intended.

mnpoonia:
@tkhurana Thanks for pointing that out. I completely missed that part. Have reverted to old behaviour of mini batch mutation time. Also created a new metric for syscat time. 
Will need your and @virajjasani  help in getting it to closure. Thanks in advance

phoenix-core/src/main/java/org/apache/phoenix/execute/MutationState.java
```diff
@@ -1198,6 +1199,7 @@ private void sendBatch(Map<TableRef, MultiRowMutationState> commitBatch, long[]
         Map<TableInfo, List<Mutation>> physicalTableMutationMap = Maps.newLinkedHashMap();
 
         // add tracing for this operation
+        long startTime = EnvironmentEdgeManager.currentTimeMillis();
```
gourabtaparia, line 1202: nitpick : the above L 1201 "add tracing" comment was for line L 1203 - maybe we can add your change just above that comment - to make the comment and code nearby as last time.

phoenix-core/src/main/java/org/apache/phoenix/execute/MutationState.java
```diff
@@ -1411,8 +1415,8 @@ public List<Mutation> getMutationList() {
                     child.stop();
                     child.stop();
                     shouldRetry = false;
-                    mutationCommitTime = EnvironmentEdgeManager.currentTimeMillis() - startTime;
-                    GLOBAL_MUTATION_COMMIT_TIME.update(mutationCommitTime);
+//                    mutationCommitTime = EnvironmentEdgeManager.currentTimeMillis() - startTime;
```
gourabtaparia, line 1418: Why this code commented ?  Please fix this

phoenix-core/src/main/java/org/apache/phoenix/monitoring/MetricType.java
```diff
@@ -208,23 +209,23 @@ private MetricType(String shortName, String description, LogLevel logLevel, PDat
     public String description() {
         return description;
     }
-
```
gourabtaparia, line 211: nitpick : there are various changes fixing the extra spaces not part of the change - should we skip this as part of this change ?

gourabtaparia (commented):
Few comments. @mnpoonia

phoenix-core/src/main/java/org/apache/phoenix/execute/MutationState.java
```diff
@@ -1214,6 +1216,8 @@ private void sendBatch(Map<TableRef, MultiRowMutationState> commitBatch, long[]
                     serverTimeStamps == null ?
                         validateAndGetServerTimestamp(tableRef, multiRowMutationState) :
                         serverTimeStamps[i++];
+                long syscatTime = EnvironmentEdgeManager.currentTimeMillis() - startTime;
```
tkhurana, line 1219: @mnpoonia This is not the only place where we call validateAndGetServerTimestamp. It is also called from here https://github.com/apache/phoenix/blob/master/phoenix-core/src/main/java/org/apache/phoenix/execute/MutationState.java#L895

tkhurana: Also, this is being called in a loop so you need to reset the value of startTime for every iteration.

tkhurana (changes requested):

phoenix-core/src/main/java/org/apache/phoenix/execute/MutationState.java
mnpoonia: Agreed. instead of keeping the time here i pushed it down to the method validateAndGetServerTimestamp.

mnpoonia: @tkhurana Let me know  if it looks good now.

phoenix-core/src/main/java/org/apache/phoenix/execute/MutationState.java
```diff
@@ -1480,8 +1500,10 @@ public List<Mutation> getMutationList() {
                         GLOBAL_MUTATION_INDEX_COMMIT_FAILURE_COUNT.update(numFailedPhase3Mutations);
                     }
                 } finally {
-                    mutationCommitTime = EnvironmentEdgeManager.currentTimeMillis() - startTime;
-                    GLOBAL_MUTATION_COMMIT_TIME.update(mutationCommitTime);
+                    if (mutationCommitTime != 0) {
```
tkhurana, line 1503: Didn't understand this check

phoenix-core/src/main/java/org/apache/phoenix/execute/MutationState.java
```diff
@@ -947,20 +963,24 @@ private long validateAndGetServerTimestamp(TableRef tableRef, MultiRowMutationSt
...
-                        NUM_METADATA_LOOKUP_FAILURES, 1);
+        } catch (Throwable e) {
+            if (table != null) {
+                TableMetricsManager.updateMetricsForSystemCatalogTableMethod(
+                    table.getTableName().toString(), NUM_METADATA_LOOKUP_FAILURES, 1);
             }
             throw e;
+        } finally {
+            long endTime = EnvironmentEdgeManager.currentTimeMillis();
+            GLOBAL_MUTATION_SYSCAT_TIME.update(endTime - startTime);
```
tkhurana, line 980: This looks good

phoenix-core/src/main/java/org/apache/phoenix/execute/MutationState.java
mnpoonia: @tkhurana this was leftover of previous commits. Cleaned it.

mnpoonia:
@virajjasani @tkhurana I have addressed most of the review comments. Please have a look now.

tkhurana (approved):

phoenix-core/src/main/java/org/apache/phoenix/monitoring/MetricType.java
virajjasani: Agree, good to keep only addition of `MUTATION_SYSCAT_TIME` for this patch

phoenix-core/src/main/java/org/apache/phoenix/execute/MutationState.java
```diff
@@ -1411,8 +1431,6 @@ public List<Mutation> getMutationList() {
                     child.stop();
                     child.stop();
                     shouldRetry = false;
-                    mutationCommitTime = EnvironmentEdgeManager.currentTimeMillis() - startTime;
-                    GLOBAL_MUTATION_COMMIT_TIME.update(mutationCommitTime);
```
virajjasani, line 1415: Nice. After [PHOENIX-6397](https://issues.apache.org/jira/browse/PHOENIX-6397) looks like we were updating this twice (same logic in `finally` as well).

mnpoonia:
@virajjasani I have removed all the whitespace and indentation changes. Please review now.
@tkhurana FYI

mnpoonia:
PR for 5.1 branch
https://github.com/apache/phoenix/pull/1557

virajjasani (approved):

PR #1547, opened 2022-12-21
PHOENIX-6849 git_jira_fix_version_check.py does not retrieve all JIRA…

tkhurana:
…s for a fixed version

Set maxResults to None so that the API does internal batching and returns back all the results

### DIFF
--- a/dev/misc_utils/git_jira_fix_version_check.py
+++ b/dev/misc_utils/git_jira_fix_version_check.py
@@ -91,11 +91,15 @@ print('Jira/Git commit message diff completed: #################################
 
 print('\nAny resolved Jira with fixVersion ' + fix_version
       + ' but corresponding commit not present')
-print('Starting diff: ##############################################')
+
+# maxResults=None results in internal batching within the API 
+# and it returns all the results back
 all_issues_with_fix_version = jira.search_issues(
     'project=' + jira_project_name + ' and status in (Resolved,Closed) and fixVersion='
-    + fix_version)
+    + fix_version, maxResults=None)
+print(f'Found {all_issues_with_fix_version.total} issues fixed in {fix_version}')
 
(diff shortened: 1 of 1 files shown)


dev/misc_utils/git_jira_fix_version_check.py
```diff
@@ -91,11 +91,15 @@
...
       + ' but corresponding commit not present')
-print('Starting diff: ##############################################')
+
+# maxResults=None results in internal batching within the API 
+# and it returns all the results back
 all_issues_with_fix_version = jira.search_issues(
     'project=' + jira_project_name + ' and status in (Resolved,Closed) and fixVersion='
-    + fix_version)
+    + fix_version, maxResults=None)
+print(f'Found {all_issues_with_fix_version.total} issues fixed in {fix_version}')
```
gourabtaparia, line 100: minor comment : Looks like the formatted string literal is supported from python3.6 onwards - https://docs.python.org/3/whatsnew/3.6.html#whatsnew36-pep498. 

Is there any minimum python version support wrt the phoenix repo where the current script should run ? As this might not be backward compatible with lower versions.

gourabtaparia (commented):
LGTM @tkhurana . One minor question :

There are few auto precheck failures - please take a look there.

dev/misc_utils/git_jira_fix_version_check.py
tkhurana: @gourabtaparia This script only works with python3. Python 3.6 is EOL so we should be fine.

stoty (approved):
+1 LGTM

As this is only run by the RM, Python version compatibility is no a priority (unlike with the startup scripts)

stoty:
nit: please fix the pylint warning before commit

PR #1548, opened 2023-01-02
PHOENIX-6851 Add spotless to format code

mnpoonia:


mnpoonia:
Yetus checks seems to fail for a comment which is a url. I am hoping that is fine. @virajjasani

virajjasani:
HBase also did something like one time format of the entire codebase. Ever since then, every PR merge follows spotless way of format to maintain same format. Do we want to start using spotless after this commit (without touching current codebase)?

FYI @stoty you might also have some thoughts.

mnpoonia:
I am planning to create a subtask to do a one time formatting for entire codebase. @virajjasani

stoty:
If we do this, then IMO the code cleanup and the check should be committed together.

One hand this would be a major PITA, for backports, etc.
On the other hand the current code looks horrible.

I don't have a strong opinion either way.
I think that this should be discussed on the dev@ list, as this is a very intrusive change.

lfrancke:
There is a new attempt here: https://github.com/apache/phoenix/pull/1995

NihalJain:
Work taken over at https://github.com/apache/phoenix/pull/2024, closing! Thank you @mnpoonia for your effort here!

PR #1549, opened 2023-01-11
PHOENIX-6786 SequenceRegionObserver should use batch mutation coproc hook

apurtell:


apurtell:
@kadirozde Is this the kind of refactoring you had in mind?

This patch:
- Adds a preBatchMutate hook that will filter any Append or Increment operations submitted in batch and handle them ahead of base HBase processing. 
- Retains existing hooks for preAppend and preIncrement that are necessary for intercepting other APIs.
- Refactor most logic to reusable private methods.

PR #1551, opened 2023-01-19
PHOENIX-6776 Abort scans of closed connections at ScanningResultIterator

lokiore:


lokiore:
@mnpoonia `org.apache.phoenix.execute.PartialCommitIT.testUpsertSelectFailure` is failing for the PR which started failing after https://github.com/apache/phoenix/commit/8b688fcbe16f89b2bf2229e2d321b267bc1fcba8, can you please take a look! Thanks

phoenix-core/src/main/java/org/apache/phoenix/iterate/TableResultIterator.java
```diff
@@ -105,6 +105,8 @@ public class TableResultIterator implements ResultIterator {
     private Map<ImmutableBytesPtr,ServerCache> caches;
     private HashCacheClient hashCacheClient;
 
+    private final boolean isConnInContext;
```
kadirozde, line 108: Let's call this isMapReduceContext and update it accordingly

phoenix-core/src/main/java/org/apache/phoenix/iterate/TableResultIterator.java
```diff
@@ -115,6 +111,7 @@ public class TableResultIterator implements ResultIterator {
         this.scanGrouper = null;
         this.caches = null;
         this.retry = 0;
+        this.isMapReduceContext = true;
```
kadirozde, line 114: I think this should be set to false now.

lokiore: Changed it Thanks!
I'll run tests once more locally, pipeline is failing because of test failing above.

lokiore:
@kadirozde, locally also only 1 test is failing which I mentioned above! and I have addressed the comments. Thanks!

kadirozde (approved):
+1, Thanks!

virajjasani:
> @mnpoonia `org.apache.phoenix.execute.PartialCommitIT.testUpsertSelectFailure` is failing for the PR which started failing after [8b688fc](https://github.com/apache/phoenix/commit/8b688fcbe16f89b2bf2229e2d321b267bc1fcba8), can you please take a look! Thanks

Thanks for reporting this @lokiore. I verified this locally and the test failure is relevant to that commit, will revert it.

PR #1552, opened 2023-01-19
PHOENIX-4906  Introduce a master coprocessor to handle cases where merging regions of salted table might cause an issue

mnpoonia:


stoty:
I suggest using a more generic name, like PhoenixMasterObserver.

This will make our lives easier if we ever need to more features to it.

mnpoonia:
@stoty i have changed the names accordingly. Please review when you get some cycles.

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/PhoenixMasterObserver.java
```diff
@@ -0,0 +1,114 @@
...
+        if (schemaName == null || schemaName.isEmpty()) {
+            sql = sql + "   and table_schem is null";
+        } else {
+            sql = sql + String.format(" and table_schem=%s", schemaName);
+        }
+
+        Configuration conf = ctx.getEnvironment().getConfiguration();
+        PreparedStatement stmt;
+        ResultSet resultSet;
+        try (Connection conn = QueryUtil.getConnectionOnServer(conf)) {
```
stoty, line 74: I wonder if we should get the salted status via a PTable object from the CQSI cache instead.
Wouldn't that be faster ?

phoenix-core/src/it/java/org/apache/phoenix/PhoenixMasterObserverIT.java
```diff
@@ -0,0 +1,107 @@
...
+            "CREATE TABLE IF NOT EXISTS %s(ID VARCHAR NOT NULL PRIMARY KEY, VAL1 INTEGER, VAL2 INTEGER) SALT_BUCKETS=4";
+    String indexName = generateUniqueName();
+    String create_index = "CREATE INDEX " + indexName + " ON %s(VAL1 DESC) INCLUDE (VAL2)";
+    String upsertStatement = "UPSERT INTO %s VALUES(?, ?, ?)";
+    String deleteTableName = generateUniqueName();
+
+    @BeforeClass
+    public static synchronized void doSetup() throws Exception {
+        Map<String, String> serverProps = Maps.newHashMapWithExpectedSize(3);
+        serverProps.put("hbase.coprocessor.master.classes", PhoenixMasterObserver.class.getName());
```
stoty, line 46: There is no way to automatically load a Master coprocessor on Phoenix startup, right ?

stoty:
Also, since this is a broad topic, perhaps we could split this feature into its separate JIRA ?

mnpoonia:
I am mostly positive about this change but need help in two things
- How can we check if a table is a phoenix table or not?
- How do we want to add this coprocessor to system? Through hbase-site config(`hbase.coprocessor.master.classes`) or through phoenix code (i am not aware of this path) ?

jpisaac:
Few questions @mnpoonia 
1. Shouldn't this be handled in HBase instead of involving Phoenix, as I understand salting is an HBase feature?

A couple of questions if Phoenix needs to be involved
2. Shouldn't there be some checks if the Phoenix subsystem is present or not?
3. What are the implications on dependency management (since there is a tight coupling here between HBase master services and phoenix) during upgrades and maintenance?

@virajjasani @apurtell your thoughts?

jpisaac:
@mnpoonia Also can we ensure that it is not a Phoenix Bug here in ScanRanges?

stoty:
Salting is not an HBase feature, it's a Phoenix feature (at least as used by Phoenix).

Ideally, HBase had a MergePolicy mechanism simliar to the SplitPolicy mechanism where we could add this logic per table, but it doesn't.
Even if we were to add it, it wouldn't be available before 2.6 or even later.

I admit I that I have never really tracked down why we have this merging restriction, but it seems to be a performance optimization.

phoenix-core/src/it/java/org/apache/phoenix/PhoenixMasterObserverIT.java
virajjasani: I think better to use config for master/regionserver coproc loading. Even if we do figure out letting phoenix update coproc at runtime, pretty sure we would end up using private/unstable APIs that can break across releases.

I was wondering this same point sometime back and hence decided to document this detail that we need to use this coproc by updating hbase config, having no other choice but to document it https://github.com/apache/phoenix/pull/1561#issue-1578828100

stoty:
Yes, the problem is caused by this assumption built into the Phoenix implementation.
If we could remove that without hurting performance, that would be brilliant.
Thank you for digging into it.

mnpoonia:
> @stoty I referring to what was explained here in the [hot-spotting](https://hbase.apache.org/book.html#_hotspotting) section of the hbase book. I think Phoenix just facilitates this under the cover for the user.

> What I was asking is shouldn't we be seeing this issue in a purely HBase table with row keys distributed with salt prefixes?

@jpisaac I am not sure but my understanding is that hbase doesnot have a concept of salting. Clients can use salting to reduce the hotpostitng but then it is upto clients that how they want to manage all aspects of it. HBase doesn't care about salting in case of split or merge.

In phoenix we have this limitation that every row in a single region should have same salt. If we mix(merge) two regions of different salts phoenix scans starting behaving weird.

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/PhoenixMasterObserver.java
```diff
@@ -29,38 +29,51 @@ public Optional<MasterObserver> getMasterObserver() {
...
+        if(phoenixTable == null) {
+            // it is not a phoenix table
+            return false;
+        }
+        int saltBuckets = phoenixTable.getBucketNum();
         System.out.println("Number of buckets="+saltBuckets);
-        if(saltBuckets > 0 ) {
-            System.out.println("Number of buckets="+saltBuckets);
-            return !regionsHaveSameSalt(regionsToMerge);
+        if(saltBuckets <= 0 ) {
```
stoty, line 57: Nit: Having negative salt buckets here points to some serious problem with the tabke.
I'd just check for 0 here.

PR #1560, opened 2023-02-09
PHOENIX-6395 Reusing Connection instance object instead of creating e…

stoty:
…verytime in PhoenixAccessController class

Co-authored-by: vmeka2020 <vmeka@salesforce.com>

### DIFF
--- a/phoenix-core/src/main/java/org/apache/phoenix/coprocessor/PhoenixAccessController.java
+++ b/phoenix-core/src/main/java/org/apache/phoenix/coprocessor/PhoenixAccessController.java
@@ -17,9 +17,17 @@
  */
 package org.apache.phoenix.coprocessor;
 
-import com.google.protobuf.ByteString;
-import com.google.protobuf.RpcCallback;
-import com.google.protobuf.RpcController;
+import java.io.IOException;
+import java.net.InetAddress;
+import java.security.PrivilegedExceptionAction;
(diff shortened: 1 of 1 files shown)


gjacoby126:
Generally looks good -- are there any concerns about what happens if the HBase connection gets into a bad state? Would the operator have to bounce the region server if that happened?

stoty:
The same question occured to me.

However, when we use ServerUtil.ConnectionFactory, we are always getting the same Connection instance (per env), so storing the reference in the coprocessor instance doesn't change that, we just avoid some cache lookups.

We already use ServerUtil.ConnectionFactory in a few places, I don't recall seeing problems stemming from this.

We also use PhoenixConnections on the server side, which implement a similar pattern with the CQSI objects holding Connection objects being cached forever. (IUC we don't check or re-create those either)

stoty:
Thank you @gjacoby126 .

My explanation above has a mistake, there is one connection per ConnectionType, not one per  RegionCoprocessorEnvironment.

The API for ServerUtil.ConnectionFactory suggests that theConnection cache is per RegionCoprocessorEnvironment , but that's not true. The cache is actually a static object, so the the Connection objects are per classloader singletons, just as the CQSI objects.

I was worried that we're creating hundreds of connections with ServerUtil.ConnectionFactory, but that factrory is doing the right thing.

PR #1563, opened 2023-02-10
PHOENIX-6873 Use cached Connection in IndexHalfStoreFileReaderGenerator

stoty:


phoenix-core/src/main/java/org/apache/hadoop/hbase/regionserver/IndexHalfStoreFileReaderGenerator.java
```diff
@@ -90,26 +89,33 @@ public StoreFileReader preStoreFileReaderOpen(ObserverContext<RegionCoprocessorE
...
-                    scvf = new SingleColumnValueFilter(HConstants.CATALOG_FAMILY,
-                        HConstants.SPLITA_QUALIFIER, CompareOperator.EQUAL, RegionInfoUtil.toByteArray(region.getRegionInfo()));
+                    scvf =
+                            new SingleColumnValueFilter(HConstants.CATALOG_FAMILY,
+                                    HConstants.SPLITA_QUALIFIER, CompareOperator.EQUAL,
+                                    RegionInfoUtil.toByteArray(region.getRegionInfo()));
                     scvf.setFilterIfMissing(true);
                 }
-                if(scvf != null) scan.setFilter(scvf);
+                if (scvf != null) scan.setFilter(scvf);
```
gjacoby126, line 118: nit: checkstyle is warning on the 1-line if statement.

gjacoby126: It's also warning on the existing 1-line if statement on 186, which would be good to fix while we're at it.

gjacoby126 (changes requested):
Looks good except for the couple of checkstyle nits.

gjacoby126 (approved):
+1, thanks @stoty

PR #1566, opened 2023-02-21
PHOENIX-6884 Phoenix to use hbase.rpc.read.timeout and hbase.rpc.writ…

kadirozde:
…e.timeout

### DIFF
--- a/phoenix-core/src/main/java/org/apache/phoenix/util/ScanUtil.java
+++ b/phoenix-core/src/main/java/org/apache/phoenix/util/ScanUtil.java
@@ -1295,6 +1295,28 @@ public class ScanUtil {
         }
     }
 
+    public static Long getPageSizeInMs(ReadOnlyProps props) {
+        if (props.getBoolean(QueryServices.PHOENIX_SERVER_PAGING_ENABLED_ATTRIB,
+                QueryServicesOptions.DEFAULT_PHOENIX_SERVER_PAGING_ENABLED)) {
+            long pageSizeMs = props.getInt(QueryServices.PHOENIX_SERVER_PAGE_SIZE_MS, -1);
+            if (pageSizeMs == -1) {
+                // Use the half of the HBase RPC read timeout value as the server page size to
+                // make sure that the HBase region server will be able to send a heartbeat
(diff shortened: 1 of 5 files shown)


kadirozde:
> @kadirozde - Makes sense. Is there anywhere else we should be making this change?

@gjacoby126, I updated the other places where RPC timeout is used. I am done with the changes. Thank you for reviewing it.

tkhurana (changes requested):
@kadirozde IndexUpgradeTool too needs changes

kadirozde:
> @kadirozde IndexUpgradeTool too needs changes

@tkhurana, Thanks! I update it too.

tkhurana (approved):

gjacoby126 (approved):
+1

phoenix-core/src/main/java/org/apache/phoenix/util/ScanUtil.java
```diff
@@ -1295,6 +1295,32 @@ public static void setScanAttributesForPhoenixTTL(Scan scan, PTable table,
         }
     }
 
+    public static Long getRPCReadTimeout(ReadOnlyProps props) {
+        if (props.get(HConstants.HBASE_RPC_READ_TIMEOUT_KEY) != null) {
+            return props.getLong(HConstants.HBASE_RPC_READ_TIMEOUT_KEY,
+                    HConstants.DEFAULT_HBASE_RPC_TIMEOUT);
```
virajjasani, line 1301: Since RPC timeout is in int, we could do this too `(long) props.getInt(HConstants.HBASE_RPC_READ_TIMEOUT_KEY, HConstants.DEFAULT_HBASE_RPC_TIMEOUT)` 

But there doesn't seem to be that much diff in time complexity of Long#parseLong vs Integer#parseInt (specifically for string values that are not int overflows, which is always the case for us). We should be good.

virajjasani: Even hbase makes this confusing, at some places this is being read as int and some places as long, should have kept same to avoid confusions :)

virajjasani (approved):

PR #1568, opened 2023-03-01
PHOENIX-6889 Improve extraction of ENCODED_QUALIFIERs

Aarchy:


Aarchy:
@stoty

phoenix-core/src/main/java/org/apache/phoenix/schema/tool/SchemaExtractionProcessor.java
```diff
@@ -502,25 +504,84 @@ private String getColumnInfoStringForTable(PTable table) {
...
-            return false;
-        }
+        return table.getColumns().size() > 0
+                && !shouldGenerateWithDefaults
+                && table.getType() == PTableType.TABLE
+                && table.getEncodingScheme() != PTable.QualifierEncodingScheme.NON_ENCODED_QUALIFIERS;
+    }
 
-        return table.getType() == PTableType.TABLE && table.getEncodingScheme() != PTable.QualifierEncodingScheme.NON_ENCODED_QUALIFIERS;
+    private boolean isConsecutive(List<Integer> list) {
```
stoty, line 515: We should also check that the first one is the normal starting value, otherwise this won't detect if the 1st column has been dropped.

phoenix-core/src/it/java/org/apache/phoenix/end2end/ShowCreateTableIT.java
```diff
@@ -108,14 +104,42 @@ public void testShowCreateTableDefaultFamily() throws Exception {
...
+        String expected = "CREATE TABLE " + tableName + "(K VARCHAR NOT NULL PRIMARY KEY, " +
+                "\"dF\".INT1 INTEGER, \"dF\".INT2 INTEGER, A.INT3 INTEGER, A.INT4 INTEGER, " +
+                "\"b\".INT5 INTEGER, B.INT6 INTEGER) " +
+                "IMMUTABLE_ROWS=true, DEFAULT_COLUMN_FAMILY='dF'";
+        assertTrue("Expected: :" + expected + "\nResult: " + rs.getString(1),
+                rs.getString(1).equals(expected));
+    }
+
+    @Test
+    public void testShowCreateTableDefaultFamilyDrop() throws Exception {
```
stoty, line 116: something like testShowCreateTableDefaultFamilyNonConsecutive would be more descriptive.
The same goes for the similar tests

phoenix-core/src/main/java/org/apache/phoenix/schema/tool/SchemaExtractionProcessor.java
```diff
@@ -502,25 +504,84 @@ private String getColumnInfoStringForTable(PTable table) {
...
+        for (int i = 1; i < list.size(); i++) {
+            if (list.get(i - 1) + 1 != list.get(i)) {
+                return false;
+            }
+        }
+        return true;
+    }
+
+    private List<String> getNonConsecutiveQualifierCounterFamilies(PTable table) {
```
stoty, line 524: I'd prefer simply getNonConsecutiveQualifierFamilies

stoty: Also for the variable names, use something like idValues, or encodedIds.
Try to avoid the word "counter" when not referring to the actual counter fields in syscat.

phoenix-core/src/main/java/org/apache/phoenix/schema/tool/SchemaExtractionProcessor.java
stoty: Now I see that you check that separately.
You may want to consolide those checks into this function.

stoty (changes requested):

stoty (approved):
+ 1 LGTM

PR #1569, opened 2023-03-01
PHOENIX-6888 Fixing TTL and Max Lookback Issues for Phoenix Tables

kadirozde:


phoenix-core/src/main/java/org/apache/phoenix/coprocessor/BaseScannerRegionObserver.java
```diff
@@ -426,58 +421,39 @@ RegionScanner getWrappedScanner(final ObserverContext<RegionCoprocessorEnvironme
...
-            if (store.getColumnFamilyDescriptor().getTimeToLive() != HConstants.FOREVER) {
-                keepDeletedCells = KeepDeletedCells.TTL;
-            }
-            options.setKeepDeletedCells(keepDeletedCells);
-        }
+        setScanOptionsForFlushesAndCompactions(options);
     }
 
+    @Override
+    public InternalScanner preCompact(ObserverContext<RegionCoprocessorEnvironment> c, Store store,
```
stoty, line 448: This disables the logic for the old indexing in UngroupedAggregateRegionObserver#preCompactScannerOpen .

We should add the capability to disable delete processing fully in StoreCompactionScanner (or subclass it), and 
move that logic to UngroupedAggregateRegionObserver#preCompact() to set  up the Scanner to never remove Deleted cells.

phoenix-core/src/main/java/org/apache/phoenix/filter/PagedFilter.java
```diff
@@ -28,16 +28,21 @@
...
 import org.apache.hadoop.io.Writable;
 import org.apache.phoenix.util.EnvironmentEdgeManager;
+import org.apache.phoenix.util.ScanUtil;
 
 /**
- * This filter overrides the behavior of delegate so that we do not scan more rows than pageSizeInRows .
+ * This is a top level Phoenix filter which injected to a scan at the server side. If the scan has
+ * already a filter then PagedFilter wraps it. There two functions this filter implements: paging
+ * and TTL. The paging function makes sure that the scan does not take more than pageSizeInMs.
+ * The TTL function is for masking expired rows.
```
apurtell, line 45: Would it be a cleaner separation of concerns to stack two filters here, `PagedFilter` implements paging, `TTLFilter` implements masking of expired rows?

apurtell (approved):
Let me conditionally approve these changes and come back to do a deep dive into the new `StoreCompactionScanner` details. At first glace it looks correct but there are a lot of fiddly internal Phoenix details I am not super familiar with. 

It would also be nice to have more unit test coverage for these changes, specifically, the construction of several predicted edge cases that `StoreCompactionScanner` might encounter. 

If I can use a bit of a tortured analogy, to confirm my understanding of the problem, the issue here is similar to in low level programming what we'd call a torn read or write. On an architecture where the hardware does 32 bit wide manipulations, a long long integer of 64 bits must be updated using multiple instructions. If a thread interrupts another thread after the instruction that does a read or write of the lower or upper 32 bit portion but before the instruction that reads or updates the other portion, and does its own updates, and then thread context switches back, the data has been corrupted, so locking must be introduced to ensure the 64 bit update is atomic. We have the same problem here, except Phoenix is implementing a row addressable relational store, where a row is comprised of one or more cells, over an underlying system that is cell addressable. (HBase does provide a limited fiction of a row based view in some API, notably Filter, but is fundamentally a cell addressable store.) So HBase can expire part of a Phoenix row without preserving Phoenix level required semantics, and the Phoenix row addressable data becomes corrupt ("torn").  

Phoenix must take the design principle that it is solely in charge of data retention. It cannot leave it up to HBase level mechanisms like HBase TTL, cell TTL, and HBase level schema cell scoped retention parameters. It must implement the equivalent functionality in a way that maintains data integrity according to its row addressable requirements. 

There is the new concept of _compaction row version_ introduced in `StoreCompactionScanner` which chronologically sorts mutations and tombstones just like HBase would while also accounting for a row based view. In retrospect this is obvious, which is the hallmark of an elegant solution.

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/BaseScannerRegionObserver.java
```diff
@@ -426,58 +421,39 @@ RegionScanner getWrappedScanner(final ObserverContext<RegionCoprocessorEnvironme
...
     }
 
     @Override
     public void preFlushScannerOpen(ObserverContext<RegionCoprocessorEnvironment> c, Store store,
                                                 ScanOptions options, FlushLifeCycleTracker tracker) throws IOException {
-        Configuration conf = c.getEnvironment().getConfiguration();
-        if (isMaxLookbackTimeEnabled(conf)) {
-            setScanOptionsForFlushesAndCompactions(conf, options, store, ScanType.COMPACT_RETAIN_DELETES);
-        }
+        setScanOptionsForFlushesAndCompactions(options);
```
jpisaac, line 430: Should the preFlush also have a specialized scanner like the StoreCompactionScanner (or maybe can be reused)?.
As I understand the preFlush currently will keep all rows, versions and delete markers.

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/BaseScannerRegionObserver.java
```diff
@@ -34,14 +34,7 @@
...
 import org.apache.hadoop.hbase.io.TimeRange;
-import org.apache.hadoop.hbase.regionserver.FlushLifeCycleTracker;
-import org.apache.hadoop.hbase.regionserver.Region;
-import org.apache.hadoop.hbase.regionserver.RegionScanner;
-import org.apache.hadoop.hbase.regionserver.ScanOptions;
-import org.apache.hadoop.hbase.regionserver.ScanType;
-import org.apache.hadoop.hbase.regionserver.ScannerContext;
-import org.apache.hadoop.hbase.regionserver.ScannerContextUtil;
-import org.apache.hadoop.hbase.regionserver.Store;
+import org.apache.hadoop.hbase.regionserver.*;
```
jpisaac, line 37: nit: remove import *

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/BaseScannerRegionObserver.java
kadirozde: Thank you for pointing out this! I will make the necessary changes.

phoenix-core/src/main/java/org/apache/phoenix/filter/PagedFilter.java
kadirozde: I could not separate them. Both functionalities needs to be done at the top level filter. If we choose to skip an expired row under or above the PagedFilter then we cannot ensure paging happens timely. I am open to change the name of the class if you have a good suggestion for this.

kadirozde:
> 

@apurtell, Thank you very much for this insightful comment!

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/BaseScannerRegionObserver.java
kadirozde: Keeping all versions and delete markers is the required behavior for memstore flushes. That is required for indexing consistency and also for max lookback.

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/StoreCompactionScanner.java
```diff
@@ -0,0 +1,422 @@
...
+        this.minVersion = cfd.getMinVersions();
+        this.maxVersion = cfd.getMaxVersions();
+        this.keepDeletedCells = cfd.getKeepDeletedCells();
+        firstStore = region.getStores().get(0).getColumnFamilyName().
+                equals(store.getColumnFamilyName());
+    }
+
+    @Override
+    public boolean next(List<Cell> result) throws IOException {
+        synchronized (storeScanner) {
```
jpisaac, line 78: @kadirozde Can u add some comments as to why we need the synchronization? Thought we always got a new scanner object for every compact run as in here?
https://github.com/apache/hbase/blob/555bf91f2b5e6240350bc6d279f35165d4596634/hbase-server/src/main/java/org/apache/hadoop/hbase/regionserver/compactions/Compactor.java#L354

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/BaseScannerRegionObserver.java
jpisaac: But doesn't running the flush thru the StoreCompactionScanner produce the same outcome, i.e keeping the right row versions taking maxLookback and delete markers and all the other settings into account?

phoenix-core/src/main/java/org/apache/phoenix/filter/PagedFilter.java
jpisaac: Isn't server-side paging a configurable feature?

kadirozde: Yes, that is another good reason for separation.

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/StoreCompactionScanner.java
```diff
@@ -0,0 +1,422 @@
...
+        }
+    }
+
+    private void filterRegionLevel(List<Cell> result) throws IOException {
+        Scan scan = new Scan();
+        scan.setRaw(true);;
+        scan.readAllVersions();
+        scan.setTimeRange(0, compactionTime);
+        byte[] rowKey = result.get(0).getRowArray();
+        scan.withStartRow(rowKey, true);
```
gjacoby126, line 405: calling scan.withStartRow twice? Is the second one supposed to be scan.withStopRow(rowKey, true)?

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/StoreCompactionScanner.java
```diff
@@ -0,0 +1,422 @@
...
+     *            next batch of {@link Cell}s
+     * @throws IOException 
+     */
+    private void filter(List<Cell> result, boolean storeLevel) throws IOException {
+        if (result.isEmpty()) {
+            return;
+        }
+        List<List<Cell>> columns = new ArrayList<>();
+        List<Cell> deleteMarkers = new ArrayList<>();
+        formColumns(result, columns, deleteMarkers);
```
gjacoby126, line 389: Some high-level comments here would be helpful

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/StoreCompactionScanner.java
```diff
@@ -0,0 +1,422 @@
...
+            CompactionRowVersion rowVersion) {
+        if (rowVersion.deleteFamilyMarker == null) {
+            if (firstStore) {
+                if (rowVersion.version < minVersion) {
+                    // Live rows whose version is less than min version are retained outside
+                    // the TTL window
+                    addCells(rowVersion, result);
+                }
+            } else if (rowVersion.version < maxVersion) {
+                // check if this row needs to be retained
```
gjacoby126, line 344: no logic here -- the check is elsewhere? If so please mention

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/StoreCompactionScanner.java
```diff
@@ -0,0 +1,422 @@
...
+                else {
+                    result.add(cell);
+                }
+            }
+        }
+    }
+    private void updateResultsWithRowOutsideMaxLookbackButInsideTTLWindow(List<Cell> result,
+            CompactionRowVersion rowVersion) {
+        // The rows that are outside the max lookback window but inside the TTL window
+        if (rowVersion.deleteFamilyMarker == null) {
```
gjacoby126, line 314: Do we need any logic about minVersions in here, or does the fact that minVersions < maxVersions mean that the rowVersion.version < maxVersion check also takes care of minVersion?

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/StoreCompactionScanner.java
```diff
@@ -0,0 +1,422 @@
...
+        firstStore = region.getStores().get(0).getColumnFamilyName().
+                equals(store.getColumnFamilyName());
+    }
+
+    @Override
+    public boolean next(List<Cell> result) throws IOException {
+        synchronized (storeScanner) {
+            boolean hasMore = storeScanner.next(result);
+            filter(result, true);
+            Collections.sort(result, CellComparator.getInstance());
```
gjacoby126, line 81: Is there anything we to do about sorting to factor in whether the "new" or "old" Cell ordering is in use? (That is, whether Deletes always win at a particular timestamp)

https://hbase.apache.org/book.html#new.version.behavior

gjacoby126 (commented):
As Andrew said, a very elegant solution to a hard problem. Some questions and a potential bug:

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/BaseScannerRegionObserver.java
kadirozde: But then why to run StoreCompactionScanner if it does not make any difference in this case? The timestamp of these rows in memory will be very recent and they will not expire in practice.

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/StoreCompactionScanner.java
kadirozde: I did it for defensive coding. I can remove it.

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/StoreCompactionScanner.java
```diff
@@ -0,0 +1,422 @@
...
+            long ts = getNextRowVersionTimestamp(columns);
+            CompactionRowVersion rowVersion = null;
+            // Form the next row version by picking the first cell from each column if the cell
+            // is not masked by a delete marker
+            for (List<Cell> column : columns) {
+                Cell firstCell = column.remove(0);
+                if (firstCell.getType() == Cell.Type.DeleteFamily ||
+                        firstCell.getType() == Cell.Type.DeleteFamilyVersion) {
+                    if (firstCell.getType() == Cell.Type.DeleteFamily) {
+                        if (firstCell.getTimestamp() >= ttlWindowStart &&
```
jpisaac, line 222: Don't we want to preserve the delete marker if it is inside the maxLookback window even if KEEP_DELETED_CELLS == FALSE?

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/StoreCompactionScanner.java
```diff
@@ -0,0 +1,422 @@
...
+        for (Cell dm : deleteMarkers) {
+            if ((storeLevel ||
+                    Bytes.compareTo(cell.getFamilyArray(), cell.getFamilyOffset(),
+                    cell.getFamilyLength(),
+                    dm.getFamilyArray(), dm.getFamilyOffset(), dm.getFamilyLength()) == 0) &&
+                    Bytes.compareTo(cell.getQualifierArray(), cell.getQualifierOffset(),
+                            cell.getQualifierLength(),
+                            dm.getQualifierArray(), dm.getQualifierOffset(),
+                            dm.getQualifierLength()) == 0) {
+                if (dm.getType() == Cell.Type.Delete) {
```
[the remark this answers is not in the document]
kadirozde: I will add comments here. It is to make sure that a Type.Delete marker is used for deleting only one cell.

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/StoreCompactionScanner.java
kadirozde: Good catch! Yes, the second one is for the stop row.

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/StoreCompactionScanner.java
kadirozde: I moved the logic for detecting the need for a region level scan to formRowVersions. I will update the code.

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/StoreCompactionScanner.java
kadirozde: The MinVersions check is required to be done outside the TTL window. If the version outside the TTL window is less than MinVersions then it is retained.

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/StoreCompactionScanner.java
kadirozde: Thank you for the link. I think we need to address this. I will look into it more later.

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/StoreCompactionScanner.java
kadirozde: Yes, we need to preserve it. The decision is done later in prepareResults. This is just an optimization to early terminate the row version generation here. However, there is a typo in the if statement. It should read if (firstCell.getTimestamp() < maxLookbackWindowStart &&. Good catch!

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/UngroupedAggregateRegionObserver.java
```diff
@@ -588,6 +588,49 @@ public InternalScanner preCompact(ObserverContext<RegionCoprocessorEnvironment>
...
+                                    LOGGER.error(
+                                            "Unable to modify compaction scanner to retain deleted "
+                                                    + "cells for a table with disabled Index; "
+                                                    + fullTableName, e);
+                                }
+                            }
+                        }
+                    }
+                    if (!isDisabled) {
+                        internalScanner = new StoreCompactionScanner(c.getEnvironment(), store, scanner,
```
stoty, line 631: IIUC this disables the new scanner for this case, and uses the default one.

TBH I haven't analyzed the new logic in detail, but wouldn't that defeat the original purpose of this patch for this case ?

Also the old code sets the keepDeletedCells and TTL, and this one doesn't. If we use the default scanner, what causes  deletes to be kept ?

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/UngroupedAggregateRegionObserver.java
```diff
@@ -588,6 +588,49 @@ public InternalScanner preCompact(ObserverContext<RegionCoprocessorEnvironment>
...
+                    boolean isDisabled = false;
+                    if (request.isMajor()) {
+                        final String
+                                fullTableName =
+                                c.getEnvironment().getRegion().getRegionInfo().getTable().getNameAsString();
+                        if (!PhoenixDatabaseMetaData.SYSTEM_CATALOG_NAME.equals(fullTableName)) {
+                            try (PhoenixConnection conn = QueryUtil.getConnectionOnServer(
+                                    compactionConfig).unwrap(PhoenixConnection.class)) {
+                                PTable table = PhoenixRuntime.getTableNoCache(conn, fullTableName);
+                                List<PTable>
```
stoty, line 600: We could also check what coprocessors the data table has, and only apply this logic if the old indexing is still used.

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/UngroupedAggregateRegionObserver.java
stoty: Thanks for the explanation.

kadirozde:
I am not done yet. I implemented everything that I planned. I may need to enhance TTL masking to cover one more case where  an expired row revives between compactions. I also need add more comments and possibly more tests. There are still some test failures to address too.

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/BaseScannerRegionObserver.java
kadirozde: Done

phoenix-core/src/main/java/org/apache/phoenix/filter/PagedFilter.java
kadirozde: Instead of implementing masking in a filter, I implemented in a region scanner (PhoenixTLLRegionScanner) as the masking logic is more complex than what I thought initially.

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/StoreCompactionScanner.java
kadirozde: I did add comments

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/StoreCompactionScanner.java
```diff
@@ -0,0 +1,422 @@
...
+                                long maxLookbackInMs) {
+        this.storeScanner = storeScanner;
+        this.region = env.getRegion();
+        this.store = store;
+        this.env = env;
+        this.config = env.getConfiguration();
+        compactionTime = EnvironmentEdgeManager.currentTimeMillis();
+        this.maxLookbackWindowStart = compactionTime - maxLookbackInMs;
+        ColumnFamilyDescriptor cfd = store.getColumnFamilyDescriptor();
+        long ttl = cfd.getTimeToLive();
```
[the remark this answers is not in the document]
kadirozde: I implemented a solution where CompactionScanner maintains a per JVM map of table:column family to max lookback age  so that the max lookback age can be overridden dynamically.

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/StoreCompactionScanner.java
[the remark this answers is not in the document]
kadirozde: Done

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/StoreCompactionScanner.java
kadirozde: The secondary indexing implementation let Deletes always win. I preserved the same behavior for Phoenix compaction policy in this PR.

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/UngroupedAggregateRegionObserver.java
kadirozde: Done

kadirozde:
> I am not done yet. I implemented everything that I planned. I may need to enhance TTL masking to cover one more case where an expired row revives between compactions. I also need add more comments and possibly more tests. There are still some test failures to address too.

I have addressed all comments provided so far, fixed test failures and introduced new tests to cover the code. The PR is now finalized. I also added a config param to disable the behavior at the cluster level when needed. @apurtell, @stoty, @gjacoby126, @jpisaac

stoty:
The fix for my issue with the old indexing code looks good.
I haven't followed the other issues, so I wouldn't comment on the rest of the patch.

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/UngroupedAggregateRegionObserver.java
```diff
@@ -588,21 +590,70 @@ public InternalScanner preCompact(ObserverContext<RegionCoprocessorEnvironment>
...
+                        }
+                        if (table != null &&
+                                !PhoenixDatabaseMetaData.SYSTEM_CATALOG_NAME.equals(fullTableName) &&
+                                !ScanUtil.hasCoprocessor(c.getEnvironment(),
+                                GlobalIndexChecker.class.getName())) {
+                            List<PTable>
+                                    indexes =
+                                    PTableType.INDEX.equals(table.getType()) ?
+                                            Lists.newArrayList(table) :
+                                            table.getIndexes();
```
virajjasani, line 620: Shall we comment just above this if condition that this condition is only for old indexing design?

virajjasani (approved):
+1, I have ran `PhoenixTableTTLIT` multiple times locally to try covering as much source changes as possible. Overall it looks good, there might be some minor follow ups that can be taken up later but given the criticality of using the new user as well as compaction scanners to mitigate data integrity issues, it's worth merging this PR.

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/CompactionScanner.java
```diff
@@ -0,0 +1,776 @@
...
+                continue;
+            }
+            if (rowContext.maxTimestamp != 0 &&
+                    column.get(0).getTimestamp() < rowContext.minTimestamp) {
+                continue;
+            }
+            column.remove(0);
+            if (cell.getType() == Cell.Type.DeleteColumn ||
+                    cell.getType() == Cell.Type.Delete) {
+                rowContext.addColumnDeleteMarker(cell);
```
gjacoby126, line 557: We add a column delete marker when we hit a row delete marker? (Cell.Type.Delete?)

kadirozde:
@gjacoby126, I feel that the code is too complicated now and I have just started working on a version that greatly simplifies the compaction logic. I appreciate if you can review it again after that. Hope you did not spend too much time on this version.

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/CompactionScanner.java
```diff
@@ -0,0 +1,776 @@
...
+    public static void overrideMaxLookback(String tableName, String columnFamilyName,
+            long maxLookbackInMillis) {
+        if (tableName == null || columnFamilyName == null) {
+            return;
+        }
+        Long old = maxLookbackMap.putIfAbsent(tableName + SEPARATOR + columnFamilyName,
+                maxLookbackInMillis);
+        if (old == null || old < maxLookbackInMillis) {
+            maxLookbackMap.put(columnFamilyName, maxLookbackInMillis);
+        }
```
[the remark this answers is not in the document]
kadirozde: Good catch, it should include the table name

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/PhoenixTTLRegionScanner.java
```diff
@@ -0,0 +1,218 @@
...
+        boolean found = false;
+        for (Cell c : result) {
+            ts = c.getTimestamp();
+            if (!found && ScanUtil.isEmptyColumn(c, emptyCF, emptyCQ)) {
+                if (ts < ttlWindowStart) {
+                    return true;
+                }
+                found = true;
+            }
+            ts = c.getTimestamp();
```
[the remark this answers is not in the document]
kadirozde: Yes, it is redundant

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/PhoenixTTLRegionScanner.java
```diff
@@ -0,0 +1,218 @@
...
+            if (maxTimestamp < ts) {
+                maxTimestamp = ts;
+            }
+            if (minTimestamp > ts) {
+                minTimestamp = ts;
+            }
```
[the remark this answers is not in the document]
kadirozde: Of course they are equal. I did not want to incur a possible function call for every scanned cell unnecessarily. I am not sure if the Java complier will replace these function calls with inline functions. If so, I will happily change it.

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/UngroupedAggregateRegionObserver.java
kadirozde: Yes, I will add it

phoenix-core/src/it/java/org/apache/phoenix/end2end/QueryDatabaseMetaDataIT.java
```diff
@@ -1126,6 +1126,7 @@ public void testDropAllKVCols() throws Exception {
 
             conn.createStatement().executeUpdate("ALTER TABLE " + tableName + " DROP COLUMN col1");
 
+            TestUtil.dumpTable(conn, TableName.valueOf(tableName));
```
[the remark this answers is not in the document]
kadirozde: I will remove it

phoenix-core/src/it/java/org/apache/phoenix/end2end/SCNIT.java
```diff
@@ -27,9 +27,11 @@
 import java.sql.SQLException;
 import java.util.Properties;
 
+import org.apache.hadoop.hbase.TableName;
```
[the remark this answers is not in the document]
kadirozde: Yes

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/BaseScannerRegionObserver.java
```diff
@@ -21,12 +21,7 @@
 import java.util.List;
 
 import org.apache.hadoop.conf.Configuration;
-import org.apache.hadoop.hbase.Cell;
-import org.apache.hadoop.hbase.DoNotRetryIOException;
-import org.apache.hadoop.hbase.HConstants;
-import org.apache.hadoop.hbase.KeepDeletedCells;
-import org.apache.hadoop.hbase.MemoryCompactionPolicy;
-import org.apache.hadoop.hbase.NotServingRegionException;
```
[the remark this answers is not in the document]
kadirozde: ok

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/BaseScannerRegionObserver.java
```diff
@@ -413,59 +435,81 @@ RegionScanner getWrappedScanner(final ObserverContext<RegionCoprocessorEnvironme
...
-     HBase 2.3 gave us this ability back.
-    We also need to override the flush compaction coproc hooks in order to implement max lookback
-     age to keep versions from being purged.
-   */
-
+    public void setScanOptionsForFlushesAndCompactions(ScanOptions options) {
+        // We want the store to give us all the deleted cells to StoreCompactionScanner
+        KeepDeletedCells keepDeletedCells = options.getKeepDeletedCells();
+        if (keepDeletedCells == KeepDeletedCells.FALSE) {
+            keepDeletedCells = KeepDeletedCells.TRUE;
```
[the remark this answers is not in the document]
kadirozde: I will set it to KeepDeletedCells.TTL always regardless of what the existing option is

phoenix-core/src/main/java/org/apache/phoenix/util/ScanUtil.java
```diff
@@ -1050,6 +1052,17 @@ public static boolean isEmptyColumn(Cell cell, byte[] emptyCF, byte[] emptyCQ) {
                         cell.getQualifierLength(), emptyCQ, 0, emptyCQ.length) == 0;
     }
 
+
+    public static boolean isEmptyColumn(Cell cell, byte[] emptyCF) {
+        return Bytes.compareTo(cell.getFamilyArray(), cell.getFamilyOffset(),
```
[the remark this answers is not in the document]
kadirozde: OK

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/CompactionScanner.java
```diff
@@ -0,0 +1,776 @@
...
+        // We can do compaction at the store level
+        if (rowContext.familyDeleteMarker == null && rowContext.familyVersionDeleteMarker == null) {
+            // The compaction row version is alive
+            if (rowVersion.version < maxVersion) {
+                // Rule 3
+                result.addAll(rowVersion.cells);
+            }
+        } else {
+            // Deleted
+            if ((rowVersion.version < maxVersion && keepDeletedCells == KeepDeletedCells.TRUE) ||
```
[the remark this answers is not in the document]
kadirozde: minVersions is considered only outside the TTL window.

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/CompactionScanner.java
kadirozde: Cell.Type.Delete is a column delete marker to delete only one cell version. It should have been named Cell.Type.DeleteColumnVersion.

kadirozde:
@jpisaac, @gjacoby126, @virajjasani, I have completed the code simplification/restructuring and updated the design doc. I also made the changes based on your feedback. You can review the code now. Thank you again!

jpisaac (commented):
Did an initial review, and have some nit comments. My local testing was also green. Will do a more detailed review of some of the critical classes.

phoenix-core/src/it/java/org/apache/phoenix/end2end/ExplainPlanWithStatsEnabledIT.java
```diff
@@ -165,7 +165,7 @@ public void testBytesRowsForSelectWithLimitIgnored() throws Exception {
             ResultSet rs = conn.createStatement().executeQuery(sql);
             assertFalse(rs.next());
             Estimate info = getByteRowEstimates(conn, sql, binds);
-            assertEquals((Long) 390L, info.estimatedBytes);
+            assertEquals((Long) 691L, info.estimatedBytes);
```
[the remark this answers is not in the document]
kadirozde: This is to undo my change on 3/17/2019. I had the impression that the Phoenix server code should never return the empty column cell back to the client. It turned out that it was not the case especially for this query. As you can see, this PR does not remove the empty column from the returned result anymore. The increase/revert is because of that.

phoenix-core/src/it/java/org/apache/phoenix/end2end/ExplainPlanWithStatsEnabledIT.java
```diff
@@ -204,7 +204,7 @@ public void testBytesRowsForSelectOnIndex() throws Exception {
                 assertTrue(rs.next());
             }
             Estimate info = getByteRowEstimates(conn, sql, binds);
-            assertEquals((Long) 390L, info.estimatedBytes);
+            assertEquals((Long) 691L, info.estimatedBytes);
```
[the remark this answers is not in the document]
kadirozde: See the previous comment.

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/TTLRegionScanner.java
```diff
@@ -0,0 +1,217 @@
...
+        super(s);
+        this.env = env;
+        this.scan = scan;
+        emptyCQ = scan.getAttribute(EMPTY_COLUMN_QUALIFIER_NAME);
+        emptyCF = scan.getAttribute(EMPTY_COLUMN_FAMILY_NAME);
+        long currentTime = EnvironmentEdgeManager.currentTimeMillis();
+        ttl = env.getRegion().getTableDescriptor().getColumnFamilies()[0].getTimeToLive();
+        ttlWindowStart = ttl == HConstants.FOREVER ? 1 : currentTime - ttl * 1000;
+        ttl *= 1000;
+        isPhoenixTableTTLEnabled = emptyCF != null && emptyCQ != null &&
```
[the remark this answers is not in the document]
kadirozde: Regardless if the TTL is disabled cluster wide or the client is an older client and did not supply the empty column parameters, the masking should not be done here. I will rename it to isMaskingEnabled to eliminate any confusion.

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/CompactionScanner.java
```diff
@@ -0,0 +1,908 @@
...
+                    // We need to check if the next delete family marker exits. If so, we need
+                    // to record its timestamp as by definition a compaction row version cannot
+                    // cross a family delete marker
+                    if (column.size() > 1) {
+                        nextDeleteFamilyTimestamp = column.get(1).getTimestamp();
+                    } else {
+                        nextDeleteFamilyTimestamp = 0;
+                    }
+                }
+            } if (firstCell.getType() == Cell.Type.Put) {
```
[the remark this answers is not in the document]
kadirozde: In oder to prevent the unnecessary evaluation of the second if condition when the first if condition is true, I wanted to have "else if" but did not type the else keyword apparently. I will add it.

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/CompactionScanner.java
```diff
@@ -0,0 +1,908 @@
...
+            }
+        }
+
+        private int compareQualifiers(Cell a, Cell b) {
+            return Bytes.compareTo(a.getQualifierArray(), a.getQualifierOffset(),
+                    a.getQualifierLength(),
+                    b.getQualifierArray(), b.getQualifierOffset(), a.getQualifierLength());
+        }
+
+        private int compareTypes(Cell a, Cell b) {
```
jpisaac, line 781: Maybe a new Comparator?

jpisaac (commented):
Added some more comments for CompactionScanner.

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/CompactionScanner.java
```diff
@@ -0,0 +1,908 @@
...
+    private final byte[] emptyCF;
+    private final byte[] emptyCQ;
+    private static Map<String, Long> maxLookbackMap = new ConcurrentHashMap<>();
+    private PhoenixLevelRowCompactor phoenixLevelRowCompactor;
+    private HBaseLevelRowCompactor hBaseLevelRowCompactor;
+
+    public CompactionScanner(RegionCoprocessorEnvironment env,
+            Store store,
+            InternalScanner storeScanner,
+            long maxLookbackInMillis,
```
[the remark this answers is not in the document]
kadirozde: This can be done when a test requiring such a constructor is added. At this moment, it is not required.

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/CompactionScanner.java
```diff
@@ -0,0 +1,908 @@
...
+                columnFamilyName.startsWith(LOCAL_INDEX_COLUMN_FAMILY_PREFIX);
+        phoenixLevelRowCompactor = new PhoenixLevelRowCompactor();
+        hBaseLevelRowCompactor = new HBaseLevelRowCompactor();
+    }
+
+    /**
+     * Any coprocessors within a JVM can extend the max lookback window for a column family
+     * by calling this static method.
+     */
+    public static void overrideMaxLookback(String tableName, String columnFamilyName,
```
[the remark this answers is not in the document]
kadirozde: I looked at sharedData but did not see how it would improve the implementation. We are not really sharing data here. Actually, what is implemented here provides better data hiding.

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/CompactionScanner.java
```diff
@@ -0,0 +1,908 @@
...
+        /**
+         * Decide if compaction row versions inside the TTL window should be retained. The
+         * versions are retained if one of the following conditions holds
+         * 1. The compaction row version is alive and its version is less than VERSIONS
+         * 2. The compaction row version is deleted and KeepDeletedCells is TTL
+         * 3. The compaction row version is deleted, its version is less than MIN_VERSIONS and
+         * KeepDeletedCells is TRUE
+         *
+         */
+        private boolean retainInsideTTLWindow(List<Cell> result,
```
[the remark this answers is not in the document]
kadirozde: Good point! We do not need these methods to return a value anymore after restructuring the code.

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/CompactionScanner.java
kadirozde: I simplified trimRegionResult.

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/BaseScannerRegionObserver.java
```diff
@@ -354,10 +361,14 @@ public boolean nextRaw(List<Cell> result) throws IOException {
                 overrideDelegate();
                 return super.nextRaw(result);
             }
+            @Override
+            public RegionScanner getNewRegionScanner(Scan scan) throws IOException {
+                return new RegionScannerHolder(c, scan,
+                        ((DelegateRegionScanner) delegate).getNewRegionScanner(scan));
```
jpisaac, line 367: Coercing without checking the instance is of type DelegateRegionScanner can lead to ClassCastException.

jpisaac (commented):
Reviewed TTLRegionScanner

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/CompactionScanner.java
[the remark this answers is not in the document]
jpisaac: I was using a modified constructor for my tests, to enable me to set maxLookbackInMillis, ttl, and compactionTime, hence the ask.

    public CompactionScanner(RegionCoprocessorEnvironment env,
                                Store store,
                                InternalScanner storeScanner,
                                long maxLookbackInMillis,
                                long ttl,
                                long compactionTime,
                                byte[] emptyCF,
                                byte[] emptyCQ) {

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/TTLRegionScanner.java
```diff
@@ -0,0 +1,220 @@
...
+    private boolean initialized = false;
+
+    public TTLRegionScanner(final RegionCoprocessorEnvironment env, final Scan scan,
+            final RegionScanner s) {
+        super(s);
+        this.env = env;
+        this.scan = scan;
+        emptyCQ = scan.getAttribute(EMPTY_COLUMN_QUALIFIER_NAME);
+        emptyCF = scan.getAttribute(EMPTY_COLUMN_FAMILY_NAME);
+        long currentTime = EnvironmentEdgeManager.currentTimeMillis();
```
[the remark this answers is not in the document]
kadirozde: Absolutely. I actually planned to make this change but forgot it. Thank you for catching this!

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/TTLRegionScanner.java
```diff
@@ -0,0 +1,220 @@
...
+        // be done here.
+        isMaskingEnabled = emptyCF != null && emptyCQ != null &&
+                env.getConfiguration().getBoolean(QueryServices.PHOENIX_TABLE_TTL_ENABLED,
+                        QueryServicesOptions.DEFAULT_PHOENIX_TABLE_TTL_ENABLED);
+    }
+
+    private void init() throws IOException {
+        // HBase PageFilter will also count the expired rows.
+        // Instead of using PageFilter for counting, we will count returned row here.
+        PageFilter pageFilter = ScanUtil.removePageFilter(scan);
```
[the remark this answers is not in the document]
kadirozde: PageFilter is an HBase builtin filter. The name should have been LimitFilter instead of PageFilter. So, there is no relation between the server paging feature and PageFilter. Phoenix has PagingRegionScanner and PagingFilter. I know it is confusing. 

The code is simpler when we count always.  Adding an if statement and counting only when the PageFilter is removed is not more efficient.

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/TTLRegionScanner.java
```diff
@@ -0,0 +1,220 @@
...
+            hasMore = raw ? delegate.nextRaw(result) : delegate.next(result);
+            if (result.isEmpty() || ScanUtil.isDummy(result)) {
+                return hasMore;
+            }
+            if (!isExpired(result)) {
+                return hasMore;
+            }
+            Cell cell = result.get(0);
+            result.clear();
+            if (EnvironmentEdgeManager.currentTimeMillis() - startTime > pageSize) {
```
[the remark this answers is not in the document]
kadirozde: Good catch! This one was supposed to be pageSizeInMs. I will fix this.

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/DelegateRegionScanner.java
```diff
@@ -87,4 +88,8 @@ public int getBatch() {
     public RegionInfo getRegionInfo() {
         return delegate.getRegionInfo();
     }
+
+    public RegionScanner getNewRegionScanner(Scan scan) throws IOException {
+        return ((DelegateRegionScanner)delegate).getNewRegionScanner(scan);
```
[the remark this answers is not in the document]
kadirozde: Regardless if you check or not, when the instance is not of the expected type, an exception should be thrown.

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/BaseScannerRegionObserver.java
kadirozde: That is the expected behavior.

kadirozde: Added try-catch for ClassCastException

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/DelegateRegionScanner.java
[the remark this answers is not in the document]
kadirozde: Added try-catch for ClassCastException

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/CompactionScanner.java
```diff
@@ -0,0 +1,879 @@
...
+            return Bytes.compareTo(a.getQualifierArray(), a.getQualifierOffset(),
+                    a.getQualifierLength(),
+                    b.getQualifierArray(), b.getQualifierOffset(), b.getQualifierLength());
+        }
+
+        /**
+         * The generates the intersection of regionResult and input. The result is the resulting
+         * intersection.
+         */
+        private void trimRegionResult(List<Cell> regionResult, List<Cell> input,
```
jpisaac, line 830: Don't quite understand this, will sync with you offline

jpisaac:
@kadirozde Other than some clarifications I am done with my review. Let me know if you have any more updates

jpisaac (approved):
LGMT +1, Thank you @kadirozde for the fix and enhancements.

PR #1572, opened 2023-03-03
PHOENIX-6890 Remove binds parameters from QueryCompiler

stoty:


gjacoby126:
Hmm...let's say that in the future we had better cost-based optimization -- wouldn't the parameters be relevant to choosing the best plan?

stoty:
We cab still get the parameters via StatemetrContex.getBindManager().  in that case.

If we need the whole List, we can add a getter for that to BindManager.

gjacoby126 (approved):
+1

PR #1575, opened 2023-03-04
PHOENIX-6141 : Ensure consistency between SYSTEM.CATALOG and SYSTEM.CHILD_LINK

palashc:
Design Doc: https://docs.google.com/document/d/1Xwz1bDQVogsZWBtwd2CVevkyN7p9UubotQrUzG_PtH4/edit


phoenix-core/src/main/java/org/apache/phoenix/coprocessor/ChildLinkMetaDataObserver.java
```diff
@@ -0,0 +1,290 @@
...
+import static org.apache.phoenix.util.ScanUtil.getDummyResult;
+import static org.apache.phoenix.util.ScanUtil.getPageSizeMsForRegionScanner;
+import static org.apache.phoenix.util.ScanUtil.isDummy;
+import static org.apache.phoenix.thirdparty.com.google.common.base.Preconditions.checkArgument;
+
+
+/**
+ * Coprocessor that verifies scanned rows of SYSTEM.CHILD_LINK table
+ */
+public class ChildLinkMetaDataObserver extends BaseScannerRegionObserver implements RegionCoprocessor {
```
jpisaac, line 45: @palashc I am assuming the verification is being done out of band right?
Any reason we do not want to reuse the ChildLinkMetaDataEndpoint coprocessor instead of having a new one?

We can extend the ChildLinkMetaDataService to have a verifyRows method and have this logic in there?

palashc: I have made the change to reuse the existing co-proc.

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/ChildLinkMetaDataEndpoint.java
```diff
@@ -111,4 +141,245 @@ private PhoenixMetaDataCoprocessorHost getCoprocessorHost() {
...
+                        byte[] rowKey = CellUtil.cloneRow(cell);
+                        result.clear();
+                        getDummyResult(rowKey, result);
+                        return true;
+                    }
+                    // skip this row as it is invalid
+                    // if there is no more row, then result will be an empty list
+                } while (hasMore);
+                rowCount++;
+                if (rowCount == pageSize) {
```
kadirozde, line 212: You need to check if the scan has the page filter. If so, you need to remove it from the scan and then reopen the underlying the region scanner. Please see GlobalIndexChecker.

kadirozde: Please add a test to make sure that limit queries return the correct number of rows in the presence of unverified rows.

kadirozde:
I expected that GlobalIndexChecker#GlobalIndexScanner would be restructured to inherit ReadRepairScanner to eliminate code duplication.

palashc:
> I expected that GlobalIndexChecker#GlobalIndexScanner would be restructured to inherit ReadRepairScanner to eliminate code duplication.
@kadirozde I was planning that for a different PR and keep this one just for this JIRA.

phoenix-core/src/it/java/org/apache/phoenix/end2end/OrphanChildLinkRowsIT.java
```diff
@@ -0,0 +1,201 @@
...
+     * 2. Create a view (same name as existing view on T1) on T2. This CREATE VIEW will fail, verify if there was no orphan child link because of that.
+     *
+     * 3. Instrument CQSI to fail phase three of CREATE VIEW. Create a new view V2 on T2 (passes) and V1 on T2 which will fail.
+     *    Both links T2->V2 and T2->V1 will be in UNVERIFIED state, repaired during read.
+     *    Check if only 2 child links are returned: T2->V2 and T1->V1.
+     */
+    @Test
+    public void testNoOrphanChildLinkRow() throws Exception {
+
+        ConnectionQueryServicesImpl.setFailPhaseThreeChildLinkWriteForTesting(false);
```
jpisaac, line 76: Can we try and extend/use ConnectionQueryServicesTestImpl for testing instead of using CQSI?

phoenix-core/src/it/java/org/apache/phoenix/end2end/OrphanChildLinkRowsIT.java
```diff
@@ -0,0 +1,201 @@
+package org.apache.phoenix.end2end;
```
jpisaac, line 1: nit: Add license

phoenix-core/src/it/java/org/apache/phoenix/end2end/OrphanChildLinkRowsIT.java
```diff
@@ -0,0 +1,201 @@
...
+     *    Both links T2->V2 and T2->V1 will be in UNVERIFIED state, repaired during read.
+     *    Check if only 2 child links are returned: T2->V2 and T1->V1.
+     */
+    @Test
+    public void testNoOrphanChildLinkRow() throws Exception {
+
+        ConnectionQueryServicesImpl.setFailPhaseThreeChildLinkWriteForTesting(false);
+        ChildLinkScanTask.disableChildLinkScanTask(true);
+
+        String v2 = "CREATE VIEW VS1.V1 (NEW_COL1 INTEGER, NEW_COL2 INTEGER) AS SELECT * FROM S2.T2 WHERE B > 10";
```
jpisaac, line 79: nit: Better to have unique where conditions to simulate updatable views. But ok since we are not testing view functionality

phoenix-core/src/it/java/org/apache/phoenix/end2end/OrphanChildLinkRowsIT.java
```diff
@@ -0,0 +1,201 @@
...
+
+        try (Connection connection = DriverManager.getConnection(getUrl())) {
+            connection.createStatement().execute(v2);
+        }
+        catch (TableAlreadyExistsException e) {
```
jpisaac, line 84: nit: empty catch

phoenix-core/src/it/java/org/apache/phoenix/end2end/OrphanChildLinkRowsIT.java
```diff
@@ -0,0 +1,201 @@
...
+        // configure CQSI to fail the last write phase of CREATE VIEW
+        // where child link mutations are set to VERIFIED or are deleted
+        ConnectionQueryServicesImpl.setFailPhaseThreeChildLinkWriteForTesting(true);
+        String v3 = "CREATE VIEW IF NOT EXISTS VS2.V2 (NEW_COL1 INTEGER, NEW_COL2 INTEGER) AS SELECT * FROM S2.T2 WHERE B > 10";
+
+        try (Connection connection = DriverManager.getConnection(getUrl())) {
+            connection.createStatement().execute(v3);
+            connection.createStatement().execute(v2);
+        }
+        catch (TableAlreadyExistsException e) {
```
jpisaac, line 98: same here

phoenix-core/src/it/java/org/apache/phoenix/end2end/OrphanChildLinkRowsIT.java
```diff
@@ -0,0 +1,201 @@
...
+                expectedChildLinks.put(table1, view);
+                try {
+                    connection.createStatement().execute(String.format(CREATE_VIEW_DDL, view, table2));
+                }
+                catch (TableAlreadyExistsException e) {
+
+                }
+            }
+
+            String childLinkQuery = "SELECT * FROM SYSTEM.CHILD_LINK LIMIT 7";
```
jpisaac, line 172: nit : why not use SELECT count(*) instead?

phoenix-core/src/it/java/org/apache/phoenix/end2end/OrphanChildLinkRowsIT.java
```diff
@@ -0,0 +1,201 @@
...
+            ResultSet rs = connection.createStatement().executeQuery(childLinkQuery);
+            int count = 0;
+            while (rs.next()) {
+                count++;
+                System.out.println(rs.getString(3) + " " + rs.getString(5));
```
jpisaac, line 177: nit: remove debug output

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/ChildLinkMetaDataEndpoint.java
```diff
@@ -111,4 +145,162 @@ private PhoenixMetaDataCoprocessorHost getCoprocessorHost() {
...
+            if ((EnvironmentEdgeManager.currentTimeMillis() - ts) > ageThreshold) {
+                Delete del = new Delete(rowKey);
+                Mutation[] mutations = new Mutation[]{del};
+                region.batchMutate(mutations);
+            }
+        }
+
+
+        private void markChildLinkVerified(byte[] rowKey, long ts, Region region) throws IOException {
+            Put put = new Put(rowKey);
```
jpisaac, line 284: Will this not create a new ts for the row

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/ChildLinkMetaDataEndpoint.java
```diff
@@ -111,4 +145,162 @@ private PhoenixMetaDataCoprocessorHost getCoprocessorHost() {
...
+            Cell cell = row.get(0);
+            byte[] rowKey = CellUtil.cloneRow(cell);
+            long ts = row.get(0).getTimestamp();
+
+            Scan sysCatScan = getExternalScanner();
+            childLinkScan = new Scan(scan);
+
+
+            // build syscat rowKey using given rowKey
+            byte[] sysCatRowKey = getSysCatRowKey(rowKey);
```
jpisaac, line 216: We should check for the existence of the view header too. The existence of only child->parent link type (3) row might be an bug.

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/ReadRepairScanner.java
```diff
@@ -0,0 +1,196 @@
...
+    public long pageSize = Long.MAX_VALUE;
+    public long rowCount = 0;
+    public long maxTimestamp;
+    public long ageThreshold;
+    public boolean restartScanDueToPageFilterRemoval = false;
+
+    /*
+    Scanner used for checking ground truth to help with read repair.
+     */
+    private Scan externalScanner = null;
```
jpisaac, line 49: nit: naming convention, reader will get confused with this being Scanner.

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/ReadRepairScanner.java
```diff
@@ -0,0 +1,196 @@
...
+    public Scan getExternalScanner() { return externalScanner; }
```
jpisaac, line 50: How is this getting set?

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/ReadRepairScanner.java
```diff
@@ -0,0 +1,196 @@
...
+                    hasMore = scanner.next(result);
+                }
+                if (result.isEmpty()) {
+                    return hasMore;
+                }
+                if (isDummy(result)) {
+                    return true;
+                }
+                Cell cell = result.get(0);
+                if (verifyRowAndRepairIfNecessary(result)) {
```
jpisaac, line 92: Can u add some comments here - about the possible outcomes.
As I understood it, it can only return true or an exception, If so the statements after this are never executed right?

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/ReadRepairScanner.java
```diff
@@ -0,0 +1,196 @@
...
+            }
+
+            if (cellList.isEmpty()) {
+                return false;
+            }
+            return true;
+        }
+    }
+
+    private PageFilter removePageFilterFromFilterList(FilterList filterList) {
```
jpisaac, line 153: Will let @kadirozde take a look

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/tasks/ChildLinkScanTask.java
```diff
@@ -0,0 +1,53 @@
...
+
+import java.sql.ResultSet;
+
+/*
+Task to run a simple select * query on SYSTEM.CHILD_LINK table to trigger read repair and verify any unverified rows.
+ */
+public class ChildLinkScanTask extends BaseTask {
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(ChildLinkScanTask.class);
+    private static final String CHILD_LINK_QUERY = "SELECT * FROM SYSTEM.CHILD_LINK";
```
jpisaac, line 19: Can we use count(*) instead?

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/tasks/ChildLinkScanTask.java
```diff
@@ -0,0 +1,53 @@
...
+    @Override
+    public TaskRegionObserver.TaskResult run(Task.TaskRecord taskRecord) {
+
+        if (isDisabled) {
+            return new TaskRegionObserver.TaskResult(TaskRegionObserver.TaskResultCode.FAIL, "ChildLinkScan task is disabled.");
+        }
+
+        int count = 0;
+        try {
+            PhoenixConnection pconn = QueryUtil.getConnectionOnServer(env.getConfiguration()).unwrap(PhoenixConnection.class);
```
jpisaac, line 36: If we can inject the scan attributes for Read repair here, then it will be more controlled. WDYT?

phoenix-core/src/main/java/org/apache/phoenix/query/ConnectionQueryServicesImpl.java
```diff
@@ -394,6 +401,11 @@ public class ConnectionQueryServicesImpl extends DelegateQueryServices implement
     private ServerSideRPCControllerFactory serverSideRPCControllerFactory;
     private boolean localIndexUpgradeRequired;
 
+    private static boolean failPhaseThreeChildLinkWriteForTesting = false;
+    public static void setFailPhaseThreeChildLinkWriteForTesting(boolean fail) {
```
jpisaac, line 405: See comments related to see if we can use ConnectionQueryServicesTestImpl

jpisaac (commented):
Made an initial pass, Left comments

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/ReadRepairScanner.java
palashc: This gets set [here](https://github.com/apache/phoenix/pull/1575/files#diff-c392040c2e8440e5df98d189022fa6504a58b39107507e4ceafbd0b5fa035650R138), the first time a row needs a repair.

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/ReadRepairScanner.java
palashc: Yes I will add a comment. 
It returns true if the row was already VERIFIED. In case it was not, this method returns false so that this row is ignored. (the row gets repaired and a new scanner is opened from that row, so it will be examined again).

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/ChildLinkMetaDataEndpoint.java
palashc: the timestamp is extracted from the first cell in the row i.e. `long ts = row.get(0).getTimestamp();`

phoenix-core/src/it/java/org/apache/phoenix/end2end/OrphanChildLinkRowsIT.java
palashc: Yes, that makes sense.

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/tasks/ChildLinkScanTask.java
palashc: Yes, will change.

phoenix-core/src/it/java/org/apache/phoenix/end2end/OrphanChildLinkRowsIT.java
palashc: @jpisaac I wanted to test the LIMIT functionality to make sure the scanner implementation is correct, as per a previous comment from Kadir.

phoenix-core/src/it/java/org/apache/phoenix/end2end/OrphanChildLinkRowsIT.java
jpisaac: I see so we want to swallow the exception.

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/ChildLinkMetaDataEndpoint.java
jpisaac: I think you may need to use the Put(byte[] row, long ts) constructor.

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/tasks/ChildLinkScanTask.java
jpisaac: See this code snippet for eg. https://github.com/apache/phoenix/blob/9a50b02ca6ce5fcca7f1b7975bd22d605ad84dc0/phoenix-core/src/it/java/org/apache/phoenix/end2end/ViewTTLIT.java#L1871

phoenix-core/src/main/java/org/apache/phoenix/util/ScanUtil.java
```diff
@@ -1333,10 +1334,22 @@ public static Long getPageSizeInMs(ReadOnlyProps props) {
         return null;
     }
 
+    private static void setScanAtrributesForChildLinkRepair(Scan scan, PTable table, PhoenixConnection phoenixConnection) {
```
jpisaac, line 1337: You can move this to the ChildLinkScanTask.java file. Thus will run the verification logic only when needed instead of every time there is a query on the CHILD_LINK table.

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/ReadRepairScanner.java
```diff
@@ -0,0 +1,219 @@
...
+    public long rowCount = 0;
+    public long maxTimestamp;
+    public long ageThreshold;
+    public boolean restartScanDueToPageFilterRemoval = false;
+
+    /*
+    Scanner used for checking ground truth to help with read repair.
+     */
+    private Scan externalScan = null;
+    public Scan getExternalScan() { return externalScan; }
```
jpisaac, line 67: Can u add some comments how this method is to be used by the framework?  Do not see a setter for this attribute? Is it supposed to be overridden by the implementing classes?

palashc: @jpisaac I will remove this from the base class and let individual use cases define their own Scans since they might need more than one.

phoenix-core/src/it/java/org/apache/phoenix/end2end/OrphanChildLinkRowsIT.java
```diff
@@ -0,0 +1,218 @@
...
+            int count = 0;
+            while (rs.next()) {
+                String parentFullName = SchemaUtil.getTableName(rs.getString(TABLE_SCHEM), rs.getString(TABLE_NAME));
+                Assert.assertTrue("Found Orphan Child Link: " + parentFullName + "->" + rs.getString(COLUMN_FAMILY), expectedChildLinks.containsKey(parentFullName));
+                Assert.assertEquals(String.format("Child was not correct in Child Link. Expected : %s, Actual: %s", expectedChildLinks.get(parentFullName), rs.getString(COLUMN_FAMILY)),
+                        expectedChildLinks.get(parentFullName), rs.getString(COLUMN_FAMILY));
+                count++;
+            }
+            Assert.assertTrue("Found Orphan Linking Row", count <= expectedChildLinks.size());
+            Assert.assertTrue("All expected Child Links not returned by query", count >= expectedChildLinks.size());
```
tkhurana, line 213: These assert checks are confusing. Isn't it equivalent to `count == expectedChildLinks.size()` How can both be true ?

phoenix-core/src/it/java/org/apache/phoenix/end2end/OrphanChildLinkRowsIT.java
```diff
@@ -0,0 +1,218 @@
...
+    @Test
+    public void testNoOrphanChildLinkRow() throws Exception {
+
+        ConnectionQueryServicesTestImpl.setFailPhaseThreeChildLinkWriteForTesting(false);
+        ChildLinkScanTask.disableChildLinkScanTask(true);
+
+        String v2 = "CREATE VIEW VS1.V1 (NEW_COL1 INTEGER, NEW_COL2 INTEGER) AS SELECT * FROM S2.T2 WHERE B > 10";
+
+        try (Connection connection = DriverManager.getConnection(getUrl())) {
+            connection.createStatement().execute(v2);
```
tkhurana, line 98: You should put an `assertFail` after this so that the test fails if we reach this statement

phoenix-core/src/it/java/org/apache/phoenix/end2end/OrphanChildLinkRowsIT.java
```diff
@@ -0,0 +1,218 @@
...
+        verifyNoOrphanChildLinkRow();
+
+        // configure CQSI to fail the last write phase of CREATE VIEW
+        // where child link mutations are set to VERIFIED or are deleted
+        ConnectionQueryServicesTestImpl.setFailPhaseThreeChildLinkWriteForTesting(true);
+        String v3 = "CREATE VIEW IF NOT EXISTS VS2.V2 (NEW_COL1 INTEGER, NEW_COL2 INTEGER) AS SELECT * FROM S2.T2 WHERE B > 10";
+
+        try (Connection connection = DriverManager.getConnection(getUrl())) {
+            connection.createStatement().execute(v3);
+            connection.createStatement().execute(v2);
```
tkhurana, line 113: Put an `assertFail()` after this

phoenix-core/src/it/java/org/apache/phoenix/end2end/OrphanChildLinkRowsIT.java
palashc: yes it is equivalent, I just wanted to differentiate between 2 cases: 
1) if child_link had more rows than expected, that means there are orphans. 
2) if child_link returned fewer rows than expected, something went wrong with the query.

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/ChildLinkMetaDataEndpoint.java
```diff
@@ -111,4 +143,197 @@ private PhoenixMetaDataCoprocessorHost getCoprocessorHost() {
...
+        return Optional.of(this);
+    }
+
+    @Override
+    public RegionScanner postScannerOpen(ObserverContext<RegionCoprocessorEnvironment> c,
+                                         Scan scan, RegionScanner s) throws IOException {
+        if (scan.getAttribute(CHECK_VERIFY_COLUMN) == null) {
+            return s;
+        }
+        return new ChildLinkMetaDataScanner(c.getEnvironment(), scan, s);
```
tkhurana, line 337: You need to wrap this around PagingRegionScanner to get paging to work. `return new ChildLinkMetaDataScanner(c.getEnvironment(), scan,
                new PagingRegionScanner(c.getEnvironment().getRegion(), s, scan));`

phoenix-core/src/it/java/org/apache/phoenix/end2end/OrphanChildLinkRowsIT.java
```diff
@@ -0,0 +1,222 @@
...
+import static org.apache.phoenix.jdbc.PhoenixDatabaseMetaData.TABLE_SCHEM;
+import static org.apache.phoenix.query.QueryConstants.VERIFIED_BYTES;
+
+@Category(NeedsOwnMiniClusterTest.class)
+public class OrphanChildLinkRowsIT extends BaseTest {
+
+    private static Map<String, String> expectedChildLinks = new HashMap<>();
+
+    @BeforeClass
+    public static synchronized void doSetup() throws Exception {
```
tkhurana, line 62: @palashc Can you also run the tests in this file with page size 0. `QueryServices.PHOENIX_SERVER_PAGE_SIZE_MS, "0");` You could make it parameterized so that first the tests will run with default page size and then run with 0 page size.

phoenix-core/src/it/java/org/apache/phoenix/end2end/ViewMetadataIT.java
```diff
@@ -712,12 +714,22 @@ static void createOrphanLink(String parentSchema, String parent1,
                 // expected
             }
 
+            /*
```
tkhurana, line 717: Remove commented code

phoenix-core/src/it/java/org/apache/phoenix/end2end/OrphanChildLinkRowsIT.java
palashc: @tkhurana Done, I have refactored that IT to make the tests independent of each other, previously they were interfering with each other after parameterizing.

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/ReadRepairScanner.java
palashc: Refactored page filter removal based on GlobalIndexChecker

palashc:
`FailoverPhoenixConnectionIT.testConnectionWhenActiveZKRestarts` is the only test failing as of now, which I think is a known flapper.

phoenix-core/src/it/java/org/apache/phoenix/end2end/OrphanChildLinkRowsIT.java
```diff
@@ -0,0 +1,270 @@
...
+     * Create tables T1 and T2. Create 3 views on T1.
+     * Create a view (same name as existing view on T1) on T2. This CREATE VIEW will fail.
+     * Verify if there was no orphan child link from T2.
+     *
+     * Instrument CQSI to fail phase three of CREATE VIEW. Create a new view V4 on T2 (passes) and V1 on T2 which will fail.
+     * Both links T2->V4 and T2->V1 will be in UNVERIFIED state, repaired during read.
+     * Check if T2 has only 1 child link.
+     */
+    @Test
+    public void testNoOrphanChildLinkRow() throws Exception {
```
tkhurana, line 91: @palashc I don't see paging being used in this test

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/tasks/ChildLinkScanTask.java
```diff
@@ -0,0 +1,58 @@
...
+            return new TaskRegionObserver.TaskResult(TaskRegionObserver.TaskResultCode.FAIL,
+                    "ChildLinkScan task is disabled.");
+        }
+
+        int count = 0;
+        try {
+            PhoenixConnection pconn = QueryUtil.getConnectionOnServer(
+                    env.getConfiguration()).unwrap(PhoenixConnection.class);
+            ResultSet rs = pconn.createStatement().executeQuery(CHILD_LINK_QUERY);
+            rs.next();
```
tkhurana, line 41: You should check the return value of rs.next()

phoenix-core/src/it/java/org/apache/phoenix/end2end/OrphanChildLinkRowsIT.java
palashc: @tkhurana this test uses a helper method to query the child_link table, that is where I set the property: https://github.com/apache/phoenix/pull/1575/files/de267ed7749ee7728e53fa45d159d4ffaa32984c#diff-eb1a64ec532ba7e19e95e6f7d6b4e2cfdc4c798b59d021fda3a8c4d80c49c349R239

phoenix-core/src/main/java/org/apache/phoenix/query/ConnectionQueryServicesImpl.java
```diff
@@ -2203,6 +2196,89 @@ public MetaDataResponse call(MetaDataService instance) throws IOException {
...
+        List<Mutation> childLinkMutations = new ArrayList<>();
+        for (Mutation m : mutations) {
+            if (isDelete) {
+                Delete delete = new Delete(m.getRow());
+                childLinkMutations.add(delete);
+            }
+            else {
+                Put put = isVerified ? new Put(m.getRow()) : (Put)m;
+                byte[] emptyColumnValue = isVerified ? VERIFIED_BYTES : UNVERIFIED_BYTES;
+                put.addColumn(emptyCF, emptyCQ, IndexUtil.getMaxTimestamp(m), emptyColumnValue);
```
tkhurana, line 2240: When you are updating the empty column to verified shouldn't that be a different timestamp. There should be two versions of the empty column cell. One with unverified and other with verified and verified one should have a later timestamp. Can you verify this ?

palashc: During debugging, I had used dumpTable and did see the verified empty column mutation with a higher timestamp.

tkhurana: But you are setting the timestamp from the existing cells in the mutation `IndexUtil.getMaxTimestamp(m)`. I am surprised you are getting a higher timestamp.

palashc: Dump from one of the tests:
```
\x00\x00T_N000001\x00\x00V_N000003/0:LINK_TYPE/1687542526662/Put/vlen=1/seqid=0 column= LINK_TYPE val=\x04
\x00\x00T_N000001\x00\x00V_N000003/0:_0/1687542526683/Put/vlen=1/seqid=0 column= _0 val=\x01
\x00\x00T_N000001\x00\x00V_N000003/0:_0/1687542526662/Put/vlen=1/seqid=0 column= _0 val=\x02
```

tkhurana: If I am reading this correctly, your verified version of the empty column will have the same timestamp as the unverified version.

palashc: Yes, you are right. I printed out all timestamps seen in `IndexUtil.getMaxTimestamp(m)` and for the test, they are always long max value. So these timestamps are being assigned at the server side?

tkhurana (approved):

tkhurana:
@palashc ChildLinkScanTask.java has missing Apache license header

palashc:
@tkhurana I have added the license.

PR #1576, opened 2023-03-06
PHOENIX-6893 Remove explicit dependency declaration on hbase-thirdpar…

stoty:
…ty artifacts


stoty:
checkstyle fix

mnpoonia:
This looks good. So effectively now we will use hbase-thirdparty that comes with hbase version that a build is using.

stoty:
Yes.
I originally added the explicit dependency to satisfy the dependency check, but it has proven to be more trouble than it's worth.

mnpoonia (approved):

PR #1577, opened 2023-03-07
PHOENIX-5066 The TimeZone is incorrectly used during writing or readi…

stoty:
…ng data

Add the phoenix.query.applyTimeZoneDisplacement option, that causes the java.sql time types to be treated as local times.
This option is disabled by default, it needs to be enabled to get the new behaviour.


stoty:
For running tests on backports

PR #1586, opened 2023-03-23
PHOENIX-6560 Fix Spotbugs security errors with code SQL_NONCONSTANT_S…

kabhishek4:
…TRING_PASSED_TO_EXECUTE

### DIFF
--- a/phoenix-core/src/main/java/org/apache/phoenix/mapreduce/OrphanViewTool.java
+++ b/phoenix-core/src/main/java/org/apache/phoenix/mapreduce/OrphanViewTool.java
@@ -553,11 +553,8 @@ public class OrphanViewTool extends Configured implements Tool {
                 if (key.getTenantId() != null) {
                     delChLink.setString(++param, key.getTenantId());
                 }
-                if (key.getSchemaName() == null) {
-                    delChLink.setString(++param, key.getTableName());
-                } else {
-                    delChLink.setString(++param,  key.getSchemaName() + "." + key.getTableName());
-                }
+                delChLink.setString(++param, key.getSchemaName() == null ?
+                    key.getTableName() : (key.getSchemaName() + "." + key.getTableName()));
                 delChLink.execute();
             }
             phoenixConnection.commit();
--- a/phoenix-core/src/main/java/org/apache/phoenix/mapreduce/OrphanViewTool.java
+++ b/phoenix-core/src/main/java/org/apache/phoenix/mapreduce/OrphanViewTool.java
@@ -553,8 +553,8 @@ public class OrphanViewTool extends Configured implements Tool {
                 if (key.getTenantId() != null) {
                     delChLink.setString(++param, key.getTenantId());
                 }
-                delChLink.setString(++param, key.getSchemaName() == null ?
-                    key.getTableName() : (key.getSchemaName() + "." + key.getTableName()));
+                delChLink.setString(++param, key.getSchemaName() == null
+                    ? key.getTableName() : (key.getSchemaName() + "." + key.getTableName()));
                 delChLink.execute();
             }
             phoenixConnection.commit();
--- a/phoenix-core/src/main/java/org/apache/phoenix/mapreduce/OrphanViewTool.java
+++ b/phoenix-core/src/main/java/org/apache/phoenix/mapreduce/OrphanViewTool.java
@@ -438,32 +438,33 @@ public class OrphanViewTool extends Configured implements Tool {
     }
 
     private void removeLink(PhoenixConnection phoenixConnection, Key src, Key dst, PTable.LinkType linkType) throws Exception {
-        String delTable = ((linkType == PTable.LinkType.PHYSICAL_TABLE || linkType == PTable.LinkType.PARENT_TABLE) ?
-            SYSTEM_CATALOG_NAME : SYSTEM_CHILD_LINK_NAME);
+        String delTable = (linkType == PTable.LinkType.PHYSICAL_TABLE
+            || linkType == PTable.LinkType.PARENT_TABLE) ? SYSTEM_CATALOG_NAME
+            : SYSTEM_CHILD_LINK_NAME;
 
         String deleteQuery = String.format(" DELETE FROM %s WHERE " +  TENANT_ID + " %s  AND "
-            + TABLE_SCHEM + " %s AND " + TABLE_NAME + " = ? " + COLUMN_NAME + " %s AND "
+            + TABLE_SCHEM + " %s AND " + TABLE_NAME + " = ? AND " + COLUMN_NAME + " %s AND "
             + COLUMN_FAMILY + " = ? ", delTable,
-            (src.getTenantId() == null ? " IS NULL" : " = ? "),
-            (src.getSchemaName() == null ? " IS NULL " : " = ? "),
-            (dst.getTenantId() == null ? " IS NULL" : " = ?"));
+            src.getTenantId() == null ? " IS NULL" : " = ? ",
+            src.getSchemaName() == null ? " IS NULL " : " = ? ",
+            dst.getTenantId() == null ? " IS NULL" : " = ?");
 
-        try(PreparedStatement stmt = phoenixConnection.prepareStatement(deleteQuery)) {
+        try (PreparedStatement stmt = phoenixConnection.prepareStatement(deleteQuery)) {
             int param = 0;
             if (src.getTenantId() != null) {
-                stmt.setString(++param,src.getTenantId());
+                stmt.setString(++param, src.getTenantId());
             }
             if (src.getSchemaName() != null) {
-                stmt.setString(++param,src.getSchemaName());
+                stmt.setString(++param, src.getSchemaName());
             }
-            stmt.setString(++param,src.getTableName());
+            stmt.setString(++param, src.getTableName());
             if (dst.getTenantId() != null) {
-                stmt.setString(++param,dst.getTenantId());
+                stmt.setString(++param, dst.getTenantId());
             }
             if (dst.getSchemaName() == null) {
-                stmt.setString(++param,dst.getTableName());
+                stmt.setString(++param, dst.getTableName());
             } else {
-                stmt.setString(++param,dst.getSchemaName() + "." +dst.getTableName());
+                stmt.setString(++param, dst.getSchemaName() + "." + dst.getTableName());
             }
             stmt.execute();
             phoenixConnection.commit();
@@ -521,36 +522,41 @@ public class OrphanViewTool extends Configured implements Tool {
             }
         }
     }
+
+
     private void forcefullyDropView(PhoenixConnection phoenixConnection,
                                     Key key) throws Exception {
-        String deleteRowsFromCatalog = String.format("DELETE FROM " + SYSTEM_CATALOG_NAME + " WHERE " + TENANT_ID
-            + " %s  AND " + TABLE_SCHEM + " %s  AND " + TABLE_NAME + " =  ? ",
-            (key.getTenantId() == null ? " IS NULL" : " = ? "),
-            (key.getSchemaName() == null ? " IS NULL " : " = ? "));
-        String deleteRowsFromChildLink = String.format("DELETE FROM " + SYSTEM_CHILD_LINK_NAME +
-            " WHERE " + COLUMN_NAME + " %s  AND " +  COLUMN_FAMILY + " = ? ",
-            (key.getTenantId() == null ? " IS NULL" : " = ? "));
+        String deleteRowsFromCatalog = String.format("DELETE FROM " + SYSTEM_CATALOG_NAME
+                + " WHERE " + TENANT_ID + " %s  AND " + TABLE_SCHEM + " %s  AND "
+                + TABLE_NAME + " =  ? ",
+            key.getTenantId() == null ? " IS NULL" : " = ? ",
+            key.getSchemaName() == null ? " IS NULL " : " = ? ");
+        String deleteRowsFromChildLink = String.format("DELETE FROM " + SYSTEM_CHILD_LINK_NAME
+                + " WHERE " + COLUMN_NAME + " %s  AND " +  COLUMN_FAMILY + " = ? ",
+            key.getTenantId() == null ? " IS NULL" : " = ? ");
         try {
-            try (PreparedStatement delSysCat = phoenixConnection.prepareStatement(deleteRowsFromCatalog)) {
+            try (PreparedStatement delSysCat =
+                phoenixConnection.prepareStatement(deleteRowsFromCatalog)) {
                 int param = 0;
                 if (key.getTenantId() != null) {
                     delSysCat.setString(++param, key.getTenantId());
                 }
-                if(key.getSchemaName() != null) {
+                if (key.getSchemaName() != null) {
                     delSysCat.setString(++param, key.getSchemaName());
                 }
-                delSysCat.setString(++param,key.getTableName());
+                delSysCat.setString(++param, key.getTableName());
                 delSysCat.execute();
             }
-            try(PreparedStatement delChLink = phoenixConnection.prepareStatement(deleteRowsFromChildLink)) {
+            try (PreparedStatement delChLink =
+                phoenixConnection.prepareStatement(deleteRowsFromChildLink)) {
                 int param = 0;
-                if(key.getTenantId() != null) {
+                if (key.getTenantId() != null) {
                     delChLink.setString(++param,key.getTenantId());
                 }
                 if (key.getSchemaName() == null) {
-                    delChLink.setString(++param,key.getTableName());
+                    delChLink.setString(++param, key.getTableName());
                 } else {
-                    delChLink.setString(++param,(key.getSchemaName() + "." + key.getTableName()));
+                    delChLink.setString(++param,  key.getSchemaName() + "." + key.getTableName());
                 }
                 delChLink.execute();
             }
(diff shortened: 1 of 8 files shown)


stoty:
Please review the checkstyle errors.
The suggested policy is to observe the 100 character limit on the lines you've changed, but you don't need to fix the existing issues.

Some of the unit test errors seem to be relevant.

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/OrphanViewTool.java
```diff
@@ -500,18 +522,44 @@ else if (!clean){
...
+                delSysCat.execute();
+            }
+            try (PreparedStatement delChLink =
+                phoenixConnection.prepareStatement(deleteRowsFromChildLink)) {
+                int param = 0;
+                if (key.getTenantId() != null) {
+                    delChLink.setString(++param, key.getTenantId());
+                }
+                if (key.getSchemaName() == null) {
+                    delChLink.setString(++param, key.getTableName());
```
stoty, line 557: This looks looks too similar to the is null case.
It would be more readable without the if, and using the original tenary operator for the value instead.

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/OrphanViewTool.java
```diff
@@ -500,18 +522,44 @@ else if (!clean){
...
+                if (key.getSchemaName() != null) {
+                    delSysCat.setString(++param, key.getSchemaName());
+                }
+                delSysCat.setString(++param, key.getTableName());
+                delSysCat.execute();
+            }
+            try (PreparedStatement delChLink =
+                phoenixConnection.prepareStatement(deleteRowsFromChildLink)) {
+                int param = 0;
+                if (key.getTenantId() != null) {
```
stoty, line 553: These are in the wrong order.

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/index/IndexTool.java
```diff
@@ -598,18 +599,27 @@ private Job configureJobForPartialBuild() throws Exception {
...
-                            + ASYNC_REBUILD_TIMESTAMP + " BIGINT) WHERE " + TABLE_SCHEM
-                            + (schemaName != null && schemaName.length() > 0 ? "='" + schemaName + "'" : " IS NULL")
-                            + " and " + TABLE_NAME + " IN (" + StringUtils.join(",", quotedIndexes) + ")")) {
+
+            String query = String.format("SELECT MAX(" + ASYNC_REBUILD_TIMESTAMP + "), "
+                    + "MAX(" + INDEX_DISABLE_TIMESTAMP + ") FROM "
+                    + SYSTEM_CATALOG_NAME + " (" + ASYNC_REBUILD_TIMESTAMP
+                    + " BIGINT) WHERE " + TABLE_SCHEM +  " %s  AND " + TABLE_NAME + " IN ( %s )",
+                        (schemaName != null && schemaName.length() > 0) ? " = ? " : " IS NULL ",
+                StringUtils.join(",", quotedIndexes));
```
stoty, line 617: The safe thing to do here is to generate as many ? as there elements in the list, and call setXXX for each one.

phoenix-core/src/main/java/org/apache/phoenix/query/ConnectionQueryServicesImpl.java
```diff
@@ -3994,21 +3994,36 @@ protected PhoenixConnection upgradeSystemCatalogIfRequired(PhoenixConnection met
...
-                    TableDescriptorBuilder.SPLIT_POLICY + "='" + SystemStatsSplitPolicy.class.getName() +"'"
-                    );
+            String altCatQry = "ALTER TABLE "
+                + PhoenixDatabaseMetaData.SYSTEM_CATALOG + " SET "
+                + HConstants.VERSIONS + "= " + props.getInt(DEFAULT_SYSTEM_MAX_VERSIONS_ATTRIB,
+                QueryServicesOptions.DEFAULT_SYSTEM_MAX_VERSIONS) + ",\n"
+                + ColumnFamilyDescriptorBuilder.KEEP_DELETED_CELLS + "="
+                + props.getBoolean(DEFAULT_SYSTEM_KEEP_DELETED_CELLS_ATTRIB,
+                QueryServicesOptions.DEFAULT_SYSTEM_KEEP_DELETED_CELLS);
+            try (PreparedStatement altCatQryStmt = metaConnection.prepareStatement(altCatQry)) {
```
stoty, line 4004: No need for PreparedStatements if we don't set any parameters.

We could just put each query in a try-with-resources block that creates a (non prepared) metaStmt, to make sure that we don't let it hanging.
Just be careful to always re-create metaStmt when replace metaConnection.

phoenix-core/src/main/java/org/apache/phoenix/schema/MetaDataClient.java
```diff
@@ -1164,19 +1164,23 @@ private long updateStatisticsInternal(PName physicalName, PTable logicalTable, M
...
-                    + " IS NULL AND " + LAST_STATS_UPDATE_TIME + " IS NOT NULL";
-            ResultSet rs = connection.createStatement().executeQuery(query);
-
-            if (rs.next()) {
-                msSinceLastUpdate = rs.getLong(1) - rs.getLong(2);
+            String query = "SELECT CURRENT_DATE()," + LAST_STATS_UPDATE_TIME + " FROM "
+                + PhoenixDatabaseMetaData.SYSTEM_STATS_NAME
+                + " WHERE " + PHYSICAL_NAME + "= ?  AND " + COLUMN_FAMILY
+                + " IS NULL AND " + LAST_STATS_UPDATE_TIME + " IS NOT NULL";
+            try (PreparedStatement stmt = connection.prepareStatement(query)) {
```
stoty, line 1178: please use a unique name

phoenix-core/src/main/java/org/apache/phoenix/schema/transform/Transform.java
```diff
@@ -472,25 +473,41 @@ public static void doCutover(PhoenixConnection connection, SystemTransformRecord
...
 
         LOGGER.info("About to do cutover via " + changeTable);
         TableViewFinderResult childViewsResult = ViewUtil.findChildViews(connection, tenantId, schema, tableName);
         boolean wasCommit = connection.getAutoCommit();
         connection.setAutoCommit(false);
         List<TableInfo> viewsToUpdateCache = new ArrayList<>();
         try {
-            connection.createStatement().execute(changeTable);
-
+            try (PreparedStatement stmt = connection.prepareStatement(changeTable)) {
```
stoty, line 492: use a uniqie name

phoenix-core/src/main/java/org/apache/phoenix/schema/transform/Transform.java
```diff
@@ -472,25 +473,41 @@ public static void doCutover(PhoenixConnection connection, SystemTransformRecord
...
+                int param = 0;
+                if (tenantId == null) {
+                    stmt.setNull(++param, Types.VARCHAR);
```
stoty, line 495: Interesting.
setString(x, null) doesn't generate a "null" string ?

phoenix-core/src/main/java/org/apache/phoenix/schema/transform/Transform.java
```diff
@@ -520,13 +537,28 @@ public static void doCutover(PhoenixConnection connection, SystemTransformRecord
             int batchSize = 0;
             for (TableInfo view : childViewsResult.getLinks()) {
                 String changeView = String.format(changeViewStmt,
-                        (columnNames.size() > 0? "," + String.join(",", columnNames):""),
-                        (view.getTenantId()==null || view.getTenantId().length == 0? null: ("'" + Bytes.toString(view.getTenantId()) + "'")),
-                        (view.getSchemaName()==null || view.getSchemaName().length == 0? null : ("'" + Bytes.toString(view.getSchemaName()) + "'")),
-                        Bytes.toString(view.getTableName()),
-                        (columnValues.size() > 0? "," + String.join(",", columnValues):""));
+                    columnNames.size() > 0 ? "," + String.join(",", columnNames) : "",
```
stoty, line 540: Same list handling as before.
(I think we may already have some helper functions for that)

phoenix-core/src/main/java/org/apache/phoenix/util/UpgradeUtil.java
```diff
@@ -1823,6 +1823,25 @@ private ViewKey(String tenantId, String schema, String viewName) {
...
     }
 
+    private static String getTableRVCWithParam(List<String> tableNames) {
+        StringBuilder query = new StringBuilder("(");
+        for (int i = 0; i < tableNames.size(); i += 3) {
+            String tenantId = tableNames.get(i);
+            String schemaName = tableNames.get(i + 1);
+            String tableName = tableNames.get(i + 2);
+            query.append('(');
+            query.append(tenantId == null ? "null" : " ? ");
```
stoty, line 1833: wouldn't setNull work here ?

phoenix-core/src/main/java/org/apache/phoenix/util/UpgradeUtil.java
```diff
@@ -1823,6 +1823,25 @@ private ViewKey(String tenantId, String schema, String viewName) {
...
+            query.append(',');
+            query.append(schemaName == null ? "null" : " ? ");
+            query.append(',');
+            query.append("'" + tableName + "'");
```
stoty, line 1837: Why is this not parametrized ?

phoenix-core/src/main/java/org/apache/phoenix/util/UpgradeUtil.java
```diff
@@ -2031,16 +2067,27 @@ private static void upgradeDescVarLengthRowKeys(PhoenixConnection upgradeConn, P
...
-                           "'" + (theSchemaName == null ? StringUtil.EMPTY_STRING : theSchemaName) + "'," +
-                           "'" + theTableName + "'," +
-                           "TRUE)");
+                String upsSyscat = String.format("UPSERT INTO "
+                    + PhoenixDatabaseMetaData.SYSTEM_CATALOG_NAME
+                    + " (" + PhoenixDatabaseMetaData.TENANT_ID + ","
+                    + PhoenixDatabaseMetaData.TABLE_SCHEM + ","
+                    + PhoenixDatabaseMetaData.TABLE_NAME + ","
+                    + MetaDataEndpointImpl.ROW_KEY_ORDER_OPTIMIZABLE + " BOOLEAN"
+                    + ") VALUES ( ?, ?," + "'" + theTableName + "'," + "TRUE)");
```
stoty, line 2076: Why is this not parametrized ?

phoenix-core/src/main/java/org/apache/phoenix/util/UpgradeUtil.java
```diff
@@ -2163,45 +2210,53 @@ public static void upgradeDescVarLengthRowKeys(PhoenixConnection conn, List<Stri
...
-                lastTenantId = tenantId;
             }
-            PTable table = PhoenixRuntime.getTable(conn, fullTableName);
-            String tableTenantId = table.getTenantId() == null ? null : table.getTenantId().getString();
-            if (Objects.equal(lastTenantId, tableTenantId) && !table.rowKeyOrderOptimizable()) {
-                upgradeDescVarLengthRowKeys(upgradeConn, globalConn, table.getSchemaName().getString(), table.getTableName().getString(), false, bypassUpgrade);
-                wasUpgraded = true;
+            rs.close();
+            if (lastTenantId != null) {
+                conn.close();
```
stoty, line 2256: Could this be replaced with a try-with-resource ?

phoenix-core/src/main/java/org/apache/phoenix/util/UpgradeUtil.java
```diff
@@ -2563,13 +2618,17 @@ private static void updateIndexesSequenceIfPresent(PhoenixConnection connection,
...
         // create new entry with new schema format
-        String upsert = "UPSERT INTO " + PhoenixDatabaseMetaData.SYSTEM_SEQUENCE + " SELECT NULL,\'" + newSchemaName +
-            "\',\'" + newSequenceName
-                + "\'," + START_WITH + "," + CURRENT_VALUE + "," + INCREMENT_BY + "," + CACHE_SIZE + "," + MIN_VALUE
-                + "," + MAX_VALUE + "," + CYCLE_FLAG + "," + LIMIT_REACHED_FLAG + " FROM "
-                + PhoenixDatabaseMetaData.SYSTEM_SEQUENCE + " WHERE " + PhoenixDatabaseMetaData.TENANT_ID
-                + " IS NULL AND " + PhoenixDatabaseMetaData.SEQUENCE_SCHEMA + " = '" + oldSchemaName + "'";
-        connection.createStatement().executeUpdate(upsert);
+        String upsert = "UPSERT INTO " + PhoenixDatabaseMetaData.SYSTEM_SEQUENCE
+            + " SELECT NULL,\'" + newSchemaName + "\',\'" + newSequenceName
```
stoty, line 2622: shouldn't these be parametrized ?

stoty (changes requested):

phoenix-core/src/main/java/org/apache/phoenix/util/UpgradeUtil.java
kabhishek4: In this statement, a 'where' predicate, SEQUENCE_SCHEMA has already been parameterised. Can you please suggest which other predicate you are referring to?

phoenix-core/src/main/java/org/apache/phoenix/util/UpgradeUtil.java
kabhishek4: This has already been added into try-with-resource block which starts at line 2223.

stoty: Thanks.
I don't quite get what is happening with conn/globalConn here, but you are not chaning that part, so it's fine to leave it as it is.

phoenix-core/src/main/java/org/apache/phoenix/util/UpgradeUtil.java
stoty: newSchemaName and newSequencename.
Any concatenated non-constant string string is potentially vulnerable to sql injection.

These seem to be the schema and name of sequences.
While realistically it's unlikely to be a real security hole, 
We cannot guarantee without auditing a lot of code that the existing sequence names don't resolve to a string that can be used for second order sql injection.

kabhishek4: Thanks for the explanation. However, after parameterising the sequence name, spotbug bug tool stopped issuing error for this statement. Do you think we still need to consider this as a security hole?  Probably, there could be several other concatenations in this code and tool is not reporting warning/error. Moreover, this PR is for addressing SQL_NONCONSTANT_STRING_PASSED_TO_EXECUTE which is a high severity security error. Considering this please let me know if we want to cover this as part of this PR.

stoty: I am quite baffled why spotbug doesn't flag this.
I wouldn't expect spotbugs to actually parse the SQL.
TBH most of these fixes are closing theoretical holes rather than real explaoitable issues.
However, the point of doing is to make sure that we are protected even if other parts of the code let an exploit through, so I think that we should be thorough.

Besides, this a super simple fix, there are no nulls or variable length lists here.

stoty (changes requested):
This looks correct, but I can see a bit of room for improvements in readability.

kabhishek4:
retest

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/index/IndexTool.java
```diff
@@ -598,18 +599,31 @@ private Job configureJobForPartialBuild() throws Exception {
...
         private long getMaxRebuildAsyncDate(String schemaName, List<String> disableIndexes) throws SQLException {
-            Long maxRebuilAsyncDate=HConstants.LATEST_TIMESTAMP;
-            Long maxDisabledTimeStamp=0L;
-            if (disableIndexes == null || disableIndexes.isEmpty()) { return 0; }
+            Long maxRebuilAsyncDate = HConstants.LATEST_TIMESTAMP;
+            Long maxDisabledTimeStamp = 0L;
+            if (disableIndexes == null || disableIndexes.isEmpty()) {
+                return 0;
+            }
             List<String> quotedIndexes = new ArrayList<String>(disableIndexes.size());
```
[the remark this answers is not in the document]
kabhishek4: I have addressed this comment in somewhat different way. Added 
 void setQuoteInListElements(PreparedStatement ps, String unQuotedString,
        int index) 
which will add quotes to one of the elements and set it to the correct index of prepared statement. Using which one of the loops got eliminated. This version of setQuoteInListElements can further be overloaded as appropriate. I hope that's ok.

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/index/IndexTool.java
```diff
@@ -604,24 +604,19 @@ private long getMaxRebuildAsyncDate(String schemaName, List<String> disableIndex
...
-                QueryUtil.generateInListParams(quotedIndexes.size()));
+                QueryUtil.generateInListParams(disableIndexes.size()));
             try (PreparedStatement selSyscat = connection.prepareStatement(query)) {
                 int param = 0;
                 if (schemaName != null && schemaName.length() > 0) {
                     selSyscat.setString(++param, schemaName);
                 }
-                for (int i = 0; i < quotedIndexes.size(); i++) {
-                    selSyscat.setString(++param, quotedIndexes.get(i));
+                for (int i = 0; i < disableIndexes.size(); i++) {
```
stoty, line 618: While this works, it would be much cleaner if we  pushed this whole for loop into the utility function.

stoty:
started a new ci build, as the results have aged out.

stoty (approved):
+1 LGTM

PR #1587, opened 2023-03-30
PHOENIX-6821: Optimize batching in auto-commit mode

haridsv:
This is a cerry-pick of the squashed change from PR https://github.com/apache/phoenix/pull/1570.


tkhurana:
@haridsv can you please fix the checkstyle errors

haridsv:
@tkhurana All but one are on existing code that I just happened to touch/move around, but I fixed them anyway. The following is in a line that follows the existing style, so I didn't want to create a deviation and so fixed the existing lines as well.

PR #1589, opened 2023-04-07
PHOENIX-6918 :- ScanningResultIterator should not retry when the query times out

lokiore:


phoenix-core/src/main/java/org/apache/phoenix/iterate/ScanningResultIterator.java
```diff
@@ -173,6 +177,12 @@ public Tuple next() throws SQLException {
         try {
             Result result = scanner.next();
             while (result != null && (result.isEmpty() || isDummy(result))) {
+                long timeOutForScan = maxQueryEndTime - EnvironmentEdgeManager.currentTimeMillis();
+                if (timeOutForScan < 0) {
+                    throw new SQLExceptionInfo.Builder(OPERATION_TIMED_OUT).setMessage(
+                            ". Query couldn't be completed in the allotted time at ScanningResultIterator: "
```
gjacoby126, line 183: nit: We probably shouldn't be mentioning Phoenix internals like ScanningResultIterator in an error message that goes back to end-users.

phoenix-core/src/it/java/org/apache/phoenix/end2end/MapReduceIT.java
```diff
@@ -135,6 +164,73 @@ private void createAndTestJob(Connection conn, String s, double v, String tenant
 
     }
 
+    private void createPagedJobAndTestFailedJobDueToTimeOut(Connection conn, String s, double v, String tenantId,
```
gjacoby126, line 167: nit: function-level variables should have meaningful names more than 1 character (not "s" or "v"). (For temp variables in a loop this would be fine)

gjacoby126 (changes requested):
Looks good, just a couple of nits

lokiore:
@gjacoby126, addressed the nits. 
Thanks for reviewing !

lokiore:
@gjacoby126  Thanks for merging, do I need to create a separate PR for 5.1 or this can be directly committed !?

PR #1596, opened 2023-04-28
PHOENIX-6944 Randomize mapper task ordering for MR tools

stoty:


stoty:
The TableSnapshotReadsMapReduceIT failure looks relevant.
I need to check if it's a test assumption or a real regression.

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/PhoenixInputFormat.java
```diff
@@ -87,74 +88,106 @@ public List<InputSplit> getSplits(JobContext context) throws IOException, Interr
...
+                                        BaseScannerRegionObserver.EXPECTED_UPPER_REGION_KEY)));
+                        }
                     }
+
+                    psplits.add(new PhoenixInputSplit(scans, regionSize, regionLocation));
                 }
+            }
 
-                psplits.add(new PhoenixInputSplit(scans, regionSize, regionLocation));
+            if (PhoenixConfigurationUtil.isMRRandomizeMapperExecutionOrder(config)) {
```
stoty, line 183: This is the only code change, the rest of the changes in thsis method are just reformats

phoenix-core/src/main/java/org/apache/phoenix/mapreduce/PhoenixInputFormat.java
```diff
@@ -87,74 +88,106 @@ public List<InputSplit> getSplits(JobContext context) throws IOException, Interr
...
+        double totalLength = splits.stream().mapToDouble(s -> {
+            try {
+                return (double) s.getLength();
+            } catch (IOException | InterruptedException e1) {
+                return defaultLength;
+            }
+        }).sum();
+        long avgLength = (long) (totalLength / splits.size());
+        splits.stream().forEach(s -> ((PhoenixInputSplit) s)
+                .setLength(avgLength + ThreadLocalRandom.current().nextInt(10000)));
```
richardantal, line 112: I assume that it performs better because there aren't big differences in the split lengths, right?

The sum of the new split lengths will be bigger than the original, why this isn't it a problem?

stoty: > I assume that it performs better because there aren't big differences in the split lengths, right?

The diffrences in split length don't matter for performance, as long we make sure that the splits for the same RS are randomly distributed.
I chose a relatively small random length so that we don't skew the reported completion performances too much, this way there will be less than 1% difference between the splits (as we originallly set the length to the size of the region in bytes)


> The sum of the new split lengths will be bigger than the original, why this isn't it a problem?

The absolute split lengths have no significance.
Only the relative split lengths are used for 
a.) determining the execution order
b.) computing the displayed completion percent 

We are already using wildly inaccurate values, as we use the region size instead of the guidepost size even when we split by guideposts, and it hasn't caused any problems.

richardantal (approved):
LGTM

PR #1595, opened 2023-04-27
PHOENIX-6932 Update LAST_DDL_TIMESTAMP for index table when add/alter indexes

shahrs87:

PR #1597, opened 2023-05-02
PHOENIX-6897 Filters on unverified index rows return wrong result

tkhurana:


tkhurana:
2 failures in AlterTableIT.testSetPropertyDoesntUpdateDDLTimestamp are fixed in #1595

phoenix-core/src/main/java/org/apache/phoenix/index/GlobalIndexChecker.java
```diff
@@ -202,7 +208,42 @@ private void init() throws IOException {
             else {
                 pageSize = Long.MAX_VALUE;
             }
+
+            Filter filter = scan.getFilter();
+            if (filter instanceof PagingFilter) {
```
kadirozde, line 213: We may also need to wrap non paging filters. So the same logic should be applied even when the scan filter is not an instance of PagingFilter.

kadirozde (changes requested):
Looks very good overall. I requested a couple of changes.

PR #1598, opened 2023-05-02
PHOENIX-6907 Explain Plan should output region locations with servers

virajjasani:
Query syntax: `EXPLAIN WITH REGIONS {query}`

Example:

Assuming the scan is prepared for 4 region locations,

When `phoenix.max.region.locations.size.explain.plan` is set as `2`, only 2 region locations are printed:
```
CLIENT PARALLEL 4-WAY RANGE SCAN OVER N000001:N000002 [1,'a'] - [1,'b']
    SERVER MERGE [0.K3]
    SERVER FILTER BY FIRST KEY ONLY
 (region locations = [region=N000001:N000002,,1683068961616.4773aca00009beabd2316e0be5828ffa., hostname=localhost,55546,1683068927671, seqNum=9, region=N000001:N000002,e\x00\x00\x00\x00\x00\x00\x00\x00\x00,1683068961616.77243f1c70242067f53018949e146393., hostname=localhost,55546,1683068927671, seqNum=9]...total size = 4) 
CLIENT MERGE SORT
```

With default value of `phoenix.max.region.locations.size.explain.plan` as `5`, all region locations are printed:
```
CLIENT PARALLEL 4-WAY RANGE SCAN OVER N000001:N000002 [1,'a'] - [1,'b']
    SERVER MERGE [0.K3]
    SERVER FILTER BY FIRST KEY ONLY
 (region locations = [region=N000001:N000002,,1683068843354.6f20f10367637072dabf805a38ac8d55., hostname=localhost,55344,1683068805683, seqNum=9, region=N000001:N000002,e\x00\x00\x00\x00\x00\x00\x00\x00\x00,1683068843354.fdfd4eb63d0fff6d80da1d08726bfeb0., hostname=localhost,55344,1683068805683, seqNum=9, region=N000001:N000002,i\x00\x00\x00\x00\x00\x00\x00\x00\x00,1683068843354.effe18db5f4a093d005706910436ed18., hostname=localhost,55344,1683068805683, seqNum=9, region=N000001:N000002,o\x00\x00\x00\x00\x00\x00\x00\x00\x00,1683068843354.d7d0cec570fe9c626d39a0b2075918d2., hostname=localhost,55344,1683068805683, seqNum=9]) 
CLIENT MERGE SORT
```

### DIFF
--- a/phoenix-core/src/main/antlr3/PhoenixSQL.g
+++ b/phoenix-core/src/main/antlr3/PhoenixSQL.g
@@ -152,6 +152,7 @@ tokens
     REVOKE = 'revoke';
     SHOW = 'show';
     UNCOVERED = 'uncovered';
+    REGIONS = 'regions';
 }
 
 
@@ -212,6 +213,7 @@ import org.apache.phoenix.util.SchemaUtil;
 import org.apache.phoenix.parse.LikeParseNode.LikeType;
 import org.apache.phoenix.trace.util.Tracing;
 import org.apache.phoenix.parse.AddJarsStatement;
+import org.apache.phoenix.parse.ExplainType;
 }
 
 @lexer::header {
@@ -452,7 +454,14 @@ oneStatement returns [BindableStatement ret]
 finally{ contextStack.pop(); }
     
 explain_node returns [BindableStatement ret]
-    :   EXPLAIN q=oneStatement {$ret=factory.explain(q);}
+    :   EXPLAIN (w=WITH)? (r=REGIONS)? q=oneStatement
+     {
+        if ((w==null && r!=null) || (w!=null && r==null)) {
+            throw new RuntimeException("Valid usage: EXPLAIN {query} OR EXPLAIN WITH REGIONS {query}");
+        }
+        ret = (w==null && r==null) ? factory.explain(q, ExplainType.DEFAULT)
+         : factory.explain(q, ExplainType.WITH_REGIONS);
+     }
     ;
 
 // Parse a create table statement.
--- a/phoenix-core/src/main/java/org/apache/phoenix/iterate/ExplainTable.java
+++ b/phoenix-core/src/main/java/org/apache/phoenix/iterate/ExplainTable.java
@@ -21,6 +21,7 @@ import java.io.IOException;
 import java.sql.SQLException;
 import java.text.Format;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Collections;
 import java.util.HashSet;
 import java.util.Iterator;
@@ -156,6 +157,8 @@ public abstract class ExplainTable {
         }
         byte[] currentKey = startKey;
         try (Table table = context.getConnection().getQueryServices().getTable(tableName)) {
+            // include all regions that include key range from the given start key
+            // and end key
             do {
                 HRegionLocation regionLocation =
                         table.getRegionLocator().getRegionLocation(currentKey, reload);
@@ -451,7 +454,9 @@ public abstract class ExplainTable {
 
         @Override
         public int hashCode() {
-            return 0;
+            int result = Arrays.hashCode(startKey);
+            result = 31 * result + Arrays.hashCode(endKey);
+            return result;
         }
     }
 
--- a/phoenix-core/src/main/java/org/apache/phoenix/iterate/ExplainTable.java
+++ b/phoenix-core/src/main/java/org/apache/phoenix/iterate/ExplainTable.java
@@ -17,24 +17,15 @@
  */
 package org.apache.phoenix.iterate;
 
-import java.io.IOException;
 import java.text.Format;
-import java.util.ArrayList;
 import java.util.Collections;
-import java.util.HashSet;
 import java.util.Iterator;
 import java.util.List;
 import java.util.NoSuchElementException;
 import java.util.Set;
 
-import org.apache.hadoop.hbase.HConstants;
-import org.apache.hadoop.hbase.HRegionLocation;
-import org.apache.hadoop.hbase.TableName;
-import org.apache.hadoop.hbase.client.Connection;
-import org.apache.hadoop.hbase.client.ConnectionFactory;
 import org.apache.hadoop.hbase.client.Consistency;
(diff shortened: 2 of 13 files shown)


phoenix-core/src/main/antlr3/PhoenixSQL.g
```diff
@@ -452,7 +454,14 @@ oneStatement returns [BindableStatement ret]
 finally{ contextStack.pop(); }
     
 explain_node returns [BindableStatement ret]
-    :   EXPLAIN q=oneStatement {$ret=factory.explain(q);}
+    :   EXPLAIN (w=WITH)? (r=REGIONS)? q=oneStatement
```
jpisaac, line 457: Would it make sense to not change what the EXPLAIN statement generates. But have all plan info/attributes generated all the time.

And have different kinds of plan output generators. 
We may want to look at how other DBs have implemented it.

virajjasani: Sounds good, let me try this tomorrow i.e. let the source generate all info in the explain plan including region info, but it's upto `ExecutableExplainStatement` to control what to print.

virajjasani:
@jpisaac addressed review comments

phoenix-core/src/main/java/org/apache/phoenix/iterate/ExplainTable.java
```diff
@@ -46,27 +61,37 @@
...
+    public static final String REGION_LOCATIONS = " (region locations = ";
+
     protected final StatementContext context;
     protected final TableRef tableRef;
     protected final GroupBy groupBy;
     protected final OrderBy orderBy;
     protected final HintNode hint;
     protected final Integer limit;
     protected final Integer offset;
+    private final ExecutorService executorService;
```
jpisaac, line 94: Why do we need an executor and async processing?

jpisaac (commented):
Mostly looks good, but not sure about the async calls?

phoenix-core/src/main/java/org/apache/phoenix/iterate/ExplainTable.java
virajjasani: this is because we want to have our own timeouts rather than relying on hbase client timeouts. 
if for some reason, we can't get region location due to some transient issue, client will keep retrying based on retry configs, but it would not be appropriate to block output of explain plan just because of that.

phoenix-core/src/main/java/org/apache/phoenix/iterate/ExplainTable.java
```diff
@@ -284,6 +367,102 @@ protected void explain(String prefix, List<String> planSteps,
...
+                                    List<List<Scan>> scansList) {
+        Future<String> task = executorService.submit(
+                () -> getRegionLocationsForExplainPlan(explainPlanAttributesBuilder, scansList));
+        try {
+            String regionLocations = task.get(5, TimeUnit.SECONDS);
+            planSteps.add(regionLocations);
+        } catch (InterruptedException | ExecutionException | TimeoutException e) {
+            LOGGER.error("Unable to retrieve region locations task result in 5s", e);
+            task.cancel(true);
+        }
```
virajjasani, line 388: this would be our own timeout, and if we can't get region locations on time, we rather cancel the task than letting hbase client retry for very long time.

@jpisaac does this sound good to you?

jpisaac: Creating an executor service for every query, might not be the right thing.

virajjasani:
latest build results: https://ci-hadoop.apache.org/job/Phoenix/job/Phoenix-PreCommit-GitHub-PR/job/PR-1598/13/PR_20General_20Check_20Report/

phoenix-core/src/main/java/org/apache/phoenix/iterate/ExplainTable.java
```diff
@@ -284,6 +355,106 @@ protected void explain(String prefix, List<String> planSteps,
...
+                return false;
+            }
+            RegionBoundary that = (RegionBoundary) o;
+            return Bytes.compareTo(startKey, that.startKey) == 0
+                    && Bytes.compareTo(endKey, that.endKey) == 0;
+        }
+
+        @Override
+        public int hashCode() {
+            return 0;
```
jpisaac, line 454: why hashcode of zero?

phoenix-core/src/main/java/org/apache/phoenix/iterate/ExplainTable.java
```diff
@@ -111,8 +126,63 @@ private String explainSkipScan() {
...
+                    regionLocations.add(regionLocation);
+                    regionBoundaries.add(regionBoundary);
+                }
+                currentKey = regionLocation.getRegion().getEndKey();
+                // condition1 = currentKey != END_ROW_KEY
+                // condition2 = endKeyIsEndOfTable == true
+                // condition3 = currentKey < endKey
+                // condition4 = includeEndKey == true
+                // condition5 = currentKey == endKey
+                // while (condition1 && (condition2 || condition3 || (condition4 && condition5)))
```
jpisaac, line 175: nit: Might be easier on the reader if it is in plain layman's language

jpisaac (commented):
Done with the review.Thanks

phoenix-core/src/main/java/org/apache/phoenix/iterate/ExplainTable.java
virajjasani: since byte[] can't be compared for hashcode (as they don't have `hashcode()` impl), i had to use unique value.
if we were doing string based comparison, then i would have used string's hashcode and equals implementations directly.

phoenix-core/src/main/java/org/apache/phoenix/iterate/ExplainTable.java
virajjasani: done, before do/while loop

jpisaac (approved):
LGTM +1 Thanks @virajjasani

PR #1602, opened 2023-05-08
PHOENIX-6949 Use j.u.Date.getTimeOffset() for displacement calculation

stoty:


stoty:
Amrked as draft, as I am not sure if / how we ought to cache the local TZ.

PR #1603, opened 2023-05-09
PHOENIX-6854 Salted global indexes do not work for queries with uncov…

tkhurana:
…ered columns


tkhurana:
One failure is being fixed in #1633

PR #1604, opened 2023-05-11
PHOENIX-6953 Creating indexes on a table with old indexing leads to i…

stoty:
…nconsistent co-processors

### DIFF
--- a/phoenix-core/src/it/java/org/apache/phoenix/end2end/CreateTableIT.java
+++ b/phoenix-core/src/it/java/org/apache/phoenix/end2end/CreateTableIT.java
@@ -1622,6 +1622,7 @@ public class CreateTableIT extends ParallelStatsDisabledIT {
         }
     }
 
+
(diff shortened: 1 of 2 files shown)


gjacoby126 (changes requested):
Change looks OK, but there are test failures on GlobalMutableTxIndexIT and UpsertSelectIT that look related.

gjacoby126 (changes requested):
Change looks OK, but there are test failures on GlobalMutableTxIndexIT and UpsertSelectIT that look related.

stoty:
I was unable to repro those, and I'm pretty sure both are unrelated

stoty:
However, we also call the function from updateDescriptorForTx(), and that one didn't check for old vs new indexing.
It doesn't matter for the current master, as changing transactional status is disabled in both directions, but it may matter for an 5.1 backport, or in the future when it is re-enabled.
Please take another look, @gjacoby126 .

PR #1606, opened 2023-05-12
PHOENIX-6955 The value of region metric--lastMajorCompactionAge is wrong，when the hfile is buckload file

Kerasone:


phoenix-core/src/main/java/org/apache/phoenix/mapreduce/MultiHfileOutputFormat.java
```diff
@@ -265,7 +265,8 @@ private WriterLength getNewWriter(final String tableName , byte[] family, Config
                                         .withBytesPerCheckSum(CompatUtil.getBytesPerChecksum(conf))
                                         .withBlockSize(blockSize)
                                         .withDataBlockEncoding(encoding)
-                                        .withCellComparator(CellComparatorImpl.COMPARATOR);
+                                        .withCellComparator(CellComparatorImpl.COMPARATOR)
+                                        .withCreateTime(System.currentTimeMillis());;
```
gjacoby126, line 269: This should probably be EnvironmentEdgeManager.currentTime() so that tests can override the timestamp correctly.

gjacoby126 (changes requested):
Thanks for the patch. Could we also please have a simple test to confirm that the bug's fixed?

Kerasone:
I'm so sorry,I'll do it quickly,thanks for your help



---- Replied Message ----
| From | Geoffrey ***@***.***> |
| Date | 05/19/2023 03:57 |
| To | ***@***.***> |
| Cc | ***@***.***>***@***.***> |
| Subject | Re: [apache/phoenix] PHOENIX-6955 The value of region metric--lastMajorCompactionAge is wrong，when the hfile is buckload file (PR #1606) |

@gjacoby126 requested changes on this pull request.

Thanks for the patch. Could we also please have a simple test to confirm that the bug's fixed?

In phoenix-core/src/main/java/org/apache/phoenix/mapreduce/MultiHfileOutputFormat.java:

> @@ -265,7 +265,8 @@ private WriterLength getNewWriter(final String tableName , byte[] family, Config
                                         .withBytesPerCheckSum(CompatUtil.getBytesPerChecksum(conf))
                                         .withBlockSize(blockSize)
                                         .withDataBlockEncoding(encoding)
-                                        .withCellComparator(CellComparatorImpl.COMPARATOR);
+                                        .withCellComparator(CellComparatorImpl.COMPARATOR)
+                                        .withCreateTime(System.currentTimeMillis());;


This should probably be EnvironmentEdgeManager.currentTime() so that tests can override the timestamp correctly.

PR #1607, opened 2023-05-15
PHOENIX-6910 Scans created during query compilation and execution aga…

stoty:
…inst salted tables need to be more resilient

add test from PHOENIX-4096

### DIFF
--- a/phoenix-core/src/it/java/org/apache/phoenix/end2end/salted/SaltedTableMergeBucketsIT.java
+++ b/phoenix-core/src/it/java/org/apache/phoenix/end2end/salted/SaltedTableMergeBucketsIT.java
@@ -1,0 +1,452 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.phoenix.end2end.salted;
+
+import static org.apache.phoenix.util.TestUtil.TEST_PROPERTIES;
+import static org.apache.phoenix.util.TestUtil.assertResultSet;
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertTrue;
+
+import java.sql.Connection;
+import java.sql.DriverManager;
+import java.sql.ResultSet;
+import java.sql.SQLException;
+import java.sql.Statement;
+import java.sql.Types;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.List;
+import java.util.Properties;
+import java.util.Set;
+import java.util.stream.Collectors;
+import java.util.stream.IntStream;
+
+import org.apache.hadoop.hbase.TableName;
+import org.apache.hadoop.hbase.client.Admin;
+import org.apache.hadoop.hbase.client.RegionInfo;
+import org.apache.phoenix.end2end.ParallelStatsDisabledIT;
+import org.apache.phoenix.end2end.ParallelStatsDisabledTest;
+import org.apache.phoenix.query.QueryServices;
+import org.apache.phoenix.schema.SortOrder;
+import org.apache.phoenix.schema.types.PDataType;
+import org.apache.phoenix.schema.types.PDecimal;
+import org.apache.phoenix.schema.types.PInteger;
+import org.apache.phoenix.schema.types.PLong;
+import org.apache.phoenix.thirdparty.com.google.common.collect.Lists;
+import org.apache.phoenix.thirdparty.com.google.common.collect.Sets;
+import org.apache.phoenix.util.PropertiesUtil;
+import org.apache.phoenix.util.QueryBuilder;
+import org.apache.phoenix.util.TestUtil;
+import org.junit.Test;
+import org.junit.experimental.categories.Category;
+
+@Category(ParallelStatsDisabledTest.class)
+public class SaltedTableMergeBucketsIT extends ParallelStatsDisabledIT {
+
+    @Test
+    public void testWithVariousPKTypes() throws Exception {
+
+        int[] variousSaltBuckets = new int[] { 11, 23, 31 };
+        SortOrder[][] sortOrdersCase1 =
+                new SortOrder[][] { { SortOrder.ASC, SortOrder.ASC, SortOrder.ASC },
+                        { SortOrder.ASC, SortOrder.ASC, SortOrder.DESC },
+                        { SortOrder.ASC, SortOrder.DESC, SortOrder.ASC },
+                        { SortOrder.ASC, SortOrder.DESC, SortOrder.DESC },
+                        { SortOrder.DESC, SortOrder.ASC, SortOrder.ASC },
+                        { SortOrder.DESC, SortOrder.ASC, SortOrder.DESC },
+                        { SortOrder.DESC, SortOrder.DESC, SortOrder.ASC },
+                        { SortOrder.DESC, SortOrder.DESC, SortOrder.DESC } };
+
+        SortOrder[][] sortOrdersCase2 =
+                new SortOrder[][] { { SortOrder.ASC, SortOrder.ASC, SortOrder.ASC, SortOrder.ASC },
+                        { SortOrder.ASC, SortOrder.ASC, SortOrder.ASC, SortOrder.DESC },
+                        { SortOrder.ASC, SortOrder.ASC, SortOrder.DESC, SortOrder.ASC },
+                        { SortOrder.ASC, SortOrder.ASC, SortOrder.DESC, SortOrder.DESC },
+                        { SortOrder.ASC, SortOrder.DESC, SortOrder.DESC, SortOrder.ASC },
+                        { SortOrder.ASC, SortOrder.DESC, SortOrder.DESC, SortOrder.DESC },
+                        { SortOrder.ASC, SortOrder.DESC, SortOrder.ASC, SortOrder.ASC },
+                        { SortOrder.ASC, SortOrder.DESC, SortOrder.ASC, SortOrder.DESC },
+                        { SortOrder.DESC, SortOrder.ASC, SortOrder.ASC, SortOrder.ASC },
+                        { SortOrder.DESC, SortOrder.ASC, SortOrder.ASC, SortOrder.DESC },
+                        { SortOrder.DESC, SortOrder.ASC, SortOrder.DESC, SortOrder.ASC },
+                        { SortOrder.DESC, SortOrder.ASC, SortOrder.DESC, SortOrder.DESC },
+                        { SortOrder.DESC, SortOrder.DESC, SortOrder.ASC, SortOrder.ASC },
+                        { SortOrder.DESC, SortOrder.DESC, SortOrder.ASC, SortOrder.DESC },
+                        { SortOrder.DESC, SortOrder.DESC, SortOrder.DESC, SortOrder.ASC },
+                        { SortOrder.DESC, SortOrder.DESC, SortOrder.DESC, SortOrder.DESC } };
+
+        for (int index = 0; index < sortOrdersCase1.length; index++) {
+            // Test Case 1: PK1 = Bigint, PK2 = Decimal, PK3 = Bigint
+            String baseTableNameCase1 = generateUniqueName();
+            PDataType[] testIntDecIntPKTypes =
+                    new PDataType[] { PLong.INSTANCE, PDecimal.INSTANCE, PLong.INSTANCE };
+            long nowTime1 =
+                    createTableCase1(baseTableNameCase1, variousSaltBuckets[0],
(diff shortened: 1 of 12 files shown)


phoenix-core/src/main/java/org/apache/phoenix/iterate/BaseResultIterators.java
```diff
@@ -1119,12 +1124,21 @@ private List<List<Scan>> getParallelScans(byte[] startKey, byte[] stopKey) throw
...
+                                keyOffset, false, splitPostfix, getTable().getBucketNum());
+                    for (Scan newScan : newScans) {
+                        if (newScan != null) {
+                            ScanUtil.setLocalIndexAttributes(newScan, keyOffset,
+                                regionInfo.getStartKey(), regionInfo.getEndKey(),
+                                newScan.getStartRow(), newScan.getStopRow());
+                            //TODO why are we passing endKey instead of startKey here ?
+                            scans =
+                                    addNewScan(parallelScans, scans, newScan, currentGuidePostBytes,
+                                        false, regionLocation);
```
virajjasani, line 1138: here `currentGuidePostBytes` could be less than `endKey` right?

virajjasani (commented):
Left minor comments, looks good overall

virajjasani:
FYI @kadirozde @jpisaac @tkhurana you might be interested, this is resolving the scan boundary problem for salted tables after split/merge.

phoenix-core/src/it/java/org/apache/phoenix/end2end/salted/SaltedTableMergeBucketsIT.java
```diff
@@ -0,0 +1,103 @@
...
+import org.apache.phoenix.end2end.ParallelStatsDisabledIT;
+import org.apache.phoenix.end2end.ParallelStatsDisabledTest;
+import org.apache.phoenix.query.QueryServices;
+import org.apache.phoenix.util.PropertiesUtil;
+import org.apache.phoenix.util.TestUtil;
+import org.junit.Test;
+import org.junit.experimental.categories.Category;
+
+@Category(ParallelStatsDisabledTest.class)
+public class SaltedTableMergeBucketsIT extends ParallelStatsDisabledIT {
```
virajjasani, line 42: shall we make this `SaltedTableSplitMergeBucketsIT` and add another test that does splitting of salted table region and then verify the result of SELECT queries against randomly picked key values from UPSERT queries?

phoenix-core/src/it/java/org/apache/phoenix/end2end/salted/SaltedTableMergeBucketsIT.java
```diff
@@ -0,0 +1,103 @@
...
+            admin.mergeRegionsAsync(regions.get(i).getEncodedNameAsBytes(),
+                regions.get(i + 1).getEncodedNameAsBytes(), false).get();
+        }
+        ResultSet rs =
+                statement.executeQuery("select c1,c2,c3 from " + tableName
+                        + " where c1='20191211' and c2 like '%HORTONWORKS_WEEKLY_TEST%'");
+        assertTrue(rs.next());
+        rs =
+                statement.executeQuery("select c1,c2,c3 from " + tableName
+                        + " where c1='20191217' and c2 like '%HORTONWORKS_WEEKLY_TEST%'");
```
virajjasani, line 100: shall we pick random c1 values from a list of values instead of only with `20191211` and `20191217`?

we can create a list with multiple values and try using UPSERT in a loop against the list, that will reduce num of unique UPSERT statements we need to run in the test and also facilitate querying against any random c1 values from the given list.

stoty: Random values in tests are fishy.
We could increase the number of rows, and check that we get back each of them.
It's a bit tricky, as the nature of hashing makes it hard to make sure that we cover all buckets.

stoty: The now generates and checks 100 rows, that should be enough to hit all buckets (statistically).

phoenix-core/src/it/java/org/apache/phoenix/end2end/salted/SaltedTableMergeBucketsIT.java
```diff
@@ -0,0 +1,103 @@
...
+            " upsert into " + tableName + " values('20191225','HORTONWORKS_WEEKLY_TEST','v3')");
+        connection.commit();
+
+        Admin admin =
+                driver.getConnectionQueryServices(getUrl(), TestUtil.TEST_PROPERTIES).getAdmin();
+        List<RegionInfo> regions = admin.getRegions(TableName.valueOf(tableName));
+        for (int i = 0; i < regions.size() - 1; i+=2) {
+            admin.mergeRegionsAsync(regions.get(i).getEncodedNameAsBytes(),
+                regions.get(i + 1).getEncodedNameAsBytes(), false).get();
+        }
```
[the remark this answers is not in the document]
stoty: Thanks, applied.

phoenix-core/src/it/java/org/apache/phoenix/end2end/salted/SaltedTableMergeBucketsIT.java
```diff
@@ -0,0 +1,103 @@
...
+    public void testQuery() throws Exception {
+        String tableName = generateUniqueName();
+
+        Properties props = PropertiesUtil.deepCopy(TEST_PROPERTIES);
+        props.put(QueryServices.FORCE_ROW_KEY_ORDER_ATTRIB, Boolean.FALSE.toString());
+        Connection connection = DriverManager.getConnection(getUrl(), props);
+        Statement statement = connection.createStatement();
+        statement.execute("CREATE TABLE " + tableName
+                + " (c1 VARCHAR NOT NULL, c2 VARCHAR NOT NULL, c3 VARCHAR NOT NULL,"
+                + " CONSTRAINT pk PRIMARY KEY(c1,c2,c3)) SALT_BUCKETS=11");
```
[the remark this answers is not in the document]
stoty: Done

phoenix-core/src/it/java/org/apache/phoenix/end2end/salted/SaltedTableMergeBucketsIT.java
stoty: I have made the test more exhaustive, to make sure that we hit all buckets.
We could add a test for splits, but AFAIK there is no problem with splits.
This specific problem only manifests on merges.

phoenix-core/src/it/java/org/apache/phoenix/end2end/salted/SaltedTableMergeBucketsIT.java
```diff
@@ -0,0 +1,103 @@
...
+                regions.get(i + 1).getEncodedNameAsBytes(), false).get();
+        }
+        ResultSet rs =
+                statement.executeQuery("select c1,c2,c3 from " + tableName
+                        + " where c1='20191211' and c2 like '%HORTONWORKS_WEEKLY_TEST%'");
+        assertTrue(rs.next());
+        rs =
+                statement.executeQuery("select c1,c2,c3 from " + tableName
+                        + " where c1='20191217' and c2 like '%HORTONWORKS_WEEKLY_TEST%'");
+        assertTrue(rs.next());
```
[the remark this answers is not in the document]
stoty: Done.
Also added checks to make sure that we get the correct row.

phoenix-core/src/main/java/org/apache/phoenix/iterate/BaseResultIterators.java
stoty: Thanks for pointing this out, @virajjasani .
This was actually a bug in my patch: 
Even if the new code split a scan because of crossing bucket boundaries, we still used the un-split region's key, which meant that we may have grouped different buckets together, which would have broken ordering.

I have removed the endKey parameter, and now use the new scan's end key.
I have also removed some logic that used the current region's start key as the end key, as it was unneccessary, and incorrect in the merged region case.

stoty:
Hmm.
Even with my last change, @jpisaac 's test fail.
Looking into it.

stoty:
I have added @jpisaac 's test, and fixed the error it deteceted.

stoty:
The current version mostly works, but my refactoring has broken query statistics.
I hope to fix that later today.

stoty:
Pushed fix for salted tables.
There is still a local index regression I am chasing.

stoty:
All relevant test pass now.
I think that the FailoverPhoenixConnectionIT is unrelated.
Can you take a look now @virajjasani @jpisaac @kadirozde @tkhurana ?

stoty:
Hmm.
I think my changes are confusing the start and end of region/bucket flags.
Gonna review and push a fix.
If you haven't started reviewing yet, hold off for a bit.

stoty:
Now it should be really ready.
The stats=enabled case wasn't tested, and was broken.

stoty:
Can you review this now that hopefully both cases are working @jpisaac , @virajjasani  ?

virajjasani:
somehow jenkins didn't trigger new build since May 25, some issues with build maybe.

just triggered a new one [here](https://ci-hadoop.apache.org/job/Phoenix/job/Phoenix-PreCommit-GitHub-PR/job/PR-1607/)

stoty:
That CI run has hung.
I'm starting another.
(We have a number of unresolved flakey tests related to the HA feature)

virajjasani:
oops, it seems some tests are taking longer and hence the build is timing out. 
@stoty do you have any vm or docker setup where you think you could run mvn verify and keep it running for a bit longer? because in this case, we might be able to get better insights from local runs i think.

stoty:
If a test run doesn't finish in 3-4 hours then it never will.
I've had trouble running tests on either too fast or too slow machines.
Starting a local test now.

stoty:
I have attached local-verify-run.gz to the ticket.
It only fails on the known flakey permission ITs.
(It seems that we have tests that tend to fail on slow machines, and also ones that tend to fail on fast ones)

virajjasani:
Finally build runs were successful, let me take a look tomorrow.

At high level, IIUC, ParallelScansCollector is going to keep adding scan objects and all places where we use parallel scans, we are going to use this new class. New version of intersectScan is going to include scan boundaries from salt buckets. Is this correct?

stoty:
Yes, that's correct.
ParallelScansCollector just simplifies the messy logic where we had to split the lists at the next entry, and maintain the state in object variables.
I have also consolidated some of the very dispersed logic into fewer places.

phoenix-core/src/main/java/org/apache/phoenix/compile/WhereOptimizer.java
```diff
@@ -155,6 +155,7 @@ public static Expression pushKeyExpressionsToScan(StatementContext context, Set<
             keySlots = whereClause.accept(visitor);
 
             if (keySlots == null && (tenantId == null || !table.isMultiTenant()) && table.getViewIndexId() == null && !minOffset.isPresent()) {
+                // FIXME this overwrites salting info in the scanRange
```
virajjasani, line 158: which query will be impacted due to the override?

virajjasani (commented):
looks good overall, left minor comments

virajjasani:
surprising that end2end tests were not run https://ci-hadoop.apache.org/job/Phoenix/job/Phoenix-PreCommit-GitHub-PR/job/PR-1607/28/testReport/

phoenix-core/src/main/java/org/apache/phoenix/compile/WhereOptimizer.java
stoty: None.
If this were fixed, we could take use the salting info from the scanRange object.
Now, we have to go back to the PTable to get it, and the salting info in ScanRange is not used.

phoenix-core/src/main/java/org/apache/phoenix/iterate/ParallelScansCollector.java
```diff
@@ -0,0 +1,57 @@
...
+
+/**
+ * Stores some state to help build the parallel Scans structure
+ */
+public class ParallelScansCollector {
+
+    private ParallelScanGrouper grouper;
+    private boolean lastScanCrossedRegionBoundary = false;
+    private List<List<Scan>> parallelScans = new ArrayList<>();
+    private List<Scan> lastBatch = new ArrayList<>();
```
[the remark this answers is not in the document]
stoty: Will fix in a moment

stoty:
I'm pretty sure that the test were run. 
We're just aging them out really quick, because the Jenkins workers have very little disk space.
The final changes will trigger a new CI run.

phoenix-core/src/main/java/org/apache/phoenix/compile/WhereOptimizer.java
stoty: Currently none.
We just need to re-read some salting info from the table, instead of taking the cached value from

stoty:
I can see that there are also quite a few warnings.
I will address (some of) those based on the new CI run.

stoty:
I took a look at the warning, they are mostly for existing issues.
I do not plan to fix those, as that would only add more noise to the patch.

phoenix-core/src/main/java/org/apache/phoenix/compile/WhereOptimizer.java
virajjasani: got it, sounds good

virajjasani (approved):
+1

phoenix-core/src/main/java/org/apache/phoenix/iterate/BaseResultIterators.java
```diff
@@ -39,6 +39,7 @@
 import java.io.EOFException;
 import java.sql.SQLException;
 import java.util.*;
+import java.util.Arrays;
```
richardantal, line 42: Can you remove this import please or list everything needed under java.util?

stoty: You are right, but this has already been committed.
Can you open a follow-up ticket ?

PR #1610, opened 2023-05-23
PHOENIX-6943 Add MetadataCache on each region server.

shahrs87:
Changes done in this PR:
1. Added DDLTimestamp maintainers as scan and mutation attribute for get and put mutations requests.
2. Deserialize DDLTimestamp maintainers on the regionserver.
3. For every DDL Timestamp maintainer, make a request to SYSCAT regionserver to get LAST_DDL_TIMESTAMP.
4. Cache the LAST_DDL_TIMESTAMP on the regionserver so that next request can be served from the cache.
5. Whole feature is gated by config parameter: `phoenix.verify.last_ddl_timestamp`, default set to false.


shahrs87:
In bunch of tests I have replaced `PHOENIX_JDBC_URL` with `getUrl()` method.
`PHOENIX_JDBC_URL` is evaluated to `jdbc:phoenix:localhost;test=true` which doesn't have zookeeper port number in the URL. It defaults to 2181 which is the default port number. Before the patch in PhoenixTestDriver, we create CQSI object once, cache it in PhoenixTestDriver and all the other calls uses the same CQSI object. But after the patch, we maintain a cache of `ConnectionInfo -> ConnectionQueryServices` within PhoenixTestDriver. To construct the right ConnectionInfo we need the actual zookeeper port number. Thats why we can't use `PHOENIX_JDBC_URL` anymore.

shahrs87:
@jpisaac  @tkhurana  Can you review this please?

phoenix-core/src/main/java/org/apache/phoenix/hbase/index/IndexRegionObserver.java
```diff
@@ -1142,7 +1143,7 @@ public void preBatchMutateWithExceptions(ObserverContext<RegionCoprocessorEnviro
         identifyIndexMaintainerTypes(indexMetaData, context);
         identifyMutationTypes(miniBatchOp, context);
         context.populateOriginalMutations(miniBatchOp);
-
+        VerifyLastDDLTimestamp.verifyLastDDLTimestamp(miniBatchOp, c.getEnvironment());
```
tkhurana, line 1146: @shahrs87 So for indexes on non immutable tables we generate mutations on the server. I think those mutations will also need the last ddl timestamp attribute.

shahrs87: The ddl timestamps are already verified when we create the mutations for the base table. When we upsert into base table, we validate the last ddl timestamp for base table as well as all the indexes.

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/VerifyLastDDLTimestamp.java
```diff
@@ -0,0 +1,130 @@
...
+import static org.apache.phoenix.coprocessor.BaseScannerRegionObserver.LAST_DDL_TIMESTAMP_MAINTAINERS;
+import static org.apache.phoenix.coprocessor.BaseScannerRegionObserver.LAST_DDL_TIMESTAMP_MAINTAINERS_VERIFIED;
+import static org.apache.phoenix.schema.types.PDataType.TRUE_BYTES;
+
+/**
+ * Client provides last DDL timestamp of tables/views/indexes included in read/write operation
+ * through scan or mutation {@link BaseScannerRegionObserver#LAST_DDL_TIMESTAMP_MAINTAINERS}
+ * attribute.
+ * This verifies that client has the latest version of LAST_DDL_TIMESTAMP version.
+ * If client's provided LAST_DDL_TIMESTAMP is less than what is present in SYSTEM.CATALOG
```
jpisaac, line 43: Can it be greater? Due to clock skew and  ....

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/VerifyLastDDLTimestamp.java
```diff
@@ -0,0 +1,130 @@
...
+        }
+        DDLTimestampMaintainersProtos.DDLTimestampMaintainers maintainers =
+                LastDDLTimestampMaintainerUtil.deserialize(maintainersBytes);
+        verifyLastDDLTimestampInternal(maintainers, env);
+    }
+
+    /**
+     * Verify that LAST_DDL_TIMESTAMP provided by the client is upto date. If it is stale it will
+     * throw StaleMetadataCacheException.
+     * This method will be called by each coprocessor in the co-proc chain. The first co-proc in
```
jpisaac, line 68: Can the LAST_DDL_TIMESTAMP change between coproc checks? And if so do we need to handle?

jpisaac (commented):
Did an initial pass, Will take a deeper dive.

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/VerifyLastDDLTimestamp.java
shahrs87: I think its fine NOT to check between co processor invocation. In current implementation we don't resolve table between multiple co processor invocation. WDYT @jpisaac

phoenix-core/src/main/java/org/apache/phoenix/cache/ServerMetadataCache.java
```diff
@@ -0,0 +1,144 @@
...
+
+    /**
+     * Creates/gets an instance of ServerMetadataCache.
+     * @param env RegionCoprocessorEnvironment
+     * @return cache
+     */
+    public static ServerMetadataCache getInstance(RegionCoprocessorEnvironment env) {
+        ServerMetadataCache result = INSTANCE;
+        if (result == null) {
+            synchronized (ServerMetadataCache.class) {
```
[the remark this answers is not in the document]
shahrs87: Is this comment just informational? or are you expecting some changes in the code? @jpisaac

phoenix-core/src/main/java/org/apache/phoenix/cache/ServerMetadataCache.java
```diff
@@ -0,0 +1,144 @@
...
+            // TODO PhoenixRuntime#getTableNoCache can throw TableNotFoundException.
+            //  In that case, do we want to throw non retryable exception back to the client?
+            // Update cache with the latest DDL timestamp from SYSCAT server.
+            lastDDLTimestampMap.put(tableKeyPtr, table.getLastDDLTimestamp());
+        } catch (SQLException sqle) {
+            // Throw IOException back to the client and let the client retry depending on
+            // the configured retry policies.
+            LOGGER.warn("Exception while calling calling getTableNoCache for tenant id: {}," +
+                            " tableName: {}", tenantIDStr, fullTableNameStr, sqle);
+            throw new IOException(sqle);
```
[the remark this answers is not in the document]
shahrs87: We are calling this from [BaseScannerRegionObserver#preScannerOpen](https://github.com/apache/phoenix/blob/2dab08504c4828119bf55ad69653477ab8085bd1/phoenix-core/src/main/java/org/apache/phoenix/coprocessor/BaseScannerRegionObserver.java#L249) and this can only throw IOException.

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/VerifyLastDDLTimestamp.java
shahrs87: Both are fetched from SYSCAT table. One is cached on the client and other is cached on the region server. I don't think we will have to worry about clock skew issue here.

PR #1612, opened 2023-05-30
PHOENIX-6963 : Bootstrap last ddl timestamp for indexes

palashc:


phoenix-core/src/main/java/org/apache/phoenix/query/ConnectionQueryServicesImpl.java
```diff
@@ -4122,6 +4122,7 @@ protected PhoenixConnection upgradeSystemCatalogIfRequired(PhoenixConnection met
             metaConnection = addColumnsIfNotExists(metaConnection, PhoenixDatabaseMetaData.SYSTEM_CATALOG,
                 MIN_SYSTEM_TABLE_TIMESTAMP_5_2_0,
                 PhoenixDatabaseMetaData.STREAMING_TOPIC_NAME + " " + PVarchar.INSTANCE.getSqlTypeName());
+            UpgradeUtil.bootstrapLastDDLTimestampForIndexes(metaConnection);
```
shahrs87, line 4125: I don't know how this UpgradeUtil code works and when it is executed? If it is executed after the upgrade then the server version would be equal to MIN_SYSTEM_TABLE_TIMESTAMP_5_2_0. But adding DDL timestamps to table/view happened in 4.16.0 version and there also we have similar condition `currentServerSideTableTimeStamp < MIN_SYSTEM_TABLE_TIMESTAMP_4_16_0` so this code clearly works.

phoenix-core/src/it/java/org/apache/phoenix/end2end/UpgradeIT.java
```diff
@@ -697,21 +698,33 @@ public void testLastDDLTimestampBootstrap() throws Exception {
...
+            UpgradeUtil.bootstrapLastDDLTimestampForTablesAndViews(conn.unwrap(PhoenixConnection.class));
             long actualTableTS = getLastTimestampForMetadata(conn, schemaName, tableName,
                 PTableType.TABLE);
             long actualViewTS = getLastTimestampForMetadata(conn, schemaName, viewName,
                 PTableType.VIEW);
+            long actualIndexTS = getLastTimestampForMetadata(conn, schemaName, indexName,
+                    PTableType.INDEX);
             assertEquals(tableTS, actualTableTS);
             assertEquals(viewTS, actualViewTS);
+            assertEquals(0L, actualIndexTS);
```
shahrs87, line 722: should actualIndexTS be NULL or 0? I think it should be null.

phoenix-core/src/it/java/org/apache/phoenix/end2end/UpgradeIT.java
```diff
@@ -697,21 +698,33 @@ public void testLastDDLTimestampBootstrap() throws Exception {
...
+            conn.createStatement().execute("CREATE INDEX " + indexName + " ON " + fullTableName + " (KV1) ASYNC");
+
             //Now we null out any existing last ddl timestamps
             nullDDLTimestamps(conn);
 
             //now get the row timestamps for each header row
             long tableTS = getRowTimestampForMetadata(conn, schemaName, tableName,
                 PTableType.TABLE);
             long viewTS = getRowTimestampForMetadata(conn, schemaName, viewName, PTableType.VIEW);
+            long indexTS = getRowTimestampForMetadata(conn, schemaName, indexName, PTableType.INDEX);
```
shahrs87, line 710: Can we assert that indexTS is not 0L? Also we can capture the start time of the test and assert that indexTS > start time of the test.

phoenix-core/src/it/java/org/apache/phoenix/end2end/UpgradeIT.java
```diff
@@ -697,21 +698,33 @@ public void testLastDDLTimestampBootstrap() throws Exception {
...
+            long actualIndexTS = getLastTimestampForMetadata(conn, schemaName, indexName,
+                    PTableType.INDEX);
             assertEquals(tableTS, actualTableTS);
             assertEquals(viewTS, actualViewTS);
+            assertEquals(0L, actualIndexTS);
+
+            // bootstrap last ddl timestamp for indexes
+            UpgradeUtil.bootstrapLastDDLTimestampForIndexes(conn.unwrap(PhoenixConnection.class));
+            actualIndexTS = getLastTimestampForMetadata(conn, schemaName, indexName, PTableType.INDEX);
+            assertEquals(indexTS, actualIndexTS);
```
shahrs87, line 727: Maybe I am missing something. 
`indexTS` is the timestamp after we set LAST_DDL_TIMESTAMP to null via `nullDDLTimestamps` method. So it should be NULL.
`actualIndexTS` is the timestamp after we call `UpgradeUtil.bootstrapLastDDLTimestampForIndexes`. So it should the LAST_ROW_TIMESTAMP() for this index.
So how are they supposed to be equal?

shahrs87 (changes requested):
Overall the PR looks good. Few minor comments. Thank you @palashc !

phoenix-core/src/it/java/org/apache/phoenix/end2end/UpgradeIT.java
palashc: It is returned using a `rs.getLong()` call which returns 0 if the actual value was sql null.

phoenix-core/src/it/java/org/apache/phoenix/end2end/UpgradeIT.java
palashc: @shahrs87 `indexTS` is the PHOENIX_ROW_TIMESTAMP() of the header row of the index in syscat (it is the timestamp of the empty column) and indicates the last time the row was modified. After we call `UpgradeUtil.bootstrapLastDDLTimestampForIndexes`, this same value for the row is used to populate the `LAST_DDL_TIMESTAMP` column which is captured in `actualIndexTS`.

shahrs87: Oh my bad, I thought indexTS is calculated from `getLastTimestampForMetadata` method.

phoenix-core/src/it/java/org/apache/phoenix/end2end/UpgradeIT.java
```diff
@@ -683,10 +683,12 @@ private void verifyExpectedCellValue(byte[] rowKey, byte[] syscatBytes,
 
     @Test
     public void testLastDDLTimestampBootstrap() throws Exception {
+        Long testStartTime = System.currentTimeMillis();
```
shahrs87, line 686: replace this with `EnvironmentEdgeManager.currentTimeMillis`

shahrs87:
In the last jenkins run, there were around 16 checkstyle errors. If those are related to you patch please fix it otherwise +1 ltgm.

shahrs87:
I don't have commit rights, @tkhurana  can you please review this patch too? Thank you.

PR #1614, opened 2023-05-31
PHOENIX-6981 Update Jackson to 2.14.1

ksobolew:
If that's too much of a jump, then we could bump instead to 2.12.7.1 :)

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -92,7 +92,7 @@
     <top.dir>${project.basedir}/..</top.dir>
 
     <!-- Dependency versions -->
-    <jackson-bom.version>2.12.6.20220326</jackson-bom.version>
+    <jackson-bom.version>2.14.1</jackson-bom.version>
     <antlr.version>3.5.2</antlr.version>
     <!-- Only used for tests with HBase 2.1-2.4 -->
     <reload4j.version>1.2.19</reload4j.version>


pom.xml
```diff
@@ -92,7 +92,7 @@
     <top.dir>${project.basedir}/..</top.dir>
 
     <!-- Dependency versions -->
-    <jackson-bom.version>2.12.6.20220326</jackson-bom.version>
+    <jackson-bom.version>2.15.1</jackson-bom.version>
```
gjacoby126, line 95: @ksobolew - thanks for the contribution. We usually try to keep our dependencies in sync with HBase. Looks like they're currently on 2.14.1, so we should probably do the same, unless we find a particular incompatibility or CVE with that version. 

See https://github.com/apache/hbase/blob/branch-2.5/pom.xml#L578

gjacoby126 (changes requested):

pom.xml
ksobolew: Sure, 2.14.1 is good too. Regarding CVEs, both 2.15.1 and 2.14.1 are clear of them, contrary to 2.12.6.

ksobolew: Changed, thanks.

gjacoby126:
@ksobolew - change looks good. Could you please change the PR and the commit message to include the JIRA number?

ksobolew:
> @ksobolew - change looks good. Could you please change the PR and the commit message to include the JIRA number?

@gjacoby126 I don't have a JIRA number. How do I get one?

gjacoby126:
@ksobolew - our issue tracking system is https://issues.apache.org/jira/ . Unfortunately, public signup for it was recently turned off because of bot problems, but if you request an account at https://selfserve.apache.org/jira-account.html  I or one of the other PMC members can approve you and make you a contributor to the project. (This step's a one-time thing.) 

Then it's just a matter of filing a JIRA ticket and rebasing the PR so that the commit message includes the JIRA number.

ksobolew:
@gjacoby126 Done!

ksobolew:
(I requested a new account, but then realized that I already have one, so I used that one instead.)

gjacoby126 (approved):
+1, thanks @ksobolew

PR #1619, opened 2023-06-08
PHOENIX-6962 Write javadoc for different mutation plans within Delete…

jinggou:
…Compiler

### DIFF
--- a/phoenix-core/src/main/java/org/apache/phoenix/compile/DeleteCompiler.java
+++ b/phoenix-core/src/main/java/org/apache/phoenix/compile/DeleteCompiler.java
@@ -357,9 +357,8 @@ public class DeleteCompiler {
 
     /**
      * Implementation of MutationPlan that is selected if
-     * 1) there is no index presented for a table,
-     * 2) no LIMIT clause in the DELETE statement, and
-     * 3) the query is strictly point lookup.
+     * 1) the query either has no WHERE clause or it is strictly point lookup, and
+     * 2) no LIMIT clause in the DELETE statement.
      */
     public class MultiRowDeleteMutationPlan implements MutationPlan {
         private final List<MutationPlan> plans;
@@ -759,11 +758,9 @@ public class DeleteCompiler {
 
     /**
      * Implementation of MutationPlan that is selected if
-     * 1) there are mutable indexes presented for the table,
-     * 2) auto commit is on,
-     * 3) the table is not transactional,
-     * 4) there is no WHERE clause, and
-     * 5) no LIMIT clause in the DELETE statement.
+     * 1) auto commit is enabled as well as server side delete mutations are enabled,
+     * 2) the table is not transactional,
+     * 3) there is neither WHERE clause nor LIMIT clause in the DELETE statement.
      */
     public class ServerSelectDeleteMutationPlan implements MutationPlan {
         private final StatementContext context;
@@ -890,8 +887,10 @@ public class DeleteCompiler {
     }
 
     /**
-     * Implementation of MutationPlan that is selected if there are immutable indexes presented
-     * for the table, and does not meet the constraints of ServerSelectDeleteMutationPlan.
+     * Implementation of MutationPlan that is selected if
+     * 1) there are immutable indexes presented for the table,
+     * 2) the query has WHERE clause and is not strictly point lookup, and
+     * 3) does not meet the constraints of ServerSelectDeleteMutationPlan.
      */
     public class ClientSelectDeleteMutationPlan implements MutationPlan {
         private final StatementContext context;
--- a/phoenix-core/src/main/java/org/apache/phoenix/compile/DeleteCompiler.java
+++ b/phoenix-core/src/main/java/org/apache/phoenix/compile/DeleteCompiler.java
@@ -356,8 +356,10 @@ public class DeleteCompiler {
     }
 
     /**
-     * Implementation of MutationPlan that is selected if there is no index presented for a table
-     * and no LIMIT clause in the DELETE statement.
+     * Implementation of MutationPlan that is selected if
+     * 1) there is no index presented for a table,
+     * 2) no LIMIT clause in the DELETE statement, and
+     * 3) the query is strictly point lookup.
      */
     public class MultiRowDeleteMutationPlan implements MutationPlan {
         private final List<MutationPlan> plans;
--- a/phoenix-core/src/main/java/org/apache/phoenix/compile/DeleteCompiler.java
+++ b/phoenix-core/src/main/java/org/apache/phoenix/compile/DeleteCompiler.java
@@ -356,7 +356,7 @@ public class DeleteCompiler {
     }
 
     /**
-     * MultiRowDeleteMutationPlan is selected if there is no index presented for a table
+     * Implementation of MutationPlan that is selected if there is no index presented for a table
      * and no LIMIT clause in the DELETE statement.
      */
     public class MultiRowDeleteMutationPlan implements MutationPlan {
@@ -670,7 +670,7 @@ public class DeleteCompiler {
     }
 
     /**
-     * Implementation of MutationPlan that is used for composing a MultiRowDeleteMutationPlan.
+     * Implementation of MutationPlan for composing a MultiRowDeleteMutationPlan.
      */
     private class SingleRowDeleteMutationPlan implements MutationPlan {
 
@@ -756,8 +756,12 @@ public class DeleteCompiler {
     }
 
     /**
-     * ServerSelectDeleteMutationPlan is selected for a DELETE statement without WHERE clause or a LIMIT clause
-     * if there is index presented for the table.
+     * Implementation of MutationPlan that is selected if
+     * 1) there are mutable indexes presented for the table,
+     * 2) auto commit is on,
+     * 3) the table is not transactional,
+     * 4) there is no WHERE clause, and
+     * 5) no LIMIT clause in the DELETE statement.
      */
     public class ServerSelectDeleteMutationPlan implements MutationPlan {
         private final StatementContext context;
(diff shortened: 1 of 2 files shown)


virajjasani (commented):
Thanks @jinggou, could you also add the implications of setting auto-commit mode? i.e. if auto-commit is on, which plan is selected with WHERE clause

phoenix-core/src/main/java/org/apache/phoenix/compile/DeleteCompiler.java
```diff
@@ -354,7 +354,11 @@ private List<PTable> getClientSideMaintainedIndexes(TableRef tableRef) {
         }
         return Collections.emptyList();
     }
-    
+
+    /**
+     * Implementation of MutationPlan that is selected if there is no index presented for a table
+     * and no LIMIT clause in the DELETE statement.
+     */
     public class MultiRowDeleteMutationPlan implements MutationPlan {
```
virajjasani, line 362: let's also add that this plan gets selected if the query is strictly point lookup?

shahrs87:
@tkhurana  Can you please review this PR for correctness? I am also not that familiar with the decisions involved in selecting mutation plan.

phoenix-core/src/main/java/org/apache/phoenix/compile/DeleteCompiler.java
```diff
@@ -748,6 +757,14 @@ public Long getEstimateInfoTimestamp() throws SQLException {
         }
     }
 
+    /**
+     * Implementation of MutationPlan that is selected if
+     * 1) there are mutable indexes presented for the table,
+     * 2) auto commit is on,
+     * 3) the table is not transactional,
+     * 4) there is no WHERE clause, and
+     * 5) no LIMIT clause in the DELETE statement.
```
virajjasani, line 766: every plan except `MultiRowDeleteMutationPlan` can have LIMIT and WHERE correct? 

i might have misread earlier, otherwise would have commented, sorry.

phoenix-core/src/main/java/org/apache/phoenix/compile/DeleteCompiler.java
virajjasani: `the query is strictly point lookup` => `the query either has no WHERE clause or it is strictly point lookup`

virajjasani (commented):
almost there

phoenix-core/src/main/java/org/apache/phoenix/compile/DeleteCompiler.java
virajjasani: basically, both `ServerSelectDeleteMutationPlan` and `ClientSelectDeleteMutationPlan` can have LIMIT and/or WHERE clause (for non-pk columns)

tkhurana:
@jinggou There are other things to like serverside deletes need a config to be set. Only then server side deletes are enabled.

virajjasani:
@tkhurana you meant any additional config on top of auto-commit?

virajjasani:
oh i think it needs `phoenix.client.enable.server.delete.mutations` to be enabled, which is enabled by default but yeah we can mention this with auto-commit. we can specify: for server plan, auto-commit and server delete mutations should be enabled

phoenix-core/src/main/java/org/apache/phoenix/compile/DeleteCompiler.java
```diff
@@ -748,6 +757,14 @@ public Long getEstimateInfoTimestamp() throws SQLException {
         }
     }
 
+    /**
+     * Implementation of MutationPlan that is selected if
+     * 1) there are mutable indexes presented for the table,
+     * 2) auto commit is on,
```
virajjasani, line 763: `auto commit is on` => `auto commit is enabled as well as server side delete mutations are enabled`

thanks @tkhurana

tkhurana:
@jinggou I would suggest you to run tests in DeleteIT and explore the various explain plans to get a better understanding of the scenarios where each plan is used.

phoenix-core/src/main/java/org/apache/phoenix/compile/DeleteCompiler.java
```diff
@@ -354,7 +354,13 @@ private List<PTable> getClientSideMaintainedIndexes(TableRef tableRef) {
         }
         return Collections.emptyList();
     }
-    
+
+    /**
+     * Implementation of MutationPlan that is selected if
+     * 1) there is no index presented for a table,
```
tkhurana, line 360: For mutable indexes it is possible to use this plan.

tkhurana: Even immutable indexes can use this plan if the indexed columns can be evaluated from the query

phoenix-core/src/main/java/org/apache/phoenix/compile/DeleteCompiler.java
jinggou: If there is LIMIT clause, it will go client side

virajjasani: i see, `boolean runOnServer = isAutoCommit && !hasPreOrPostProcessing && .....`
so `!hasPreOrPostProcessing` is applicable to both `MultiRowDeleteMutationPlan` and `ServerSelectDeleteMutationPlan`.

what this means is, if `hasPreOrPostProcessing` is true, only `ClientSelectDeleteMutationPlan` will get selected.

`hasPreOrPostProcessing` is always true for LIMIT clause, alright this one looks good, thanks

phoenix-core/src/main/java/org/apache/phoenix/compile/DeleteCompiler.java
jinggou: Oh yeah, you are right. I will update that.

jinggou:
> @jinggou I would suggest you to run tests in DeleteIT and explore the various explain plans to get a better understanding of the scenarios where each plan is used.

jinggou:
> @jinggou I would suggest you to run tests in DeleteIT and explore the various explain plans to get a better understanding of the scenarios where each plan is used.

I've run all tests in DeleteIT and try more by modifying delete statement. I've updated javadoc based on above comments. Feel free to comment more if you find anything is missing/incorrect. Thanks!

phoenix-core/src/main/java/org/apache/phoenix/compile/DeleteCompiler.java
```diff
@@ -748,6 +756,12 @@ public Long getEstimateInfoTimestamp() throws SQLException {
         }
     }
 
+    /**
+     * Implementation of MutationPlan that is selected if
+     * 1) auto commit is enabled as well as server side delete mutations are enabled,
+     * 2) the table is not transactional,
+     * 3) there is neither WHERE clause nor LIMIT clause in the DELETE statement.
```
tkhurana, line 763: WHERE clause is allowed @jinggou

tkhurana: No immutable index (or client side index)

jinggou: I compare the tests for ClientSelectDeleteMutationPlan and ServerSelectDeleteMutationPlan and the difference is the WHERE clause. In what case does WHERE clause be allowed ?

tkhurana: @jinggou This test uses server delete plan and it has where clause. https://github.com/apache/phoenix/blob/master/phoenix-core/src/it/java/org/apache/phoenix/end2end/DeleteIT.java#L113

You can change the test by asserting on the explain plan. To get the explain plan you can use this method:
https://github.com/apache/phoenix/blob/master/phoenix-core/src/test/java/org/apache/phoenix/util/TestUtil.java#L1255

Then you can check the explain plan. For server it will have the string "DELETE ROWS SERVER SELECT"

tkhurana (changes requested):

phoenix-core/src/main/java/org/apache/phoenix/compile/DeleteCompiler.java
```diff
@@ -872,6 +887,12 @@ public QueryPlan getQueryPlan() {
         }
     }
 
+    /**
+     * Implementation of MutationPlan that is selected if
+     * 1) there are immutable indexes presented for the table,
+     * 2) the query has WHERE clause and is not strictly point lookup, and
```
tkhurana, line 893: Point 2 is applicable to ServerSelectPlan also.

phoenix-core/src/main/java/org/apache/phoenix/compile/DeleteCompiler.java
```diff
@@ -354,7 +354,12 @@ private List<PTable> getClientSideMaintainedIndexes(TableRef tableRef) {
         }
         return Collections.emptyList();
     }
-    
+
+    /**
+     * Implementation of MutationPlan that is selected if
+     * 1) the query either has no WHERE clause or it is strictly point lookup, and
```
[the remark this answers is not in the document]
virajjasani: > Just `point lookup` is correct in my opinion.

now i understand why this is true, even though there is no client indexes, for this to be true, there is at least one query plan available
`noQueryReqd &= queryPlans.size() == 1 + clientSideIndexes.size();`

and when noQueryReqd is true, it will at least go through this once:
```
            noQueryReqd = (!context.getScan().hasFilter()
                    || context.getScan().getFilter() instanceof SkipScanFilter)
                && context.getScanRanges().isPointLookup();
```

hence, i agree that point lookup is necessary for MultiRowDeleteMutation to get selected.

phoenix-core/src/main/java/org/apache/phoenix/compile/DeleteCompiler.java
tkhurana: So I would recommend we move bullet 2 to serverselect as bullet 5. And just one bullet point here.

tkhurana (approved):

tkhurana:
Thanks @jinggou

virajjasani:
Thank you @jinggou, +1

virajjasani:
@shahrs87 anything from you sir before closing this PR?

shahrs87 (approved):
+1 ltgm Thank you @jinggou

PR #1629, opened 2023-06-19
PHOENIX-6985 Setting server-side masking flag default to false

lokiore:


stoty:
Please make sure that the commit message has the JIRA id in the proper format.

jpisaac:
@lokiore The following JIRA title may be more appropriate or something along these -

"Setting server-side masking config default to false"

lokiore:
@jpisaac I see 2 test failures, one is passing locally and Failover one is failing even without the commit. Can you please take a look. I'll push an empty commit to rerun the tests.

jpisaac:
@lokiore I understand they are flappers after talking to @virajjasani @tkhurana, Is there a way to mark them as such?

jpisaac (approved):
LGTM +1

lokiore:
> @lokiore I understand they are flappers after talking to @virajjasani @tkhurana, Is there a way to mark them as such?

I am not sure if we can do something here other than Ignore which I assume we don't want. maybe some plugin can be added at jenkins like Result Test Analyser or do we already have that @virajjasani !?

gjacoby126 (approved):
+1

virajjasani (approved):

virajjasani:
@jpisaac we want this in 5.1 too?

lokiore:
@virajjasani create PR for 5.1 to let tests run :- https://github.com/apache/phoenix/pull/1634

PR #1630, opened 2023-06-20
Combined PR for testing Phoenix-6984 and phoenix-6983 together

stoty:


stoty:
Unfortunately, PHOENIX-6983 cannot be tested without PHOENIX-6984, so either we have to do reviews and commits in two passes, or we need to do it together.
There is a separate PR for PHOENIX-6984 only: #1625

PR #1631, opened 2023-06-20
Branch for testinf Phoenix-6984,  phoenix-6983 and phoenix 6959 together on 5.1

stoty:


stoty:
This contains backports of Phoenix-6984 and phoenix-6983, and adds PHOENIX-6959 on top of it.
Phoenix-6959 is specific to 5.1, as it backports some fixes from master.

We need to review these together, or do the reviews and commits on two or three passes.

phoenix-core/src/main/java/org/apache/phoenix/optimize/QueryOptimizer.java
```diff
@@ -369,6 +384,10 @@ private QueryPlan addPlan(PhoenixStatement statement, SelectStatement select, PT
                  * otherwise we just don't use this index (as opposed to trying to join back from
                  * the index table to the data table.
                  */
+                // Reset the state changes from the attempt above
+                indexTableRef.setHinted(false);
```
virajjasani, line 388: without this reset, the compiler might still end up using uncovered index right?

stoty: We'd use the skip-scan-join plan with the projector for the server merge plan, which doesn't work.

phoenix-core/src/it/java/org/apache/phoenix/end2end/index/GlobalIndexOptimizationIT.java
```diff
@@ -155,14 +155,13 @@ private void testOptimization(String dataTableName, String dataTableFullName, St
...
-            String expected = 
-                    "CLIENT PARALLEL 1-WAY FULL SCAN OVER " + dataTableName + "\n" +
-                    "    SKIP-SCAN-JOIN TABLE 0\n" +
-                    "        CLIENT PARALLEL 1-WAY RANGE SCAN OVER " + indexTableName + " \\['a'\\]\n" +
-                    "            SERVER FILTER BY FIRST KEY ONLY\n" +
-                    "    DYNAMIC SERVER FILTER BY \\(\"" + dataTableName + ".T_ID\", \"" + dataTableName + ".K1\", \"" + dataTableName + ".K2\"\\) IN \\(\\(\\$\\d+.\\$\\d+, \\$\\d+.\\$\\d+, \\$\\d+.\\$\\d+\\)\\)";
+            String expected =
+                    "CLIENT PARALLEL 1-WAY RANGE SCAN OVER " + indexTableName + " ['a']\n" +
+                            "    SERVER MERGE [0.K3]\n" +
+                            "    SERVER FILTER BY FIRST KEY ONLY";
```
virajjasani, line 161: do we not need `NO_INDEX_SERVER_MERGE` with `INDEX("dt" "it")` for this to be changed?

phoenix-core/src/main/java/org/apache/phoenix/compile/ProjectionCompiler.java
```diff
@@ -505,7 +505,15 @@ public static RowProjector compile(StatementContext context, SelectStatement sta
...
             for (byte[] family : projectedFamilies) {
-                projectColumnFamily(table, scan, family);
+                try {
+                    if (table.getColumnFamily(family) != null) {
+                        projectColumnFamily(table, scan, family);
+                    }
+                } catch (ColumnFamilyNotFoundException e) {
+                    if (!(tableRef.getTable().getIndexType() == IndexType.LOCAL || isHintedGlobalIndex(tableRef))) {
+                        throw e;
+                    }
```
virajjasani, line 515: for else case, would a DEBUG log be helpful?

phoenix-core/src/main/java/org/apache/phoenix/optimize/QueryOptimizer.java
stoty: For example we saw

Error: ERROR 504 (42703): Undefined column. columnName=BILLING_ORDER.0:PERIOD_END (state=42703,code=504)
org.apache.phoenix.schema.ColumnNotFoundException: ERROR 504 (42703): Undefined column. columnName=BILLING_ORDER.0:PERIOD_END

which would refer to the projected data table in the server merge plan, but the actual Table is unprojected data table in the skip-join-scan, which doesn't have a column of that name.

Resetting those flags makes sure that we generate and use the correct projector.

phoenix-core/src/it/java/org/apache/phoenix/end2end/index/GlobalIndexOptimizationIT.java
stoty: With PHOENIX-6959 the behaviour is the same as 5.2, and these changes are taken from 5.2 branch.

I have added a single new test case for the NO_INDEX_SERVER_MERGE option below.

phoenix-core/src/main/java/org/apache/phoenix/compile/ProjectionCompiler.java
stoty: This is again synced from 5.2, which does not have a debug statement there.
This applies to all server merged columns, so it would generate a log line for each merged column, which sounds excessive.

PR #1637, opened 2023-07-11
PHOENIX-6943 Add MetadataCache on each region server.

shahrs87:


phoenix-core/src/test/java/org/apache/phoenix/query/BaseTest.java
```diff
@@ -493,6 +494,8 @@ public static synchronized void tearDownMiniCluster(final int numTables) {
         } finally {
             try {
                 utility.shutdownMiniCluster();
+                // Clear ServerMetadataCache.
+                ServerMetadataCache.resetCache();
```
tkhurana, line 498: move this call before the shutdown call in case shutdown throws an exception

phoenix-core/src/test/java/org/apache/phoenix/cache/ServerMetadataCacheTest.java
```diff
@@ -0,0 +1,206 @@
...
+import static org.mockito.Mockito.times;
+import static org.mockito.Mockito.verify;
+
+public class ServerMetadataCacheTest extends BaseTest {
+    @BeforeClass
+    public static synchronized void doSetup() throws Exception {
+        Map<String, String> props = Maps.newHashMapWithExpectedSize(1);
+        props.put(BaseScannerRegionObserver.PHOENIX_MAX_LOOKBACK_AGE_CONF_KEY,
+                Integer.toString(60*60)); // An hour
+        props.put(QueryServices.USE_STATS_FOR_PARALLELIZATION, Boolean.toString(false));
```
tkhurana, line 54: Do we need these properties ?

tkhurana (approved):

shahrs87:
There is just 1 test failure. `org.apache.phoenix.jdbc.FailoverPhoenixConnectionIT#testConnectionWhenActiveZKRestarts` which is a known flapper.
There is 1 spotbugs warning where I am setting the connection to use in test for mock/spying purpose.
@tkhurana  Can you please merge this PR? Thank you!

PR #1640, opened 2023-07-14
PHOENIX-6986 Add property to disable server merges for hinted uncover…

kabhishek4:
…ed indexes

Introduces config parameter to enable or disable server merges of uncovered hinted indexes.

### DIFF
--- a/phoenix-core/src/main/java/org/apache/phoenix/optimize/QueryOptimizer.java
+++ b/phoenix-core/src/main/java/org/apache/phoenix/optimize/QueryOptimizer.java
@@ -336,10 +336,15 @@ public class QueryOptimizer {
         Map<TableRef, QueryPlan> dataPlans = Collections.singletonMap(indexTableRef, dataPlan);
         PTable indexTable = indexTableRef.getTable();
         PIndexState indexState = indexTable.getIndexState();
+        boolean isServerMergeFoeUncoveredIndexEnabled = statement.getConnection().
+            getQueryServices().getProps().getBoolean(
+                QueryServices.SERVER_MERGE_FOR_UNCOVERED_INDEX,
+                QueryServicesOptions.DEFAULT_SERVER_MERGE_FOR_UNCOVERED_INDEX);
         if (indexState == PIndexState.ACTIVE || indexState == PIndexState.PENDING_ACTIVE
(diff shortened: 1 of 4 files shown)


phoenix-core/src/main/java/org/apache/phoenix/optimize/QueryOptimizer.java
```diff
@@ -336,10 +336,15 @@ private QueryPlan addPlan(PhoenixStatement statement, SelectStatement select, PT
         Map<TableRef, QueryPlan> dataPlans = Collections.singletonMap(indexTableRef, dataPlan);
         PTable indexTable = indexTableRef.getTable();
         PIndexState indexState = indexTable.getIndexState();
+        boolean isServerMergeFoeUncoveredIndexEnabled = statement.getConnection().
```
stoty, line 339: typo

stoty: Also, CQSI is already available in the "services" object variable, no need to go back to statement.

stoty (changes requested):

phoenix-core/src/main/java/org/apache/phoenix/optimize/QueryOptimizer.java
kabhishek4: Corrected the typo. 

I am reading the value of parameter from statement because my test is failing. 
I see that value of parameter is true in QueryOptimizer.addPlan,

this.services.getProps().getBoolean(
QueryServices.SERVER_MERGE_FOR_UNCOVERED_INDEX,
QueryServicesOptions.DEFAULT_SERVER_MERGE_FOR_UNCOVERED_INDEX);

despite setting it to

props.setProperty(QueryServices.SERVER_MERGE_FOR_UNCOVERED_INDEX,
    Boolean.toString(false));

in the test. This is because there is nothing in the overriden values map.

I verified another property USE_REVERSE_SCAN_ATTRIB and it shows the same behaviour. 
I could see appropriate value only when read from the ‘statement’ object.


Failed Checks analysis:
The checkstyle errors are not due to my code.
I ran ParallelPhoenixConnectionFailureTest.testExecuteQueryChainFailure locally and it passed. 
It’s a flaky test.

stoty:
Hmm.
I think that this may be a limitation of the "unit" test infrastructure.
It would be interesting to see if the using the existing services object in the code works in a unit test, or on a real / sandbox cluster.

stoty (approved):
+1 LGTM

PR #1646, opened 2023-07-24
PHOENIX-6999 Point lookups fail with reverse scan

stoty:


ankitsinghal (approved):
scan.includeRow() is it new

stoty:
That API was added in 1.4

PR #1654, opened 2023-08-08
PHOENIX-6988 Create new regionserver coprocessor named PhoenixRegionServerEndpoint in Phoenix

shahrs87:
1. Create a new region server level coproc: PhoenixRegionServerCoprocessor
2. Added a sample method to validate last DDL timestamp for list of tables.
3. Added test cases.

### DIFF
--- a/phoenix-core/src/main/java/org/apache/phoenix/coprocessor/VerifyLastDDLTimestamp.java
+++ b/phoenix-core/src/main/java/org/apache/phoenix/coprocessor/VerifyLastDDLTimestamp.java
@@ -50,12 +50,12 @@ public class VerifyLastDDLTimestamp {
      * @param tableName              table name
      * @param clientLastDDLTimestamp last ddl timestamp provided by client
      * @param conf                   configuration
-     * @throws IOException
+     * @throws SQLException         StaleMetadataCacheException if client provided timestamp
+     *                              is stale.
      */
     public static void verifyLastDDLTimestamp(Configuration conf, byte[] tenantID,
             byte[] schemaName, byte[] tableName, long clientLastDDLTimestamp) throws SQLException {
         ServerMetadataCache cache = ServerMetadataCache.getInstance(conf);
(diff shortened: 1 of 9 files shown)


shahrs87:
@virajjasani  All the tests passed in the last 2 commits and the checkstyle comments are not relevant.
Can you please review?

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/VerifyLastDDLTimestamp.java
```diff
@@ -0,0 +1,70 @@
...
+     * @param schemaName             schema name
+     * @param tableName              table name
+     * @param clientLastDDLTimestamp last ddl timestamp provided by client
+     * @param conf                   configuration
+     * @throws IOException
+     */
+    public static void verifyLastDDLTimestamp(Configuration conf, byte[] tenantID,
+            byte[] schemaName, byte[] tableName, long clientLastDDLTimestamp) throws SQLException {
+        ServerMetadataCache cache = ServerMetadataCache.getInstance(conf);
+        String fullTableName = SchemaUtil.getTableName(schemaName, tableName);
```
virajjasani, line 58: good to pass `fullTableName` as argument to avoid calling `SchemaUtil#getTableName` twice: in `PhoenixRegionServerEndpoint#validateLastDDLTimestamp` as well as here?

virajjasani (approved):
left few nits, +1 overall

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/VerifyLastDDLTimestamp.java
shahrs87: But we still need byte[] tablename separately. However I need fullTableName only when the timestamp is stale which should be rare. So I have moved creating fullTableName only in exception block.

shahrs87:
Thank you @virajjasani  for the review.

PR #1656, opened 2023-08-16
PHOENIX-6942: Some config properties do not have phoenix prefix

Aarchy:


stoty (commented):
Can you add a test where you check the deprecation mechanism ?

Also check that we're not using these strings as Scan attributes (we'd need extra code to handle the both versions)

phoenix-core/src/it/java/org/apache/phoenix/end2end/index/IndexCoprocIT.java
```diff
@@ -58,12 +58,12 @@ public class IndexCoprocIT extends ParallelStatsDisabledIT {
         "|org.apache.phoenix.index.GlobalIndexChecker|805306365|";
     public static final String INDEX_REGION_OBSERVER_CONFIG =
         "|org.apache.phoenix.hbase.index.IndexRegionObserver|805306366|" +
-            "index.builder=org.apache.phoenix.index.PhoenixIndexBuilder," +
```
stoty, line 61: index.builder and org.apache.hadoop.hbase.index.codec.class are present in the HBase table descriptor.
For data backwards compatibility reasons I'd prefer not change these two (even though they have the same problems as others).

i.e. Changing these would break older versions of Phoenix when trying to use them if we changed these.

stoty (changes requested):

phoenix-core/src/it/java/org/apache/phoenix/end2end/index/IndexCoprocIT.java
stoty: And these are not read from hbase-site.xml, but from the Hbase table metadata directly.

Aarchy: Thank you for pointing this out, will restore the original strings. 
The string "index.builder" was declared twice though, I think one of them was unnecessary.

stoty: Can you revert the whitespace changes ?
If we don't change these, it's better to leave the file alone.

phoenix-core/src/main/java/org/apache/phoenix/hbase/index/IndexRegionObserver.java
```diff
@@ -333,7 +330,6 @@ public int getMaxPendingRowCount() {
 
   private static final int DEFAULT_ROWLOCK_WAIT_DURATION = 30000;
   private static final int DEFAULT_CONCURRENT_MUTATION_WAIT_DURATION_IN_MS = 100;
-
```
stoty, line 336: nit: don't delete the empty line

phoenix-core/src/main/java/org/apache/phoenix/hbase/index/Indexer.java
```diff
@@ -194,7 +194,6 @@ public BatchMutateContext(int clientVersion) {
             .encodeVersion("0.94.9");
 
   private static final int DEFAULT_ROWLOCK_WAIT_DURATION = 30000;
-
```
stoty, line 197: nit: don't delete the empty line

stoty (changes requested):
Some nits and whitespace issues

stoty (approved):
+1 LGTM

PR #1658, opened 2023-08-22
PHOENIX-6523 Support for HBase Registry Implementations through Phoenix connection URL

stoty:


phoenix-core/src/main/java/org/apache/phoenix/jdbc/MasterConnectionInfo.java
```diff
@@ -0,0 +1,103 @@
...
+import org.apache.phoenix.util.ReadOnlyProps;
+
+/**
+ * ConnectionInfo used for org.apache.hadoop.hbase.client.MasterRegistry
+ *
+ */
+public class MasterConnectionInfo extends AbstractRPCConnectionInfo {
+
+    private static final String MASTER_REGISTRY_CLASS_NAME =
+            "org.apache.hadoop.hbase.client.MasterRegistry";
```
virajjasani, line 37: How about using MasterRegistry class directly to retrieve it's full name?
It has also been deprecated with HBASE-26172

stoty: Referring the class directly would require compiling with HBase 2.4+, and would preclude a straight-ish backport to 5.1.

Also, RPCRegistry is only available in HBase 2.5, so we cannot do this for each registry even in master.
I have duplicated all property and class strings so that we can compile with HBase 2.1+, and added some explicit HBase version checks instead.

(I also noticed that we are duplicating a lot of string constants that have been stable and available for a long time, but I left those alone).

The extra code required to support MasterRegistry is also minimal, it's the same as RPCClientRegistry, it just takes the hosts lists from different properties.

stoty: Opened PHOENIX-7059 for the constant de-duplication.

virajjasani:
we need rebase

virajjasani (commented):
It looks good overall with ConnectionInfo refactoring, will take another look

stoty:
Thank you for taking a look, @virajjasani .
I have rebased the patch.

phoenix-core/src/main/java/org/apache/phoenix/jdbc/ConnectionInfo.java
```diff
@@ -0,0 +1,546 @@
...
+                // No registry class set in config. Use version-dependent default
+                if (VersionInfo.getMajorVersion(VersionInfo.getVersion()) >= 3) {
+                    builder = new RPCConnectionInfo.Builder(url, configuration, props, info);
+                } else {
+                    builder = new ZKConnectionInfo.Builder(url, configuration, props, info);
+                }
+            }
+        } else {
+            throw getMalFormedUrlException(url);
+        }
```
virajjasani, line 172: I wonder how best we can document this and what would be the right place

virajjasani (commented):
overall good, if we could get review for HA changes, would be more helpful

phoenix-core/src/main/java/org/apache/phoenix/jdbc/ConnectionInfo.java
stoty: I have added very long release note to the ticket as a stopgap, but we're gonna have to update the web page for sure.

stoty:
> overall good, if we could get review for HA changes, would be more helpful

Thank you.
I have pinged @dbwong , but he hasn't reacted yet. Can you reach him directly ?

phoenix-core/src/main/java/org/apache/phoenix/jdbc/ConnectionInfo.java
stoty: I'm gonna do that after the patch has landed.
Probably a dedicated page would be the best, then we could link to it from every place where the URLs are referred.
I would also like to emphasize in the docs the best solution is just to rely the defaults in hbase-site.xml

stoty:
And BTW this is something I'd also like to backport to 5.1.
While this is a new feature, itis meant to be fully backwards compatible, and if it is not, we need to fix that anyway.

virajjasani (approved):

virajjasani:
Thanks @stoty! 
I believe we might also need a heads up thread on dev@ to let everyone know of this big change getting committed to master and 5.1 branches?

> While this is a new feature, itis meant to be fully backwards compatible, and if it is not, we need to fix that anyway.

Makes sense, good candidate for 5.1.4 release

stoty:
Sure, I will announce it, and copy the release notes to the email.

stoty:
This has been merged some tine ago.

PR #1659, opened 2023-08-22
PHOENIX-7024 Fix issues in Server Paging

tkhurana:


phoenix-core/src/main/java/org/apache/phoenix/iterate/NonAggregateRegionScannerFactory.java
```diff
@@ -303,12 +282,34 @@ public boolean isFilterDone() {
...
+                            // with any additional rows then this special cell is treated as part
+                            // of application result set.
+                            KeyValue kv = new KeyValue(QueryConstants.OFFSET_ROW_KEY_BYTES,
+                                    QueryConstants.OFFSET_FAMILY, QueryConstants.OFFSET_COLUMN,
+                                    PInteger.INSTANCE.toBytes(iterator.getRemainingOffset()));
+                            results.add(kv);
+                        }
+                        return false;
+                    } else {
+                        // tuple is non null
```
kadirozde, line 304: There is no check for null tuple. Is this comment still valid?

kadirozde: Oh I remember isFilterDone checks for null tuple. Ignore my previous comment

kadirozde: We do not handle the case where the region closes before returning the row at the starting offset. To support this, we need to get the region lock using region.startRegionOperation() while scanning the rows before the starting offset.  If we get a dummy row while the region is locked (i.e., while scanning the rows before the starting offset), we need to close the region and return a special result (not dummy) which includes the current row key and the remaining offset. The client should reopen a scanner with starting with the next row and pass the remaining offset.

kadirozde (approved):
+1. Thank you for extensive testing and weeding out the bugs!

PR #1666, opened 2023-08-30
PHOENIX-7025 : Create a new RPC to validate last ddl timestamp for read requests.

palashc:
On the read path, if `phoenix.ddl.timestamp.validation.enabled` is set:

- Pick a random live region server and call the `validateLastDDLTimestamp` RPC for the table/view/index (and all its ancestors) in the optimized query plan. 
    - List of live region servers is cached at the CQSI level and initialized in `CQSI.init`. 
- If `StaleMetadataCache` exception is encountered, force update client side cache and retry `executeQuery`.
    - Skip validating last ddl timestamp again in this retry. 
    - In case of a view, also update the cache for all its ancestors.
    - In case of an index, just update the cache for its parent table. That will pull in the latest index metadata as well. 
- If an error is encountered in performing the `validateLastDDLTimestamp` RPC, retry once independent of the executeQuery retry which is already implemented currently. 
    - Refresh the list of live region servers in case we tried the rpc on a bad region server.


phoenix-core/src/main/java/org/apache/phoenix/query/QueryServices.java
```diff
@@ -313,6 +313,9 @@ public interface QueryServices extends SQLCloseable {
     //Update Cache Frequency default config attribute
     public static final String DEFAULT_UPDATE_CACHE_FREQUENCY_ATRRIB  = "phoenix.default.update.cache.frequency";
 
+    // whether to validate last ddl timestamps during client operations
+    public static final String LAST_DDL_TIMESTAMP_VALIDATION_ENABLED = "phoenix.ddl.timestamp.validation.enabled";
```
virajjasani, line 317: can we make it true at BaseTest level so that it applies to all tests? that way we will get more confidence?

virajjasani (commented):
though `UPDATE_CACHE_FREQUENCY` matters, still worth doing it at BaseTest level?

phoenix-core/src/main/java/org/apache/phoenix/jdbc/PhoenixStatement.java
```diff
@@ -417,11 +504,22 @@ private PhoenixResultSet executeQuery(final CompilableStatement stmt,
...
                                         //TODO we can log retry count and error for debugging in LOG table
-                                        return executeQuery(stmt, false, queryLogger, noCommit);
+                                        return executeQuery(stmt, false, queryLogger, noCommit, validateDDLTimestamp);
                                     }
                                 }
                                 throw e;
-                            } catch (RuntimeException e) {
+                            }
+                            catch (SQLException e) {
+                                // force update cache in case of StaleMetadataCacheException and retry
```
shahrs87, line 513: Trying to understand why we have retries here and also in validateDDLTimestamp? Can we consolidate them just in 1 place? Preferably in validateDDLTimestamp?

shahrs87 (changes requested):
Went through the non test code, posted few comments. Thank you @palashc  for the patch.

phoenix-core/src/main/java/org/apache/phoenix/jdbc/PhoenixStatement.java
```diff
@@ -317,16 +331,83 @@ protected QueryPlan optimizeQuery(CompilableStatement stmt) throws SQLException
     
     protected PhoenixResultSet executeQuery(final CompilableStatement stmt, final QueryLogger queryLogger)
             throws SQLException {
-        return executeQuery(stmt, true, queryLogger, false);
+        return executeQuery(stmt, true, queryLogger, false, this.ddlTimestampValidationEnabled);
```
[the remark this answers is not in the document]
palashc: Good point, can just check that class level variable.

phoenix-core/src/main/java/org/apache/phoenix/jdbc/PhoenixStatement.java
```diff
@@ -317,16 +331,83 @@ protected QueryPlan optimizeQuery(CompilableStatement stmt) throws SQLException
...
+     */
+    private void validateDDLTimestamp(TableRef tableRef, boolean doRetry) throws SQLException {
+
+        try (Admin admin = this.connection.getQueryServices().getAdmin()) {
+            // get all live region servers
+            List<ServerName> regionServers = new ArrayList<>(admin.getRegionServers(true));
+            // pick one at random
+            ServerName regionServer = regionServers.get(ThreadLocalRandom.current().nextInt(regionServers.size()));
+
+            LOGGER.debug("Sending DDL timestamp validation request for table {} at regionserver {}",
```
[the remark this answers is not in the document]
palashc: `tableRef.getTable().getName()` returns the full name if I am not wrong. I can break it down to `tableRef.getTable().getTableName()` and `tableRef.getTable().getSchemaName()` to avoid any confusion.

phoenix-core/src/main/java/org/apache/phoenix/query/QueryServices.java
palashc: @virajjasani that makes sense, I can try running all tests with this property set.

phoenix-core/src/main/java/org/apache/phoenix/jdbc/PhoenixStatement.java
```diff
@@ -317,16 +331,83 @@ protected QueryPlan optimizeQuery(CompilableStatement stmt) throws SQLException
...
+                    tableRef.getTable().getTableName().getString(), regionServer.toString());
+
+            // RPC
+            CoprocessorRpcChannel channel = admin.coprocessorService(regionServer);
+            PhoenixRegionServerEndpoint.BlockingInterface service = PhoenixRegionServerEndpoint.newBlockingStub(channel);
+            service.validateLastDDLTimestamp(null, getValidateDDLTimestampRequest(tableRef));
+        }
+        // handle server side exceptions
+        catch (ServiceException | SQLException | IOException e) {
+            if (e instanceof ServiceException) {
```
shahrs87, line 389: You can refactor most of the catch block like this:
```suggestion
        catch (Exception e) {
          SQLException parsedException = ServerUtil.parseServerException(e);
          if (parsedException instanceof StaleMetadataCacheException) {
            throw parsedException;
          }
          //retry once for any other errors
          LOGGER.error("Error in validating DDL timestamp for table {}: {}",
                    tableRef.getTable().getName().getString(), parsedException);
          if (doRetry) {
              validateDDLTimestamp(tableRef, false);
              return;
          }
          throw parsedException;
```

shahrs87 (changes requested):
more comments.

phoenix-core/src/it/java/org/apache/phoenix/end2end/ServerMetadataCachingIT.java
```diff
@@ -0,0 +1,264 @@
...
+
+import static org.apache.hadoop.hbase.coprocessor.CoprocessorHost.REGIONSERVER_COPROCESSOR_CONF_KEY;
+import static org.apache.phoenix.util.TestUtil.TEST_PROPERTIES;
+import static org.mockito.Matchers.any;
+import static org.mockito.Matchers.anyLong;
+import static org.mockito.Matchers.eq;
+import static org.mockito.Matchers.isNull;
+
+@Category({NeedsOwnMiniClusterTest.class })
+public class ServerMetadataCachingIT extends BaseTest {
```
shahrs87, line 38: Do you want to add all the tests in ServerMetadataCacheTest?

phoenix-core/src/main/java/org/apache/phoenix/jdbc/PhoenixStatement.java
palashc: @shahrs87 any server side exception is wrapped in a `com.google.protobuf.ServiceException`, that is what the RPC returns. I have tried to differentiate between that case and the case when there is an IOException or SQLException raised by `CQSI.getAdmin` or `admin.getRegionServers`. We should parse server exception only in the former case, right?

phoenix-core/src/main/java/org/apache/phoenix/jdbc/PhoenixStatement.java
palashc: As discussed, validateDDLTimestamp retry is just for the RPC call. This retry is for executeQuery() after force-updating the client cache.

phoenix-core/src/main/java/org/apache/phoenix/jdbc/PhoenixStatement.java
palashc: Done!

phoenix-core/src/it/java/org/apache/phoenix/end2end/ServerMetadataCachingIT.java
```diff
@@ -0,0 +1,281 @@
...
+import java.util.Random;
+
+import static org.apache.hadoop.hbase.coprocessor.CoprocessorHost.REGIONSERVER_COPROCESSOR_CONF_KEY;
+import static org.apache.phoenix.util.TestUtil.TEST_PROPERTIES;
+import static org.mockito.Matchers.any;
+import static org.mockito.Matchers.anyLong;
+import static org.mockito.Matchers.eq;
+import static org.mockito.Matchers.isNull;
+
+@Category({NeedsOwnMiniClusterTest.class })
```
shahrs87, line 54: All the tests are great but would like to see one more test.
1. Make ConnectionQueryServices#getAdmin() always throw IOException. In this case executeQuery will throw Exception and query will fail.

shahrs87 (changes requested):

phoenix-core/src/main/java/org/apache/phoenix/jdbc/PhoenixStatement.java
```diff
@@ -421,7 +514,20 @@ private PhoenixResultSet executeQuery(final CompilableStatement stmt,
...
-                            } catch (RuntimeException e) {
+                            } catch (SQLException e) {
+                                // force update cache if StaleMetadataCacheException and retry
+                                if (e instanceof StaleMetadataCacheException) {
+                                    String planSchemaName = getLastQueryPlan().getTableRef().getTable().getSchemaName().toString();
+                                    String planTableName = getLastQueryPlan().getTableRef().getTable().getTableName().toString();
+                                    // update cache
+                                    new MetaDataClient(connection).updateCache(connection.getTenantId(), planSchemaName, planTableName, true);
+                                    // skip last ddl timestamp validation in the retry
+                                    setValidateLastDdlTimestamp(false);
```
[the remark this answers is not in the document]
palashc: Yeah I was thinking of that, do we ever re-use the statement for different queries? I could only find usage of `conn.createStatement` in our tests but we can not guarantee that usage from customers. 

I can revert back to passing a boolean parameter through the execute* methods.

phoenix-core/src/it/java/org/apache/phoenix/end2end/ServerMetadataCachingIT.java
```diff
@@ -0,0 +1,281 @@
...
+public class ServerMetadataCachingIT extends BaseTest {
+
+    private final Random RANDOM = new Random(42);
```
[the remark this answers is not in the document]
palashc: Needed some seed value!

phoenix-core/src/main/java/org/apache/phoenix/jdbc/PhoenixStatement.java
```diff
@@ -325,6 +337,80 @@ protected PhoenixResultSet executeQuery(final CompilableStatement stmt, final Qu
...
+            PhoenixRegionServerEndpoint.BlockingInterface service
+                    = PhoenixRegionServerEndpoint.newBlockingStub(channel);
+            service.validateLastDDLTimestamp(null, getValidateDDLTimestampRequest(tableRef));
+        } catch (Exception e) {
+            SQLException parsedException = ServerUtil.parseServerException(e);
+            if (parsedException instanceof StaleMetadataCacheException) {
+                throw parsedException;
+            }
+            //retry once for any exceptions other than StaleMetadataCacheException
+            LOGGER.error("Error in validating DDL timestamp for {}: {}",
```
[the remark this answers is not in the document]
palashc: Done.

phoenix-core/src/main/java/org/apache/phoenix/jdbc/PhoenixStatement.java
```diff
@@ -421,7 +514,20 @@ private PhoenixResultSet executeQuery(final CompilableStatement stmt,
                                     }
                                 }
                                 throw e;
-                            } catch (RuntimeException e) {
+                            } catch (SQLException e) {
```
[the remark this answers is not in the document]
palashc: Yes.

phoenix-core/src/it/java/org/apache/phoenix/end2end/ServerMetadataCachingIT.java
palashc: Added!

phoenix-core/src/it/java/org/apache/phoenix/end2end/ServerMetadataCachingIT.java
```diff
@@ -0,0 +1,281 @@
...
+
+            // create table and upsert using client-1
+            createTable(conn1, tableName, Long.MAX_VALUE);
+            upsert(conn1, tableName);
+
+            // Instrument ServerMetadataCache to throw a SQLException once
+            ServerMetadataCache spyCache = Mockito.spy(ServerMetadataCache.getInstance(config));
+            Mockito.doThrow(new SQLException()).doCallRealMethod().when(spyCache)
+                    .getLastDDLTimestampForTable(any(), any(), eq(Bytes.toBytes(tableName)));
+            ServerMetadataCache.setInstance(spyCache);
```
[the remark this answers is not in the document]
palashc: I am replacing the cache instance with a spy of the current cache instance which thrown an exception.

phoenix-core/src/it/java/org/apache/phoenix/end2end/ServerMetadataCachingIT.java
```diff
@@ -0,0 +1,281 @@
...
+            alterTableAddColumn(conn1, tableName, "newCol1");
+
+            // invalidate region server cache
+            ServerMetadataCache.resetCache();
+
+            // select query from client-2 with old ddl timestamp works
+            // there should be one more getTable RPC
+            query(conn2, tableName);
+            expectedNumGetTableRPCs += 1;
+            Mockito.verify(spyCqs2, Mockito.times(expectedNumGetTableRPCs))
```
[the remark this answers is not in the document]
palashc: The getTable RPC should imply that the cache was updated? I can look into how to verify that the exception was encountered and we indeed did a retry in PhoenixStatement.

phoenix-core/src/main/java/org/apache/phoenix/jdbc/PhoenixStatement.java
```diff
@@ -315,18 +327,91 @@ protected QueryPlan optimizeQuery(CompilableStatement stmt) throws SQLException
...
+     * Retry once if there was an error performing the RPC, otherwise throw the Exception.
+     * @param tableRef
+     * @throws SQLException
+     */
+    private void validateLastDDLTimestamp(TableRef tableRef, boolean doRetry) throws SQLException {
+
+        String infoString = getInfoString(tableRef);
+        try (Admin admin = this.connection.getQueryServices().getAdmin()) {
+            // get all live region servers
+            List<ServerName> regionServers = new ArrayList<>(admin.getRegionServers(true));
```
tkhurana, line 384: Is this a rpc to active hmaster ? For every DDL in every client we are going to make this rpc. This could be a bottleneck.

phoenix-core/src/main/java/org/apache/phoenix/jdbc/PhoenixStatement.java
```diff
@@ -315,18 +327,91 @@ protected QueryPlan optimizeQuery(CompilableStatement stmt) throws SQLException
...
 
-    protected PhoenixResultSet executeQuery(final CompilableStatement stmt, final QueryLogger queryLogger, boolean noCommit)
+    protected PhoenixResultSet executeQuery(final CompilableStatement stmt, final QueryLogger queryLogger, boolean noCommit, boolean validateLastDdlTimestamp)
             throws SQLException {
-        return executeQuery(stmt, true, queryLogger, noCommit);
+        return executeQuery(stmt, true, queryLogger, noCommit, validateLastDdlTimestamp);
+    }
+
+    private String getInfoString(TableRef tableRef) {
+        return String.format("Tenant: %s, Schema: %s, Table: %s",
```
tkhurana, line 341: I think this should be <schema> <table> <tenant>

phoenix-core/src/main/java/org/apache/phoenix/jdbc/PhoenixStatement.java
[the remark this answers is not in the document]
tkhurana: You can use the same prepared statement object in a loop by binding different values to the parameters in the query and then calling execute query on it.

phoenix-core/src/main/java/org/apache/phoenix/jdbc/PhoenixStatement.java
palashc: @tkhurana @shahrs87 does it make sense to cache live region servers at the CQSI level? Although I am not sure how that will be kept updated.

phoenix-core/src/main/java/org/apache/phoenix/jdbc/PhoenixStatement.java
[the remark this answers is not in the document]
palashc: Just keeping a class variable to store the config property, it can be used every time for executeQuery or executeMutation even if the same Statement object is used.

phoenix-core/src/it/java/org/apache/phoenix/end2end/ServerMetadataCachingIT.java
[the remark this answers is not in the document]
palashc: @shahrs87 I changed the tests to verify number of `cqsi.addTable` invocations instead.

phoenix-core/src/main/java/org/apache/phoenix/jdbc/PhoenixStatement.java
```diff
@@ -417,11 +508,20 @@ private PhoenixResultSet executeQuery(final CompilableStatement stmt,
                                                     e.getSchemaName(), e.getTableName(), true)
                                             .wasUpdated()) {
                                         //TODO we can log retry count and error for debugging in LOG table
-                                        return executeQuery(stmt, false, queryLogger, noCommit);
+                                        return executeQuery(stmt, false, queryLogger, noCommit, validateLastDdlTimestamp);
                                     }
                                 }
                                 throw e;
-                            } catch (RuntimeException e) {
+                            } catch (StaleMetadataCacheException e) {
```
tkhurana, line 515: I think you need to move this catch block inside where you are calling validateTimestamp otherwise the finally block will be executed twice when StaleMetadataCacheException is raised.

phoenix-core/src/main/java/org/apache/phoenix/jdbc/PhoenixStatement.java
tkhurana: connection objects are short lived so you will not gain much. Caching at the cqsi level should help.

palashc:
@shahrs87 Refreshing live region servers in CQSI.init creates issues for the Permission related ITs. I have added a feature flag check in CQSI.init for now. We can look into those failing ITs as part of another sub-task to turn the feature on.

palashc:
Test Report: https://ci-hadoop.apache.org/job/Phoenix/job/Phoenix-PreCommit-GitHub-PR/job/PR-1666/17/testReport/

phoenix-core/src/main/java/org/apache/phoenix/jdbc/PhoenixStatement.java
```diff
@@ -287,10 +298,14 @@ public String toString() {
     private int queryTimeoutMillis;
     // Caching per Statement
     protected final Calendar localCalendar = Calendar.getInstance();
+    private boolean validateLastDdlTimestamp;
 
     public PhoenixStatement(PhoenixConnection connection) {
         this.connection = connection;
         this.queryTimeoutMillis = getDefaultQueryTimeoutMillis();
+        this.validateLastDdlTimestamp = this.connection.getQueryServices().getProps()
```
shahrs87, line 306: Do you want to follow `getDefaultQueryTimeoutMillis` way to initialize `validateLastDdlTimestamp`?  I think it keeps the constructor clean.

phoenix-core/src/main/java/org/apache/phoenix/jdbc/PhoenixStatement.java
```diff
@@ -317,16 +332,112 @@ protected QueryPlan optimizeQuery(CompilableStatement stmt) throws SQLException
...
+        return String.format("Tenant: %s, Schema: %s, Table: %s",
+                this.connection.getTenantId(),
+                tableRef.getTable().getSchemaName(),
+                tableRef.getTable().getTableName());
+    }
+
+    private void setLastDDLTimestampRequestParameters(
+            RegionServerEndpointProtos.LastDDLTimestampRequest.Builder innerBuilder,
+            byte[] tenantId, PTable pTable) {
+        byte[] schemaBytes = pTable.getSchemaName() == null
```
shahrs87, line 353: for schemaBytes, we are doing null check here but for tenantID bytes, we are doing null check in `getValidateDDLTimestampRequest`. This is little bit confusing. Lets do all null check in one place. I think it should be in this function.

phoenix-core/src/main/java/org/apache/phoenix/jdbc/PhoenixStatement.java
```diff
@@ -317,16 +332,112 @@ protected QueryPlan optimizeQuery(CompilableStatement stmt) throws SQLException
...
+
+    private String getInfoString(TableRef tableRef) {
+        return String.format("Tenant: %s, Schema: %s, Table: %s",
+                this.connection.getTenantId(),
+                tableRef.getTable().getSchemaName(),
+                tableRef.getTable().getTableName());
+    }
+
+    private void setLastDDLTimestampRequestParameters(
+            RegionServerEndpointProtos.LastDDLTimestampRequest.Builder innerBuilder,
```
shahrs87, line 351: Change `innerBuilder` to `builder`?  There are no inner and outer builder in this method.

phoenix-core/src/main/java/org/apache/phoenix/jdbc/PhoenixStatement.java
```diff
@@ -317,16 +332,112 @@ protected QueryPlan optimizeQuery(CompilableStatement stmt) throws SQLException
...
+     * @param tableRef
+     * @throws SQLException
+     */
+    private void validateLastDDLTimestamp(TableRef tableRef, boolean doRetry) throws SQLException {
+
+        String infoString = getInfoString(tableRef);
+        try (Admin admin = this.connection.getQueryServices().getAdmin()) {
+            // if this is a retry, update the list of live region servers in CQSI
+            if (!doRetry) {
+                this.connection.getQueryServices().refreshLiveRegionServers();
```
shahrs87, line 407: This seems little unusual to me. Normally whenever you encounter an exception, you need to clear the cache. So we should move refreshing live regionserver list call to the catch exception block.

phoenix-core/src/main/java/org/apache/phoenix/jdbc/PhoenixStatement.java
```diff
@@ -317,16 +332,112 @@ protected QueryPlan optimizeQuery(CompilableStatement stmt) throws SQLException
...
+            }
+            // get all live region servers
+            List<ServerName> regionServers = this.connection.getQueryServices().getLiveRegionServers();
+            // pick one at random
+            ServerName regionServer
+                    = regionServers.get(ThreadLocalRandom.current().nextInt(regionServers.size()));
+
+            LOGGER.debug("Sending DDL timestamp validation request for {} to regionserver {}",
+                    infoString, regionServer.toString());
```
shahrs87, line 416:
```suggestion
                    infoString, regionServer);
```

phoenix-core/src/main/java/org/apache/phoenix/jdbc/PhoenixStatement.java
```diff
@@ -371,6 +482,23 @@ private PhoenixResultSet executeQuery(final CompilableStatement stmt,
                                 plan =
                                         connection.getQueryServices().getOptimizer()
                                                 .optimize(PhoenixStatement.this, plan);
+                                setLastQueryPlan(plan);
+
+                                //verify metadata for the table/view/index in the query plan
+                                if (plan.getTableRef() != null && validateLastDdlTimestamp) {
```
shahrs87, line 488: Is it even possible to have `plan.getTableRef()` as null ?

phoenix-core/src/main/java/org/apache/phoenix/jdbc/PhoenixStatement.java
```diff
@@ -371,6 +482,23 @@ private PhoenixResultSet executeQuery(final CompilableStatement stmt,
...
+                                    try {
+                                        validateLastDDLTimestamp(plan.getTableRef(), true);
+                                    } catch (StaleMetadataCacheException e) {
+                                        String planSchemaName = plan.getTableRef().getTable().getSchemaName().toString();
+                                        String planTableName = plan.getTableRef().getTable().getTableName().toString();
+                                        // force update client metadata cache
+                                        LOGGER.debug("Force updating client metadata cache for {}", getInfoString(getLastQueryPlan().getTableRef()));
```
shahrs87, line 495: If we receive StaleMetadataCacheException for any view in the view hierarchy, then we need to invlidate cache for the whole hierarchy, correct? since we don't know from the exception for which view the validation failed?

phoenix-core/src/main/java/org/apache/phoenix/query/ConnectionQueryServicesImpl.java
```diff
@@ -5181,6 +5192,24 @@ private void flushTable(byte[] tableName) throws SQLException {
...
+            catch (IOException e) {
+                throw ServerUtil.parseServerException(e);
+            }
+        }
+        LOGGER.info("Refreshed list of live region servers.");
+    }
+
+    @Override
+    public List<ServerName> getLiveRegionServers() {
+        return this.liveRegionServers;
```
shahrs87, line 5210: Do we want to have synchronize block here also? Assume a case where refreshing the live regionservers is ongoing and some client tries to access liveRegionServers. That client will get stale copy of region servers.

phoenix-core/src/main/java/org/apache/phoenix/schema/MetaDataClient.java
```diff
@@ -903,7 +903,7 @@ private boolean addColumnsAndIndexesFromAncestors(MetaDataMutationResult result,
                 String parentSchemaName = SchemaUtil.getSchemaNameFromFullName(parentName);
                 tableName = SchemaUtil.getTableNameFromFullName(parentName);
                 MetaDataMutationResult parentResult = updateCache(connection.getTenantId(), parentSchemaName, tableName,
-                        false, resolvedTimestamp);
+                        true, resolvedTimestamp);
```
shahrs87, line 906: Why is this change needed?

phoenix-core/src/it/java/org/apache/phoenix/end2end/ServerMetadataCachingIT.java
```diff
@@ -0,0 +1,334 @@
...
+
+            // query using client-2 should succeed
+            query(conn2, tableName);
+
+            // verify live region servers were refreshed
+            Mockito.verify(spyCqs2, Mockito.times(1)).refreshLiveRegionServers();
+        }
+        finally {
+            // reset cache instance so that it does not interfere with any other test
+            ServerMetadataCache.setInstance(cache);
```
shahrs87, line 185:
```suggestion
            ServerMetadataCache.resetCache();
```

phoenix-core/src/it/java/org/apache/phoenix/end2end/ServerMetadataCachingIT.java
```diff
@@ -0,0 +1,334 @@
...
+            // query using client-2 to populate cache
+            query(conn2, tableName);
+            expectedNumCacheUpdates = 1;
+            Mockito.verify(spyCqs2, Mockito.times(expectedNumCacheUpdates))
+                    .addTable(any(PTable.class), anyLong());
+
+            // add column using client-1 to update last ddl timestamp
+            alterTableAddColumn(conn1, tableName, "newCol1");
+
+            // invalidate region server cache
```
shahrs87, line 219: Ideally you don't need to call `resetCache` here since we should invalidate the cache if we run any DDL statements. Can you add TODO statement to remove this call once [PHOENIX-6968](https://issues.apache.org/jira/browse/PHOENIX-6968) is committed.

phoenix-core/src/it/java/org/apache/phoenix/end2end/ServerMetadataCachingIT.java
```diff
@@ -0,0 +1,334 @@
...
+            expectedNumCacheUpdates += 1;
+            Mockito.verify(spyCqs2, Mockito.times(expectedNumCacheUpdates))
+                    .addTable(any(PTable.class), anyLong());
+
+            // verify live region servers were refreshed
+            Mockito.verify(spyCqs2, Mockito.times(1)).refreshLiveRegionServers();
+        }
+        finally {
+            // reset cache instance so that it does not interfere with any other test
+            ServerMetadataCache.setInstance(cache);
```
shahrs87, line 240: Instead of resetting cache in every test, you can create method with After annotation and reset it there. See [here](https://github.com/apache/phoenix/blob/PHOENIX-6883-feature/phoenix-core/src/test/java/org/apache/phoenix/cache/ServerMetadataCacheTest.java#L55-L58)

shahrs87 (changes requested):
Left some comments. Thank you for the nice work @palashc !

phoenix-core/src/it/java/org/apache/phoenix/end2end/ServerMetadataCachingIT.java
```diff
@@ -0,0 +1,334 @@
...
+
+            // alter first level view using client-1 to update its last ddl timestamp
+            alterViewAddColumn(conn1, view1, "foo");
+
+            // invalidate region server cache
+            ServerMetadataCache.resetCache();
+
+            // query second level view
+            query(conn2, view2);
+            expectedNumCacheUpdates += 3; // table, view1
```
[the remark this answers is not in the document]
shahrs87: From the code, I don't see where it will update the cache for the whole hierarchy.

phoenix-core/src/main/java/org/apache/phoenix/jdbc/PhoenixStatement.java
```diff
@@ -371,6 +482,23 @@ private PhoenixResultSet executeQuery(final CompilableStatement stmt,
...
+                                        new MetaDataClient(connection).updateCache(connection.getTenantId(), planSchemaName, planTableName, true);
+                                        // skip last ddl timestamp validation in the retry
+                                        return executeQuery(stmt, doRetryOnMetaNotFoundError, queryLogger, noCommit, false);
```
shahrs87, line 498: Isn't this still part of the outer try block and hence we will update the metrics twice? Cc @tkhurana

phoenix-core/src/it/java/org/apache/phoenix/end2end/ServerMetadataCachingIT.java
palashc: Since these tests are more end to end, would it make sense to keep them here in the ITs package? We can also add tests for write path to the same class.

phoenix-core/src/main/java/org/apache/phoenix/jdbc/PhoenixStatement.java
palashc: In case of `EXPLAIN <query>`, I had noticed that `plan.getTableRef()` was null.

phoenix-core/src/main/java/org/apache/phoenix/query/ConnectionQueryServicesImpl.java
palashc: @shahrs87 I think a client getting a stale copy of region server list should be okay. It will pick another region server randomly which will likely be different from the regionserver which had issues. Also the list variable is `volatile` so that all threads will get the latest updated value. This weaker and light-weight form of synchronization should suffice for our purpose, most of the shared state in CQSI is handled this way. Let me know if I am missing something.

phoenix-core/src/it/java/org/apache/phoenix/end2end/ServerMetadataCachingIT.java
shahrs87: I would say keep all the tests in the same class. If you want to move tests from ServerMetadataCacheTest to this class, I am fine too. Both the tests are starting mini cluster, adding PhoenixRegionServerEndpoint co proc and testing some behavior. Hope this helps.

phoenix-core/src/main/java/org/apache/phoenix/jdbc/PhoenixStatement.java
shahrs87: Lets add a comment so that it is clear to everyone.

phoenix-core/src/main/java/org/apache/phoenix/schema/MetaDataClient.java
palashc: @shahrs87  `MetaDataClient.updateCache` eventually calls this method `addColumnsAndIndexesFromAncestors` which looks up ancestors of views and view indexes and recursively calls `MetaDataClient.updateCache` on every parent table up the hierarchy. Currently, the value of `alwaysHitServer` is false for all these recursive calls i.e. we will look up the ancestors in the local cache if UPDATE_CACHE_FREQUENCY is set to NEVER. Setting it to true bypasses all checks for avoiding the rpc to get table ([code](https://github.com/apache/phoenix/blob/master/phoenix-core/src/main/java/org/apache/phoenix/schema/MetaDataClient.java#L749)) and updates the entire hierarchy.

phoenix-core/src/it/java/org/apache/phoenix/end2end/ServerMetadataCachingIT.java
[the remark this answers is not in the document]
palashc: The change in MetadataClient takes care of that as discussed here: https://github.com/apache/phoenix/pull/1666#discussion_r1342971828

phoenix-core/src/main/java/org/apache/phoenix/jdbc/PhoenixStatement.java
palashc: The change in MetadataClient should take care of that as discussed here: https://github.com/apache/phoenix/pull/1666#discussion_r1342971828

phoenix-core/src/main/java/org/apache/phoenix/schema/MetaDataClient.java
palashc: I think to avoid any side effects, it will be better to pass the `alwaysHitServer` value to `addColumnsAndIndexesFromAncestors` instead of just setting it to true always. I will make that change.

phoenix-core/src/it/java/org/apache/phoenix/end2end/ServerMetadataCachingIT.java
palashc: Moved.

phoenix-core/src/main/java/org/apache/phoenix/jdbc/PhoenixStatement.java
palashc: @shahrs87 added a boolean flag for updating metrics in the finally block. Setting it to false in the catch blocks.

phoenix-core/src/main/java/org/apache/phoenix/jdbc/PhoenixStatement.java
```diff
@@ -417,54 +538,66 @@ private PhoenixResultSet executeQuery(final CompilableStatement stmt,
...
-                            } catch (RuntimeException e) {
-
+                            }
+                            catch (StaleMetadataCacheException e) {
+                                updateMetrics = false;
+                                String planSchemaName = lastQueryPlan.getTableRef().getTable().getSchemaName().toString();
+                                String planTableName = lastQueryPlan.getTableRef().getTable().getTableName().toString();
+                                // force update client metadata cache
+                                LOGGER.debug("Force updating client metadata cache for {}", getInfoString(getLastQueryPlan().getTableRef()));
+                                new MetaDataClient(connection).updateCache(connection.getTenantId(), planSchemaName, planTableName, true);
```
shahrs87, line 552: Add a comment saying that this will update the cache for all the parents in the hieracrchy.

phoenix-core/src/main/java/org/apache/phoenix/query/ConnectionQueryServicesImpl.java
```diff
@@ -3471,6 +3476,12 @@ public Void call() throws Exception {
                             LOGGER.info("An instance of ConnectionQueryServices was created.");
                             openConnection();
                             hConnectionEstablished = true;
+                            boolean lastDDLTimestampValidationEnabled
```
shahrs87, line 3479: This is fine as of now. But any reason why populating the region servers is guarded by LAST_DDL_TIMESTAMP_VALIDATION_ENABLED?

palashc: It was breaking some Permission related ITs. I could not look into it in detail at the time. https://github.com/apache/phoenix/pull/1666#issuecomment-1730322149

phoenix-core/src/test/java/org/apache/phoenix/cache/ServerMetadataCacheTest.java
```diff
@@ -419,16 +423,291 @@ public void testInvalidateCacheForBaseTableWithUpdateIndexStatement() throws Exc
...
+
+            // alter first level view using client-1 to update its last ddl timestamp
+            alterViewAddColumn(conn1, view1, "foo");
+
+            // invalidate region server cache
+            // TODO: remove this call after PHOENIX-6968 is committed.
+            ServerMetadataCache.resetCache();
+
+            // reset the spy CQSI object
+            Mockito.reset(spyCqs2);
```
shahrs87, line 651: Normally it is advised never to use mockito.reset ever in tests. 
See [here](https://javadoc.io/doc/org.mockito/mockito-core/latest/org/mockito/Mockito.html#resetting_mocks) 
But I think we can use it here.

palashc: I see, did not know that. My initial approach without reset was a better way then, I can revert to that.

shahrs87: But then we won't be able to test for which tables the updateCache was called. I think using Mockito.reset is fine since it is improving our test coverage.

phoenix-core/src/test/java/org/apache/phoenix/cache/ServerMetadataCacheTest.java
```diff
@@ -419,16 +425,296 @@ public void testInvalidateCacheForBaseTableWithUpdateIndexStatement() throws Exc
...
+            // add column using client-1 to update last ddl timestamp
+            alterTableAddColumn(conn1, tableName, "newCol1");
+
+            // reset the spy CQSI object
+            Mockito.reset(spyCqs2);
+
+            // Instrument ServerMetadataCache to throw a SQLException once
+            cache = ServerMetadataCache.getInstance(config);
+            ServerMetadataCache spyCache = Mockito.spy(cache);
+            Mockito.doThrow(new SQLException("FAIL")).doCallRealMethod().when(spyCache)
```
shahrs87, line 550: Here we are testing 2 things.
1. ServerMetadataCache throwing SQL Exception first time and succeeding 2nd time.
2. In the second attempt, ServerMetadataCache is throwing StaleMetadaCacheException since it is using old PTable object and the retry within PhoenixStatement#executeQuery will kick in and refresh the client metadata cache. 

Is that correct? If yes, then please update the javadoc

phoenix-core/src/test/java/org/apache/phoenix/cache/ServerMetadataCacheTest.java
```diff
@@ -419,16 +425,296 @@ public void testInvalidateCacheForBaseTableWithUpdateIndexStatement() throws Exc
...
+
+            // create table and upsert using client-1
+            createTable(conn1, tableName, NEVER);
+            upsert(conn1, tableName);
+
+            // Instrument ServerMetadataCache to throw a SQLException twice
+            cache = ServerMetadataCache.getInstance(config);
+            ServerMetadataCache spyCache = Mockito.spy(cache);
+            SQLException e = new SQLException("FAIL");
+            Mockito.doThrow(e).doThrow(e).when(spyCache)
```
shahrs87, line 589: Do we need `doThrow` twice?  If you just declare it once, it will throw the exception always.

shahrs87 (approved):
Will wait for jenkins run to succeed before merging.

palashc:
@shahrs87 Lets wait for https://github.com/apache/phoenix/pull/1702 before merging this PR.

phoenix-core/src/test/java/org/apache/phoenix/cache/ServerMetadataCacheTest.java
```diff
@@ -678,6 +677,139 @@ public void testSelectQueryOnSystemTables() throws Exception {
...
+            createIndex(conn1, tableName, indexName, "v1");
+
+            //client-2 populates its cache
+            query(conn2, tableName);
+
+            //client-1 drops the index
+            dropIndex(conn1, tableName, indexName);
+
+            //simulate changing table's ddl timestamp by altering it for now
+            //TODO: remove this when PHOENIX-7056 is committed.
```
shahrs87: @palashc  PHOENIX-7056 is committed now. Can you rebase this PR?

palashc: Yes I have done that already.

phoenix-core/src/test/java/org/apache/phoenix/cache/ServerMetadataCacheTest.java
```diff
@@ -678,6 +677,139 @@ public void testSelectQueryOnSystemTables() throws Exception {
...
+             Connection conn2 = spyCqs2.connect(url2, props)) {
+
+            //client-1 creates a table and an index on it
+            createTable(conn1, tableName, NEVER);
+            createIndex(conn1, tableName, indexName, "v1");
+            TestUtil.waitForIndexState(conn1, indexName, PIndexState.ACTIVE);
+
+            //client-2 populates its cache, 1 getTable&addTable call for the table
+            //no getTable calls for Index since we add all indexes in the cache from a table's PTable object
+            queryWithNoIndexHint(conn2, tableName);
```
shahrs87, line 811: Why do we need to add NO_INDEX hint?

phoenix-core/src/test/java/org/apache/phoenix/cache/ServerMetadataCacheTest.java
```diff
@@ -678,6 +677,139 @@ public void testSelectQueryOnSystemTables() throws Exception {
...
+
+            //client-1 updates index property
+            alterIndexChangeStateToRebuild(conn1, tableName, indexName);
+
+            //client-2's query using the index should work
+            queryWithIndexHint(conn2, tableName, indexName);
```
shahrs87, line 817: why we need to add INDEX hint?

palashc: I wanted to make sure the query plan uses the index table.

phoenix-core/src/test/java/org/apache/phoenix/cache/ServerMetadataCacheTest.java
palashc: The goal was to populate the client cache with both the table and index. If the query is on the index table, I think the data table will not be populated in the client cache.

phoenix-core/src/test/java/org/apache/phoenix/cache/ServerMetadataCacheTest.java
```diff
@@ -536,7 +525,391 @@ public void testUpdateLastDDLTimestampViewAfterIndexCreation() throws Exception
...
+            TestUtil.waitForIndexState(conn1, indexName, PIndexState.ACTIVE);
+
+            //client-2 populates its cache, 1 getTable&addTable call for the table
+            query(conn2, tableName);
+
+            //client-1 updates index property
+            alterIndexChangeStateToRebuild(conn1, tableName, indexName);
+
+            //client-2's query using the index should work
+            PhoenixStatement stmt = conn2.createStatement().unwrap(PhoenixStatement.class);
```
shahrs87, line 814: Do you want to replace these 2 lines with queryWithIndex() method?

phoenix-core/src/test/java/org/apache/phoenix/cache/ServerMetadataCacheTest.java
```diff
@@ -536,7 +525,391 @@ public void testUpdateLastDDLTimestampViewAfterIndexCreation() throws Exception
...
+
+        try (Connection conn1 = spyCqs1.connect(url1, props);
+             Connection conn2 = spyCqs2.connect(url2, props)) {
+
+            //client-1 creates table and index on it
+            createTable(conn1, tableName, NEVER);
+            createIndex(conn1, tableName, indexName, "v1");
+
+            //client-2 populates its cache
+            query(conn2, tableName);
```
shahrs87, line 897: Instead of `query` method (which runs COUNT(*)) query, do you want to use `queryWithIndex` method ?

shahrs87 (changes requested):
Overall the patch looks good. Few minor comments and waiting for jenkins build to run.

phoenix-core/src/test/java/org/apache/phoenix/cache/ServerMetadataCacheTest.java
palashc: I think it should not matter? Both queries will load the table and index into the client's cache.

phoenix-core/src/test/java/org/apache/phoenix/cache/ServerMetadataCacheTest.java
palashc: I wanted to make sure the statement's query plan used the index, I have an assert on that following this. Everywhere else I use queryWithIndex() method.

palashc:
Test Report: https://ci-hadoop.apache.org/job/Phoenix/job/Phoenix-PreCommit-GitHub-PR/job/PR-1666/31/testReport/

shahrs87:
Overall the patch looks good. 
There are 7 test failures in RegexBulkLoadToolIT.
Along with test failures, there are asflicense, blanks, checkstyle and spotbugs warnings.
@palashc  Please take a look.

palashc:
@shahrs87 RegexBulkLoadToolIT was probably a flapper. I have fixed most of the checkstyle and other issues.

shahrs87:
Still there are 31 checkstyle warnings and 2 spotbugs warnings. Can you see if they are relevant?

phoenix-core/src/main/java/org/apache/phoenix/query/ConnectionQueryServicesImpl.java
shahrs87: I don't think think we can keep populating regionservers guarded by LAST_DDL_TIMESTAMP_VALIDATION_ENABLED conf for a long time. In future if some other feature wants to read this regionservers list but don't want to enable LAST_DDL_TIMESTAMP_VALIDATION_ENABLED conf property.
For time being it is fine, but lets open a jira to see why Permission related IT are failing with this change. Before we merge this feature into master branch, we will have to fix that jira.

shahrs87 (approved):
+1 pending some checkstyle and spotbugs warnings.

shahrs87:
Merged. Thank you @palashc  for working on this.

PR #1670, opened 2023-09-06
PHOENIX-6979 (Addendum) Use HBase TTL as TTL for Tables only at Phoenix Level when phoenix.table.ttl.enabled is true

lokiore:


lokiore:
Not Sure why this is not showing build but all tests passed :- https://ci-hadoop.apache.org/job/Phoenix/job/Phoenix-PreCommit-GitHub-PR/job/PR-1670/2/artifact/yetus-general-check/output/patch-unit-phoenix-core.txt
@jpisaac

jpisaac:
Looks good.
@lokiore Need to handle the case for TTL=FOREVER.
Previously NONE was supported, but we should support it going forward. Since HBase TTL spec does not allow it.

lokiore:
> Looks good. @lokiore Need to handle the case for TTL=FOREVER. Previously NONE was supported, but we should support it going forward. Since HBase TTL spec does not allow it.

@jpisaac I have updated the PR to handle FOREVER and NONE value in both cases, when phoenix.table.level.ttl is enabled or disabled. Will create a JIRA to remove NONE value from TTL as it will require other changes as well and out of scope of this JIRA. Thanks!!!

jpisaac:
LGTM +1 Thanks @lokiore

jpisaac (approved):

virajjasani (approved):

PR #1672, opened 2023-09-13
PHOENIX-6973 Add option to CREATE TABLE to skip verification of HBase table

payert:
This Pull Request introduces a NOVERIFY option for CREATE TABLE that allows skipping the verification the existence of columns with empty qualifier.

### DIFF
--- a/phoenix-core/src/test/java/org/apache/phoenix/compile/PostIndexDDLCompilerTest.java
+++ b/phoenix-core/src/test/java/org/apache/phoenix/compile/PostIndexDDLCompilerTest.java
@@ -23,6 +23,7 @@ import java.sql.Connection;
 import java.sql.DriverManager;
 import java.sql.SQLException;
 
+import org.apache.phoenix.execute.MutationState;
 import org.apache.phoenix.jdbc.PhoenixConnection;
 import org.apache.phoenix.jdbc.PhoenixStatement;
 import org.apache.phoenix.query.BaseConnectionlessQueryTest;
@@ -57,6 +58,20 @@ public class PostIndexDDLCompilerTest extends BaseConnectionlessQueryTest {
             PostIndexDDLCompiler compiler = new PostIndexDDLCompiler(pConn, new TableRef(pDataTable));
             MutationPlan plan = compiler.compile(pConn.getTable(new PTableKey(null, "IDX")));
             assertEquals("T", plan.getQueryPlan().getTableRef().getTable().getTableName().getString());
+        }
+    }
+
+    @Test
+    public void testCreateTableWithNoVerify() throws SQLException {
+        String ddl = "CREATE TABLE A (K VARCHAR PRIMARY KEY DESC) NOVERIFY";
+        try (Connection conn = DriverManager.getConnection(getUrl())) {
+            PhoenixStatement stmt = conn.createStatement().unwrap(PhoenixStatement.class);
+            MutationPlan plan = stmt.compileMutation(ddl);
+            MutationState state = plan.execute();
+
+            assertEquals("CREATE TABLE\n", plan.getExplainPlan().toString());
+            assertEquals(0, state.getMaxSize());
+            assertEquals(0, state.getMaxSizeBytes());
         }
     }
 
--- a/phoenix-core/src/it/java/org/apache/phoenix/end2end/CreateTableIT.java
+++ b/phoenix-core/src/it/java/org/apache/phoenix/end2end/CreateTableIT.java
@@ -41,19 +41,28 @@ import java.util.Properties;
 import java.util.UUID;
 
 import org.apache.hadoop.hbase.HConstants;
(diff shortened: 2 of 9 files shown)


payert:
@stoty @richardantal Could you please review this PR ?

phoenix-core/src/test/java/org/apache/phoenix/jdbc/CreateTableCompilerTest.java
```diff
@@ -0,0 +1,52 @@
...
+import java.sql.SQLException;
+
+public class CreateTableCompilerTest extends BaseConnectionlessQueryTest {
+
+  @Test
+  public void testCreateTableWithNoVerify() throws SQLException {
+    String ddl = "CREATE TABLE A (K VARCHAR PRIMARY KEY DESC) NOVERIFY";
+    try (Connection conn = DriverManager.getConnection(getUrl());
+         PhoenixStatement phoenixStatement = conn.createStatement().unwrap(PhoenixStatement.class)) {
+      /* Test is placed to 'org.apache.phoenix.jdbc' package to be able to access 'CompilableStatement'
```
stoty, line 40: Can we move this to org.apache.phoenix.compile.PostIndexDDLCompilerTest ?
Based on that test, we don't actually need to acsess CompilableStatement.

phoenix-core/src/it/java/org/apache/phoenix/end2end/CreateTableNoVerifyIT.java
```diff
@@ -0,0 +1,134 @@
...
+import java.sql.Connection;
+import java.sql.DriverManager;
+import java.sql.SQLException;
+import java.util.ArrayList;
+import java.util.List;
+
+/**
+ * Test the NOVERIFY option of CREATE TABLE.
+ */
+@Category(ParallelStatsDisabledTest.class)
```
stoty, line 54: Move this to CreateTableIt ?

stoty (changes requested):

phoenix-core/src/test/java/org/apache/phoenix/jdbc/CreateTableCompilerTest.java
payert: I need to get the `CompilableStatement` instance to cast it to `CreateTableStatement` and check if the `noverify` is set. 
If we want to check for the `noverify` value in the `Statement` object we need to keep this test here, or if we want to move it we can drop the `assertTrue(((CreateTableStatement) compilableStatement).isNoVerify());` assertion.

phoenix-core/src/it/java/org/apache/phoenix/end2end/CreateTableNoVerifyIT.java
payert: done

phoenix-core/src/test/java/org/apache/phoenix/compile/PostIndexDDLCompilerTest.java
```diff
@@ -17,18 +17,25 @@
...
 
+import org.apache.phoenix.execute.MutationState;
 import org.apache.phoenix.jdbc.PhoenixConnection;
 import org.apache.phoenix.jdbc.PhoenixStatement;
+import org.apache.phoenix.parse.CreateTableStatement;
 import org.apache.phoenix.query.BaseConnectionlessQueryTest;
 import org.apache.phoenix.schema.PTable;
 import org.apache.phoenix.schema.PTableKey;
 import org.apache.phoenix.schema.TableRef;
+import org.apache.phoenix.util.PropertiesUtil;
```
stoty, line 38: You left the additional imports from the attempt in.

phoenix-core/src/test/java/org/apache/phoenix/jdbc/CreateTableCompilerTest.java
stoty: I'm not a fan of adding a new test class just for this feature.

If the point here is testing that the parser adds a flag, it could be done like in
org.apache.phoenix.parse.QueryParserTest.testParseCreateTableInlinePrimaryKeyWithOrder()
, and add the new test there.

stoty (changes requested):

phoenix-core/src/test/java/org/apache/phoenix/jdbc/CreateTableCompilerTest.java
payert: Yes, I agree about not introducing new test class. I have separated the tests to check the statement and the mutation in different tests and different test classes, so I could delete this test class without compromising on what we test.

stoty (approved):
+1 LGTM

stoty:
Please rebase and squash @payert .

stoty:
Sorry, I have already committed the previous version. 
No need to anything more.

PR #1673, opened 2023-09-14
PHOENIX-6900 Enable paging feature when the phoenix query is made on …

sdave33:
…a table snapshot.

This patch enables the paging feature when the phoenix query is made on the snapshot


sdave33 (commented):
@kadirozde we have this fixed already in the internal phoenix fork, this patch is to also push it into opensource. 
Kindly review it whenever you get a chance. thanks.

virajjasani:
the build had test failures (not exactly relevant), triggered another run

virajjasani (approved):

PR #1674, opened 2023-09-14
PHOENIX-7002 : Added log line for Stale Region Boundary Cache Exception

Divneet18:
Added a log line for understanding Stale Region Boundary error better

### DIFF
--- a/phoenix-core/src/main/java/org/apache/phoenix/iterate/BaseResultIterators.java
+++ b/phoenix-core/src/main/java/org/apache/phoenix/iterate/BaseResultIterators.java
@@ -1415,8 +1415,8 @@ public abstract class BaseResultIterators extends ExplainTable implements Result
                             // Resubmit just this portion of work again
                             Scan oldScan = scanPair.getFirst();
                             byte[] startKey = oldScan.getAttribute(SCAN_ACTUAL_START_ROW);
-                            if (e2 instanceof StaleRegionBoundaryCacheException){
-                                LOGGER.warn("Region boundary is stale and throwing exception with the following stack trace ", e2);
+                            if (e2 instanceof StaleRegionBoundaryCacheException) {
+                                LOGGER.warn("Found stale region boundary with the following stack trace ", e2);
                             }
                             if (e2 instanceof HashJoinCacheNotFoundException) {
                                 LOGGER.debug(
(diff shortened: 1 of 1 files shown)


virajjasani (approved):

lokiore (approved):

virajjasani:
Thank you for the contribution @Divneet18!!

phoenix-core/src/main/java/org/apache/phoenix/iterate/BaseResultIterators.java
```diff
@@ -1415,6 +1415,9 @@ private List<PeekingResultIterator> getIterators(List<List<Scan>> scan, Connecti
                             // Resubmit just this portion of work again
                             Scan oldScan = scanPair.getFirst();
                             byte[] startKey = oldScan.getAttribute(SCAN_ACTUAL_START_ROW);
+                            if (e2 instanceof StaleRegionBoundaryCacheException) {
```
shahrs87, line 1418: Sorry for the late comment. While I created the jira, I wanted to log all the exceptions coming from the server side not just StaleRegionBoundaryCacheException. I was thinking to logging the exception at L#1408, something like this.
```
SQLException sqle = ServerUtil.parseServerException(e);
LOG.warn("Exception encountered", sqle);
throw sqle;
```
Cc @Divneet18 @virajjasani @lokiore

virajjasani: Not bad idea but i wonder whether this would result in redundant whole stacktrace logging. Perhaps, we can just log `e.getMessage()` and let this change to report whole stacktrace for `StaleRegionBoundaryCacheException` be as is?

shahrs87: I don't think just printing `e.getMessage()`is a good idea. We should remove the current logging and add at L#1408. It would be duplicate for HashJoinCacheNotFoundException but thats ok.

virajjasani: Removing from here is fine but adding full stacktraces might be overwhelming amount of logs is what i think. For every single SELECT query with some/any exception, we would log and that might be too much for logs.

It's fine though, let's give it a shot but I fear that during some genuine problem at the server side, we might be overwhelmed with long stacktraces that are on the other hand obvious errors and likely be ignored by anyone debugging.

PR #1675, opened 2023-09-15
PHOENIX-7035 Index on data table with only pk columns result in invalid state

virajjasani:


phoenix-core/src/main/java/org/apache/phoenix/index/IndexMaintainer.java
```diff
@@ -634,6 +638,23 @@ private Void addDataColInfo(final PTable dataTable, Expression expression) {
...
                     }
+
+                    private boolean tableHasOnlyPkCols(PTable dataTable, byte[] cf) {
+                        try {
+                            dataTable.getColumnFamily(cf);
+                        } catch (ColumnFamilyNotFoundException e) {
+                            List<PColumn> columns = dataTable.getColumns();
+                            if (columns != null && columns.size() > 0) {
+                                boolean nonPkPresent = columns.stream()
+                                        .anyMatch(column -> column.getFamilyName() != null);
```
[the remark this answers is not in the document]
virajjasani: I was thinking of doing that but then realized that it doesn't look any better either, so kept it this way in an attempt to keep it a bit cleaner. But i am fine either ways.

stoty (approved):
+1 LGTM

phoenix-core/src/main/java/org/apache/phoenix/index/IndexMaintainer.java
```diff
@@ -634,6 +638,23 @@ private Void addDataColInfo(final PTable dataTable, Expression expression) {
                         }
                         return null;
                     }
+
+                    private boolean tableHasOnlyPkCols(PTable dataTable, byte[] cf) {
+                        try {
+                            dataTable.getColumnFamily(cf);
```
tkhurana, line 644: why is this call necessary ?

tkhurana: We could make this an API on the PTable object itself ? `HasOnlyPKCols` that might be cleaner.

virajjasani: pk columns in the PTable do not have any CF associated with them, hence this call is necessary.

```
    public static boolean isPKColumn(PColumn column) {
        return column.getFamilyName() == null;
    }
```

tkhurana: But why is the function `tableHasOnlyPkCols()` taking a column object as an input is something I am not able to understand

virajjasani: because `dataTable.getColumnFamily(cf)` will fail with CFNFE for certain since the PTable has only pk columns but you are correct that PTable should have this utility rather than keeping it here

virajjasani: @tkhurana this is taken care of in the latest revision

phoenix-core/src/main/java/org/apache/phoenix/index/IndexMaintainer.java
```diff
@@ -612,6 +612,10 @@ private Void addDataColInfo(final PTable dataTable, Expression expression) {
                         byte[] cf = colExpression.getColumnFamily();
                         byte[] cq = colExpression.getColumnQualifier();
                         try {
+                            if (dataTable.hasOnlyPkColumns()) {
```
tkhurana, line 615: @virajjasani can we move this into the catch block as @stoty was also suggesting ?

virajjasani: yeah now it seems doable, meaning it is clean even when we put in catch.
done

tkhurana: Thanks @virajjasani .

tkhurana (approved):

PR #1677, opened 2023-09-18
PHOENIX-7039 Snapshot scanner should skip replay WAL and update seqid while opening region

virajjasani:


gjacoby126 (commented):
Both CommonFsUtils and MobFileCache are IA.Private. Is there no way around this that only uses IA.Public and LimitedPrivate(COPROC or PHOENIX)? I'm guessing the answer's no, but thought I'd check. 

For the future is there some kind of abstraction that would be good to have in HBase to make this low-level use of HBase internals unnecessary?

virajjasani:
Yeah i agree, i wish we could use public API directly, let me take some discussion to dev@hbase reg this. Scanning snapshot from client from restored region is valid usecase. While hbase supports it using ClientSideRegionScanner, it is also marked private. Moreover, even HRegion#initialize is deprecated, because the recommended way is to use HRegion#open directly, but that doesn't allow us to pass restored region flag.

apurtell (approved):
I agree we need a story that does not involve HBase private interfaces. 

The very first line of code, `HRegion.newHRegion()` is an even bigger problem. You can't do this. `HRegion` is `Private` and will remain so. We created `Region` to be `Public`. So this is instantiation of something private and there are a number of ways this may not work as you expect soon. Never mind additional use of private interfaces to set up parameters for this instantiation or to set properties of the resulting object as @gjacoby126 pointed out. 

I am approving this change so Phoenix can continue to function with recent versions of HBase, but this approach is fundamentally wrong. Don't create your own snapshot scanner. Use HBase's provided snapshot scanner. If it is insufficient, work with HBase to make it sufficient.

virajjasani:
Let me add TODO before merging this PR

PR #1679, opened 2023-09-19
PHOENIX-7027 Add compatibility module for Hbase 2.5.4 and upgrade Hba…

stoty:
…se version to 2.5.5

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -40,6 +40,7 @@
   </organization>
 
   <modules>
+    <module>phoenix-hbase-compat-2.5.4</module>
     <module>phoenix-hbase-compat-2.5.0</module>
     <module>phoenix-hbase-compat-2.4.1</module>
     <module>phoenix-hbase-compat-2.4.0</module>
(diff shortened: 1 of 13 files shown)


pom.xml
```diff
@@ -1649,10 +1666,25 @@
...
+      <id>phoenix-hbase-compat-2.5.0</id>
+      <activation>
+        <property>
+          <name>hbase.profile</name>
+          <value>2.5.0</value>
+        </property>
+      </activation>
+      <properties>
+        <hbase.profile>2.5.0</hbase.profile>
+        <hbase.compat.version>2.5.0</hbase.compat.version>
```
virajjasani, line 1685: shall we keep this also `2.5.4` just to maintain same compat version as before?

virajjasani (approved):

virajjasani:
build#1 timed out after 9+ hr, triggered new build: https://ci-hadoop.apache.org/job/Phoenix/job/Phoenix-PreCommit-GitHub-PR/job/PR-1679/2/

pom.xml
stoty: The convention we have used so far is that the profile without patch level, i.e. 2.5 always points to the latest supported HBase patch level (and corresponding compatibility module) within that minor version.
When we need to add a new profile, we move the patch level-less profile to point to the latest one, and create a new profile which is named after the first patch level it supports.

Before this patch we had the 2.5 profile which used the existing 2.5.0 compatibility moudule.
Now the 2.5 profile depends on the new 2.5.4 compatibility module instead, and we've added the new 2.5.0 profile, which depends on the old 2.5.0 compatibility module.

Yes, this means that you have to change the profile if you keep using an older HBase version, but 2.5 profile always works with the latest 2.5 release. We can't have it both ways unfortunately.

We could add an explicit 2.5.4 profile as an alias to 2.5, but I don't think that would improve things.

PR #1662, opened 2023-08-28
PHOENIX-7008 Shallow grammar support for CREATE CDC

haridsv:
This PR introduces basic shallow support for CREATE CDC grammar. It is being tested using UTs, but execution from a client will generate an NPE.

Jira: PHOENIX-7008

PR #1681, opened 2023-09-20
PHOENIX-7008 Implementation for CREATE CDC 

haridsv:
This PR builds on top of the shallow grammar support added for CREATE CDC grammar as part of #1662. This change makes the required syscat changes along with the creation of the underly index. There are quite a few TODO items that need to be discussed, but this PR can still be merged as we can address them in future PRs.

Jira: [PHOENIX-7008](https://issues.apache.org/jira/browse/PHOENIX-7008)


bin/sqlline.py
```diff
@@ -114,6 +116,7 @@ def kill_child():
     disable_jna = ""
 
 java_cmd = java + ' $PHOENIX_OPTS ' + \
+    (args.debug and "-Xrunjdwp:transport=dt_socket,server=y,suspend=n,address=127.0.0.1:8071" or "") + \
```
virajjasani, line 119: this is local change right?

bin/sqlline.py
```diff
@@ -123,8 +126,11 @@ def kill_child():
     '" -Dlog4j2.configurationFile=file:' + os.path.join(phoenix_utils.current_dir, "log4j2.properties") + \
     disable_jna + \
     " sqlline.SqlLine -d org.apache.phoenix.jdbc.PhoenixDriver" + \
-    " -u jdbc:phoenix:" + phoenix_utils.shell_quote([zookeeper]) + \
-    " -n none -p none --color=" + colorSetting + " --fastConnect=" + tryDecode(args.fastconnect) + \
-    " --verbose=" + tryDecode(args.verbose) + " --incremental=false --isolation=TRANSACTION_READ_COMMITTED " + sqlfile
+    (not args.noconnect and " -u jdbc:phoenix:" + phoenix_utils.shell_quote([zookeeper]) or "") + \
```
virajjasani, line 129: same here, all changes in this file can be removed

phoenix-core/src/it/java/org/apache/phoenix/end2end/CDCMiscIT.java
```diff
@@ -0,0 +1,127 @@
...
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertNull;
+import static org.junit.Assert.fail;
+
+@Category(ParallelStatsDisabledTest.class)
+public class CDCMiscIT extends ParallelStatsDisabledIT {
+    private void assertCDCState(Connection conn, String cdcName, String expInclude,
+                                int idxType) throws SQLException {
+        try (ResultSet rs = conn.createStatement().executeQuery("SELECT cdc_include FROM " +
```
virajjasani, line 40: Once all the SELECT queries to SYSTEM.CATALOG are replaced with PTable based comparison, we are good to go

bin/sqlline.py
haridsv: I am hoping to include this as it is quite useful. Do you see any problem?

bin/sqlline.py
haridsv: I would like to include the changes in the sqlline, unless there is a problem with the approach. I can also raise it as a separate PR if that is preferred.

phoenix-core/src/it/java/org/apache/phoenix/end2end/CDCMiscIT.java
haridsv: I have added the check for the change scope on `PTable` as it corresponds to the protobuf change but left the existing SYSCAT checks as is, as these are existing fields and don't necessarily have to be validated through `PTable`. WDYT?

bin/sqlline.py
virajjasani: yeah i think separate Jira would be better to track it and others can also review, i am not actually sure whether we can make the change or not because at least we don't use this script as is in production env

virajjasani:
Looks like PTableType change from previous commit caused the conflict

virajjasani:
looks good otherwise

bin/sqlline.py
haridsv: I now moved the change into [PHOENIX-7055](https://issues.apache.org/jira/browse/PHOENIX-7055) and created a separate PR #1695. Please take a look.

haridsv:
FIxed the conflict and moved out the sqlline.py change.

virajjasani:
BackwardCompatibilityIT tests are failing with error
```
ERROR 504 (42703): Undefined column. columnName=SYSTEM.CATALOG.CDC_INCLUDE
```

phoenix-core/src/main/java/org/apache/phoenix/query/ConnectionQueryServicesImpl.java
```diff
@@ -4107,6 +4109,9 @@ protected PhoenixConnection upgradeSystemCatalogIfRequired(PhoenixConnection met
             metaConnection = addColumnsIfNotExists(metaConnection, PhoenixDatabaseMetaData.SYSTEM_CATALOG,
                 MIN_SYSTEM_TABLE_TIMESTAMP_5_2_0,
```
lokiore, line 4110: Here MIN_SYSTEM_TABLE_TIMESTAMP_5_2_0 should be increased by 1 and then each addColumns above will decrease it by 1. This can lead to BackwardCompatibilityIT tests to fail, which we are seeing,

haridsv: Done, thank you!

haridsv: The tests are still failing for the same reason, but this time on the `PHYSICAL_TABLE_NAME` column (the very first one). Could you please check to see what I might have done wrong?

lokiore: Value of MIN_SYSTEM_TABLE_TIMESTAMP needs to be increased by 1 as well here :- https://github.com/apache/phoenix/blob/6125f0244ce69b86235c19b39c6abaf3095f7eab/phoenix-core/src/main/java/org/apache/phoenix/coprocessor/MetaDataProtocol.java#L102

virajjasani (approved):

PR #1687, opened 2023-09-22
PHOENIX-7034 Disallow mapped view creation when the schema does not exists

Aarchy:


phoenix-core/src/main/java/org/apache/phoenix/compile/FromCompiler.java
```diff
@@ -163,16 +163,22 @@ public static ColumnResolver getResolverForCreation(final CreateTableStatement s
...
-        if (baseTable == null) {
-            if (SchemaUtil.isSchemaCheckRequired(statement.getTableType(), connection.getQueryServices().getProps())) {
-                schemaName = statement.getTableName().getSchemaName();
-                if (schemaName != null) {
-                    new SchemaResolver(connection, statement.getTableName().getSchemaName(), true);
-                } else if (connection.getSchema() != null) {
-                    // To ensure schema set through properties or connection string exists before creating table
-                    new SchemaResolver(connection, connection.getSchema(), true);
-                }
+        if (SchemaUtil.isSchemaCheckRequired(statement.getTableType(),
```
stoty, line 166: This is the actual change.
The rest is tests and reformats.

phoenix-core/src/main/java/org/apache/phoenix/compile/FromCompiler.java
```diff
@@ -163,16 +163,22 @@ public static ColumnResolver getResolverForCreation(final CreateTableStatement s
...
+                connection.getQueryServices().getProps())) {
+            schemaName = statement.getTableName().getSchemaName();
+            if (schemaName != null) {
+                new SchemaResolver(connection,
```
stoty, line 170: This is not an intuitive way of checking for the existence of a schema.
Please add a comment with some explanation.

phoenix-core/src/main/java/org/apache/phoenix/compile/FromCompiler.java
stoty: Looking at isSchemaCheckRequired(), it only returns true when namespacemapping is enabled.
I wonder if that is correct behaviour.

stoty: Even if it isn't, we don't have to fix that in this patch.

PR #1688, opened 2023-09-24
PHOENIX-7023 : Deprecate columns PHOENIX_TTL and PHOENIX_TTL_HWM and related code

lokiore:


phoenix-core/src/main/protobuf/PTable.proto
```diff
@@ -106,19 +106,16 @@ message PTable {
   optional int32 viewIndexIdType = 39 [default = 5];
   optional bool viewModifiedUpdateCacheFrequency = 40;
   optional bool viewModifiedUseStatsForParallelization = 41;
-  optional int64 phoenixTTL = 42;
```
jpisaac, line 109: Let's not change the message format in this release. Maybe safer to remove it in a later release when the columns are removed. We will need to ensure backward and forward compatibility of messages against various versions of the client and server.

jpisaac (commented):
LGTM +1, Few nits.
Let's chat offline about the message format change

jpisaac (approved):
LGTM +1

PR #1690, opened 2023-09-25
PHOENIX-6053 Split Server Side Code into a Separate Module

Aarchy:


phoenix-core/src/main/java/org/apache/phoenix/schema/task/Task.java
```diff
@@ -62,67 +57,70 @@
...
-        }
-        else {
-            try {
-                stmt.execute();
-                conn.commit();
-            } catch (SQLException e) {
-                throw new IOException(e);
-            }
+
+//    private static List<Mutation> getMutationsForSystemTaskTable(
```
stoty, line 61: In the final version remove code instead of commenting it out.

phoenix-core/src/main/java/org/apache/phoenix/schema/transform/TransformMaintainer.java
```diff
@@ -21,7 +21,6 @@
 import org.apache.phoenix.thirdparty.com.google.common.collect.Maps;
 import org.apache.phoenix.thirdparty.com.google.common.collect.Sets;
 
-import org.apache.hadoop.hbase.client.Delete;
```
stoty, line 24: I can see a lot of files where you are just randomly removing unused imports.
This change is already huge, leaving these alone for now would make it somewhat more managable.

We can handle unrelated cleanups in separate tickets.

This only applies when you only remove non-server unused imports.
When there are other changes, it's fine to clean up them.

phoenix-core/src/main/java/org/apache/phoenix/util/ClientUtil.java
```diff
@@ -0,0 +1,195 @@
...
+import org.apache.hadoop.hbase.NotServingRegionException;
+import org.apache.hadoop.hbase.client.Admin;
+import org.apache.hadoop.hbase.client.RetriesExhaustedWithDetailsException;
+import org.apache.phoenix.coprocessor.HashJoinCacheNotFoundException;
+import org.apache.phoenix.exception.PhoenixIOException;
+import org.apache.phoenix.exception.SQLExceptionCode;
+import org.apache.phoenix.exception.SQLExceptionInfo;
+import org.apache.phoenix.schema.StaleRegionBoundaryCacheException;
+
+public class ClientUtil {
```
stoty, line 39: When splitting classes to client and server ones, check where it used more often.
In some cases it may beneficial to keep the old name for the client class, and in some cases it may be better to keep it for the server class, whichever makes more sense, or makes the patch smaller.

stoty (changes requested):
Thank you.

I've left some general comments.
We can talk them over tomorrow.

phoenix-core/src/main/java/org/apache/phoenix/index/PhoenixIndexBuilderStatic.java
```diff
@@ -0,0 +1,146 @@
...
+import org.apache.hadoop.hbase.util.Bytes;
+import org.apache.hadoop.io.WritableUtils;
+import org.apache.phoenix.coprocessor.generated.PTableProtos;
+import org.apache.phoenix.expression.Expression;
+import org.apache.phoenix.expression.ExpressionType;
+import org.apache.phoenix.schema.PTable;
+import org.apache.phoenix.schema.PTableImpl;
+import org.apache.phoenix.util.TrustedByteArrayOutputStream;
+
+public final class PhoenixIndexBuilderStatic {
```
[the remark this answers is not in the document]
Aarchy: Yes, I used those names.,I'll rename them

phoenix-coprocessors/src/main/java/org/apache/phoenix/schema/transform/Transform.java
```diff
@@ -17,19 +17,14 @@
  */
 package org.apache.phoenix.schema.transform;
 
-import com.fasterxml.jackson.core.JsonProcessingException;
 import org.apache.hadoop.hbase.util.Bytes;
-import org.apache.phoenix.coprocessor.TableInfo;
+import org.apache.phoenix.scanattributes.TableInfo;
```
stoty, line 21: Find some other package name, this is not a scan attribute.
(I know it's because I asked you to rename the package to scanatributes.)

phoenix-coprocessors/src/main/java/org/apache/phoenix/schema/transform/Transform.java
```diff
@@ -67,192 +57,24 @@
...
 import static org.apache.phoenix.query.QueryServices.MUTATE_BATCH_SIZE_ATTRIB;
 import static org.apache.phoenix.schema.ColumnMetaDataOps.addColumnMutation;
-import static org.apache.phoenix.schema.MetaDataClient.CREATE_LINK;
 import static org.apache.phoenix.schema.MetaDataClient.UPDATE_INDEX_STATE_TO_ACTIVE;
 import static org.apache.phoenix.schema.PTable.ImmutableStorageScheme.SINGLE_CELL_ARRAY_WITH_OFFSETS;
 import static org.apache.phoenix.schema.PTableType.INDEX;
-import static org.apache.phoenix.schema.PTableType.VIEW;
 
-public class Transform {
+public class Transform extends TransformHelper {
```
stoty, line 73: Would it mke sense to call TransaformHelper TransformClient ?

phoenix-core/src/main/java/org/apache/phoenix/cache/ServerCacheClient.java
```diff
@@ -51,15 +51,14 @@
 import org.apache.hadoop.hbase.util.Bytes;
 import org.apache.phoenix.compile.QueryPlan;
 import org.apache.phoenix.compile.ScanRanges;
-import org.apache.phoenix.coprocessor.MetaDataProtocol;
-import org.apache.phoenix.coprocessor.ServerCachingProtocol.ServerCacheFactory;
+import org.apache.phoenix.scanattributes.MetaDataProtocol;
```
stoty, line 54: scanattributes package name is a poor fit for these.
maybe we should add another package for the non-attributes, or move them into another package.

phoenix-core/src/main/java/org/apache/phoenix/cache/TenantCache.java
```diff
@@ -21,10 +21,9 @@
 import java.sql.SQLException;
 
 import org.apache.hadoop.hbase.io.ImmutableBytesWritable;
-import org.apache.phoenix.coprocessor.ServerCachingProtocol.ServerCacheFactory;
 import org.apache.phoenix.hbase.index.util.ImmutableBytesPtr;
 import org.apache.phoenix.memory.MemoryManager;
-
+import org.apache.phoenix.scanattributes.ServerCachingProtocol.ServerCacheFactory;
```
stoty, line 26: ditto

phoenix-core/src/main/java/org/apache/phoenix/exception/SQLExceptionCode.java
```diff
@@ -22,11 +22,11 @@
 import java.sql.SQLTimeoutException;
 import java.util.Map;
 
-import org.apache.phoenix.coprocessor.MetaDataProtocol;
 import org.apache.phoenix.hbase.index.util.IndexManagementUtil;
 import org.apache.phoenix.jdbc.PhoenixDatabaseMetaData;
 import org.apache.phoenix.query.QueryConstants;
 import org.apache.phoenix.query.QueryServices;
+import org.apache.phoenix.scanattributes.MetaDataProtocol;
```
stoty, line 29: change not needed

phoenix-core/src/main/java/org/apache/phoenix/execute/PhoenixTxIndexMutationGenerator.java
```diff
@@ -64,6 +62,7 @@
 import org.apache.phoenix.jdbc.PhoenixConnection;
 import org.apache.phoenix.query.ConnectionQueryServices;
 import org.apache.phoenix.query.KeyRange;
+import org.apache.phoenix.scanattributes.MetaDataProtocol;
```
stoty, line 65: change not needed

phoenix-core/src/main/java/org/apache/phoenix/filter/SystemCatalogViewIndexIdFilter.java
```diff
@@ -36,10 +36,10 @@
 import java.util.List;
 
 
-import static org.apache.phoenix.coprocessor.MetaDataProtocol.MIN_SPLITTABLE_SYSTEM_CATALOG;
 import static org.apache.phoenix.jdbc.PhoenixDatabaseMetaData.VIEW_INDEX_ID_BYTES;
 import static org.apache.phoenix.jdbc.PhoenixDatabaseMetaData.VIEW_INDEX_ID_DATA_TYPE_BYTES;
 import static org.apache.phoenix.query.QueryConstants.DEFAULT_COLUMN_FAMILY_BYTES;
+import static org.apache.phoenix.scanattributes.MetaDataProtocol.MIN_SPLITTABLE_SYSTEM_CATALOG;
```
stoty, line 42: change not needed

phoenix-core/src/main/java/org/apache/phoenix/hbase/index/IndexerStatic.java
```diff
@@ -0,0 +1,51 @@
...
+import java.io.IOException;
+import java.util.HashMap;
+import java.util.Map;
+
+import org.apache.hadoop.hbase.client.CoprocessorDescriptorBuilder;
+import org.apache.hadoop.hbase.client.TableDescriptor;
+import org.apache.hadoop.hbase.client.TableDescriptorBuilder;
+import org.apache.phoenix.query.QueryConstants;
+
+public class IndexerStatic {
```
stoty, line 29: Move to IndexUtil or rename to IndexerUtil ?

stoty (changes requested):
Left some suggestions about names and packages.
Please do another pass to undo the rest of the changes that only affect import order or unused imports.

phoenix-coprocessors/src/main/java/org/apache/phoenix/mapreduce/transform/PhoenixTransformWithViewsInputFormat.java
```diff
@@ -27,11 +27,11 @@
 import org.apache.phoenix.compile.MutationPlan;
 import org.apache.phoenix.compile.QueryPlan;
 import org.apache.phoenix.compile.ServerBuildTransformingTableCompiler;
-import org.apache.phoenix.coprocessor.TableInfo;
 import org.apache.phoenix.jdbc.PhoenixConnection;
 import org.apache.phoenix.mapreduce.PhoenixServerBuildIndexInputFormat;
 import org.apache.phoenix.mapreduce.util.ConnectionUtil;
 import org.apache.phoenix.mapreduce.util.PhoenixConfigurationUtil;
+import org.apache.phoenix.scanattributes.TableInfo;
```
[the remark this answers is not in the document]
Aarchy: Also package name is changed (from coprocessor to scanatribbutes)

stoty:
This looks OK to me.
I will invite the community for review.

stoty:
Can you rebase to the current master (unfortunately a few more rebases will likely be needed while the discussion is ongoing) ?
You may also want to squash the commits.

stoty:
I see you haven't touched phoenix-client.
While the ultimate goal is to have a slimmer client, we plan to make a new artifact for that.
For now, phoenix-client should depend on phoenix-coprocessor, so it is basically unchanged.

stoty:
Unfortunately the Zookeeperless code made another large-ish rebase necessary.

richardantal (commented):
This change is huge, I tried to check it reasonably thorough but there might be things that I missed.
The change looks mostly good I only left some minor comments on it.

phoenix-coprocessors/src/main/java/org/apache/phoenix/coprocessor/UncoveredIndexRegionScanner.java
```diff
@@ -321,10 +324,12 @@ private boolean getNextCoveredIndexRow(List<Cell> result) throws IOException {
                     }
                 } else {
                     if (indexMaintainer.isUncovered()) {
+                        long ts = indexRow.get(0).getTimestamp();
                         // Since we also scan the empty column for uncovered global indexes, this mean the data row
                         // does not exist. Delete the index row if the index is an uncovered global index
-                        indexMaintainer.deleteRowIfAgedEnough(indexRowKey, indexRow.get(0).getTimestamp(),
-                                ageThreshold, false, region);
+                        if (indexMaintainer.isAgedEnough(ts, ageThreshold)) {
+                            region.delete(indexMaintainer.createDelete(indexRowKey, ts, false));
```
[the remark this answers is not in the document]
Aarchy: According to hbase documentation:

batchMutate: Perform a batch of mutations. Note this supports only Put and Delete mutations and will ignore other types passed.

delete: Deletes the specified cells/row.

In this specific case the batch contained only 1 delete, so changing it to a single delete operation should do the same.

phoenix-core/src/main/java/org/apache/phoenix/coprocessorclient/UngroupedAggregateRegionObserverConstants.java
```diff
@@ -0,0 +1,59 @@
...
+import java.util.List;
+
+import org.apache.hadoop.io.WritableUtils;
+import org.apache.phoenix.coprocessor.generated.PTableProtos;
+import org.apache.phoenix.expression.Expression;
+import org.apache.phoenix.expression.ExpressionType;
+import org.apache.phoenix.schema.PTable;
+import org.apache.phoenix.schema.PTableImpl;
+
+public class UngroupedAggregateRegionObserverConstants {
```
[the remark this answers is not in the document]
Aarchy: that's right, this should definitely be a helper class. thanks.

phoenix-coprocessors/src/main/java/org/apache/hadoop/hbase/regionserver/IndexHalfStoreFileReaderGenerator.java
```diff
@@ -210,7 +210,7 @@ public InternalScanner preCompact(ObserverContext<RegionCoprocessorEnvironment>
             InternalScanner s, ScanType scanType, CompactionLifeCycleTracker tracker,
             CompactionRequest request) throws IOException {
 
-        if (!IndexUtil.isLocalIndexStore(store)) { return s; }
```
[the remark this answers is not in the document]
Aarchy: I think Store is defined in hbase-server package, IndexUtil is in phoenix-core. We want to break (phoenix-core -> hbase-server) dependency so this function needs to be moved. As its only used in this file, which is indeed in pheonix-coprocessors (which can use hbase-server), I moved it here.

Aarchy:
@richardantal updated the PR based on your comments.
Thank you for taking time to review this huge patch, really appreciate the feedback!

phoenix-coprocessors/src/main/java/org/apache/phoenix/coprocessor/UncoveredIndexRegionScanner.java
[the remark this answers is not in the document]
richardantal: yep, that is right, that should do the same.
I was curious if you know if there is some performance difference between the 2.

richardantal (approved):
Thank you for doing this change.
I really appreciate the effort that went into this one.
Keep up the fantastic work.

stoty (changes requested):
Phoenix-server is broken.
You need to add the phoenix-coprocessors dependency (or rather replace the phoenix-core dependency with phoenix-coprocessors which should pull in core transitively)

stoty:
I suggested moving BaseIndexCodec one package higher, but now I see that it was unneccessary.
Can you move it back ? Sorry for flip-flopping.

stoty:
ServerIndexUtil and ServerViewUtil get added to phoenix-client-embedded.
I'm trying to figure out why.

Aarchy:
@stoty Added phoenix-coprocessors dependency to phoenix-server and modified the commit msg.

stoty:
I got this error when running mvn install.

[INFO] --- maven-dependency-plugin:3.1.1:analyze-only (enforce-dependencies) @ phoenix-tests ---
[WARNING] Used undeclared dependencies found:
[WARNING]    org.apache.zookeeper:zookeeper-jute:jar:3.5.7:compile
[INFO] ------------------------------------------------------------------------

Test dependencies are not transitive, you need to add everything that phoenix-core depends on.
(Or just add phoenix-core as a non-test dependency)

stoty:
Come to think of it, we should do this properly.
Remove the transitive dependencies from phoenix-tests, and set the phoenix-core dependency to compile.
You may have to add an exclusion for this in the maven-dependency-plugin config

stoty:
Ignore my previous comment, I was wrong.

stoty:
We could in fact change the scope to test for most dependencies.

stoty:
You only need to add the zookeeper-jute dependency

Aarchy:
Weird, if I add zookeeper-jute I get the following error:

[INFO] --- dependency:3.1.1:analyze-only (enforce-dependencies) @ phoenix-tests ---
[WARNING] Unused declared dependencies found:
[WARNING]    org.apache.zookeeper:zookeeper-jute:jar:3.5.7:test

stoty:
It depends on the java version.
Java 11 fails with "Used undeclared dependencies found:"

But if we add it, then Java 8 fails with "Unused declared dependencies found:"

Please add the exclusion both ways into the maven-dependency-plugin config, with a comment explaining the situation.

Aarchy:
Do we need to add two way exclusion? 
I suppose just one of these warnings is right, so wouldn't it be better to just add exclusion for the faulty case? This way, at least the declared dependencies would be correct.

stoty:
That's also fine.

I have disassembled tests, and even though there is no include for zookeeper-jute, it does in fact call a method whoose return value is defined in zk-jute.

Why this is dependent on java version is beyond me.

Aarchy:
@stoty I've added the zk-jute dependency and also the exclusion.

stoty:
Thank you, @Aarchy .

stoty:
As discussed ofline, we have found a more compatible module naming scheme:
phoenix-core -> phoenix-core-client
phoenix-coprocessors -> phoenix-core-server
phoenix-test > phoenix-core (the original module)

phoenix-core depends on phoenix-core-server and phoenix-server-client, and all the tests.

This maximizes compatbility with existing maven consumers and downstream projects.

stoty:
You also need adopt (or rather revert) phoenix-sandbox.py to these changes.

stoty (approved):
+1 LGTM

stoty:
The hang is definielty caused by this patch.
Maybe a bad merge ?

stoty:
So I have confirmed that the hang is caused by this patch. The test ran fine more than five times in a row without the patch, but failed every time I tried with the patch.

It doesn't seem to be a problem with the test code itself, so the patch must have introduced some other error, which causes the hang.

Now we just need to figure out which one of the 1300 changed files went wrong.

Aarchy:
@stoty Thanks for pointing this out and the offline help as well.
There were two enableIndexing functions (one in Indexer.java and another one in IndexRegionObserver.java) but they set different coprocessors. I missed it, and just kept the IndexRegionObserver version in IndexerUtil. Changing it to use the correct coprocessors fixed the issue.

PR #1691, opened 2023-09-25
PHOENIX-6968 Create PhoenixRegionServerEndpoint#invalidateCache method to invalidate cache.

shahrs87:


phoenix-core/src/test/java/org/apache/phoenix/query/BaseTest.java
```diff
@@ -418,7 +420,7 @@ public abstract class BaseTest {
     }
     
     private static final String ORG_ID = "00D300000000XHP";
-    protected static int NUM_SLAVES_BASE = 1;
+    protected static int NUM_SLAVES_BASE = 3;
```
stoty, line 423: This doesn't look like a good idea.
It would be better to override this on the tests where it is needed

shahrs87: @stoty  This PR is not ready for review yet. I was just trying to run all the tests to see which ones fails in the first pass. 
I should have labelled the PR as Draft PR. Sorry for the confusion.

shahrs87:
@stoty  @tkhurana  @virajjasani  This PR is ready for review now. Can you please help review? Thank you.

phoenix-core/src/it/java/org/apache/phoenix/end2end/InvalidateMetadataCacheIT.java
```diff
@@ -0,0 +1,141 @@
...
+                getFailingPhoenixRegionServerEndpoint(regionServerZero);
+        coprocForRS0.sleep();
+        try(Connection conn = DriverManager.getConnection(getUrl(), props)) {
+            conn.createStatement().execute(ddl);
+            conn.createStatement().execute("ALTER TABLE " + dataTableFullName +
+                    " ADD CF.col2 integer");
+            fail("Shouldn't reach heere");
+        } catch (Exception e) {
+            LOGGER.error("Exception while adding column", e);
+            // This is expected
```
ranganathg, line 80: should we add assertTrue in this exception? similarly below?

phoenix-core/src/it/java/org/apache/phoenix/end2end/FailingPhoenixRegionServerEndpoint.java
```diff
@@ -0,0 +1,72 @@
...
+            return;
+        } else if (shouldSleep) {
+            try {
+                // Sleeping for 2 seconds more than metadataCacheInvalidationTimeoutMs.
+                Thread.sleep(metadataCacheInvalidationTimeoutMs + 2000);
+            } catch (InterruptedException e) {
+                LOGGER.warn("Exception while sleeping in FailingPhoenixRegionServerEndpoint", e);
+            }
+        } else {
+            LOGGER.info("Invalidating server metadata cache");
```
ranganathg, line 59: should there be a call to invalidateServerMetadataCache()? how does the cache get invalidated without any call?

shahrs87: > how does the cache get invalidated without any call?
We don't want to invalidate anything here. This is FailingPhoenixRegionServerEndpoint which just simulates failure case like timeout and exception throwing.

phoenix-core/src/it/java/org/apache/phoenix/end2end/InvalidateMetadataCacheIT.java
shahrs87: I thought that is kind of implicit. There is an `assert fail` statement in the try block and if it throws anything other than Exception, the test will fail. So it has to go in `catch (Exception)` block for the test to succeed.

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/MetaDataEndpointImpl.java
```diff
@@ -3416,6 +3437,167 @@ private MetaDataMutationResult mutateColumn(
...
+     * @param serverNames
+     * @param tenantId
+     * @param schemaName
+     * @param tableOrViewName
+     * @param retryAttempt
+     * @throws Throwable
+     */
+    private void invalidateServerMetadataCacheWithRetries(Admin admin,
+            Collection<ServerName> serverNames, byte[] tenantId, byte[] schemaName,
+            byte[] tableOrViewName, boolean retryAttempt) throws Throwable {
```
tkhurana, line 3489: better to rename the variable to `isRetry` which signifies its a boolean rather than retryAttempt which denotes a counter

phoenix-core/src/it/java/org/apache/phoenix/end2end/InvalidateMetadataCacheIT.java
```diff
@@ -0,0 +1,141 @@
...
+    public void testAddColumnWithTimeout() {
+        Properties props = PropertiesUtil.deepCopy(TEST_PROPERTIES);
+        String dataTableFullName = generateUniqueName();
+        String ddl = getCreateTableStmt(dataTableFullName);
+        HRegionServer regionServerZero = utility.getMiniHBaseCluster().getRegionServer(0);
+        FailingPhoenixRegionServerEndpoint coprocForRS0 =
+                getFailingPhoenixRegionServerEndpoint(regionServerZero);
+        coprocForRS0.sleep();
+        try(Connection conn = DriverManager.getConnection(getUrl(), props)) {
+            conn.createStatement().execute(ddl);
```
tkhurana, line 74: So the CREATE TABLE ddl will not trigger cache invalidation ?

phoenix-core/src/it/java/org/apache/phoenix/end2end/InvalidateMetadataCacheIT.java
```diff
@@ -0,0 +1,141 @@
...
+            conn.createStatement().execute("ALTER TABLE " + dataTableFullName +
+                    " ADD CF.col2 integer");
+            fail("Shouldn't reach heere");
```
tkhurana, line 77: typo heere

phoenix-core/src/it/java/org/apache/phoenix/end2end/FailingPhoenixRegionServerEndpoint.java
```diff
@@ -0,0 +1,72 @@
...
+        }
+    }
+
+    public void throwException() {
+        this.throwException = true;
+        this.shouldSleep = false;
+    }
+
+    public void sleep() {
```
tkhurana, line 68: maybe add another API to reset these states

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/MetaDataEndpointImpl.java
```diff
@@ -3416,6 +3437,167 @@ private MetaDataMutationResult mutateColumn(
...
+        String fullTableName = SchemaUtil.getTableName(schemaName, tableOrViewName);
+        String tenantIDStr = Bytes.toString(tenantId);
+        LOGGER.info("Invalidating metadata cache for tenantID: {}, tableName: {} for"
+                        + " region servers: {}, isRetryAttempt: {}", tenantIDStr, fullTableName,
+                serverNames, retryAttempt);
+        RegionServerEndpointProtos.InvalidateServerMetadataCacheRequest request =
+                getRequest(tenantId, schemaName, tableOrViewName);
+        // TODO Do I need my own executor or can I re-use QueryServices#Executor
```
tkhurana, line 3497: What is the threadpool size of the default executor ? Maybe it is better to create a new executor so that you can control its size

phoenix-core/src/it/java/org/apache/phoenix/end2end/InvalidateMetadataCacheIT.java
```diff
@@ -0,0 +1,141 @@
...
+    }
+
+    private FailingPhoenixRegionServerEndpoint getFailingPhoenixRegionServerEndpoint(
+            HRegionServer regionServer) {
+        FailingPhoenixRegionServerEndpoint coproc = regionServer
+                .getRegionServerCoprocessorHost()
+                .findCoprocessor(FailingPhoenixRegionServerEndpoint.class);
+        return coproc;
+    }
+}
```
tkhurana, line 141: You should also add a test where the invalidation is successful on retry

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/MetaDataEndpointImpl.java
```diff
@@ -3165,10 +3184,12 @@ private MetaDataMutationResult mutateColumn(
                 if (mutationResult.isPresent()) {
                     return mutationResult.get();
                 }
-
+                // We take a write row lock for tenantId, schemaName, tableOrViewName
                 acquireLock(region, key, locks);
+                // Invalidate the cache from all the regionservers.
+                invalidateServerMetadataCache(tenantId, schemaName, tableOrViewName);
```
tkhurana, line 3190: We need to call invalidate cache in other scenarios also like create index. Will that come in a separate JIRA ?

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/MetaDataEndpointImpl.java
ranganathg: also when I looked at the function name "WithRetries" I thought there would a RetryManager that would use some retry strategy. May be the name can be with just invalidateServerMetadataCache - with one param being isRetry?

shahrs87: > also when I looked at the function name "WithRetries" I thought there would a RetryManager that would use some retry strategy.

We can add RetryManager later if needed. We already have method name `invalidateServerMetadataCache` above.

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/MetaDataEndpointImpl.java
shahrs87: > What is the threadpool size of the default executor ? 

https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/ForkJoinPool.html#ForkJoinPool--
Currently default is ForkJoinPool which uses the number of threads as many as available processors.

shahrs87: >  Maybe it is better to create a new executor so that you can control its size

@tkhurana  Do you think currently we can keep it as it is and use custom threadpool if needed?

phoenix-core/src/it/java/org/apache/phoenix/end2end/InvalidateMetadataCacheIT.java
shahrs87: Trying to think how can I make it succeed on the retry. Let me try something.

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/MetaDataEndpointImpl.java
shahrs87: > We need to call invalidate cache in other scenarios also like create index. 

Yes, I have already created jira to update LAST_DDL_TIMESTAMP of the parent table/view when we add an index. [PHOENIX-7056](https://issues.apache.org/jira/browse/PHOENIX-7056)

shahrs87: > Will that come in a separate JIRA ?

Yes.

phoenix-core/src/it/java/org/apache/phoenix/end2end/FailingPhoenixRegionServerEndpoint.java
shahrs87: Addressed in latest commit. Please review again. Thank you.

phoenix-core/src/it/java/org/apache/phoenix/end2end/InvalidateMetadataCacheIT.java
shahrs87: Addressed in latest commit. Please review again. Thank you.

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/MetaDataEndpointImpl.java
tkhurana: Sounds good @shahrs87

tkhurana (approved):

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/MetaDataEndpointImpl.java
```diff
@@ -3416,6 +3437,167 @@ private MetaDataMutationResult mutateColumn(
...
+            map.put(future, serverName);
+        }
+
+        // Here we create one master like future which tracks individual future
+        // for each region server.
+        CompletableFuture<Void> allFutures = CompletableFuture.allOf(
+                futures.toArray(new CompletableFuture[0]));
+        try {
+            allFutures.get(metadataCacheInvalidationTimeoutMs, TimeUnit.MILLISECONDS);
+        } catch (Throwable t) {
```
haridsv, line 3542: This catches every kind of exception, is it desirable to retry in error cases? Why not be specific to those that are just declared?
```suggestion
        } catch (CancellationException | ExecutionException | InterruptedException | TimeoutException t) {
```

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/MetaDataEndpointImpl.java
```diff
@@ -3416,6 +3437,167 @@ private MetaDataMutationResult mutateColumn(
...
+            futures.add(future);
+            map.put(future, serverName);
+        }
+
+        // Here we create one master like future which tracks individual future
+        // for each region server.
+        CompletableFuture<Void> allFutures = CompletableFuture.allOf(
+                futures.toArray(new CompletableFuture[0]));
+        try {
+            allFutures.get(metadataCacheInvalidationTimeoutMs, TimeUnit.MILLISECONDS);
```
haridsv, line 3541: Perhaps we should have a discount for retries?

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/MetaDataEndpointImpl.java
```diff
@@ -3416,6 +3437,167 @@ private MetaDataMutationResult mutateColumn(
...
+        } catch (Throwable t) {
+            List<ServerName> failedServers = getFailedServers(futures, map);
+            LOGGER.error("Invalidating metadata cache for tenantID: {}, tableName: {} failed for "
```
haridsv, line 3544: This should be an error only if we are not going to retry?

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/MetaDataEndpointImpl.java
```diff
@@ -3416,6 +3437,167 @@ private MetaDataMutationResult mutateColumn(
...
+    }
+
+    /*
+        Get the list of regionservers that failed the invalidateCache rpc.
+     */
+    private List<ServerName> getFailedServers(List<CompletableFuture<Void>> futures,
+                                              Map<Future, ServerName> map) {
+        List<ServerName> failedServers = new ArrayList<>();
+        for (CompletableFuture completedFuture : futures) {
+            if (completedFuture.isDone() == false) {
```
haridsv, line 3570: I am not familiar with futures, but what happens in case of a timeout? Will it have a `false` value? In the timeout case, won't the remote RPC thread be in a stuck state? What is the chance of a retry making progress, besides initiating a concurrent request?

phoenix-core/src/main/protobuf/RegionServerEndpointService.proto
```diff
@@ -36,7 +36,24 @@ message LastDDLTimestampRequest {
   required int64 lastDDLTimestamp = 4;
 }
 
-service RegionServerCoprocService {
+message InvalidateServerMetadataCache {
```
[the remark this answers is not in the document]
shahrs87: We haven't deployed the previous changes anywhere. This was still in the dev branch so it was not important to handle backwards compatibility. But I am not sure how we would handle the backwards compatibility once the release was cut off.

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/MetaDataEndpointImpl.java
shahrs87: In case of non declared exception, what do you think we should do? Should we just fail the DDL operation and not retry? 
I was thinking to catch all the exception, retry once and if it still fails then fail the operation.

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/MetaDataEndpointImpl.java
shahrs87: I can change it to WARN. WDYT? @haridsv

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/MetaDataEndpointImpl.java
shahrs87: > Will it have a false value?

Yes, it will have false value. We also have [test case](https://github.com/apache/phoenix/blob/PHOENIX-6883-feature/phoenix-core/src/it/java/org/apache/phoenix/end2end/InvalidateMetadataCacheIT.java#L67) for this.

> In the timeout case, won't the remote RPC thread be in a stuck state? What is the chance of a retry making progress, besides initiating a concurrent request?

In a later jira [PHOENIX-7115](https://issues.apache.org/jira/browse/PHOENIX-7115), we have added special handler threads on each regionserver just to handle invalidate cache RPCs. Earlier invalidate cache RPCs were handled by default handler threads. So the probability of these RPC threads being stuck is very small but if it happens we can increase the number of threads if necessary.

PR #1697, opened 2023-09-29
PHOENIX-7004 : Validate the result returned by phoenix query is within scan range, if provided

Divneet18:
Jira: PHOENIX-7004

Added a check to make sure query is within start and end bounds

### DIFF
--- a/phoenix-core/src/main/java/org/apache/phoenix/exception/QueryOutOfScanRangeException.java
+++ b/phoenix-core/src/main/java/org/apache/phoenix/exception/QueryOutOfScanRangeException.java
@@ -1,3 +1,22 @@
+/*
+ *
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
 package org.apache.phoenix.exception;
 
 import java.sql.SQLException;
@@ -6,7 +25,8 @@ public class QueryOutOfScanRangeException extends SQLException {
 
     private static SQLExceptionCode code = SQLExceptionCode.VALUE_OUTSIDE_RANGE;
 
-    public QueryOutOfScanRangeException(String message){
-        super(new SQLExceptionInfo.Builder(code).setMessage(message).build().toString(), code.getSQLState(), code.getErrorCode());
+    public QueryOutOfScanRangeException(String message) {
+        super(new SQLExceptionInfo.Builder(code).setMessage(message).build().toString(),
+                code.getSQLState(), code.getErrorCode());
     }
 }
--- a/phoenix-core/src/main/java/org/apache/phoenix/exception/QueryOutOfScanRangeException.java
+++ b/phoenix-core/src/main/java/org/apache/phoenix/exception/QueryOutOfScanRangeException.java
@@ -1,0 +1,12 @@
+package org.apache.phoenix.exception;
+
+import java.sql.SQLException;
+
+public class QueryOutOfScanRangeException extends SQLException {
+
+    private static SQLExceptionCode code = SQLExceptionCode.VALUE_OUTSIDE_RANGE;
+
+    public QueryOutOfScanRangeException(String message){
+        super(new SQLExceptionInfo.Builder(code).setMessage(message).build().toString(), code.getSQLState(), code.getErrorCode());
+    }
+}
--- a/phoenix-core/src/main/java/org/apache/phoenix/iterate/TableResultIterator.java
+++ b/phoenix-core/src/main/java/org/apache/phoenix/iterate/TableResultIterator.java
@@ -36,6 +36,7 @@ import java.util.concurrent.locks.ReentrantLock;
 
 import javax.annotation.concurrent.GuardedBy;
 
+import org.apache.hadoop.hbase.HConstants;
 import org.apache.hadoop.hbase.client.AbstractClientScanner;
 import org.apache.hadoop.hbase.client.Scan;
 import org.apache.hadoop.hbase.client.Table;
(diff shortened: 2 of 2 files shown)


phoenix-core/src/main/java/org/apache/phoenix/exception/QueryOutOfScanRangeException.java
```diff
@@ -0,0 +1,12 @@
+package org.apache.phoenix.exception;
```
virajjasani, line 1: please add apache header info

phoenix-core/src/main/java/org/apache/phoenix/iterate/TableResultIterator.java
```diff
@@ -186,20 +188,45 @@ public Tuple next() throws SQLException {
                 if (lastTuple != null) {
                     ImmutableBytesWritable ptr = new ImmutableBytesWritable();
                     lastTuple.getKey(ptr);
+                    if (scan.getStartRow() != null
+                            && Bytes.compareTo(scan.getStartRow(), HConstants.EMPTY_START_ROW) != 0) {
+                        if (scan.includeStartRow()) {
+                            if (Bytes.compareTo(scan.getStartRow(), ptr.get()) > 0) {
+                                throw new QueryOutOfScanRangeException("Query returned is out of scan range");
```
virajjasani, line 195: nit: `\s"out of scan range"\"out of scan start rowkey range"`

phoenix-core/src/main/java/org/apache/phoenix/iterate/TableResultIterator.java
```diff
@@ -186,20 +188,45 @@ public Tuple next() throws SQLException {
...
+                            }
+                        }
+                    }
+
+                    if (scan.getStopRow() != null
+                            && Bytes.compareTo(scan.getStopRow(), HConstants.EMPTY_END_ROW) != 0) {
+                        if (scan.includeStopRow()) {
+                            if (Bytes.compareTo(scan.getStopRow(), ptr.get()) < 0) {
+                                throw new QueryOutOfScanRangeException("Query returned is out of scan range");
```
virajjasani, line 208: nit: `nit: \s"out of scan range"\"out of scan stop rowkey range"`

virajjasani (dismissed):
left nits reg error msg, otherwise good to go

shahrs87:
@Divneet18  There are 615 test failures.  https://ci-hadoop.apache.org/job/Phoenix/job/Phoenix-PreCommit-GitHub-PR/job/PR-1697/2/testReport/

shahrs87:
Also is there any way to mock the result set and return the mocked value and make sure the following test scenarios.
1. start key is set but returned value is less than start key and the exception is thrown.
2. stop key is specified but returned value is greater than stop key and the exception is thrown.
@Divneet18

virajjasani:
Test failures are relevant

phoenix-core/src/main/java/org/apache/phoenix/exception/QueryOutOfScanRangeException.java
```diff
@@ -0,0 +1,32 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.phoenix.exception;
+
+import java.sql.SQLException;
+
+public class QueryOutOfScanRangeException extends SQLException {
```
shahrs87, line 24: s/QueryOutOfScanRangeException/ResultSetOutOfScanRangeException

virajjasani:
Will need to revisit this, looks like we missed something fundamental here

phoenix-core/src/main/java/org/apache/phoenix/iterate/TableResultIterator.java
```diff
@@ -186,20 +188,45 @@ public Tuple next() throws SQLException {
                 if (lastTuple != null) {
                     ImmutableBytesWritable ptr = new ImmutableBytesWritable();
                     lastTuple.getKey(ptr);
+                    if (scan.getStartRow() != null
+                            && Bytes.compareTo(scan.getStartRow(), HConstants.EMPTY_START_ROW) != 0) {
+                        if (scan.includeStartRow()) {
```
shahrs87, line 193: What is the difference between `scan.getStartRow() != null` and `scan.includeStartRow()` ? 
Can we exclude one check if both are the same?

phoenix-core/src/main/java/org/apache/phoenix/iterate/TableResultIterator.java
```diff
@@ -186,20 +188,45 @@ public Tuple next() throws SQLException {
...
+                        } else {
+                            if (Bytes.compareTo(scan.getStartRow(), ptr.get()) >= 0) {
+                                throw new QueryOutOfScanRangeException("Query returned is out of scan range");
+                            }
+                        }
+                    }
+
+                    if (scan.getStopRow() != null
+                            && Bytes.compareTo(scan.getStopRow(), HConstants.EMPTY_END_ROW) != 0) {
+                        if (scan.includeStopRow()) {
```
shahrs87, line 206: What is the difference between `scan. getStopRow() != null` and `scan.includeStopRow()` ? 
Can we exclude one check if both are the same?

phoenix-core/src/main/java/org/apache/phoenix/iterate/TableResultIterator.java
```diff
@@ -186,20 +188,45 @@ public Tuple next() throws SQLException {
...
                     ImmutableBytesWritable ptr = new ImmutableBytesWritable();
                     lastTuple.getKey(ptr);
+                    if (scan.getStartRow() != null
+                            && Bytes.compareTo(scan.getStartRow(), HConstants.EMPTY_START_ROW) != 0) {
+                        if (scan.includeStartRow()) {
+                            if (Bytes.compareTo(scan.getStartRow(), ptr.get()) > 0) {
+                                throw new QueryOutOfScanRangeException("Query returned is out of scan range");
+                            }
+                        } else {
+                            if (Bytes.compareTo(scan.getStartRow(), ptr.get()) >= 0) {
```
ranganathg, line 198: why is equality check required for start and stop both? are both the end exclusive?

dbwong (commented):
In general where are the test cases that make sure this exception is thrown when necessary?  Especially reverse scan failure type cases.  Also validating every row seems expensive and not even possible for cases where we don't project rowkeys for a server side scan right?

virajjasani:
Now that i think about this Jira, we are adding O(1) time complexity for every single row returned so for N rows, so this can turn out to be quite expensive, you are correct @dbwong.

@Divneet18 no worries, let's think about this and see whether we want to go forward with this change because the primary responsibility for returning the right results should stay within hbase scan and filter boundaries. Even though we check scan boundaries, what about filters? We can't introduce check for every single filter. This is going to be quite expensive.

virajjasani:
FYI, we are resuming this effort on https://github.com/apache/phoenix/pull/1724
The latest revision has all tests passing

virajjasani:
> FYI, we are resuming this effort on #1724 The latest revision has all tests passing

sorry for the confusion, i meant to create that PR only to check how many cases need fixes at server side, so that PR was just meant to be the initial phase for correctly identifying problems, we can continue more work in this current PR.

PR #1699, opened 2023-10-02
PHOENIX-7057 Potential bug in MetadataEndpointImpl#updateIndexState.

stoty:


shahrs87 (approved):
@stoty  Do you think we can add a test case for this fix? Already approved the PR if you think it is ok not to have test case.

stoty:
Thanks for the review @shahrs87 .
As this code is only used for the old non globally consistent index implementation, and the bug comes from a copy-paste error in my refactor, I wouldn't spend a lot of time on writing a test for this.

PR #1702, opened 2023-10-05
PHOENIX-7056 Update the LAST_DDL_TIMESTAMP of the parent table/view when we add/drop an index.

shahrs87:


shahrs87:
@tkhurana @palashc  Can you please help review the patch?

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/MetaDataEndpointImpl.java
```diff
@@ -2417,6 +2417,15 @@ public void createTable(RpcController controller, CreateTableRequest request,
                 // table/index/views.
                 tableMetadata.add(MetaDataUtil.getLastDDLTimestampUpdate(tableKey,
                     clientTimeStamp, EnvironmentEdgeManager.currentTimeMillis()));
+                if (tableType == INDEX) {
```
palashc, line 2420: @shahrs87 We also need to do this when dropping an index, right?

shahrs87: Yeah, I realized that and withdrew the review ask. :P

shahrs87:
@palashc  The patch is ready to review now.

phoenix-core/src/test/java/org/apache/phoenix/cache/ServerMetadataCacheTest.java
```diff
@@ -420,6 +423,126 @@ public void testInvalidateCacheForBaseTableWithUpdateIndexStatement() throws Exc
...
+            // Verify that last ddl timestamp after index deletion is greater than
+            // the previous last ddl timestamp.
+            assertNotNull(tableLastDDLTimestampAfterIndexDeletion);
+            assertTrue(tableLastDDLTimestampAfterIndexDeletion >
+                    tableLastDDLTimestampAfterIndexCreation);
+        }
+    }
+
+    /**
+     *  Test that we invalidate the cache of the immediate parent view
```
palashc, line 480: We are also testing the drop-index case so the comment can be updated.

phoenix-core/src/test/java/org/apache/phoenix/cache/ServerMetadataCacheTest.java
```diff
@@ -420,6 +423,126 @@ public void testInvalidateCacheForBaseTableWithUpdateIndexStatement() throws Exc
...
+        String createIndexDDL = "create index  " + indexName + " on " + tableName + " (v1)";
+        String dropIndexDDL = "DROP INDEX " + indexName + " ON " + tableName;
+        try (Connection conn = DriverManager.getConnection(getUrl());
+             Statement stmt = conn.createStatement()) {
+            conn.setAutoCommit(true);
+            stmt.execute(ddl);
+            long tableLastDDLTimestampBeforeIndexCreation = getLastDDLTimestamp(tableName);
+            // Populate the cache
+            assertNotNull(cache.getLastDDLTimestampForTable(null, null, tableNameBytes));
+            Thread.sleep(1);
```
palashc, line 452: For my understanding, why is sleep required in these tests?

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/MetaDataEndpointImpl.java
```diff
@@ -3507,23 +3525,30 @@ private void invalidateServerMetadataCacheWithRetries(Admin admin,
                     // DDL operations. We also need to think of we need separate RPC handler
                     // threads for this?
                     ServerRpcController controller = new ServerRpcController();
+                    for (InvalidateServerMetadataCacheRequest invalidateCacheRequest:
```
palashc, line 3528: We are already iterating over the list of `InvalidateServerMetadataCacheRequests` in `getRequest` to create the proto request. Can we do the logging there? The list is very small so maybe it does not matter?

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/InvalidateServerMetadataCacheRequest.java
```diff
@@ -0,0 +1,42 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.phoenix.coprocessor;
+
+public class InvalidateServerMetadataCacheRequest {
```
palashc, line 20: Maybe a `toString` method could be useful when we need to log the contents of this request?

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/InvalidateServerMetadataCacheRequest.java
```diff
@@ -0,0 +1,42 @@
...
+    private final byte[] tenantId;
+    private final byte[] schema;
+    private final byte[] name;
+
+    public InvalidateServerMetadataCacheRequest(byte[] tenantId, byte[] schema, byte[] name) {
+        this.tenantId = tenantId;
+        this.schema = schema;
+        this.name = name;
```
palashc, line 28: nit: Should these be `schemaName` and  `tableName` ?

palashc (commented):
@shahrs87 Few comments, but lgtm otherwise.

phoenix-core/src/test/java/org/apache/phoenix/cache/ServerMetadataCacheTest.java
shahrs87: We are generating the last ddl timestamp using `EnvironmentEdgeManager.currentTimeMillis()` when we create/drop an index. I just want to be very sure that we don't create the same time when we created an index as well as when we dropped an index. So I sleep for 1 milliseconds in between.

palashc: Got it. Will be nice to add a comment mentioning this.

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/MetaDataEndpointImpl.java
shahrs87: > The list is very small so maybe it does not matter?

Thats what I was thinking too. Atmost there will be 2 entries.

palashc: It might still get rid of some code and bytes to string conversion.

shahrs87: But then it won't have the region server name. Having a regionserver name would be helpful.

palashc: Yeah that makes sense. In that case, do you think we can override `toString` in `InvalidateServerMetadataCacheRequest` to make it cleaner here?

shahrs87: Yes, in the process of making that change.

palashc (approved):
+1, thank you

shahrs87:
Will wait for jenkins build before merging.

PR #1706, opened 2023-10-12
PHOENIX-7063 Track and account garbage collected phoenix connections

lokiore:
FYI @jpisaac

### DIFF
--- a/phoenix-core/src/it/java/org/apache/phoenix/end2end/ConnectionQueryServicesTestImpl.java
+++ b/phoenix-core/src/it/java/org/apache/phoenix/end2end/ConnectionQueryServicesTestImpl.java
@@ -69,14 +69,12 @@ public class ConnectionQueryServicesTestImpl extends ConnectionQueryServicesImpl
     
     @Override
     public void addConnection(PhoenixConnection connection) throws SQLException {
-        super.addConnection(connection);
         connections.add(connection);
         super.addConnection(connection);
     }
     
     @Override
     public void removeConnection(PhoenixConnection connection) throws SQLException {
-        super.removeConnection(connection);
         connections.remove(connection);
         super.removeConnection(connection);
     }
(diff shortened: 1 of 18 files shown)


phoenix-core/src/it/java/org/apache/phoenix/end2end/ConnectionQueryServicesTestImpl.java
```diff
@@ -69,12 +69,14 @@ public ConnectionQueryServicesTestImpl(QueryServices services, ConnectionInfo in
     
     @Override
     public void addConnection(PhoenixConnection connection) throws SQLException {
+        super.addConnection(connection);
```
jpisaac, line 72: Was this added intentionally? Do see a similar statement below?

phoenix-core/src/it/java/org/apache/phoenix/end2end/ConnectionQueryServicesTestImpl.java
```diff
@@ -69,12 +69,14 @@ public ConnectionQueryServicesTestImpl(QueryServices services, ConnectionInfo in
...
         connections.add(connection);
         super.addConnection(connection);
     }
     
     @Override
     public void removeConnection(PhoenixConnection connection) throws SQLException {
+        super.removeConnection(connection);
```
jpisaac, line 79: Same here

jpisaac (commented):
LGTM, except for some additional extraneous stmts

phoenix-core/src/it/java/org/apache/phoenix/end2end/ConnectionQueryServicesTestImpl.java
lokiore: Yes, We are testing connection counts for different scenarios in LoggingConnectionLimiterIT, and these counts gets updated when we call acquire connection in CQSI's addConnection, which was not being used before as part of CQSTestImpl.

jpisaac: But there is already a super.addConnection(..) below?

lokiore: Aaah I see....... ✌️

jpisaac (approved):
LGTM +1

PR #1708, opened 2023-10-13
PHOENIX-7044 : Support Index on a generated column that extracts a scalar value from JSON column

ranganathg:
[PHOENIX-7044](https://issues.apache.org/jira/browse/PHOENIX-7044)

### DIFF
--- a/phoenix-core/src/it/java/org/apache/phoenix/end2end/json/JsonFunctionsIT.java
+++ b/phoenix-core/src/it/java/org/apache/phoenix/end2end/json/JsonFunctionsIT.java
@@ -23,8 +23,11 @@ import com.jayway.jsonpath.Configuration;
 import com.jayway.jsonpath.JsonPath;
 import com.jayway.jsonpath.spi.json.GsonJsonProvider;
 import org.apache.hadoop.hbase.TableName;
+import org.apache.phoenix.end2end.IndexToolIT;
 import org.apache.phoenix.end2end.ParallelStatsDisabledIT;
+import org.apache.phoenix.exception.SQLExceptionCode;
 import org.apache.phoenix.util.PropertiesUtil;
+import org.apache.phoenix.util.QueryUtil;
 import org.apache.phoenix.util.TestUtil;
 import org.junit.Test;
 
@@ -32,6 +35,7 @@ import java.sql.Connection;
 import java.sql.DriverManager;
 import java.sql.PreparedStatement;
 import java.sql.ResultSet;
+import java.sql.SQLException;
 import java.util.Properties;
 
 import static org.apache.phoenix.util.TestUtil.TEST_PROPERTIES;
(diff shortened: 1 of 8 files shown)


phoenix-core/src/it/java/org/apache/phoenix/end2end/json/JsonFunctionsIT.java
```diff
@@ -633,4 +637,34 @@ public void testJsonQuery() throws Exception {
...
+            String
+                    ddl =
+                    "create table if not exists " + tableName +
+                            " (pk integer primary key, col integer, jsoncol.jsoncol json)";
+            conn.createStatement().execute(ddl);
+            conn.createStatement().execute(
+                    "UPSERT INTO " + tableName + " (pk, col, jsoncol) VALUES (1,2, '" + JsonDoc1 + "')");
+            conn.createStatement().execute(
+                    "CREATE INDEX " + indexName + " ON " + tableName
+                            + " (JSON_VALUE(JSONCOL,'$.type'), JSON_VALUE(JSONCOL,'$.info.address.town')) include (col)");
```
tkhurana, line 657: Is it possible to have a JSON function in included columns for an index ? Can we add a test for the same

phoenix-core/src/main/java/org/apache/phoenix/schema/MetaDataClient.java
```diff
@@ -1529,6 +1526,11 @@ public MutationState createIndex(CreateIndexStatement statement, byte[][] splits
                 if (expressionIndexCompiler.isAggregate()) {
                     throw new SQLExceptionInfo.Builder(SQLExceptionCode.AGGREGATE_EXPRESSION_NOT_ALLOWED_IN_INDEX).build().buildException();
                 }
+                if (expressionIndexCompiler.isJsonFragment()) {
```
tkhurana, line 1529: Can you add a test case for this too

phoenix-core/src/it/java/org/apache/phoenix/end2end/json/JsonFunctionsIT.java
ranganathg: I remember discussing this with @kadirozde . I don't think we support functions in include columns currently. Instead we can use uncovered index to achieve what we want with functions in include columns.

phoenix-core/src/it/java/org/apache/phoenix/end2end/json/JsonFunctionsIT.java
```diff
@@ -633,4 +637,34 @@ public void testJsonQuery() throws Exception {
...
+            String
+                    selectSql =
+                    "SELECT JSON_VALUE(JSONCOL,'$.type'), " +
+                    "JSON_VALUE(JSONCOL,'$.info.address.town') FROM " + tableName +
+                    " WHERE JSON_VALUE(JSONCOL,'$.type') = 'Basic'";
+            ResultSet rs = conn.createStatement().executeQuery("EXPLAIN " + selectSql);
+            String actualExplainPlan = QueryUtil.getExplainPlan(rs);
+            IndexToolIT.assertExplainPlan(false, actualExplainPlan, tableName, indexName);
+        } catch (SQLException e) {
+            assertEquals(SQLExceptionCode.DIVIDE_BY_ZERO.getErrorCode(), e.getErrorCode());
```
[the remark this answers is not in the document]
ranganathg: Fixed this.

phoenix-core/src/it/java/org/apache/phoenix/end2end/json/JsonFunctionsIT.java
ranganathg: Tested it this way:
`0: jdbc:phoenix:localhost> create index indexmytable1 on mytable1 (colA) include (PHOENIX_ROW_TIMESTAMP());
Error: ERROR 601 (42P00): Syntax error. Encountered "(" at line 1, column 77. (state=42P00,code=601)`

`0: jdbc:phoenix:localhost> create index idx_jsontest6 on jsontest6 (JSON_VALUE(JSONCOL,'$.type'), JSON_VALUE(JSONCOL,'$.info.address.town')) include (JSON_VALUE(JSONCOL, '$.info.address.county'));

Error: ERROR 601 (42P00): Syntax error. Encountered "(" at line 1, column 134. (state=42P00,code=601)`

phoenix-core/src/it/java/org/apache/phoenix/end2end/json/JsonFunctionsIT.java
```diff
@@ -633,4 +637,64 @@ public void testJsonQuery() throws Exception {
...
+            String
+                    ddl =
+                    "create table if not exists " + tableName +
+                            " (pk integer primary key, col integer, jsoncol.jsoncol json)";
+            conn.createStatement().execute(ddl);
+            conn.createStatement().execute(
+                    "UPSERT INTO " + tableName + " (pk, col, jsoncol) VALUES (1,2, '" + JsonDoc1 + "')");
+            conn.createStatement().execute(
+                    "CREATE INDEX " + indexName + " ON " + tableName
+                            + " (JSON_VALUE(JSONCOL,'$.type'), JSON_VALUE(JSONCOL,'$.info.address.town')) include (col)");
```
tkhurana, line 657: What if we create an index on a json field that is not of type string ? Does the index schema correctly infer that ?

ranganathg: Return type of JSON_VALUE is Varchar (String). So all the fields are of type String only and we can use explicit CAST to cast to different types if we need. Do you see some problem here?

tkhurana: As I think more about it, it will be fine.

phoenix-core/src/it/java/org/apache/phoenix/end2end/json/JsonFunctionsIT.java
```diff
@@ -633,4 +637,64 @@ public void testJsonQuery() throws Exception {
...
+        String indexName = "IDX_" + generateUniqueName();
+        try (Connection conn = DriverManager.getConnection(getUrl(), props)) {
+            conn.setAutoCommit(true);
+            String
+                    ddl =
+                    "create table if not exists " + tableName +
+                            " (pk integer primary key, col integer, jsoncol.jsoncol json)";
+            conn.createStatement().execute(ddl);
+            conn.createStatement().execute(
+                    "UPSERT INTO " + tableName + " (pk, col, jsoncol) VALUES (1,2, '" + JsonDoc1 + "')");
```
tkhurana, line 654: I would recommend expanding this testcase to add more rows and also test delete queries

phoenix-core/src/it/java/org/apache/phoenix/end2end/json/JsonFunctionsIT.java
```diff
@@ -633,4 +637,64 @@ public void testJsonQuery() throws Exception {
...
+            conn.createStatement().execute(
+                    "CREATE INDEX " + indexName + " ON " + tableName
```
tkhurana, line 656: Another thing to test is drop the json column from the base table. That should drop the index if any scalar field of json column is part of index pk. See `DropColumnIT#testDroppingIndexedColDropsIndex`

tkhurana (approved):

PR #1711, opened 2023-10-16
PHOENIX-7026 : Validate LAST_DDL_TIMESTAMP for write requests

palashc:


palashc:
Need to rethink the approach here. Validating metadata on the write path in the `execute` phase might be too early since there can be arbitrary number of executes for different tables before client calls `commit` and we might end up creating mutations using old metadata.

PR #1713, opened 2023-10-17
PHOENIX-7074 DROP CDC Implementation

TheNamesRai:
This PR builds on top of the shallow grammar support added for DROP CDC grammar as part of https://github.com/apache/phoenix/pull/1694. This change makes the required changes to DROP CDC object and the Uncovered Index associated with it.

Jira: PHOENIX-7074


phoenix-core/src/main/java/org/apache/phoenix/schema/MetaDataClient.java
```diff
@@ -3545,6 +3545,18 @@ public MutationState dropIndex(DropIndexStatement statement) throws SQLException
...
+    public MutationState dropCDC(DropCDCStatement statement) throws SQLException {
+        String schemaName = statement.getTableName().getSchemaName();
+        String cdcTableName = statement.getCdcObjName().getName();
+        String parentTableName = statement.getTableName().getTableName();
+        // Dropping the virtual CDC Table
+        dropTable(schemaName, cdcTableName, parentTableName, PTableType.CDC, statement.ifExists(), false, false);
+
+        String indexName = CDCUtil.getCDCIndexName(statement.getCdcObjName().getName());
+        // Dropping the uncovered index associated with the CDC Table
+        return dropTable(schemaName, indexName, parentTableName, PTableType.INDEX, statement.ifExists(), false, false);
```
haridsv, line 3557: In create, we are catching SQLExceptions on index creation and translating them on to CDC so that user wouldn't be exposed to the index name that is internal to the CDC implementation. We should probably do something similar here.

haridsv (commented):
Looks good overall. There is one requirement that came up in a recent discussion that involves protecting the CDC index from getting dropped by the user via DROP INDEX. I wonder if we can include that change as part of the scope of this PR.

phoenix-core/src/main/java/org/apache/phoenix/schema/MetaDataClient.java
TheNamesRai: Updated: Handled the SQLExceptions from dropping the Index and created a CDC Wrapper Exception on it.

I think we need not worry about dropping the CDC index directly as we are creating the index with the 2 underscores in the beginning which is preventing the index to be dropped directly.
I tried doing the same but was getting parser exceptions - 

`0: jdbc:phoenix:localhost> DROP INDEX __CDC__CDCTABLE ON DATATABLE;
Error: ERROR 601 (42P00): Syntax error. Unexpected char: '_' (state=42P00,code=601)
org.apache.phoenix.exception.PhoenixParserException: ERROR 601 (42P00): Syntax error. Unexpected char: '_'
	at org.apache.phoenix.exception.PhoenixParserException.newException(PhoenixParserException.java:33)`

haridsv: > underscores in the beginning which is preventing the index to be dropped directly.

You can still drop it by putting double quotes around the index name.

phoenix-core/src/main/java/org/apache/phoenix/schema/MetaDataClient.java
```diff
@@ -3554,7 +3555,12 @@ public MutationState dropCDC(DropCDCStatement statement) throws SQLException {
 
         String indexName = CDCUtil.getCDCIndexName(statement.getCdcObjName().getName());
         // Dropping the uncovered index associated with the CDC Table
-        return dropTable(schemaName, indexName, parentTableName, PTableType.INDEX, statement.ifExists(), false, false);
+        try {
+            return dropTable(schemaName, indexName, parentTableName, PTableType.INDEX, statement.ifExists(), false, false);
+        } catch (SQLException e) {
+            throw new SQLExceptionInfo.Builder(SQLExceptionCode.fromErrorCode(e.getErrorCode()))
+                    .setTableName(statement.getCdcObjName().getName()).build().buildException();
```
haridsv, line 3564: Missing `setRootCause()`. The same applies to create CDC as well, which I will fix in the next PR.

phoenix-core/src/main/java/org/apache/phoenix/jdbc/PhoenixStatement.java
```diff
@@ -1578,6 +1580,12 @@ public ExplainPlan getExplainPlan() throws SQLException {
 
                 @Override
                 public MutationState execute() throws SQLException {
+                    if (CDCUtil.isACDCIndex(
+                            ExecutableDropIndexStatement.this.getIndexName().getName())) {
+                        throw new SQLExceptionInfo.Builder(CANNOT_DROP_CDC_INDEX).setTableName(
+                                ExecutableDropIndexStatement.this.getIndexName().getName())
```
haridsv, line 1586: Nit: `ExecutableDropIndexStatement.this.getIndexName().getName()` could be assigned to a local variable to avoid repetition.

TheNamesRai: @haridsv , updated

haridsv (approved):
lgtm

virajjasani (approved):
+1

PR #1717, opened 2023-10-18
PHOENIX-5980 : MUTATION_BATCH_FAILED_SIZE metric is incorrectly updated for failing delete mutations

Divneet18:
JIRA :- [PHOENIX-5980](https://issues.apache.org/jira/browse/PHOENIX-5980)

### DIFF
--- a/phoenix-core/src/it/java/org/apache/phoenix/end2end/MutationBatchFailedStateMetricWithAllDeleteIT.java
+++ b/phoenix-core/src/it/java/org/apache/phoenix/end2end/MutationBatchFailedStateMetricWithAllDeleteIT.java
@@ -27,6 +27,7 @@ import java.sql.Connection;
 import java.sql.DriverManager;
 import java.sql.SQLException;
 import java.sql.Statement;
+import java.sql.ResultSet;
 import java.util.Map;
 import java.util.Properties;
 
@@ -59,6 +60,7 @@ public class MutationBatchFailedStateMetricWithAllDeleteIT extends BaseMutationB
      */
     @Test
     public void testFailedDelete() throws Exception {
+        int numDeleteCount = 0;
         Properties props = PropertiesUtil.deepCopy(TEST_PROPERTIES);
         Connection conn = null;
         try {
@@ -69,7 +71,12 @@ public class MutationBatchFailedStateMetricWithAllDeleteIT extends BaseMutationB
             // Note :- it is called before applying mutation to region
             TestUtil.addCoprocessor(conn, deleteTableName,
                     ImmutableIndexIT.DeleteFailingRegionObserver.class);
+
             // trying to delete 4 rows with this single delete statement
+            ResultSet rs = stmt.executeQuery("SELECT COUNT(*) FROM "+ deleteTableName + " where val1 > 1");
+            rs.next();
+            numDeleteCount = rs.getInt(1);
+
             String dml = String.format("DELETE FROM %s where val1 >  1", deleteTableName);
             stmt.execute(dml);
             conn.commit();
@@ -78,24 +85,24 @@ public class MutationBatchFailedStateMetricWithAllDeleteIT extends BaseMutationB
         } catch (SQLException e) {
             Map<String, Map<MetricType, Long>> mutationMetrics =
                     conn.unwrap(PhoenixConnection.class).getMutationMetrics();
-            int mfs =
-                    mutationMetrics.get(deleteTableName).get(MUTATION_BATCH_FAILED_SIZE).intValue();
             int dbfs =
                     mutationMetrics.get(deleteTableName).get(DELETE_BATCH_FAILED_SIZE).intValue();
             int upfs =
                     mutationMetrics.get(deleteTableName).get(UPSERT_BATCH_FAILED_SIZE).intValue();
+            int mfs =
+                    mutationMetrics.get(deleteTableName).get(MUTATION_BATCH_FAILED_SIZE).intValue();
             long gfs =
                     GLOBAL_MUTATION_BATCH_FAILED_COUNT.getMetric().getValue();
 
-            Assert.assertEquals(4, dbfs);
-            Assert.assertEquals(4, mfs);
+            Assert.assertEquals(numDeleteCount, dbfs);
             Assert.assertEquals(0, upfs);
+            Assert.assertEquals(numDeleteCount, mfs);
             //for the second parameter the global mutation batch failed count is double the original parameter because
             //global metrics do not reset to 0 for each parameter
             if(this.transactionProvider == null){
-                Assert.assertEquals(8, gfs);
+                Assert.assertEquals(2*numDeleteCount, gfs);
             } else if(this.transactionProvider.equals("OMID")){
-                Assert.assertEquals(4, gfs);
+                Assert.assertEquals(numDeleteCount, gfs);
             }
         }
     }
--- a/phoenix-core/src/it/java/org/apache/phoenix/end2end/MutationBatchFailedStateMetricWithAllDeleteIT.java
+++ b/phoenix-core/src/it/java/org/apache/phoenix/end2end/MutationBatchFailedStateMetricWithAllDeleteIT.java
@@ -1,0 +1,175 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.phoenix.end2end;
+
+import static org.apache.phoenix.monitoring.MetricType.DELETE_BATCH_FAILED_SIZE;
+import static org.apache.phoenix.monitoring.MetricType.UPSERT_BATCH_FAILED_SIZE;
+import static org.apache.phoenix.monitoring.MetricType.MUTATION_BATCH_FAILED_SIZE;
+import static org.apache.phoenix.util.TestUtil.TEST_PROPERTIES;
+import static org.apache.phoenix.monitoring.GlobalClientMetrics.GLOBAL_MUTATION_BATCH_FAILED_COUNT;
+
+
+import java.sql.Connection;
+import java.sql.DriverManager;
+import java.sql.PreparedStatement;
+import java.sql.SQLException;
+import java.sql.Statement;
+import java.util.Arrays;
+import java.util.Collection;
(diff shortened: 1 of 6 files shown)


phoenix-core/src/main/java/org/apache/phoenix/execute/MutationState.java
```diff
@@ -1584,12 +1588,27 @@ public static MutationMetricQueue.MutationMetric updateMutationBatchFailureMetri
...
+
+        long totalFailedMutation = numUpsertMutationsInBatch + numDeleteMutationsInBatch;
+        //TODO: this case is something that should not happen
+        //but still if condition makes sese if this ever happens
+        if (totalFailedMutation < numFailedMutations) {
+            LOGGER.warn(
+                    "total failed mutation less than num of failed mutation.  This is not expected.");
+            totalFailedMutation = numFailedMutations;
+        }
+        System.out.println(GLOBAL_MUTATION_BATCH_FAILED_COUNT.getMetric().getValue());
```
virajjasani, line 1600: looks like this was for your debugging purpose, feel free to remove :)

phoenix-core/src/main/java/org/apache/phoenix/execute/MutationState.java
```diff
@@ -1584,12 +1588,27 @@ public static MutationMetricQueue.MutationMetric updateMutationBatchFailureMetri
...
+
+        GLOBAL_MUTATION_BATCH_FAILED_COUNT.update(
+                allDeletesMutations && !isTransactional ? numDeleteMutationsInBatch :
+                        totalFailedMutation);
```
virajjasani, line 1604: we can extract the logic and keep value in one variable so the same variable can be used for both `GLOBAL_MUTATION_BATCH_FAILED_COUNT.update` and `MutationMetricQueue.MutationMetric`

virajjasani (commented):
Thanks @Divneet18, just did an initial high level review

phoenix-core/src/it/java/org/apache/phoenix/end2end/MutationBatchFailedStateMetricWithAllDeleteIT.java
```diff
@@ -0,0 +1,175 @@
...
+            // the mutations
+            // it is called before applying mutation to region
+            TestUtil.addCoprocessor(conn, deleteTableName,
+                    ImmutableIndexIT.DeleteFailingRegionObserver.class);
+            // trying to delete 4 rows with this single delete statement
+            String dml = String.format("DELETE FROM %s where val1 >  1", deleteTableName);
+            stmt.execute(dml);
+            conn.commit();
+
+            throw new AssertionError("Commit should not have succeeded");
```
gjacoby126, line 122: nit: You can use Assert.fail() here.

phoenix-core/src/it/java/org/apache/phoenix/end2end/MutationBatchFailedStateMetricWithAllDeleteIT.java
```diff
@@ -0,0 +1,175 @@
...
+                    mutationMetrics.get(deleteTableName).get(DELETE_BATCH_FAILED_SIZE).intValue();
+            int upfs =
+                    mutationMetrics.get(deleteTableName).get(UPSERT_BATCH_FAILED_SIZE).intValue();
+            long gfs =
+                    GLOBAL_MUTATION_BATCH_FAILED_COUNT.getMetric().getValue();
+
+            Assert.assertEquals(4, dbfs);
+            Assert.assertEquals(4, mfs);
+            Assert.assertEquals(0, upfs);
+            //for the second parameter the global mutation batch failed count is double the original parameter because
```
gjacoby126, line 138: Rather than assuming an execution order, might be cleaner just to keep track of which parameter (1 or 2) we're on and multiply it by 4.

phoenix-core/src/it/java/org/apache/phoenix/end2end/MutationBatchFailedStateMetricWithAllUpsertIT.java
```diff
@@ -0,0 +1,180 @@
...
+        this.transactional = transactionProvider != null;
+        if (this.transactional) {
+            create_table =
+                    create_table + (this.transactional
+                            ? (" TRANSACTIONAL=true,TRANSACTION_PROVIDER='" + transactionProvider
+                            + "'")
+                            : "");
+        }
+        this.transactionProvider = transactionProvider;
+        System.out.println("transaction provider " + this.transactionProvider);
```
gjacoby126, line 76: please remove System.out.println calls

phoenix-core/src/it/java/org/apache/phoenix/end2end/MutationBatchFailedStateMetricWithAllDeleteIT.java
```diff
@@ -0,0 +1,175 @@
...
+        this.transactional = transactionProvider != null;
+        if (this.transactional) {
+            create_table =
+                    create_table + (this.transactional
+                            ? (" TRANSACTIONAL=true,TRANSACTION_PROVIDER='" + transactionProvider
+                            + "'")
+                            : "");
+        }
+        this.transactionProvider = transactionProvider;
+        System.out.println("transaction provider " + this.transactionProvider);
```
gjacoby126, line 75: Please remove system.out.println calls

phoenix-core/src/it/java/org/apache/phoenix/end2end/MutationBatchFailedStateMetricWithAllUpsertIT.java
```diff
@@ -0,0 +1,180 @@
...
+            String upsertSQL = String.format(upsertStatement, deleteTableName);
+            PreparedStatement preparedStatement = conn.prepareStatement(upsertSQL);
+            preparedStatement.setString(1, "ROW_6");
+            preparedStatement.setInt(2, 6);
+            preparedStatement.setInt(3, 12);
+            preparedStatement.execute();
+
+            conn.commit();
+
+            throw new AssertionError("Commit should not have succeeded");
```
gjacoby126, line 128: Assert.fail()

phoenix-core/src/it/java/org/apache/phoenix/end2end/MutationBatchFailedStateMetricWithAllUpsertIT.java
```diff
@@ -0,0 +1,180 @@
...
+@RunWith(Parameterized.class)
+public class MutationBatchFailedStateMetricWithAllUpsertIT extends ParallelStatsDisabledIT {
+    String create_table =
+            "CREATE TABLE IF NOT EXISTS %s(ID VARCHAR NOT NULL PRIMARY KEY, VAL1 INTEGER, VAL2 INTEGER)";
+    String indexName = generateUniqueName();
+    String create_index = "CREATE INDEX " + indexName + " ON %s(VAL1 DESC) INCLUDE (VAL2)";
+    String upsertStatement = "UPSERT INTO %s VALUES(?, ?, ?)";
+    String deleteTableName = generateUniqueName();
+    private final boolean transactional;
+    private String transactionProvider;
```
gjacoby126, line 64: There's a lot of common boilerplate between the three new ITs. How difficult would it be to consolidate to one, or have a common superclass?

phoenix-core/src/it/java/org/apache/phoenix/end2end/MutationBatchFailedStateMetricWithDeleteAndUpsertIT.java
```diff
@@ -0,0 +1,181 @@
...
+            String upsertSQL = String.format(upsertStatement, deleteTableName);
+            PreparedStatement preparedStatement = conn.prepareStatement(upsertSQL);
+            preparedStatement.setString(1, "ROW_6");
+            preparedStatement.setInt(2, 6);
+            preparedStatement.setInt(3, 12);
+            preparedStatement.execute();
+
+            conn.commit();
+
+            throw new AssertionError("Commit should not have succeeded");
```
gjacoby126, line 126: Assert.fail()

phoenix-core/src/it/java/org/apache/phoenix/end2end/MutationBatchFailedStateMetricWithDeleteAndUpsertIT.java
```diff
@@ -0,0 +1,181 @@
...
+            long gfs =
+                    GLOBAL_MUTATION_BATCH_FAILED_COUNT.getMetric().getValue();
+
+
+            Assert.assertEquals(4, dbfs);
+            Assert.assertEquals(5, mfs);
+            Assert.assertEquals(1, upfs);
+            //for the second parameter the global mutation batch failed count is double the original parameter because
+            //global metrics do not reset to 0 for each parameter
+            if(this.transactionProvider == null){
```
gjacoby126, line 145: Likewise we're assuming an order here when we really care about whether this is the first or second parameter.

phoenix-core/src/it/java/org/apache/phoenix/end2end/MutationBatchFailedStateMetricWithDeleteAndUpsertIT.java
```diff
@@ -0,0 +1,181 @@
...
+import org.junit.experimental.categories.Category;
+import org.junit.runner.RunWith;
+import org.junit.runners.Parameterized;
+
+@Category(NeedsOwnMiniClusterTest.class)
+@RunWith(Parameterized.class)
+public class MutationBatchFailedStateMetricWithDeleteAndUpsertIT extends ParallelStatsDisabledIT {
+    String create_table =
+            "CREATE TABLE IF NOT EXISTS %s(ID VARCHAR NOT NULL PRIMARY KEY, VAL1 INTEGER, VAL2 INTEGER)";
+    String indexName = generateUniqueName();
```
gjacoby126, line 58: What do the Delete and Upsert ITs added above show that this IT (which does both delete and upsert) doesn't?

phoenix-core/src/main/java/org/apache/phoenix/execute/MutationState.java
```diff
@@ -1584,12 +1588,26 @@ public static MutationMetricQueue.MutationMetric updateMutationBatchFailureMetri
                 numDeleteMutationsInBatch++;
             }
         }
+
+        long totalFailedMutation = numUpsertMutationsInBatch + numDeleteMutationsInBatch;
+        //TODO: this case is something that should not happen
```
gjacoby126, line 1593: Comment seems valid but it doesn't seem like a TODO.

phoenix-core/src/it/java/org/apache/phoenix/end2end/MutationBatchFailedStateMetricWithAllUpsertIT.java
Divneet18: The reason the tests are in three different classes is because we are testing for global metrics. every time a test is run and for every parameter the global metric always gets added on to the previous value rather than reseting to 0 (like mutation metrics does). Hence the actual answers for the metrics were different than expected and would change based on the way tests are run and sometimes the order too. To avoid that, we came up with making different ITs. However, I will make a superclass to avoid the recurring lines of code.

virajjasani (approved):
+1, pending build results

virajjasani:
It seems we can't backport this to 5.1 because 5.1 does not have PHOENIX-6397

virajjasani:
Thanks for the contribution @Divneet18, Thanks for the review @gjacoby126

PR #1718, opened 2023-10-19
PHOENIX-7078 java.lang.NoSuchMethodError in CreateTableIT with HBase 2.1

stoty:


stoty:
Note that the HBase 2.1 build still fails with a netty version conflict that is caused by the Omid 1.1.0 version bump.
However, this fixes the build with the older Omid version.

virajjasani (approved):
+1, pending build results

stoty:
Thanks @virajjasani .
Unfortunately, we don't have precommit with HBase 2.1 , so the CI results are not relevant.

PR #1722, opened 2023-10-20
Enable pre-split on index with compound keys

potatochip:


stoty:
Can you add a JIRA and a test case @potatochip  ?

PR #1726, opened 2023-10-26
PHOENIX-7026 : Validate LAST_DDL_TIMESTAMP for write requests

palashc:
On the write path, if `phoenix.ddl.timestamp.validation.enabled` is set:
- Pick a random live region server and call the `validateLastDDLTimestamp` RPC for all the tables/views in the `mutationsMap`
    -  Include all indexes on the table/view in case their state had changed
    -  Include all ancestors of views
-  If `StaleMetadataCacheException` is encountered, force update client side cache for all the tables/views. 
    - ~~Replace the `PTable` object in all the `TableRefs` with the latest `PTable` for every table/view.~~
        - `MutationState#validateAndGetServerTimestamp` should take care of this.

### DIFF
--- a/phoenix-core/src/main/java/org/apache/phoenix/monitoring/MetricType.java
+++ b/phoenix-core/src/main/java/org/apache/phoenix/monitoring/MetricType.java
@@ -150,7 +150,7 @@ public enum MetricType {
     CLIENT_METADATA_CACHE_MISS_COUNTER("cmcm", "Number of cache misses for the CQSI cache.", LogLevel.DEBUG, PLong.INSTANCE),
     CLIENT_METADATA_CACHE_HIT_COUNTER("cmch", "Number of cache hits for the CQSI cache.", LogLevel.DEBUG, PLong.INSTANCE),
     PAGED_ROWS_COUNTER("prc", "Number of dummy rows returned to client due to paging.", LogLevel.DEBUG, PLong.INSTANCE),
-    STALE_METADATA_CACHE_EXCEPTION_COUNTER("smcem", "Number of StaleMetadataCacheException encountered.", LogLevel.DEBUG, PLong.INSTANCE),
+    STALE_METADATA_CACHE_EXCEPTION_COUNTER("smce", "Number of StaleMetadataCacheException encountered.", LogLevel.DEBUG, PLong.INSTANCE),
 
     // hbase metrics
     COUNT_RPC_CALLS("rp", "Number of RPC calls",LogLevel.DEBUG, PLong.INSTANCE),
--- a/phoenix-core/src/main/java/org/apache/phoenix/monitoring/MetricType.java
+++ b/phoenix-core/src/main/java/org/apache/phoenix/monitoring/MetricType.java
@@ -150,6 +150,7 @@ public enum MetricType {
     CLIENT_METADATA_CACHE_MISS_COUNTER("cmcm", "Number of cache misses for the CQSI cache.", LogLevel.DEBUG, PLong.INSTANCE),
     CLIENT_METADATA_CACHE_HIT_COUNTER("cmch", "Number of cache hits for the CQSI cache.", LogLevel.DEBUG, PLong.INSTANCE),
     PAGED_ROWS_COUNTER("prc", "Number of dummy rows returned to client due to paging.", LogLevel.DEBUG, PLong.INSTANCE),
+    STALE_METADATA_CACHE_EXCEPTION_COUNTER("smcem", "Number of StaleMetadataCacheException encountered.", LogLevel.DEBUG, PLong.INSTANCE),
 
     // hbase metrics
     COUNT_RPC_CALLS("rp", "Number of RPC calls",LogLevel.DEBUG, PLong.INSTANCE),
--- a/phoenix-core/src/main/java/org/apache/phoenix/monitoring/MetricType.java
+++ b/phoenix-core/src/main/java/org/apache/phoenix/monitoring/MetricType.java
@@ -150,7 +150,9 @@ public enum MetricType {
     CLIENT_METADATA_CACHE_MISS_COUNTER("cmcm", "Number of cache misses for the CQSI cache.", LogLevel.DEBUG, PLong.INSTANCE),
     CLIENT_METADATA_CACHE_HIT_COUNTER("cmch", "Number of cache hits for the CQSI cache.", LogLevel.DEBUG, PLong.INSTANCE),
     PAGED_ROWS_COUNTER("prc", "Number of dummy rows returned to client due to paging.", LogLevel.DEBUG, PLong.INSTANCE),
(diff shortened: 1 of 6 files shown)


phoenix-core/src/test/java/org/apache/phoenix/cache/ServerMetadataCacheTest.java
```diff
@@ -911,7 +913,330 @@ public void testSelectQueryDropIndex() throws Exception {
...
+            //client-1 creates a table and views
+            createTable(conn1, tableName, NEVER);
+            createView(conn1, tableName, viewName1);
+            createView(conn1, viewName1, viewName2);
+
+            //client-2 populates its cache, 1 getTable RPC each for table, view1, view2
+            query(conn2, viewName2);
+
+            //client-1 alters first level view
+            alterViewAddColumn(conn1, viewName1, "v3");
```
shahrs87, line 982: Can we rename `v3` to `col3`? It gets confusing if it is a view name or column name.

phoenix-core/src/test/java/org/apache/phoenix/cache/ServerMetadataCacheTest.java
```diff
@@ -911,7 +913,330 @@ public void testSelectQueryDropIndex() throws Exception {
...
+        try (Connection conn1 = spyCqs1.connect(url1, props);
+             Connection conn2 = spyCqs2.connect(url2, props)) {
+
+            //client-1 creates a table and views
+            createTable(conn1, tableName, NEVER);
+            createView(conn1, tableName, viewName1);
+            createView(conn1, viewName1, viewName2);
+
+            //client-2 populates its cache, 1 getTable RPC each for table, view1, view2
+            query(conn2, viewName2);
```
shahrs87, line 979: Can we verify that getTable call was made for tableName, viewName1 and viewName2?

phoenix-core/src/test/java/org/apache/phoenix/cache/ServerMetadataCacheTest.java
```diff
@@ -911,7 +913,330 @@ public void testSelectQueryDropIndex() throws Exception {
...
+
+            // client-2 drops the table
+            conn2.createStatement().execute("DROP TABLE " + tableName);
+
+            //client-1 commits
+            conn1.commit();
+            Assert.fail("Commit should have failed with TableNotFoundException");
+        }
+        catch (Exception e) {
+            Assert.assertTrue("TableNotFoundException was not thrown when table was dropped concurrently with upserts.", e instanceof TableNotFoundException);
```
shahrs87, line 1043: nit: line length warning.

phoenix-core/src/test/java/org/apache/phoenix/cache/ServerMetadataCacheTest.java
```diff
@@ -911,7 +913,330 @@ public void testSelectQueryDropIndex() throws Exception {
...
+                    any(byte[].class), eq(PVarchar.INSTANCE.toBytes(tableName2)),
+                    anyLong(), anyLong());
+        }
+    }
+
+    /**
+     * Test upserts into a multi-level view hierarchy.
+     */
+    @Test
+    public void testUpsertViewWithOldDDLTimestamp() throws Exception {
```
shahrs87, line 960: Can we add a test case where we add/drop view in the view hierarchy?

shahrs87 (changes requested):
Overall the PR looks v good. Few comments, most of them are in test related code. Thank you for the nice work @palashc

phoenix-core/src/test/java/org/apache/phoenix/cache/ServerMetadataCacheTest.java
palashc: @shahrs87 Can you please share the scenario you have in mind here?

palashc: I added one where a client drops a parent view concurrently with another client upserting into its child view.

phoenix-core/src/main/java/org/apache/phoenix/monitoring/MetricType.java
```diff
@@ -150,6 +150,7 @@ public enum MetricType {
     CLIENT_METADATA_CACHE_MISS_COUNTER("cmcm", "Number of cache misses for the CQSI cache.", LogLevel.DEBUG, PLong.INSTANCE),
     CLIENT_METADATA_CACHE_HIT_COUNTER("cmch", "Number of cache hits for the CQSI cache.", LogLevel.DEBUG, PLong.INSTANCE),
     PAGED_ROWS_COUNTER("prc", "Number of dummy rows returned to client due to paging.", LogLevel.DEBUG, PLong.INSTANCE),
+    STALE_METADATA_CACHE_EXCEPTION_COUNTER("smcem", "Number of StaleMetadataCacheException encountered.", LogLevel.DEBUG, PLong.INSTANCE),
```
shahrs87, line 153: Why `smcem` ?  What does the last m indicate? (metric?)  If yes, then we don't have in other metrics.

shahrs87 (changes requested):
Just couple of minor comments and then we are good to go.

phoenix-core/src/main/java/org/apache/phoenix/monitoring/MetricType.java
palashc: I think I wanted to add a `c` for counter and added `m` instead. I will remove it.

shahrs87:
@palashc  There are 15 test failures, all in 2 classes: 
`org.apache.phoenix.end2end.index.MutableIndexFailureIT` and `org.apache.phoenix.end2end.index.MutableIndexFailureWithNamespaceIT` 
Looks like these are related to your changes.
Also there are 15 new checkstyle warnings. Can you please fix that too? Thank you !

palashc:
@shahrs87 Those tests pass locally. Maybe some issue with the test environment caused them to fail. We have not turned on the feature flag for other tests yet so ideally they should not fail here. Most of the checkstyle warnings for non-test code introduced in this PR are for line length being more than the limit only by a couple of characters.

shahrs87:
> Most of the checkstyle warnings for non-test code introduced in this PR are for line length being more than the limit only by a couple of characters.

Lets fix them. 

> Those tests pass locally. Maybe some issue with the test environment caused them to fail.

Let me trigger the build again. Lets monitor this build: https://ci-hadoop.apache.org/job/Phoenix/job/Phoenix-PreCommit-GitHub-PR/job/PR-1726/7/
@palashc

shahrs87:
@palashc  There are 10 checkstyle warnings. Are they all irrelevant?

shahrs87:
In the latest build, looks like most of the checkstyle warnings are due to this patch. @palashc Can you please fix them and I will merge then

palashc:
@shahrs87 I have fixed the checkstyle warnings coming from code. I am not sure how to fix the remaining two.

PR #1728, opened 2023-10-31
PHOENIX-7090 Hash join throw NullPointerException when subquery retur…

chaijunjie0101:
…n null

### DIFF
--- a/phoenix-core/src/main/java/org/apache/phoenix/execute/HashJoinPlan.java
+++ b/phoenix-core/src/main/java/org/apache/phoenix/execute/HashJoinPlan.java
@@ -462,7 +462,9 @@ public class HashJoinPlan extends DelegateQueryPlan {
                 }
 
                 Object result = expectSingleRow ? (values.isEmpty() ? null : values.get(0)) : PArrayDataType.instantiatePhoenixArray(baseType, values.toArray());
-                parent.getContext().setSubqueryResult(select, result);
+                if (result != null) {
+                    parent.getContext().setSubqueryResult(select, result);
+                }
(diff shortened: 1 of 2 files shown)


phoenix-core/src/main/java/org/apache/phoenix/execute/HashJoinPlan.java
```diff
@@ -462,7 +462,9 @@ public ServerCache execute(HashJoinPlan parent) throws SQLException {
                 }
 
                 Object result = expectSingleRow ? (values.isEmpty() ? null : values.get(0)) : PArrayDataType.instantiatePhoenixArray(baseType, values.toArray());
-                parent.getContext().setSubqueryResult(select, result);
+                if (result != null) {
```
stoty, line 465: Nit: Can you switch the if to return null in the then clause ?

chaijunjie0101: thanks for reviewing
for the org.apache.phoenix.execute.HashJoinPlan.WhereClauseSubPlan#execute, it always return null;
you mean change here like this?
                if (result == null) {
                    return null;
                }
                parent.getContext().setSubqueryResult(select, result);
                return null;
here i changed just judge to avoid put a null value to subqueryResults as a subquery result

stoty: Forget it, your code is fine.
We always have to return null anyway.

stoty (approved):
+1 LGTM

PR #1729, opened 2023-11-02
PHOENIX-7095 Implement Statement.closeOnCompletion() and fix related close() bugs

stoty:


stoty:
No tests yet, I just want to see if this causes any regressions.

phoenix-core/src/main/java/org/apache/phoenix/jdbc/PhoenixConnection.java
```diff
@@ -720,17 +717,17 @@ public void clearWarnings() throws SQLException {
...
             mutationState.rollback();
         } catch (SQLException e) {
             // ignore any exceptions while rolling back
         } finally {
             try {
-                SQLCloseables.closeAll(statements);
+                // create new list to prevent close of statements from modifying this collection.
+                // TODO This could be optimized out by decoupling closing the stmt and removing it
+                // from the connection.
+                List<? extends PhoenixStatement> statementsCopy = new ArrayList<>(this.statements);
```
richardantal, line 729: Is there a reason why the statements got changed to HashSet?
This copy is a list and it is not consistent this way.

phoenix-core/src/main/java/org/apache/phoenix/jdbc/PhoenixStatement.java
```diff
@@ -330,161 +333,162 @@ protected PhoenixResultSet executeQuery(final CompilableStatement stmt, final Qu
...
-                                }
-                                Iterator<TableRef> tableRefs = plan.getSourceRefs().iterator();
-                                connection.getMutationState().sendUncommitted(tableRefs);
-                                plan =
-                                        connection.getQueryServices().getOptimizer()
-                                                .optimize(PhoenixStatement.this, plan);
-                                // this will create its own trace internally, so we don't wrap this
-                                // whole thing in tracing
-                                ResultIterator resultIterator = plan.iterator();
+            return CallRunner.run(
```
richardantal, line 341: I've checked this part of code locally because the here on the website it is really hard to review reformatted code.
As I see there is only a minor change in this part that setLastResultSet is not called.
To be honest I would not do this reformatting because it changes the git log/blame, makes it hard to search.

phoenix-core/src/main/java/org/apache/phoenix/jdbc/PhoenixConnection.java
stoty: Connections may have several open statements.
Removing closed connections from an ArrayList is O(n), while remving them from a Hash is O(1).
(Though for a few open statements this doesn't matter much)

But you are right, since closeAll() eventually removes the elements from the copy, it's better to use a HashSet for the copy as well.

phoenix-core/src/main/java/org/apache/phoenix/jdbc/PhoenixStatement.java
stoty: It's the usual fight with the usless 120 character line width limit.
Since the code was already not well formatted, I may as well revert it.

stoty: Addressed both concerns.

richardantal (approved):
Thanks for addressing my concerns, it Looks good to me

PR #1736, opened 2023-11-10
PHOENIX-7106 Data Integrity issues due to invalid rowkeys returned by various coprocessors

virajjasani:
Jira: PHOENIX-7106


virajjasani:
Need to resolve merge conflict

phoenix-core-client/src/main/java/org/apache/phoenix/iterate/OffsetResultIterator.java
```diff
@@ -32,32 +34,49 @@
...
+            }
             if (tuple.size() == 0 || isDummy(tuple)) {
                 // while rowCount < offset absorb the dummy and call next on the underlying scanner
                 continue;
             }
             rowCount++;
-            // no page timeout check at this level because we cannot correctly resume
-            // scans for OFFSET queries until the offset is reached
+            lastScannedTuple = tuple;
+            if (!isIncompatibleClient) {
```
ranganathg, line 69: what doesn an incompatibleclient mean? and in which cases will you have scenario where currentTime - startTime is more than LONG_MAX? and why do you need to do that check? 
Shouldn't you return immediately with dummyTuple when its an incompatible client?

phoenix-core-client/src/main/java/org/apache/phoenix/iterate/OrderedResultIterator.java
```diff
@@ -151,6 +159,14 @@ public Expression apply(OrderByExpression column) {
     private Tuple dummyTuple = null;
     private long byteSize;
     private long pageSizeMs;
+    private Scan scan;
+    private byte[] scanStartRowKey;
+    private byte[] prevScanStartRowKey;
+    private Boolean prevScanIncludeStartRowKey;
```
ranganathg, line 165: nit: boolean? (small b?)

phoenix-core-client/src/main/java/org/apache/phoenix/iterate/OrderedResultIterator.java
```diff
@@ -301,21 +415,51 @@ private PeekingResultIterator getResultIterator() throws SQLException {
...
         
         return resultIterator;
     }
 
+    /**
+     * Retrieve dummy rowkey and return to the client.
+     *
+     * @return the result iterator.
+     */
+    private PeekingResultIterator getDummyResult() {
```
kadirozde, line 439: This should not return PeekingResultIterator. Should it be private void updateDummyResult()?

phoenix-core-client/src/main/java/org/apache/phoenix/filter/PagingFilter.java
```diff
@@ -71,30 +68,21 @@ public void setDelegateFilter (Filter delegate) {
...
     }
 
-    public byte[] getRowKeyAtStop() {
+    public byte[] getCurrentRowKeyToBeExcluded() {
         byte[] rowKeyAtStop = null;
-        if (nextHintCell != null) {
-            // if we have already seeked to the next cell use that when we resume the scan
-            rowKeyAtStop = CellUtil.cloneRow(nextHintCell);
-        } else if (currentCell != null) {
+        if (currentCell != null) {
```
kadirozde, line 73: Is it possible that currentCell is null? If so then PagingRegionScanner should check if the row key is null, and handle the null row key.

phoenix-core-client/src/main/java/org/apache/phoenix/iterate/OffsetResultIterator.java
```diff
@@ -32,32 +34,49 @@
...
+                if (EnvironmentEdgeManager.currentTimeMillis() - startTime >= pageSizeMs) {
```
kadirozde, line 70: Since we cannot handle paging before skipping offset number of rows, we should not time out here either. We should also lock the region to prevent region movement.

phoenix-core-client/src/main/java/org/apache/phoenix/util/ByteUtil.java
```diff
@@ -469,6 +474,105 @@ public static byte[] previousKey(byte[] key) {
         return previousKey;
     }
 
+    public static byte[] getLargestPossibleRowKeyInRange(byte[] startKey, byte[] endKey) {
```
kadirozde, line 477: This needs documentation comments to describe what it does. It also assumes that startKey and endKey are not null and so it will be good to check them. It looks likes the range here is [startKey, endKey), i.e., endKey is not inclusive. Let's state it in the comments.

phoenix-core-client/src/main/java/org/apache/phoenix/util/ByteUtil.java
```diff
@@ -469,6 +474,105 @@ public static byte[] previousKey(byte[] key) {
...
+                    }
+                }
+            } else if (endKey.length > 0) {
+                rowKey = ByteUtil.previousKeyWithLength(ByteUtil.concat(endKey,
+                        new byte[1]), endKey.length + 1);
+            } else {
+                rowKey = ByteUtil.nextKeyWithLength(ByteUtil.concat(startKey,
+                        new byte[1]), startKey.length + 1);
+            }
+            if (rowKey == null) {
```
kadirozde, line 533: Should we suppress the error by returning null or generate a RuntimeException?

phoenix-core-server/src/main/java/org/apache/phoenix/coprocessor/IndexerRegionScanner.java
```diff
@@ -435,12 +437,33 @@ public boolean next(List<Cell> results) throws IOException {
...
+            byte[] startKey = scan.getStartRow().length > 0 ? scan.getStartRow() :
+                    region.getRegionInfo().getStartKey();
+            byte[] endKey = scan.getStopRow().length > 0 ? scan.getStopRow() :
+                    region.getRegionInfo().getEndKey();
+            final boolean isIncompatibleClient =
+                    ScanUtil.isIncompatibleClientForServerReturnValidRowKey(scan);
+            if (!isIncompatibleClient) {
+                rowKey = ByteUtil.getLargestPossibleRowKeyInRange(startKey, endKey);
+                if (rowKey == null) {
+                    if (scan.includeStartRow()) {
```
kadirozde, line 453: If lastCell is null then there is no row left to scan. Why do we need to worry about the start row here?

phoenix-core-server/src/main/java/org/apache/phoenix/coprocessor/UncoveredIndexRegionScanner.java
```diff
@@ -404,4 +421,37 @@ public boolean next(List<Cell> result) throws IOException {
...
+        } else {
+            if (includeInitStartRowKey && initStartRowKey.length > 0) {
+                byte[] prevKey;
+                if (Bytes.compareTo(initStartRowKey, initStartRowKey.length - 1,
+                        1, Bytes.toBytesBinary("\\x00"), 0, 1) == 0) {
+                    prevKey = new byte[initStartRowKey.length - 1];
+                    System.arraycopy(initStartRowKey, 0, prevKey, 0, prevKey.length);
+                } else {
+                    prevKey = ByteUtil.previousKeyWithLength(ByteUtil.concat(initStartRowKey,
+                            new byte[10]), initStartRowKey.length + 10);
```
kadirozde, line 448: Padding 10 zero bytes needs some explanation here.

phoenix-core-server/src/main/java/org/apache/phoenix/coprocessor/PagingRegionScanner.java
```diff
@@ -72,15 +96,10 @@ private boolean next(List<Cell> results, boolean raw) throws IOException {
                 // There is no more row from the HBase region scanner. We need to check if PageFilter
                 // has stopped the region scanner
                 if (pagingFilter.isStopped()) {
-                    // Close the current region scanner, start a new one and return a dummy result
-                    delegate.close();
-                    byte[] rowKey = pagingFilter.getRowKeyAtStop();
-                    boolean isInclusive = pagingFilter.isNextRowInclusive();
-                    scan.withStartRow(rowKey, isInclusive);
-                    delegate = region.getScanner(scan);
+                    byte[] rowKey = pagingFilter.getCurrentRowKeyToBeExcluded();
```
tkhurana, line 99: move this to inside the `if (results.isEmpty())`

phoenix-core-server/src/main/java/org/apache/phoenix/coprocessor/PagingRegionScanner.java
```diff
@@ -61,6 +63,28 @@ public PagingRegionScanner(Region region, RegionScanner scanner, Scan scan) {
 
     private boolean next(List<Cell> results, boolean raw) throws IOException {
         try {
+            byte[] adjustedStartRowKey =
```
tkhurana, line 66: why do we need to do this check on every next call ? Can't we do this in the constructor ?

phoenix-core-client/src/main/java/org/apache/phoenix/iterate/OffsetResultIterator.java
virajjasani: We have region lock in `getOffsetScanner()` so this whole operation is under region lock.
However, the reason why any lower/upper level dummy is getting ignored as of now is because `getOffsetKvWithLastScannedRowKey()` returns the last scanned rowkey and discard dummy.

So in reality, it's true that we can remove this logic of timeout here, but even if we don't, `RegionScannerResultIterator` will continue `scanner#nextRaw` as dummy is discarded by `getOffsetKvWithLastScannedRowKey()`.

However, now i see that we should not timeout here because even if dummy is discarded, when the offset scanner returns the result, region lock is lost and hence before `RegionScannerResultIterator` performs next `scanner#nextRaw` call, region can potentially move. Let me revisit this.

phoenix-core-client/src/main/java/org/apache/phoenix/iterate/OffsetResultIterator.java
virajjasani: Incompatible client means that the client that does not understand new format of cells returned by this particular change and hence we expect server to return same old format of cells so that old (incompatible) client can continue to work.

> in which cases will you have scenario where currentTime - startTime is more than LONG_MAX?

LONG_MAX is only if paging size is not set. If it is set, then page timeout value will be used.

phoenix-core-client/src/main/java/org/apache/phoenix/filter/PagingFilter.java
virajjasani: currentCell would never be null because this is called only when paging filter stops. PagingFilter is stopped by:
```
        if (state == State.STARTED && currentCell != null
                && currentTime - startTime >= pageSizeMs) {
            state = State.TIME_TO_STOP;
        }
```

Hence, only if currentCell is not null, `TIME_TO_STOP` is set to stop the filter.

phoenix-core-client/src/main/java/org/apache/phoenix/iterate/OffsetResultIterator.java
tkhurana: Another issue with offset is that when the region moves the hbase client will adjust the scan start key but it has no idea about the `OFFSET` attribute so the new scan will start with the correct key but wrong offset. @virajjasani are you handling this case ?

virajjasani: Hmm that's also good point, let me check what happens to `_RowOffset`, i think scan attributes are preserved because when scanner is reset, on the same scan object, start rowkey is changed so the scan attributes should remain the same but let me spend sometime to confirm this.

phoenix-core-server/src/main/java/org/apache/phoenix/coprocessor/PagingRegionScanner.java
virajjasani: GROUP BY and ORDER BY on non-pk columns would need to reset this but they can figure out whether to reset the scanner only as part of `scanner#next`. This is why we can't move it to constructor.

phoenix-core-server/src/main/java/org/apache/phoenix/coprocessor/IndexerRegionScanner.java
virajjasani: This is because if `getLargestPossibleRowKeyInRange()` is not able to figure out previous largest possible rowkey, it returns null and most likely that happens when either start or end key are empty.

So, we try to find [startKey, endKey) and if we can't, then if startKey is included, we return scan startkey (or region start key if scan startKey is empty), if stopKey is included (very rare case), we return endKey, otherwise we return empty rowKey (none of our scan boundary changes have both exclusive start and stop keys).

virajjasani: This logic applies to the places where we just want to return count value to client such that we would not come back to the region again even if it moves. AFAIK, these cases are safe because scan operation is not continued but let me spend a bit more time to ensure this is certainly true.

virajjasani:
Let me address few comments in single commit so that we can check test results after the change. I will address documentation changes too in separate commit, Thanks for the in-progress reviews.

phoenix-core-server/src/main/java/org/apache/phoenix/coprocessor/UncoveredIndexRegionScanner.java
virajjasani: Addressed this so far, awaiting full build results.

virajjasani: I still need to document this, keeping note.

virajjasani:
https://github.com/apache/phoenix/pull/1736/commits/41144ee9f9694c6c558256c2e6094cdd5ea60962 caused some issues, working on it.

gjacoby126:
A general question: looks like we're duplicating quite a lot of IT tests so that we can test them with region moves. This will require future test-writers to either duplicate their tests going forward, or the tests will gradually diverge. This will also increase the time to run the full test suite.

Is there a way we can do region moves as a parameterized use case, and/or is there a more limited set of functionality we need to verify works with region moves that covers all the different scanner types but not necessarily all the higher level logic the existing IT tests need to verify?

gjacoby126:
That would also shrink the size of the patch significantly.

virajjasani:
@gjacoby126 i agree with your concerns and i really wish there was some level of possible way to make the tests generic, but honestly its almost impossible to make tests generic by using parameterized argument because of several reasons:

1. We need to use `hbase.client.scanner.max.result.size` before setting up `PhoenixTestDriver`
2. Set `phoenix.scanning.result.post.dummy.process` to a class that can process region moves after dummy is received by client (depending on how aggressive tests are, we need custom logic here)
3. Set `phoenix.scanning.result.post.valid.process` for some tests
4. Set `phoenix.server.page.size.ms` to 0ms to generate frequent timeouts at server side
5. Depending on tests, we also need to move regions after specific num of `rs#next` calls
6. Region move logic has to be very specific: only move regions of tables that are really going to be used by queries e.g. data table and index table for the given query. If we move all regions (including hbase and phoenix system tables), tests take 5x more time than what they do with the current setup.

In fact, initially i was able to reduce the patch size from ~19k to < 15k by adding `ParallelStatsDisabledWithRegionMovesIT`, which is now being extended by several tests.

phoenix-core-server/src/main/java/org/apache/phoenix/coprocessor/UncoveredIndexRegionScanner.java
virajjasani: Addressing this with max row length logic turned out to be much complicated that what i thought earlier. Very few tests are able to cover the scenario where hbase client has to lookup the given row in meta after meta cache expires at the client, and when they do, meta lookup fails because meta lookup scan start rowkey becomes much larger by combining table name, the dummy rowkey (max length - 1) and some constant values added by client.

i have addressed this in the latest commit: https://github.com/apache/phoenix/pull/1736/commits/fcbec0f676b08ce757036d33cf4687b797e2c781

awaiting the build result before moving forward.

phoenix-core-client/src/main/java/org/apache/phoenix/iterate/OffsetResultIterator.java
tkhurana: @virajjasani Did you take a look at this

virajjasani: yes, the scan attributes remain unaffected because hbase client keeps the same scan object and it just updates start rowkey:

```
    // Else, its signal from depths of ScannerCallable that we need to reset the scanner.
    if (this.lastResult != null) {
      // The region has moved. We need to open a brand new scanner at the new location.
      // Reset the startRow to the row we've seen last so that the new scanner starts at
      // the correct row. Otherwise we may see previously returned rows again.
      // If the lastRow is not partial, then we should start from the next row. As now we can
      // exclude the start row, the logic here is the same for both normal scan and reversed scan.
      // If lastResult is partial then include it, otherwise exclude it.
      scan.withStartRow(lastResult.getRow(), lastResult.mayHaveMoreCellsInRow());
    }
```

So, we are good from scan attributes viewpoint. I will address the remaining comments, including offset queries today.

tkhurana: But the offset should change because some rows have already been scanned so this won't work right @virajjasani ?

virajjasani: I am not able to reproduce any issue with this.

OffsetScanner will return valid tuple with cell CF:CQ as `f_offset:c_offset`, from which the cell value will be used by SerialIterators to set offset on next scan:

```
            while (index < scans.size()) {
                Scan currentScan = scans.get(index++);
                if (remainingOffset != null) {
// <======= set offset value
                    currentScan.setAttribute(BaseScannerRegionObserverConstants.SCAN_OFFSET, PInteger.INSTANCE.toBytes(remainingOffset));
                }
                ScanMetricsHolder scanMetricsHolder =
                        ScanMetricsHolder.getInstance(readMetrics, tableName, currentScan,
                            context.getConnection().getLogLevel());
                TableResultIterator itr =
                        new TableResultIterator(mutationState, currentScan, scanMetricsHolder,
                                renewLeaseThreshold, plan, scanGrouper, caches, maxQueryEndTime);
                PeekingResultIterator peekingItr = iteratorFactory.newIterator(context, itr, currentScan, tableName, plan);
                Tuple tuple;
                if ((tuple = peekingItr.peek()) == null) {
                    peekingItr.close();
                    continue;
// <======== retrieve offset value from server
                } else if ((remainingOffset = QueryUtil.getRemainingOffset(tuple)) != null) {
                    peekingItr.next();
                    peekingItr.close();
                    continue;
                }
                context.getConnection().addIteratorForLeaseRenewal(itr);
                return peekingItr;
            }
```

Hence, depending on the cell returned by server, next scan's offset attribute will be set and accordingly new scanner will take that value.

Basically, i will have to remove page timeout logic from offset because of it's tricky to manage returning dummy, and if we do that, we will anyways ensure that server is able to return valid offset value to client and client will check whether more serial iteration needs to happen depending on how many rows were scanned.

The logic remains same even if region moves after returning valid row. The problem happens only when region moves after dummy is returned, so we will remove page timeout for offset scanner.

tkhurana: That is why we had the original comment 
`// no page timeout check at this level because we cannot correctly resume
          scans for OFFSET queries until the offset is reached`
 I was under the impression that you were fixing that issue too

virajjasani: I fixed handling of dummy for offset, the tests also fail if we don't return right rowkey for offset dummy.
For every dummy cell that we send from server, we need to send rowkey as scan start rowkey (inclusive or exclusive depending on whether it was set as inclusive).

virajjasani:
Offset is broken, without paging. Offset maintains row count at server side in memory and increments it until offset is reached, so even if we restrict dummy processing until offset is reached, it can still cause issues with valid row.
e.g. rows are scanned until offset is reached and then we return valid row, and then region moves. Now when scan is resumed on moved region, row count state is lost and it is reset to 0 to begin with. This can easily lead to data integrity issues for offset because the expected behavior in this case is to resume scanning rows considering row count is already at offset.

I am working on fixing this.

phoenix-core-client/src/main/java/org/apache/phoenix/iterate/OffsetResultIterator.java
virajjasani: Fixing page timeout for offset https://github.com/apache/phoenix/pull/1736/commits/5677efde8fc6d6046d5a8fd7a81d4e0c0ade2b45

phoenix-core-client/src/main/java/org/apache/phoenix/util/ByteUtil.java
virajjasani: This is not required because callers are making enough attempt to ride over null value returned here. e.g. if this returns null, either start or end key is returned as rowkey (depending on which one is inclusive in the scan range).

phoenix-core-client/src/main/java/org/apache/phoenix/iterate/OrderedResultIterator.java
virajjasani: We need Boolean here because only server side execution of this iterator sets this value so we have check in place to see if the value is not null.
e.g.
```
            if (firstScan && serverSideIterator && prevScanStartRowKey != null &&
                    prevScanIncludeStartRowKey != null) {
```

phoenix-core-client/src/main/java/org/apache/phoenix/util/ByteUtil.java
virajjasani: Done

phoenix-core-server/src/main/java/org/apache/phoenix/coprocessor/PagingRegionScanner.java
virajjasani: Done

virajjasani:
@kadirozde @tkhurana @gjacoby126 @ranganathg

Fixed offset issues with valid row return with region movement causing data integrity issues, and reintroduced page timeout for offset.

I have also addressed all comments so far.

virajjasani:
Added backward compatibility tests for unordered + ordered group by queries. Need to add more backward compatibility tests.

virajjasani:
https://ci-hadoop.apache.org/job/Phoenix/job/Phoenix-PreCommit-GitHub-PR/job/PR-1736/

phoenix-core-server/src/main/java/org/apache/phoenix/coprocessor/GroupedAggregateRegionObserver.java
```diff
@@ -230,10 +238,17 @@ private static final class InMemoryGroupByCache implements GroupByCache {
         private final ServerAggregators aggregators;
         private final RegionCoprocessorEnvironment env;
         private final byte[] customAnnotations;
+        private final ConcurrentMap<ImmutableBytesWritable, ImmutableBytesWritable>
```
tkhurana, line 241: Any reason why you are using `ImmutableBytesWritable` vs aggregateMap is using `ImmutableBytesPtr` ?

phoenix-core-server/src/main/java/org/apache/phoenix/coprocessor/GroupedAggregateRegionObserver.java
```diff
@@ -243,6 +258,7 @@ private static final class InMemoryGroupByCache implements GroupByCache {
             this.aggregateMap = Maps.newHashMapWithExpectedSize(estDistVals);
             this.chunk = tenantCache.getMemoryManager().allocate(estSize);
             this.customAnnotations = customAnnotations;
+            aggregateValueToLastScannedRowKeys = new ConcurrentHashMap<>();
```
tkhurana, line 261: Maybe use `Maps.new....` API here to remain consistent with aggregateMap

phoenix-core-server/src/main/java/org/apache/phoenix/coprocessor/GroupedAggregateRegionObserver.java
```diff
@@ -371,12 +425,32 @@ private static class UnorderedGroupByRegionScanner extends BaseRegionScanner {
...
+            this.scan = scan;
+            // If scan start rowkey is empty, use region boundaries. Reverse region boundaries
+            // for reverse scan.
+            scanStartRowKey = scan.getStartRow().length > 0 ? scan.getStartRow() :
+                    (scan.isReversed() ? region.getRegionInfo().getEndKey() :
+                            region.getRegionInfo().getStartKey());
+            includeStartRowKey = scan.includeStartRow();
+            // Retrieve start rowkey of the previous scan. This would be different than
+            // current scan start rowkey if the region has recently moved or split or merged.
+            this.prevScanStartRowKey =
```
tkhurana, line 451: rename this to actualScanStartRowKey

phoenix-core-server/src/main/java/org/apache/phoenix/coprocessor/GroupedAggregateRegionObserver.java
```diff
@@ -371,12 +425,32 @@ private static class UnorderedGroupByRegionScanner extends BaseRegionScanner {
...
+                    scan.getAttribute(BaseScannerRegionObserverConstants.SCAN_ACTUAL_START_ROW);
+            this.prevScanIncludeStartRowKey = true;
```
tkhurana, line 453: rename this to actualScanIncludeStartRowKey

phoenix-core-server/src/main/java/org/apache/phoenix/coprocessor/GroupedAggregateRegionObserver.java
```diff
@@ -436,36 +622,79 @@ public boolean next(List<Cell> resultsToReturn) throws IOException {
...
             } finally {
                 if (acquiredLock) region.closeRegionOperation();
             }
         }
 
+        /**
+         * Retrieve dummy rowkey and return to the client.
+         *
+         * @param resultsToReturn dummy cell.
+         * @return true if more rows exist after this one, false if scanner is done.
```
tkhurana, line 667: This comment can be updated since we are always returning true

phoenix-core-server/src/main/java/org/apache/phoenix/coprocessor/GroupedAggregateRegionObserver.java
```diff
@@ -436,36 +622,79 @@ public boolean next(List<Cell> resultsToReturn) throws IOException {
...
+         */
+        private boolean getDummyResult(List<Cell> resultsToReturn) {
+            if (lastReturnedRowKey != null) {
+                ScanUtil.getDummyResult(lastReturnedRowKey, resultsToReturn);
+                return true;
+            }
+            if (scanStartRowKey.length > 0 && !ScanUtil.isLocalIndex(scan)) {
+                if (Bytes.compareTo(prevScanStartRowKey, scanStartRowKey) !=
```
tkhurana, line 675: This is the same condition we used above on line 493..495 for detecting region move Maybe introduce a util function in this class. That will also make it self descriptive that we are implementing this logic to handle region move

phoenix-core-server/src/main/java/org/apache/phoenix/coprocessor/GroupedAggregateRegionObserver.java
```diff
@@ -436,36 +622,79 @@ public boolean next(List<Cell> resultsToReturn) throws IOException {
...
+                        0 || prevScanIncludeStartRowKey != includeStartRowKey) {
+                    byte[] lastByte =
+                            new byte[]{scanStartRowKey[scanStartRowKey.length - 1]};
+                    if (scanStartRowKey.length > 1 && Bytes.compareTo(lastByte,
+                            Bytes.toBytesBinary("\\x00")) == 0) {
+                        byte[] prevKey = new byte[scanStartRowKey.length - 1];
```
tkhurana, line 681: This can be made a byte util function to calculate the previous key given a key.

phoenix-core-server/src/main/java/org/apache/phoenix/coprocessor/GroupedAggregateRegionObserver.java
```diff
@@ -411,6 +487,115 @@ private UnorderedGroupByRegionScanner(final ObserverContext<RegionCoprocessorEnv
...
+                        if (ScanUtil.isDummy(resultsToReturn)) {
+                            return true;
+                        }
+                        if (resultsToReturn.size() > 0) {
+                            lastReturnedRowKey = CellUtil.cloneRow(resultsToReturn.get(0));
+                        }
+                        return moreRows;
+                    } else if (
+                            Bytes.compareTo(
+                                    ByteUtil.concat(resultRowKey, Bytes.toBytesBinary("\\x00")),
```
tkhurana, line 563: We can create a static final object for `Bytes.toBytesBinary("\\x00"))`

phoenix-core-server/src/main/java/org/apache/phoenix/coprocessor/GroupedAggregateRegionObserver.java
tkhurana: And this is always set to true so can it even change when the scan resumes. Won't it always be true ?

phoenix-core-server/src/main/java/org/apache/phoenix/coprocessor/GroupedAggregateRegionObserver.java
```diff
@@ -411,6 +487,115 @@ private UnorderedGroupByRegionScanner(final ObserverContext<RegionCoprocessorEnv
 
         @Override
         public boolean next(List<Cell> resultsToReturn) throws IOException {
+            if (firstScan && prevScanStartRowKey != null && prevScanIncludeStartRowKey != null) {
+                firstScan = false;
```
tkhurana, line 491: Can't this logic when `firstScan` is true moved into the constructor and then we can remove this variable. One less state to maintain.

virajjasani: If the region has moved, only then `firstScan` will be true while resuming the scan.

phoenix-core-server/src/main/java/org/apache/phoenix/coprocessor/GroupedAggregateRegionObserver.java
```diff
@@ -411,6 +487,115 @@ private UnorderedGroupByRegionScanner(final ObserverContext<RegionCoprocessorEnv
...
+                            // use updateScannerBasedOnPrevRowKey flag and also reset the scanner
+                            // at paging region scanner level to re-read the previously sent
+                            // values in order to re-compute the aggregation and then return
+                            // only the next rowkey that was not yet sent back to the client.
+                            updateScannerBasedOnPrevRowKey = true;
+                            scan.setAttribute(QueryServices.PHOENIX_PAGING_NEW_SCAN_START_ROWKEY,
+                                    prevScanStartRowKey);
+                            scan.setAttribute(
+                                    QueryServices.PHOENIX_PAGING_NEW_SCAN_START_ROWKEY_INCLUDE,
+                                    Bytes.toBytes(prevScanIncludeStartRowKey));
```
tkhurana, line 522: The difference between these 2 is not clear. Both the conditions detect that the region has moved and set attributes to adjust the paging region scan. And if the `prevScanStartRowKey` will not change ever because it is a scan attribute when will concatenating it with `\x00` give you the current scan's start row key ?

phoenix-core-server/src/main/java/org/apache/phoenix/coprocessor/GroupedAggregateRegionObserver.java
```diff
@@ -423,6 +608,7 @@ public boolean next(List<Cell> resultsToReturn) throws IOException {
                     if (regionScanner != null) {
                         return regionScanner.next(resultsToReturn);
                     }
+                    ImmutableBytesPtr currentRowKey = null;
```
tkhurana, line 611: redundant variable. Not being used

phoenix-core-server/src/main/java/org/apache/phoenix/coprocessor/GroupedAggregateRegionObserver.java
```diff
@@ -436,36 +622,79 @@ public boolean next(List<Cell> resultsToReturn) throws IOException {
...
-                                getDummyResult(resultsToReturn);
-                                return true;
+                                return getDummyResult(resultsToReturn);
                             }
                             result.setKeyValues(results);
                             ImmutableBytesPtr key =
                                     TupleUtil.getConcatenatedValue(result, expressions);
+                            ImmutableBytesPtr originalRowKey = new ImmutableBytesPtr();
+                            result.getKey(originalRowKey);
+                            currentRowKey = originalRowKey;
```
tkhurana, line 632: currentRowKey is assigned but seems it is never used.

tkhurana (changes requested):

phoenix-core-client/src/main/java/org/apache/phoenix/coprocessorclient/BaseScannerRegionObserverConstants.java
```diff
@@ -135,6 +135,12 @@ public static long getMaxLookbackInMillis(Configuration conf){
     public static final String EMPTY_COLUMN_QUALIFIER_NAME = "_EmptyCQName";
     public static final String INDEX_ROW_KEY = "_IndexRowKey";
     public static final String READ_REPAIR_TRANSFORMING_TABLE = "_ReadRepairTransformingTable";
+    public static final String SCAN_ANALYZE_ACTUAL_START_ROW = "_ScanAnalyzeActualStartRow";
+    public static final String SCAN_ANALYZE_ACTUAL_STOP_ROW = "_ScanAnalyzeActualStopRow";
+    public static final String SCAN_ANALYZE_INCLUDE_START_ROW = "_ScanAnalyzeIncludeStartRow";
+    public static final String SCAN_ANALYZE_INCLUDE_STOP_ROW = "_ScanAnalyzeIncludeStopRow";
+    public static final String SCAN_SERVER_RETURN_VALID_ROW_KEY = "_ScanServerValidRowKey";
+
```
tkhurana, line 143: Good to explain what these attributes mean and are used for

phoenix-core-client/src/main/java/org/apache/phoenix/util/ByteUtil.java
```diff
@@ -469,6 +474,116 @@ public static byte[] previousKey(byte[] key) {
...
+     * range of [startKey, endKey). If startKey and endKey are empty, the empty key is returned.
+     * This function is used to return valid rowkey for some ungrouped aggregation e.g. while
+     * returning count value after scanning the rows. If any error or validation issues (e.g.
+     * startKey > endKey) are encountered, null value is returned.
+     *
+     * @param startKey start rowkey for the range.
+     * @param endKey end rowkey for the range.
+     * @return best attempt of largest rowkey in the range of [startKey, endKey).
+     */
+    public static byte[] getLargestPossibleRowKeyInRange(byte[] startKey, byte[] endKey) {
```
tkhurana, line 488: Can we add some unit tests for this ?

phoenix-core-server/src/main/java/org/apache/phoenix/coprocessor/GroupedAggregateRegionObserver.java
tkhurana: But it is set to true where it is declared

phoenix-core-server/src/main/java/org/apache/phoenix/coprocessor/GroupedAggregateRegionObserver.java
tkhurana: OK, I got it. The first condition is true when we haven't returned any valid rows yet.

phoenix-core-client/src/main/java/org/apache/phoenix/iterate/OrderedResultIterator.java
```diff
@@ -172,6 +188,28 @@ public OrderedResultIterator(ResultIterator delegate,
...
+                                 boolean spoolingEnabled,
+                                 long thresholdBytes, Integer limit, Integer offset,
+                                 int estimatedRowSize, long pageSizeMs, Scan scan,
+                                 RegionInfo regionInfo) {
+        this(delegate, orderByExpressions, spoolingEnabled, thresholdBytes, limit, offset,
+                estimatedRowSize, pageSizeMs);
+        this.scan = scan;
+        // If scan start rowkey is empty, use region boundaries. Reverse region boundaries
+        // for reverse scan.
+        this.scanStartRowKey = scan.getStartRow().length > 0 ? scan.getStartRow() :
```
tkhurana, line 202: I see this logic being used in many places. If we can add a util function for this too it will be helpful

phoenix-core-server/src/main/java/org/apache/phoenix/coprocessor/GroupedAggregateRegionObserver.java
```diff
@@ -411,6 +487,115 @@ private UnorderedGroupByRegionScanner(final ObserverContext<RegionCoprocessorEnv
...
+                                    QueryServices.PHOENIX_PAGING_NEW_SCAN_START_ROWKEY_INCLUDE,
+                                    Bytes.toBytes(prevScanIncludeStartRowKey));
+                        } else {
+                            // This happens when the server side scanner has already sent some
+                            // rows back to the client and region has moved, so now we need to
+                            // use updateScannerBasedOnPrevRowKey flag and also reset the scanner
+                            // at paging region scanner level to re-read the previously sent
+                            // values in order to re-compute the aggregation and then return
+                            // only the next rowkey that was not yet sent back to the client.
+                            updateScannerBasedOnPrevRowKey = true;
```
tkhurana, line 517: A better name for this variable could be skipValidRowsSent

phoenix-core-client/src/main/java/org/apache/phoenix/iterate/OrderedResultIterator.java
virajjasani: Can do this but now that we have server and client modules, we won't be able to use one util by all coprocs. Except for `OrderedResultIterator`, every other coproc can use it. This happens if we move util function to server util class.

On the other hand, we can't move util function to core client module because client module does not have access to hbase server classes e.g. `Region`.

phoenix-core-server/src/main/java/org/apache/phoenix/coprocessor/GroupedAggregateRegionObserver.java
virajjasani: The reason why it is kept true here is because everytime when we set `_ScanActualStartRow` scan attribute, we always use `scan#withStartRow` function which internally sets include start row as true. So, yes this never changes, this is meant to be true only, even across region moves after scanner resets.

phoenix-core-server/src/main/java/org/apache/phoenix/coprocessor/GroupedAggregateRegionObserver.java
virajjasani: No reason as such, ImmutableBytesPtr extends ImmutableBytesWritable anyway, and both have similar hashcode and equals functions. We can use either one, `ImmutableBytesWritable` is more generic and while retrieving key/value pair from the map, we actually use `ImmutableBytesWritable` type data:
```
            for (Map.Entry<ImmutableBytesPtr, Aggregator[]> entry : aggregateMap.entrySet()) {
                ImmutableBytesWritable aggregateGroupValPtr = entry.getKey();
                Aggregator[] rowAggregators = entry.getValue();
```

Hence, even though aggregateMap has `ImmutableBytesPtr` as key, we still use `ImmutableBytesWritable` as generic key type.

virajjasani:
Added a few tests for region split also.

tkhurana (approved):

PR #1738, opened 2023-11-14
PHOENIX-7014 Query compiler/optimizer changes for SELECT queries against CDC

haridsv:
This change tries to short-circuit the existing optimizer logic to produce a plan that is suitable for CDC queries. It also reimplements the query hints so that it is simpler and easier to maintain using regex. It includes a PoC query test and some debug test cases for short term reference (can be removed before merging to master).

https://issues.apache.org/jira/browse/PHOENIX-7014


ranganathg:
Just trying to think loud (I haven't looked at all the CDC related PRs also). While looking at the PR at a high level, I was thinking would it better to refactor the code to follow a design pattern which could possibly avoid having if conditions to check if(feature == CDC) to using inheritance? May be while INDEXing changes were added it could have been much over head to do such refactoring. Does my comment make sense?

haridsv:
Superseded by #1766

PR #1741, opened 2023-11-22
PHOENIX-7109 Incorrect query results when using OFFSET

Aarchy:


phoenix-core/src/main/java/org/apache/phoenix/iterate/OffsetResultIterator.java
```diff
@@ -49,7 +49,7 @@ public Tuple next() throws SQLException {
         while (rowCount < offset) {
             Tuple tuple = super.next();
             if (tuple == null) { return null; }
-            if (isDummy(tuple)) {
```
stoty, line 52: very minor nit:
.size() is probably cheaper than isDummy(), so switching the two predicates may be advised.

stoty (approved):
+1 LGTM

PR #1745, opened 2023-11-29
PHOENIX-7130 Support skipping of shade sources jar creation

NihalJain:


stoty (approved):
+1 LGTM

NihalJain:
Hi @stoty are we good here? Also please let me know if should take this to 5.1. Will need to raise another PR  as seems an additional file change is needed in branch-5.1.

pom.xml
```diff
@@ -1887,6 +1888,17 @@
           @{jacocoArgLine}</argLine>
       </properties>
     </profile>
+    <profile>
```
stoty, line 1891: Do we need a profile for this ?
Specifying -DshadeSources=false would have the same effect.

But if you  prefer this way, it's fine, too.

stoty (approved):

pom.xml
NihalJain: Yes actually just having a profile for a variable which can be passed from outside does not make sense. But I was thinking if we also add `maven.source.skip` to be true `here` and rename this profile to `skipSourceProcessing` or something.

NihalJain: Hi @stoty Removed the profile as suggested by you.

NihalJain:
Please donot push this to 5.1, have raised another PR for 5.1 as code is a bit different.

stoty:
I have tested this now, but the shading time difference is seems to be minimal.
On my machine it takes ~3:20 with the shaded sources, and ~3:05 without them.

I don't have a problem with committing this, but I wander what kind of improvement do you see with the patch, @NihalJain  ?

Also I have just merged phoenix-client-lite, you may want to update the patch to add same change there.

NihalJain:
> I have tested this now, but the shading time difference is seems to be minimal. On my machine it takes ~3:20 with the shaded sources, and ~3:05 without them.

In our internal jenkins, which is based on branch-5.1, the difference is substantial. Maybe because we download all jars for every build and also possibly our local artifactory + VPN is not as fast, so downloading the source jars take almost 33% of build time.

Without  `-PskipShadeSources -Dmaven.source.skip=true`
```
.
.
[INFO] Phoenix Client ..................................... SUCCESS [22:40 min]
[INFO] Phoenix Client Embedded ............................ SUCCESS [08:56 min]
[INFO] Phoenix Server JAR ................................. SUCCESS [01:54 min]
[INFO] Phoenix Assembly ................................... SUCCESS [ 38.548 s]
[INFO] ------------------------------------------------------------------------
[INFO] BUILD SUCCESS
[INFO] ------------------------------------------------------------------------
[INFO] Total time:  36:06 min
[INFO] Finished at: 2024-01-03T10:55:52Z
[INFO] ------------------------------------------------------------------------
```

Vs 

With  `-PskipShadeSources -Dmaven.source.skip=true`
```
.
.

[INFO] Phoenix Client ..................................... SUCCESS [09:37 min]
[INFO] Phoenix Client Embedded ............................ SUCCESS [08:14 min]
[INFO] Phoenix Server JAR ................................. SUCCESS [01:49 min]
[INFO] Phoenix Assembly ................................... SUCCESS [ 38.593 s]
[INFO] ------------------------------------------------------------------------
[INFO] BUILD SUCCESS
[INFO] ------------------------------------------------------------------------
[INFO] Total time:  24:08 min
[INFO] Finished at: 2024-01-03T11:29:24Z
[INFO] ------------------------------------------------------------------------
```


> Also I have just merged phoenix-client-lite, you may want to update the patch to add same change there.

Sure let me rebase and make appropriate changes, if any

stoty:
Ouch.
That's extremley slow.
Is it on a spinning rust HDD ? Or some slow NAS ?

gjacoby126 (approved):
+1

gjacoby126:
(Note that we're waiting on #1736 before merging since it's a blocker.)

stoty:
This is a pom only change, and #1736 doesn't touch the poms, so I wouldn't expect them to confict in any way.

virajjasani:
I think it's fine to merge this since it adds the option to skip the shaded jar creation.

virajjasani (approved):

stoty:
Also, I'm going to merge the phoenix-server shading refactor soon (another pom only change), It would probably be a good idea to  add this change there as well.

virajjasani:
pom changes are fine so far, but yeah any other big changes or feature changes that have potential to cause any additional issues (more than what is known with #1736 and make it worse) are potential blockers e.g. JSON support, which is technically ready for merge (though still in final review phase) but i would like to block it.

NihalJain:
Rebased code and updated all new instances of createSourcesJar in code. Could you please review and merge this.?

CC: @gjacoby126 @stoty @virajjasani

virajjasani (approved):
+1

chrajeshbabu:
Since build failures are not related this. Merging

PR #1751, opened 2023-12-05
PHOENIX-7006: Configure maxLookbackAge at table level 

sanjeet006py:
[Design doc](https://docs.google.com/document/d/12q_ujY1FwftCkQGbeSzFrRJRGy4Vb-JRbK_3N-XiDzU/edit)

- Support for table level max lookback age in SYSCAT.
- Support for table level max lookback age in CREATE/ALTER TABLE.
- Block setting/altering max lookback age for views and indexes.
- Retrieve max lookback age for view and index from data table via hierarchical lookup and store it only for data table.
- Invalidate metadata cache on alter table for child views and associated indexes (view indexes and data table indexes) so that new PTable instance in loaded in cache with update max lookback age.
- Use table level max lookback age in coproc hooks wherever cluster level max lookback age is being used. Cluster level max lookback age will be fallback if table level max lookback age is not set.
- Use table level max lookback in Index verification in GlobalIndexRegionScanner. Cluster level max lookback age will be fallback if table level max lookback age is not set.
- Use table level max lookback age in IndexScrutinyTool.
- Test coverage for all the above items.


phoenix-core/src/it/java/org/apache/phoenix/end2end/CreateTableIT.java
```diff
@@ -1699,6 +1701,27 @@ public void testCreateTableWithNoVerify() throws SQLException, IOException, Inte
...
+    @Test
+    public void testCreateTableWithTableLevelMaxLookbackAge() throws Exception {
+        String schemaName = generateUniqueName();
+        String dataTableName = generateUniqueName();
+        String fullTableName = SchemaUtil.getTableName(schemaName, dataTableName);
+        try(Connection conn = DriverManager.getConnection(getUrl())) {
+            String createDdl = "CREATE TABLE " + fullTableName +
+                    " (id char(1) NOT NULL," + " col1 integer NOT NULL," + " col2 bigint NOT NULL," +
+                    " CONSTRAINT NAME_PK PRIMARY KEY (id, col1, col2)) " +
+                    "MAX_LOOKBACK_AGE=259200000"; // Set table level max lookback age to 3 days in milli-seconds
```
gjacoby126, line 1713: nit: 259200000 can be a variable or constant since it's used twice in this method.

phoenix-core/src/main/java/org/apache/phoenix/schema/MetaDataClient.java
```diff
@@ -3153,6 +3158,7 @@ public boolean isViewReferenced() {
             } else {
                 tableUpsert.setNull(36, Types.VARCHAR);
             }
+            tableUpsert.setLong(37, maxLookbackAge);
```
gjacoby126, line 3161: When creating a table from now on maxlookbackage will default to 0 instead of NULL? Are we sure that we aren't introducing assumptions that max lookback age is non null elsewhere, since existing tables will have the value be null?

phoenix-core/src/main/java/org/apache/phoenix/schema/MetaDataClient.java
```diff
@@ -5882,6 +5912,14 @@ public void setSchemaVersion(String schemaVersion) {
         public void setStreamingTopicName(String streamingTopicName) {
             this.streamingTopicName = streamingTopicName;
         }
+
+        public Long getMaxLookbackAge() {
```
gjacoby126, line 5916: So here max lookback age is a Long, and we have some logic dealing with it being null at, e.g line 5578...but elsewhere like on PTable we're assuming it's a long that can't be null.

phoenix-core/src/main/java/org/apache/phoenix/schema/PTable.java
```diff
@@ -997,6 +997,8 @@ IndexMaintainer getIndexMaintainer(PTable dataTable, PhoenixConnection connectio
      * @throws SQLException
      */
     Set<ColumnReference> getIndexWhereColumns(PhoenixConnection connection) throws SQLException;
+
+    long getMaxLookbackAge();
```
gjacoby126, line 1001: We're absolutely sure that every path that can possibly create a PTable from a table without a max lookback age will transparently use the default?

phoenix-core/src/it/java/org/apache/phoenix/end2end/AlterTableIT.java
```diff
@@ -1795,4 +1795,45 @@ public void testChangePropertiesUpdatesLASTDDLTimestamp() throws Exception {
...
+
+    private void assertMaxLookbackAge(String schemaName, String dataTableName, long expectedMaxLookbackAge) throws Exception {
+        String query = "SELECT MAX_LOOKBACK_AGE FROM \"SYSTEM\".\"CATALOG\"\n"
+                + "WHERE TENANT_ID IS NULL AND\n"
+                + "(TABLE_SCHEM, TABLE_NAME) = ('" + schemaName + "','"+ dataTableName + "') AND\n"
+                + "COLUMN_FAMILY IS NULL AND COLUMN_NAME IS NULL";
+        try (Connection conn = DriverManager.getConnection(getUrl())) {
+            ResultSet rs = conn.createStatement().executeQuery(query);
+            assertTrue(rs.next());
+            assertEquals(expectedMaxLookbackAge, rs.getLong(1));
```
lokiore, line 1835: Can you also add test in which you are asserting maxlookback through PTable object, I mean getting maxlookback from getMaxLookbackAge() API so that, that code path is also tested. 
Also add assertions in some old tests or new one where value is null in SYSCAT.

tkhurana:
Another thing we should consider is that maxlookback on index should inherit the value of the base table. The behavior should be exactly similar to TTL where both the data table and index have same values and you cannot directly set ttl on index. It always inherits the TTL value from base table.

phoenix-core/src/main/java/org/apache/phoenix/schema/MetaDataClient.java
sanjeet006py: Got it, so I think the max lookback age as undefined should be either null or 0. And, preferrable will be null as already for existing tables we will have it as NULL. Please correct me if I am understanding it wrong.

sanjeet006py:
> Another thing we should consider is that maxlookback on index should inherit the value of the base table. The behavior should be exactly similar to TTL where both the data table and index have same values and you cannot directly set ttl on index. It always inherits the TTL value from base table.

Yes, I will be doing that only. Current set of changes include SYSCAT schema changes and CREATE/ALTER Table support. In the subsequent changes I will be adding support to block max lookback age modifications for index.

phoenix-core/src/main/java/org/apache/phoenix/schema/MetaDataClient.java
sanjeet006py: Actually I was of the understanding that I took care of converting existing NULLs in SYSCAT to 0 (default) via: https://github.com/sanjeet006py/phoenix/blob/PHOENIX-7006/phoenix-core/src/main/java/org/apache/phoenix/coprocessor/MetaDataEndpointImpl.java#L1448, but I agree it good to be consistent around definition of max lookback age not defined.

sanjeet006py:
Jenkins test this please

phoenix-core-client/src/main/java/org/apache/phoenix/schema/TableProperty.java
```diff
@@ -336,6 +336,29 @@ public Object getValue(Object value) {
...
+                return null;
+            }
+            else if (value instanceof Integer) {
+                return Long.valueOf((Integer) value);
+            }
+            else if (value instanceof Long) {
+                return value;
+            }
+            else {
+                throw new IllegalArgumentException("Table level MAX_LOOKBACK_AGE should be a numeric value in milli-seconds");
```
gjacoby126, line 354: nit: instead of "numeric" specify BIGINT, since "numeric" might refer to a floating point value that isn't supported.

phoenix-core/src/it/java/org/apache/phoenix/end2end/AlterTableIT.java
```diff
@@ -1795,4 +1796,65 @@ public void testChangePropertiesUpdatesLASTDDLTimestamp() throws Exception {
...
     }
-}
\ No newline at end of file
+
+    @Test
+    public void testChangeTableLevelMaxLookbackAge() throws Exception {
+        String schemaName = generateUniqueName();
+        String dataTableName = generateUniqueName();
+        String fullTableName = SchemaUtil.getTableName(schemaName, dataTableName);
+        long baseMaxLookbackAge = 86400000; // 1 day
```
gjacoby126, line 1805: this can be a constant because it's used at 1830. Also it should be 86400000L to avoid a cast. Alternately you could use Duration.ofDays(1).getSeconds() * 1000 if you want to make it more explicit.

phoenix-core/src/it/java/org/apache/phoenix/end2end/AlterTableIT.java
gjacoby126: This would be simpler to do as PhoenixRuntime.getTableNoCache().getMaxLookbackAge()

phoenix-core/src/it/java/org/apache/phoenix/end2end/AlterTableIT.java
```diff
@@ -1795,4 +1796,65 @@ public void testChangePropertiesUpdatesLASTDDLTimestamp() throws Exception {
                     newLastDDLTimestamp > oldLastDDLTimestamp);
         }
     }
-}
\ No newline at end of file
+
```
gjacoby126, line 1799: What about if someone sets MAX_LOOKBACK_AGE to NULL?

phoenix-core/src/it/java/org/apache/phoenix/end2end/CreateTableIT.java
```diff
@@ -1727,4 +1767,26 @@ private int checkGuidePostWidth(String tableName) throws Exception {
...
+            String createDdl = "CREATE TABLE " + fullTableName +
+                    " (id char(1) NOT NULL," + " col1 integer NOT NULL," + " col2 bigint NOT NULL," +
+                    " CONSTRAINT NAME_PK PRIMARY KEY (id, col1, col2)) " +
+                    "MAX_LOOKBACK_AGE="+maxLookbackAge;
+            conn.createStatement().execute(createDdl);
+        }
+    }
+
+    private Long queryTableLevelMaxLookbackAge(String schemaName, String dataTableName) throws Exception {
+        try(Connection conn = DriverManager.getConnection(getUrl())) {
```
gjacoby126, line 1781: Likewise this would also be simpler using PhoenixRuntime.getTableNoCache() and then looking up from the PTable that's returned.

phoenix-core-client/src/main/java/org/apache/phoenix/schema/TableProperty.java
```diff
@@ -336,6 +336,29 @@ public Object getValue(Object value) {
...
+        @Override
+        public Object getValue(Object value) {
+            if (value == null) {
+                return null;
+            }
+            else if (value instanceof Integer) {
+                return Long.valueOf((Integer) value);
+            }
+            else if (value instanceof Long) {
+                return value;
```
haridsv, line 351: What about -ve numbers?

phoenix-core/src/it/java/org/apache/phoenix/end2end/AlterTableIT.java
```diff
@@ -1795,4 +1796,65 @@ public void testChangePropertiesUpdatesLASTDDLTimestamp() throws Exception {
...
+        try (Connection conn = DriverManager.getConnection(getUrl())) {
+            String ddl = "CREATE TABLE  " + fullTableName +
+                    "  (a_string varchar not null, a_binary VARCHAR not null, col1 integer" +
+                    "  CONSTRAINT pk PRIMARY KEY (a_string, a_binary)) " + "MAX_LOOKBACK_AGE=" + maxLookbackAge;
+            conn.createStatement().execute(ddl);
+            assertMaxLookbackAge(schemaName, dataTableName, maxLookbackAge);
+            maxLookbackAge = 3L * baseMaxLookbackAge;
+            alterTableLevelMaxLookbackAge(fullTableName, maxLookbackAge.toString());
+            assertMaxLookbackAge(schemaName, dataTableName, maxLookbackAge);
+            maxLookbackAge = 2L * baseMaxLookbackAge;
```
haridsv, line 1816: How is this different from above?

phoenix-core/src/it/java/org/apache/phoenix/end2end/AlterTableIT.java
```diff
@@ -1795,4 +1796,65 @@ public void testChangePropertiesUpdatesLASTDDLTimestamp() throws Exception {
...
+        }
+    }
+
+    private void alterTableLevelMaxLookbackAge(String fullTableName, String maxLookbackAge) throws Exception {
+        try(Connection conn = DriverManager.getConnection(getUrl())) {
+            String ddl  = "ALTER TABLE " + fullTableName + " SET MAX_LOOKBACK_AGE = " + maxLookbackAge;
+            conn.createStatement().execute(ddl);
+        }
+    }
+}
```
haridsv, line 1860: What about testing with parent table?

phoenix-core/src/it/java/org/apache/phoenix/end2end/CreateTableIT.java
```diff
@@ -1699,6 +1701,44 @@ public void testCreateTableWithNoVerify() throws SQLException, IOException, Inte
...
+    public void testCreateTableWithTableLevelMaxLookbackAgeAsNull() throws Exception {
+        String schemaName = generateUniqueName();
+        String dataTableName = generateUniqueName();
+        String fullTableName = SchemaUtil.getTableName(schemaName, dataTableName);
+        createTableWithTableLevelMaxLookbackAge(fullTableName, "NULL");
+        assertNull(queryTableLevelMaxLookbackAge(schemaName, dataTableName));
+        schemaName = generateUniqueName();
+        dataTableName = generateUniqueName();
+        fullTableName = SchemaUtil.getTableName(schemaName, dataTableName);
+        createTableWithTableLevelMaxLookbackAge(fullTableName, "null");
```
haridsv, line 1730: This variant is not nneeded, it is really only testing the parser literal token.

phoenix-core/src/it/java/org/apache/phoenix/end2end/CreateTableIT.java
```diff
@@ -1727,4 +1767,26 @@ private int checkGuidePostWidth(String tableName) throws Exception {
         }
     }
 
+    private void createTableWithTableLevelMaxLookbackAge(String fullTableName, String maxLookbackAge) throws Exception {
+        try(Connection conn = DriverManager.getConnection(getUrl())) {
+            String createDdl = "CREATE TABLE " + fullTableName +
+                    " (id char(1) NOT NULL," + " col1 integer NOT NULL," + " col2 bigint NOT NULL," +
+                    " CONSTRAINT NAME_PK PRIMARY KEY (id, col1, col2)) " +
+                    "MAX_LOOKBACK_AGE="+maxLookbackAge;
```
haridsv, line 1775: As I commented earlier, there is no need to use anything but the simplest table schema here.

haridsv (changes requested):

phoenix-core-client/src/main/java/org/apache/phoenix/schema/TableProperty.java
sanjeet006py: Negative numbers are not parsed by grammar as I tried and got error. As per grammar it seems like only whole numbers are expected.

phoenix-core/src/it/java/org/apache/phoenix/end2end/AlterTableIT.java
sanjeet006py: First I created a table with maxlookback age 1 day, then I increased it to 3 days and now decreasing it to 2 days. Lastly setting it to 0.

phoenix-core/src/it/java/org/apache/phoenix/end2end/AlterTableIT.java
sanjeet006py: You mean for indexes? If so then it will be part of next batch of changes.

phoenix-core/src/it/java/org/apache/phoenix/end2end/AlterTableIT.java
sanjeet006py: Currently, for most table level properties setting them to NULL via Alter is no-op. Created a JIRA issue: https://issues.apache.org/jira/browse/PHOENIX-7161 for the same. Thanks

sanjeet006py:
@gjacoby126 @haridsv please re-review the PR. Thanks

phoenix-core-client/src/main/java/org/apache/phoenix/compile/QueryCompiler.java
```diff
@@ -190,8 +192,20 @@ private void verifySCN() throws SQLException {
         if (scn == null) {
             return;
         }
-        long maxLookBackAgeInMillis =
-            BaseScannerRegionObserverConstants.getMaxLookbackInMillis(conn.getQueryServices().getConfiguration());
+        ColumnResolver resolver = FromCompiler.getResolverForQuery(select, conn);
```
kadirozde, line 195: I do not think we need to get a new ColumnResolver here. Can we use the column resolver of the QueryCompiler object. Please note FromCompiler.getResolverForQuery() makes an RPC call to the syscat region.

phoenix-core-server/src/main/java/org/apache/phoenix/coprocessor/BaseScannerRegionObserver.java
```diff
@@ -527,6 +535,28 @@ public static boolean isMaxLookbackTimeEnabled(long maxLookbackTime){
...
+        PTable table;
+        try(PhoenixConnection conn = QueryUtil.getConnectionOnServer(
+                conf).unwrap(PhoenixConnection.class)) {
+            table = conn.getTableNoCache(fullTableName);
+        }
+        catch (Exception e) {
+            if (e instanceof TableNotFoundException) {
+                LOGGER.debug("Ignoring HBase table that is not a Phoenix table: "
+                        + fullTableName);
+                // non-Phoenix HBase tables won't be found, do nothing
```
kadirozde, line 551: We need to see if the HBase table name for a given Phoenix table has changed by the online data format change feature. We need query the system catalog using WHERE  PHYSICAL_TABLE_NAME = '<this HBase table name>' to find the Phoenix table name.

phoenix-core/src/it/java/org/apache/phoenix/end2end/CreateTableIT.java
kadirozde: Please use conn.getTableNoCache instead.

phoenix-core-server/src/main/java/org/apache/phoenix/coprocessor/BaseScannerRegionObserver.java
sanjeet006py: Thanks I missed this scenario. Wondering if same can happen in here also: https://github.com/apache/phoenix/blob/master/phoenix-core-server/src/main/java/org/apache/phoenix/coprocessor/UngroupedAggregateRegionObserver.java#L601-L614.

sanjeet006py: If we have a table `T1` and we do a transform on it, seems like SYSCAT will have following two records: 

- A record with `TABLE_NAME` as `T1_1` and `PHYSICAL_TABLE_NAME` also as `T1_1`.
- A record with `TABLE_NAME` as `T1` and `PHYSICAL_TABLE_NAME` as `T1_1`.

As we are looking up Phoenix table from HBase table shall we skip the records where `TABLE_NAME` == `PHYSICAL_TABLE_NAME` and `PHYSICAL_TABLE_NAME is NOT NULL`?

sanjeet006py: As above WHERE clause was causing FULL SCAN over SYSCAT and to avoid that went with an alternate approach. Please take a look. Thanks

sanjeet006py: > What is the alternative approach?

In `TransformClient` we derive new table name from existing table name. I saw we are appending sequence num to eixtsing table's full name separated by underscore. I added a logic to derive logical table name from physical table name by checking if physical table is underscore separated and if part of table name before underscore is valid logical table name in SYSCAT along with checking if the logical table found has physical table as original physical table name.

I thought this will avoid FULL SCAN along max be one RPC call but even any READ operation on any table even if we create an index will be at least one RPC call. Please correct me if I misunderstood something. Thanks

kadirozde: Based on a side conversation with @sanjeet006py, I learned that  the IT tests for Transform is not running currently due to missing @Category annotation and some of these tests are broken. We need to fix the transform tool issues in a separate Jira. We can assume that the transform tool is not supported yet and the name of a Phoenix table and the name of its HBase table are the same for now. When we fix transform tool issues, we need to handle the table name mismatch issues in general.

kadirozde: > I saw we are appending sequence num to eixtsing table's full name separated by underscore
I see. We need to revisit this later to see if it is safe to rely on this.

sanjeet006py:
@kadirozde I have addressed all the review comments. Please re-review. Thanks

phoenix-core-server/src/main/java/org/apache/phoenix/coprocessor/MetaDataRegionObserver.java
```diff
@@ -61,6 +61,7 @@
 import org.apache.hadoop.hbase.security.User;
 import org.apache.hadoop.hbase.shaded.protobuf.generated.AdminProtos;
 import org.apache.hadoop.hbase.util.Pair;
+import org.apache.hadoop.hbase.util.Bytes;
```
virajjasani, line 64: unused

virajjasani (commented):
Left some comments, will take another look later

phoenix-core-server/src/main/java/org/apache/phoenix/coprocessor/MetaDataEndpointImpl.java
```diff
@@ -1587,6 +1623,56 @@ private PTable getTableFromCells(List<Cell> tableCellList, List<List<Cell>> allC
         return builder.build();
     }
 
+    private Long scanMaxLookbackAgeFromParent(byte[] key, long clientTimeStamp) throws IOException {
+        Scan scan = MetaDataUtil.newTableRowsScan(key, MIN_TABLE_TIMESTAMP, clientTimeStamp);
+        Table sysCat = ServerUtil.getHTableForCoprocessorScan(this.env,
+                SchemaUtil.getPhysicalTableName(SYSTEM_CATALOG_NAME_BYTES, env.getConfiguration()));
```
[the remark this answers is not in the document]
haridsv: There is a recursive call and the resource is not needed for it, so if we close it before the call begins, the resource can be released a bit sooner and so can be beneficial.

phoenix-core-server/src/main/java/org/apache/phoenix/coprocessor/BaseScannerRegionObserver.java
```diff
@@ -527,6 +535,28 @@ public static boolean isMaxLookbackTimeEnabled(long maxLookbackTime){
...
+        Configuration conf = c.getEnvironment().getConfiguration();
+        PTable table;
+        try(PhoenixConnection conn = QueryUtil.getConnectionOnServer(
+                conf).unwrap(PhoenixConnection.class)) {
+            table = conn.getTableNoCache(fullTableName);
+        }
+        catch (Exception e) {
+            if (e instanceof TableNotFoundException) {
+                LOGGER.debug("Ignoring HBase table that is not a Phoenix table: "
+                        + fullTableName);
```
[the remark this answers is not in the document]
sanjeet006py: Sure, but wondering any drawback with non-parameterized version, just for my knowledge? Thanks

phoenix-core-server/src/main/java/org/apache/phoenix/coprocessor/MetaDataEndpointImpl.java
```diff
@@ -1463,6 +1485,7 @@ private PTable getTableFromCells(List<Cell> tableCellList, List<List<Cell>> allC
 
         PTable transformingNewTable = null;
         boolean isRegularView = (tableType == PTableType.VIEW && viewType != ViewType.MAPPED);
+        boolean isThisAViewIndex = false;
```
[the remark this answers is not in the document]
sanjeet006py: How about `isViewIndex` as we are storing if the PTable instance which we are deserializing is that a view index or not?

phoenix-core-server/src/main/java/org/apache/phoenix/coprocessor/MetaDataEndpointImpl.java
```diff
@@ -1587,6 +1623,56 @@ private PTable getTableFromCells(List<Cell> tableCellList, List<List<Cell>> allC
...
+                if (linkTypeBytes != null) {
+                    LinkType linkType = LinkType.fromSerializedValue(linkTypeBytes[0]);
+                    int rowKeyColMetadataLength = 5;
+                    byte[][] rowKeyMetaData = new byte[rowKeyColMetadataLength][];
+                    getVarChars(result.getRow(), rowKeyColMetadataLength, rowKeyMetaData);
+                    if (linkType == VIEW_INDEX_PARENT_TABLE) {
+                        parentTableKey = getParentTableKeyFromChildRowKeyMetaData(rowKeyMetaData);
+                        return scanMaxLookbackAgeFromParent(parentTableKey, clientTimeStamp);
+                    }
+                    else if (linkType == PHYSICAL_TABLE) {
```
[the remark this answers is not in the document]
sanjeet006py: Yes, because we need to use `DATA_TABLE_NAME` filed in syscat so needs to handled special.

phoenix-core-server/src/main/java/org/apache/phoenix/coprocessor/MetaDataEndpointImpl.java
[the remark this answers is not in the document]
sanjeet006py: In this method there are 4 `return` clauses. Having two `close()` calls before each of them seems a bit unclean while using try-with-resources seems clean. If mem by these extra resources is not a major concern then I think try-with-resource might be better choice.

phoenix-core-server/src/main/java/org/apache/phoenix/coprocessor/BaseScannerRegionObserver.java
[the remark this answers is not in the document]
virajjasani: It's perf improvement depending on the size of the string. When heavy objects are to be logged and if we use non-param version, the string concatenation happens even though the given logger level is not enabled (e.g. even if log level is INFO and the above DEBUG log is not going to be printed, the string concatenation still happens), whereas if we use parameterized version, the string concatenation happens only if the given logger level is enabled and the log is expected to be printed. With parameterized version, the string concatenation takes place within log4j implementation.

virajjasani:
@sanjeet006py the changes overall looks good now on the latest revision. There are some merge conflicts though.

sanjeet006py:
Thanks a lot @virajjasani . I have resolved the conflicts. The conflicts were around the version changes for 5.3.0 creation. Please re-review. Thanks

sanjeet006py:
Once current PR validation is done. Shall I squash and rebase all my changes into one commit or merge procedure will take care of that?

virajjasani (approved):
+1

virajjasani:
Triggered one more build. Once we get results, will merge the PR unless there are any objections.

sanjeet006py:
@virajjasani the latest build is complete. Are we good to merge the PR? Thanks

virajjasani:
Test failures on the build are not relevant.

haridsv (commented):
I had a couple of comments and forgot to submit the review, sorry about doing this after the merge. May be it can help with a future change.

PR #1752, opened 2023-12-07
PHOENIX-6996 :- Provide an upgrade path for Phoenix tables with HBase TTL to move the…

lokiore:
…ir TTL spec to SYSTEM.CATALOG


gjacoby126:
@lokiore - what happens if I don't want to upgrade HBase TTLs to Phoenix TTLs at upgrade time, but I want to do it later?

lokiore:
> @lokiore - what happens if I don't want to upgrade HBase TTLs to Phoenix TTLs at upgrade time, but I want to do it later?

@gjacoby126 No I added this only at upgrade time but after upgrade phoenix.table.ttl.enabled if set true will expect that TTL defined in SYSCAT is the TTL to follow, won't that break functionality if we don't do this at upgrade time!?

I was expecting to know at upgrade time if we want older tables to follow new TTL mechanism or not.
We can try to do this whenever we set phoenix.table.ttl.enabled to true and check if this is done or not. But we would have to store the state.

phoenix-core/src/main/java/org/apache/phoenix/query/QueryServices.java
```diff
@@ -360,6 +360,7 @@ public interface QueryServices extends SQLCloseable {
     public static final String PHOENIX_SERVER_PAGE_SIZE_MS = "phoenix.server.page.size.ms";
     // Phoenix TTL implemented by CompactionScanner and TTLRegionScanner is enabled
     public static final String PHOENIX_TABLE_TTL_ENABLED = "phoenix.table.ttl.enabled";
+    public static final String PHOENIX_TABLE_TTL_ENABLED_FOR_UPGRADE = "phoenix.table.ttl.enabled.upgrade";
```
jpisaac, line 363: nit: Should it be called PHOENIX_TABLE_TTL_ENABLED_IN_SYSCAT?

jpisaac (commented):
LGTM, one nit. Will do one more pass and some local testing.

phoenix-core/src/main/java/org/apache/phoenix/util/UpgradeUtil.java
```diff
@@ -1470,6 +1475,121 @@ public static void copyTTLValuesFromPhoenixTTLColumnToTTLColumn(
 
     }
 
+    public static void moveHBaseLevelTTLToSYSCAT(PhoenixConnection oldMetaConnection,
+                                                 Map<String, String> options) throws IOException {
+        long numOfTableThatHasTTLMoved = 0;
+        ReadOnlyProps readOnlyProps = oldMetaConnection.getQueryServices().getProps();
+        TableName sysCat = SchemaUtil.getPhysicalTableName(SYSTEM_CATALOG_NAME, readOnlyProps);
+
+        LOGGER.debug(String.format("SYSTEM CATALOG tabled use for copying TTL values: %s", sysCat.toString()));
```
jpisaac, line 1484: nit: typo "table used'?

phoenix-core/src/main/java/org/apache/phoenix/exception/SQLExceptionCode.java
```diff
@@ -368,6 +368,8 @@ public SQLException newException(SQLExceptionInfo info) {
 
     TTL_ALREADY_DEFINED_IN_HIERARCHY(10958, "44A40", TTL
             + " property is already defined in hierarchy for this entity"),
+    VIEW_TTL_NOT_ENABLED(10959,"44A41", TTL +
+            " property can not be set on Views as phoenix.view.ttl.enabled is false"),
```
jpisaac, line 372: nit: typo should be lowercase "views"

jpisaac (approved):
Couple of nits. LGTM +1

PR #1753, opened 2023-12-11
PHOENIX-7113 Create a feature flag for invoking MDEI#invalidateServerMetadataCache

shahrs87:


phoenix-core/src/test/java/org/apache/phoenix/query/BaseTest.java
```diff
@@ -650,16 +650,13 @@ public static Configuration setUpConfigForMiniCluster(Configuration conf, ReadOn
...
-        by default. If some other regionserver coprocs are already present then append
-        PhoenixRegionServerEndpoint to the existing coprocs.
+        Set property hbase.coprocessor.regionserver.classes to include PhoenixRegionServerEndpoint
+        by default, if some other regionserver coprocs are not already present.
      */
     private static void setPhoenixRegionServerEndpoint(Configuration conf) {
         String value = conf.get(REGIONSERVER_COPROCESSOR_CONF_KEY);
         if (value == null) {
             value = PhoenixRegionServerEndpoint.class.getName();
-        } else {
```
palashc, line 661: @shahrs87 why was this change needed?

shahrs87: Earlier this was the check to decide whether we invalidate the cache on all regionservers or not. In InvalidateMetadataCacheIT test, we override the regionserver coproc to FailingPhoenixRegionServerEndpoint [here](https://github.com/apache/phoenix/blob/PHOENIX-6883-feature/phoenix-core/src/it/java/org/apache/phoenix/end2end/InvalidateMetadataCacheIT.java#L51).  But we still add `PhoenixRegionServerEndpoint` to the conf property to pass the below check. But now since the check is determined by a separate boolean conf property, we can remove the unnecessary mess that I have created earlier.
```
        String value = conf.get(REGIONSERVER_COPROCESSOR_CONF_KEY);
        if (value == null
                || !value.contains(PhoenixRegionServerEndpoint.class.getName()))
```

palashc: Got it, thanks.

palashc (approved):
+1, thank you.

PR #1760, opened 2023-12-13
PHOENIX-7140 Update Apache Parent and Maven Plugin Versions in Core

NihalJain:
- Apache Parent pom updated to v30
- Versions for the following plugins removed as they are derived from parent pom:
  - maven-enforcer-plugin
  - maven-shade-plugin
  - maven-project-info-reports-plugin
- Unused property surefire.version has been removed
- Properties findbugs-annotations.version, maven-owasp-plugin.version, and os.maven.version have been relocated
- Versions for these plugins are now managed via properties:
  - jasmine-maven-plugin
  - lifecycle-mapping
  - maven-bundle-plugin
  - exec-maven-plugin
  - maven-checkstyle-plugin
- Upgraded versions for these plugins:
  - maven-eclipse-plugin to 2.10
  - maven-build-helper-plugin to 3.5.0
  - restrict-imports.enforcer to 2.4.0
  - spotbugs-maven-plugin to 4.8.1.0
  - spotbugs to 4.8.1
  - jacoco-maven-plugin to 0.8.11
  - sonar-maven-plugin to 3.10.0.2594
  - maven-owasp-plugin to 8.4.0
  - os.maven to 1.7.1
  - maven-bundle-plugin to 5.1.9
  - exec-maven-plugin to 3.1.1
  - maven-checkstyle-plugin to 3.3.0
- Updated checker.xml based on updated checkstyle, refer https://github.com/checkstyle/checkstyle/issues/2116
- Switched to <RestrictImports> as the old declaration variant no longer supported, refer https://github.com/skuzzle/restrict-imports-enforcer-rule/releases/tag/v2.0.0
- Temporarily retained old version of apache parent for maven-dependency-plugin as upgrading is causing build to fail

### DIFF
--- a/phoenix-tracing-webapp/src/main/config/checkstyle/checker.xml
+++ b/phoenix-tracing-webapp/src/main/config/checkstyle/checker.xml
@@ -27,8 +27,8 @@ limitations under the License.
 -->
 
 <!DOCTYPE module PUBLIC
-        "-//Puppy Crawl//DTD Check Configuration 1.3//EN"
-        "http://www.puppycrawl.com/dtds/configuration_1_3.dtd">
+        "-//Checkstyle//DTD Checkstyle Configuration 1.3//EN"
+        "https://checkstyle.org/dtds/configuration_1_3.dtd">
 
 <!-- Calcite customization of default Checkstyle behavior -->
 <module name="Checker">
--- a/phoenix-tracing-webapp/src/main/config/checkstyle/checker.xml
+++ b/phoenix-tracing-webapp/src/main/config/checkstyle/checker.xml
@@ -26,7 +26,9 @@ limitations under the License.
   http://checkstyle.sourceforge.net
 -->
 
-<!DOCTYPE module PUBLIC "-//Puppy Crawl//DTD Check Configuration 1.1//EN" "http://www.puppycrawl.com/dtds/configuration_1_1.dtd">
+<!DOCTYPE module PUBLIC
+        "-//Puppy Crawl//DTD Check Configuration 1.3//EN"
+        "http://www.puppycrawl.com/dtds/configuration_1_3.dtd">
 
 <!-- Calcite customization of default Checkstyle behavior -->
 <module name="Checker">
@@ -55,6 +57,15 @@ limitations under the License.
   <!-- See http://checkstyle.sf.net/config_whitespace.html -->
     <!-- No tabs allowed! -->
   <module name="FileTabCharacter"/>
+
+  <!-- Checks for Size Violations.                    -->
+  <!-- See http://checkstyle.sf.net/config_sizes.html -->
+  <!-- Lines cannot exceed 100 chars, except if they are hyperlinks
+       or strings (possibly preceded by '+' and followed by say '),'. -->
+  <module name="LineLength">
+    <property name="max" value="100"/>
+    <property name="ignorePattern" value="^import|@see|@link|@BaseMessage|href|^[ +]*&quot;.*&quot;[);,]*$"/>
+  </module>
 
   <module name="TreeWalker">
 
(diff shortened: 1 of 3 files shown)


pom.xml
```diff
@@ -401,6 +400,7 @@
         </plugin>
         <plugin>
           <artifactId>maven-dependency-plugin</artifactId>
+          <version>3.1.1</version>
```
NihalJain, line 403: Created PHOENIX-7151, will take this separately.

NihalJain:
Ran following locally and verified build is SUCCESS:
- `mvn clean install -DskipTests`
- `mvn clean compile test-compile site -Dspotbugs.site`
- `mvn checkstyle:checkstyle`
- `mvn pmd:pmd`
- `mvn exec:exec -q -N -Dexec.executable="echo" -Dexec.args='${hbase.profile.list}'`

CC: @stoty

pom.xml
NihalJain: Please let me if you want me to do with this commit. Just don't want to complicate the patch, hence broke into another ticket.

stoty:
You can just use git commit --amend, and edit the patch in your favourite unix-ish command line editor.
Maybe your editor uses CR instead of NL ?

stoty:
Sorry, misclicked.

NihalJain:
> You can just use git commit --amend, and edit the patch in your favourite unix-ish command line editor. Maybe your editor uses CR instead of NL ?

Phew! Fixed the message: 
https://github.com/apache/phoenix/commit/47bf1371abe9ab43d200c1e436bdfdc9dae34030.patch Thank you.

Force pushed: updated DTD version to be same as omid, not sure how to validate as `mvn checkstyle:checkstyle` runs seemlessly locally without the change also, so waiting for build result.

phoenix-tracing-webapp/src/main/config/checkstyle/checker.xml
```diff
@@ -27,8 +27,8 @@ limitations under the License.
 -->
 
 <!DOCTYPE module PUBLIC
-        "-//Puppy Crawl//DTD Check Configuration 1.3//EN"
-        "http://www.puppycrawl.com/dtds/configuration_1_3.dtd">
+        "-//Checkstyle//DTD Checkstyle Configuration 1.3//EN"
```
NihalJain, line 30: Updated based on comment in file http://www.puppycrawl.com/dtds/configuration_1_3.dtd, let see if this helps

NihalJain: its possible this may fix it as we were using http before and now it changes to https as mentioned by someone in https://stackoverflow.com/questions/44333939/the-markup-declarations-contained-or-pointed-to-by-the-document-type-declaration. also possible this check was never passing? lets see what build says

stoty: Looks good according to the tests.

pom.xml
```diff
@@ -62,7 +62,7 @@
   <parent>
     <groupId>org.apache</groupId>
     <artifactId>apache</artifactId>
-    <version>23</version>
+    <version>30</version>
```
stoty, line 65: Is there a reason for not using 31 ?
If you are trying to keep the version in sync, I have unfortunately already used 31 in connectors.

NihalJain: Yes I was just trying to keep in sync acrossthe repos. Can bump to 31 in that case?

stoty: It's fine either way.

stoty: LMK if you consider this ready, and I will merge it.

phoenix-tracing-webapp/src/main/config/checkstyle/checker.xml
NihalJain: > Looks good according to the tests.
> 

Yes this got fixed now. Thankfully :) I am assuming the XML test was never triggered for a long time and was broken for a while.

PR #1765, opened 2023-12-14
PHOENIX-6721 CSV bulkload tool fails with FileNotFoundException if --…

stoty:
…output points to the S3 location

This is @ss77892 's patch, rebased, and using getWorkPath() instead of getOutputPath()


stoty:
FYI @steveloughran 

Is this version correct ?

I have tested this with both HDFS and S3 outputdir.

ss77892:
It looks good to me. I don't quite remember why I didn't use getWorkPath in the first place, but if it works as expected, then I have no objections. Thank you so much, Istvan!

ss77892 (approved):

chrajeshbabu (approved):

gjacoby126 (approved):
+1

stoty:
Merged manually.

steveloughran:
> I don't quite remember why I didn't use getWorkPath in the first place,

@ss77892 we only pulled it up from a FileOutputFormat into an interface with the s3a committer work

PR #1767, opened 2023-12-18
PHOENIX-7155 Validate Partial Index support with JSON

ranganathg:
[PHOENIX-7155](https://issues.apache.org/jira/browse/PHOENIX-7155)


phoenix-core/src/it/java/org/apache/phoenix/end2end/index/PartialIndexIT.java
```diff
@@ -764,4 +768,180 @@ public void testPartialIndexPreferredOverFullIndex() throws Exception {
...
+            // Verify that index table includes only the rows with A > 50
+            selectSql = "SELECT * from " + indexTableName;
+            rs = conn.createStatement().executeQuery(selectSql);
+            assertTrue(rs.next());
+            assertEquals(75, rs.getInt(1));
+            assertTrue(rs.next());
+            assertEquals(100, rs.getInt(1));
+            assertFalse(rs.next());
+
+            // Overwrite an existing row that satisfies the index WHERE clause using an atomic
```
kadirozde, line 832: Remove "using an atomic upsert" from the comment

phoenix-core/src/it/java/org/apache/phoenix/end2end/index/PartialIndexIT.java
```diff
@@ -764,4 +768,180 @@ public void testPartialIndexPreferredOverFullIndex() throws Exception {
...
+            // anymore. This should result in deleting the index row.
+            String dml = "UPSERT INTO " + dataTableName + " values ('id2', 0, 2, 9.5, 'd', JSON_MODIFY(jsoncol, '$.info.age', '0')) "; //, JSON_MODIFY(jsoncol, '$.info.age', '"0"')
+            conn.createStatement().execute(dml);
+            conn.commit();
+            rs = conn.createStatement().executeQuery(selectSql);
+            assertTrue(rs.next());
+            assertEquals(75, rs.getInt(1));
+            assertFalse(rs.next());
+
+            // Retrieve update row from the data table and verify that the index table is not used
```
kadirozde, line 843: Retrieve the updated row ...

phoenix-core/src/it/java/org/apache/phoenix/end2end/index/PartialIndexIT.java
```diff
@@ -764,4 +768,180 @@ public void testPartialIndexPreferredOverFullIndex() throws Exception {
...
+            // Add rows to the data table before creating a partial index to test that the index
+            // will be built correctly by IndexTool
+            conn.createStatement().execute(
+                    "upsert into " + dataTableName + " values ('id1', 25, 2, 3.14, 'a','" + getJsonString(
+                            partialIndexJson, "$[0]") + "')");
+
+            conn.createStatement().execute(
+                    "upsert into " + dataTableName + " (id, A, D, jsoncol) values ('id2', 100, 'b','" + getJsonString(
+                            partialIndexJson, "$[3]") + "')");
+            conn.commit();
```
kadirozde, line 880: Commit is not necessary on a auto-commit connnection

phoenix-core/src/main/java/org/apache/phoenix/compile/CreateIndexCompiler.java
```diff
@@ -110,6 +111,8 @@ private String getValue(PDataType type) {
             return "ARRAY[" + getValue(PDate.INSTANCE) + "]";
         } else if (type instanceof PArrayDataType) {
             return "ARRAY" + type.getSampleValue().toString();
+        } else if (type instanceof PJson) {
+            return null;
```
kadirozde, line 115: Can we return a simple JSON doc instead of null here?

phoenix-core/src/it/java/org/apache/phoenix/end2end/index/PartialIndexIT.java
```diff
@@ -764,4 +768,180 @@ public void testPartialIndexPreferredOverFullIndex() throws Exception {
...
+            conn.createStatement().execute("CREATE " + (uncovered ? "UNCOVERED " : " ") +
+                    (local ? "LOCAL " : " ") +"INDEX "
+                    + indexTableName + " on " + dataTableName + " (CAST(TO_NUMBER(JSON_VALUE(jsoncol, '$.info.age')) AS INTEGER)) " +
+                    (uncovered ? "" : "INCLUDE (B, C, D)") + " WHERE (CAST(TO_NUMBER(JSON_VALUE(jsoncol, '$.info.age')) AS INTEGER)) > 50 ASYNC");
```
kadirozde, line 884: I assume the where clause will include JSON_EXISTS

phoenix-core/src/main/java/org/apache/phoenix/compile/CreateIndexCompiler.java
ranganathg: There was a problem with that - initially i was returning '{"a":1}' but when I was running tests creating the index with the expression, I found that the jsonPath in the expression is applied on the sample json and that ran into issue. Didn't debug it further and rather choose to return null.

phoenix-core/src/it/java/org/apache/phoenix/end2end/index/PartialIndexIT.java
ranganathg: yes yes, i have changes locally which im debugging

phoenix-core/src/main/java/org/apache/phoenix/compile/CreateIndexCompiler.java
ranganathg: Done

gjacoby126 (changes requested):
Haven't had a chance to look at the patch, but wanted to point out that we have Apache License failures on this PR coming from the new JSON files.

ranganathg:
> Haven't had a chance to look at the patch, but wanted to point out that we have Apache License failures on this PR coming from the new JSON files.

Thanks @gjacoby126 . As per my understanding - The [JSON specification](https://www.json.org/json-en.html) does not include comments, and they cannot be used in JSON files.

kadirozde (changes requested):
We should also add JSON specific tests to WhereCompilerTest#testWhereInclusion().

kadirozde:
> > Haven't had a chance to look at the patch, but wanted to point out that we have Apache License failures on this PR coming from the new JSON files.
> 
> Thanks @gjacoby126 . As per my understanding - The [JSON specification](https://www.json.org/json-en.html) does not include comments, and they cannot be used in JSON files.

We can eliminate the json file and use a static map (see BaseTest) or an array instead.

ranganathg:
> We can eliminate the json file and use a static map (see BaseTest) or an array instead.

I would prefer to have json data in separate json files itself. If it's within a Java file, its not readable and not IDE support to modify JSON with java static map wouldn't be clean.
Does that make sense?

ranganathg:
> We should also add JSON specific tests to WhereCompilerTest#testWhereInclusion().

Pushed new changes for this.

phoenix-core/src/it/java/org/apache/phoenix/end2end/index/PartialIndexIT.java
```diff
@@ -764,4 +768,174 @@ public void testPartialIndexPreferredOverFullIndex() throws Exception {
...
+            conn.createStatement().execute("create table " + dataTableName +
+                    " (id varchar not null primary key, " +
+                    "A integer, B integer, C double, D varchar, jsoncol json)"
+                    + (salted ? " SALT_BUCKETS=4" : ""));
+            String indexTableName = generateUniqueName();
+            String partialIndexJson = "json/json_partialindex_tests.json";
+            // Add rows to the data table before creating a partial index to test that the index
+            // will be built correctly by IndexTool
+            conn.createStatement().execute(
+                    "upsert into " + dataTableName + " values ('id1', 70, 2, 3.14, 'a','" + getJsonString(
```
kadirozde, line 872: To improve the readability of the tests, we should include the json doc payload directly in the string for the upsert statement instead of retrieving from a file. The json doc should include only the content necessary for this test. For example, for this test the string for the doc could be  ‘{"info": {"address ": {"exits" : true}}}’. If the same json string is used multiple times then we can define a string constant.

ranganathg: Sure makes sense. Will make these changes

gjacoby126:
@ranganathg - If you keep the JSON, exceptions to the license check can be added here: https://github.com/apache/phoenix/blob/master/pom.xml#L636

ranganathg:
> @ranganathg - If you keep the JSON, exceptions to the license check can be added here: https://github.com/apache/phoenix/blob/master/pom.xml#L636

This is nice. Thank you @gjacoby126

ranganathg:
> It is not easy to read the test when the input is in a non-java file that cannot be accessed with a shortcut or link. If the input is large then this is acceptable but for a tiny input, it is inconvenient.

Took your comment and this made the code much simpler and easy to read. Thank you :)

ranganathg:
> @ranganathg - If you keep the JSON, exceptions to the license check can be added here: https://github.com/apache/phoenix/blob/master/pom.xml#L636

What is the right change here?
<exclude>phoenix-core/src/it/resources/*.json</exclude>
or
<exclude>**/*.json</exclude>

How do I check if this fixes the validation on my local machine before pushing changes?

kadirozde (approved):
+1, Thanks!

PR #1774, opened 2023-12-26
PHOENIX-7160 Change the TSO default port to be compatible with Omid 1.1.1

luocooong:
Since [Omid-247](https://github.com/apache/phoenix-omid/commit/7d3cf3e83586bc523e20277113ecb844172cefc0) has been merged, the default port has changed from 54758 to 24758. The TSO configuration in the Phoenix component also needs to be updated.

### DIFF
--- a/bin/hbase-omid-client-config.yml
+++ b/bin/hbase-omid-client-config.yml
@@ -14,9 +14,10 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-#Omid TSO connection
+# Omid TSO connection
+# The default port from 54758 has changed to 24758 since Omid 1.1.1
 connectionType: !!org.apache.omid.tso.client.OmidClientConfiguration$ConnType DIRECT
-connectionString: "localhost:54758"
+connectionString: "localhost:24758"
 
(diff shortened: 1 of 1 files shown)


luocooong:
Is it necessary to backport to the 5.1 branch?

dbwong (approved):

dbwong:
> Is it necessary to backport to the 5.1 branch?

Is this Omid 1.1.1 in 5.1?

richardantal:
TSO default port change happenes in OMID 1.1.1 as it can be seen in the fix version https://issues.apache.org/jira/browse/OMID-247 
And in phoenix we still use version 1.1.0
https://github.com/apache/phoenix/blob/a90b87deec6a70078f5fdc9b8c564913b170dfc3/pom.xml#L128

I think this change will be needed only after Omid 1.1.1 is released and we use that version in phoenix.

luocooong:
@richardantal Hi, thank you for your confirmation. We do need to wait for the release of Omid 1.1.1, but because Omid does not provide a binary release program, Users are required to compile and test the source code themselves. And compared to the 1.1.0 branch, the master branch has many useful changes, the user has a high chance of selecting the master branch directly to build.

PR #1777, opened 2024-01-02
PHOENIX-7165 getTable should retrieve PTable from server if not cached

kadirozde:


stoty (approved):
+1 LGTM

stoty:
Also, I wonder if we should make this functionalitiy available from the Connection object as a new method.

kadirozde:
> Also, I wonder if we should make this functionalitiy available from the Connection object as a new method.

Ideally yes we should. We tired that before while working on PHOENIX-6761. However, there are tests that expect TableNotFoundException from PhoenixConnection#getTable. So, we gave up on that.

stoty:
> > Also, I wonder if we should make this functionalitiy available from the Connection object as a new method.
> 
> Ideally yes we should. We tried that before while working on [PHOENIX-6761](https://issues.apache.org/jira/browse/PHOENIX-6761). However, there are tests that expect TableNotFoundException from PhoenixConnection#getTable. So, we gave up on that.

What is your opinion on renaming the current getTable to something like getCachedTable or getTableFromCache, and adding a fixed getTable to handles tables not in the cache ?
That would not break the tests (and current code that relies on the bahviour), and would make the method names better match the behaviour ?

kadirozde:
> > > Also, I wonder if we should make this functionalitiy available from the Connection object as a new method.
> > 
> > 
> > Ideally yes we should. We tried that before while working on [PHOENIX-6761](https://issues.apache.org/jira/browse/PHOENIX-6761). However, there are tests that expect TableNotFoundException from PhoenixConnection#getTable. So, we gave up on that.
> 
> What is your opinion on renaming the current getTable to something like getCachedTable or getTableFromCache, and adding a fixed getTable to handles tables not in the cache ? That would not break the tests (and current code that relies on the bahviour), and would make the method names better match the behaviour ?

I will look into unifying get table and related methods and possibly remove them from PhoenixRuntime to simply the code.

dbwong:
I think my questions are on the same lines as @stoty .  Can this just be pushed into Conection::getTable as well?  In general I dislike the spread/increase use of PhoenixRuntime as much as possible.

stoty:
Most of the the checkstyle errors look relevant (like unused imports), it's probably a good idea to fix those (even if some of them are pre-existing problems)

stoty:
BTW I have just committed @Aarchy's fix for the QueryTimeoutIT test failure.

kadirozde:
> Most of the the checkstyle errors look relevant (like unused imports), it's probably a good idea to fix those (even if some of them are pre-existing problems)

I will fix them. Is there a mvn command to fix them or do we fix them manually?

stoty:
I'm not aware of a maven auto-fix option (I tend to use my IDE to clean up the imports automatically).

PR #1780, opened 2024-01-03
PHOENIX-628 Support native JSON data type

ranganathg:
* PHOENIX-628 - Support native JSON data type


stoty:
Can you clean up the commit message ?

Use of colons, dashes, etc is inconsitent.

stoty:
The commit message doesn't match the JIRA ticket description

ranganathg:
> The commit message doesn't match the JIRA ticket description

Can it be different from the JIRA description? I think the description has a recommendation - can I add the message as "Adding JSON native datatype and JSON functions" - That would be more relevant to the comment instead of the JIRA description?

stoty:
The convention is to use the JIRA description for the first line of the commit message.

If you think that the JIRA description is not sufficient, you may want to change the JIRA ticket description, just keep it in sync with the commit message.

virajjasani:
It's good to continue the review, however just wanted to bring up this discuss thread: Discuss thread: https://lists.apache.org/thread/xhwqfwytlklylt6kwqm1botsxr7dd4zb

I am hopeful that in a couple of weeks, we should be well positioned to get this PR move forward, but review can definitely continue in parallel.
Thank you @ranganathg!

phoenix-core-client/src/main/java/org/apache/phoenix/compile/ProjectionCompiler.java
```diff
@@ -497,42 +502,106 @@ public static RowProjector compile(StatementContext context, SelectStatement sta
...
-                arrayOldFuncs.remove(i);
+                serverParsedKVRefs.remove(i);
+                serverParsedKVFuncs.remove(i);
+                serverParsedOldFuncs.remove(i);
             }
         }
 
-        if (arrayKVFuncs.size() > 0 && arrayKVRefs.size() > 0) {
-            serailizeArrayIndexInformationAndSetInScan(context, arrayKVFuncs, arrayKVRefs);
+        if (serverParsedKVFuncs.size() > 0 && serverParsedKVRefs.size() > 0) {
```
kadirozde, line 515: The rest of the code is confusing since JSON functions and ArrayIndex functions are returned in the same output parameter. Can we refactor SelectClauseVisitor and add different getters for ArrayIndex and JSON functions instead of getting them using parameters passed to the constructor?

phoenix-core-client/src/main/java/org/apache/phoenix/compile/ProjectionCompiler.java
```diff
@@ -641,19 +710,32 @@ public <T> T accept(ExpressionVisitor<T> visitor) {
             return null;
         }
     }
-    private static void serailizeArrayIndexInformationAndSetInScan(StatementContext context, List<Expression> arrayKVFuncs,
-            List<KeyValueColumnExpression> arrayKVRefs) {
+
+    static class JsonPathExpression extends ArrayIndexExpression {
```
kadirozde, line 714: This inheritance relationship does not sound right to me. Can we introduce an abstract class both JsonPathExpression and ArrayIndexExpression can extend or something similar?

ranganathg: Actually JsonPath and ArrayIndex expressions in a way look identical like for eg: ARRAY_ELEM(column, index) and JSON_VALUE(column, Path). By identical I don't mean function name and number of params but like operations on a map. one is my index and other is by path(or key). Actually there wasn't really a need to have a JsonPathExpression separately as ArrayIndexExpression has all the logic required - I have added JsonPathExp just to have a code level separation in case we happen to make any modifications in the future we can only modify JsonPath without changing ArrayIndex.
I'm not sure if we need to extend from ScalarFunction for JsonPathExp. It can be from BaseTerminalExpression but then would be a complete copy of ArrayIndexExp

phoenix-core-client/src/main/java/org/apache/phoenix/compile/ProjectionCompiler.java
ranganathg: Sorry I didn't understand : SelectClauseVisitor constructor didn't change - which params passed to the constructor are you referring to?
You mean to say instead of using same ServerParsedkvFunctions and Refs - use different variables for arrays and Json and then have different getters for array index and json expressions?. That can be done but these variables are bit spread around for ref counting and other things.

kadirozde: The replacement map was only for array index functions (old vs new). Now that JSON path function inherits from the array index function, the code gives the impression that the replacement is needed for the JSON path function too. The code has been introduce to handle this replacement even though there is no such replacement is needed for the JSON functions as far as I understand. That is another reason I suggested that JSON path should not inherit from ArrayIndex but from Scalar. 

The refactoring I suggested is to keep track of array index and JSON functions separately and old vs new replacement should be applied to array index only. Maybe I misunderstood the whole thing.

ranganathg: Replacement map is required for JSON functions too. One refactoring I can think of is having a new base class for both ArrayIndexExpression and JSONPathExpression.

phoenix-core-client/src/main/java/org/apache/phoenix/util/json/BsonJsonProvider.java
```diff
@@ -0,0 +1,250 @@
...
+    private static BsonArray toBsonArray(final Object o) {
+        return (BsonArray) o;
+    }
+
+    private static BsonDocument toBsonDocument(final Object o) {
+        return (BsonDocument) o;
+    }
+
+    /**
+     * https://github.com/spring-projects/spring-data-mongodb/blob/main/spring-data-mongodb/src/main/java/org/springframework/data/mongodb/util/BsonUtils.java#L66
```
kadirozde, line 164: What is the purpose of this link here (line 66?)?

ranganathg: This is the reference code for this function implementation. We can actually remove this.

virajjasani:
Re-triggered after fixing build failure on master branch: https://ci-hadoop.apache.org/job/Phoenix/job/Phoenix-PreCommit-GitHub-PR/job/PR-1780/16/

virajjasani (commented):
Instead of using Upsert Select at server side plan, if we can set the JSON_MODIFY expressions (serialized) with Mutation and let the server evaluate the expression by reading the row, that would be great. Otherwise, with UPSERT SELECT plan, we loose the capability to return atomic upsert status code as a result of `statement#executeUpdate`.

If auto-commit is on, we would like client to send the mutations to server side. If the auto-commit is off, we can accummulate mutations and let server perform them in batches?

phoenix-core-client/src/main/java/org/apache/phoenix/util/json/BsonDataFormat.java
```diff
@@ -0,0 +1,201 @@
...
+            }
+        }
+        return null;
+    }
+
+    @Override
+    public ByteBuffer updateValue(Object top, String jsonPathExprStr, String newVal) {
+        Configuration conf = Configuration.builder().jsonProvider(new BsonJsonProvider()).build();
+        BsonValue newValue = JsonPath.using(conf).parse(newVal).json();
+        BsonDocument root = fromRaw((RawBsonDocument) top);
```
virajjasani, line 109: Do we need to use fromRaw() here which uses BsonBinaryReader and then decode RawBsonDocument to BsonDocument?
Can we use this directly? It seems safe type cast.

```
BsonDocument root = ((RawBsonDocument) top).asDocument()
```

phoenix-core-client/src/main/java/org/apache/phoenix/util/json/BsonDataFormat.java
```diff
@@ -0,0 +1,201 @@
...
+                        DecoderContext.builder().build());
+        return updated.getByteBuffer().asNIO();
+    }
+
+    // Ref: https://github.com/json-path/JsonPath/pull/828
+    @Override
+    public boolean isPathValid(Object top, String path) {
+        try{
+            Configuration conf = Configuration.builder().jsonProvider(new BsonJsonProvider()).build();
+            BsonDocument root = fromRaw((RawBsonDocument) top);
```
virajjasani, line 123: Same here, can we just type cast `BsonDocument root = ((RawBsonDocument) top).asDocument()` ?

virajjasani (commented):
RawBsonDocument extends BsonDocument, I am not sure if we need to use BsonBinaryReader just to convert RawBsonDocument to BsonDocument.

phoenix-core-client/src/main/java/org/apache/phoenix/util/json/BsonDataFormat.java
tkhurana: The reason we use RawBSONDocument is because we want to do lazy parsing of bson bytes. Once we have the byte buffer representing json if we call `asDocument` it will parse the entire byte buffer and construct a BsonDocument. But then you have lost the efficiency gains. We only want to parse the byte[] which is queried in the jsonpath expression leaving the rest of the bytes uninterpreted. By uninterpreted I mean we don't parse those bytes  but just read them into memory into the buffer. This is where the performance gains come from. If you parse everything you lose that.

virajjasani: Sounds good.

@ranganathg could you document this as comments for `fromRaw()`? 
`fromRaw()` has only one comment `// Transform to an in memory BsonDocument instance` as of now.

virajjasani: Btw the definition of `asDocument()`

```
  /**
     * Gets this value as a BsonDocument if it is one, otherwise throws exception
     *
     * @return a BsonDocument
     * @throws org.bson.BsonInvalidOperationException if this value is not of the expected type
     */
    public BsonDocument asDocument() {
        throwIfInvalidType(BsonType.DOCUMENT);
        return (BsonDocument) this;
    }
```

virajjasani: The problem here is that, `asDocument()` just does the typecast, so even if we use it, the underlying object is still of Raw type, meaning we will not be able to update the document. Hence, we need to create updatable `BsonDocument` object using `BsonBinaryReader`.

ranganathg: Will take this another iteration

phoenix-core-client/src/main/java/org/apache/phoenix/util/json/BsonDataFormat.java
ranganathg: Will take this another iteration with additional feature

virajjasani (approved):
+1

virajjasani:
Since the last build had flaky, triggered a new build: https://ci-hadoop.apache.org/job/Phoenix/job/Phoenix-PreCommit-GitHub-PR/job/PR-1780/23/

PR #1784, opened 2024-01-08
PHOENIX-7138 Import netty-bom to make sure all netty 4.1 modules shar…

stoty:
…e the same version

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -100,6 +100,7 @@
 
     <!-- Dependency versions -->
     <jackson-bom.version>2.14.1</jackson-bom.version>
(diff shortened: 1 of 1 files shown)


stoty:
I've re-run the failing tests successfully locally.

richardantal (approved):
LGTM
I checked the dependency tree, all netty 4.1 version are 4.1.104.Final
However we have netty jar 3.10.6 coming from com.google.inject.extensions:guice-servlet, but that should be OK

stoty:
Yes, Netty 3 and 4 can coexist, so that's fine (We could probably exclude netty3, but that's a different hill to climb)

stoty:
merged manually

PR #1785, opened 2024-01-09
PHOENIX-7173 Update default HBase versions to 2.4.17 and 2.5.7 respec…

stoty:
…tively

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -84,9 +84,9 @@
     <hbase.profile.list>2.4.0 2.4 2.5.0 2.5</hbase.profile.list>
     <!-- The default hbase versions to build with (override with hbase.version) -->
     <hbase-2.4.0.runtime.version>2.4.0</hbase-2.4.0.runtime.version>
-    <hbase-2.4.runtime.version>2.4.16</hbase-2.4.runtime.version>
+    <hbase-2.4.runtime.version>2.4.17</hbase-2.4.runtime.version>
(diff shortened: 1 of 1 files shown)


virajjasani:
I wonder if this should be part of IT test so that it won't block all ITs from running https://ci-hadoop.apache.org/job/Phoenix/job/Phoenix-PreCommit-GitHub-PR/job/PR-1785/2/testReport/junit/org.apache.phoenix.hbase.index.write.recovery/TestPerRegionIndexWriteCache/testMultipleRegions/

virajjasani:
Otherwise if tests have already been verified locally, +1

stoty:
The problem is that I added an optimization to the Yetus personality to only rebuild HBase once at the start, and Yetus won't rebuild the new HBase version when we change the HBase version.

This wouldn't be very hard to fix, I just haven't gotten around to doing it.
https://issues.apache.org/jira/browse/PHOENIX-6593

stoty:
At least I assume that that is the problem, as the TestPerRegionIndexWriteCache doesn't fail for me locally, and I don't recall seeing it fail before.

virajjasani:
> and I don't recall seeing it fail before.

yes, me neither.

virajjasani (approved):

stoty:
I've run the full test suite successfully locally both for 2.4 and 2.5.

stoty:
merged manually

PR #1787, opened 2024-01-09
PHOENIX-7043 : Split FailoverPhoenixConnectionIT

Divneet18:
FailoverPhoenixConnectionIT takes 20 mins to run. Split it into two files for parallel execution


stoty (changes requested):
The new class name name doesn't make sense to me.
Every failover IT uses a ZK Cluster.

I suggest simply calling the split class FailoverPhoenixConnection2IT , as it is logically the same suite, and is only split for performance. This is the convention the project usually uses.

stoty (approved):
+1 LGTM

stoty:
merged manually

PR #1793, opened 2024-01-11
PHOENIX-7172: Support HBase 2.6

richardantal:
Remove files left from Hbase 2.3 compat module


richardantal:
I removed the HBase 2.3 compat module as it should have been removed in [PHOENIX-6877](https://issues.apache.org/jira/browse/PHOENIX-6877)
And added the new 2.6 modules, git recognises these as renamed some files which is not ideal, maybe it should be done separately.
Also this change can only be merged after the 2.6 release

stoty:
We still have time, I agree that it'd be cleaner if we removed any 2.3 leftovers in a separate ticket.

stoty:
Please update the commit message, @richardantal

pom.xml
```diff
@@ -1724,6 +1752,22 @@
         <hbase.version>${hbase-2.4.runtime.version}</hbase.version>
       </properties>
     </profile>
+    <profile>
```
stoty, line 1755: nit:
Please keep the ascending order, and move this under the 2.5 profiles.

stoty (changes requested):

stoty:
Thank you.
Let's get get back to this once 2.6.0 is out.

stoty (commented):
removing "request changes" state

phoenix-hbase-compat-2.6.0/src/main/java/org/apache/phoenix/compat/hbase/CompatIndexHalfStoreFileReader.java
```diff
@@ -0,0 +1,42 @@
...
+
+import java.io.IOException;
+import java.util.Map;
+import java.util.concurrent.atomic.AtomicInteger;
+
+public class CompatIndexHalfStoreFileReader  extends StoreFileReader {
+
+    public CompatIndexHalfStoreFileReader(final FileSystem fs, final CacheConfig cacheConf, final Configuration conf,
+                                          final ReaderContext readerContext, final HFileInfo hFileInfo) throws IOException {
+        super(readerContext, hFileInfo, cacheConf, new StoreFileInfo(conf, fs, new Path(fs.getHomeDirectory().toUri()),
```
stoty, line 37: This is broken.

You need to pass the Path here, and use it directly.

stoty: This will point to the FS root, and fail.
At least fails on 3.0, and it doesn't ,make sense either.

stoty: You can see how I fixed this in #1815

virajjasani:
Now that HBase 2.6.0 is released, we can resume here?

stoty:
Kicked off a new CI build

stoty:
Can you rebase @richardantal ? The patch doesn't apply cleanly anymore.

stoty:
It is based on an old version, from before we removed 2.4.0. support from HEAD.

stoty:
Also this still uses 2.6.0-SNAPSHOT.
Updating to 2.6.0-hadoop3 should let the CI run fully.

stoty:
Also update the Phoenix version to 5.3.0-SNAPSHOT in the new compatibility module, I've spent an hour trying to compile it.

stoty:
CompatOmidTransactionTable also needs to be updated.

stoty:
I see that hose issues are fixed.

Sorry, I must have used an older patch.

stoty:
However, most of the checkstyle warnings are relevant.

I'm not sure why the javadoc check failed.

stoty (approved):
+1 LGTM.

I have no idea why the javadoc check has failed, but I can't see anything wrong in the patch that could have caused that.

virajjasani:
Does this work fine locally?
`mvn clean javadoc:javadoc -Dskip.code-coverage  -DskipTests=true`

stoty:
Yes, it works after mvn install.
Yetus should have taken care of that.

virajjasani:
Looks like we need to add support to run tests against 2.6 profile https://ci-hadoop.apache.org/job/Phoenix/job/Phoenix-mulitbranch/job/master/688/ ?

stoty:
Yes, opened https://issues.apache.org/jira/browse/PHOENIX-7333 for that.

virajjasani:
Thanks @stoty @richardantal for this great work!

Reg the precommit jenkins builds, i just noticed that 2.6 builds are getting aborted (whereas 2.4 and 2.5 builds get failed with few test failures but at least the builds are not aborted).

stoty:
You mean the multibranch jenkis job, @virajjasani  ?
We all precommit builds are running with 2.5 only at the moment.

virajjasani:
My bad, yes multibranch builds of master and 5.2

stoty:
Opened https://issues.apache.org/jira/browse/PHOENIX-7339

virajjasani:
Thanks a lot @stoty @richardantal for getting HBase 2.6 support in and stabilizing tests.
It seems like one last thing we might have to do is adjust client version for 2.6 profile? https://ci-hadoop.apache.org/job/Phoenix/job/Phoenix-mulitbranch/job/master/703/testReport/junit/org.apache.phoenix.end2end/BackwardCompatibilityIT/MatrixBuild___Matrix___HBASE_PROFILE____2_6____BuildAndTest___initializationError/

PR #1794, opened 2024-01-12
PHOENIX-7015 Implementing CDCGlobalIndexRegionScanner: JSON Response for CDC

TheNamesRai:
Implementing CDCGlobalIndexRegionScanner: JSON Response for CDC

This pull request introduces changes to the server-side CDCGlobalIndexRegionScanner, specifically focusing on capability to form JSON responses for Change Data Capture (CDC)

Screenshot 2024-01-12 at 10 27 55 AM

https://issues.apache.org/jira/browse/PHOENIX-7015

### DIFF
--- a/phoenix-core/src/main/java/org/apache/phoenix/coprocessor/CDCGlobalIndexRegionScanner.java
+++ b/phoenix-core/src/main/java/org/apache/phoenix/coprocessor/CDCGlobalIndexRegionScanner.java
@@ -23,24 +23,23 @@ import org.apache.hadoop.hbase.Cell;
 import org.apache.hadoop.hbase.CellBuilder;
 import org.apache.hadoop.hbase.CellBuilderFactory;
 import org.apache.hadoop.hbase.CellBuilderType;
-import org.apache.hadoop.hbase.CellComparator;
 import org.apache.hadoop.hbase.client.Result;
 import org.apache.hadoop.hbase.client.Scan;
 import org.apache.hadoop.hbase.coprocessor.RegionCoprocessorEnvironment;
 import org.apache.hadoop.hbase.io.ImmutableBytesWritable;
 import org.apache.hadoop.hbase.regionserver.Region;
 import org.apache.hadoop.hbase.regionserver.RegionScanner;
+import org.apache.phoenix.coprocessor.generated.CDCInfoProtos;
 import org.apache.phoenix.execute.TupleProjector;
 import org.apache.phoenix.hbase.index.util.ImmutableBytesPtr;
+import org.apache.phoenix.index.CDCTableInfo;
 import org.apache.phoenix.index.IndexMaintainer;
+import org.apache.phoenix.schema.PColumn;
 import org.apache.phoenix.schema.PTable;
+import org.apache.phoenix.schema.PTableImpl;
 import org.apache.phoenix.schema.SortOrder;
-import org.apache.phoenix.schema.tuple.ResultTuple;
-import org.apache.phoenix.schema.types.PDataType;
 import org.apache.phoenix.schema.types.PLong;
 import org.apache.phoenix.util.CDCUtil;
-import org.apache.phoenix.util.IndexUtil;
-import org.apache.phoenix.util.ScanUtil;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -53,32 +52,20 @@ import java.util.Arrays;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.HashMap;
-import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
-import java.util.stream.Collectors;
 
-import static org.apache.phoenix.coprocessor.BaseScannerRegionObserver.CDC_INCLUDE_SCOPES;
-import static org.apache.phoenix.coprocessor.BaseScannerRegionObserver.CDC_JSON_COL_QUALIFIER;
-import static org.apache.phoenix.coprocessor.BaseScannerRegionObserver.DATA_COL_QUALIFIER_TO_NAME_MAP;
-import static org.apache.phoenix.coprocessor.BaseScannerRegionObserver.DATA_COL_QUALIFIER_TO_TYPE_MAP;
-import static org.apache.phoenix.query.QueryConstants.CHANGE_IMAGE;
-import static org.apache.phoenix.query.QueryConstants.DELETE_EVENT_TYPE;
-import static org.apache.phoenix.query.QueryConstants.EVENT_TYPE;
-import static org.apache.phoenix.query.QueryConstants.POST_IMAGE;
-import static org.apache.phoenix.query.QueryConstants.PRE_IMAGE;
-import static org.apache.phoenix.query.QueryConstants.UPSERT_EVENT_TYPE;
-import static org.apache.phoenix.util.ByteUtil.EMPTY_BYTE_ARRAY;
+import static org.apache.phoenix.coprocessor.BaseScannerRegionObserver.*;
 
 public class CDCGlobalIndexRegionScanner extends UncoveredGlobalIndexRegionScanner {
     private static final Logger LOGGER =
             LoggerFactory.getLogger(CDCGlobalIndexRegionScanner.class);
 
-    private Map<ImmutableBytesPtr, String> dataColQualNameMap;
-    private Map<ImmutableBytesPtr, PDataType> dataColQualTypeMap;
+    private CDCTableInfo cdcDataTableInfo;
     // Map<dataRowKey: Map<TS: Map<qualifier: Cell>>>
     private Set<PTable.CDCChangeScope> cdcChangeScopeSet;
+    //private List<CDCColumnInfo> columnList;
     private Long startTimeRange;
     private Long stopTimeRange;
 
@@ -96,10 +83,8 @@ public class CDCGlobalIndexRegionScanner extends UncoveredGlobalIndexRegionScann
         super(innerScanner, region, scan, env, dataTableScan, tupleProjector, indexMaintainer,
                 viewConstants, ptr, pageSizeMs, queryLimit);
         CDCUtil.initForRawScan(dataTableScan);
-        dataColQualNameMap = ScanUtil.deserializeColumnQualifierToNameMap(
-                scan.getAttribute(DATA_COL_QUALIFIER_TO_NAME_MAP));
-        dataColQualTypeMap = ScanUtil.deserializeColumnQualifierToTypeMap(
-                scan.getAttribute(DATA_COL_QUALIFIER_TO_TYPE_MAP));
+        cdcDataTableInfo = CDCTableInfo.createFromProto(CDCInfoProtos.CDCTableDef
+                .parseFrom(scan.getAttribute(CDC_DATA_TABLE_DEF)));
         Charset utf8Charset = StandardCharsets.UTF_8;
         String cdcChangeScopeStr = utf8Charset.decode(ByteBuffer.wrap(scan.getAttribute(CDC_INCLUDE_SCOPES))).toString();
         cdcChangeScopeSet = CDCUtil.makeChangeScopeEnumsFromString(cdcChangeScopeStr);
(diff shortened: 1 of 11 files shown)


phoenix-core/src/main/java/org/apache/phoenix/coprocessor/CDCGlobalIndexRegionScanner.java
```diff
@@ -104,115 +94,74 @@ protected Scan prepareDataTableScan(Collection<byte[]> dataRowKeys) throws IOExc
...
+            ImmutableBytesPtr dataRowKey = new ImmutableBytesPtr(
+                    indexToDataRowKeyMap.get(indexRowKey));
+            Result dataRow = dataRows.get(dataRowKey);
+            Long indexCellTs = firstCell.getTimestamp();
+            Cell.Type indexCellType = firstCell.getType();
+
+            Map<ImmutableBytesPtr, Cell> preImageObj = new HashMap<>();
+            Map<ImmutableBytesPtr, Cell> changeImageObj = new HashMap<>();
+            List<Cell> resultCells = Arrays.asList(dataRow.rawCells());
+            Collections.sort(resultCells, CellComparator.getInstance().reversed());
```
haridsv, line 110: Cells are already sorted, so we can just reverse the list without needing a comparator.
```suggestion
            Collections.reverse(resultCells);
```

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/CDCGlobalIndexRegionScanner.java
```diff
@@ -104,115 +94,74 @@ protected Scan prepareDataTableScan(Collection<byte[]> dataRowKeys) throws IOExc
...
-                                columns.add(currentColumn);
-                                currentColumn = new LinkedList<>();
-                                currentColumnCell = cell;
-                                currentColumn.add(cell);
-                            } else {
-                                currentColumn.add(cell);
-                            }
+                for (Cell cell : resultCells) {
+                    if (cell.getType() == Cell.Type.DeleteColumn) {
+                        // DDL is not supported in CDC
```
haridsv, line 117: Also, there can be multiple columns set to NULL either in the same or multiple UPSERT statements, so we would have to detect and surface these changes as is.

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/CDCGlobalIndexRegionScanner.java
```diff
@@ -223,4 +172,120 @@ protected boolean getNextCoveredIndexRow(List<Cell> result) throws IOException {
...
+                    }
+                }
+                for (Map.Entry<ImmutableBytesPtr, Cell> preImageObjCell : preImageObj.entrySet()) {
+                    if(dataColQualNameMap.get(preImageObjCell.getKey()) != null
+                            && postImage.get(dataColQualNameMap.get(preImageObjCell.getKey())) == null) {
+                        postImage.put(dataColQualNameMap.get(preImageObjCell.getKey()),
+                                dataColQualTypeMap.get(preImageObjCell.getKey()).toObject(
+                                        preImageObjCell.getValue().getValueArray()));
+                    }
+                }
```
haridsv, line 226: If you flip these two, you can simply clone the `preImage` map and overwrite it with the `changeMap` which will be both simpler and faster.

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/CDCGlobalIndexRegionScanner.java
```diff
@@ -104,115 +94,74 @@ protected Scan prepareDataTableScan(Collection<byte[]> dataRowKeys) throws IOExc
...
-                                currentColumn.add(cell);
-                            }
+                for (Cell cell : resultCells) {
+                    if (cell.getType() == Cell.Type.DeleteColumn) {
+                        // DDL is not supported in CDC
+                        if (cell.getTimestamp() == indexCellTs) {
+                            isIndexCellDeleteColumn = true;
+                            break;
+                        }
+                    } else if (cell.getType() == Cell.Type.Put) {
```
haridsv, line 122: I think we can skip over the Put cells once we are past the indexCellTs, correct?

haridsv (changes requested):
Submitting what I have in the first pass, will have another pass later.

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/CDCGlobalIndexRegionScanner.java
```diff
@@ -104,115 +94,74 @@ protected Scan prepareDataTableScan(Collection<byte[]> dataRowKeys) throws IOExc
...
-                                currentColumn.add(cell);
-                            } else {
-                                currentColumn.add(cell);
-                            }
+                for (Cell cell : resultCells) {
+                    if (cell.getType() == Cell.Type.DeleteColumn) {
+                        // DDL is not supported in CDC
+                        if (cell.getTimestamp() == indexCellTs) {
+                            isIndexCellDeleteColumn = true;
+                            break;
```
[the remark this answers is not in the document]
TheNamesRai: As we just have to ignore all the DeleteColumn cells, even if one of them matches with the timestamp of the current index row cell. We can just break out of this loop.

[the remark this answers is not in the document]
haridsv: What about the handling of setting NULL value as mentioned in the other [comment](https://github.com/apache/phoenix/pull/1794#discussion_r1453565178)?

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/CDCGlobalIndexRegionScanner.java
```diff
@@ -223,4 +172,120 @@ protected boolean getNextCoveredIndexRow(List<Cell> result) throws IOException {
...
+                    if (cell.getType() == Cell.Type.DeleteFamily) {
+                        byte[] indexRowKey = new ImmutableBytesPtr(cell.getRowArray(),
+                                cell.getRowOffset(), cell.getRowLength())
+                                .copyBytesIfNecessary();
+                        ImmutableBytesPtr dataRowKey = new ImmutableBytesPtr(
+                                indexToDataRowKeyMap.get(indexRowKey));
+                        Result dataRow = dataRows.get(dataRowKey);
+                        List<Cell> resultCells = Arrays.asList(dataRow.rawCells());
+                        for (Cell dataRowCell : resultCells) {
+                            if (dataRowCell.getType() == Cell.Type.DeleteFamily && dataRowCell.getTimestamp() == cell.getTimestamp()) {
```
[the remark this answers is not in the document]
TheNamesRai: Yes, that is a correct understanding. Will add the comment as well.

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/CDCGlobalIndexRegionScanner.java
TheNamesRai: result cells are sorted in this order colQual -> timestamp.
For example: all delete markers are at the end of the result cells together. 
So we have to iterate over all result cells.

haridsv: I wasn't suggesting to terminate the loop, just skip over the `Put` cells only once we know that the image won't change.

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/CDCGlobalIndexRegionScanner.java
```diff
@@ -223,4 +186,126 @@ protected boolean getNextCoveredIndexRow(List<Cell> result) throws IOException {
...
+            Map<ImmutableBytesPtr, Cell> preImageObj,
+            Map<ImmutableBytesPtr, Cell> changeImageObj,
+            boolean isIndexCellDeleteRow, Long indexCellTS, Cell firstCell) {
+        Map<String, Object> rowValueMap = new HashMap<>();
+
+        Map<String, Object> preImage = new HashMap<>();
+        if (this.cdcChangeScopeSet.size() == 0
+                || (this.cdcChangeScopeSet.contains(PTable.CDCChangeScope.PRE))) {
+            for (Map.Entry<ImmutableBytesPtr, Cell> preImageObjCell : preImageObj.entrySet()) {
+                if (dataColQualNameMap.get(preImageObjCell.getKey()) != null) {
```
haridsv, line 200:
```suggestion
                if (dataColQualNameMap.containsKey(preImageObjCell.getKey())) {
```

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/CDCGlobalIndexRegionScanner.java
```diff
@@ -223,4 +186,126 @@ protected boolean getNextCoveredIndexRow(List<Cell> result) throws IOException {
...
+            }
+            rowValueMap.put(PRE_IMAGE, preImage);
+        }
+
+        Map<String, Object> changeImage = new HashMap<>();
+        if (this.cdcChangeScopeSet.size() == 0
+                || (this.cdcChangeScopeSet.contains(PTable.CDCChangeScope.CHANGE))) {
+            for (Map.Entry<ImmutableBytesPtr, Cell> changeImageObjCell
+                    : changeImageObj.entrySet()) {
+                if (dataColQualNameMap.get(changeImageObjCell.getKey()) != null) {
```
haridsv, line 214:
```suggestion
                if (dataColQualNameMap.containsKey(changeImageObjCell.getKey())) {
```

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/CDCGlobalIndexRegionScanner.java
```diff
@@ -223,4 +186,126 @@ protected boolean getNextCoveredIndexRow(List<Cell> result) throws IOException {
...
+    @Override
+    protected void scanDataTableRows(long startTime) throws IOException {
+        super.scanDataTableRows(startTime);
+        List<List<Cell>> indexRowList = new ArrayList<>();
+        // Creating new Index Rows for Delete Row events
+        for (int rowIndex = 0; rowIndex < indexRows.size(); rowIndex++) {
+            List<Cell> indexRow = indexRows.get(rowIndex);
+            indexRowList.add(indexRow);
+            if (indexRow.size() > 1) {
+                List<Cell> deleteRow = null;
```
haridsv, line 279: Nit: Outside the loop you are just using this as a flag, so why not use a boolean flag here and make use of a local list at 294?

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/CDCGlobalIndexRegionScanner.java
```diff
@@ -223,4 +186,126 @@ protected boolean getNextCoveredIndexRow(List<Cell> result) throws IOException {
...
+                        for (Cell dataRowCell : dataRow.rawCells()) {
+                            // Note: Upsert adds delete family marker in the index table but not in the datatable.
+                            // Delete operation adds delete family marker in datatable as well as index table.
+                            if (dataRowCell.getType() == Cell.Type.DeleteFamily
+                                    && dataRowCell.getTimestamp() == cell.getTimestamp()) {
+                                if (deleteRow == null) {
+                                    deleteRow = new ArrayList<>();
+                                }
+                                deleteRow.add(cell);
+                                indexRowList.add(deleteRow);
```
haridsv, line 298: If you change the deleteRow to be a boolean flag, this can just be:
```suggestion
                                indexRowList.add(Arrays.asList(cell));
```

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/CDCGlobalIndexRegionScanner.java
```diff
@@ -223,4 +186,126 @@ protected boolean getNextCoveredIndexRow(List<Cell> result) throws IOException {
...
+                }
+                for (Map.Entry<ImmutableBytesPtr, Cell> changeImageObjCell
+                        : changeImageObj.entrySet()) {
+                    if (dataColQualNameMap.get(changeImageObjCell.getKey()) != null) {
+                        postImage.put(dataColQualNameMap.get(changeImageObjCell.getKey()),
+                                dataColQualTypeMap.get(changeImageObjCell.getKey()).toObject(
+                                        changeImageObjCell.getValue().getValueArray()));
+                    }
+                }
+            }
```
haridsv, line 243: Instead of using a separate loop for for building postImage, why not do it as part of both the pre and change image? I mean, you would repeat the lines 201 and 215 for postImage as well.

haridsv:
Closed in favor of PR #1813

PR #1796, opened 2024-01-15
PHOENIX-7181 Do not declare commons-configuration2 dependency

stoty:


stoty:
This runs the UTs fine with both HBase 2.5 and 2.4.

I have also tested this with Hadoop 3.2.5-SNAPSHOT. The UTs ran, and maven dependency:tree shows that it picks up the new commons-configuration2 version from Hadoop.

gjacoby126 (approved):
+1, assuming that asflicense failure is from something else. (Don't see how this patch could be causing one)

stoty:
Thank you.
The Asflicence check often gives false positives, as it sometimes finds and flags generated files.
We could probably improve that test in Yetus. (Or maybe a Yetus upgrade would solve that problem)

stoty:
merged manually

PR #1797, opened 2024-01-16
PHOENIX-7171 Update Zookeeper to 3.8.3 when building with HBase 2.4+

stoty:


stoty:
This is the same as #1797 which has gone into some error state that I cannot recover it from.

gjacoby126 (approved):
+1

stoty:
merged manually

PR #1798, opened 2024-01-16
PHOENIX-7176 QueryTimeoutIT#testQueryTimeout fails with incorrect err…

Aarchy:
…or message (addendum: modify & add message to assert)


Aarchy:
The test still fails sometimes, but on another assert, which checks whether the elapsed time was greater than the timeout set. 
Logging elapsedTime showed that it was equal to the timeout (1000 ms). 

I'm wondering if this is caused by relying on System.currentTimeMillis() instead of system.nanotime(). Using (>=) in the assert seems to fix it though.

stoty (approved):
+1 LGTM

stoty:
merged manually

PR #1799, opened 2024-01-16
PHOENIX-7108 Provide support for pruning expired rows of views using Phoenix level compactions

jpisaac:


phoenix-core/src/main/java/org/apache/phoenix/coprocessor/CompactionScanner.java
```diff
@@ -160,17 +241,312 @@ public void close() throws IOException {
...
+                // case multi-tenant + index-table => startingPKPosition = 0
+                // case salted + index-table => startingPKPosition = 1
+                // case salted => startingPKPosition = 1
+                // case index-table => startingPKPosition = 0
+                // case multi-tenant => startingPKPosition = 1, 0
+                int offset = pkPositions.get(pkPosition);
+                byte[] rowKey = CellUtil.cloneRow(firstCell);
+                // Search using the starting offset (startingPKPosition offset)
+                Integer tableId = index.getTableIdWithPrefix(rowKey, offset);
+                if ((tableId == null) && (!isIndexTable) && (isMultiTenant || isSalted)) {
```
tkhurana, line 432: @jpisaac What if my partition key is a prefix of tenantid ? I define a ttl on global view and a ttl on a tenant view created on a different global view. Basically there is one ttl on "KP1" and another on "KP1Tenant#KP2"

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/CompactionScanner.java
```diff
@@ -160,17 +241,312 @@ public void close() throws IOException {
...
+                // case multi-tenant + salted => startingPKPosition = 2, 1
+                // case multi-tenant + index-table => startingPKPosition = 0
+                // case salted + index-table => startingPKPosition = 1
+                // case salted => startingPKPosition = 1
+                // case index-table => startingPKPosition = 0
+                // case multi-tenant => startingPKPosition = 1, 0
+                int offset = pkPositions.get(pkPosition);
+                byte[] rowKey = CellUtil.cloneRow(firstCell);
+                // Search using the starting offset (startingPKPosition offset)
+                Integer tableId = index.getTableIdWithPrefix(rowKey, offset);
```
tkhurana, line 431: @jpisaac What if my partition key is a prefix of tenantid ? I define a ttl on global view and a ttl on a tenant view created on a different global view. Basically there is one ttl on "KP1" and another on "KP1Tenant#KP2"

jpisaac: This is an interesting case of overlap between the global space and tenant space,  Currently KP1Tenant#KP2 row will match the global keyprefix "KP1".  The fix will be to keep the two prefix indexes (global and tenant) separate.

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/CompactionScanner.java
```diff
@@ -160,17 +240,390 @@ public void close() throws IOException {
...
+    }
+    private class NonPartitionedTableTTLTracker implements TTLTracker {
+
+        private long ttl;
+        private RowContext rowContext;
+
+        public NonPartitionedTableTTLTracker(PTable pTable, RegionCoprocessorEnvironment env, Store store) {
+            boolean isSystemTable = pTable.getType() == PTableType.SYSTEM;
+            if (isSystemTable) {
+                ColumnFamilyDescriptor cfd = store.getColumnFamilyDescriptor();
```
kadirozde, line 259: Can we have a config param, say, phoenix.system.catalog.ttl, to specify the ttl for the system table? We want to eliminate HBase TTL for all tables.

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/CompactionScanner.java
```diff
@@ -84,45 +110,99 @@ public CompactionScanner(RegionCoprocessorEnvironment env,
...
+            // If there is atleast one child view for this table then it is a partitioned table.
+            boolean isPartitioned = ViewUtil.hasChildViews(
+                    childLinkHTable,
+                    EMPTY_BYTE_ARRAY,
+                    Bytes.toBytes(schemaName),
+                    Bytes.toBytes(tableName),
+                    currentTime);
+
+            return isPartitioned ?
+                    new PartitionedTableTTLTracker(baseTable, env, isSalted, isPartitionedIndexTable) :
```
kadirozde, line 200: View indexes are partitioned table but they do not have any children in the child link table. This code does not handle them properly, does it?

jpisaac: >> View indexes are partitioned table but they do not have any children in the child link table
Correct!, but if there are no views defined in SYSCAT for this table, we do not expect any rows to be in them.

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/CompactionScanner.java
```diff
@@ -84,45 +110,99 @@ public CompactionScanner(RegionCoprocessorEnvironment env,
...
+     * @param env
+     * @param store
+     * @param baseTable
+     * @return
+     */
+    private TTLTracker createTTLTrackerFor(RegionCoprocessorEnvironment env,
+            Store store,  PTable baseTable) throws IOException {
+
+        long currentTime = EnvironmentEdgeManager.currentTimeMillis();
+        String compactionTableName = env.getRegion().getRegionInfo().getTable().getNameAsString();
```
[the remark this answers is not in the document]
jpisaac: The compactionTableName (local variable) from regionInfo is used only for logging purposes. 
This code was not modified in this PR,  the only other place it is used is for the maxLookbackMap.

PR #1803, opened 2024-01-23
PHOENIX-7187 Improvement of Integration test case with Explain Plan for Partial Index

nikita15p:
Currently in the integration test for partial index, partial index table obtained from result set is compared against expected schema and table name. But there is no test case to validate explain plan for partial index.

This PR adds query plan test case for partial index.

### DIFF
--- a/phoenix-core/src/it/java/org/apache/phoenix/end2end/index/PartialIndexIT.java
+++ b/phoenix-core/src/it/java/org/apache/phoenix/end2end/index/PartialIndexIT.java
@@ -228,11 +228,19 @@ public class PartialIndexIT extends BaseTest {
             assertTrue(rs.next());
             assertEquals("b", rs.getString(1));
             assertFalse(rs.next());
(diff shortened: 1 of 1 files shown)


phoenix-core/src/it/java/org/apache/phoenix/end2end/index/PartialIndexIT.java
```diff
@@ -764,4 +764,47 @@ public void testPartialIndexPreferredOverFullIndex() throws Exception {
             assertFalse(rs.next());
         }
     }
+
```
tkhurana, line 767: My recommendation would be to modify some of the existing tests in this class and add this additional check there rather than creating a new test.

nikita15p: @tkhurana sure. Let me look into it.

tkhurana (approved):

PR #1805, opened 2024-01-25
PHOENIX-7064 Improve prepare of local index mutations

fan1emon:
This commit try to improve the performance of prepare local index.

The prepare logic is similar to global index and may be combined with global index.

IT needs to be added


fan1emon:
May I ask how to trigger a jenkins test

stoty:
The Jenkins tests are triggered automatically.

PR #1807, opened 2024-01-27
PHOENIX-7190 : Store LAST_DDL_TIMESTAMP of all ancestors in a View's PTable

palashc:
- Add a map from `PTableKey` to `LAST_DDL_TIMESTAMP` in `PTableImpl`.
- This map is populated recursively during the `updateCache` workflow, similar to `addColumnsAndIndexesFromAncestors`.
- During a query on a view, use timestamps from this map for validation.


shahrs87:
https://ci-hadoop.apache.org/job/Phoenix/job/Phoenix-PreCommit-GitHub-PR/job/PR-1807/1/testReport/
@palashc  I don't think this build ran all the tests. It only ran 2097 test. IIRC there are around 8000 tests.

phoenix-core-client/src/main/java/org/apache/phoenix/util/ValidateLastDDLTimestampUtil.java
```diff
@@ -143,14 +143,17 @@ public static void validateLastDDLTimestamp(
...
+            // view -> parent view and its ancestors
+            // view index -> view and its ancestors
+            for (Map.Entry<PTableKey, Long> entry :
+                    tableRef.getTable().getAncestorLastDDLTimestampMap().entrySet()) {
                 innerBuilder = RegionServerEndpointProtos.LastDDLTimestampRequest.newBuilder();
-                PTableKey key = new PTableKey(conn.getTenantId(),
-                        tableRef.getTable().getParentName().getString());
-                PTable parentTable = conn.getTable(key);
-                setLastDDLTimestampRequestParameters(innerBuilder, conn.getTenantId(), parentTable);
+                PTable ancestorTable = conn.getTable(entry.getKey());
```
shahrs87, line 153: Do we still need to call this method `conn.getTable(entry.getKey())` ? 
We have all the things required to construct `LastDDLTimestampRequest` object in `table. getAncestorLastDDLTimestampMap` map.
The only downside is getAncestorLastDDLTimestampMap contains PTableKey which only has tenantID and name. Here the name is fully qualified name which is the combination of schema name and tableName.
Whenever we want to call validateLastDDLTimestamp RPC, we need to split the fully qualified name to schema name and tableName which can increase pressure on memory. But definitely we don't want to call conn.getTable method.

shahrs87 (changes requested):

phoenix-core-client/src/main/java/org/apache/phoenix/util/ValidateLastDDLTimestampUtil.java
palashc: Yeah I was thinking if we can store both table name and schema name when creating a PTableKey?

phoenix-core-client/src/main/java/org/apache/phoenix/schema/PMetaDataImpl.java
```diff
@@ -162,7 +162,11 @@ public void addTable(PTable table, long resolvedTime) throws SQLException {
             metaData.put(table.getKey(), tableRef);
         }
         for (PTable index : table.getIndexes()) {
-            metaData.put(index.getKey(), tableRefFactory.makePTableRef(index, this.timeKeeper.getCurrentTime(), resolvedTime));
+            PTable indexPTable = index;
```
shahrs87, line 165: We add an index to the cache at L160 and L162 above. 
There is a specific check whether the table is an index or not at L133 above. 
If the table  is an index, then codepath between L133-L156 is executed where we will miss adding the parent link to the cache.

phoenix-core-client/src/main/java/org/apache/phoenix/schema/MetaDataClient.java
```diff
@@ -898,9 +898,9 @@ public MetaDataMutationResult updateCache(PName tenantId, List<String> functionN
      * @return true if the PTable contained by result was modified and false otherwise
      * @throws SQLException if the physical table cannot be found
      */
-    private boolean addColumnsAndIndexesFromAncestors(MetaDataMutationResult result, Long resolvedTimestamp,
-                                                      boolean alwaysAddAncestorColumnsAndIndexes,
-                                                      boolean alwaysHitServerForAncestors)
+    private boolean addColumnsIndexesAndLastDDLTimestampsFromAncestors(MetaDataMutationResult result, Long resolvedTimestamp,
```
tkhurana, line 901: I believe we should refactor this function a little so that it works for views, view indexes and regular indexes. This way there is only one place in the code where we are adding information from the ancestors. For view/view index we add columns, indexes, lastddltimestamp. For indexes we just add lastddltimestamp.

tkhurana: We could rename this function to something like `addMetadataInformationFromAncestors` and then add a javadoc explaining what metdata we need from ancestors depending upon the table type.

phoenix-core-client/src/main/java/org/apache/phoenix/schema/MetaDataClient.java
```diff
@@ -940,8 +940,13 @@ private boolean addColumnsAndIndexesFromAncestors(MetaDataMutationResult result,
                 if (!alwaysAddAncestorColumnsAndIndexes && !result.wasUpdated() && !parentResult.wasUpdated()) {
                     return false;
                 }
-                result.setTable(ViewUtil.addDerivedColumnsAndIndexesFromParent(
-                        connection, table, parentTable));
+
+                PTable pTableWithDerivedColumnsAndIndexes
```
tkhurana, line 944: You can remove the hasIndexId check from line 908 and use that here something like `if (!index || hasIndexId) call ViewUtil.addDerivedColumnsAndIndexesFromParent`

phoenix-core-client/src/main/java/org/apache/phoenix/schema/MetaDataClient.java
palashc: > This way there is only one place in the code where we are adding information from the ancestors.
@tkhurana What if updateCache is not called for the index table directly? If the client resolves just the base table for a query, this method will not execute for the index, right? One thing we could do is to create a new PTable in the case of tables with a new list of indexes inside it which have the ancestor map.

tkhurana: How does it handle view indexes in that case @palashc ? I am assuming if this method is called for all types of index.

palashc: Good point. I just verified with a test that a new client will not be able to populate the map for both indexes and view indexes. I created a table, view and view-index using client-1. Then I queried the view using client-2. In client2's cache, view index is not populated with the map.

palashc: > I am assuming this method is called for all types of index.

Yes, but I think it is called only when the index is directly being used in the plan. Thus, we might have to populate the map for indexes when we add an index to the cache in `PMetaDataImpl.addTable`

phoenix-core-client/src/main/java/org/apache/phoenix/schema/MetaDataClient.java
palashc: Made this change, it was required to make sure the parent of an index table was present in cache so that we can add the ancestor map to the index.

phoenix-core-client/src/main/java/org/apache/phoenix/schema/PMetaDataImpl.java
```diff
@@ -135,6 +135,8 @@ public void addTable(PTable table, long resolvedTime) throws SQLException {
             PTableRef oldParentRef = metaData.get(new PTableKey(table.getTenantId(), parentName));
             // If parentTable isn't cached, that's ok we can skip this
             if (oldParentRef != null) {
+                table = MetaDataUtil.getPTableWithAncestorLastDDLTimestampMap(table, oldParentRef.getTable());
```
palashc, line 138: not required

phoenix-core-client/src/main/java/org/apache/phoenix/schema/MetaDataClient.java
```diff
@@ -5163,8 +5193,8 @@ private void addTableToCache(MetaDataMutationResult result, boolean alwaysHitSer
 
     private void addTableToCache(MetaDataMutationResult result, boolean alwaysHitServerForAncestors,
                                  long timestamp) throws SQLException {
-        addColumnsAndIndexesFromAncestors(result, null, false, alwaysHitServerForAncestors);
-        PTable table = result.getTable();
+        addColumnsIndexesAndLastDDLTimestampsFromAncestors(result, null, false, alwaysHitServerForAncestors);
+        PTable table = updateIndexesWithAncestorMap(result.getTable());
```
tkhurana, line 5197: The PTable in result is now different from the table added to cache

phoenix-core-client/src/main/java/org/apache/phoenix/schema/MetaDataClient.java
```diff
@@ -716,9 +716,10 @@ public MetaDataMutationResult updateCache(PName origTenantId, String schemaName,
                             // In this case, we update the parent table which may in turn pull
                             // in indexes to add to this table.
                             long resolvedTime = TransactionUtil.getResolvedTime(connection, result);
-                            if (addColumnsAndIndexesFromAncestors(result, resolvedTimestamp,
+                            if (addColumnsIndexesAndLastDDLTimestampsFromAncestors(result, resolvedTimestamp,
                                     true, false)) {
-                                connection.addTable(result.getTable(), resolvedTime);
+                                connection.addTable(updateIndexesWithAncestorMap(result.getTable()),
```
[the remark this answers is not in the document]
palashc: Oh right. Missed that!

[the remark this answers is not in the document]
tkhurana: The Ptable in result is different from the Ptable in cache

phoenix-core-client/src/main/java/org/apache/phoenix/schema/PMetaDataImpl.java
```diff
@@ -20,10 +20,10 @@
 import static org.apache.phoenix.schema.PTableImpl.getColumnsToClone;
 
 import java.sql.SQLException;
+import java.util.ArrayList;
```
tkhurana, line 23: Unused imports

phoenix-core-client/src/main/java/org/apache/phoenix/schema/PMetaDataImpl.java
```diff
@@ -33,6 +33,7 @@
 import org.apache.phoenix.query.QueryConstants;
 import org.apache.phoenix.query.QueryServices;
 import org.apache.phoenix.query.QueryServicesOptions;
+import org.apache.phoenix.util.MetaDataUtil;
```
tkhurana, line 36: Unused imports ?

phoenix-core-client/src/main/java/org/apache/phoenix/util/MetaDataUtil.java
```diff
@@ -1182,4 +1184,24 @@ public static void deleteFromStatsTable(PhoenixConnection connection,
             connection.setAutoCommit(isAutoCommit);
         }
     }
+
```
tkhurana, line 1187: Maybe we can move this function to metadataclient class only since that is where everything is being used

tkhurana (approved):

palashc:
@tkhurana @shahrs87 Can we merge if this looks good now? 
Latest Test Report - https://ci-hadoop.apache.org/job/Phoenix/job/Phoenix-PreCommit-GitHub-PR/view/change-requests/job/PR-1807/15/testReport/

PR #1808, opened 2024-01-30
PHOENIX-7081 : Replace /tmp with {java.io.tmpdir} in tests

Divneet18:


phoenix-core/src/it/java/org/apache/phoenix/end2end/CreateTableIT.java
```diff
@@ -1557,7 +1558,7 @@ public void testCoprocessorsForCreateIndexOnOldImplementation() throws Exception
             // Now roll back to the old indexing
             IndexUpgradeTool iut =
                     new IndexUpgradeTool(ROLLBACK_OP, tableName, null,
-                            "/tmp/index_upgrade_" + UUID.randomUUID().toString(), false, null,
+                            filePath + "/index_upgrade_" + UUID.randomUUID().toString(), false, null,
```
stoty, line 1561: In the Hadoop Tools, path parameters are some times HDFS paths, so that is a case where simple "/tmp" is the right solution.
Those are also separate HDFS file systems, so there is no danger of coliision.

Have you checked that all of these instances are used as local paths, as opposed to HadoopFS paths ?

(I have checked this one, and this is correct, as this is a local FS path)

stoty (changes requested):

phoenix-core/src/it/java/org/apache/phoenix/end2end/CsvBulkLoadToolIT.java
```diff
@@ -62,7 +62,7 @@ public class CsvBulkLoadToolIT extends BaseOwnClusterIT {
 
     private static Connection conn;
     private static String zkQuorum;
-    private static final String filePath = System.getProperty("java.io.tmpdir");
```
stoty: This is a HDFS path. not a local path.
This should NOT use java.io.tmpdir.

Please check each changed test  to make sure that only local paths use java.io.tmpdir, and not HadoopFS paths.

stoty (changes requested):

phoenix-core-server/src/main/java/org/apache/phoenix/mapreduce/index/IndexUpgradeTool.java
```diff
@@ -739,7 +740,7 @@ private int startIndexRebuilds(HashMap<String, IndexInfo> indexInfos,
             String tenantId = indexInfo.getTenantId();
             String baseTable = indexInfo.getBaseTable();
             String schema = indexInfo.getSchemaName();
-            String outFile = "/tmp/index_rebuild_" +schema+"_"+ indexName +
+            String outFile = tmpPath + "/index_rebuild_" +schema+"_"+ indexName +
```
stoty, line 743: This one eventually becomes an HadoopFS path.

stoty (changes requested):
Please fix IndexUpgradeTool

phoenix-core-server/src/main/java/org/apache/phoenix/mapreduce/AbstractBulkLoadTool.java
```diff
@@ -249,7 +250,7 @@ private int loadData(Configuration conf, CommandLine cmdLine) throws Exception {
         if (cmdLine.hasOption(OUTPUT_PATH_OPT.getOpt())) {
             outputPath = new Path(cmdLine.getOptionValue(OUTPUT_PATH_OPT.getOpt()));
         } else {
-            outputPath = new Path("/tmp/" + UUID.randomUUID());
+            outputPath = new Path(tmpPath + "/" + UUID.randomUUID());
```
stoty, line 253: Please check EACH PATH manually.
This is another obvious HadoopFS path, the type is org.apache.hadoop.fs.Path. It is defined just a few lines above.

stoty (changes requested):
Please verify  EACH path instance, as I wrote above.

stoty (changes requested):
Please revert the files that only have whitespace changes, so that it is immediately apparent which files were changed.

phoenix-core/src/it/java/org/apache/phoenix/end2end/IndexToolIT.java
```diff
@@ -167,9 +167,9 @@ public static synchronized void setup(
             //Maybe we're gonna have to refactor the namespace
             if(clusterInitialized) {
                 tearDownMiniCluster(0);
-                System.setProperty("java.io.tmpdir", tmpDir);
+                System.setProperty("java.io.tmpdir", tmpPath);
```
stoty, line 170: The old variable name is better here.
This is not used a base for paths, it's only used for saving the value of the system property.

stoty: and the old variable name reflects this better.

phoenix-core/src/it/java/org/apache/phoenix/end2end/ParameterizedIndexUpgradeToolIT.java
```diff
@@ -115,7 +114,7 @@ public class ParameterizedIndexUpgradeToolIT extends BaseTest {
 
     @BeforeClass
     public static synchronized void saveTmp () throws Exception {
-        tmpDir = System.getProperty("java.io.tmpdir");
+        tmpPath = System.getProperty("java.io.tmpdir");
```
stoty, line 117: nit: I find it better to use tmpDir here, too.

stoty (approved):
+1 LGTM.
I will address my nits on commit.

stoty (changes requested):
Come to think of it, this is initaiizing the path variables too early.
MiniCluster DOES change "java.io.tmpdir" , and there is no guarantee that the miniCluster used for the tests is already running when the test is loaded.

We should take tmpdir value from after the miniCluster initialization, i.e. either
* in a @Before method
* At the end of @BeforeClass (for @NeedsOwnCluster tests)
* or in the test itself.

phoenix-core/src/it/java/org/apache/phoenix/end2end/ParameterizedIndexUpgradeToolIT.java
```diff
@@ -87,9 +87,9 @@ public class ParameterizedIndexUpgradeToolIT extends BaseTest {
     private static final String [] TABLE_LIST_NAMESPACE_SIMPLIFIED = new String[1];
 
     private static final String [] TRANSACTIONAL_TABLE_LIST = new String[1];
-
+    private static String tmpPath = System.getProperty("java.io.tmpdir");
```
stoty, line 90: This is supposed to be value of tmpdir after minicluster has been started.
(But it is not, see my previous commenft()

phoenix-core/src/it/java/org/apache/phoenix/end2end/ParameterizedIndexUpgradeToolIT.java
stoty, line 117: This is the value of tmpDir before starting the minicluster.
It should only be used for restoring its value, and not be used in tests.

Divneet18: could you elaborate on this please since all i did here was change the variable name?

phoenix-core/src/it/java/org/apache/phoenix/end2end/ParameterizedIndexUpgradeToolIT.java
```diff
@@ -87,9 +87,9 @@ public class ParameterizedIndexUpgradeToolIT extends BaseTest {
...
     private static String INPUT_LIST = "";
-    private static final String INPUT_FILE = "/tmp/input_file_index_upgrade.csv";
+    private static final String INPUT_FILE = tmpPath + "/input_file_index_upgrade.csv";
```
stoty, line 92: This is where you use the the saved tmpdir.

You should instead use tmpPath which is read AFTER minicluster is started, which can be a different value (depending onm HBase / Hadoop version)

phoenix-core/src/it/java/org/apache/phoenix/end2end/ParameterizedIndexUpgradeToolIT.java
stoty: See the comment above.
the tmpdir values originally set, and re-set by minicluster must be kept separate.

stoty (changes requested):

stoty:
As there was no activity here, I fixed the remaining issues in #1992

PR #1810, opened 2024-01-31
PHOENIX-7193 Fix cluster override for mapreduce jobs for non-ZK registries

stoty:


wendigo:
@stoty sorry for unrelated question, but is the new release happening anytime soon?

In trinodb/trino we are considering dropping Phoenix 5 support because of the high number of critical CVEs included in the shaded client jar.

chrajeshbabu (approved):

stoty:
Yes, it is, @wendigo.
I suggest starting a thread with your concerns on the phoenix mailing list, metioning some specific dependencies.
That would be a better forum.

stoty:
merged manually.

PR #1813, opened 2024-02-01
PHOENIX-7015 Implementing CDCGlobalIndexRegionScanner

TheNamesRai:
Implementing CDCGlobalIndexRegionScanner: JSON Response for CDC

This pull request introduces changes to the server-side CDCGlobalIndexRegionScanner, specifically focusing on capability to form JSON responses for Change Data Capture (CDC)

Screenshot 2024-01-12 at 10 27 55 AM

https://issues.apache.org/jira/browse/PHOENIX-7015


phoenix-core/src/main/java/org/apache/phoenix/coprocessor/CDCGlobalIndexRegionScanner.java
```diff
@@ -90,128 +88,144 @@ public CDCGlobalIndexRegionScanner(final RegionScanner innerScanner,
...
+                                preImageObj.get(columnFamily).put(columnQualifier,
+                                        this.getColumnValue(cell, cdcColumnInfoList
+                                                .get(columnListIndex).getColumnType()));
+                            } else if (cell.getTimestamp() == indexCellTS) {
+                                if (!changeImageObj.containsKey(columnFamily)) {
+                                    changeImageObj.put(columnFamily, new HashMap<>());
+                                }
+                                changeImageObj.get(columnFamily).put(columnQualifier,
+                                        this.getColumnValue(cell, cdcColumnInfoList
+                                                .get(columnListIndex).getColumnType()));
```
haridsv, line 207: Also, we could avoid building change image if only PRE is selected.

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/CDCGlobalIndexRegionScanner.java
```diff
@@ -223,4 +237,85 @@ protected boolean getNextCoveredIndexRow(List<Cell> result) throws IOException {
...
+                || (this.cdcChangeScopeSet.contains(PTable.CDCChangeScope.POST))) {
+            if (!isIndexCellDeleteRow) {
+                for (Map.Entry<String, Map<String, Object>> preImageObjFamily
+                        : preImageObj.entrySet()) {
+                    String columnFamily = preImageObjFamily.getKey();
+                    postImageObj.put(columnFamily, new HashMap<>());
+                    for (Map.Entry<String, Object> preImageColQual :
+                            preImageObjFamily.getValue().entrySet()) {
+                        postImageObj.get(columnFamily).put(preImageColQual.getKey(),
+                                preImageColQual.getValue());
```
haridsv, line 268: We should be able to just clone the original map.

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/CDCGlobalIndexRegionScanner.java
```diff
@@ -223,4 +237,85 @@ protected boolean getNextCoveredIndexRow(List<Cell> result) throws IOException {
...
+                for (Map.Entry<String, Map<String, Object>> changeImageObjFamily
+                        : changeImageObj.entrySet()) {
+                    String columnFamily = changeImageObjFamily.getKey();
+                    if (!postImageObj.containsKey(columnFamily)) {
+                        postImageObj.put(columnFamily, new HashMap<>());
+                    }
+                    for (Map.Entry<String, Object> changeImageColQual :
+                            changeImageObjFamily.getValue().entrySet()) {
+                        postImageObj.get(columnFamily).put(changeImageColQual.getKey(),
+                                changeImageColQual.getValue());
```
haridsv, line 280: Again, clone the original map and use putAll.

phoenix-core/src/main/java/org/apache/phoenix/query/QueryConstants.java
```diff
@@ -308,6 +308,14 @@ enum JoinType {INNER, LEFT_OUTER}
 
     String CDC_JSON_COL_NAME = "CDC JSON";
 
+    String EVENT_TYPE = "event_type";
+    String PRE_IMAGE = "pre_image";
+    String POST_IMAGE = "post_image";
+    String CHANGE_IMAGE = "change_image";
+    String UPSERT_EVENT_TYPE = "upsert";
+    String DELETE_EVENT_TYPE = "delete";
```
haridsv, line 316: I think we should put CDC_ prerfix on all of these keys.

haridsv (commented):
We need ample comments in the core algorithm.

phoenix-core/src/main/java/org/apache/phoenix/util/CDCUtil.java
```diff
@@ -110,4 +109,25 @@ public static Scan initForRawScan(Scan scan) {
...
+
+    public static int compareCellFamilyAndQualifier(byte[] columnFamily1,
+                                                     byte[] columnQual1,
+                                                     byte[] columnFamily2,
+                                                     byte[] columnQual2) {
+        int familyNameComparison = CDCUtil.compare(columnFamily1, columnFamily2);
+        if (familyNameComparison != 0) {
+            return familyNameComparison;
+        }
+        return CDCUtil.compare(columnQual1, columnQual2);
```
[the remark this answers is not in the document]
TheNamesRai: Arrays.compare is available after Java 9. To keep supporting Java 8, i created a new function to do the same.
As per it's implementation, both seems equally efficient.

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/CDCGlobalIndexRegionScanner.java
```diff
@@ -90,128 +88,144 @@ public CDCGlobalIndexRegionScanner(final RegionScanner innerScanner,
...
+        try {
+            cdcChangeScopeSet = CDCUtil.makeChangeScopeEnumsFromString(cdcChangeScopeStr);
+        } catch (SQLException e) {
+            throw new RuntimeException(e);
+        }
     }
 
     @Override
     protected Scan prepareDataTableScan(Collection<byte[]> dataRowKeys) throws IOException {
+        //TODO: Get Timerange from the start row and end row of the index scan object
```
kadirozde, line 105: Can we finalize this TODO?

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/CDCGlobalIndexRegionScanner.java
```diff
@@ -90,128 +88,144 @@ public CDCGlobalIndexRegionScanner(final RegionScanner innerScanner,
...
     protected boolean getNextCoveredIndexRow(List<Cell> result) throws IOException {
         if (indexRowIterator.hasNext()) {
             List<Cell> indexRow = indexRowIterator.next();
-            for (Cell c: indexRow) {
-                if (c.getType() == Cell.Type.Put) {
-                    result.add(c);
-                }
-            }
+            Cell firstCell = indexRow.get(indexRow.size() - 1);
+            byte[] indexRowKey = new ImmutableBytesPtr(firstCell.getRowArray(),
```
TheNamesRai, line 122: Use CellUtil.cloneRow

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/CDCGlobalIndexRegionScanner.java
```diff
@@ -90,128 +88,144 @@ public CDCGlobalIndexRegionScanner(final RegionScanner innerScanner,
...
+                    firstCell.getRowOffset(), firstCell.getRowLength())
+                    .copyBytesIfNecessary();
+            ImmutableBytesPtr dataRowKey = new ImmutableBytesPtr(
+                    indexToDataRowKeyMap.get(indexRowKey));
+            Result dataRow = dataRows.get(dataRowKey);
+            Long indexCellTS = firstCell.getTimestamp();
+            Map<String, Map<String, Object>> preImageObj = new HashMap<>();
```
TheNamesRai, line 129: Create these object conditionally based on cdc include scope

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/CDCGlobalIndexRegionScanner.java
```diff
@@ -90,128 +88,144 @@ public CDCGlobalIndexRegionScanner(final RegionScanner innerScanner,
...
+            Map<String, Map<String, Object>> changeImageObj = new HashMap<>();
+            Long lowerBoundForPreImage = 0L;
```
TheNamesRai, line 131: Rename - lowerBoundTsForPreImage

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/CDCGlobalIndexRegionScanner.java
```diff
@@ -90,128 +88,144 @@ public CDCGlobalIndexRegionScanner(final RegionScanner innerScanner,
...
+            boolean isIndexCellDeleteRow = false;
+            byte[] emptyCQ = indexMaintainer.getEmptyKeyValueQualifier();
```
TheNamesRai, line 133: Verify this method for datatable empty key qualifier

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/CDCGlobalIndexRegionScanner.java
```diff
@@ -90,128 +88,144 @@ public CDCGlobalIndexRegionScanner(final RegionScanner innerScanner,
...
+                        this.cdcDataTableInfo.getColumnInfoList();
+                for (Cell cell : dataRow.rawCells()) {
+                    if (cell.getType() == Cell.Type.DeleteFamily) {
+                        if (columnListIndex > 0) {
+                            continue;
+                        }
+                        if (indexCellTS == cell.getTimestamp()) {
+                            isIndexCellDeleteRow = true;
+                        } else if (indexCellTS > cell.getTimestamp()) {
+                            lowerBoundForPreImage = cell.getTimestamp();
```
TheNamesRai, line 146: Do it once.

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/CDCGlobalIndexRegionScanner.java
```diff
@@ -90,128 +88,144 @@ public CDCGlobalIndexRegionScanner(final RegionScanner innerScanner,
...
+                                        .getColumnQualifier()) == 0) {
+                            String columnFamily = StandardCharsets.UTF_8
+                                    .decode(ByteBuffer.wrap(cdcColumnInfoList.get(columnListIndex)
+                                            .getColumnFamily())).toString();
+                            String columnQualifier = cdcColumnInfoList.get(columnListIndex)
+                                    .getColumnName();
+                            if (Arrays.equals(
+                                    cdcColumnInfoList.get(columnListIndex).getColumnFamily(),
+                                    cdcDataTableInfo.getDefaultColumnFamily())) {
+                                columnFamily = DEFAULT_COLUMN_FAMILY_STR;
```
TheNamesRai, line 188: use dot notation for as a key (family.qualifier)

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/CDCGlobalIndexRegionScanner.java
```diff
@@ -90,128 +88,144 @@ public CDCGlobalIndexRegionScanner(final RegionScanner innerScanner,
...
-                                                    cell.getQualifierOffset(),
-                                                    cell.getQualifierLength()),
-                                            cell);
-                                    ++columnPointers[i];
+                        if (CDCUtil.compareCellFamilyAndQualifier(
+                                cell.getFamilyArray(), cell.getQualifierArray(),
+                                cdcColumnInfoList.get(columnListIndex).getColumnFamily(),
+                                cdcColumnInfoList.get(columnListIndex)
+                                        .getColumnQualifier()) == 0) {
+                            String columnFamily = StandardCharsets.UTF_8
```
TheNamesRai, line 180: Add Family string in the CDCTableInfo Obj

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/CDCGlobalIndexRegionScanner.java
```diff
@@ -90,128 +88,144 @@ public CDCGlobalIndexRegionScanner(final RegionScanner innerScanner,
...
-                                build()));
-                    }
                 }
-                if (dataRow != null && tupleProjector != null) {
-                    IndexUtil.addTupleAsOneCell(result, new ResultTuple(dataRow),
+                Result cdcRow = getCDCImage(
+                        preImageObj, changeImageObj, isIndexCellDeleteRow, indexCellTS, firstCell);
+                if (cdcRow != null && tupleProjector != null) {
+                    if (firstCell.getType() == Cell.Type.DeleteFamily) {
+                        result.add(CellBuilderFactory.create(CellBuilderType.DEEP_COPY)
```
TheNamesRai, line 216: Add a comment

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/CDCGlobalIndexRegionScanner.java
```diff
@@ -223,4 +237,85 @@ protected boolean getNextCoveredIndexRow(List<Cell> result) throws IOException {
...
         return false;
     }
+
+    private Result getCDCImage(
+            Map<String, Map<String, Object>> preImageObj,
+            Map<String, Map<String, Object>> changeImageObj,
+            boolean isIndexCellDeleteRow, Long indexCellTS, Cell firstCell) {
+        Map<String, Object> rowValueMap = new HashMap<>();
+
+        if (this.cdcChangeScopeSet.size() == 0
```
TheNamesRai, line 247: Default change scope should be just Change (Not All)

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/CDCGlobalIndexRegionScanner.java
```diff
@@ -223,4 +237,85 @@ protected boolean getNextCoveredIndexRow(List<Cell> result) throws IOException {
...
+                build()));
+
+        return cdcRow;
+    }
+
+    private Object getColumnValue (Cell cell, PDataType dataType) {
+        if (dataType.getSqlType() == Types.BINARY) {
+            return Base64.getEncoder().encodeToString(cell.getValueArray());
+        } else if (dataType.getSqlType() == Types.DATE) {
+            return ((Date) dataType.toObject(cell.getValueArray())).getTime();
```
TheNamesRai, line 315: Change this to Stringified value of Date/Timestamp

phoenix-core/src/main/java/org/apache/phoenix/util/CDCUtil.java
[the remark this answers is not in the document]
haridsv: Oh, I didn't know it was added in Java 9. There is also a `org.apache.hadoop.io.FastByteComparisons#compareTo`? I found a copy of this class in Phoenix that is named as `DescVarLengthFastByteComparisons` that is already being used in several places. I think it is using some native comparator so should be more efficient, besides reducing the code duplication.


There is also [org.apache.avro.io.BinaryData#compareBytes](https://github.com/apache/avro/blob/d143d6262f7d6c688e695d1656f8609605835604/lang/java/avro/src/main/java/org/apache/avro/io/BinaryData.java#L183) (requires anew dependency for Phoenix, but it is an existing dependency for HBase)

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/CDCGlobalIndexRegionScanner.java
```diff
@@ -90,128 +88,144 @@ public CDCGlobalIndexRegionScanner(final RegionScanner innerScanner,
...
 
     protected boolean getNextCoveredIndexRow(List<Cell> result) throws IOException {
         if (indexRowIterator.hasNext()) {
             List<Cell> indexRow = indexRowIterator.next();
-            for (Cell c: indexRow) {
-                if (c.getType() == Cell.Type.Put) {
-                    result.add(c);
-                }
-            }
+            Cell firstCell = indexRow.get(indexRow.size() - 1);
```
TheNamesRai, line 121: Add a comment

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/CDCGlobalIndexRegionScanner.java
TheNamesRai: As we have to handle different cases like, multitenant tables, salted tables, we decided to take care of it in later iterations.

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/CDCGlobalIndexRegionScanner.java
```diff
@@ -90,128 +88,144 @@ public CDCGlobalIndexRegionScanner(final RegionScanner innerScanner,
...
-                                deleteMarkers.add(cell);
-                            }
-                            if (CellUtil.matchingColumn(cell, QueryConstants.DEFAULT_COLUMN_FAMILY_BYTES,
-                                    emptyKV.getFirst())) {
-                                continue;
+                int columnListIndex = 0;
+                List<CDCTableInfo.CDCColumnInfo> cdcColumnInfoList =
+                        this.cdcDataTableInfo.getColumnInfoList();
+                for (Cell cell : dataRow.rawCells()) {
+                    if (cell.getType() == Cell.Type.DeleteFamily) {
```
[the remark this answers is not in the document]
TheNamesRai: We will only compare with the first Column Family for Delete Family cells because there is no way to delete column family in Phoenix, We can only delete the entire row, which will create same Delete Family marks for all column family

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/CDCGlobalIndexRegionScanner.java
```diff
@@ -90,128 +88,144 @@ public CDCGlobalIndexRegionScanner(final RegionScanner innerScanner,
...
+                            lowerBoundForPreImage = cell.getTimestamp();
+                        }
+                    } else if (cell.getType() == Cell.Type.DeleteColumn
+                            || cell.getType() == Cell.Type.Put) {
+                        if (!Arrays.equals(cell.getQualifierArray(), emptyCQ)
+                                && CDCUtil.compareCellFamilyAndQualifier(
+                                        cell.getFamilyArray(), cell.getQualifierArray(),
+                                        cdcColumnInfoList.get(columnListIndex).getColumnFamily(),
+                                        cdcColumnInfoList.get(columnListIndex)
+                                                .getColumnQualifier()) > 0) {
```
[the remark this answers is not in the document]
TheNamesRai: Yes, its not required. We will only need the empty column qualifier check. Will update. thanks

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/CDCGlobalIndexRegionScanner.java
```diff
@@ -90,128 +90,157 @@ public CDCGlobalIndexRegionScanner(final RegionScanner innerScanner,
...
     }
 
     protected boolean getNextCoveredIndexRow(List<Cell> result) throws IOException {
         if (indexRowIterator.hasNext()) {
             List<Cell> indexRow = indexRowIterator.next();
-            for (Cell c: indexRow) {
-                if (c.getType() == Cell.Type.Put) {
-                    result.add(c);
-                }
+            // firstCell: Picking the earliest cell in the index row so that
```
kadirozde, line 121: Please add this comment : A given CDC index row can have either one put cell, or a delete marker (Delete Family) and a put cell due to DML operations. When a data table column is dropped by a DDL operation and a second put cell with a higher timestamp is inserted to an index row. The timestamp of this put cell matches with the timestamp of the corresponding DeleteColumn mutation on the data table row. This DDL mutation will not show up in the CDC stream.  Index row delete marker timestamp is always higher than or equal to the put cell timestamp. The delete markers are for maintaining the index rows and can be ignored. The oldest put cell will be used to point back to the data table row version. The timestamp of the put cell is used the identify the corresponding data table mutation which can a put or delete mutation. This is the reason we only consider the oldest cell (i.e. first cell) here.

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/CDCGlobalIndexRegionScanner.java
```diff
@@ -90,128 +90,157 @@ public CDCGlobalIndexRegionScanner(final RegionScanner innerScanner,
...
+            // timestamp of the cell and the row will be same.
+            Cell firstCell = indexRow.get(indexRow.size() - 1);
+            ImmutableBytesPtr dataRowKey = new ImmutableBytesPtr(
+                    indexToDataRowKeyMap.get(CellUtil.cloneRow(firstCell)));
+            Result dataRow = dataRows.get(dataRowKey);
```
kadirozde, line 126: It should be possible to have a null dataRow. This happens when the index table write succeeds but the data table write fails. We should handle it here. In that case we need to skip this index row.

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/CDCGlobalIndexRegionScanner.java
```diff
@@ -90,128 +90,157 @@ public CDCGlobalIndexRegionScanner(final RegionScanner innerScanner,
...
-                                build()));
-                    }
                 }
-                if (dataRow != null && tupleProjector != null) {
-                    IndexUtil.addTupleAsOneCell(result, new ResultTuple(dataRow),
+                Result cdcRow = getCDCImage(preImageObj, changeImageObj, isIndexCellDeleteRow,
+                        indexCellTS, firstCell, isChangeImageInScope, isPreImageInScope,
+                        isPostImageInScope);
+                if (cdcRow != null && tupleProjector != null) {
+                    if (firstCell.getType() == Cell.Type.DeleteFamily) {
```
kadirozde, line 226: Is it possible to have a Delete Family first cell? Should not be the first cell a put cell?

phoenix-core/src/main/java/org/apache/phoenix/coprocessor/CDCGlobalIndexRegionScanner.java
```diff
@@ -223,4 +252,67 @@ protected boolean getNextCoveredIndexRow(List<Cell> result) throws IOException {
...
+                    postImageObj.put(preImageObjCol.getKey(), preImageObjCol.getValue());
+                }
+                for (Map.Entry<String, Object> changeImageObjCol : changeImageObj.entrySet()) {
+                    postImageObj.put(changeImageObjCol.getKey(), changeImageObjCol.getValue());
+                }
+            }
+            rowValueMap.put(CDC_POST_IMAGE, postImageObj);
+        }
+
+        if (isIndexCellDeleteRow) {
```
kadirozde, line 283: We should not decide if the data table mutation is delete or upsert based on the index row mutation type. It should be determined based on the data table mutation type.

phoenix-core/src/it/java/org/apache/phoenix/end2end/CDCMiscIT.java
```diff
@@ -416,6 +706,116 @@ public void testSelectTimeRangeQueries() throws Exception {
...
+        props.put("phoenix.query.timeoutMs", "6000000");
+        props.put("zookeeper.session.timeout", "6000000");
+        props.put("hbase.rpc.timeout", "6000000");
+        Connection conn = DriverManager.getConnection(getUrl(), props);
+        String tableName = generateUniqueName();
+        conn.createStatement().execute(
+                "CREATE TABLE  " + tableName + " ( k INTEGER PRIMARY KEY," + " v1 INTEGER, v2 INTEGER, v3 INTEGER, v4 INTEGER)");
+        String cdcName = generateUniqueName();
+
+        conn.createStatement().execute("UPSERT INTO " + tableName + " (k, v1, v2) VALUES (1, 100, 1000)");
```
kadirozde, line 723: Let's have more row versions here and retrieve these versions with multiple select statements with different time ranges, that is the where clauses should include the corresponding PHOENIX_ROW_TIMESTAMP() ranges.

phoenix-core/src/it/java/org/apache/phoenix/end2end/CDCMiscIT.java
```diff
@@ -446,4 +846,91 @@ public void testSelectUncoveredIndex() throws Exception {
...
+                    + " ON " + tableName;
+            createAndWait(conn, tableName, cdcName, cdc_sql);
+            assertCDCState(conn, cdcName, null, 3);
+            assertCDCBinaryAndDateColumn(conn.createStatement().executeQuery
+                    ("SELECT /*+ CDC_INCLUDE(PRE, POST, CHANGE) */ * " + "FROM " + cdcName),
+                    byteColumnValues, dateColumnValues, timestampColumnValue);
+        } finally {
+            conn.close();
+        }
+    }
```
kadirozde, line 935: We need to have tests with successful index writes but failed data table writes. Please see GlobalIndexCheckerIT for how to fail data table writes. We should not see any CDC event for these failed transactions.

kadirozde (changes requested):

virajjasani:
testCreate is failing: https://ci-hadoop.apache.org/job/Phoenix/job/Phoenix-PreCommit-GitHub-PR/job/PR-1813/15/testReport/junit/org.apache.phoenix.end2end/CDCDefinitionIT/

`CDCBaseIT` needs license

virajjasani:
Merging to unblock final PR against master branch.

PR #1816, opened 2024-02-01
PHOENIX-6769 Align mockito version with Hadoop and HBase

stoty:
* Update mockito to 4.11.0, same as Hbase branch-3
* Remove org.hamcrest usage and direct dependency


gjacoby126 (approved):
+1. It would be good at some point to replace the use of reflection to change internal state with some kind of dependency injection, but that can be a future JIRA.

stoty:
Or maybe we could just expose them as public, with a @PublicForTesting annotation, like we do in other classes.

PR #1817, opened 2024-02-05
PHOENIX-7195 Exclude reload4j dependencies from Hadoop and add explic…

stoty:
…it log4j2 test dependencies

### DIFF
--- a/phoenix-core/pom.xml
+++ b/phoenix-core/pom.xml
@@ -444,5 +444,28 @@
             <artifactId>hamcrest-core</artifactId>
             <scope>test</scope>
         </dependency>
+
+        <!-- logging -->
+        <dependency>
+            <groupId>org.apache.logging.log4j</groupId>
+            <artifactId>log4j-api</artifactId>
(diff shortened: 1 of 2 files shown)


virajjasani (commented):
Since we are excluding reload4j, we need log4j at test scope?

stoty:
We've already switched to log4j2 previously.
I suspect that this was broken during the code refactor, when we forgot to re-add the log4j2 dependencies to phoenix-core.
So that part is just a regression fix.

Recent Hadoop releases has switched for lof4j1 to reload4j, and we haven't filtered that yet, so this is needed to restore the correct logging libraries (log4j2) in case a new Hadoop is used (otherwise we'd have both reload4j and log4j2 jars). This is a net new change.

stoty:
So this is in effect two different fixes rolled into one, one regression fix (re-add log4j2 libs), and one for new Hadoop releases (filter reload4j).

virajjasani (approved):

stoty:
merged manually.

PR #1821, opened 2024-02-09
PHOENIX-7041 (ADDENDUM ): Populate ROW_KEY_PREFIX column when creating views

jpisaac:


kadirozde (approved):
+1, Thanks!

kadirozde:
@jpisaac, I approved this PR and then noticed the test failures. Can you look into the failures?

phoenix-core/src/main/java/org/apache/phoenix/schema/MetaDataClient.java
```diff
@@ -3243,10 +3243,10 @@ public boolean isViewReferenced() {
                 tableUpsert.setInt(34, ttl);
             }
 
-            if (rowKeyPrefix == null) {
+            if (rowKeyMatcher == null) {
```
lokiore, line 3246: @jpisaac a question do we want to store null in SYSCAT in case of Empty_Byte_array as well!?
We are not doing any scans based on it bring null!?

lokiore (approved):
LGTM, thanks for fixing the tests

phoenix-core/src/main/java/org/apache/phoenix/schema/MetaDataClient.java
jpisaac: Added that check too!

PR #1822, opened 2024-02-13
PHOENIX-7213 Add option for unlimited phoenix.query.QueueSize

tkhurana:


stoty (changes requested):
The change looks good, but it doesn't do what the commit message / JIRA id says.

I suggest changing the message/description to something like 
"Add option for unlimited phoenix.query.QueueSize" , or something similar.

tkhurana:
Updated JIRA title and commit message. Thanks @stoty

stoty (approved):
+1 LGTM

PR #1823, opened 2024-02-13
PHOENIX-7212 : Handle inherited view indexes when validating LAST_DDL_TIMESTAMPS

palashc:


palashc:
@tkhurana wanted to track all changes for inherited view indexes with this JIRA, if I can get a +1. Thank you!

phoenix-core-client/src/main/java/org/apache/phoenix/util/ViewUtil.java
```diff
@@ -484,7 +484,8 @@ public static void addIndexesFromParent(PhoenixConnection connection, PTable vie
                             .setTableName(modifiedIndexName)
                             .setViewStatement(viewStatement)
                             .setUpdateCacheFrequency(view.getUpdateCacheFrequency())
-                            .setTenantId(view.getTenantId())
+                            //retain the tenantId from the index being inherited
+                            .setTenantId(index.getTenantId())
```
tkhurana, line 488: @palashc Can you add another test case where there are multiple tenant views on the same global view ? And what would happen if those tenant views have the same name (very common)  because they have a different tenant id ? I think they will now point to the same entry in the metadata cache.

phoenix-core/src/test/java/org/apache/phoenix/cache/ServerMetadataCacheTest.java
```diff
@@ -1591,8 +1593,93 @@ public void testAncestorLastDDLMapPopulatedInDifferentClient() throws Exception
...
+    public void testInheritedIndexOnTenantViewsDifferentNames() throws Exception {
+        testInheritedIndexOnTenantViews(false);
+    }
+
+    /**
+     * Test that tenant connections are able to learn about state change of an inherited index
+     * on their tenant views with same names.
+     */
+    @Test
+    public void testInheritedIndexOnTenantViewsSameNames() throws Exception {
```
palashc, line 1610: @tkhurana Added this test.

phoenix-core-client/src/main/java/org/apache/phoenix/util/ViewUtil.java
palashc: Added the test. 

> I think they will now point to the same entry in the metadata cache.

Yes. There was only one entry in the cache for the index with modified name and tenant id is null. Both tenant view PTables also have the index ptable with the modified name and no tenant id.

phoenix-core/src/test/java/org/apache/phoenix/cache/ServerMetadataCacheTest.java
```diff
@@ -1591,8 +1593,93 @@ public void testAncestorLastDDLMapPopulatedInDifferentClient() throws Exception
...
+
+            // query on secondary key should use inherited index for all tenant views.
+            try (Connection tenantConn1 = cqs.connect(url, tenantProps1);
+                 Connection tenantConn2 = cqs.connect(url, tenantProps2)) {
+
+                String query1 = "SELECT KV2 FROM  " + tenantViewName1 + " WHERE KV = 'KV'";
+                String query2 = "SELECT KV2 FROM  " + tenantViewName2 + " WHERE KV = 'KV'";
+
+                ResultSet rs = tenantConn1.createStatement().executeQuery(query1);
+                assertPlan((PhoenixResultSet) rs,  "", tenantViewName1 + "#" + globalViewIndexName);
```
tkhurana, line 1662: Small nit: Instead of using `#` use the constant `CHILD_VIEW_INDEX_NAME_SEPARATOR` in the test code. There are multiple places where `#` is being used.

tkhurana (approved):

shahrs87 (approved):
+1 ltgm.

PR #1825, opened 2024-02-15
PHOENIX-7216 Bump Hadoop version to 3.2.4 for 2.5.x profile

stoty:


pom.xml
```diff
@@ -74,7 +74,7 @@
     <!-- Hadoop and Hbase-thirdparty version -->
     <!-- These are expected to be overridden to conform to cluster versions
     along with hbase.version (defined in profiles) -->
-    <hadoop.version>3.1.4</hadoop.version>
+    <hadoop.version>3.2.4</hadoop.version>
```
stoty, line 77: I've changed trhe default to be the newest to appease static checkers.
We are overriding it every profile now, anyway.

virajjasani (approved):

PR #1832, opened 2024-02-23
PHOENIX-7229 Leverage bloom filters for single key point lookups

tkhurana:
Jira: PHOENIX-7229

### DIFF
--- a/phoenix-core-client/src/main/java/org/apache/phoenix/iterate/BaseResultIterators.java
+++ b/phoenix-core-client/src/main/java/org/apache/phoenix/iterate/BaseResultIterators.java
@@ -37,6 +37,7 @@ import java.io.ByteArrayInputStream;
 import java.io.DataInput;
 import java.io.DataInputStream;
 import java.io.EOFException;
+import java.io.IOException;
 import java.sql.SQLException;
 import java.util.*;
 import java.util.Arrays;
@@ -72,6 +73,7 @@ import org.apache.phoenix.compile.StatementContext;
 import org.apache.phoenix.coprocessorclient.BaseScannerRegionObserverConstants;
 import org.apache.phoenix.coprocessorclient.HashJoinCacheNotFoundException;
 import org.apache.phoenix.coprocessorclient.UngroupedAggregateRegionObserverHelper;
+import org.apache.phoenix.exception.PhoenixIOException;
 import org.apache.phoenix.exception.SQLExceptionInfo;
 import org.apache.phoenix.execute.MutationState;
 import org.apache.phoenix.execute.ScanPlan;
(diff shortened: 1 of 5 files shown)


phoenix-core/src/it/java/org/apache/phoenix/end2end/BloomFilterIT.java
```diff
@@ -0,0 +1,227 @@
+package org.apache.phoenix.end2end;
```
virajjasani, line 1: ASF lincense

virajjasani:
@tkhurana the reason why this PR has build failure is because of `MetricsRegionServerWrapper` usage. It comes from `hbase-hadoop-compat`, otherwise the master/5.2 branches don't have that issue.

kadirozde (approved):
+1, Thanks

tkhurana:
> Need updates in WhereOptimizerTest, since we are updating high level Scan object's stop rowkey. Updated in the latest revision.

@virajjasani No, actually scanFromContext is not changing. It is the scan we generate from the iterator is changing.

virajjasani:
I made this commit https://github.com/apache/phoenix/commit/1e3fa9dae41c361ba2b7f4b381ecbf08568ee126
For single pointlookup, scan context can also be changed. Does it hurt?

virajjasani:
Earlier also scan context object had same scan boundary as the scans generated from it.

phoenix-core-client/src/main/java/org/apache/phoenix/iterate/BaseResultIterators.java
```diff
@@ -941,12 +940,7 @@ private ScansWithRegionLocations getParallelScans(byte[] startKey, byte[] stopKe
             Scan scanFromContext = context.getScan();
             if (scanRanges.getPointLookupCount() == 1) {
                 // leverage bloom filter for single key point lookup by turning scan to Get Scan#isGetScan()
-                try {
-                    scanFromContext = new Scan(context.getScan());
-                } catch (IOException e) {
-                    throw new PhoenixIOException(e);
-                }
-                scanFromContext.withStopRow(scanFromContext.getStartRow(), scanFromContext.includeStartRow());
+                scanFromContext.withStopRow(scanFromContext.getStartRow(), true);
```
tkhurana, line 943: I actually didn't want to change the scan stored in StatementContext. I wanted to modify the scan we generate in getIterators.

tkhurana: You generate a scan in StatementContext and then we generate the actual scans when we are creating the iterators like a high level scan in statementcontext could generate multiple parallel scans one for each region server. That is why I was copying the original scan stored in StatementContext.

virajjasani: I see, then we can revert it.

tkhurana: Earlier the scan object had stop key as next(start key) for point lookups. Take a look at this test case https://github.com/apache/phoenix/blob/master/phoenix-core/src/test/java/org/apache/phoenix/compile/WhereOptimizerTest.java#L988-L992

tkhurana:
> @tkhurana the reason why this PR has build failure is because of `MetricsRegionServerWrapper` usage. It comes from `hbase-hadoop-compat`, otherwise the master/5.2 branches don't have that issue.

Thanks @virajjasani It makes sense.

virajjasani:
@tkhurana HBASE-27241 is only present on hbase 2.4 based build fails if we try to access bloom filter metrics with IT test, i will have to revert the IT test, we can figure out how best to write test later.

PR #1835, opened 2024-02-26
PHOENIX-7237 Pherf unit tests PherfTest and ResourceTest are failing

ddupg:


stoty:
I don't like this, as this is very brittle, and we will have the same problem again soon. (The previous patch has the same problem)

The real problem is that pherf seems to try to load its test config from anywhere on the classpath.

I would prefer that the logic in getResourcesPath is fixed so that it does not recursively search the whole classpath.

https://github.com/apache/phoenix/blob/bc1e2e7bea40c7d03940748e8f1d9f6b23339867/phoenix-pherf/src/main/java/org/apache/phoenix/pherf/util/ResourceList.java#L103

ddupg:
> I don't like this, as this is very brittle, and we will have the same problem again soon. (The previous patch has the same problem)
> 
> The real problem is that pherf seems to try to load its test config from anywhere on the classpath.
> 
> I would prefer that the logic in getResourcesPath is fixed so that it does not recursively search the whole classpath.
> 
> https://github.com/apache/phoenix/blob/bc1e2e7bea40c7d03940748e8f1d9f6b23339867/phoenix-pherf/src/main/java/org/apache/phoenix/pherf/util/ResourceList.java#L103
Thanks @stoty for reviewing and opinion, I'll try to fix it.

PR #1838, opened 2024-02-28
PHOENIX-7243 : Add isServerConnection property to ConnectionInfo

palashc:


phoenix-core/src/test/java/org/apache/phoenix/jdbc/PhoenixTestDriver.java
```diff
@@ -92,7 +93,8 @@ public synchronized Connection connect(String url, Properties info) throws SQLEx
     public synchronized ConnectionQueryServices getConnectionQueryServices(String url, Properties infoIn) throws SQLException {
         checkClosed();
         final Properties info = PropertiesUtil.deepCopy(infoIn);
-        ConnectionInfo connInfo = ConnectionInfo.create(url, null, info);
+        boolean isServerConnection = Boolean.valueOf(info.getProperty(QueryUtil.IS_SERVER_CONNECTION));
+        ConnectionInfo connInfo = ConnectionInfo.create(url, null, info, isServerConnection);
```
shahrs87, line 97: Why do we want to pass as a separate argument for `isServerConnection`?  It is present in `info` object anyways.

phoenix-core-client/src/main/java/org/apache/phoenix/jdbc/ConnectionInfo.java
```diff
@@ -114,6 +115,15 @@ public static ConnectionInfo create(String url, ReadOnlyProps props, Properties
         return create(url, conf, props, info);
     }
 
+    public static ConnectionInfo create(String url, ReadOnlyProps props,
+            Properties info, boolean isServerConnection)
+            throws SQLException {
+        Configuration conf = HBaseFactoryProvider.getConfigurationFactory().getConfiguration();
+        ConnectionInfo connInfo = create(url, conf, props, info);
+        connInfo.isServerConnection = isServerConnection;
```
shahrs87, line 123: I think we should initialize `isServerConnection` the same way we are initializing principal, keytab, user and haGroup variables.

shahrs87: All the other variables are being set via Builder. Lets not break that pattern for `isServerConnection`

phoenix-core/src/test/java/org/apache/phoenix/jdbc/PhoenixTestDriver.java
palashc: Pushed the refactor, no changes in Drivers needed.

phoenix-core-client/src/main/java/org/apache/phoenix/jdbc/ConnectionInfo.java
palashc: Done.

palashc:
Opened https://github.com/apache/phoenix/pull/1839 for master branch @shahrs87

stoty:
Server connections are always opened on the server, and they are only special in that they are cached and cannot be closed.

How would creating a new server connection help if we already have a similar non-server connection ?

palashc:
@stoty This came up in [PHOENIX-6883](https://issues.apache.org/jira/browse/PHOENIX-6883) where we want to invalidate the metadata cache on a region server and only a server connection should be allowed to do so. Since we enforced that condition, some tests which tried to perform invalidation on a non-server connection failed.

shahrs87:
Thank you @stoty for your comment.
Mostly this is a problem on the test side where PhoenixTestDriver is caching the server and non server connections.
Looking at the code, MDEI always creates server connections but there is nothing enforcing in the code so there is a possibility that someone might create non server connection also.

PR #1844, opened 2024-03-04
PHOENIX-4555 Only mark view as updatable if rows cannot overlap with other updatable views

jinggou:
Jira: [PHOENIX-4555](https://issues.apache.org/jira/browse/PHOENIX-4555)

Restrict view to be UPDATABLE if the view specification:
1. uses only the PK columns;
2. starts from the first PK column (ignore the prefix PK columns, TENANT_ID and/or _SALTED, if the parent table is multi-tenant and/or salted);
3. PK columns should be in the order they are defined;
4. uses the same set of PK columns as its sibling views' specification;

Otherwise, mark the view as READ_ONLY.


phoenix-core-client/src/main/java/org/apache/phoenix/compile/CreateTableCompiler.java
```diff
@@ -209,6 +247,137 @@ public MutationPlan compile(CreateTableStatement create) throws SQLException {
...
+                Expression siblingViewWhere = getWhereFromView(connection, siblingView);
+                Set<PColumn> siblingViewPkColsInWhere = new HashSet<>();
+                if (siblingViewWhere != null) {
+                    ViewWhereExpressionValidatorVisitor siblingViewValidatorVisitor =
+                            new ViewWhereExpressionValidatorVisitor(parentToBe,
+                                    siblingViewPkColsInWhere, new HashSet<>());
+                    siblingViewWhere.accept(siblingViewValidatorVisitor);
+                }
+                if (!pkColumnsInWhere.equals(siblingViewPkColsInWhere)) {
+                    return ViewType.READ_ONLY;
```
virajjasani, line 323: Every case where we return READ_ONLY view, can we add a log explaining why we need to create this as a read-only view?

phoenix-core-client/src/main/java/org/apache/phoenix/compile/CreateTableCompiler.java
```diff
@@ -164,9 +185,29 @@ public MutationPlan compile(CreateTableStatement create) throws SQLException {
...
+                                new ViewWhereExpressionValidatorVisitor(parentToBe,
+                                        pkColumnsInWhere, nonPkColumnsInWhere);
+                        where.accept(validatorVisitor);
+                        if (!(connection.getQueryServices()
+                                instanceof ConnectionlessQueryServicesImpl)) {
+                            try {
+                                viewTypeToBe = setViewTypeToBe(connection, parentToBe,
+                                        pkColumnsInWhere, nonPkColumnsInWhere);
+                            } catch (IOException e) {
+                                throw new RuntimeException(e);
```
jpisaac, line 204: Do we want to throw a RuntimeException? OR build a SQLException and throw that?

phoenix-core-client/src/main/java/org/apache/phoenix/compile/CreateTableCompiler.java
```diff
@@ -209,6 +250,151 @@ public MutationPlan compile(CreateTableStatement create) throws SQLException {
...
+        List<Integer> pkPositions = new ArrayList<>();
+        parentToBe.getPKColumns().forEach(tablePkColumn ->
+                tablePkPositions.add(tablePkColumn.getPosition()));
+        pkColumnsInWhere.forEach(pkColumn -> pkPositions.add(pkColumn.getPosition()));
+        Collections.sort(pkPositions);
+
+        // 2. If not multi tenant, view specification WHERE clause should start from the first PK
+        // column; otherwise, start from the second PK column
+        boolean isMultiTenant = parentToBe.isMultiTenant();
+        int firstPkPosition = pkPositions.get(0);
```
jpisaac, line 293: Can we test with SALTED tables and views, My understanding is the PK position might be different.

phoenix-core-client/src/main/java/org/apache/phoenix/compile/CreateTableCompiler.java
```diff
@@ -209,6 +250,151 @@ public MutationPlan compile(CreateTableStatement create) throws SQLException {
...
+
+    /**
+     * Check if the primary key columns are in order (consecutive in position) as they are
+     * defined, providing their positions list.
+     * @param pkPositions A positions list of PK columns to be checked
+     * @param tablePkPositions The positions list of the table's PK columns
+     * @param isMultiTenant Whether the parent table is multi tenant
+     * @return
+     */
+    private boolean isPkColumnsInOrder(final List<Integer> pkPositions,
```
jpisaac, line 378: Can you make sure, this holds for SALTED tables.

phoenix-core-client/src/main/java/org/apache/phoenix/compile/CreateTableCompiler.java
jinggou: You are right. I tested with SALTED table, the first PK column would be _SALT, so the first PK defined in view would be in the second position in the table, just like the multi-tenant table. What do you think we should do here? @jpisaac

phoenix-core-client/src/main/java/org/apache/phoenix/compile/CreateTableCompiler.java
jinggou: The column position will move one position back, just like the multi-tenant table

phoenix-core-client/src/main/java/org/apache/phoenix/compile/CreateTableCompiler.java
jpisaac: @jinggou 
Adjust the logic if it is a salted table, you can find out if it is a salted table by boolean isSalted = parentToBe.getBucketNum() != null;

virajjasani:
1. Update the description on Javadocs and PR/Jira to specify Salt bucket case
2. Pass null non-pk column set and if it is not null, add values from KeyValueColumnExpression
3. Loggers with view name and more details if required e.g. for sibling views, if the columns don't match, log the column set for view and sibling view, and also sibling view name
4. Tests with comments on what criteria is covered

virajjasani:
We are really close now, thank you for working on this @jinggou!

virajjasani (approved):
+1, pending build results

PR #1845, opened 2024-03-05
PHOENIX-7251 : Refactor server-side code to support multiple ServerMetadataCache for ITs which create multiple RSs or mini clusters

palashc:


phoenix-core-client/src/main/java/org/apache/phoenix/cache/ServerMetadataCacheImpl.java
```diff
@@ -0,0 +1,169 @@
...
+        byte[] tableKey = SchemaUtil.getTableKey(tenantID, schemaName, tableName);
+        ImmutableBytesPtr tableKeyPtr = new ImmutableBytesPtr(tableKey);
+        lastDDLTimestampMap.invalidate(tableKeyPtr);
+    }
+
+    protected Connection getConnection(Properties properties) throws SQLException {
+        return QueryUtil.getConnectionOnServer(properties, this.conf);
+    }
+
+    protected static void resetCache() {
```
shahrs87, line 165: Do we need this method here?

phoenix-core/src/it/java/org/apache/phoenix/end2end/ServerMetadataCacheTestImpl.java
```diff
@@ -0,0 +1,66 @@
...
+import java.util.Properties;
+
+/**
+ * Implementation of {@link ServerMetadataCache} for Integration Tests.
+ * Supports keeping more than one instance keyed on the regionserver ServerName.
+ */
+public class ServerMetadataCacheTestImpl extends ServerMetadataCacheImpl {
+    private static volatile Map<ServerName, ServerMetadataCacheTestImpl> INSTANCES = new HashMap<>();
+    private Connection connectionForTesting;
+    ServerMetadataCacheTestImpl(Configuration conf) {
```
shahrs87, line 23: new line before constructor.

phoenix-core/src/it/java/org/apache/phoenix/end2end/ServerMetadataCacheTestImpl.java
```diff
@@ -0,0 +1,66 @@
...
+        this.connectionForTesting = connection;
+    }
+
+    public static void resetCache() {
+        INSTANCES.clear();
+    }
+
+    @Override
+    protected Connection getConnection(Properties properties) throws SQLException {
+        System.out.println("USED");
```
shahrs87, line 62: remove sysout comment.

phoenix-core/src/it/java/org/apache/phoenix/end2end/ServerMetadataCacheTestImpl.java
```diff
@@ -0,0 +1,66 @@
...
+        return connectionForTesting != null ? connectionForTesting
+                : QueryUtil.getConnectionOnServer(properties, this.conf);
```
shahrs87, line 64:
```suggestion
                return connectionForTesting != null ? connectionForTesting : super.getConnection(properties);
```

phoenix-core/src/test/java/org/apache/phoenix/cache/ServerMetadataCacheImplTest.java
```diff
@@ -76,11 +78,13 @@
 import static org.mockito.Mockito.verify;
 
 @Category(ParallelStatsDisabledIT.class)
-public class ServerMetadataCacheTest extends ParallelStatsDisabledIT {
+public class ServerMetadataCacheImplTest extends ParallelStatsDisabledIT {
```
shahrs87, line 81: Do we want to rename this class to `ServerMetadataCacheIT` to avoid confusion between `ServerMetadataCacheImplTest` and `ServerMetadataCacheTestImpl`

phoenix-core/src/test/java/org/apache/phoenix/cache/ServerMetadataCacheImplTest.java
```diff
@@ -121,7 +127,7 @@ public void testCacheForBaseTable() throws Exception {
             createTable(conn, tableNameStr, NEVER);
             pTable = PhoenixRuntime.getTableNoCache(conn,
                     tableNameStr);// --> First call to CQSI#getTable
-            ServerMetadataCache cache = ServerMetadataCache.getInstance(config);
+            ServerMetadataCacheTestImpl cache = ServerMetadataCacheTestImpl.getInstance(config, serverName);
```
shahrs87, line 130: Instead of directly calling ServerMetadataCacheTestImpl, we should call it via findCoprocessor method something like this:
```
private ServerMetadataCache getCache() {
  String phoenixRegionServerEndpoint = conf.getValue(REGIONSERVER_COPROCESSOR_CONF_KEY);
  assertNotNull(phoenixRegionServerEndpoint);
  PhoenixRegionServerEndpoint coproc = regionServer.getRegionServerCoprocessorHost()               
  .findCoprocessor(phoenixRegionServerEndpoint.class); 
  assertNotNull(coproc);
  ServerMetadataCache cache = coproc.getServerMetadataCache();
  assertNotNull(cache);
  return cache;
}
```

shahrs87 (changes requested):
Posting some initial comments.

palashc:
> we need to call ServerMetadaCacheTestImpl#resetCache in all the tests with category: NeedsOwnMiniClusterTest
@shahrs87 Does it make sense to have that change as part of [PHOENIX-7166](https://github.com/apache/phoenix/pull/1778)?

phoenix-core/src/it/java/org/apache/phoenix/end2end/ServerMetadataCacheTestImpl.java
```diff
@@ -0,0 +1,66 @@
...
+import java.sql.SQLException;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Properties;
+
+/**
+ * Implementation of {@link ServerMetadataCache} for Integration Tests.
+ * Supports keeping more than one instance keyed on the regionserver ServerName.
+ */
+public class ServerMetadataCacheTestImpl extends ServerMetadataCacheImpl {
```
shahrs87, line 20: Can you add some more context in the comment something like:

PhoenixRegionServerEndpoint is a region server coproc. There is a 1-1 correspondence between PhoenixRegionServerEndpoint and ServerMetadataCache.
In ITs we can have multiple regionservers per cluster so we need multiple instances of ServerMetadataCache.
And HighAvailabilityTestingUtility creates 2 clusters so we need to have one instance of ServerMetadataCache for each regionserver in each cluster.

shahrs87 (changes requested):
Just one comment. Other than that looks good.

phoenix-core/src/it/java/org/apache/phoenix/end2end/ServerMetadataCacheTestImpl.java
palashc: done.

palashc:
@shahrs87 looks like no test failures apart from the one flapper - https://ci-hadoop.apache.org/job/Phoenix/job/Phoenix-PreCommit-GitHub-PR/job/PR-1845/6/

shahrs87:
There are 3 checkstyle warning, 1 asf license warning, 1 spotbug warning. Can you please fix them? @palashc

shahrs87:
@palashc  There are still 2 checkstyle and 1 spotbugs warning. Mind taking a look? 
spotbugs warning is relevant in this case.

palashc:
Checkstyle
- INSTANCE variable does not match static variable name pattern but it was already present in code, we just refactored in this PR.
- new class does have license but it's not a line by line match to what is expected, I took it from other classes in the codebase.
 
Spotbug
- This is about storing the conf object inside the server metadata cache, also something which was only refactored into a new class in this PR.

shahrs87:
> INSTANCE variable does not match static variable name pattern but it was already present in code, we just refactored in this PR.

Can we fix this just by following the pattern that it expects? 

> new class does have license but it's not a line by line match to what is expected, I took it from other classes in the codebase.

Do we know what is expected? If yes then lets make checkstyle happy.


> This is about storing the conf object inside the server metadata cache, also something which was only refactored into a new class in this PR.

Agree that this is just refactoring. But I think the correct fix is creating a new conf object something like this:
`this.conf = new Configuration(conf);
`

Lets do these changes to make checkstyle and spotbugs happy. @palashc

palashc:
`this.conf = new Configuration(conf); ` - This change breaks our tests, need to look into it.

shahrs87:
> Maybe we need HBaseConfiguration.create(conf); - Yes, that works!

Lets not do this. This reads all site.xml files (hbase-site.xml, hbase-default.xml, hdfs-site.xml) and then overrides the properties with the passed conf. 

>  This change breaks our tests, need to look into it.

Lets see why this is breaking our test.

palashc:
> Lets see why this is breaking our test.
Creating a connection on the server side fails because of malformed URL exception. 

```
Caused by: java.sql.SQLException: ERROR 102 (08001): Malformed connection url. Quorum not specified and hbase.client.zookeeper.quorum is not set in configuration : jdbc:phoenix
	at org.apache.phoenix.exception.SQLExceptionCode$Factory$1.newException(SQLExceptionCode.java:656)
	at org.apache.phoenix.exception.SQLExceptionInfo.buildException(SQLExceptionInfo.java:229)
	at org.apache.phoenix.jdbc.ConnectionInfo.getMalFormedUrlException(ConnectionInfo.java:82)
	at org.apache.phoenix.jdbc.ZKConnectionInfo$Builder.normalize(ZKConnectionInfo.java:322)
	at org.apache.phoenix.jdbc.ZKConnectionInfo$Builder.create(ZKConnectionInfo.java:175)
	at org.apache.phoenix.jdbc.ConnectionInfo.create(ConnectionInfo.java:174)
	at org.apache.phoenix.jdbc.ConnectionInfo.createNoLogin(ConnectionInfo.java:119)
	at org.apache.phoenix.util.QueryUtil.getConnectionUrl(QueryUtil.java:454)
	at org.apache.phoenix.util.QueryUtil.getConnectionUrl(QueryUtil.java:438)
	at org.apache.phoenix.util.QueryUtil.getConnection(QueryUtil.java:429)
	at org.apache.phoenix.util.QueryUtil.getConnectionOnServer(QueryUtil.java:410)
	at org.apache.phoenix.cache.ServerMetadataCacheImpl.getConnection(ServerMetadataCacheImpl.java:162)
	at org.apache.phoenix.end2end.ServerMetadataCacheTestImpl.getConnection(ServerMetadataCacheTestImpl.java:87)
	at org.apache.phoenix.cache.ServerMetadataCacheImpl.getLastDDLTimestampForTable(ServerMetadataCacheImpl.java:134)
	at org.apache.phoenix.coprocessor.VerifyLastDDLTimestamp.verifyLastDDLTimestamp(VerifyLastDDLTimestamp.java:57)
	at org.apache.phoenix.coprocessor.PhoenixRegionServerEndpoint.validateLastDDLTimestamp(PhoenixRegionServerEndpoint.java:76)
```


So I think we do need to read all site files?

shahrs87:
> So I think we do need to read all site files?

Looks like we do have to use `this.conf = HBaseConfiguration.create(conf)` 
But since we are caching ServerMetadataCache object and this instance will be created just once for each RS, I think it is fine.

shahrs87 (approved):
+1 ltgm

PR #1848, opened 2024-03-05
PHOENIX-7253 Metadata lookup performance improvement for range scan queries

virajjasani:
Jira: PHOENIX-7253

### DIFF
--- a/phoenix-core-client/src/main/java/org/apache/phoenix/iterate/BaseResultIterators.java
+++ b/phoenix-core-client/src/main/java/org/apache/phoenix/iterate/BaseResultIterators.java
@@ -599,9 +599,10 @@ public abstract class BaseResultIterators extends ExplainTable implements Result
             return scans;
     }
 
-    private List<HRegionLocation> getRegionBoundaries(ParallelScanGrouper scanGrouper)
-        throws SQLException{
-        return scanGrouper.getRegionBoundaries(context, physicalTableName);
+    private List<HRegionLocation> getRegionBoundaries(ParallelScanGrouper scanGrouper,
+        byte[] startRegionBoundaryKey, byte[] stopRegionBoundaryKey) throws SQLException {
+        return scanGrouper.getRegionBoundaries(context, physicalTableName, startRegionBoundaryKey,
+            stopRegionBoundaryKey);
     }
 
     private static List<byte[]> toBoundaries(List<HRegionLocation> regionLocations) {
@@ -695,7 +696,8 @@ public abstract class BaseResultIterators extends ExplainTable implements Result
      * @throws SQLException
      */
     private ScansWithRegionLocations getParallelScans(Scan scan) throws SQLException {
-        List<HRegionLocation> regionLocations = getRegionBoundaries(scanGrouper);
+        List<HRegionLocation> regionLocations =
+            getRegionBoundaries(scanGrouper, scan.getStartRow(), scan.getStopRow());
         List<byte[]> regionBoundaries = toBoundaries(regionLocations);
         int regionIndex = 0;
         int stopIndex = regionBoundaries.size();
@@ -963,8 +965,6 @@ public abstract class BaseResultIterators extends ExplainTable implements Result
                 SchemaUtil.processSplit(new byte[] { 0 }, table.getPKColumns());
         byte[] splitPostfix =
                 Arrays.copyOfRange(sampleProcessedSaltByte, 1, sampleProcessedSaltByte.length);
-        List<HRegionLocation> regionLocations = getRegionBoundaries(scanGrouper);
-        List<byte[]> regionBoundaries = toBoundaries(regionLocations);
         boolean isSalted = table.getBucketNum() != null;
         GuidePostsInfo gps = getGuidePosts();
         // case when stats wasn't collected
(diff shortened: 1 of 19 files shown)


dbwong:
Please check the JIRA notes prior to merging this as we want to make sure the cache is used as often as possible or be aware of the possible perf impacts...

virajjasani:
Indeed @dbwong, this is bit sensitive part that affects all queries, so will need to do more testing.

The primary purpose here is to get as much info from HBase client Connection level cached regions as well as only retrieve the regions that cover the scan range boundaries, required for the given (non pointlookup) query.

Let me get a couple of test rounds done, will get back with more explanations, i also need to update javadocs to make things more clear.

virajjasani:
@dbwong here are some of the observations:
In both HBase 1 and 2, region locations for the given table are cached at HBase Connection level. Phoenix CQSI connections are by default cached for ~24 hr. In fact, the issue for the large table range scan queries that we have seen occurs after 24 hr such that multiple range scan queries get hit by performance issue. Sometimes the queries take ~5-10 min worth of time and sometimes the thread that performs meta table lookup gets interrupted. However, even after 24 hr, still significant num of queries get affected.

In the incident, what we have observed is that the base table has ~138k regions. However, the queries are being done using tenant connections. Because of the fact that we get all table regions regardless of the nature of the query, we end up spending significant time retrieving region locations and filling up the connection cache, even when the given query on the tenant view likely does not require going through more than 5-10 regions.

Hence, this fix would improve the performance of queries being done on the large tables (and usually tables that share more tenants are larger) significantly. I am not proposing any change to how the getAllTableRegions API is written, we can continue to follow the same pattern because HBase still does not provide API where it can take start and end key of the scan range and provide all list of region locations in single API, I can file a jira for the same as well.

The current state of the PR address the performance issues of the queries:

- Range Scan
- Any queries using tenant connection - full scan, range scan, point lookup
- Point lookup or Range scan on Salted table
- Point lookup or Range scan on Salted table with Tenant id and/or View index id

The only queries that should require all table region locations are the ones that need full base table scan.

Does this look good to you?

phoenix-core-client/src/main/java/org/apache/phoenix/iterate/BaseResultIterators.java
```diff
@@ -1005,6 +1005,43 @@ private ScansWithRegionLocations getParallelScans(byte[] startKey, byte[] stopKe
...
+                        scan.getStartRow().length - 1);
+
+                    byte[] saltStopRegionKey = new byte[scan.getStopRow().length];
+                    saltStopRegionKey[0] = (byte) i;
+                    System.arraycopy(scan.getStopRow(), 1, saltStopRegionKey, 1,
+                        scan.getStopRow().length - 1);
+
+                    regionLocations.addAll(
+                        getRegionBoundaries(scanGrouper, saltStartRegionKey, saltStopRegionKey));
+                }
```
tkhurana, line 1031: Is all this necessary ? The variable `traverseAllRegions`  is set to true for salted tables so what are we gaining with this extra work

virajjasani: `traverseAllRegions` is only used here, to use scan boundaries:
```
        } else if (!traverseAllRegions) {
            byte[] scanStartRow = scan.getStartRow();
            if (scanStartRow.length != 0 && Bytes.compareTo(scanStartRow, startKey) > 0) {
                startRegionBoundaryKey = startKey = scanStartRow;
            }
            byte[] scanStopRow = scan.getStopRow();
            if (stopKey.length == 0
                    || (scanStopRow.length != 0 && Bytes.compareTo(scanStopRow, stopKey) < 0)) {
                stopRegionBoundaryKey = stopKey = scanStopRow;
            }
        }
```

Here, with this work, we will be able to restrict the range scan for salted table to only specific regions, given that single salt can have many more regions as the table size grows.
Moreover, salted table with tenant id makes the restriction even narrower for large table. Hence, with this work, we are covering optimization for large salted tables.

virajjasani:
Even earlier, without this patch, once CQSI connection is cached, multiple queries executed from the same client, using the same CQSI, do end up calling the same metadata API from HBase ClusterConnection:
```
  /**
   * Find region location hosting passed row
   * @param tableName table name
   * @param row       Row to find.
   * @param reload    If true do not use cache, otherwise bypass.
   * @return Location of row.
   * @throws IOException if a remote or network exception occurs
   */
  HRegionLocation getRegionLocation(TableName tableName, byte[] row, boolean reload)
    throws IOException;
```

kadirozde:
@virajjasani,  PR looks good. The title of the jira does not reflect the PR well. This is not a general query performance improvement. It is a metadata lookup performance improvement. I would suggest something like "Metadata lookup performance improvement for range queries". This PR does not have any tests for checking the number of metadata lookups for range queries. Can we add tests for that?

virajjasani:
For multi-tenant, even full scan is included. So i can update title to reflect "metadata lookup perf improvement for range scan and multi-tenant queries"?

virajjasani:
@kadirozde i have introduced new attribute to ExplainPlan V2 API that can be used to check how many region location lookups were done, added a test to use it.

kadirozde (approved):
+1, Thanks!

PR #1850, opened 2024-03-07
PHOENIX-7263 Row value constructor split keys not allowed on indexes

chrajeshbabu:


phoenix-core/src/it/java/org/apache/phoenix/end2end/index/BaseIndexIT.java
```diff
@@ -134,6 +134,51 @@ protected BaseIndexIT(boolean localIndex, boolean uncovered, boolean mutable, St
...
+            assertEquals("chara", rs.getString("char_col1"));
+            assertEquals(2, rs.getInt(2));
+            assertTrue(rs.next());
+            assertEquals("chara", rs.getString(1));
+            assertEquals(3, rs.getInt(2));
+            assertTrue(rs.next());
+            assertEquals("chara", rs.getString(1));
+            assertEquals(4, rs.getInt(2));
+            assertFalse(rs.next());
+            conn.createStatement().execute("DROP INDEX " + indexName + " ON " + fullTableName);
```
stoty, line 178: Nit: We already have a stmt object. We could also create the stmt in the try-with-reources block.

phoenix-core/src/it/java/org/apache/phoenix/end2end/index/BaseIndexIT.java
```diff
@@ -134,6 +134,51 @@ protected BaseIndexIT(boolean localIndex, boolean uncovered, boolean mutable, St
...
+            String indexSplitKeys = "('a',111, 'foo', 'a', 1, 100, 1.0, TO_DATE('2024-01-01')), " +
+                    "('e','222','john', 'b', 2, 101, 2.0, TO_DATE('2024-01-01')), " +
+                    "('i', '333', 'tom', 'c', 3, 102, 3.0, TO_DATE('2024-01-01'))";
+            ddl = "CREATE " + (localIndex ? "LOCAL" : "") + (uncovered ? "UNCOVERED" : "")
+                    + " INDEX " + indexName + " ON " + fullTableName
+                    + " (char_col1 ASC, int_col1 ASC)"
+                    + (uncovered ? "" : " INCLUDE (long_col1, long_col2)") + (localIndex ? "" :
+                    " SPLIT ON (" + indexSplitKeys + ")");
+            stmt.execute(ddl);
+
```
stoty, line 164: We could check the regions and start keys explicitly.
I think that there is already test code for that and we'd just need to use / copy that.

phoenix-core-client/src/main/java/org/apache/phoenix/compile/SplitKeyUtil.java
```diff
@@ -0,0 +1,89 @@
...
+import org.apache.phoenix.schema.PDatum;
+import org.apache.phoenix.schema.SortOrder;
+import org.apache.phoenix.schema.types.PDataType;
+import org.apache.phoenix.schema.types.PVarbinary;
+import org.apache.phoenix.util.ByteUtil;
+
+import java.sql.SQLException;
+import java.util.List;
+
+public class SplitKeyUtil {
```
stoty, line 35: Do we need a new Util class ?
Do we have an existing Util class that would make sense for this ?

stoty (commented):
I would prefer that that the test enumerated and checked the created regions explicitly.

stoty:
The checkstyle errors also look relevant.

chrajeshbabu:
@stoty  while preparing the split keys index key constraints like order  also should be considered. Will change the patch accordingly and will check whether it can be moved to existing utils.

PR #1851, opened 2024-03-07
PHOENIX-7258: Query Optimizer should pick Index hint even for point lookup queries

sanjeet006py:
Jira: PHOENIX-7258

### DIFF
--- a/phoenix-core-client/src/main/java/org/apache/phoenix/optimize/QueryOptimizer.java
+++ b/phoenix-core-client/src/main/java/org/apache/phoenix/optimize/QueryOptimizer.java
@@ -210,8 +210,9 @@ public class QueryOptimizer {
 
     private List<QueryPlan> getApplicablePlansForSingleFlatQuery(QueryPlan dataPlan, PhoenixStatement statement, List<? extends PDatum> targetColumns, ParallelIteratorFactory parallelIteratorFactory, boolean stopAtBestPlan) throws SQLException {
         SelectStatement select = (SelectStatement)dataPlan.getStatement();
-        // Exit early if we have a point lookup as we can't get better than that
-        if (dataPlan.getContext().getScanRanges().isPointLookup()
+        String indexHint = select.getHint().getHint(Hint.INDEX);
+        // Exit early if we have a point lookup w/o index hint as we can't get better than that
+        if (indexHint == null && dataPlan.getContext().getScanRanges().isPointLookup()
                 && stopAtBestPlan && dataPlan.isApplicable()) {
             return Collections.<QueryPlan> singletonList(dataPlan);
         }
@@ -246,6 +247,10 @@ public class QueryOptimizer {
                 return Collections.singletonList(hintedPlan);
             }
             plans.add(0, hintedPlan);
+        }
+        if (dataPlan.getContext().getScanRanges().isPointLookup()
+                && stopAtBestPlan && dataPlan.isApplicable()) {
+            return Collections.<QueryPlan> singletonList(dataPlan);
         }
         
         for (PTable index : indexes) {
--- a/phoenix-core-client/src/main/java/org/apache/phoenix/optimize/QueryOptimizer.java
+++ b/phoenix-core-client/src/main/java/org/apache/phoenix/optimize/QueryOptimizer.java
@@ -248,10 +248,6 @@ public class QueryOptimizer {
             }
             plans.add(0, hintedPlan);
         }
-        if (dataPlan.getContext().getScanRanges().isPointLookup()
-                && stopAtBestPlan && dataPlan.isApplicable()) {
-            return Collections.<QueryPlan> singletonList(dataPlan);
-        }
         
         for (PTable index : indexes) {
             QueryPlan plan = addPlan(statement, translatedIndexSelect, index, targetColumns, parallelIteratorFactory, dataPlan, false);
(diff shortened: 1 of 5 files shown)


phoenix-core/src/it/java/org/apache/phoenix/end2end/TenantSpecificTablesDDLIT.java
```diff
@@ -616,6 +621,52 @@ public void testTableMetadataScan() throws Exception {
...
+            String physicalViewIndexTableName = MetaDataUtil.getViewIndexPhysicalName(fullDataTableName);
+            TableName viewIndexHBaseTable = TableName.valueOf(physicalViewIndexTableName);
+            TestUtil.assertRawRowCount(conn, viewIndexHBaseTable, 2);
+            String sql = "SELECT /*+ INDEX(" + fullGrandChildViewName + " " + viewIndexName + ")*/ "
+                    + "val2, id2, val1, id3, id1 FROM " + fullGrandChildViewName
+                    + " WHERE id2 = 'a2' AND (id1 = 'a1' OR id1 = 'b1') AND id3 = 3";
+            ResultSet rs = stmt.executeQuery("EXPLAIN " + sql);
+            String actualQueryPlan = QueryUtil.getExplainPlan(rs);
+            String expectedQueryPlan = "CLIENT PARALLEL 1-WAY POINT LOOKUP ON 2 KEYS OVER "
+                    + physicalViewIndexTableName;
```
virajjasani, line 663: Could you please also use the same query without hint and assert that the Explain plan now uses 1-WAY POINT LOOKUP on the base table?

virajjasani (commented):
Left one nit, looks good otherwise. The changes are much simplified.

phoenix-core/src/it/java/org/apache/phoenix/end2end/RowValueConstructorOffsetIT.java
```diff
@@ -1221,4 +1222,29 @@ public void rvcOffsetTrailingNullKeyTest() throws Exception {
...
+        try (Statement statement = conn.createStatement()){
+            ResultSet rs = statement.executeQuery("EXPLAIN " + sql);
+            String actualQueryPlan = QueryUtil.getExplainPlan(rs);
+            // As hinted plan is not applicable so use data plan which is point lookup
+            assertTrue(actualQueryPlan.contains("POINT LOOKUP ON 1 KEY OVER " + TABLE_NAME));
+        }
+    }
+
+    @Test
+    public void testRVCOffsetWithNotApplicableAndPointLookup() throws Exception {
```
kadirozde, line 1240: is this test applicable to this PR?

phoenix-core-client/src/main/java/org/apache/phoenix/optimize/QueryOptimizer.java
```diff
@@ -247,6 +248,10 @@ private List<QueryPlan> getApplicablePlansForSingleFlatQuery(QueryPlan dataPlan,
             }
             plans.add(0, hintedPlan);
         }
+        if (dataPlan.getContext().getScanRanges().isPointLookup()
+                && stopAtBestPlan && dataPlan.isApplicable()) {
+            return Collections.<QueryPlan> singletonList(dataPlan);
+        }
```
tkhurana, line 254: Why is this logic added here ? The dataPlan is already added to the list of plans. This feels like code smell.

virajjasani: I was thinking about the same.

It seems that we might have more plans in the list of plans than just data plan, so if we return single plan here as data plan, it would help cover this case: Even if the hinted index is not applicable, and the data table has point lookup, we return best plan as point lookup on data table.

sanjeet006py: Yes, this fix honors index hint even if original data plan is point lookup. But if hinted plan is null or not applicable then we need to return original data plan as we are already maximum optimized.

phoenix-core/src/it/java/org/apache/phoenix/end2end/RowValueConstructorOffsetIT.java
sanjeet006py: Yes, I added this test as I am doing changes for point lookup case and wanted to assert that I am not breaking the scenario where its point lookup but no query plan is applicable. Just wanted to be extra sure.

phoenix-core-client/src/main/java/org/apache/phoenix/optimize/QueryOptimizer.java
tkhurana: The optimizer will select the optimal plan from the list of plans. You can test by commenting this logic. There is no need to duplicate code. It is confusing to maintain. Same logic is in multiple places.

sanjeet006py: Got it, thanks. Checking

sanjeet006py: @tkhurana I see that if we remove this block then code goes on to create query plan for each index of the original data table. And, to create query plan for each index at least 1 RPC call is done so, I think this can be expensive when the end result we know is original data plan but we will still do RPC calls for other indexes.

tkhurana: But this will not be a common scenario. Most of the time index hint will be null and you will return early. If index hint is not null then most of the times you will use the hinted index plan. It will be rare that you will end up here that you have to exit early.

sanjeet006py: I have addressed the suggestion and also added an IT for partial index. Please take a look. Thanks

kadirozde (approved):

virajjasani (approved):
+1

tkhurana (approved):

PR #1853, opened 2024-03-08
PHOENIX-7185 Error message in SequenceNotFoundException is misleading

anchal246:


stoty (changes requested):
This looks good, but please fix the checkstyle and formatting errors.

PR #1802, opened 2024-01-22
PHOENIX-7008: Addressing Jira spec and review feedback changes

haridsv:
Address the changes in the CREATE CDC spec in PHOENIX-7001
Also includes some review feedback changes across the changes in PRs #1681, #1703 and #1766

PR #1703, opened 2023-10-06
PHOENIX-7008 Tweaks, fixes and additional test coverage for CREATE CDC

haridsv:
Follow up PR for #1681

Jira: [PHOENIX-7008](https://issues.apache.org/jira/browse/PHOENIX-7008)

PR #1866, opened 2024-03-29
PHOENIX-7001: Initial implementation of Change Data Capture (CDC) feature

haridsv:
This commit includes a squash of all the below changes from all the CDC feature PRs and further narrows the diff down by removing non-functional changes for the ease of review. On top of it, the changes have been reworked for the changed state of master, especially the new submodules.

* 4c9827a43 Hari.. PHOENIX-7008 Shallow grammar support for CREATE CDC (#1662)
* e5220e01e TheN..  PHOENIX-7054 Shallow grammar support for DROP CDC and ALTER CDC  (#1694)
* 581e61365 Hari.. PHOENIX-7008 Implementation for CREATE CDC  (#1681)
* e2ef88603 Hari.. PHOENIX-7008 Tweaks, fixes and additional test coverage for CREATE CDC (#1703)
* 5d3fd4033 TheN.. PHOENIX-7074 DROP CDC Implementation (#1713)
* 7420443d8 Hari.. PHOENIX-7014: Query compiler/optimizer changes along with some PHOENIX-7015 changes (#1766)
* da6ddad04 Kadi.. Add an extra delete mutation for CDC
* 93d586efb Kadi.. Add an extra delete mutation during rebuild for CDC index
* f07898f1f Hari.. PHOENIX-7008: Addressing Jira spec and review feedback changes (#1802)
* 87a2ea101 Hari.. PHOENIX-7008: Fix for parser gap and fix for failing test (#1812)
* e395780e9 TheN.. PHOENIX-7015 Implementing CDCGlobalIndexRegionScanner (#1813)


phoenix-core/src/it/java/org/apache/phoenix/end2end/CDCBaseIT.java
```diff
@@ -0,0 +1,655 @@
...
+                                    Integer nSaltBuckets, PTable.IndexType indexType) throws Exception {
+        // For CDC, multitenancy gets derived automatically via the parent table.
+        createTable(conn, cdc_sql, encodingScheme, false, nSaltBuckets, indexType, false, null);
+        String schemaName = SchemaUtil.getSchemaNameFromFullName(tableName);
+        tableName = SchemaUtil.getTableNameFromFullName(tableName);
+        IndexToolIT.runIndexTool(false, schemaName, tableName,
+                "\""+CDCUtil.getCDCIndexName(cdcName)+"\"");
+        String indexFullName = SchemaUtil.getTableName(schemaName,
+                CDCUtil.getCDCIndexName(cdcName));
+        TestUtil.waitForIndexState(conn, indexFullName, PIndexState.ACTIVE);
```
kadirozde, line 187: Is this necessary? IndexTool runs and waits for the job to complete.

phoenix-core-client/src/main/java/org/apache/phoenix/util/CDCUtil.java
```diff
@@ -0,0 +1,118 @@
...
+            cdcChangeScopes = StringUtils.join(",", tmpStream);
+        }
+        return cdcChangeScopes;
+    }
+
+    public static String getCDCIndexName(String cdcName) {
+        return CDC_INDEX_PREFIX + SchemaUtil.getTableNameFromFullName(cdcName.toUpperCase());
+    }
+
+    public static boolean isCDCIndex(String indexName) {
```
tkhurana, line 87: Seems like this API assumes that the indexName doesn't have the schema qualifier ?

phoenix-core-client/src/main/java/org/apache/phoenix/util/CDCUtil.java
```diff
@@ -0,0 +1,118 @@
...
+        return indexName.startsWith(CDC_INDEX_PREFIX);
+    }
+
+    public static boolean isCDCIndex(PTable indexTable) {
+        return isCDCIndex(indexTable.getTableName().getString());
+    }
+
+    public static Scan initForRawScan(Scan scan) {
```
tkhurana, line 95: A better name for this API could be initializeScanForCDC or setupScanForCDC because it is doing more than just setting the raw attribute.

phoenix-core/src/it/java/org/apache/phoenix/end2end/CDCDefinitionIT.java
```diff
@@ -0,0 +1,328 @@
...
+            assertEquals(SQLExceptionCode.TABLE_UNDEFINED.getErrorCode(), e.getErrorCode());
+        }
+
+        cdc_sql = "CREATE CDC " + cdcName + " ON " + tableName;
+        createCDCAndWait(conn, tableName, cdcName, cdc_sql, null, null, null);
+        assertCDCState(conn, cdcName, null, 3);
+        assertNoResults(conn, cdcName);
+
+        try {
+            conn.createStatement().execute(cdc_sql);
```
tkhurana, line 94: Can I create multiple cdc's on the same table with different cdc name but same change scope ?

phoenix-core/src/it/java/org/apache/phoenix/end2end/CDCDefinitionIT.java
```diff
@@ -0,0 +1,328 @@
...
+        } catch (SQLException e) {
+            assertEquals(SQLExceptionCode.TABLE_ALREADY_EXIST.getErrorCode(), e.getErrorCode());
+            assertTrue(e.getMessage().endsWith(cdcName));
+        }
+
+        conn.createStatement().execute("CREATE CDC IF NOT EXISTS " + cdcName + " ON " + tableName +
+                " INCLUDE (pre, post) INDEX_TYPE=g");
+
+        cdcName = generateUniqueName();
+        cdc_sql = "CREATE CDC " + cdcName + " ON " + tableName + " INCLUDE (pre, post)";
```
tkhurana, line 105: Instead of using hard coded strings PRE, POST use the constants defined in the code

phoenix-core/src/it/java/org/apache/phoenix/end2end/CDCDefinitionIT.java
```diff
@@ -0,0 +1,328 @@
...
+                                null, false, saltingConfig[0], false, null);
+                assertSaltBuckets(conn, tableName, saltingConfig[0]);
+
+                String cdcName = generateUniqueName();
+                String cdc_sql = "CREATE CDC " + cdcName + " ON " + tableName;
+                createCDCAndWait(conn, tableName, cdcName, cdc_sql, null,
+                        saltingConfig[1], null);
+                try {
+                    assertCDCState(conn, cdcName, null, 3);
+                    // Index inherits table salt buckets.
```
tkhurana, line 154: This comment is not right.

phoenix-core/src/it/java/org/apache/phoenix/end2end/CDCBaseIT.java
haridsv: This was copied from existing test code (e.g, see [1](https://github.com/apache/phoenix/blob/1e96a2756eaf0a2201a50579789190e8c10747df/phoenix-core/src/it/java/org/apache/phoenix/end2end/index/BaseIndexIT.java#L1516) [2](https://github.com/apache/phoenix/blob/1e96a2756eaf0a2201a50579789190e8c10747df/phoenix-core/src/it/java/org/apache/phoenix/end2end/index/BaseIndexWithRegionMovesIT.java#L1711)). But now, I see plenty of other places where this wasn't used, so I will try without and remove if it is working fine.

phoenix-core-client/src/main/java/org/apache/phoenix/util/CDCUtil.java
haridsv: Good point! It started purely for setting raw scan attributes so the name was appropriate back then, but your suggested name sounds good.

phoenix-core/src/it/java/org/apache/phoenix/end2end/CDCDefinitionIT.java
haridsv: Yes, the CDC objects are completely independent of each other, with their own change scopes, so they can use same or different change scopes.

phoenix-core/src/it/java/org/apache/phoenix/end2end/CDCDefinitionIT.java
haridsv: Sure, just to confirm, you are referring to the line 107, correct? I would like to leave the usage in 102 and 105 as is.

phoenix-core/src/it/java/org/apache/phoenix/end2end/CDCDefinitionIT.java
haridsv: It is meant for the next line, will move it.

phoenix-core-client/src/main/java/org/apache/phoenix/util/CDCUtil.java
haridsv: Yes, that is right. There are 2 calls to this right now, 1 is from the overloaded method that works on PTable and it is passing the name without schema qualifier. The other call comes from the "DROP INDEX" where you won't expect a schema qualifier anyway.

kadirozde:
I found usage of maps to keep track of mutations and pre and post images are too complex in ITs. The test code should be easy to read and verify. We need a list of mutations and then check the result CDC against this list of mutations. For example, each mutation can be modeled as a timestamp, mutation type, list of column names and list of values represented as strings. This list can be initialized when it is constructed. Creating upsert and delete statements from this mutation list should be straightforward. Comparing the result of CDC against this list also should simple. The pre and post images can be retrived from the data table using SCN connections and verified against the CDC result.

phoenix-core-server/src/main/java/org/apache/phoenix/coprocessor/CDCGlobalIndexRegionScanner.java
```diff
@@ -0,0 +1,274 @@
...
+        CellBuilder builder = CellBuilderFactory.create(CellBuilderType.SHALLOW_COPY);
+        Result cdcRow = Result.create(Arrays.asList(builder
+                .setRow(indexRowKey)
+                .setFamily(ImmutableBytesPtr.cloneCellFamilyIfNecessary(firstCell))
+                .setQualifier(cdcDataTableInfo.getCdcJsonColQualBytes())
+                .setTimestamp(changeBuilder.getChangeTimestamp())
+                .setValue(value)
+                .setType(Cell.Type.Put)
+                .build()));
+        return cdcRow;
```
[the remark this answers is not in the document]
virajjasani: We also use JacksonUtil at some places: https://github.com/apache/phoenix/blob/master/phoenix-core-client/src/main/java/org/apache/phoenix/util/JacksonUtil.java

pom.xml
```diff
@@ -1756,6 +1757,11 @@
         <artifactId>javax.el</artifactId>
         <version>${glassfish.el.version}</version>
       </dependency>
+      <dependency>
+        <groupId>com.google.code.gson</groupId>
+        <artifactId>gson</artifactId>
+        <version>${gson.version}</version>
+      </dependency>
```
virajjasani, line 1764: This can be avoided altogether if we use JacksonUtil

phoenix-core-client/src/main/java/org/apache/phoenix/util/CDCChangeBuilder.java
```diff
@@ -0,0 +1,151 @@
...
+        } else if (cell.getTimestamp() == changeTimestamp) {
+            assert !isDeletionEvent() : "Not expected to find a change for delete event";
+            changeType = CDC_UPSERT_EVENT_TYPE;
+            if (isChangeImageInScope || isPostImageInScope) {
+                changeImage.put(cdcColumnName, value);
+            }
+        }
+    }
+
+    public Map buildCDCEvent() {
```
virajjasani, line 124: All the places where raw `Map` is used, let's replace it with `Map<String, Object>`

phoenix-core/src/it/java/org/apache/phoenix/end2end/CDCQueryIT.java
```diff
@@ -0,0 +1,667 @@
...
+        String cdcName, cdc_sql;
+        String schemaName = withSchemaName ? generateUniqueName() : null;
+        String tableName = SchemaUtil.getTableName(schemaName, generateUniqueName());
+        try (Connection conn = newConnection()) {
+            createTable(conn, "CREATE TABLE  " + tableName + " (" +
+                    (multitenant ? "TENANT_ID CHAR(5) NOT NULL, " : "") +
+                    "k INTEGER NOT NULL, a_binary binary(10), d Date, t TIMESTAMP, " +
+                    "CONSTRAINT PK PRIMARY KEY " +
+                    (multitenant ? "(TENANT_ID, k) " : "(k)") + ")", encodingScheme, multitenant,
+                    tableSaltBuckets, false, null);
```
virajjasani, line 575: Let's also add `VARBINARY` test?

virajjasani:
Now that JSON PR is merged, we can also address it with changes/pre/post image and include some tests for JSON data.

phoenix-core-server/src/main/java/org/apache/phoenix/coprocessor/CDCGlobalIndexRegionScanner.java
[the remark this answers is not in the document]
haridsv: I switched from using Gson to Jackson.

kadirozde (approved):
+1, Thanks!

kadirozde:
> Now that JSON PR is merged, we can also address it with changes/pre/post image and include some tests for JSON data.

Let us open a separate Jira to support complex data types including Array and JSON.

virajjasani:
Belated +1

phoenix-core/src/it/java/org/apache/phoenix/end2end/index/BaseIndexIT.java
```diff
@@ -1514,7 +1514,6 @@ public void testLastDDLTimestampOnAsyncIndexes() throws Exception {
 
             // run the index MR job.
             IndexToolIT.runIndexTool(false, TestUtil.DEFAULT_SCHEMA_NAME, tableName, indexName);
-            TestUtil.waitForIndexState(conn, fullIndexName, PIndexState.ACTIVE);
```
palashc, line 1517: @haridsv Would you remember why this change was made in this PR? This breaks a few tests on my PR https://github.com/apache/phoenix/pull/1883. 
Rushabh made the [change](https://github.com/apache/phoenix/pull/1595/files#diff-6d91af246300bfabfe5d43b94abe177fa549c3dc60d2e25e795e1425a0ff4bb4R1466) originally and I think it is needed. Please let me know if there was a reason to remove it in this PR.

tkhurana: Actually, this change is causing these 3 tests to fail continuously on all PRs on master:
`GlobalImmutableTxIndexWithRegionMovesIT.testLastDDLTimestampOnAsyncIndexes`
`GlobalMutableTxIndexIT.testLastDDLTimestampOnAsyncIndexes`
`GlobalImmutableTxIndexIT.testLastDDLTimestampOnAsyncIndexes`

palashc: I see it was removed based on Kadir's comment here - https://github.com/apache/phoenix/pull/1866#discussion_r1561771543

palashc: @tkhurana Should I add it back to BaseIndexIT and BaseIndexWithRegionMovesIT in my PR? Not sure why the behaviour of IndexTool is different in these 3 cases.

virajjasani: We will have to put this back, we are just discussing this internally.

PR #1871, opened 2024-04-02
PHOENIX 7233 : CQSI openConnection should timeout to unblock other connection threads

Divneet18:


phoenix-core-client/src/main/java/org/apache/phoenix/query/ConnectionQueryServicesImpl.java
```diff
@@ -3501,7 +3503,18 @@ public Void call() throws Exception {
...
-                            openConnection();
+                            CompletableFuture<Void> supplyFuture = CompletableFuture.runAsync(() -> {
+                                try {
+                                    openConnection();
+                                } catch (SQLException e) {
+                                    throw new RuntimeException(e);
+                                }
+                            });
+                            try {
+                                supplyFuture.get(120000, MILLISECONDS);
```
gjacoby126, line 3514: The timeout should be a configurable value

phoenix-core-client/src/main/java/org/apache/phoenix/query/ConnectionQueryServicesImpl.java
```diff
@@ -3501,7 +3503,18 @@ public Void call() throws Exception {
...
+                            } catch (TimeoutException e) {
+                                LOGGER.error("openConnection did not return any value");
```
gjacoby126, line 3516: Why are we swallowing the exception rather than rethrowing it? Shouldn't the timeout be a failure condition rather than proceeding?

gjacoby126 (changes requested):

PR #1872, opened 2024-04-03
PHOENIX-7299 :- ScanningResultIterator should not time out a query after receiving a valid result

lokiore:


phoenix-core-client/src/main/java/org/apache/phoenix/iterate/ScanningResultIterator.java
```diff
@@ -85,6 +86,8 @@ public class ScanningResultIterator implements ResultIterator {
 
     private long dummyRowCounter = 0;
 
+    private AtomicBoolean hasFirstValidResult = new AtomicBoolean(false);
```
kadirozde, line 89: Phoenix result iterators are single threaded. We do not need to use AtomicBoolean.

phoenix-core/src/it/java/org/apache/phoenix/iterate/PhoenixQueryTimeoutIT.java
```diff
@@ -202,6 +204,57 @@ public void testScanningResultIteratorQueryTimeoutForPagingWithVeryLowTimeout()
...
+    public void testScanningResultIteratorShouldNotQueryTimeoutForPagingAfterReceivingAValidRow() throws Exception {
+        //Arrange
+        PreparedStatement ps = loadDataAndPreparePagedQuery(10,1);
+
+        ManualEnvironmentEdge injectEdge;
+        injectEdge = new ManualEnvironmentEdge();
+        injectEdge.setValue(EnvironmentEdgeManager.currentTimeMillis());
+        EnvironmentEdgeManager.injectEdge(injectEdge);
+
+        //Act + Assert
```
kadirozde, line 217: Please add more readable comment here. For example, this query should not timeout as the first row will be retrieved within 10 ms.

phoenix-core/src/it/java/org/apache/phoenix/iterate/PhoenixQueryTimeoutIT.java
```diff
@@ -202,6 +204,57 @@ public void testScanningResultIteratorQueryTimeoutForPagingWithVeryLowTimeout()
...
+            ResultSet rs = ps.executeQuery();
+            while(rs.next()) {
+                injectEdge.incrementValue(10);
+            }
+        } catch (SQLException e) {
+            fail("Query should have run smoothly");
+        } finally {
+            BaseResultIterators.setForTestingSetTimeoutToMaxToLetQueryPassHere(false);
+        }
+
```
kadirozde, line 230: Please add more readable comments here. For example, this query should time out as the query timeout value is set to 0.

kadirozde (approved):
+1, Thanks!

PR #1874, opened 2024-04-05
PHOENIX-7302 Server Paging doesn't work on scans with limit

tkhurana:
Fix paging for scans with limit but no other filter

### DIFF
--- a/phoenix-core-client/src/main/java/org/apache/phoenix/util/ScanUtil.java
+++ b/phoenix-core-client/src/main/java/org/apache/phoenix/util/ScanUtil.java
@@ -1615,14 +1615,21 @@ public class ScanUtil {
     public static PageFilter removePageFilter(Scan scan) {
         Filter filter = scan.getFilter();
         if (filter != null) {
+            PagingFilter pagingFilter = null;
             if (filter instanceof PagingFilter) {
-                filter = ((PagingFilter) filter).getDelegateFilter();
+                pagingFilter = (PagingFilter) filter;
+                filter = pagingFilter.getDelegateFilter();
                 if (filter == null) {
                     return null;
                 }
             }
             if (filter instanceof PageFilter) {
-                scan.setFilter(null);
+                if (pagingFilter != null) {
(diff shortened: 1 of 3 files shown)


phoenix-core-client/src/main/java/org/apache/phoenix/util/ScanUtil.java
```diff
@@ -1615,14 +1615,21 @@ public static boolean isIncompatibleClientForServerReturnValidRowKey(Scan scan)
...
+                filter = pagingFilter.getDelegateFilter();
                 if (filter == null) {
                     return null;
                 }
             }
             if (filter instanceof PageFilter) {
-                scan.setFilter(null);
+                if (pagingFilter != null) {
+                    pagingFilter.setDelegateFilter(null);
+                    scan.setFilter(pagingFilter);
```
virajjasani, line 1629: This helps when filter is of both types: HBase PageFilter and Phoenix PagingFilter (which happens only when LIMIT is present), so that we want to retain PagingFilter for performing page timeouts?

virajjasani (commented):
Looks good overall

phoenix-core-client/src/main/java/org/apache/phoenix/util/ScanUtil.java
tkhurana: Yes, the LIMIT translates to HBase PageFilter and the Phoenix PagingFilter prevents rpc timeouts.

virajjasani:
Really nice find btw

virajjasani (approved):

tkhurana:
Test results look good. 2 tests failed. Known flappers
[ERROR] Failures: 
[ERROR]   WALRecoveryRegionPostOpenIT.testRecoveryRegionPostOpen
[ERROR] org.apache.phoenix.end2end.index.MutableIndexFailureIT.testIndexWriteFailure

kadirozde (approved):
+1, Thanks!

PR #1876, opened 2024-04-09
PHOENIX-7299 (Addendum) :- ScanningResultIterator should not time out a query after receiving a valid result

lokiore:


phoenix-core-client/src/main/java/org/apache/phoenix/compile/StatementContext.java
```diff
@@ -111,6 +113,7 @@ public StatementContext(StatementContext context) {
         this.queryLogger = context.queryLogger;
         this.isClientSideUpsertSelect = context.isClientSideUpsertSelect;
         this.isUncoveredIndex = context.isUncoveredIndex;
+        this.hasFirstValidResult = context.hasFirstValidResult;
```
kadirozde, line 116: This should be this.hasFirstValidResult = new AtomicBoolean(context.getHasFirstValidResult());

kadirozde (approved):
+1, Thanks!

PR #1880, opened 2024-04-12
PHOENIX-7306 Metadata lookup should be permitted only within query timeout

virajjasani:
Jira: PHOENIX-7306

### DIFF
--- a/phoenix-core-client/src/main/java/org/apache/phoenix/query/ConnectionQueryServices.java
+++ b/phoenix-core-client/src/main/java/org/apache/phoenix/query/ConnectionQueryServices.java
@@ -99,22 +99,27 @@ public interface ConnectionQueryServices extends QueryServices, MetaDataMutated 
     public TableDescriptor getTableDescriptor(byte[] tableName) throws SQLException;
 
     public HRegionLocation getTableRegionLocation(byte[] tableName, byte[] row) throws SQLException;
-    public List<HRegionLocation> getAllTableRegions(byte[] tableName) throws SQLException;
+
+    public List<HRegionLocation> getAllTableRegions(byte[] tableName, int queryTimeout)
+            throws SQLException;
 
     /**
      * Retrieve table region locations that cover the startRowKey and endRowKey. The start key
      * of the first region of the returned list must be less than or equal to startRowKey.
      * The end key of the last region of the returned list must be greater than or equal to
-     * endRowKey.
+     * endRowKey. The operation to retrieve the table region locations must be completed within
+     * the query timeout.
      *
-     * @param tableName Table name.
-     * @param startRowKey Start RowKey.
-     * @param endRowKey End RowKey.
+     * @param tableName    Table name.
+     * @param startRowKey  Start RowKey.
+     * @param endRowKey    End RowKey.
+     * @param queryTimeout Phoenix query timeout.
      * @return The list of region locations that cover the startRowKey and endRowKey key boundary.
      * @throws SQLException If fails to retrieve region locations.
      */
     public List<HRegionLocation> getTableRegions(byte[] tableName, byte[] startRowKey,
-        byte[] endRowKey) throws SQLException;
+                                                 byte[] endRowKey,
+                                                 int queryTimeout) throws SQLException;
 
     public PhoenixConnection connect(String url, Properties info) throws SQLException;
 
--- a/phoenix-core-client/src/main/java/org/apache/phoenix/query/ConnectionQueryServices.java
+++ b/phoenix-core-client/src/main/java/org/apache/phoenix/query/ConnectionQueryServices.java
@@ -128,6 +128,23 @@ public interface ConnectionQueryServices extends QueryServices, MetaDataMutated 
      * Retrieve table region locations that cover the startRowKey and endRowKey. The start key
      * of the first region of the returned list must be less than or equal to startRowKey.
      * The end key of the last region of the returned list must be greater than or equal to
+     * endRowKey.
+     * This method is Deprecated. Use {@link #getTableRegions(byte[], byte[], byte[], int)} instead.
+     *
+     * @param tableName Table name.
+     * @param startRowKey Start RowKey.
+     * @param endRowKey End RowKey.
+     * @return The list of region locations that cover the startRowKey and endRowKey key boundary.
+     * @throws SQLException If fails to retrieve region locations.
+     */
+    @Deprecated
+    public List<HRegionLocation> getTableRegions(byte[] tableName, byte[] startRowKey,
+                                                 byte[] endRowKey) throws SQLException;
+
+    /**
+     * Retrieve table region locations that cover the startRowKey and endRowKey. The start key
+     * of the first region of the returned list must be less than or equal to startRowKey.
+     * The end key of the last region of the returned list must be greater than or equal to
      * endRowKey. The operation to retrieve the table region locations must be completed within
      * the query timeout.
      *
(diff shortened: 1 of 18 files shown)


phoenix-core-client/src/main/java/org/apache/phoenix/query/ConnectionQueryServicesImpl.java
```diff
@@ -739,8 +739,6 @@ public byte[] getNextRegionStartKey(HRegionLocation regionLocation, byte[] curre
             && !Bytes.equals(regionLocation.getRegion().getEndKey(), HConstants.EMPTY_END_ROW);
         if (conditionOne || conditionTwo) {
             GLOBAL_HBASE_COUNTER_METADATA_INCONSISTENCY.increment();
-            String regionNameString =
-                new String(regionLocation.getRegion().getRegionName(), StandardCharsets.UTF_8);
             LOGGER.error(
```
kadirozde, line 742: This should be warning.

virajjasani: Done

kadirozde (approved):

phoenix-core-client/src/main/java/org/apache/phoenix/query/ConnectionQueryServicesImpl.java
```diff
@@ -780,6 +775,10 @@ public List<HRegionLocation> getTableRegions(byte[] tableName, byte[] startRowKe
             config.getInt(PHOENIX_GET_REGIONS_RETRIES, DEFAULT_PHOENIX_GET_REGIONS_RETRIES);
         TableName table = TableName.valueOf(tableName);
         byte[] currentKey = null;
+        final int queryTimeout = this.getProps().getInt(QueryServices.THREAD_TIMEOUT_MS_ATTRIB,
```
tkhurana, line 778: querytimeout can be overridden at the phoenix statement level. We need to get it from the statementcontext `context.getStatement().getQueryTimeoutInMillis();`

phoenix-core-client/src/main/java/org/apache/phoenix/query/ConnectionQueryServicesImpl.java
```diff
@@ -800,6 +799,8 @@ public List<HRegionLocation> getTableRegions(byte[] tableName, byte[] startRowKe
                         && Bytes.compareTo(currentKey, endRowKey) >= 0) {
                         break;
                     }
+                    throwErrorIfQueryTimedOut(startRowKey, endRowKey, maxQueryEndTime,
```
tkhurana, line 802: I think we can test this by setting a low value for a query timeout and creating a table with multiple regions either through salting or split points. PhoenixQueryTimeoutIT seems like a good candidate for this test.

phoenix-core-client/src/main/java/org/apache/phoenix/query/ConnectionQueryServicesImpl.java
tkhurana: If it is not overridden at the statement level the statementcontext API  gets it from cqsi

phoenix-core-client/src/main/java/org/apache/phoenix/query/ConnectionQueryServicesImpl.java
virajjasani: One of the tests in PhoenixQueryTimeoutIT is already covering this. However, the way `BaseResultIterators.setForTestingSetTimeoutToMaxToLetQueryPassHere` is being used is causing issues. We will have to clean it up or separate the test using it to different class. Given that it is static, it has potential to run into issues while running any other test in parallel.

phoenix-core-client/src/main/java/org/apache/phoenix/query/ConnectionQueryServicesImpl.java
virajjasani: Sounds good, done

phoenix-core-client/src/main/java/org/apache/phoenix/query/ConnectionQueryServices.java
```diff
@@ -99,22 +99,27 @@ public interface ConnectionQueryServices extends QueryServices, MetaDataMutated
     public TableDescriptor getTableDescriptor(byte[] tableName) throws SQLException;
 
     public HRegionLocation getTableRegionLocation(byte[] tableName, byte[] row) throws SQLException;
-    public List<HRegionLocation> getAllTableRegions(byte[] tableName) throws SQLException;
+
+    public List<HRegionLocation> getAllTableRegions(byte[] tableName, int queryTimeout)
```
tkhurana, line 103: I would recommend not changing the public API but add a new API. We had similar problems in the past when we introduced a timeout in ScanningResultIterator. The existing APIs can send values like LONG.max.

virajjasani: The clients using the API either have StatementContext object or they have Config object so either way they need to provide timeout anyway.
If we keep the API as is, anyone would only use that directly even though they might have PhoenixStatement level overridden timeout. So using public API without timeout is not good practice anyway right?

For ScanningResultIterator, the constructor has timeout param and the clients are either providing Long.MAX_VALUE or Context level overridden timeout but the pattern is same i.e. clients using the APIs are providing timeout.

tkhurana: Sorry, it is TableResultIterator which broke when we modified its constructor. This was fixed via https://issues.apache.org/jira/browse/PHOENIX-7102

tkhurana: The point is we need to maintain API compatibility

virajjasani: I see, sounds good. If we maintain the old API contract, we still want it to figure out the timeout value from the config right?

Let me also mark the old API deprecated.

tkhurana: That is even better. The old API can fetch the timeout from config instead of using Long.max. That would have saved you some test changes where you are doing exactly that.

virajjasani: I think many of those test changes will be required because many of them don't have real config as such, we are either mocking or using constant. 

No problem, this can be taken care of in general. Though we have this test recently written (this month) that is getting stuck and not moving forward. This seems bit of a concern, either for using static test method directly on BaseResultIterator is becoming problematic or EnvironmentEdgeManager is causing troubles. Either way, the test might be blocking the whole build.

virajjasani:
The latest build: https://ci-hadoop.apache.org/job/Phoenix/job/Phoenix-PreCommit-GitHub-PR/job/PR-1880/7/

phoenix-core-client/src/main/java/org/apache/phoenix/query/ConnectionQueryServices.java
```diff
@@ -99,22 +99,48 @@ public interface ConnectionQueryServices extends QueryServices, MetaDataMutated
...
+     * @param startRowKey  Start RowKey.
+     * @param endRowKey    End RowKey.
+     * @param queryTimeout Phoenix query timeout.
      * @return The list of region locations that cover the startRowKey and endRowKey key boundary.
      * @throws SQLException If fails to retrieve region locations.
      */
     public List<HRegionLocation> getTableRegions(byte[] tableName, byte[] startRowKey,
-        byte[] endRowKey) throws SQLException;
+                                                 byte[] endRowKey,
+                                                 int queryTimeout) throws SQLException;
```
tkhurana, line 143: We should keep the existing API for this too and mark it deprecated just like the other one.

tkhurana:
@virajjasani One comment, other than that looks good.

tkhurana (approved):

virajjasani:
Will merge this in sometime, need to confirm whether the patch can be directly applied on 5.2 branch.

PR #1882, opened 2024-04-13
PHOENIX-7305 : Invalidate server metadata cache when dropping indexes as part of DROP column

palashc:


phoenix-core-client/src/main/java/org/apache/phoenix/jdbc/PhoenixStatement.java
```diff
@@ -464,14 +464,6 @@ private PhoenixResultSet executeQuery(final CompilableStatement stmt,
                                 LOGGER.debug("Force updating client metadata cache for {}",
                                         ValidateLastDDLTimestampUtil.getInfoString(tenantId,
                                                 Arrays.asList(getLastQueryPlan().getTableRef())));
-
```
palashc, line 467: A remnant of old design I believe, should not be needed now.

phoenix-core/src/test/java/org/apache/phoenix/cache/ServerMetadataCacheIT.java
```diff
@@ -891,12 +891,14 @@ public void testSelectQueryAfterAlterIndex() throws Exception {
             stmt.executeQuery("SELECT k FROM " + tableName + " WHERE v1=1");
             Assert.assertEquals("Query on secondary key should have used index.", indexName, stmt.getQueryPlan().getTableRef().getTable().getTableName().toString());
 
-            //verify client-2 cache was updated with the index's base table metadata
```
palashc, line 894: change needed because earlier I was calling MetadataClient.updateCache for the parent table but the correct way is to call it for the index.

phoenix-core-server/src/main/java/org/apache/phoenix/coprocessor/MetaDataEndpointImpl.java
```diff
@@ -3904,6 +3904,13 @@ private MetaDataMutationResult dropIndexes(RegionCoprocessorEnvironment env, PTa
                 // that a change in index state doesn't
                 // occur while we're dropping it.
                 acquireLock(region, indexKey, locks);
+                // invalidate server metadata cache when dropping index
+                List<InvalidateServerMetadataCacheRequest> requests = new ArrayList<>();
+                requests.add(new InvalidateServerMetadataCacheRequest(tenantId,
```
shahrs87, line 3909: We are invalidating server cache for table in MDEI#mutateColumn [here](https://github.com/apache/phoenix/blob/PHOENIX-6883-feature/phoenix-core-server/src/main/java/org/apache/phoenix/coprocessor/MetaDataEndpointImpl.java#L3296-L3299).

@palashc  Do you mean to invalidate the indexes?

palashc: @shahrs87 Yes, we need to invalidate the entry for any indexes which we are dropping here. Are you suggesting I change the comment here?

palashc: @shahrs87 Pushed the change as discussed offline.

phoenix-core/src/test/java/org/apache/phoenix/cache/ServerMetadataCacheIT.java
shahrs87: > change needed because earlier I was calling MetadataClient.updateCache for the parent table but the correct way is to call it for the index.

Where is this change being made?  @palashc

palashc: PhoenixStatement.java, there are only 3 changes in this PR

shahrs87 (approved):
+1 lgtm.

PR #1748, opened 2023-12-04
PHOENIX-7115 Create separate handler thread pool for invalidating server metadata cache

shahrs87:

PR #1883, opened 2024-04-24
PHOENIX-6883 : Phoenix Metadata Caching Redesign

palashc:
JIRA - https://issues.apache.org/jira/browse/PHOENIX-6883

1. Introduce a region server endpoint for phoenix and server side metadata caching on region servers. 
2. Server side cache kept consistent by invalidating cache entries synchronously during a DDL operation.
3. Client sets `UPDATE_CACHE_FREQUENCY=never` and uses LAST_DDL_TIMESTAMP to validate metadata before query/upsert. 
4. Client refreshes its cache using the `getTable` RPC only when it receives a stale cache exception from the server.


phoenix-core-client/src/main/java/org/apache/phoenix/query/ConnectionQueryServicesImpl.java
```diff
@@ -6248,4 +6320,170 @@ public synchronized PhoenixTransactionClient initTransactionClient(Provider prov
...
+     * @param isRetry
+     * @throws Throwable
+     */
+    private void invalidateServerMetadataCacheWithRetries(Admin admin,
+            Collection<ServerName> serverNames,
+            List<InvalidateServerMetadataCacheRequest> invalidateCacheRequests,
+            boolean isRetry) throws Throwable {
+        RegionServerEndpointProtos.InvalidateServerMetadataCacheRequest protoRequest =
+                getRequest(invalidateCacheRequests);
+        // TODO Do I need my own executor or can I re-use QueryServices#Executor
```
palashc, line 6380: This came up in discussion today and the recommendation is to create a separate thread pool for executing cache invalidations. I can create a subtask for the same and add the changes to this PR.
@shahrs87 @tkhurana @kadirozde

kadirozde: Assuming that these new methods are only called at the server side, I think we should move this code to a separate or existing server side class.

palashc: Since the cache invalidation code is in CQSI and we would want to avoid creating a new thread pool for every DDL operation, how do you suggest we save the executor? We can have an `executor` in CQSI but we would not want client side CQSI to create an executor unnecessarily - or is that okay?

palashc: @kadirozde This code did start out in a separate class but I remember @shahrs87 and @jpisaac had some reasons to move it to CQSI. I could not find which PR that was. @shahrs87 Can you recall those details?

palashc: This was the PR - https://github.com/apache/phoenix/pull/1748
@kadirozde Please see comments here https://github.com/apache/phoenix/pull/1748#discussion_r1419676088

phoenix-core-client/src/main/java/org/apache/phoenix/util/PhoenixRuntime.java
```diff
@@ -539,6 +541,18 @@ public static PTable getTable(Connection conn, @Nullable String tenantId, String
         return pconn.getTable(tenantId, fullTableName, timestamp);
     }
 
+    /**
+     * Returns the most recent PTable fetched from the server without updating the CQSI cache.
+     */
+    public static PTable getTableFromServerNoCache(Connection conn, byte[] schemaName,
```
tkhurana, line 547: The API signature is not consistent with other getTable APIs in this class.

phoenix-core-client/src/main/java/org/apache/phoenix/compile/UpsertCompiler.java
```diff
@@ -636,7 +636,7 @@ public MutationPlan compile(UpsertStatement upsert) throws SQLException {
                 // as max TS, so that the query can safely restarted and still work of a snapshot
                 // (so it won't see its own data in case of concurrent splits)
                 // see PHOENIX-4849
-                long serverTime = selectResolver.getTables().get(0).getCurrentTime();
+                long serverTime = selectResolver.getTables().get(0).getTimeStamp();
```
tkhurana, line 639: What is the difference between getCurrentTime() and getTimeStamp() ?

phoenix-core-client/src/main/java/org/apache/phoenix/cache/ServerMetadataCacheImpl.java
```diff
@@ -0,0 +1,165 @@
...
+
+    protected Configuration conf;
+    // key is the combination of <tenantID, schema name, table name>, value is the lastDDLTimestamp
+    protected final Cache<ImmutableBytesPtr, Long> lastDDLTimestampMap;
+    private static final Logger LOGGER = LoggerFactory.getLogger(ServerMetadataCacheImpl.class);
+    private static final String PHOENIX_COPROC_REGIONSERVER_CACHE_TTL_MS =
+            "phoenix.coprocessor.regionserver.cache.ttl.ms";
+    // Keeping default cache expiry for 30 mins since we won't have stale entry
+    // for more than 30 mins.
+    private static final long DEFAULT_PHOENIX_COPROC_REGIONSERVER_CACHE_TTL_MS
```
tkhurana, line 56: Does this mean we will go to syscat after 30 minutes to get the lastDDLTimestamp ? Isn't 30 mins to less ?

phoenix-core-client/src/main/java/org/apache/phoenix/cache/ServerMetadataCacheImpl.java
```diff
@@ -0,0 +1,165 @@
...
+                DEFAULT_PHOENIX_COPROC_REGIONSERVER_CACHE_SIZE);
+        lastDDLTimestampMap = CacheBuilder.newBuilder()
+                .removalListener((RemovalListener<ImmutableBytesPtr, Long>) notification -> {
+                    String key = notification.getKey().toString();
+                    LOGGER.debug("Expiring " + key + " because of "
+                            + notification.getCause().name());
+                })
+                // maximum number of entries this cache can handle.
+                .maximumSize(maxSize)
+                .expireAfterAccess(maxTTL, TimeUnit.MILLISECONDS)
```
tkhurana, line 99: We should consider adding some jitter here otherwise all the RS will expire roughly around the same time.

phoenix-core-client/src/main/java/org/apache/phoenix/cache/ServerMetadataCacheImpl.java
palashc: @tkhurana I will move it to the config so that we can change it and also increase the default. What do you think would be a good default value? Since we do not run DDL operations very often, maybe we could keep it to a larger value like 1 day or even more.

phoenix-core-client/src/main/java/org/apache/phoenix/compile/UpsertCompiler.java
palashc: @tkhurana Currently both of those methods return the same value i.e. `upperBoundTimestamp` which is either -1 or whatever is provided to the TableRef constructor. Current code:
```
// if UPDATE_CACHE_FREQUENCY is set, always let the server set timestamps
this.upperBoundTimeStamp = table.getUpdateCacheFrequency()!=0 ? QueryConstants.UNSET_TIMESTAMP : upperBoundTimeStamp;
this.currentTime = this.upperBoundTimeStamp;

public long getTimeStamp() {
        return this.upperBoundTimeStamp;
}
public long getCurrentTime() {
        return this.currentTime;
 }
```
Once we set UPDATE_CACHE_FREQ to NEVER, currentTime was being set to -1 and features using currentTime were breaking. This change keeps currentTime to whatever is provided to the TableRef constructor so that features using currentTime can continue to use it. There is effectively no change in UpsertCompiler - it will still be using the same value i.e. TableRef.upperBoundTimeStamp.
```
this.currentTime = upperBoundTimeStamp;
```

phoenix-core-client/src/main/java/org/apache/phoenix/cache/ServerMetadataCacheImpl.java
palashc: @tkhurana Access time is reset by a read operation and since clients can pick regionservers at random, can we expect expiry of entries to also be a bit random?

tkhurana: @palashc Actually I don't see how this ttl will help. It will be reset on every access so for popular entries the ttl expiry will never kick in. And for less popular entries your size limit will take care of evicting them. I would recommend just keeping the size limit.

phoenix-core-client/src/main/java/org/apache/phoenix/cache/ServerMetadataCacheImpl.java
```diff
@@ -0,0 +1,165 @@
...
+     * If not found in cache, then query SYSCAT regionserver.
+     * @param tenantID tenant id
+     * @param schemaName schema name
+     * @param tableName table name
+     * @return last DDL timestamp
+     * @throws Exception
+     */
+    public long getLastDDLTimestampForTable(byte[] tenantID, byte[] schemaName, byte[] tableName)
+            throws SQLException {
+        String fullTableNameStr = SchemaUtil.getTableName(schemaName, tableName);
```
tkhurana, line 114: This seems to be used only for logging. We could eliminate this call and the log itself or just log the byte[] formatted as strings.

phoenix-core-client/src/main/java/org/apache/phoenix/cache/ServerMetadataCacheImpl.java
palashc: That makes sense!

phoenix-core-client/src/main/java/org/apache/phoenix/cache/ServerMetadataCacheImpl.java
```diff
@@ -0,0 +1,165 @@
...
+    /**
+     * Invalidate cache for the given tenantID, schema name and table name.
+     * Guava cache is thread safe so we don't have to synchronize it explicitly.
+     * @param tenantID tenantID
+     * @param schemaName schemaName
+     * @param tableName tableName
+     */
+    public void invalidate(byte[] tenantID, byte[] schemaName, byte[] tableName) {
+        String fullTableNameStr = SchemaUtil.getTableName(schemaName, tableName);
+        LOGGER.debug("Invalidating server metadata cache for tenantID: {}, full table: {}",
```
tkhurana, line 155: I would make this log INFO.

phoenix-core-client/src/main/java/org/apache/phoenix/cache/ServerMetadataCacheImpl.java
palashc: Resolving, since we will remove this ttl.

phoenix-core-client/src/main/java/org/apache/phoenix/compile/FromCompiler.java
```diff
@@ -180,7 +180,8 @@ public static ColumnResolver getResolverForCreation(final CreateTableStatement s
         NamedTableNode tableNode = NamedTableNode.create(null, baseTable, Collections.<ColumnDef>emptyList());
         // Always use non-tenant-specific connection here
         try {
-            SingleTableColumnResolver visitor = new SingleTableColumnResolver(connection, tableNode, true);
+            SingleTableColumnResolver visitor
```
tkhurana, line 183: Why are we always hitting the server when resolving the table ?

phoenix-core-client/src/main/java/org/apache/phoenix/compile/FromCompiler.java
```diff
@@ -280,7 +281,8 @@ public static ColumnResolver getResolver(NamedTableNode tableNode, PhoenixConnec
 
     public static ColumnResolver getResolver(SingleTableStatement statement, PhoenixConnection connection)
             throws SQLException {
-        SingleTableColumnResolver visitor = new SingleTableColumnResolver(connection, statement.getTable(), true);
+        SingleTableColumnResolver visitor
+                = new SingleTableColumnResolver(connection, statement.getTable(), true, true);
```
tkhurana, line 285: Why are we setting alwaysHitServer to true ?

phoenix-core-client/src/main/java/org/apache/phoenix/compile/FromCompiler.java
palashc: @tkhurana This creates a resolver for table creation. As discussed in PHOENIX-7271, we will always resolve before DDL operations.

phoenix-core-client/src/main/java/org/apache/phoenix/compile/FromCompiler.java
palashc: @tkhurana This resolver is also used in several DDL paths. 
We create resolvers for query/mutations in `getResolverForQuery` and `getResolverForMutation` and that is where we set it to false.

phoenix-core-server/src/main/java/org/apache/phoenix/coprocessor/PhoenixRegionServerEndpoint.java
```diff
@@ -0,0 +1,119 @@
...
+        for (RegionServerEndpointProtos.LastDDLTimestampRequest lastDDLTimestampRequest
+                : request.getLastDDLTimestampRequestsList()) {
+            byte[] tenantID = lastDDLTimestampRequest.getTenantId().toByteArray();
+            byte[] schemaName = lastDDLTimestampRequest.getSchemaName().toByteArray();
+            byte[] tableName = lastDDLTimestampRequest.getTableName().toByteArray();
+            long clientLastDDLTimestamp = lastDDLTimestampRequest.getLastDDLTimestamp();
+            String tenantIDStr = Bytes.toString(tenantID);
+            String fullTableName = SchemaUtil.getTableName(schemaName, tableName);
+            try {
+                LOGGER.debug("Verifying last ddl timestamp for tenantID: {}, tableName: {}",
```
tkhurana, line 74: This log will be very noisy.

palashc: @tkhurana I can remove this one. We have a similar [log line client side](https://github.com/apache/phoenix/pull/1883/files#diff-dd4686c6fc4cfc76ac9f187f3c9cf25e171516197e11d4cdd3a2d43eb26c7f2aR113) as well to log which region server the client is sending the request to. Should we remove that one too?

tkhurana: Yes. We already have a metric for how many validation requests a RS got.

phoenix-core-client/src/main/java/org/apache/phoenix/query/QueryServicesOptions.java
```diff
@@ -368,10 +371,27 @@ public class QueryServicesOptions {
...
+     * is currently incompatible and clients may not learn about the physical table change.
+     * See https://issues.apache.org/jira/browse/PHOENIX-7284.
+     *
+     * Disable caching re-design if your clients will not have ADMIN perms to call region server
+     * RPC. See https://issues.apache.org/jira/browse/HBASE-28508
+     */
+    public static final long DEFAULT_UPDATE_CACHE_FREQUENCY
+                = (long) ConnectionProperty.UPDATE_CACHE_FREQUENCY.getValue("ALWAYS");
+    public static final boolean DEFAULT_LAST_DDL_TIMESTAMP_VALIDATION_ENABLED = false;
+    public static final boolean DEFAULT_PHOENIX_METADATA_INVALIDATE_CACHE_ENABLED = false;
```
tkhurana, line 390: Where is this being used ?

phoenix-core-client/src/main/java/org/apache/phoenix/jdbc/PhoenixStatement.java
```diff
@@ -330,16 +336,20 @@ protected QueryPlan optimizeQuery(CompilableStatement stmt) throws SQLException
     
     protected PhoenixResultSet executeQuery(final CompilableStatement stmt, final QueryLogger queryLogger)
             throws SQLException {
-        return executeQuery(stmt, true, queryLogger, false);
+        return executeQuery(stmt, true, queryLogger, false, this.validateLastDdlTimestamp);
```
tkhurana, line 339: validateLastDDlTimestamp is a member field. Why pass it as a function argument ?

phoenix-core-client/src/main/java/org/apache/phoenix/jdbc/PhoenixStatement.java
```diff
@@ -377,6 +387,14 @@ private PhoenixResultSet executeQuery(final CompilableStatement stmt,
                                 plan =
                                         connection.getQueryServices().getOptimizer()
                                                 .optimize(PhoenixStatement.this, plan);
+                                setLastQueryPlan(plan);
+
+                                //verify metadata for the table/view/index in the query plan
+                                //plan.getTableRef can be null in some cases like EXPLAIN <query>
```
tkhurana, line 393: Nice comment

phoenix-core-client/src/main/java/org/apache/phoenix/util/ValidateLastDDLTimestampUtil.java
```diff
@@ -0,0 +1,231 @@
...
+
+            // add the current table to the request
+            PTable ptable = tableRef.getTable();
+            innerBuilder = RegionServerEndpointProtos.LastDDLTimestampRequest.newBuilder();
+            setLastDDLTimestampRequestParameters(innerBuilder, ptable.getKey(),
+                    ptable.getLastDDLTimestamp());
+            requestBuilder.addLastDDLTimestampRequests(innerBuilder);
+
+            // add all indexes of the current table
+            for (PTable idxPTable : tableRef.getTable().getIndexes()) {
```
tkhurana, line 180: When validating timestamps we validate the timestamps of all the ancestors and the current table as well as indexes. When the server returns StaleMetadataCacheException are we also re-populating the client cache for all the objects that we are validating ? Are we also re-populating the indexes of the current table ?

phoenix-core-client/src/main/java/org/apache/phoenix/jdbc/PhoenixStatement.java
palashc: yeah, used to implement one retry

phoenix-core-client/src/main/java/org/apache/phoenix/util/ValidateLastDDLTimestampUtil.java
palashc: Yes. Currently, we are re-populating everything since we use `MetaDataClient.updateCache` workflow - it resolves ancestors and goes through the process of adding inherited columns, indexes from them, it also re-populates the indexes into the client cache.

phoenix-core-client/src/main/java/org/apache/phoenix/cache/ServerMetadataCacheImpl.java
palashc: done

phoenix-core-server/src/main/java/org/apache/phoenix/coprocessor/PhoenixRegionServerEndpoint.java
palashc: done

phoenix-core-client/src/main/java/org/apache/phoenix/query/QueryServicesOptions.java
palashc: `DEFAULT_LAST_DDL_TIMESTAMP_VALIDATION_ENABLED` used in helper methods in `ValidateLastDDLTimestampUtil` which client uses to decide whether to validate timestamps and `DEFAULT_PHOENIX_METADATA_INVALIDATE_CACHE_ENABLED` in `CQSI.invalidateServerMetadataCache` to decide whether to invalidate cache on server side.

phoenix-core-server/src/main/java/org/apache/phoenix/coprocessor/MetaDataEndpointImpl.java
```diff
@@ -3529,6 +3566,19 @@ private MetaDataMutationResult mutateColumn(
         }
     }
 
+    private void invalidateServerMetadataCache(List<InvalidateServerMetadataCacheRequest> requests)
+            throws Throwable {
+        Properties properties = new Properties();
```
tkhurana, line 3571: Shouldn't we check the invalidate metadata cache config earlier and short circuit earlier before doing all the work ?

phoenix-core-client/src/main/java/org/apache/phoenix/util/PhoenixRuntime.java
tkhurana: Maybe then you shouldn't add an API in PhoenixRuntime and directly use connection.

phoenix-core-client/src/main/java/org/apache/hadoop/hbase/ipc/controller/InvalidateMetadataCacheControllerFactory.java
```diff
@@ -0,0 +1,41 @@
...
+package org.apache.hadoop.hbase.ipc.controller;
+
+import org.apache.hadoop.conf.Configuration;
+import org.apache.hadoop.hbase.ipc.HBaseRpcController;
+import org.apache.hadoop.hbase.ipc.RpcControllerFactory;
+
+/**
+ * Factory to instantiate InvalidateMetadataCacheControllers
+ */
+public class InvalidateMetadataCacheControllerFactory extends RpcControllerFactory {
```
tkhurana, line 27: Why didn't we consider using the ServerSideRpcControllerFactory instead of creating a new factory ?

palashc: I am not sure about this one

phoenix-core-server/src/main/java/org/apache/phoenix/coprocessor/MetaDataEndpointImpl.java
palashc: Yes I believe we could!
But there are multiple places we are invalidating server cache from and there could be more in the future. Assuming we don't do a lot of DDL changes which invalidate the cache, maybe it is okay to keep the check in one place? We could move it to this method in MDEI instead of keeping it in CQSI.

tkhurana: Yes, that's what I was getting to. Doing the check early. You could keep the check in CQSI also.

palashc: > before doing all the work

@tkhurana By this, do you mean creating the list of invalidation requests?

palashc: Okay I guess we can keep a boolean flag in MDEI itself with the config value which is set when the coproc is loaded.

tkhurana: @palashc What I meant was we can check the config and return early before creating the server connection

palashc: @tkhurana I see. Do you think keeping the boolean flag in MDEI and checking that before we create the invalidation requests would be even better?

tkhurana: Yes, that should be even better.

palashc: Done. Stuck to your previous suggestion to just keep it in one place before we create server connection.

tkhurana (approved):

palashc:
Latest test report - https://ci-hadoop.apache.org/job/Phoenix/job/Phoenix-PreCommit-GitHub-PR/job/PR-1883/13/testReport/

@virajjasani I think the 3 failing tests are known flappers? If yes, I will go ahead and merge this.

virajjasani:
Yes they are known flappers. All yours, go ahead!!

PR #1884, opened 2024-04-24
PHOENIX-6714 Return update status from Conditional Upserts

jinggou:


phoenix-core/src/it/java/org/apache/phoenix/end2end/OnDuplicateKeyIT.java
```diff
@@ -720,6 +726,402 @@ public void testMultiplePartialUpdatesInSameBatch() throws Exception {
         }
     }
 
+    @Test
+    public void testColumnsTimestampUpdateWithAllCombinations() throws Exception {
```
virajjasani, line 730: `testColumnsTimestampUpdateWithAllCombinations()` and `testColumnsTimestampUpdateWithIntentionalUpdate()` are the only tests that don't need to be ignored because we are not asserting return status?

jinggou: We are asserting return status, but since the return values are 1 in these two test cases, they will always pass even though hbase change is not included.

tkhurana:
@jinggou can you resolve the merge conflicts

phoenix-core-client/src/main/java/org/apache/phoenix/execute/MutationState.java
```diff
@@ -1415,8 +1425,23 @@ public List<Mutation> getMutationList() {
                             }, iwe, connection, connection.getQueryServices().getProps());
                             shouldRetryIndexedMutation = false;
                         } else {
-                            hTable.batch(mutationBatch, null);
+                            hTable.batch(mutationBatch, resultObjects);
+                        }
+
+                        if (connection.getAutoCommit()) {
+                            for (int i = 0; i < mutationBatch.size(); i++) {
```
tkhurana, line 1432: Instead of using `mutationBatch.size()` use `resultObjects.size()`

phoenix-core-client/src/main/java/org/apache/phoenix/execute/MutationState.java
```diff
@@ -1366,6 +1375,7 @@ private void sendMutations(Iterator<Entry<TableInfo, List<Mutation>>> mutationsI
                     while (itrListMutation.hasNext()) {
                         final List<Mutation> mutationBatch = itrListMutation.next();
                         currentMutationBatch = mutationBatch;
+                        Object[] resultObjects = new Object[mutationBatch.size()];
```
tkhurana, line 1378: I see that we are only reading resultObjects if autocommit is set to true. So we should only allocate this object when autocommit is set. If the autocommit is not set the batch size can be 100s to 1000s and we are just wasting memory by allocating.

phoenix-core-client/src/main/java/org/apache/phoenix/execute/MutationState.java
```diff
@@ -1415,8 +1425,23 @@ public List<Mutation> getMutationList() {
...
-                            hTable.batch(mutationBatch, null);
+                            hTable.batch(mutationBatch, resultObjects);
+                        }
+
+                        if (connection.getAutoCommit()) {
+                            for (int i = 0; i < mutationBatch.size(); i++) {
+                                Result result = (Result) resultObjects[i];
+                                if (result != null && !result.isEmpty()) {
+                                    Cell cell = result.rawCells()[0];
+                                    numUpdatedRowsForAutoCommit = PInteger.INSTANCE.getCodec()
```
tkhurana, line 1436: Now we have 2 APIs on the MutationState object `getUpdateCount()` and `getNumUpdatedRowsForAutoCommit()`. They server a similar purpose but can give different results under some conditions. We should just keep the original API `getUpdateCount` which should handle all the conditions and give the correct result. 
One idea is we could introduce a new field called `numRowsIgnored` initialized to 0. Then if the Result cell is 0 that means the row is ignored and we increment `numRowsIgnored`.  Now in `getUpdateCount` we simply return `numRows - numRowsIgnored`

tkhurana: `numRowsIgnored` is keeping track of the updates that are ignored

phoenix-core-client/src/main/java/org/apache/phoenix/jdbc/PhoenixStatement.java
```diff
@@ -568,13 +568,18 @@ public Integer call() throws SQLException {
...
+                                // Unfortunately, JDBC uses an int for update count, so we
+                                // just max out at Integer.MAX_VALUE
+                                int lastUpdateCount = (int) Math.min(Integer.MAX_VALUE,
+                                        lastState.getUpdateCount());
                                 if (connection.getAutoCommit()) {
                                     connection.commit();
+                                    if (isAtomicUpsert) {
+                                        lastUpdateCount = connection.getMutationState()
+                                                .getNumUpdatedRowsForAutoCommit();
+                                    }
```
tkhurana, line 580: If you do the `numRowsIgnored` suggestion I stated above then we don't need to change anything here.

tkhurana: I thought more about this. I think we will still need some changes here even with my suggestion.

tkhurana: The problem is `connection.commit()` calls `resetState` which resets everything including `numRows`

virajjasani: Yes, that was one of the reasons to introduce `numUpdatedRowsForAutoCommit`

phoenix-core-client/src/main/java/org/apache/phoenix/execute/MutationState.java
tkhurana: Actually, we need to think more carefully about the client side changes. The code written so far is only handling auto-commit on a batch of size 1. It will not work when there is a batch of atomic upserts with auto-commit set to true. See [PHOENIX-6821](https://issues.apache.org/jira/browse/PHOENIX-6821)

tkhurana: Discussed offline, we will only handle the case when auto-commit is set to true and the batch size is 1.

phoenix-core-client/src/main/java/org/apache/phoenix/execute/MutationState.java
tkhurana: @jinggou We should only pass the results array when auto-commit is set to true and batch size is 1 since this is the only case we are handling currently. Let's make that explicit in the code.

tkhurana:
@jinggou I think there seems to be some formatting changes to IndexRegionObserver.java which has made the diff huge. Can we revert and just keep what's relevant.

phoenix-core-client/src/main/java/org/apache/phoenix/execute/MutationState.java
jinggou: done

jinggou:
@tkhurana @virajjasani  Please review my changes
1. reverted the formatting changes to IndexRegionObserver.java and keep what's relevant
2. allocated results array only when autocommit is set and batch size is 1
3. added tests for batched upsert

virajjasani (commented):
I will take another look for any changes if required but overall since I am aware of the changes, let's wait for final review from Tanuj.

Thanks @jinggou @tkhurana!!

phoenix-core/src/it/java/org/apache/phoenix/end2end/OnDuplicateKeyIT.java
```diff
@@ -720,6 +747,470 @@ public void testMultiplePartialUpdatesInSameBatch() throws Exception {
...
+                    " counter2 = CASE WHEN counter2 < 11 THEN counter2 + 1 ELSE counter2 END," +
+                    " approval = CASE WHEN counter2 < 10 THEN 'NONE'" +
+                    " WHEN counter2 < 11 THEN 'MANAGER_APPROVAL'" +
+                    " ELSE approval END";
+            stmt.addBatch(dml);
+            stmt.addBatch(dml);
+            stmt.addBatch(dml);
+
+            int[] actualReturnValues = stmt.executeBatch();
+            int[] expectedReturnValues = new int[]{0, 1, 1, 1, 1, 0};
```
tkhurana, line 1192: I don't think this is correct for the case when auto-commit is true and batch API is used. AFAIK for batching with/without auto-commit the return values will always be {1,1,....1}

phoenix-core-server/src/main/java/org/apache/phoenix/hbase/index/IndexRegionObserver.java
```diff
@@ -1308,6 +1345,19 @@ public void postBatchMutateIndispensably(ObserverContext<RegionCoprocessorEnviro
...
+                      byte[] retVal = PInteger.INSTANCE.toBytes(1);
+                      Cell cell = PhoenixKeyValueUtil.newKeyValue(
+                              miniBatchOp.getOperation(i).getRow(),
+                              EMPTY_BYTE_ARRAY, EMPTY_BYTE_ARRAY, 0, retVal, 0,
+                              retVal.length);
+                      Result result = Result.create(new ArrayList<>(Arrays.asList(cell)));
+                      miniBatchOp.setOperationStatus(i,
+                              new OperationStatus(OperationStatusCode.SUCCESS, result));
+                  }
+              }
```
tkhurana, line 1360: Why do we need this logic again in `postBatchMutateIndispensably` ?

virajjasani: Since HBase treats Put/Delete mutations as non-atomic mutations (unlike Increment/Append mutations), if we don't set Result object here in `postBatchMutateIndispensably()`, no Result object would stay around by the time mutation is executed till this point.

This is the reason why value `0` is provided with Result object early on (as part of `addOnDupMutationsToBatch()` as we short-circuit there by ignoring the mutation) whereas value `1` is provided with Result object here.

tkhurana: I didn't understand your comment "No result object would stay around" ?. We are only setting Result with value `1` here. Result with value `0` are already set earlier and are still around right when this method is called ?

virajjasani: > Result with value `0` are already set earlier and are still around right when this method is called ?

That's correct, because from HBase side, no remaining logic of batchMutate() gets executed as we short-circuit with value `0`.
Only when the whole batchMutate() execution goes through, Result object no longer stays around.
```
      batchOp.visitBatchOperations(true, finalLastIndexExclusive, (int i) -> {
        Mutation mutation = batchOp.getMutation(i);
        if (mutation instanceof Increment || mutation instanceof Append) {
          if (finalSuccess) {
            batchOp.retCodeDetails[i] =
              new OperationStatus(OperationStatusCode.SUCCESS, batchOp.results[i]);
          } else {
            batchOp.retCodeDetails[i] = OperationStatus.FAILURE;
          }
        } else {
          batchOp.retCodeDetails[i] =
            finalSuccess ? OperationStatus.SUCCESS : OperationStatus.FAILURE;
        }
        return true;
      });
```

Here, HBase sets `OperationStatus.SUCCESS` to `batchOp.retCodeDetails[i]` for Put/Delete mutations, leaving no Result object around with OperationStatus.
However, in case of ignore, the above logic is not executed, hence what we set with value `0` with `addOnDupMutationsToBatch()` stays around but for setting `1`, we need to set it with `postBatchMutateIndispensably()`.

tkhurana: But for our purposes we don't need to send the Result with value `1`. Implicit `null` is good enough for us. If we don't set the Result for success a null will be returned which the client is already handling. Then we don't even need to maintain this map of update status. Also, since postBatchMutateIndispensably is also called for regular updates we will do a lookup in the map for every mutation in the batch and that lookup will not return anything but it is adding additional overhead.

phoenix-core-server/src/main/java/org/apache/phoenix/hbase/index/IndexRegionObserver.java
```diff
@@ -1588,10 +1680,36 @@ private List<Mutation> generateOnDupMutations(BatchMutateContext context, Put at
...
+        Set<byte[]> familySet = new HashSet<>();
+        if (!put.isEmpty()) {
+            familySet.addAll(put.getFamilyCellMap().keySet());
+        }
+        if (delete != null) {
+            familySet.addAll(delete.getFamilyCellMap().keySet());
+        }
+        for (byte[] familyBytes : familySet) {
+            Cell emptyKVCell = tuple.getValue(familyBytes,
+                    QueryConstants.ENCODED_EMPTY_COLUMN_BYTES);
```
tkhurana, line 1707: We can't always assume the empty column will be encoded.

phoenix-core/src/it/java/org/apache/phoenix/end2end/OnDuplicateKeyIT.java
```diff
@@ -720,6 +747,470 @@ public void testMultiplePartialUpdatesInSameBatch() throws Exception {
...
+    }
+
+    @Test
+    public void testColumnsTimestampUpdateWithIntentionalUpdate() throws Exception {
+        Properties props = PropertiesUtil.deepCopy(TEST_PROPERTIES);
+        String tableName = generateUniqueName();
+        try (Connection conn = DriverManager.getConnection(getUrl(), props)) {
+            conn.setAutoCommit(true);
+
+            String ddl = "create table " + tableName +
```
tkhurana, line 1055: Can you also run this test by adding `COLUMN_ENCODED_BYTES = 0` to your create table ddl.

phoenix-core-server/src/main/java/org/apache/phoenix/hbase/index/IndexRegionObserver.java
virajjasani: While that is true, i was thinking of two cases:

1. If there is a bug at server side, and the client starts receiving null Result object always, we would not know the difference b/ server returning null object by intention vs the buggy behavior causing override to what server is supposed to send.
2. In future, when we need to introduce JDBC API returning Result object with values, we anyways need to incorporate this logic.

How about we do this: keep the above code same, but wrap it within if condition: 
`if(context.hasAtomic && miniBatchOp.size() == 1)`

phoenix-core-server/src/main/java/org/apache/phoenix/hbase/index/IndexRegionObserver.java
virajjasani: +1, we need to use `EncodedColumnsUtil#usesEncodedColumnNames(PTable)` with the PTable object to determine whether empty column need to be encoded, else use `EMPTY_COLUMN_BYTES`.

virajjasani: Since we don't seem to have PTable object here (unless used by expression evaluation), probably it might be good to send boolean flag (isTableQualifiersEncoded) from client side as Mutation attribute. WDYT @tkhurana?

phoenix-core-server/src/main/java/org/apache/phoenix/hbase/index/IndexRegionObserver.java
tkhurana: After we introduce the check `if(context.hasAtomic && miniBatchOp.size() == 1)` what value is the map bringing ? Is it enough to justify maintaining a new map considering we are only using for batch size 1 ? I would recommend not maintaining the map at all for now. We can revisit this later in a subsequent PR on how to return Result object

phoenix-core-server/src/main/java/org/apache/phoenix/hbase/index/IndexRegionObserver.java
tkhurana: We already serializing the PTable object and re-creating on the server so we can get the encoding information https://github.com/apache/phoenix/blob/master/phoenix-core-server/src/main/java/org/apache/phoenix/hbase/index/IndexRegionObserver.java#L1482

phoenix-core-server/src/main/java/org/apache/phoenix/hbase/index/IndexRegionObserver.java
```diff
@@ -546,9 +568,22 @@ private void addOnDupMutationsToBatch(MiniBatchOperationInProgress<Mutation> min
...
+                  context.setUpdateStatus(i, BatchMutateContext.UpdateStatus.UPDATE);
               } else {
-                  // empty list of generated mutations implies ON DUPLICATE KEY IGNORE
-                  miniBatchOp.setOperationStatus(i, IGNORE);
+                  // empty list of generated mutations implies
+                  // 1) ON DUPLICATE KEY IGNORE if row already exists, OR
+                  // 2) ON DUPLICATE KEY UPDATE if CASE expression is specified and in each of
+                  // them the new value is the same as the old value in the ELSE-clause (empty
+                  // cell timestamp will NOT be updated)
+                  byte[] retVal = PInteger.INSTANCE.toBytes(0);
```
tkhurana, line 578: Should we directly send an integer or should we introduce here a protobuf object which has a field like updateStatus. By sending a protobuf object inside the Result will allow us to expand this object in the future by adding more fields to it.

phoenix-core/src/it/java/org/apache/phoenix/end2end/OnDuplicateKeyIT.java
jinggou: @tkhurana With auto-commit set to true, the size of mutationBatch will be 1 each time,  so the `returnCodes` will be updated by `statement.getUpdateCount()` for each statement in `PhoenixStatement#executeBatch()`.

tkhurana: @jinggou PhoenixStatement#executeBatch internally turns off auto-commit

tkhurana: @jinggou So when will it return 0 ?

tkhurana: Won't getUpdateCount() always return 1 ? We are not changing that API right ?

phoenix-core-server/src/main/java/org/apache/phoenix/hbase/index/IndexRegionObserver.java
virajjasani: Getting rid of Map makes sense. Instead of `Map<Integer, UpdateStatus> updateStatusMap`, let's use `UpdateStatus updateStatus` and update it's value when `miniBatchOp.size() == 1` and operation is atomic.

phoenix-core/src/it/java/org/apache/phoenix/end2end/OnDuplicateKeyIT.java
jinggou: @tkhurana We didn't change `PhoenixStatement#getUpdateCount()` API, but since it calls `PhoenixStatement#getLastUpdateCount()`, 
```
    @Override
    public int getUpdateCount() throws SQLException {
        int updateCount = getLastUpdateCount();
        // Only first call can get the update count, otherwise
        // some SQL clients get into an infinite loop when an
        // update occurs.
        setLastUpdateCount(NO_UPDATE);
        return updateCount;
    }
```
while `lastUpdateCount` will be changed and `setLastUpdateCount` in our new logic:
```
                                int lastUpdateCount = (int) Math.min(Integer.MAX_VALUE,
                                        lastState.getUpdateCount());
                                if (connection.getAutoCommit()) {
                                    connection.commit();
                                    if (isAtomicUpsert) {
                                        lastUpdateCount = connection.getMutationState()
                                                .getNumUpdatedRowsForAutoCommit();
                                    }
                                }
                                ...
                                setLastUpdateCount(lastUpdateCount);
```
I think that's why it won't always return 1.
FYI @virajjasani

jinggou: @tkhurana Now I see that PhoenixStatement#executeBatch internally turns off auto-commit, and the code of `PhoenixStatement#executeBatch` is different in our lightfork. And I'm testing using lightfork, that's why this test will pass.

jinggou: I'll sync to the current `PhoenixStatement#executeBatch` and change the expected value to all 1's and test again.

phoenix-core-server/src/main/java/org/apache/phoenix/hbase/index/IndexRegionObserver.java
jinggou: `updateStatusMap` is designed to replace the check of `miniBatchOp.getOperationStatus(i) == IGNORE`. If we get rid of Map, then we should change this check to something like `miniBatchOp.size() == 1 && updateStatus == IGNORE`, right?

virajjasani: Yes @jinggou, I think that should work. Let's try to change it this way and see how it goes.

virajjasani: Oh but we still need to preserve IGNORE status for batch atomic upsert, and without map it's tricky to know which index of the batch needs to be ignored.

virajjasani: Oh but we still need to preserve IGNORE status for batch atomic upsert, and without map it's tricky to know which index of the batch needs to be ignored.
If not Map, we at least need ArrayList (or array) to get and set UpdateStatus for the indexed Mutation. Or how about we do this: we keep an array of indexes to be ignored i.e. 
`List<Integer> ignoreStatusIndexes = new ArrayList<>()`

tkhurana: In-order to disntinguish between IGNORE and UPDATED can't we simply check if the Result object is null or not in the `OperationStatus` ? All those that will be ignored will have a non-null Result object.

virajjasani: Good idea, this is worth trying!

jinggou: But now we have the Result object with value 0 for ignored and 1 for updated, right? Or do we want to go back to only sending the Result with value 0 for ignored, and implicit null is good enough to represent updated ?

jinggou: Discussed offline and my question is resolved.

phoenix-core-server/src/main/java/org/apache/phoenix/hbase/index/IndexRegionObserver.java
jinggou: I still don't know how to pair the put cells with the PTable object

virajjasani:
HBase 2.6 support is likely going to land very soon https://github.com/apache/phoenix/pull/1793

Also, 2.4.18 is already released: https://github.com/apache/hbase/releases/tag/rel%2F2.4.18
2.5.9 is about to be released soon.

@jinggou could you please remove `@ignore` for tests and apply logic similar to https://github.com/apache/phoenix/blob/master/phoenix-core/src/it/java/org/apache/phoenix/end2end/index/ReplicationWithWALAnnotationIT.java#L173-L201 ?

For the tests, we will need to keep min versions: 2.6.0, 2.5.9 and 2.4.18.

phoenix-core-server/src/main/java/org/apache/phoenix/hbase/index/IndexRegionObserver.java
```diff
@@ -602,7 +616,7 @@ private Collection<? extends Mutation> groupMutations(MiniBatchOperationInProgre
             // unfortunately, we really should ask if the raw mutation (rather than the combined mutation)
             // should be indexed, which means we need to expose another method on the builder. Such is the
             // way optimization go though.
-            if (miniBatchOp.getOperationStatus(i) != IGNORE && this.builder.isEnabled(m)) {
+            if (miniBatchOp.getOperationStatus(i).getResult() == null && this.builder.isEnabled(m)) {
```
tkhurana, line 619: @jinggou Do we need this change ? Will the original change not work

phoenix-core-server/src/main/java/org/apache/phoenix/hbase/index/IndexRegionObserver.java
```diff
@@ -625,7 +639,7 @@ private Collection<? extends Mutation> groupMutations(MiniBatchOperationInProgre
     public static void setTimestamps(MiniBatchOperationInProgress<Mutation> miniBatchOp,
                                      IndexBuildManager builder, long ts) throws IOException {
         for (Integer i = 0; i < miniBatchOp.size(); i++) {
-            if (miniBatchOp.getOperationStatus(i) == IGNORE) {
+            if (miniBatchOp.getOperationStatus(i).getResult() != null) {
```
tkhurana, line 642: Same here. Will the IGNORE check won't work ?

phoenix-core-server/src/main/java/org/apache/phoenix/hbase/index/IndexRegionObserver.java
```diff
@@ -1388,6 +1398,18 @@ public void postBatchMutateIndispensably(ObserverContext<RegionCoprocessorEnviro
       try {
           if (success) {
               context.currentPhase = BatchMutatePhase.POST;
+              if(context.hasAtomic && miniBatchOp.size() == 1) {
+                  if (miniBatchOp.getOperationStatus(0).getResult() == null) {
+                      byte[] retVal = PInteger.INSTANCE.toBytes(1);
+                      Cell cell = PhoenixKeyValueUtil.newKeyValue(
+                              miniBatchOp.getOperation(0).getRow(),
+                              EMPTY_BYTE_ARRAY, EMPTY_BYTE_ARRAY, 0, retVal, 0,
```
tkhurana, line 1406: We need to give the column family and column qualifier

phoenix-core-client/src/main/java/org/apache/phoenix/coprocessorclient/BaseScannerRegionObserverConstants.java
```diff
@@ -62,6 +62,8 @@ public static long getMaxLookbackInMillis(Configuration conf){
     public static final String UPSERT_SELECT_EXPRS = "_UpsertSelectExprs";
     public static final String DELETE_CQ = "_DeleteCQ";
     public static final String DELETE_CF = "_DeleteCF";
+    public static final String UPSERT_STATUS_CQ = "_UpsertStatusCQ";
```
tkhurana, line 65: We need to use this column family and qualifier when generating the Result object on the server and when reading from the client.

phoenix-core-server/src/main/java/org/apache/phoenix/hbase/index/IndexRegionObserver.java
virajjasani: We no longer rely on `IGNORE` because this code has been removed:

```
                  // empty list of generated mutations implies ON DUPLICATE KEY IGNORE
                  miniBatchOp.setOperationStatus(i, IGNORE);
```

phoenix-core-client/src/main/java/org/apache/phoenix/execute/MutationState.java
```diff
@@ -1489,8 +1502,21 @@ public List<Mutation> getMutationList() {
...
+                            hTable.batch(mutationBatch, resultObjects);
+                        }
+
+                        if (resultObjects != null) {
+                            Result result = (Result) resultObjects[0];
+                            if (result != null && !result.isEmpty()) {
+                                Cell cell = result.rawCells()[0];
+                                numUpdatedRowsForAutoCommit = PInteger.INSTANCE.getCodec()
+                                        .decodeInt(cell.getValueArray(), cell.getValueOffset(),
+                                                SortOrder.getDefault());
```
virajjasani, line 1514: Here also, we need to use `_UpsertCF:_UpsertStatusCQ` to get the Cell

phoenix-core-server/src/main/java/org/apache/phoenix/hbase/index/IndexRegionObserver.java
virajjasani: +1

jinggou: yea I neglected this place, will change

phoenix-core-server/src/main/java/org/apache/phoenix/hbase/index/IndexRegionObserver.java
tkhurana: We can't remove it. Otherwise how will ON DUPLICATE KEY IGNORE work ? You have to mark it IGNORED.

tkhurana: How is then hbase ignoring the mutation because it is still there in the batch ?

tkhurana: IGNORE and SUCCESS are equivalent so I don't understand what's the difference.

phoenix-core-server/src/main/java/org/apache/phoenix/hbase/index/IndexRegionObserver.java
```diff
@@ -573,8 +576,19 @@ private void addOnDupMutationsToBatch(MiniBatchOperationInProgress<Mutation> min
...
+                  // them the new value is the same as the old value in the ELSE-clause (empty
+                  // cell timestamp will NOT be updated)
+                  byte[] retVal = PInteger.INSTANCE.toBytes(0);
+                  Cell cell = PhoenixKeyValueUtil.newKeyValue(m.getRow(), Bytes.toBytes(UPSERT_CF),
+                          Bytes.toBytes(UPSERT_STATUS_CQ), 0, retVal, 0, retVal.length);
+                  // put Result in OperationStatus for returning update status from conditional
+                  // upserts, where 0 represents the row is not updated
+                  Result result = Result.create(new ArrayList<>(Arrays.asList(cell)));
+                  miniBatchOp.setOperationStatus(i,
+                          new OperationStatus(OperationStatusCode.SUCCESS, result));
```
virajjasani, line 591: FYI @tkhurana

phoenix-core-server/src/main/java/org/apache/phoenix/hbase/index/IndexRegionObserver.java
virajjasani: FYI https://github.com/apache/phoenix/pull/1884#discussion_r1618041568

> Here, HBase sets OperationStatus.SUCCESS to batchOp.retCodeDetails[i] for Put/Delete mutations, leaving no Result object around with OperationStatus.

phoenix-core-server/src/main/java/org/apache/phoenix/hbase/index/IndexRegionObserver.java
```diff
@@ -1652,10 +1717,37 @@ private List<Mutation> generateOnDupMutations(BatchMutateContext context, Put at
...
-
+    private void addEmptyKVCells(Put put, Delete delete, MultiKeyValueTuple tuple, PTable table) throws IOException {
+        Set<byte[]> familySet = new HashSet<>();
+        if (!put.isEmpty()) {
+            familySet.addAll(put.getFamilyCellMap().keySet());
+        }
+        if (delete != null) {
+            familySet.addAll(delete.getFamilyCellMap().keySet());
+        }
+        for (byte[] familyBytes : familySet) {
```
tkhurana, line 1743: Shouldn't we use this to get empty column family `byte[] emptyCF = SchemaUtil.getEmptyColumnFamily(table)`

phoenix-core-server/src/main/java/org/apache/phoenix/hbase/index/IndexRegionObserver.java
tkhurana: What does it mean for the result to be null vs non null ? It is not very intuitive  when reading the code. We can introduce a utility function here something like this:
 `isAtomicOperationComplete(OperationStatus status) {
return status.getOperationStatusCode() == SUCCESS && status.getResult() != null
}`

phoenix-core-server/src/main/java/org/apache/phoenix/hbase/index/IndexRegionObserver.java
jinggou: @tkhurana Yes, the original code doesn't work because we changed the code `miniBatchOp.setOperationStatus(i, IGNORE);` as @virajjasani mentioned below.

phoenix-core-server/src/main/java/org/apache/phoenix/hbase/index/IndexRegionObserver.java
virajjasani: Creating utility function is fine, which can be used in all similar places.

virajjasani: Though even then, any dev would still not get clarity on why we differentiate null vs non null, so as part of the javadoc of the new utility method, we should mention specifically that 
"HBase returns null Result by default for successful operation, only Increment and Append return non-null Result."

This will make things clear for whoever reads this code in future.

phoenix-core-server/src/main/java/org/apache/phoenix/hbase/index/IndexRegionObserver.java
```diff
@@ -1628,6 +1689,7 @@ private List<Mutation> generateOnDupMutations(BatchMutateContext context, Put at
               }
               // update the cells to the latest values calculated above
               flattenedCells = mergeCells(flattenedCells, updatedCells);
+              flattenedCells.sort(CellComparator.getInstance());
```
tkhurana, line 1692: Can you add a comment here why we need to sort ? I couldn't figure out.

virajjasani: This is really interesting part :)
As part of `addEmptyKVCells()`, we need to get empty cell: `Cell emptyKVCell = tuple.getValue(familyBytes, emptyCQ);`

If we don't sort cells here, tuple#getValue might not be able to retrieve the right empty column cell because getValue() relies on binary search to get the right Cell.

Let's add a short comment here that "we need to retrieve empty cell later on"?

phoenix-core-server/src/main/java/org/apache/phoenix/hbase/index/IndexRegionObserver.java
```diff
@@ -1565,6 +1589,13 @@ private List<Mutation> generateOnDupMutations(BatchMutateContext context, Put at
       // read the column values requested in the get from the current data row
       List<Cell> cells = IndexUtil.readColumnsFromRow(currentDataRowState, colsReadInExpr);
 
+      // store current cells into a map where the key is column family and the value is a map with
+      // column qualifier as key and a pair of cell and a boolean (is true if the expression is
+      // CaseExpression and Else-clause is evaluated to be true, is null if there is no
+      // expression on this column, otherwise false)
+      Map<ImmutableBytesPtr, Map<ImmutableBytesPtr, Pair<Cell, Boolean>>> currCfCqCellExprMap =
```
tkhurana, line 1596: We should use ColumnReference as the key here. It is already being used in this file. It could simplify the map here.

phoenix-core-server/src/main/java/org/apache/phoenix/hbase/index/IndexRegionObserver.java
```diff
@@ -1583,6 +1614,17 @@ private List<Mutation> generateOnDupMutations(BatchMutateContext context, Put at
       } else {
           // Base current state off of existing row
           flattenedCells = cells;
+          // store all current cells from currentDataRowState
```
tkhurana, line 1617: Technically, you don't need to read all the cells. You only need columns which are being updated. You have that information in the PTable object.  The PTable that you get as part of OnDupKey is a special PTable with only the subset of columns which are being updated. This could be an important optimization because the number of cells in the row could be lot compared to the number of columns which are actually being updated.

phoenix-core-server/src/main/java/org/apache/phoenix/hbase/index/IndexRegionObserver.java
```diff
@@ -1642,7 +1704,9 @@ private List<Mutation> generateOnDupMutations(BatchMutateContext context, Put at
                   transferAttributes(atomicPut, put);
                   mutations.add(put);
               }
-              put.add(cell);
+              if (checkCellNeedUpdate(cell, currCfCqCellExprMap)) {
```
tkhurana, line 1707: I think if you do this check before allocating the put object and adding it to the mutations list it will simplify your code.

phoenix-core-server/src/main/java/org/apache/phoenix/hbase/index/IndexRegionObserver.java
```diff
@@ -1652,10 +1716,37 @@ private List<Mutation> generateOnDupMutations(BatchMutateContext context, Put at
               delete.add(cell);
           }
       }
+
```
tkhurana, line 1719: This logic will also become simpler if you try my suggestion above.

phoenix-core-server/src/main/java/org/apache/phoenix/hbase/index/IndexRegionObserver.java
virajjasani: How about a follow-up Jira for this? More Jiras in this area are expected anyway.

tkhurana: I am ok with doing this in a follow-up JIRA.

phoenix-core-server/src/main/java/org/apache/phoenix/hbase/index/IndexRegionObserver.java
jinggou: @tkhurana I don't get your point. Adding the put object to mutations only need to be done once, but adding cell to put should be done each time when the cell need to be updated. What is the difference if we do this check earlier?

phoenix-core-server/src/main/java/org/apache/phoenix/hbase/index/IndexRegionObserver.java
jinggou: Done

phoenix-core-server/src/main/java/org/apache/phoenix/hbase/index/IndexRegionObserver.java
jinggou: Done

phoenix-core-client/src/main/java/org/apache/phoenix/execute/MutationState.java
jinggou: Done

phoenix-core-server/src/main/java/org/apache/phoenix/hbase/index/IndexRegionObserver.java
jinggou: Done

phoenix-core-server/src/main/java/org/apache/phoenix/hbase/index/IndexRegionObserver.java
jinggou: Done

phoenix-core-server/src/main/java/org/apache/phoenix/hbase/index/IndexRegionObserver.java
jinggou: Done

phoenix-core-server/src/main/java/org/apache/phoenix/hbase/index/IndexRegionObserver.java
```diff
@@ -1642,7 +1701,14 @@ private List<Mutation> generateOnDupMutations(BatchMutateContext context, Put at
                   transferAttributes(atomicPut, put);
                   mutations.add(put);
               }
-              put.add(cell);
+              if (checkCellNeedUpdate(cell, currColumnCellExprMap)) {
+                  put.add(cell);
+              }
+
+              if (checkCellNeedUpdate(cell, currColumnCellExprMap)) {
```
tkhurana, line 1708: You are checking twice. Seems like a typo

tkhurana: @jinggou Will doing like this simplify your logic
```suggestion
              for (int i = 0; i < tuple.size(); i++) {
                   Cell cell = tuple.getValue(i);
                   if (cell.getType() == Cell.Type.Put && checkCellNeedUpdate(cell, currColumnCellExprMap)) {
                     if (put == null) {
                       put = new Put(rowKey);
                       transferAttributes(atomicPut, put);
                       mutations.add(put);
                    }
                    put.add(cell);
```

jinggou: This makes sense. But if the put is null (no cells need update) and delete is not null, we still need to add the empty cell to put. I'm not sure about the specific logic of delete, but if we don't add an empty cell, NullPointerException will be thrown in the delete test. 
I'll update the code based on your suggested change, and we can see which way is simpler.

phoenix-core-server/src/main/java/org/apache/phoenix/hbase/index/IndexRegionObserver.java
```diff
@@ -1669,6 +1748,44 @@ private static void flattenCells(Mutation m, List<Cell> flattenedCells) {
...
+        if (colCellExprMap.isEmpty() || !colCellExprMap.containsKey(colRef)) {
+            return true;
+        }
+
+        Pair<Cell, Boolean> valuePair = colCellExprMap.get(colRef);
+        Boolean isInCaseExpressionElseClause = valuePair.getSecond();
+        if (isInCaseExpressionElseClause == null) {
+            return false;
+        }
+        if (isInCaseExpressionElseClause == false) {
```
virajjasani, line 1777: nit: `if(!isInCaseExpressionElseClause)`

virajjasani (dismissed):
+1, pending build results and +1 from @tkhurana

virajjasani:
All the latest builds are timing out, aborting or getting much longer to complete. The cause is also not clear and tests are passing locally. We should try breaking `OnDuplicateKeyIT` into two classes: move half of the tests to new class `OnDuplicateKeyIT2`.

tkhurana (approved):

virajjasani:
It's still not going well, the latest build is running for almost 6 hr, it will either get aborted or timed out: https://ci-hadoop.apache.org/job/Phoenix/job/Phoenix-PreCommit-GitHub-PR/job/PR-1884/21/

virajjasani (approved):
+1, once tests are divided into multiple classes and we get successful build results, this is ready for merge, given source code remains

virajjasani:
can see that the test was running when the `OutOfMemoryError` was thrown from logs: https://ci-hadoop.apache.org/job/Phoenix/job/Phoenix-PreCommit-GitHub-PR/job/PR-1884/21/artifact/yetus-general-check/output/patch-unit-root.txt

PR #1885, opened 2024-04-26
PHOENIX-7229 Addendum to handle cases where a child view extends the

tkhurana:
parent's PK but we do a point lookup using the parent's PK.


kadirozde (approved):
+1, Thanks!

virajjasani (approved):

tkhurana:
I will also need to handle aggregate queries.

kadirozde (approved):
+1, Thanks!

PR #1888, opened 2024-05-10
PHOENIX-7313 All cell versions should not be retained during flushes …

kadirozde:
…and minor compaction when maxlookback is disabled

### DIFF
--- a/phoenix-core-server/src/main/java/org/apache/phoenix/coprocessor/BaseScannerRegionObserver.java
+++ b/phoenix-core-server/src/main/java/org/apache/phoenix/coprocessor/BaseScannerRegionObserver.java
@@ -357,7 +357,7 @@ abstract public class BaseScannerRegionObserver implements RegionObserver {
                 dataRegion, indexMaintainer, null, viewConstants, null, null, projector, ptr, useQualiferAsListIndex);
     }
 
-    public void setScanOptionsForFlushesAndCompactions(ScanOptions options,
+    public void setScanOptionsForFlushesAndCompactions(Store store, ScanOptions options,
             boolean retainAllVersions) {
         // We want the store to give us all the deleted cells to StoreCompactionScanner
         options.setKeepDeletedCells(KeepDeletedCells.TTL);
@@ -365,6 +365,11 @@ abstract public class BaseScannerRegionObserver implements RegionObserver {
         if (retainAllVersions) {
             options.setMaxVersions(Integer.MAX_VALUE);
             options.setMinVersions(Integer.MAX_VALUE);
+        } else {
+            options.setMinVersions(Math.max(Math.max(options.getMaxVersions(),
+                    store.getColumnFamilyDescriptor().getMaxVersions()), 1));
+            options.setMinVersions(Math.max(Math.max(options.getMinVersions(),
+                    store.getColumnFamilyDescriptor().getMaxVersions()), 1));
         }
 
     }
@@ -378,7 +383,7 @@ abstract public class BaseScannerRegionObserver implements RegionObserver {
             boolean retainAllVersions =  isMaxLookbackTimeEnabled(
                     BaseScannerRegionObserverConstants.getMaxLookbackInMillis(conf))
                     || request.isMajor();
-            setScanOptionsForFlushesAndCompactions(options, retainAllVersions);
+            setScanOptionsForFlushesAndCompactions(store, options, retainAllVersions);
             return;
         }
         long maxLookbackAge = getMaxLookbackAge(c);
@@ -396,7 +401,7 @@ abstract public class BaseScannerRegionObserver implements RegionObserver {
         if (isPhoenixTableTTLEnabled(conf)) {
             boolean retainAllVersions =  isMaxLookbackTimeEnabled(
                     BaseScannerRegionObserverConstants.getMaxLookbackInMillis(conf));
-            setScanOptionsForFlushesAndCompactions(options, retainAllVersions);
+            setScanOptionsForFlushesAndCompactions(store, options, retainAllVersions);
             return;
         }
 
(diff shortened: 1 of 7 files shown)


kadirozde:
The last commit for fixing the same issue reported in https://issues.apache.org/jira/browse/PHOENIX-7232

tkhurana:
@kadirozde Can you also fix the issue to disable masking if TTL is set to forever. Thanks

virajjasani:
First build timed out after ~9 hr, and second one ran for only ~16 min.
Triggered new one: https://ci-hadoop.apache.org/job/Phoenix/job/Phoenix-PreCommit-GitHub-PR/job/PR-1888/3/

virajjasani:
Almost every test is getting aborted after ~9 hr https://ci-hadoop.apache.org/job/Phoenix/job/Phoenix-PreCommit-GitHub-PR/job/PR-1888/

This usually means that some test is getting stuck forever and that's why the whole test suite is not getting completed.

phoenix-core-server/src/main/java/org/apache/phoenix/coprocessor/CompactionScanner.java
```diff
@@ -94,19 +94,18 @@ public CompactionScanner(RegionCoprocessorEnvironment env,
...
-        maxLookbackInMillis = overriddenMaxLookback == null ?
+        this.maxLookbackInMillis = overriddenMaxLookback == null ?
                 maxLookbackInMillis : Math.max(maxLookbackInMillis, overriddenMaxLookback);
         // The oldest scn is current time - maxLookbackInMillis. Phoenix sets the scan time range
         // for scn queries [0, scn). This means that the maxlookback size should be
         // maxLookbackInMillis + 1 so that the oldest scn does not return empty row
-        this.maxLookbackWindowStart = maxLookbackInMillis == 0 ?
-                compactionTime : compactionTime - (maxLookbackInMillis + 1);
+        this.maxLookbackWindowStart = this.maxLookbackInMillis == 0 ?
+                compactionTime : compactionTime - (this.maxLookbackInMillis + 1);
```
tkhurana, line 108: @kadirozde It will be useful to have some logging here so that we know we are using Phoenix compactor

kadirozde: In which case we should log and why?

kadirozde: I misunderstood this comment. This was for CompactionScanner. I added logging for that.

jpisaac (approved):
LGTM +1, Will need to check the TFs

jpisaac:
@kadirozde I may have prematurely approved it. Can u add some IT tests for compacting shared Index tables.

kadirozde:
Tests are passing now. Only one test failed which passes when I run. I think it (ConcurrentMutationsIT.testSetIndexedColumnToNullAndValueAtSameTSWithStoreNulls2) is a flapper. @tkhurana, @virajjasani, I want to merge this if you do not have further comments.

tkhurana (approved):

phoenix-core-server/src/main/java/org/apache/phoenix/coprocessor/BaseScannerRegionObserver.java
```diff
@@ -357,20 +357,33 @@ RegionScanner getWrappedScanner(final ObserverContext<RegionCoprocessorEnvironme
...
         // We want the store to give us all the deleted cells to StoreCompactionScanner
         options.setKeepDeletedCells(KeepDeletedCells.TTL);
         options.setTTL(HConstants.FOREVER);
-        options.setMaxVersions(Integer.MAX_VALUE);
-        options.setMinVersions(Integer.MAX_VALUE);
+        if (retainAllVersions) {
+            options.setMaxVersions(Integer.MAX_VALUE);
+            options.setMinVersions(Integer.MAX_VALUE);
+        } else {
+            options.setMinVersions(Math.max(Math.max(options.getMaxVersions(),
```
d-c-manning, line 369: is this intended to be `options.setMaxVersions`

phoenix-core-server/src/main/java/org/apache/phoenix/coprocessor/BaseScannerRegionObserver.java
```diff
@@ -357,20 +357,33 @@ RegionScanner getWrappedScanner(final ObserverContext<RegionCoprocessorEnvironme
...
+                    store.getColumnFamilyDescriptor().getMaxVersions()), 1));
+            options.setMinVersions(Math.max(Math.max(options.getMinVersions(),
+                    store.getColumnFamilyDescriptor().getMaxVersions()), 1));
```
d-c-manning, line 372: is min versions supposed to be columnFamilyDescription max versions if it exists, or should this also be min? `store.getColumnFamilyDescriptor().getMinVersions()`

kadirozde: This code is borrowed from the previous implementation (before CompactionScanner).  It is interesting that some tests for old indexes or SYCAT fails if getMaxVersions is not used above. I do not know the exact reason for it. This will be changed very likely again when we enable CompactionScanner for minor compaction. I am planning to get that change (for https://issues.apache.org/jira/browse/PHOENIX-7314) soon.

d-c-manning: Yeah I could see how potentially we still want to retain all versions here for whatever reason... I did not try to understand it. But on line 369 above we are also calling `options.setMinVersions` so we call `options.setMinVersions` twice, but never call `options.setMaxVersions`. If the CFDescriptor can have a higher number for maxVersions than what `options` contains, then min will be greater than max.

kadirozde: I just noticed that there is a typo here. The old code just had `options.setMinVersions(Math.max(Math.max(options.getMinVersions(),
                    store.getColumnFamilyDescriptor().getMaxVersions()), 1));`.  I wanted to have the same logic for max version and even though it was not necessary for existing tests. So my intention was to add `options.setMaxVersions(Math.max(Math.max(options.getMaxVersions(),
                    store.getColumnFamilyDescriptor().getMaxVersions()), 1));` However, because of the typo, I ended up not adding this logic for max version. The code is harmless but confusing. I am going to change this in the jira. Thank you for catching this.

phoenix-core-server/src/main/java/org/apache/phoenix/coprocessor/BaseScannerRegionObserver.java
kadirozde: Yes, it was. Please see the next comment.

PR #1896, opened 2024-05-26
PHOENIX-7314 Enable CompactionScanner for flushes and minor compaction

kadirozde:


phoenix-core-server/src/main/java/org/apache/phoenix/coprocessor/MetaDataEndpointImpl.java
```diff
@@ -1654,7 +1655,8 @@ else if (linkType == PHYSICAL_TABLE) {
                 else {
                     byte[] maxLookbackAgeInBytes = result.getValue(TABLE_FAMILY_BYTES, MAX_LOOKBACK_AGE_BYTES);
                     if (maxLookbackAgeInBytes != null) {
-                        return PLong.INSTANCE.getCodec().decodeLong(maxLookbackAgeInBytes, 0, SortOrder.getDefault());
+                        return PInteger.INSTANCE.getCodec().decodeInt(maxLookbackAgeInBytes, 0,
```
sanjeet006py, line 1658: The test in MaxLookbackIT seems to be failing as MAX_LOOKBACK_AGE is of type BIGINT in SYSCAT while we are decoding it via PInteger decoder instead of PLong decoder. I think we need to change the type of MAX_LOOKBACK_AGE column in SYSCAT to INTERGER. 
I am also checking other places where we might we using max lookback as long and will update on comment. Thanks

kadirozde: That is great! Thank you very much! I will update the PR accordingly.

phoenix-core-server/src/main/java/org/apache/phoenix/coprocessor/UngroupedAggregateRegionObserver.java
```diff
@@ -581,92 +601,135 @@ private boolean areMutationsInSameTable(Table targetHTable, Region region) {
...
+            return User.runAsLoginUser(new PrivilegedExceptionAction<InternalScanner>() {
+                @Override public InternalScanner run() throws Exception {
+                    String tableName = c.getEnvironment().getRegion().getRegionInfo().getTable()
+                            .getNameAsString();
+                    long maxLookbackInMillis =
+                            UngroupedAggregateRegionObserver.getMaxLookbackInMillis(
+                                    tableName, store.getColumnFamilyName(),
+                                    BaseScannerRegionObserverConstants.getMaxLookbackInMillis(
+                                            c.getEnvironment().getConfiguration()));
+                    maxLookbackInMillis = CompactionScanner.getMaxLookbackInMillis(tableName,
```
sanjeet006py, line 619: Why are we maintaining two maps of max lookback age: one in this class and other in Compaction Scanner?

phoenix-core-server/src/main/java/org/apache/phoenix/coprocessor/UngroupedAggregateRegionObserver.java
```diff
@@ -581,92 +601,135 @@ private boolean areMutationsInSameTable(Table targetHTable, Region region) {
...
+                        SchemaUtil.getParentTableNameFromIndexTable(tableName.getNameAsString(),
+                                MetaDataUtil.VIEW_INDEX_TABLE_PREFIX) :
+                        tableName.getNameAsString();
+                PTable table = null;
+                Integer maxLookbackAge = null;
+                try (PhoenixConnection conn = QueryUtil.getConnectionOnServer(
+                        compactionConfig).unwrap(PhoenixConnection.class)) {
+                    table = conn.getTableNoCache(fullTableName);
+                    maxLookbackAge = table.getMaxLookbackAge();
+                    if (table.getType() == PTableType.INDEX) {
```
sanjeet006py, line 657: I think we don't need this `if` check now as Indexes inherit max lookback of tables and view indexes inherit max lookback of parent view?

phoenix-core-server/src/main/java/org/apache/phoenix/coprocessor/UngroupedAggregateRegionObserver.java
```diff
@@ -581,92 +601,135 @@ private boolean areMutationsInSameTable(Table targetHTable, Region region) {
...
+            InternalScanner scanner, FlushLifeCycleTracker tracker) throws IOException {
+        if (!isPhoenixTableTTLEnabled(c.getEnvironment().getConfiguration())) {
+            return scanner;
+        } else {
+            return User.runAsLoginUser(new PrivilegedExceptionAction<InternalScanner>() {
+                @Override public InternalScanner run() throws Exception {
+                    String tableName = c.getEnvironment().getRegion().getRegionInfo().getTable()
+                            .getNameAsString();
+                    long maxLookbackInMillis =
+                            UngroupedAggregateRegionObserver.getMaxLookbackInMillis(
```
sanjeet006py, line 615: This guarantees eventual consistency and the static map is filled up by pre-compact hook. So that means once DDL to change max lookback for a table is run, new value will start getting used on region server (updated in static map) when pre-compact hook on that table is called. So, I have following doubts:
- Is there a set frequency within which pre-compact will get called at least once for each table on each region server hosting regions of that table?
- For short duration within the max lookback map just after changing max lookback, value of max lookback for table and index/view and view index might not match, is that fine?

phoenix-core-server/src/main/java/org/apache/phoenix/coprocessor/UngroupedAggregateRegionObserver.java
kadirozde: The one in UngroupedAggregateRegionObserver is for caching the max lookback age in order to prevent getTable calls in the coproc flush hook. The other is for overriding the cached value by other coprocs, (for example a backup coproc).

phoenix-core-server/src/main/java/org/apache/phoenix/coprocessor/UngroupedAggregateRegionObserver.java
kadirozde: Yes, you are right that this solution here is eventual consistent. The precompact is called for minor and major compaction. Yes, major compaction is done periodically typically once a week. However, the compaction policy may decide to skip it.  Minor compaction is supposed to run whenever the number of HFiles exceeds the configured threshold. If the eventual consistency is fine for data table regions, then eventual consistency between data table region and index from the max lookback perspective should be also fine. However, I think we need a better solution than this such that we need to populate and invalidate this cache whenever the max lookback config changes. The solution here is a stopgap. Hopefully, we will implement a better solution before 5.3.  Please note we can continue to configure "phoenix.max.lookback.age.seconds" at the table level to using the Phoenix alter table command to get much better eventual consistency meanwhile.

kadirozde:
I run the failed tests locally and all of them passes for me.

phoenix-core-server/src/main/java/org/apache/phoenix/coprocessor/CompactionScanner.java
```diff
@@ -419,13 +551,15 @@ private void formNextCompactionRowVersion(LinkedList<LinkedList<Cell>> columns,
                     if (cell.getTimestamp() >= rowContext.maxTimestamp) {
                         rowContext.familyDeleteMarker = cell;
                         column.removeFirst();
+                        break;
                     }
                     continue;
                 }
                 else if (cell.getType() == Cell.Type.DeleteFamilyVersion) {
-                    if (cell.getTimestamp() >= rowContext.maxTimestamp) {
+                    if (cell.getTimestamp() == rowVersion.ts) {
```
virajjasani, line 559: Nice

virajjasani (approved):

PR #1902, opened 2024-06-06
PHOENIX-7326 Simplify LockManager and make it more efficient

kadirozde:


phoenix-core-server/src/main/java/org/apache/phoenix/hbase/index/LockManager.java
```diff
@@ -21,224 +21,148 @@
...
-            this.context = context;
-            this.lock = lock;
-        }
-
-        Lock getLock() {
-            return lock;
+            if (success) {
+                return this;
+            }
+            return null;
```
virajjasani, line 134: Since `InterruptedException` is being re-thrown, `success` here will always be true so we will never return null here. We can directly return `this` instead of checking for `success` value.

phoenix-core-server/src/main/java/org/apache/phoenix/hbase/index/LockManager.java
```diff
@@ -21,224 +21,148 @@
...
-        RowLockContext(ImmutableBytesPtr rowKey) {
+        private RowLockImpl(ImmutableBytesPtr rowKey) {
             this.rowKey = rowKey;
+            lock.lock();
+            threadName = Thread.currentThread().getName();
         }
 
-        RowLockImpl newRowLock() {
-            count.incrementAndGet();
+        RowLockImpl lock(long waitDuration) throws IOException{
```
virajjasani, line 104: nit: `waitDurationMs` to indicate the unit is milliseconds?

phoenix-core-server/src/main/java/org/apache/phoenix/hbase/index/LockManager.java
```diff
@@ -21,224 +21,148 @@
...
-            if (traceScope != null) {
-                traceScope.close();
+            // The existing lock was released and removed from the hash map before the current
+            // thread attempt to lock
+            long now = EnvironmentEdgeManager.currentTimeMillis();
+            long timePassed = now - startTime;
+            if (timePassed > waitDuration) {
+                throw new TimeoutIOException("Timed out waiting for lock for row: " + rowKey);
             }
+            waitDuration -= timePassed;
```
virajjasani, line 79: nit: `waitDuration -= (int) timePassed`

phoenix-core-server/src/main/java/org/apache/phoenix/hbase/index/LockManager.java
```diff
@@ -21,224 +21,148 @@
...
+    private final ConcurrentHashMap<ImmutableBytesPtr, RowLockImpl> lockedRows =
+            new ConcurrentHashMap<>();
 
     public LockManager () {
     }
 
     /**
      * Lock the row or throw otherwise
-     * @param rowKey the row key
-     * @return RowLock used to eventually release the lock
```
virajjasani, line 55: nit: shall we keep `@return` with description as is?

virajjasani (approved):
Left few minor comments, overall +1 for this nice simplification

virajjasani (approved):

PR #1904, opened 2024-06-12
PHOENIX-7318 : Support JSON_MODIFY in Upserts

ranganathg:


phoenix-core-client/src/main/java/org/apache/phoenix/compile/UpsertCompiler.java
```diff
@@ -143,6 +147,9 @@ private static void setValues(byte[][] values, int[] pkSlotIndex, int[] columnIn
         for (int i = 0, j = numSplColumns; j < values.length; j++, i++) {
             byte[] value = values[j];
             PColumn column = table.getColumns().get(columnIndexes[i]);
+            if (value == null) {
```
tkhurana, line 150: nit: Move this check before the column assignment

phoenix-core-client/src/main/java/org/apache/phoenix/compile/UpsertCompiler.java
```diff
@@ -931,6 +906,78 @@ public int getPosition() {
                 connection, pkSlotIndexes, useServerTimestamp, onDupKeyBytes, maxSize, maxSizeBytes);
     }
 
+    private static byte[] getBytes(PTable table, StatementContext context,
```
tkhurana, line 909: getBytes is a terrible name. Suggestion `getOnDuplicateKeyBytes`

phoenix-core-client/src/main/java/org/apache/phoenix/compile/UpsertCompiler.java
```diff
@@ -842,11 +851,27 @@ public MutationPlan compile(UpsertStatement upsert) throws SQLException {
                         expression.getDataType(), column.getDataType(), "expression: "
                                 + expression.toString() + " in column " + column);
             }
+            if (!SchemaUtil.isPKColumn(column) && !isJsonNode(valueNode)) {
+                nonPKColumns.add(
+                        new Pair<>(ColumnName.caseSensitiveColumnName(column.getName().getString()),
+                                valueNode));
+            }
+            if (isJsonNode(valueNode)) {
```
tkhurana, line 859: Shouldn't this be `else if`

phoenix-core-client/src/main/java/org/apache/phoenix/compile/UpsertCompiler.java
```diff
@@ -842,11 +851,27 @@ public MutationPlan compile(UpsertStatement upsert) throws SQLException {
                         expression.getDataType(), column.getDataType(), "expression: "
                                 + expression.toString() + " in column " + column);
             }
+            if (!SchemaUtil.isPKColumn(column) && !isJsonNode(valueNode)) {
+                nonPKColumns.add(
+                        new Pair<>(ColumnName.caseSensitiveColumnName(column.getName().getString()),
```
tkhurana, line 856: You are ignoring the column family name here. This is not correct

phoenix-core-client/src/main/java/org/apache/phoenix/compile/UpsertCompiler.java
```diff
@@ -842,11 +851,27 @@ public MutationPlan compile(UpsertStatement upsert) throws SQLException {
...
+                                valueNode));
+            }
+            if (isJsonNode(valueNode)) {
+                jsonExpressions.add(
+                        new Pair<>(ColumnName.caseSensitiveColumnName(column.getName().getString()),
+                                valueNode));
+                jsonExpressions.addAll(nonPKColumns);
+                nonPKColumns = Lists.newArrayList();
```
tkhurana, line 864: Why are you doing this here and resetting the nonPKColumns list ?

tkhurana: You are adding again when you exit the loop. I didn't quite understand this logic.

phoenix-core-client/src/main/java/org/apache/phoenix/compile/UpsertCompiler.java
tkhurana: Actually, can you verify this by adding a test case with the same column name in multiple column families.

phoenix-core/src/it/java/org/apache/phoenix/end2end/json/JsonFunctionsIT.java
```diff
@@ -749,4 +761,186 @@ public void testJsonWithSetGetObjectAPI() throws Exception {
...
+                    "strcol5 varchar, " +
+                    "jsoncol json)";
+            conn.createStatement().execute(ddl);
+            PreparedStatement stmt = conn.prepareStatement("UPSERT INTO " + tableName + " (pk,col,strcol,jsoncol) VALUES (?,?,?,?)");
+            stmt.setInt(1, 1);
+            stmt.setInt(2, 2);
+            stmt.setString(3, "");
+            stmt.setString(4, basicJson);
+            stmt.execute();
+            String upsert = "UPSERT INTO " + tableName + "(pk,col,strcol,jsoncol) VALUES(1,2,JSON_VALUE(jsoncol, '$.info.address.town'),JSON_MODIFY(jsoncol, '$.info.address.town', '\"Manchester\"')) ";
```
tkhurana, line 788: So this query doesn't work:
`UPSERT INTO TABLENAME (pk,col,strcol,jsoncol) VALUES(1,2,JSON_VALUE(jsoncol, '$.info.address.town') || 'City',JSON_MODIFY(jsoncol, '$.info.address.town', 'Manchester')`

tkhurana: But if I change the query to `UPSERT INTO TABLENAME (pk) VALUES (1) ON DUPLICATE KEY UPDATE strcol = JSON_VALUE(jsoncol, '$.info.address.town') || 'City', jsoncol = JSON_MODIFY(jsoncol, '$.info.address.town', 'Manchester')` then it works

phoenix-core-client/src/main/java/org/apache/phoenix/compile/UpsertCompiler.java
ranganathg: I see "SchemaUtil.isPKColumn(column)" being used at different places without checking to column family. What is the problem with multiple column families you find?

phoenix-core-client/src/main/java/org/apache/phoenix/compile/UpsertCompiler.java
ranganathg: This is bascically say you have something like "UPSERT nonPkcol1,nonPkcol2,jsoncol1, jsoncol2,nonpkcol3,jsoncol3, nonpkcol4". as you find nonPK column add to list till you find a jsoncol at which point you include all those columns and reset because you want to add the next set of nonpk cols. and adding after exiting the loop to add the remaining nonpk cols which came after all the json cols.

phoenix-core/src/it/java/org/apache/phoenix/end2end/json/JsonFunctionsIT.java
ranganathg: what does this mean? "|| 'City'"?

ranganathg: I will add this case

phoenix-core-client/src/main/java/org/apache/phoenix/compile/UpsertCompiler.java
tkhurana: The problem is not in `isPKColumn` but column.getName() which only returns the column qualifier

phoenix-core-client/src/main/java/org/apache/phoenix/compile/UpsertCompiler.java
tkhurana: But ultimately everything merges to one list so why we need to reset. Why can't we just add all non pk cols in one list and json expressions in another list. At then end of the loop if json expressions list is not empty add all the non pk columns to json expression list.

ranganathg: Yeah your comment perfectly makes sense

phoenix-core/src/it/java/org/apache/phoenix/end2end/json/JsonFunctionsIT.java
ranganathg: I will create a separate JIRA to handle nested JSON Functions.

phoenix-core-client/src/main/java/org/apache/phoenix/compile/UpsertCompiler.java
ranganathg: Got it. Will add an IT to test these and add appropriate tests.

phoenix-core/src/it/java/org/apache/phoenix/end2end/json/JsonFunctionsIT.java
ranganathg: I have handled this scenario

tkhurana:
@ranganathg Changes look good. I will wait for the CI to finish and then merge your PR.

tkhurana:
@ranganathg Can you fix the checkstyle errors. They are mostly unsued imports so should be easy fix. Also there is one test failure. Can you try running the test locally ?

tkhurana (approved):

ranganathg:
> @ranganathg Can you fix the checkstyle errors. They are mostly unsued imports so should be easy fix. Also there is one test failure. Can you try running the test locally ?

fixed the checkstyle. and that test is running fine locally. Thank you.

PR #1905, opened 2024-06-12
PHOENIX-7329 Change TTL column type to VARCHAR in syscat

tkhurana:
This PR changes the TTL column data type in syscat to VARCHAR. The in-memory representation in PTable is still integer. That will change in a subsequent PR with PHOENIX-7170

### DIFF
--- a/phoenix-core/src/it/java/org/apache/phoenix/end2end/MoveTTLDuringUpgradeIT.java
+++ b/phoenix-core/src/it/java/org/apache/phoenix/end2end/MoveTTLDuringUpgradeIT.java
@@ -79,7 +79,7 @@ public class MoveTTLDuringUpgradeIT extends ParallelStatsDisabledIT {
                 int ttl = tableTTLMap.get(table);
                 //Check if TTL is moved to SYSCAT.
                 if (ttl != HConstants.FOREVER) {
-                    assertEquals(ttl, rs.getInt(2));
+                    assertEquals(ttl, Integer.valueOf(rs.getString(2)).intValue());
                 } else {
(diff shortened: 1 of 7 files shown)


tkhurana:
I ran the following ITs manually:

1. TableTTLIT
2. TTLAsPhoenixTTLIT
3. ViewTTLNotEnabledIT
4. MoveTTLDuringUpgradeIT

tkhurana:
Only failure is known flapper MutableIndexFailureIT.testIndexWriteFailure

lokiore:
@tkhurana in MetaProperties we are storing ttl as Integer, When conditional TTL will come we are allowed to change condition with Alter!? I think we should move those to String as well. https://github.com/apache/phoenix/blob/6425350bab63c0c52624e48e0f3836fc5a025d56/phoenix-core/src/main/java/org/apache/phoenix/schema/MetaDataClient.java#L5771

tkhurana:
@lokiore This PR is not changing the in memory representation. In MetadataClient, currently we take the integer value and then store it as a varchar in syscat. What I am thinking is that ALTER TABLE SET TTL=<val> where <val> can be integer or val can be a ttl expression string. I am working on a new way to represent TTL in memory something that can represent both integers and expressions. But that will a subsequent PR which changes how TTL is represented in memory. In syscat it will still be a varchar.

tkhurana:
Similarly in PTable, ttl is also an integer for now. We are just converting it to/from the varchar value in syscat.

lokiore (approved):
OKayz that makes sense, in that perspective LGTM!

PR #1906, opened 2024-06-15
PHOENIX-7330 Introducing Binary JSON (BSON) with Complex Document structures in Phoenix

virajjasani:
Jira: PHOENIX-7330

- Design Doc 1: https://docs.google.com/document/d/1mlHdXP3e8Ol9G1GPGQDQQjUIOTyVkV5R4w15U6IePzQ/edit?usp=sharing
- Design Doc 2: https://docs.google.com/document/d/1tuxcPCpoopRncuqZQvC7Gcec7Vj1nPmoK1rpkTmoLdo/edit?usp=sharing
- Design Doc 3: https://docs.google.com/document/d/102ZPkf7zIFrAKRfeLIqk69BoVWxurMl2MQGzXS-nqzY/edit?usp=sharing

The purpose of this Jira is to introduce new data type in Phoenix: Binary JSON (BSON) to manage more complex document data structures in Phoenix.

BSON or Binary JSON is a Binary-Encoded serialization of JSON-like documents. BSON data type is specifically used for users to store, update and query part or whole of the BsonDocument in the most performant way without having to serialize/deserialize the document to/from binary format. Bson allows deserializing only part of the nested documents such that querying or indexing any attributes within the nested structure becomes more efficient and performant as the deserialization happens at runtime. Any other document structure would require deserializing the binary into the document, and then perform the query.

BSONSpec: https://bsonspec.org/

JSON and BSON are closely related by design. BSON serves as a binary representation of JSON data, tailored with specialized extensions for wider application scenarios, and finely tuned for efficient data storage and traversal. Similar to JSON, BSON facilitates the embedding of objects and arrays.

One particular way in which BSON differs from JSON is in its support for some more advanced data types. For instance, JSON does not differentiate between integers (round numbers), and floating-point numbers (with decimal precision). BSON does distinguish between the two and store them in the corresponding BSON data type (e.g. BsonInt32 vs BsonDouble). Many server-side programming languages offer advanced numeric data types (standards include integer, regular precision floating point number i.e. “float”, double-precision floating point i.e. “double”, and boolean values), each with its own optimal usage for efficient mathematical operations.

Another key distinction between BSON and JSON is that BSON documents have the capability to include Date or Binary objects, which cannot be directly represented in pure JSON format. BSON also provides the ability to store and retrieve user defined Binary objects. Likewise, by integrating advanced data structures like Sets into BSON documents, we can significantly enhance the capabilities of Phoenix for storing, retrieving, and updating Binary, Sets, Lists, and Documents as nested or complex data types.

Moreover, JSON format is human as well as machine readable, whereas BSON format is only machine readable. Hence, as part of introducing BSON data type, we also need to provide a user interface such that users can provide human readable JSON as input for BSON datatype.

This Jira also introduces access and update functions for BSON documents.

BSON_CONDITION_EXPRESSION can evaluate condition expression on the document fields, similar to how WHERE clause evaluates condition expression on various columns of the given row(s) for the relational tables.

BSON_UPDATE_EXPRESSION can perform one or more document field updates similar to how UPSERT statements can perform update to one or more columns of the given row(s) for the relational tables.

Phoenix can introduce more complex data structures like sets of scalar types, in addition to the nested documents and nested arrays provided by BSON.

Overall, by combining various functionalities available in Phoenix like secondary indexes, conditional updates, high throughput read/write with BSON, we can evolve Phoenix into highly scalable Document Database.


phoenix-core-client/src/main/java/org/apache/phoenix/expression/function/BsonUpdateExpressionFunction.java
```diff
@@ -0,0 +1,133 @@
...
+import org.apache.phoenix.parse.FunctionParseNode;
+import org.apache.phoenix.schema.tuple.Tuple;
+import org.apache.phoenix.schema.types.PBson;
+import org.apache.phoenix.schema.types.PDataType;
+import org.apache.phoenix.schema.types.PVarbinary;
+import org.apache.phoenix.schema.types.PVarchar;
+import org.apache.phoenix.thirdparty.com.google.common.base.Preconditions;
+
+@FunctionParseNode.BuiltInFunction(name = BsonUpdateExpressionFunction.NAME,
+    nodeClass = BsonUpdateExpressionParseNode.class,
```
ranganathg, line 43: I am not really sure but just thinking that we can call these BsonUpdateFunction and BsonUpdateParseNode? similarly for the other node types and functions?

phoenix-core-client/src/main/java/org/apache/phoenix/expression/function/BsonUpdateExpressionFunction.java
```diff
@@ -0,0 +1,133 @@
...
+    args =
+        {
+            @FunctionParseNode.Argument(allowedTypes = {PBson.class, PVarbinary.class}),
+            @FunctionParseNode.Argument(allowedTypes = {PBson.class, PVarbinary.class},
+                    isConstant = true)
+        })
+public class BsonUpdateExpressionFunction extends ScalarFunction {
+
+    public static final String NAME = "BSON_UPDATE_EXPRESSION";
```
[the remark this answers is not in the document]
virajjasani: I thought so too, keeping JSON_MODIFY in mind. But JSON_MODIFY (or if we could say JSON_UPDATE) does one field update, it's purpose is to either set new key-value pair or replace value of existing key with new value.

Whereas BSON_UPDATE_EXPRESSION is in the form of Document based Expression i.e. it has several ways of updating document: set value, remove field and value pair, add/increment numerical value, update Set data structure with addition and removal of elements, remove list elements from specific index etc.
Hence, I kept name as update expression.

phoenix-core-client/src/main/java/org/apache/phoenix/expression/util/bson/CommonComparisonExpressionUtils.java
```diff
@@ -0,0 +1,271 @@
...
+import org.bson.BsonValue;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import org.apache.hadoop.hbase.util.Bytes;
+
+/**
+ * Common Util functions to help retrieve BSON Document values based on the given field expressions.
+ */
+public class CommonComparisonExpressionUtils {
```
ranganathg, line 36: this seems to have comparison and other helper functions - May be BsonUtils class?

phoenix-core-client/src/main/java/org/apache/phoenix/expression/util/bson/DocumentComparisonExpressionUtils.java
```diff
@@ -0,0 +1,370 @@
...
+import org.bson.BsonValue;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import org.apache.phoenix.thirdparty.com.google.common.base.Preconditions;
+
+/**
+ * Document style condition expression evaluation support.
+ */
+public class DocumentComparisonExpressionUtils {
```
ranganathg, line 37: Is Document and BSON object synonymous or used differently in different contexts?

virajjasani: They are synonymous, the package name itself is `org.apache.phoenix.expression.util.bson` so every utility within this class that refers to documents is for BsonDocument.

phoenix-core-client/src/main/java/org/apache/phoenix/expression/util/bson/CommonComparisonExpressionUtils.java
virajjasani: I can update the class name, but the package name itself has bson in it `org.apache.phoenix.expression.util.bson`, so all classes under the package refers to Bson document update/access utility classes.

ranganathg: After looking at other files I see there is a reason you named it this way.

phoenix-core-client/src/main/java/org/apache/phoenix/compile/CreateIndexCompiler.java
```diff
@@ -113,6 +114,8 @@ private String getValue(PDataType type) {
             return "ARRAY" + type.getSampleValue().toString();
         } else if (type instanceof PJson) {
             return "'{a:1}'";
+        } else if (type instanceof PBson) {
```
kadirozde, line 117: The previous line for PJson should be corrected. Can we merge two if statements into one with the OR operator?

phoenix-core-client/src/main/java/org/apache/phoenix/expression/function/BsonConditionExpressionFunction.java
```diff
@@ -0,0 +1,137 @@
...
+    public boolean evaluate(Tuple tuple, ImmutableBytesWritable ptr) {
+        if (!getColValExpr().evaluate(tuple, ptr)) {
+            return false;
+        }
+        if (ptr == null || ptr.getLength() == 0) {
+            return false;
+        }
+
+        RawBsonDocument rawBsonDocument =
+            (RawBsonDocument) PBson.INSTANCE.toObject(ptr, getColValExpr().getSortOrder());
```
kadirozde, line 76: Do we need to provide the sort oder here? It is ignored by PBson.

phoenix-core-client/src/main/java/org/apache/phoenix/expression/function/BsonConditionExpressionFunction.java
```diff
@@ -0,0 +1,137 @@
...
+    }
+
+    @Override
+    public String getName() {
+        return NAME;
+    }
+
+    @Override
+    public boolean evaluate(Tuple tuple, ImmutableBytesWritable ptr) {
+        if (!getColValExpr().evaluate(tuple, ptr)) {
```
kadirozde, line 68: Can we have a comment here to state why we need to do this first? Should the method be called getColValExpr? What does it mean here? It refers to the first child.

phoenix-core-client/src/main/java/org/apache/phoenix/expression/util/bson/CommonComparisonExpressionUtils.java
```diff
@@ -0,0 +1,271 @@
...
+
+  /**
+   * Compares two values: value and compareValue. If both have same data type and the value of
+   * actual value is less than the compareValue, then returns true. Otherwise, returns false.
+   *
+   * @param value The actual value to compare against compareValue.
+   * @param compareValue The value against which the actual value is compared.
+   * @return True if both have same data type and the value is less than compareValue.
+   */
+  public static boolean lessThan(BsonValue value, BsonValue compareValue) {
```
kadirozde, line 154: We can combine the following four methods into one compare by providing the comparison operator.

phoenix-core-client/src/main/java/org/apache/phoenix/expression/function/BsonConditionExpressionFunction.java
```diff
@@ -0,0 +1,137 @@
...
+            return false;
+        }
+
+        final RawBsonDocument conditionExpressionBsonDoc;
+        if (getConditionExpression().getDataType() == PVarchar.INSTANCE) {
+            String conditionExpression =
+                    (String) PVarchar.INSTANCE.toObject(ptr,
+                            getConditionExpression().getSortOrder());
+            if (conditionExpression == null || conditionExpression.isEmpty()) {
+                return true;
```
ranganathg, line 91: does true here imply returning the complete bson? mean like no condition so return all?

phoenix-core-client/src/main/java/org/apache/phoenix/expression/util/bson/SQLComparisonExpressionUtils.java
```diff
@@ -0,0 +1,278 @@
...
+    }
+  }
+
+  public String convertExpression(String expression) {
+    expression =
+        expression.replaceAll("field_not_exists\\(([^)]+)\\)", "!exists('$1')");
+    expression = expression.replaceAll("field_exists\\(([^)]+)\\)", "exists('$1')");
+    //    expression = expression.replaceAll("\\b([\\w.\\[\\]]+)\\b\\s*=\\s*(:\\w+)", "isEquals('$1', '$2')");
+    expression =
+        expression.replaceAll("\\b([\\w.\\[\\]]+)\\s*=\\s*([#:$]*\\w+)", "isEquals('$1', '$2')");
```
ranganathg, line 65: these regex should be declared as some constants?

phoenix-core-client/src/main/java/org/apache/phoenix/expression/util/bson/UpdateExpressionUtils.java
```diff
@@ -0,0 +1,790 @@
...
+import org.slf4j.LoggerFactory;
+
+/**
+ * BSON Update Expression Utility to perform the Document updates. All update expressions
+ * provided by this utility supports operations on nested document fields. The field key can
+ * represent any top level or nested fields within the document. The caller should use "."
+ * notation for accessing nested document elements and "[n]" notation for accessing nested array
+ * elements. Top level fields do not require any additional character.
+ */
+public class UpdateExpressionUtils {
```
ranganathg, line 51: are there any bson utils in the bson library which can expose some functionality implemented here?

phoenix-core-client/src/main/java/org/apache/phoenix/expression/function/BsonConditionExpressionFunction.java
virajjasani: since condition is empty, it is true so it is meant to be evaluated to true

phoenix-core-client/src/main/java/org/apache/phoenix/expression/util/bson/UpdateExpressionUtils.java
virajjasani: i could not find common utilities and also going through DFS traversal for nested document/array structures. however even if they had all utilities in place, we would still not be able to use it as is, because we are supporting new data structure: `Sets of same data-type elements`, so we need our own utility anyway.

phoenix-core-client/src/main/java/org/apache/phoenix/expression/function/BsonConditionExpressionFunction.java
virajjasani: done, it is evaluating BSON cell value

phoenix-core-client/src/main/java/org/apache/phoenix/expression/util/bson/CommonComparisonExpressionUtils.java
virajjasani: done

phoenix-core-client/src/main/java/org/apache/phoenix/expression/util/bson/SQLComparisonExpressionUtils.java
virajjasani: done

phoenix-core-client/src/main/java/org/apache/phoenix/expression/function/BsonConditionExpressionFunction.java
ranganathg: right but the ptr which contains the result will be currently set as the total bson document right?

phoenix-core-client/src/main/java/org/apache/phoenix/expression/util/bson/DocumentComparisonExpressionUtils.java
```diff
@@ -0,0 +1,370 @@
...
+   *     "$gte": <value>
+   *   }
+   * }
+   *
+   * @param document The document used for comparison.
+   * @param conditionExpression Condition Expression Document.
+   * @return True if the field provided in the condition expression has value greater than
+   * or equals to the value provided in the condition expression.
+   */
+  private static boolean greaterThanOrEquals(final BsonDocument document,
```
kadirozde, line 162: Can we combine all these comparison methods into one by passing the comparison operator instead of having a separate method for each operator?

phoenix-core-client/src/main/java/org/apache/phoenix/expression/util/bson/SQLComparisonExpressionUtils.java
```diff
@@ -0,0 +1,290 @@
...
+    try {
+      Object result = MVEL.eval(expression, this);
+      return result instanceof Boolean && (Boolean) result;
+    } catch (Exception e) {
+      LOGGER.error("Error while evaluating expression: {}", expression, e);
+      throw new RuntimeException("Expression could not be evaluated: " + expression);
+    }
+  }
+
+  public String convertExpression(String expression) {
```
kadirozde, line 106: This method scans an expression string 15 times for pattern matching and then 15 times for replacing matching strings. A conventional parsing would need only one pass.  Also it creates 15 matcher objects and 15 new expressions via this process. This looks computationally pretty expensive.

virajjasani: Yes, the string is parsed 15 times for pattern matching and this could be a bit time consuming and I have been exploring a couple of options:

1. Using Pattern matcher replacement with StringBuffer
2. SearchAndReplaceBuilder from DFALEX: https://github.com/mtimmerm/dfalex/commit/a1bc6aca051b7b2cbdc7e2c9e639fe8a1e3cd16a
3. commons-text StringSubstitutor

None of these options are exactly helping with what we want.

I have also been measuring as large string as i could generate. With all the strings available within tests, the max time spent in the computation is 1ms (it fluctuates b/ 0 and 1 ms).

I am still exploring more options but so far have not seen more efficient multi-pattern matcher for single string in Java.

virajjasani: Also, earlier i was doing String#replaceAll, which would compile regex pattern each time string replacement happens. With the latest commit, i have separated out pattern compilation and kept it static so it happens only once: https://github.com/apache/phoenix/pull/1906/commits/ace8240f1df4c027ce2458309f8ed78c05f5484b

phoenix-core-client/src/main/java/org/apache/phoenix/expression/function/BsonConditionExpressionFunction.java
virajjasani: Correct

phoenix-core-client/src/main/java/org/apache/phoenix/expression/util/bson/SQLComparisonExpressionUtils.java
virajjasani: Updated PR to reduce the num of pattern matching and replacement. The pattern matching with replacement will be in direct correlation with the num of unique operators used in the String.

virajjasani: Moreover, to further reduce the dependency on the pattern-matcher, we can also document option for client to use executable functions directly i.e. instead of `a = :b`, client can directly provide `isEquals('a', ':b')`, similarly instead of `AND`, client can provide `&&` etc, in which case we will no longer perform pattern-matcher-replacer for the input string.

virajjasani: In the latest revision, I just added additional tests that no longer relies on pattern-matcher for each matcher replacement, as the string input itself is executable.

virajjasani: I had an offline discussion with @kadirozde reg this. I will create a separate follow-up Jira to use ANTLR based parsing for Bson SQL style condition expression evaluation.

phoenix-core-client/src/main/java/org/apache/phoenix/expression/util/bson/UpdateExpressionUtils.java
```diff
@@ -0,0 +1,790 @@
...
+          updateNestedFieldByDelete(value, i, fieldKey, newVal);
+          return;
+        } else {
+          sb.append(fieldKey.charAt(i));
+        }
+      }
+    }
+  }
+
+  private static void updateNestedFieldByDelete(final BsonValue value, final int idx,
```
kadirozde, line 146: It will be good to explain the meaning of the parameters and the basic logic in the method.

phoenix-core-client/src/main/java/org/apache/phoenix/expression/util/bson/DocumentComparisonExpressionUtils.java
virajjasani: Yes i can make this, but since we need separate methods in SQL style comparison, i kept it this way:

- SQL style expression -> contains individual comparison methods -> CommonComparisonExpressionUtils -> single method for all comparisons
- Document style expression -> contains individual comparison methods -> CommonComparisonExpressionUtils -> single method for all comparisons

When i introduce ANTLR style expression evaluation, i can make this change as well:

- SQL style expression -> single method for all comparisons -> CommonComparisonExpressionUtils -> single method for all comparisons
- Document style expression -> single method for all comparisons -> CommonComparisonExpressionUtils -> single method for all comparisons

@kadirozde does this sound good?

phoenix-core-client/src/main/java/org/apache/phoenix/expression/util/bson/DocumentComparisonExpressionUtils.java
```diff
@@ -0,0 +1,370 @@
...
+
+  private static final String EXISTS_OP = "$exists";
+  private static final String EQUALS_OP = "$eq";
+  private static final String NOT_EQUALS_OP = "$ne";
+  private static final String LESS_THAN_OP = "$lt";
+  private static final String LESS_THAN_OR_EQUALS_OP = "$lte";
+  private static final String GREATER_THAN_OP = "$gt";
+  private static final String GREATER_THAN_OR_EQUALS_OP = "$gte";
+
+  public static boolean isConditionExpressionMatching(final BsonDocument document,
```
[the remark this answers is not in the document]
virajjasani: Sure, will do

phoenix-core-client/src/main/java/org/apache/phoenix/expression/util/bson/SQLComparisonExpressionUtils.java
```diff
@@ -0,0 +1,344 @@
...
+  private final RawBsonDocument rawBsonDocument;
+  private final BsonDocument comparisonValuesDocument;
+
+  public SQLComparisonExpressionUtils(RawBsonDocument rawBsonDocument,
+      BsonDocument comparisonValuesDocument) {
+    this.rawBsonDocument = rawBsonDocument;
+    this.comparisonValuesDocument = comparisonValuesDocument;
+  }
+
+  public boolean isConditionExpressionMatching(final String conditionExpression) {
```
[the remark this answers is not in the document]
virajjasani: Sounds good

phoenix-core-client/src/main/java/org/apache/phoenix/expression/util/bson/SQLComparisonExpressionUtils.java
```diff
@@ -0,0 +1,344 @@
...
+
+  /**
+   * Returns true if the value of the field is less than the value represented by {@code
+   * expectedFieldValue}. The comparison can happen only if the data type of both values match.
+   *
+   * @param fieldKey The field key for which value is compared against expectedFieldValue.
+   * @param expectedFieldValue The literal value to compare against the field value.
+   * @return True if the value of the field is less than expectedFieldValue.
+   */
+  public boolean lessThan(final String fieldKey, final String expectedFieldValue) {
```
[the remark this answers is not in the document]
virajjasani: Yes i can make this, but since we need separate methods in SQL style comparison as this PR uses MVEL expression language and having separate function calls are clean for expression language, i kept it this way:

- SQL style expression -> contains individual comparison methods -> CommonComparisonExpressionUtils -> single method for all comparisons
- Document style expression -> contains individual comparison methods -> CommonComparisonExpressionUtils -> single method for all comparisons

When i introduce ANTLR style expression evaluation as follow up Jira, i can make this change as well:

- SQL style expression -> single method for all comparisons -> CommonComparisonExpressionUtils -> single method for all comparisons
- Document style expression -> single method for all comparisons -> CommonComparisonExpressionUtils -> single method for all comparisons

@kadirozde does this sound good?

phoenix-core-client/src/main/java/org/apache/phoenix/expression/util/bson/DocumentComparisonExpressionUtils.java
kadirozde: You can still have a separate method for each of them if you need but they should call a common method to eliminate duplicate code.

phoenix-core-client/src/main/java/org/apache/phoenix/expression/util/bson/SQLComparisonExpressionUtils.java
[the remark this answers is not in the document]
kadirozde: You can still have a separate method for each of them if you need but they should call a common method to eliminate duplicate code.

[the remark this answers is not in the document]
virajjasani: Yes, they all call common method eventually: https://github.com/apache/phoenix/pull/1906/files#diff-d0b48fb7d6da791c1eb8a7abccad83a3e891396695be2e744983583217aeac23R154

```
  public static boolean compareValues(BsonValue value, BsonValue compareValue, CompareOp op) {
    Preconditions.checkArgument(op != null, "Comparison operator should not be null");
    if (value.isString() && compareValue.isString()) {
      int compare =
              ((BsonString) value).getValue().compareTo(((BsonString) compareValue).getValue());
      switch (op) {
        case LESS:
          return compare < 0;
        case LESS_OR_EQUAL:
          return compare <= 0;
        case GREATER:
          return compare > 0;
        case GREATER_OR_EQUAL:
          return compare >= 0;
      }
    }
...
...
```

Here, CompareOp takes the comparison operator enum value and evaluates accordingly

phoenix-core-client/src/main/java/org/apache/phoenix/expression/util/bson/DocumentComparisonExpressionUtils.java
kadirozde: I am suggesting something like this:
```
private static boolean compare(final BsonDocument document,
      final BsonDocument conditionExpression, CommonComparisonExpressionUtils.CompareOp op) {
    Set<Map.Entry<String, BsonValue>> entrySet = conditionExpression.entrySet();
    Preconditions.checkArgument(entrySet.size() == 1, "Expected entry size for a compare operation"
        + " is 1");
    for (Map.Entry<String, BsonValue> bsonValueEntry : entrySet) {
      String fieldKey = bsonValueEntry.getKey();
      BsonValue bsonValue = bsonValueEntry.getValue();
      Preconditions.checkArgument(bsonValue instanceof BsonDocument,
          "Expected type for Bson value is Document for a compare operation");
      BsonDocument bsonDocument = (BsonDocument) bsonValue;
      BsonValue compareValue = bsonDocument.get(getBsonCompareOp(op)); // need to implement getBsonCompareOp
      BsonValue topLevelValue = document.get(fieldKey);
      BsonValue actualValue = topLevelValue != null ? topLevelValue
          : CommonComparisonExpressionUtils.getFieldFromDocument(fieldKey, document);
      return actualValue != null && CommonComparisonExpressionUtils.compareValues(actualValue,
          compareValue, op);
    }
    return false;
  }
private static boolean greaterThanOrEquals(final BsonDocument document,
      final BsonDocument conditionExpression) {

    return compare(document, conditionExpression, CommonComparisonExpressionUtils.CompareOp.GREATER_OR_EQUAL);
}
private static boolean greaterThan(final BsonDocument document,
      final BsonDocument conditionExpression) {

    return compare(document, conditionExpression, CommonComparisonExpressionUtils.CompareOp.GREATER);
}
and so on
```

virajjasani: I see, you mean within the individual classes SQLComparisonExpressionUtils and DocumentComparisonExpressionUtils also, we should have common comparison method, which can directly call common comparison method from CommonComparisonExpressionUtils.

Sounds good improvement, let me change it now.

virajjasani: Done

phoenix-core-client/src/main/java/org/apache/phoenix/expression/util/bson/SQLComparisonExpressionUtils.java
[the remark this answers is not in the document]
virajjasani: This is taken care of in the latest revision.

phoenix-core-client/src/main/java/org/apache/phoenix/expression/util/bson/UpdateExpressionUtils.java
virajjasani: Done

phoenix-core-client/src/main/java/org/apache/phoenix/expression/util/bson/UpdateExpressionUtils.java
```diff
@@ -0,0 +1,949 @@
...
+   *    "$UNSET": { &lt;field1&gt;: null, &lt;field2&gt;: null, ... },
+   *    "$ADD": { &lt;field1&gt;: &lt;value1&gt;, &lt;field2&gt;: &lt;value2&gt;, .... },
+   *    "$DELETE_FROM_SET": { &lt;field1&gt;: &lt;value1&gt;, &lt;field2&gt;: &lt;value2&gt;, .... }
+   * }
+   * <p/>
+   * "$SET": Use the SET action in an update expression to add one or more fields to a BSON
+   * Document. If any of these fields already exists, they are overwritten by the new values.
+   * To perform multiple SET actions, provide multiple fields key-value entries within nested
+   * document under $SET field key.
+   * "$UNSET": Use the REMOVE action in an update expression to remove one or more fields from a
```
kadirozde, line 69: Both the document and PR makes references to keyword "REMOVE" which does not exist. Should we replace REMOVE with UNSET? Or is REMOVE a better choice than UNSET as a keyword here?

virajjasani: Ok let me fix the description and refer to `$UNSET` only.

REMOVE is also fine, i started off with "$REMOVE" and then switched to "$UNSET".

phoenix-core-client/src/main/java/org/apache/phoenix/expression/util/bson/UpdateExpressionUtils.java
```diff
@@ -0,0 +1,950 @@
...
+   * @param value Bson value at the given level of the document hierarchy.
+   * @param idx The index value for the given field key. The function is expected to retrieve
+   * the value of the nested document or array at the given level of the tree.
+   * @param fieldKey The full field key.
+   * @param setValuesToDelete The set values that need to be removed from the existing set.
+   */
+  private static void updateNestedFieldByDelete(final BsonValue value,
+      final int idx, final String fieldKey, final BsonValue setValuesToDelete) {
+    int curIdx = idx;
+    if (fieldKey.charAt(curIdx) == '.') {
```
kadirozde, line 189: There is too much repeated code here to find a entry (key-value) in a document. We can reduce the code by having a common method that returns an entry for a given field from an update expression. The same method can be reused for the fields from a condition expression too, I think. By doing so, the recursion and iterating over keys for . and [ can be contained and implemented in one place.

virajjasani: For condition expression we already have common utility method that perform tree traversal one level at a time and retrieves the deeply nested fields: `CommonComparisonExpressionUtils#getFieldFromDocument()`. This is being used at many places.
For update expression, it is really complicated to maintain such common tree traversal logic because we not only need to traverse the tree, we also need to update the leaf node or intermediate node of the tree. Doing so requires different BSON APIs for different update operation. I have tried to keep common logic separate but its really making update expression very complicated. Some behaviors are unique to individual update operation e.g. DELETE_FROM_SET needs to remove the field if we remove all elements of the set. ADD needs to perform arithmetic or set union operation depending on the data type. Maintaining all this logic with tree traversal as separate common function is becoming too complex to manage.

virajjasani: Let me try refactoring some portion within update expression, however sharing the common portion with condition expression does not seem feasible due to inherent complexity involved with individual update operators.

virajjasani: Refactor for update expression done https://github.com/apache/phoenix/pull/1906/commits/34f3675bbb623275e7344c195a2f0e66afdde364

kadirozde: Your refactoring reduced the code significantly. However, you did it in a different way than what I suggested. I did not suggested writing a common method for updating a field. As you can see from your refactoring, the code reduction comes from not repeating the code that traverses the BSON document to find the entry for a given field name (the path to an entry). Your recursive methods return void currently. They are written in procedural programming (using side effects). The common (recursive) method I suggested should be written in functional programming (without any side effect). It should take a Bson document and a filed name and return an entry. Does that make sense?

virajjasani: While condition expression requires the retrieval of the field name key-value entry by traversing the tree, update expression has complicated behavior depending on the update operation. So if we were to write function that returns BsonValue, it would be for the ancestor node rather than the leaf node that requires update. For example, path1.path2.path3[4] requires function returning path1.path2.path3 so that update expression can perform its logic on path3, whereas for path1.path2.path3, the function needs to return path1.path2 value. While this seems doable, but we need separate function anyway because now it needs to return the ancestor node and not the leaf node itself.

Such function cannot be shared with the condition expression.

virajjasani: Also, even if we use function to return ancestor node, for (deeply) nested arrays, the caller still needs the array index on which to perform the update operation so the caller will need to parse the field key string from the last index position to find the leaf node (nested array) index position and then perform the operation. Hence, this approach is also not any simpler than the current way of doing recursion (procedural way) in which while we return void, we do take care of updating the value at the target leaf node. The recursion followed by updating the target node is not expected to be buggy.

kadirozde: I did not understand the complication you mentioned if the method I suggested returns the list of Bson documents instead of just the document corresponding to the last component of the path. The Bson docs can be added to the list when the method traverses the entire document from top to bottom. I am not suggesting to change the restructuring that you have just done so far. I am suggesting that you can further improve your restructuring by extracting the find method as a common method. The problem statement begs for this method as it is "update/delete/check the value of the bson document for a given path".

virajjasani: The list of BsonValue can be returned but depending on whether the update method (caller) overrides the already present BsonValue or needs to add new BsonValue, it needs to have field key associated with each BsonValue in the list. While we can use LinkedHashMap and let caller get last pair from LinkedHashMap<String, BsonValue>, the caller still needs to traverse the field key again to get the last portion of the key back. For example, for path1.path2.path3[4], the caller still needs to recognize and have different logic for path3[4] vs path4 (if the field key was path1.path2.path3.path4). The caller still needs to retrieve the value of path4 or path3[4] as an additional parsing.

The current version of update recursive function looks far simpler.

kadirozde: I did not understand the need for LinkedHashMap instead of just ArrayList and did not understand why doc vs array update cannot be resolved easily. Let's have a side conversation on this.

phoenix-core-client/src/main/java/org/apache/phoenix/expression/util/bson/UpdateExpressionUtils.java
```diff
@@ -0,0 +1,794 @@
...
+   *
+   * @param fieldKey The full field key.
+   * @param newVal New value to be used while performing update operation on the existing node.
+   * @param updateOp Type of the update operation to be performed.
+   * @param arrayIdx The index of the array at which the target node value is to be updated.
+   * @param nestedList The parent or ancestor node, expected to be nested array only.
+   * For example, for "category.subcategories.brands[5]" field key, the nestedList is expected to
+   * be "brands" array and arrayIdx is expected to be 5.
+   */
+  private static void updateValueAtLeafNode(final String fieldKey,
```
kadirozde, line 463: This should be called updateArray instead of updateValue as this is specific to an Array field.

phoenix-core-client/src/main/java/org/apache/phoenix/expression/util/bson/UpdateExpressionUtils.java
```diff
@@ -0,0 +1,794 @@
...
+   * Update the existing set by removing the set values that are present in
+   * {@code setValuesToDelete}. For this operation to be successful, both {@code currentValue} and
+   * {@code setValuesToDelete} must be of same set data type. For instance, both must be either
+   * set of string, set of numbers or set of binary values.
+   *
+   * @param currentValue The value that needs to be updated by performing deletion operation.
+   * @param setValuesToDelete The set values that need to be deleted from the currentValue set.
+   * @return Updated set after performing the set difference operation.
+   */
+  private static BsonValue modifyFieldValueByDelete(final BsonValue currentValue,
```
kadirozde, line 182: This should be called modifyFieldValueByDeleteFromSet

kadirozde (approved):
+1, Thanks!

phoenix-core-client/src/main/java/org/apache/phoenix/expression/util/bson/UpdateExpressionUtils.java
virajjasani: looks like i missed this, let me make this change.

PR #1907, opened 2024-06-18
PHOENIX-7331 Fix incompatibilities with HBASE-28644

stoty:


stoty:
I haven't reviewed if we really need on-heap cells, I only copied the ensureKeyValue* methods from HBase.

Apache9 (approved):
For now I think it is OK, but I suggest that we use ExtendedCell instead of KeyValue in the future as lots of newly introduced offheap cell types are not KeyValue...

stoty:
The problem is that in Phoenix it is not always clear why we convert to KeyValue.
In some cases it's just lazyness, as we didn't upgrade some internal APIs in to Cells.

However, in other cases it's to make sure that we copy off-heap cells to on-heap to avoid use-after-free issues.

Unfortunately, it takes a lot of investigation to separate the two use cases, and no-one made the time for it yet.

virajjasani (approved):

stoty:
merged manually

PR #1900, opened 2024-05-31
PHOENIX-7041 (ADDENDUM-2): ROW_KEY_MATCHER should support various tenantId data types

jpisaac:

PR #1855, opened 2024-03-12
PHOENIX-7108 Provide support for pruning expired rows of views using Phoenix level compactions

jpisaac:

PR #1908, opened 2024-06-18
PHOENIX-7108 Provide support for pruning expired rows of views using Phoenix level compactions

jpisaac:


jpisaac:
Rebased the old [PR](https://github.com/apache/phoenix/pull/1855) to include the PHOENIX-7041 (ADDENDUM-2) - [PR](https://github.com/apache/phoenix/pull/1900/commits)

phoenix-core/src/main/java/org/apache/phoenix/util/ByteUtil.java
```diff
@@ -508,6 +508,26 @@ public static byte[] fillKey(byte[] key, int length) {
...
+        if (b == null) {
+            return true;
+        }
+        if (Bytes.compareTo(b, HConstants.EMPTY_BYTE_ARRAY) == 0 ) {
+            return true;
+        }
+        if (b.length < offset + length) {
+            throw new IllegalStateException();
+        }
+        boolean result = true;
```
ranganathg, line 521: I think we can rewrite to something like this?
`        
        for (int i = (offset + length) - 1; i >= offset; i--) {
            if (b[i] != 0x0) {
                return false;
            }
        }
        return true;`

phoenix-core/src/main/java/org/apache/phoenix/util/ByteUtil.java
```diff
@@ -508,6 +508,26 @@ public static byte[] fillKey(byte[] key, int length) {
         return newBound;
     }
 
+    public static boolean isEmptyOrNull(byte[] b, int offset, int length) {
+        if (b == null) {
```
ranganathg, line 512: we can merge these 2 ifs to if(b=null || bytes.compare..) return true;?

phoenix-core/src/main/java/org/apache/phoenix/util/matcher/TableTTLInfo.java
```diff
@@ -0,0 +1,115 @@
...
+	public int hashCode() {
+		return Arrays.hashCode(tenantId) +  Arrays.hashCode(entityName);
+	}
+	@Override
+	public int compareTo(Object obj) {
+		if (this == obj)
+			return 0;
+		if (obj == null)
+			throw new NullPointerException();
+		TableTTLInfo other = (TableTTLInfo) obj;
```
ranganathg, line 93: is there a need to do similar check as this? if(getClass() != o.getClass() ) throw somException?

phoenix-core/src/main/java/org/apache/phoenix/util/matcher/TableTTLInfo.java
```diff
@@ -0,0 +1,115 @@
...
+		return physicalTableName;
+	}
+
+
+	@Override
+	public boolean equals(Object o) {
+		if (this == o) return true;
+		if (o == null || getClass() != o.getClass()) return false;
+		TableTTLInfo that = (TableTTLInfo) o;
+		return Arrays.equals(physicalTableName, that.physicalTableName) &&
```
ranganathg, line 78: which one is the right one to compare? Bytes.BYTES_COMPARATOR.compare or Arrays.equals? below I see BYTES_COMPARATOR is being used. so trying to understand.

kadirozde (approved):
+1, Thanks!

phoenix-core/src/main/java/org/apache/phoenix/util/matcher/TableTTLInfo.java
jpisaac: I think the BYTES_COMPARATOR makes sense when you want to compare and especially when comparing lexographically.  When checking for equality either works fine.

phoenix-core/src/main/java/org/apache/phoenix/util/matcher/TableTTLInfo.java
jpisaac: The caller will get a ClassCastException if the such is the case.

PR #1891, opened 2024-05-13
Phoenix CSV BulkloadTool fails with "Global tag is not allowed" error on transactional table

nikita15p:
This changes removes the shading of `org.apache.omid` and hence trusted lists defines in OMID goes through successfully and don't error out.

PR #1914, opened 2024-06-20
 Phoenix CSV BulkloadTool fails with "Global tag is not allowed" error on transactional table

nikita15p:
This changes removes the shading of org.apache.omid and hence trusted lists defines in OMID goes through successfully and don't error out.

Raised another PR from new branch name. Changes are same as https://github.com/apache/phoenix/pull/1891

@stoty please review


stoty:
merged manually.

PR #1915, opened 2024-06-20
PHOENIX-6066 : MetaDataEndpointImpl.doGetTable should acquire a readLock instead of an exclusive writeLock 

palashc:


phoenix-core-client/src/main/java/org/apache/phoenix/query/QueryServices.java
```diff
@@ -486,6 +486,8 @@ public interface QueryServices extends SQLCloseable {
 
     int DEFAULT_PHOENIX_GET_REGIONS_RETRIES = 10;
 
+    String PHOENIX_GET_METADATA_READ_LOCK_ENABLED = "phoenix.get.metadata.read.lock.enabled";
```
dbwong, line 489: Whats the risk that requires this be parameterized?  Also when did the previous change with exclusive writelock go ou?

phoenix-core/src/it/java/org/apache/phoenix/end2end/MetadataGetTableReadLockIT.java
```diff
@@ -0,0 +1,134 @@
...
+ */
+@Category(ParallelStatsDisabledTest.class)
+public class MetadataGetTableReadLockIT extends BaseTest {
+
+    @BeforeClass
+    public static synchronized void doSetup() throws Exception {
+        Map<String, String> props = Maps.newHashMapWithExpectedSize(1);
+        // Disable system task handling
+        props.put(QueryServices.TASK_HANDLING_INITIAL_DELAY_MS_ATTRIB, Long.toString(Long.MAX_VALUE));
+        props.put(QueryServices.PHOENIX_GET_METADATA_READ_LOCK_ENABLED, Boolean.toString(true));
```
dbwong, line 35: Technically don't have either an integration/unit test that checks that when this is false there is no change.

dbwong (commented):
Generally good, couple of questions.

phoenix-core-client/src/main/java/org/apache/phoenix/query/QueryServices.java
palashc: Just the ability to quickly to disable this change in production if it causes any issues.

dbwong: Is it not enabled prior to this release?

phoenix-core/src/it/java/org/apache/phoenix/end2end/MetadataGetTableReadLockIT.java
palashc: We decided to keep the config enabled by default.

phoenix-core-client/src/main/java/org/apache/phoenix/query/QueryServices.java
palashc: > Also when did the previous change with exclusive write lock go out?
@dbwong This has always been the behaviour in production. We want to change it to using read locks for getTable RPC calls.

PR #1919, opened 2024-06-27
PHOENIX-6066 : MetaDataEndpointImpl.doGetTable should acquire a readLock instead of an exclusive writeLock

palashc:

PR #1921, opened 2024-06-28
PHOENIX-6066 : MetaDataEndpointImpl.doGetTable should acquire a readL…

palashc:
…ock instead of an exclusive writeLock (#1919)

* non exclusive row lock for doGet apis

* add config

* add test

* change test annotation

* config enabled by default

* add license to test class

---------


tkhurana (approved):

tkhurana:
@palashc Let's wait for the test results before merging

PR #1923, opened 2024-06-30
PHOENIX-7287 Leverage bloom filters for multi-key point lookups

kadirozde:


phoenix-core-server/src/main/java/org/apache/phoenix/coprocessor/PagingRegionScanner.java
```diff
@@ -59,8 +70,32 @@ public PagingRegionScanner(Region region, RegionScanner scanner, Scan scan) {
...
+            // Check if the scan is a multi-point-lookup scan
+            skipScanFilter = ScanUtil.getSkipScanFilter(scan);
+            if (skipScanFilter != null && skipScanFilter.isMultiKeyPointLookup()) {
+                isMultiKeyPointLookup = true;
+                multiKeyPointLookupRanges = skipScanFilter.getPointLookupKeyRanges();
+            }
+        }
     }
 
+    private void setMultiKeyPointLookupPosition(byte[] rowKey) throws IOException {
```
virajjasani, line 86: Since this is meant to be used when high level scanners need to re-scan the already scanned rows, this is kind of resetting `multiKeyPointLookupPosition` to 0 and then iterating through key ranges. So, shall we call this `resetMultiKeyPointLookupPosition` or `setMultiKeyPointLookupPositionToAdjustedRowKey`?

phoenix-core-server/src/main/java/org/apache/phoenix/coprocessor/PagingRegionScanner.java
```diff
@@ -101,20 +149,28 @@ private boolean next(List<Cell> results, boolean raw) throws IOException {
...
                 pagingFilter.init();
             }
+            if (isMultiKeyPointLookup) {
+                --multiKeyPointLookupPosition;
+            }
             throw e;
+        } finally {
+            if (isMultiKeyPointLookup) {
+                delegate.close();
+            }
```
virajjasani, line 173: Can we close the delegate within `if (isMultiKeyPointLookup)` condition, just to maintain the pattern similar to adjusted scan boundaries?

phoenix-core-server/src/main/java/org/apache/phoenix/coprocessor/PagingRegionScanner.java
```diff
@@ -59,8 +70,32 @@ public PagingRegionScanner(Region region, RegionScanner scanner, Scan scan) {
...
+    private void setMultiKeyPointLookupPosition(byte[] rowKey) throws IOException {
+        multiKeyPointLookupPosition = 0;
+        for (KeyRange keyRange : multiKeyPointLookupRanges) {
+            if (Bytes.compareTo(keyRange.getLowerRange(), rowKey) == 0) {
+                return;
+            } else {
+                multiKeyPointLookupPosition++;
+            }
+        }
+        throw new DoNotRetryIOException("Row key " + rowKey
```
virajjasani, line 95: `Bytes.toStringBinary(rowKey)` instead of printing raw bytes?

phoenix-core-server/src/main/java/org/apache/phoenix/coprocessor/PagingRegionScanner.java
```diff
@@ -48,6 +53,12 @@ public class PagingRegionScanner extends BaseRegionScanner {
     private Region region;
     private Scan scan;
     private PagingFilter pagingFilter;
+    private SkipScanFilter skipScanFilter;
+    private int keyIndex = 0;
+    private List<KeyRange> multiKeyPointLookupRanges = null;
+    private int multiKeyPointLookupPosition = 0;
+    private boolean isMultiKeyPointLookup = false;
+    private Cell lastCell = null;
```
virajjasani, line 61: nit: looks like it is not being used

phoenix-core-server/src/main/java/org/apache/phoenix/coprocessor/PagingRegionScanner.java
kadirozde: Yes, we should do that.

phoenix-core-server/src/main/java/org/apache/phoenix/coprocessor/PagingRegionScanner.java
kadirozde: We also need to close always even an exception is thrown. Please note we close the current scanner after scanning a row. In adjusted scan boundaries, we close the previous scanner not the one to be scanned. So, closing it at finally makes sense. Does not it?

virajjasani: Sounds good

virajjasani:
Overall changes look good

phoenix-core-client/src/main/java/org/apache/phoenix/filter/SkipScanFilter.java
```diff
@@ -618,7 +632,8 @@ public void readFields(DataInput in) throws IOException {
                 orClause.add(range);
             }
         }
-        this.init(slots, slotSpan, schema, includeMultipleVersions);
+        boolean isPointLookup = in.readBoolean();
```
tkhurana, line 635: For backward compatibility we need to handle EOFException because the older clients won't be sending the boolean field

kadirozde: Good point! I will do that.

phoenix-core-client/src/main/java/org/apache/phoenix/filter/SkipScanFilter.java
```diff
@@ -91,27 +92,39 @@ private enum Terminate {AT, AFTER};
...
     
     public void setOffset(int offset) {
         this.offset = offset;
     }
+    public boolean isMultiKeyPointLookup() {
+        return isMultiKeyPointLookup;
+    }
+
+    public List<KeyRange> getPointLookupKeyRanges() {
+        return slots.get(0);
```
tkhurana, line 123: This API should check if isMultiKeyPointLookup is true then return the 0th element else return an empty list. This will be similar to how some of the APIs in ScanRanges class work like `getPointLookupCount` and `getPointLookupKeyIterator`

phoenix-core-server/src/main/java/org/apache/phoenix/coprocessor/PagingRegionScanner.java
```diff
@@ -59,8 +70,32 @@ public PagingRegionScanner(Region region, RegionScanner scanner, Scan scan) {
         if (pagingFilter != null) {
             pagingFilter.init();
         }
+
+        TableDescriptor tableDescriptor = region.getTableDescriptor();
+        BloomType bloomFilterType = tableDescriptor.getColumnFamilies()[0].getBloomFilterType();
+        if (bloomFilterType == BloomType.ROW) {
+            // Check if the scan is a multi-point-lookup scan
+            skipScanFilter = ScanUtil.getSkipScanFilter(scan);
```
tkhurana, line 78: Do we need to remove the skipscan filter from the scan also ?

kadirozde:
The support for local indexes is not fully working yet and so I converted the PR to a draft PR.

phoenix-core-server/src/main/java/org/apache/phoenix/coprocessor/PagingRegionScanner.java
kadirozde: We do not have to remove the skip scan filter and but it makes sense to remove it for simplicity and efficiency (that is one less filter).

phoenix-core-server/src/main/java/org/apache/phoenix/coprocessor/PagingRegionScanner.java
```diff
@@ -81,40 +201,70 @@ private boolean next(List<Cell> results, boolean raw) throws IOException {
...
                     if (results.isEmpty()) {
                         byte[] rowKey = pagingFilter.getCurrentRowKeyToBeExcluded();
                         LOGGER.info("Page filter stopped, generating dummy key {} ",
                                 Bytes.toStringBinary(rowKey));
                         ScanUtil.getDummyResult(rowKey, results);
+                        if (multiKeyPointLookup != null) {
+                            multiKeyPointLookup.lastDummyCell = results.get(0);
+                            multiKeyPointLookup.lookupPosition--;
+                        }
+                        return true;
```
virajjasani, line 243: I think the problem here is that if we return from here, region can move and then it will start off with dummy rowkey which we cannot use for point lookup. Each coproc that performs complex or long running scans have handling of this logic but simple point lookup or skip scan might not need this handling, so i think if we can resume scan here rather than return to client, that could help with the problem of region moving immediately after returning dummy rowkey.

Since we know it's point lookup, we can resume the scan here from PagingRegionScanner only, because getting dummy for point lookup might only happen with tests (due to 0ms timeout) but in real production getting dummy for point lookup would be extremely rare event and even if it does, it might be better to resume scan here because it's single row only.

kadirozde: When it happens, the PR reverts to doing a range scan with the skip scan filter to retrieve the next row, and then goes back to the point lookup scans.

phoenix-core-server/src/main/java/org/apache/phoenix/coprocessor/PagingRegionScanner.java
kadirozde: It turned out that we have to keep the skip scan filter to handle region moves. In that case, we need to do range scan with the skip scan filter to retrieve the next row.

kadirozde:
The tests are locally passing for me. The PR is ready for review now.

phoenix-core-server/src/main/java/org/apache/phoenix/coprocessor/PagingRegionScanner.java
virajjasani: Got it, this would be good point to add comment here?

phoenix-core-server/src/main/java/org/apache/phoenix/coprocessor/PagingRegionScanner.java
virajjasani: We already have some tests that would fail if we remove the filter right?

kadirozde: Yes, there are such tests with region moves.

phoenix-core-client/src/main/java/org/apache/phoenix/filter/SkipScanFilter.java
```diff
@@ -208,10 +226,11 @@ private boolean areSlotsSingleKey(int startPosInclusive, int endPosExclusive) {
         return true;
     }
     
-    private void resetState() {
+    public void resetState() {
```
kadirozde, line 229: No need.for this any more

tkhurana (approved):

PR #1932, opened 2024-07-10
PHOENIX-7354 Disable TransformToolIT.testInvalidRowsWithTableLevelMax…

stoty:
…Lookback() on HBase 2.6+

also revert the ineffective fix added in PHOENIX-7353

### DIFF
--- a/phoenix-core/src/it/java/org/apache/phoenix/end2end/transform/TransformToolIT.java
+++ b/phoenix-core/src/it/java/org/apache/phoenix/end2end/transform/TransformToolIT.java
@@ -32,6 +32,7 @@ import org.apache.phoenix.thirdparty.com.google.common.collect.Maps;
 import org.apache.hadoop.conf.Configuration;
 import org.apache.hadoop.hbase.TableName;
 import org.apache.hadoop.hbase.util.Bytes;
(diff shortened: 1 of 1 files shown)


virajjasani (commented):
I think we should just disable the tests until someone properly fixes it. WDYT @stoty?

stoty:
That's what assumeFalse does, @virajjasani .
It disables the test if the HBase version is 2.6.0 or higher.

I see no reason to disable the test when running on HBase versions that can handle the stopped clock.
We already a few similar tests that are only ran with certain HBase versions.

virajjasani:
Correct, but in those tests, we usually ignore tests for lower versions because new feature might have dependency on new version HBase clients whereas here we have some buggy behaviour that doesn't work for newer versions. So what I meant is that if anything doesn't work for newer versions, should be marked ignore for all versions because we might either have bug in Phoenix side or HBase side. If test keeps running for older versions, no one might fix it :)

But it's fine either way, upto you.

virajjasani (approved):

stoty:
I'm not positive that @Ignore-ing the test results in better chances of it having fixed than only running it for old versions.
I will add a comment explaining the problem, and open a follow-up ticket.

stoty:
merged manually.

PR #1936, opened 2024-07-11
PHOENIX-5117 : Return the count of rows scanned in HBase

palashc:


phoenix-core-client/src/main/java/org/apache/phoenix/compile/QueryCompiler.java
```diff
@@ -125,21 +125,25 @@ public class QueryCompiler {
...
-        this(statement, select, resolver, new BindManager(statement.getParameters()), targetColumns, parallelIteratorFactory, sequenceManager, projectTuples, optimizeSubquery, dataPlans);
+        this(statement, select, resolver, new BindManager(statement.getParameters()), targetColumns, parallelIteratorFactory, sequenceManager, projectTuples, optimizeSubquery, dataPlans, null);
     }
 
-    public QueryCompiler(PhoenixStatement statement, SelectStatement select, ColumnResolver resolver, BindManager bindManager, List<? extends PDatum> targetColumns, ParallelIteratorFactory parallelIteratorFactory, SequenceManager sequenceManager, boolean projectTuples, boolean optimizeSubquery, Map<TableRef, QueryPlan> dataPlans) throws SQLException {
-        this.statement = statement;
+    public QueryCompiler(PhoenixStatement statement, SelectStatement select, ColumnResolver resolver, List<? extends PDatum> targetColumns, ParallelIteratorFactory parallelIteratorFactory, SequenceManager sequenceManager, boolean projectTuples, boolean optimizeSubquery, Map<TableRef, QueryPlan> dataPlans, StatementContext parentContext) throws SQLException {
```
tkhurana, line 142: Don't change public API signatures. It causes API compatibility issues.

phoenix-core-client/src/main/java/org/apache/phoenix/compile/QueryCompiler.java
```diff
@@ -705,6 +717,9 @@ protected QueryPlan compileSubquery(
                     statement,
                     queryPlan);
         }
+        if (parentContext != null) {
+            parentContext.addSubStatementContext(queryPlan.getContext());
```
tkhurana, line 721: What is the reasoning behind this change?

phoenix-core/src/it/java/org/apache/phoenix/monitoring/CountRowsScannedIT.java
```diff
@@ -0,0 +1,299 @@
...
+        }
+        return getRowsScanned(rs);
+    }
+
+    private long getRowsScanned(ResultSet rs) {
+        if (!(rs instanceof PhoenixResultSet)) {
+            return -1;
+        }
+        PhoenixResultSet phoenixResultSet = (PhoenixResultSet) rs;
+        Map<String, Map<MetricType, Long>> metrics = phoenixResultSet.getReadMetrics();
```
tkhurana, line 282: Better to use `PhoenixRuntime.getRequestReadMetricInfo`

tkhurana: Also, check the tableName which is the key in the map to ensure we are only extracting the relevant metrics.

phoenix-core/src/it/java/org/apache/phoenix/monitoring/CountRowsScannedIT.java
```diff
@@ -0,0 +1,299 @@
...
+        Map<String, String> props = Maps.newHashMapWithExpectedSize(1);
+        setUpTestDriver(new ReadOnlyProps(props));
+    }
+
+    @Test
+    public void testSinglePrimaryKey() throws Exception {
+        Properties props = new Properties();
+        props.put(QueryServices.COLLECT_REQUEST_LEVEL_METRICS, "true");
+        // force many rpc calls
+        props.put(QueryServices.SCAN_CACHE_SIZE_ATTRIB, "10");
```
tkhurana, line 56: I see this code being repeated in multiple tests. We could move it to the setup method.

phoenix-core/src/it/java/org/apache/phoenix/monitoring/CountRowsScannedIT.java
```diff
@@ -0,0 +1,299 @@
...
+        Connection conn = DriverManager.getConnection(getUrl(), props);
+        String tableName = generateUniqueName();
+        PhoenixStatement stmt = conn.createStatement().unwrap(PhoenixStatement.class);
+        stmt.execute("CREATE TABLE " + tableName
```
tkhurana, line 60: Can you also add a testcase with skip scan. Like `WHERE A IN (20, 45, 68, 3)`

tkhurana:
@palashc Another test I want you to write is where there are delete markers in the beginning of the table and then you run a query like this `SELECT * FROM <TABLE> LIMIT 1`. It should skip the delete markers till it finds the first valid row.  You could run this query in a loop where in every iteration you could delete more rows which have to be skipped in the subsequent SELECT query.

phoenix-core-server/src/main/java/org/apache/phoenix/cache/aggcache/SpillableGroupByCache.java
```diff
@@ -365,6 +366,10 @@ public void close() throws IOException {
                 }
             }
 
+            public boolean next(List<Cell> result, ScannerContext scannerContext) throws IOException {
+                return next(result);
```
tkhurana, line 370: Why is the scannerContext being silently ignored here ?

phoenix-core-client/pom.xml
```diff
@@ -438,5 +438,9 @@
       <groupId>org.hdrhistogram</groupId>
       <artifactId>HdrHistogram</artifactId>
     </dependency>
+    <dependency>
```
tkhurana, line 441: Is it OK to add this dependency ? @stoty can you confirm ?

phoenix-core/src/it/java/org/apache/phoenix/monitoring/CountRowsScannedIT.java
palashc: done

phoenix-core/src/it/java/org/apache/phoenix/monitoring/CountRowsScannedIT.java
palashc: @tkhurana done

phoenix-core/src/it/java/org/apache/phoenix/monitoring/CountRowsScannedIT.java
```diff
@@ -0,0 +1,312 @@
...
+        assertEquals(79, count5);
+
+        // pk1 and pk2, group by, ordered
+        long count6 = countRowsScannedFromSql(stmt, "SELECT B, SUM(A), SUM(Z) FROM " + tableName
+                + " WHERE A >= 2 AND B >= 3 GROUP BY B ORDER BY B DESC");
+        assertEquals(79, count6);
+    }
+
+    @Test
+    public void testQueryWithDeleteMarkers() throws Exception {
```
palashc, line 177: @tkhurana Added this test for deletes, let me know if this makes sense.

phoenix-core/src/it/java/org/apache/phoenix/monitoring/CountRowsScannedIT.java
palashc: done

phoenix-core-client/pom.xml
virajjasani: Having phoenix-core-client depend on hbase-server is really bad practice because it ultimately kills the whole purpose of having two separate phoenix-core modules, one for client and one of server.

But the bigger problem here is that we use the same iterator class at both client and server side, and depending on which side uses it, the context like this can be used.

It would have been much simpler if we had iterators of type `ResultIterator` only being used by clients and scanners of type `BaseRegionScanner` only being used by server. Then we could make `ScannerContext` available to only scanners of type `BaseRegionScanner`.

phoenix-core-server/src/main/java/org/apache/phoenix/iterate/NonAggregateRegionScannerFactory.java
```diff
@@ -465,6 +475,10 @@ public boolean next(List<Cell> results) throws IOException {
                         }
                     }
                     tuple = nextTuple;
+                    if (regionScannerContext != null) {
+                        ScannerContextUtil.updateMetrics(regionScannerContext, scannerContext);
+                        regionScannerContext = null;
+                    }
```
virajjasani, line 481: I wonder if it is still possible to not pass the ScannerContext to iterators above and still use it here?
Probably we can, since we are not really updating any context in `OrderedResultIterator` and `OffsetResultIterator`? Or i am missing something?

phoenix-core-client/pom.xml
virajjasani: Now i think probably we might still be able to get around this by not passing scanner context to `OrderedResultIterator` and `OffsetResultIterator`?

phoenix-core-server/src/main/java/org/apache/phoenix/iterate/NonAggregateRegionScannerFactory.java
palashc: @virajjasani how would we pass the context from RegionScannerResultIterator to NonAggregateRegionScannerFactory?

phoenix-core-client/src/main/java/org/apache/phoenix/compile/QueryCompiler.java
palashc: @tkhurana This is to keep track of the Context for nested/inner Select queries. During `getReadMetrics` in `PhoenixResultSet`, we combine the metrics from the inner contexts with the outer context.

phoenix-core-client/src/main/java/org/apache/phoenix/compile/QueryCompiler.java
palashc: @tkhurana The existing overloaded constructors did not change, they just provide a null value for the new field `parentContext`.

tkhurana: @palashc But one constructor API which is public is  changing right ?

palashc: @tkhurana Only the one which sets all the fields and only has usage in the class itself.

tkhurana: We have seen in the past where other applications are using the public API. Simply searching in the Phoenix code space is not enough. The rule of thumb is don't change the signature of public API.

phoenix-core-server/src/main/java/org/apache/phoenix/cache/aggcache/SpillableGroupByCache.java
palashc: @tkhurana I am not sure what the implementation should be for scanners returned by SpillableGroupByCache and InMemoryGroupByCache

phoenix-core-client/src/main/java/org/apache/phoenix/compile/QueryCompiler.java
palashc: Got it, understood. I have made the fix to not change any of the 5 existing constructors.

phoenix-core-server/src/main/java/org/apache/phoenix/iterate/NonAggregateRegionScannerFactory.java
```diff
@@ -298,6 +303,24 @@ static OrderedResultIterator deserializeFromScan(Scan scan, RegionScanner s,
         }
     }
 
+    private static class IteratorAndContext {
```
virajjasani, line 306: @palashc you think this is what seems tricky? Let me take a look next week in detail.

palashc: @virajjasani Yes, exactly. Maybe there is another way to refactor here.

phoenix-core-server/src/main/java/org/apache/phoenix/iterate/NonAggregateRegionScannerFactory.java
palashc: Tried to address in latest commit

phoenix-core-server/src/main/java/org/apache/phoenix/coprocessor/BaseRegionScanner.java
```diff
@@ -42,9 +42,7 @@ public boolean isFilterDone() {
     public abstract boolean next(List<Cell> results) throws IOException;
 
     @Override
-    public boolean next(List<Cell> result, ScannerContext scannerContext) throws IOException {
```
stoty, line 45: Why is that ?
I can see this implemented in child classes.

phoenix-core-client/pom.xml
stoty: Can you point me where this is a problem in the code ?
There are too many change files for me to check easily.

palashc: @stoty initial implementation had introduced a dependency on hbase-server in phoenix-client since I was passing ScannerContext through `OrderedResultIterator` and `OffsetResultIterator`, which was not acceptable. I have now tried an alternate implementation where I avoid doing this.

phoenix-core-server/src/main/java/org/apache/phoenix/coprocessor/BaseRegionScanner.java
palashc: Declared next with scannerContext as abstract to make sure all child classes implement it.

phoenix-core-server/src/main/java/org/apache/hadoop/hbase/regionserver/ScannerContextUtil.java
```diff
@@ -36,4 +36,28 @@ public static void incrementSizeProgress(ScannerContext sc, List<Cell> cells) {
...
+        sc.updateTimeProgress();
+    }
+
+    /**
+     * Increment the heapSize progress to a large value for server paging.
+     * Not using Long.MAX_VALUE since that can cause overflow.
+     */
+    public static void incrementHeapProgressForPaging(ScannerContext sc) {
+        sc.incrementSizeProgress(0, Long.MAX_VALUE/2);
+    }
```
virajjasani, line 62: @kadirozde @palashc This is nice trick. Is `Long.MAX_VALUE/2` guaranteed to cause large enough size for HBase to return response?

kadirozde: Yes, 4096 PB will be beyond any practical heap size for a very long time.

kadirozde: However, you have a good point. @palashc, the default heap size is -1, which means it is not enforced. What happens in that case?

kadirozde: @palashc, we may have to set the heap limit to Long.MAX_VALUE/2 on the scanner context on the way down and increment heap progress by Long.MAX_VALUE/2 when dummy result is received. Also we need to make sure that this manipulation of the scanner context is done only once for a given scanner.

virajjasani: @palashc it seems we have better alternative i.e. instead of size limit, we can enforce time limit reached by using

```
  // Set this to true will have the same behavior with reaching the time limit.
  // This is used when you want to make the current RSRpcService.scan returns immediately. For
  // example, when we want to switch from pread to stream, we can only do it after the rpc call is
  // returned.
  private boolean returnImmediately;
```

```
  void returnImmediately() {
    this.returnImmediately = true;
  }
```

This enforces `timeLimitReached`. HBase uses three factors to determine whether to return immediately:
1. sizeLimitReached
2. timeLimitReached
3. resultsLimitReached

kadirozde: @palashc, the above by @virajjasani is a much simpler and better approach. Please verify that it works.

palashc: Made the change in the latest commit after verifying using the new IT I added in ServerPagingIT. 
Thanks @virajjasani , great find!

palashc: @kadirozde @virajjasani Please let me know if you would be able to review the rest of the changes in this PR regarding scannerContext.

phoenix-core-server/src/main/java/org/apache/phoenix/coprocessor/DelegateRegionScanner.java
```diff
@@ -97,4 +152,10 @@ public RegionScanner getNewRegionScanner(Scan scan) throws IOException {
             throw new DoNotRetryIOException(e);
         }
     }
+
+    public static class ScannerContextNextNotImplementedException extends IOException {
```
kadirozde, line 156: Why is this necessary?

palashc: It can be used to inform DelegateRegionScanner that a lower level does not implement next/nextRaw with scannerContext, for example in this change in HashJoinRegionScanner: https://github.com/apache/phoenix/pull/1936/files#diff-77ea25574d36a1657d82cedf3384ebca3d8b11ac7a0de74dd5bcb18c59069afcR352

kadirozde: Why does HashJoinRegionScanner not support it? Should all layers support this now without exception?

palashc: `HashJoinRegionScanner.nextRaw` supports it. Should `next()` also support it?

kadirozde: Yes, all should support it from now on

palashc: What should be the implementation for next in HashJoinRegionScanner, with and without ScannerContext? Should it just call the corresponding nextRaw?

palashc: I made the change in the latest commit.

palashc:
Latest test report: https://ci-hadoop.apache.org/job/Phoenix/job/Phoenix-PreCommit-GitHub-PR/job/PR-1936/29/testReport/

phoenix-core-server/src/main/java/org/apache/phoenix/iterate/NonAggregateRegionScannerFactory.java
kadirozde: I did not understand what we are achieving here. OrderedResultIterator is used in both client and server side. So, we need to add the next() with ScannerContext method to OrderedResultIterator too. This method will be used on the server side. The scanner context should not serialized/deserialized. We also need to pass through ScannerContext context for the other result iterators used on the server side (for example, OffsetResultIterator). Also let us add tests for these result iterators.

phoenix-core/src/it/java/org/apache/phoenix/monitoring/CountRowsScannedIT.java
```diff
@@ -0,0 +1,311 @@
...
+import org.apache.phoenix.query.QueryServices;
+import org.apache.phoenix.thirdparty.com.google.common.collect.Maps;
+import org.apache.phoenix.util.PhoenixRuntime;
+import org.apache.phoenix.util.ReadOnlyProps;
+import org.junit.BeforeClass;
+import org.junit.Test;
+import org.junit.experimental.categories.Category;
+
+@Category(NeedsOwnMiniClusterTest.class)
+public class CountRowsScannedIT extends BaseTest {
```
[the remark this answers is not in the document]
palashc: @tkhurana added a test which scans the index

phoenix-core-server/src/main/java/org/apache/phoenix/iterate/NonAggregateRegionScannerFactory.java
palashc: As discussed offline, the changes in this class help with linking the RegionScanner from NonAggregateRegionScannerFactory to the RegionScannerResultIterator, thus bypassing the ResultIterators in between since the current codebase structure makes it difficult to pass the ScannerContext through them on the server side. 
I have also renamed the inner class used here to more meaningful `OrderedResultIteratorWithScannerContext`

kadirozde (approved):
+1, Thanks!

PR #1937, opened 2024-07-12
PHOENIX-7357 New variable length binary data type: VARBINARY_ENCODED

virajjasani:
Jira: PHOENIX-7357

Design Doc: https://docs.google.com/document/d/1-eZ4cbLVEPC4z7Iy1U_VJgNaKC9odkRVzcVPV3ZqOcI/edit?usp=sharing

As of today, Phoenix provides several variable length as well as fixed length data types. One of the variable length data types is VARBINARY. It is variable length binary blob. Using VARBINARY as only primary key can be considered as if using HBase row key.

HBase provides a single row key. Any client application that requires using more than one column for primary keys, using HBase requires special handling of storing both column values as a single binary row key. Phoenix provides the ability to use more than one primary key by providing composite primary keys. Composite primary key can contain any number of primary key columns. Phoenix also provides the ability to add new nullable primary key columns to the existing composite primary keys. Phoenix uses HBase as its backing store. In order to provide the ability for users to define multiple primary keys, Phoenix internally concatenates binary encoded values of each primary key column value and uses concatenated binary value as HBase row key. In order to efficiently concatenate as well as retrieve individual primary key values, Phoenix implements two ways:

For fixed length columns: The length of the given column is determined by the maximum length of the column. As part of the read flow, while iterating through the row key, fixed length numbers of bytes are retrieved while reading. While writing, if the original encoded value of the given column has less number of bytes, additional null bytes (\x00) are padded until the fixed length is filled up. Hence, for smaller values, we end up wasting some space.
For variable length columns: Since we cannot know the length of the value of variable length data type in advance, a separator or terminator byte is used. Phoenix uses null byte as separator (\x00) byte. As of today, VARCHAR is the most commonly used variable length data type and since VARCHAR represents String, null byte is not part of valid String characters. Hence, it can be effectively used to determine when to terminate the given VARCHAR value.
 

The null byte (\x00) works fine as a separator for VARCHAR. However, it cannot be used as a separator byte for VARBINARY because VARBINARY can contain any binary blob values. Due to this, Phoenix has restrictions for VARBINARY type: 

 

It can only be used as the last part of the composite primary key.
It cannot be used as a DESC order primary key column.
 

Using VARBINARY data type as an earlier portion of the composite primary key is a valid use case. One can also use multiple VARBINARY primary key columns. After all, Phoenix provides the ability to use multiple primary key columns for users.

Besides, using secondary index on data table means that the composite primary key of secondary index table includes: 

```
<secondary-index-col1> <secondary-index-col2> … <secondary-index-colN> <primary-key-col1> <primary-key-col2> … <primary-key-colN>
```

As primary key columns are appended to the secondary indexes columns, one cannot create a secondary index on any VARBINARY column.

The proposal of this Jira is to introduce new data type VARBINARY_ENCODED, which has no restriction of being considered as composite primary key prefix or using it as DESC ordered column.

This means, we need to effectively distinguish where the variable length binary data terminates in the absence of fixed length information.


phoenix-core-client/src/main/java/org/apache/phoenix/expression/CoerceExpression.java
```diff
@@ -145,11 +147,30 @@ public void write(DataOutput output) throws IOException {
 
     @Override
     public boolean evaluate(Tuple tuple, ImmutableBytesWritable ptr) {
-        if (getChild().evaluate(tuple, ptr)) {
-            getDataType().coerceBytes(ptr, null, getChild().getDataType(),
-                    getChild().getMaxLength(), null, getChild().getSortOrder(), 
-                    maxLength, null, getSortOrder(), rowKeyOrderOptimizable);
-            return true;
+        if (getChild() instanceof LiteralExpression
+            && getChild().getDataType() == PVarbinary.INSTANCE
```
ranganathg, line 151: what is this check fo?

phoenix-core-client/src/main/java/org/apache/phoenix/expression/RowValueConstructorExpression.java
```diff
@@ -241,12 +254,16 @@ public boolean evaluate(Tuple tuple, ImmutableBytesWritable ptr) {
...
-                            k >=0 &&  getChildren().get(k).getDataType() != null 
-                                  && !getChildren().get(k).getDataType().isFixedWidth()
-                                  && outputBytes[outputSize-1] == SchemaUtil.getSeparatorByte(true, false, getChildren().get(k))
-                                  &&  (getChildren().get(k).getSortOrder() == SortOrder.ASC ? true : isStripTrailingSepByte()) ; k--) {
+                    for (int k = expressionCount - 1; k >= 0
+                        && getChildren().get(k).getDataType() != null
+                        && !getChildren().get(k).getDataType().isFixedWidth()
+                        && hasSeparatorBytes(outputBytes, outputSize, k)
+                        && (getChildren().get(k).getSortOrder() == SortOrder.ASC
+                        || isStripTrailingSepByte()); k--) {
```
ranganathg, line 262: large condition in for block, can this be extracted to a method?

phoenix-core-client/src/main/java/org/apache/phoenix/expression/RowValueConstructorExpression.java
```diff
@@ -259,7 +276,23 @@ public boolean evaluate(Tuple tuple, ImmutableBytesWritable ptr) {
...
+                getChildren().get(k));
+        } else {
+            byte[] sepBytes = SchemaUtil.getSeparatorBytesForVarBinaryEncoded(true, false,
+                getChildren().get(k).getSortOrder());
+            return outputSize >= 2 && outputBytes[outputSize - 1] == sepBytes[1]
+                && outputBytes[outputSize - 2] == sepBytes[0];
+        }
+    }
+
+    private static int getSeparatorBytesLen(Expression expression) {
```
ranganathg, line 292: you can keep it getSeparatorBytesLength?

phoenix-core/src/it/java/org/apache/phoenix/end2end/VarBinary1IT.java
```diff
@@ -0,0 +1,2205 @@
...
+
+import org.apache.hadoop.hbase.util.Bytes;
+import org.apache.phoenix.schema.PTableImpl;
+import org.apache.phoenix.util.PropertiesUtil;
+
+import static org.apache.phoenix.util.TestUtil.TEST_PROPERTIES;
+
+@Category(ParallelStatsDisabledTest.class)
+@RunWith(Parameterized.class)
+public class VarBinary1IT extends ParallelStatsDisabledIT {
```
ranganathg, line 43: VarBinaryEncodedIT?

phoenix-core-client/src/main/java/org/apache/phoenix/expression/CoerceExpression.java
virajjasani: I will write more comments, filling up design doc section on this as well.

phoenix-core-client/src/main/java/org/apache/phoenix/expression/RowValueConstructorExpression.java
virajjasani: Sure, will do

phoenix-core/src/it/java/org/apache/phoenix/end2end/VarBinary1IT.java
virajjasani: We don't even have much coverage for VARBINARY (obviously because it is not used as tenant id, it is not used as any but last part of composite pk), so maybe that's why i kept it VarBinary but yes let me change this.

phoenix-core-client/src/main/java/org/apache/phoenix/filter/SkipScanFilter.java
```diff
@@ -576,13 +577,21 @@ private void setStartKey() {
...
-                startKey[startKeyLength++] = SchemaUtil.getSeparatorByte(schema.rowKeyOrderOptimizable(), true, field);
+                if (field.getDataType() != PVarbinaryEncoded.INSTANCE) {
+                    startKey[startKeyLength++] =
+                        SchemaUtil.getSeparatorByte(schema.rowKeyOrderOptimizable(), true, field);
+                } else {
+                    byte[] sepBytes = SchemaUtil.getSeparatorBytesForVarBinaryEncoded(
+                        schema.rowKeyOrderOptimizable(), true, field.getSortOrder());
+                    startKey[startKeyLength++] = sepBytes[0];
+                    startKey[startKeyLength++] = sepBytes[1];
+                }
```
tkhurana, line 594: Could we introduce an API which always returns a byte array and we simply iterate over the array. That should work regardless whether there is one separator byte or multiple and reduce the amount of code duplication.

phoenix-core-client/src/main/java/org/apache/phoenix/expression/RowValueConstructorExpression.java
```diff
@@ -220,14 +223,24 @@ public boolean evaluate(Tuple tuple, ImmutableBytesWritable ptr) {
...
                             output.write(tempPtr.get(), tempPtr.getOffset(), tempPtr.getLength());
                             if (!childType.isFixedWidth()) {
-                                output.write(SchemaUtil.getSeparatorByte(true, false, child));
+                                if (childType != PVarbinaryEncoded.INSTANCE) {
+                                    output.write(SchemaUtil.getSeparatorByte(true, false, child));
+                                } else {
+                                    output.write(
+                                        SchemaUtil.getSeparatorBytesForVarBinaryEncoded(true, false,
+                                            child.getSortOrder()));
+                                }
```
tkhurana, line 243: Could we introduce an API which always returns a byte array and we simply iterate over the array. That should work regardless whether there is one separator byte or multiple and reduce the amount of code duplication.

tkhurana:
Could we introduce an API which always returns a byte array and we simply iterate over the array. That should work regardless whether there is one separator byte or multiple and reduce the amount of code duplication.

virajjasani:
> Could we introduce an API which always returns a byte array and we simply iterate over the array. That should work regardless whether there is one separator byte or multiple and reduce the amount of code duplication.

This is done

phoenix-core-client/src/main/java/org/apache/phoenix/schema/RowKeySchema.java
```diff
@@ -232,11 +262,24 @@ public Boolean previous(ImmutableBytesWritable ptr, int position, int minOffset)
...
+                        offset--;
+                    } while (offset > minOffset && buf[offset] != sepByte);
+                }
+            } else {
+                if (offset > minOffset && SchemaUtil.areSeparatorBytesForVarBinaryEncoded(buf,
+                    offset, SortOrder.ASC)) {
+                    boolean ascSort =
+                        !rowKeyOrderOptimizable || field.getSortOrder() == SortOrder.ASC;
+                    do {
+                        offset++;
```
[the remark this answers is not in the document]
virajjasani: Excellent catch, previous() is not being tested by the tests

phoenix-core-client/src/main/java/org/apache/phoenix/schema/RowKeySchema.java
```diff
@@ -159,30 +160,59 @@ private Boolean next(ImmutableBytesWritable ptr, int position, int maxOffset, bo
         // If positioned at SEPARATOR_BYTE, skip it.
         // Don't look back at previous fields if this is our first next call, as
         // we may have a partial key for RVCs that doesn't include the leading field.
-        if (position > 0 && !isFirst && !getField(position-1).getDataType().isFixedWidth()) {
-            ptr.set(ptr.get(), ptr.getOffset()+ptr.getLength()+1, 0);
+        if (position > 0 && !isFirst && !getField(position - 1).getDataType().isFixedWidth()) {
+            if (getField(position - 1).getDataType() != PVarbinaryEncoded.INSTANCE) {
```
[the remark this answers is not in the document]
virajjasani: Done

phoenix-core-client/src/main/java/org/apache/phoenix/schema/RowKeyValueAccessor.java
```diff
@@ -37,14 +47,58 @@
...
+
+    /**
+     * The class used to keep list of booleans representing whether the order of the given
+     * field is Ascending. For example, let's say we have rowkey with pk columns
+     * (A ASC, B DESC, C ASC). The corresponding object of this class will have sortOrderAsc list
+     * with values (true, false, true). The array of this class is serialized and deserialized
+     * together with other fields of RowKeyValueAccessor. The value of the list determines which
+     * separator bytes need to be used while iterating through the rowkey bytes.
+     */
+    static class SortOrderLists {
```
kadirozde, line 90: This should read SortOrderList.

phoenix-core-client/src/main/java/org/apache/phoenix/schema/RowKeyValueAccessor.java
```diff
@@ -37,14 +47,58 @@
...
+    /**
+     * The class used to keep list of booleans representing whether the data type of the given
+     * field is of type VARBINARY_ENCODED. For example, let's say we have variable length
+     * data types in the order (VARCHAR, VARBINARY_ENCODED, VARCHAR). The corresponding object
+     * of this class will have binaryEncodedDataTypes list with values (false, true, false).
+     * The array of this class is serialized and deserialized together with other fields of
+     * RowKeyValueAccessor. The value of the list determines which separator bytes need to be
+     * used while iterating through the rowkey bytes.
+     */
+    static class BinaryEncodedTypesLists {
```
kadirozde, line 70: BinaryEncodedTypesLists does not tell what this is about.  It gets confusing when you declare BinaryEncodedTypesLists[] binaryEncodedTypesLists later.  I think ListOfBinaryEncodedTypeFlags or something similar would be more readable. Also did you consider using BitSet? If you use it, you do not need these two classes.

kadirozde:
Can you add tests for both varbinary and varbinary encoded to WhereCompilerTest# testWhereInclusion and PartialIndexIT?

virajjasani:
> Can you add tests for both varbinary and varbinary encoded to WhereCompilerTest# testWhereInclusion and PartialIndexIT?

Thanks @kadirozde! Added the tests in the latest revision, and addressed previous review comments.

kadirozde (approved):
+1, Thank you!

PR #1939, opened 2024-07-15
PHOENIX-7359 BackwardCompatibilityIT throws NPE with Hbase 2.6 profile

stoty:


virajjasani (approved):
+1 for the change, though i think `5.1.3` might work for `2.6` profile also

stoty:
No, we have never backported 2.6 support to 5.1, much less into 5.1.3.

I think that backporting 2.6 support to 5.1 would be a goodea BTW.

virajjasani:
Ah right, my bad

stoty:
merged manually.

PR #1943, opened 2024-07-23
PHOENIX-7363 Protect server side metadata cache updates for the given PTable

virajjasani:
Jira: PHOENIX-7363

**All meta handlers on SYSTEM.CATALOG regionserver exhausted within CQSI#init**

After allowing readLock for getTable API at the server side ([PHOENIX-6066](https://issues.apache.org/jira/browse/PHOENIX-6066)), it seems that under heavy load, all meta handler threads can get exhausted within ConnectionQueryServicesImpl initialization as part of any of the MetaDataEndpointImpl coproc operations. When the table details are not present in the cache, MetaDataEndpointImpl coproc can attempt to create new connection on the server side in order to scan SYSTEM.CATALOG. Under heavy load, several (all of) meta handlers – which are dedicated for all metadata (system table) operations – could attempt to create server side connection, which can further lead into creating new PhoenixConnection to execute CREATE TABLE DDL for SYSTEM.CATALOG in order to ensure that the System tables exist.

Only one thread should update the metadata cache for a given table at a given time.

### DIFF
--- a/phoenix-core-server/src/main/java/org/apache/phoenix/coprocessor/MetaDataEndpointImpl.java
+++ b/phoenix-core-server/src/main/java/org/apache/phoenix/coprocessor/MetaDataEndpointImpl.java
@@ -3911,7 +3911,7 @@ TABLE_FAMILY_BYTES, TABLE_SEQ_NUM_BYTES);
          * Lock directly on key, though it may be an index table. This will just prevent a table
          * from getting rebuilt too often.
          */
-        boolean wasLocked = (rowLock != null);
+        final boolean wasLocked = (rowLock != null);
         try {
             if (!wasLocked) {
                 rowLock = acquireLock(region, key, null, true);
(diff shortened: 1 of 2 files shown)


phoenix-core-server/src/main/java/org/apache/phoenix/coprocessor/MetaDataEndpointImpl.java
```diff
@@ -3920,24 +3930,59 @@ private PTable doGetTable(byte[] tenantId, byte[] schemaName, byte[] tableName,
                 }
                 return table;
             }
-            // Query for the latest table first, since it's not cached
-            table =
-                    buildTable(key, cacheKey, region, HConstants.LATEST_TIMESTAMP, clientVersion);
-            if ((table != null && table.getTimeStamp() <= clientTimeStamp) ||
-                    (blockWriteRebuildIndex && table.getIndexDisableTimestamp() > 0)) {
-                return table;
+            if (!wasLocked && rowLock != null) {
```
kadirozde, line 3933: Since we are using two different locking mechanisms, we should not release the original read lock here. Otherwise, we attempt to get the table from the cache without a lock essentially and this will race with another thread that takes a write lock and make a change on the object. We do not want this to happen.

virajjasani: Makes sense, this was only needed if we were to use HBase row level write-lock.

kadirozde (approved):
+1, Thanks!

PR #1944, opened 2024-07-24
PHOENIX-7369 Avoid redundant recursive getTable() RPC calls

virajjasani:
Jira: PHOENIX-7369


virajjasani:
@stoty i see that all builds are going to https://ci-hadoop.apache.org/computer/hadoop1/ and the host does not seem healthy as all builds get aborted. What can we do? Can we disconnect the host using jenkins ui? Or shall i inform ASF Infra reg this by creating INFRA Jira?

kadirozde (approved):

stoty:
Yes, the pre-yetus scripts did list and exclude single CI hosts explictily, so that is possible (or at least it was possible at the time).

However, I think that asking for help from INFRA is the way to go, and we should only resort to excluding hosts if infra cannot help with the problem.

virajjasani:
created Jira [INFRA-25995](https://issues.apache.org/jira/browse/INFRA-25995)

virajjasani:
Marked Hadoop1 node offline as per recommendation on ASF slack channel asfinfra, now the build is progressing: https://ci-hadoop.apache.org/job/Phoenix/job/Phoenix-PreCommit-GitHub-PR/job/PR-1944/

PR #1950, opened 2024-08-06
PHOENIX-7375 CQSI connection init from regionserver hosting SYSTEM.CATALOG does not require RPC calls to system tables

virajjasani:
Jira: PHOENIX-7375

In order to execute any query at the server side, Phoenix client requires creating CQSI (ConnectionQueryServicesImpl) connection, which internally initiates and maintains long lasting connection against HBase server. CQSI connection is unique per JDBC url used to initiate the connection. Once created, CQSI connections are cached per 24 hr (by default) for every unique JDBC url provided.

When client initiates CQSI connection, the connection initialization also attempts to execute some metadata queries to ensure that the system tables like SYSTEM.CATALOG exist and the client version is compatible against the server version. For this, CQSI#init makes RPC calls against MetaDataEndpointImpl coproc.

This operation is valid for every CQSI connection initiated for every unique JDBC url by every client. However, when server hosting SYSTEM.CATALOG initiates CQSI connection, it means that SYSTEM.CATALOG and other system tables already exist. Moreover, client/server version compatibility check is not required because the connection is being created from the same server that is hosting SYSTEM.CATALOG.

Metadata operations performed by the regionserver hosting SYSTEM.CATALOG region(s) also hold row level write lock for the given PTable entry. Hence, this improvement is also expected to bring some perf improvement for the metadata operations.

### DIFF
--- a/phoenix-core-server/src/main/java/org/apache/phoenix/coprocessor/MetaDataEndpointImpl.java
+++ b/phoenix-core-server/src/main/java/org/apache/phoenix/coprocessor/MetaDataEndpointImpl.java
@@ -104,6 +104,7 @@ import static org.apache.phoenix.util.ViewUtil.getSystemTableForChildLinks;
 import java.io.IOException;
 import java.nio.charset.StandardCharsets;
 import java.security.PrivilegedExceptionAction;
+import java.sql.Connection;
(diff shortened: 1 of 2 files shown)


kadirozde (approved):
+1, Thanks

PR #1951, opened 2024-08-11
PHOENIX-7379 Improve handling of concurrent index mutations with the …

kadirozde:
…same timestamp

### DIFF
--- a/phoenix-core-server/src/main/java/org/apache/phoenix/hbase/index/IndexRegionObserver.java
+++ b/phoenix-core-server/src/main/java/org/apache/phoenix/hbase/index/IndexRegionObserver.java
@@ -58,7 +58,6 @@ import org.apache.hadoop.hbase.CellUtil;
 import org.apache.hadoop.hbase.CoprocessorEnvironment;
 import org.apache.hadoop.hbase.DoNotRetryIOException;
 import org.apache.hadoop.hbase.HConstants;
-import org.apache.hadoop.hbase.HConstants.OperationStatusCode;
 import org.apache.hadoop.hbase.TableName;
 import org.apache.hadoop.hbase.client.Delete;
 import org.apache.hadoop.hbase.client.Durability;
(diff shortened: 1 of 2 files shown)


kadirozde:
This PR drastically reduces the number of times a thread sleeps (more than 6x) and visibly reduces total upsert time in ConcurrentMutationsExtendedIT#testConcurrentUpserts().

phoenix-core-server/src/main/java/org/apache/phoenix/hbase/index/IndexRegionObserver.java
```diff
@@ -366,7 +365,8 @@ public int getMaxPendingRowCount() {
   private boolean shouldWALAppend = DEFAULT_PHOENIX_APPEND_METADATA_TO_WAL;
   private boolean isNamespaceEnabled = false;
   private boolean useBloomFilter = false;
-
+  private long lastTimestamp = 0;
```
stoty, line 368: Longs are not atomic in Java.
Shouldn't this be atomic / volatile or be accessed in synchronized blocks everywhere ?

phoenix-core/src/it/java/org/apache/phoenix/end2end/ConcurrentMutationsExtendedIT.java
```diff
@@ -357,6 +358,8 @@ public void testConcurrentUpserts() throws Exception {
         }
 
         assertTrue("Ran out of time", doneSignal.await(120, TimeUnit.SECONDS));
+        System.out.println("Total upsert time in ms : "
```
stoty, line 361: use LOGGER.info() instead ?

phoenix-core-server/src/main/java/org/apache/phoenix/hbase/index/IndexRegionObserver.java
kadirozde: It is accessed only in synchronized block and should be accessed only in synchronized block. So there is no need to define it as volatile or atomic, which will give the wrong impression.

stoty: Thanks, you are right.

stoty (approved):
+1 LGTM

PR #1952, opened 2024-08-13
PHOENIX-7381 : Skip last_ddl_timestamp validation when UCF is defined on table and cache entry is not old

palashc:


phoenix-core/src/it/java/org/apache/phoenix/cache/ServerMetadataCacheIT.java
```diff
@@ -1812,6 +1812,47 @@ public void testCacheUpdatedBeforeDDLOperations() throws Exception {
...
+                upsert(conn1, tableName, true);
+            }
+
+            MetricsMetadataCachingSource.MetadataCachingMetricValues newMetricValues
+                    = metricsSource.getCurrentMetricValues();
+
+            assertEquals("There should have been no timestamp validation requests.",
+                    0,
+                    newMetricValues.getValidateDDLTimestampRequestsCount()
+                            - oldMetricValues.getValidateDDLTimestampRequestsCount());
```
virajjasani, line 1853: `newMetricValues.getValidateDDLTimestampRequestsCount()` should also be 0?

palashc: There will be requests from other tests, that's why I am checking the difference before and after just this test.

palashc:
@virajjasani Does this change look good? If yes, we can get it in.

phoenix-core-client/src/main/java/org/apache/phoenix/schema/MetaDataClient.java
```diff
@@ -810,12 +810,10 @@ private boolean avoidRpcToGetTable(boolean alwaysHitServer, Long resolvedTimesta
...
             //QueryOptimizer needs that to decide whether the index can be used
-            if (overrideUcfToDefault) {
+            if (overrideUcfToAlways) {
                 effectiveUpdateCacheFreq =
-                    (Long) ConnectionProperty.UPDATE_CACHE_FREQUENCY.getValue(
-                        connection.getQueryServices().getProps()
-                            .get(QueryServices.DEFAULT_UPDATE_CACHE_FREQUENCY_ATRRIB));
-                ucfInfoForLogging = "connection-level-default";
+                    (Long) ConnectionProperty.UPDATE_CACHE_FREQUENCY.getValue("ALWAYS");
+                ucfInfoForLogging = "index-level-override";
```
virajjasani, line 816: nit: for the logging purpose, let's use `override-to-always` or something similar

phoenix-core/src/it/java/org/apache/phoenix/end2end/UCFWithDisabledIndexWithDDLValidationIT.java
```diff
@@ -40,7 +40,7 @@ private static void initCluster() throws Exception {
         Map<String, String> props = Maps.newConcurrentMap();
         props.put(BaseScannerRegionObserverConstants.PHOENIX_MAX_LOOKBACK_AGE_CONF_KEY,
                 Integer.toString(60 * 60 * 1000));
-        props.put(QueryServices.DEFAULT_UPDATE_CACHE_FREQUENCY_ATRRIB, "ALWAYS");
+        props.put(QueryServices.DEFAULT_UPDATE_CACHE_FREQUENCY_ATRRIB, "NEVER");
```
virajjasani, line 43: do we want to test with both `ALWAYS` and `NEVER`?

virajjasani (commented):
Few nits, +1 otherwise

phoenix-core/src/it/java/org/apache/phoenix/end2end/UCFWithDisabledIndexWithDDLValidationIT.java
palashc: @virajjasani With DDL timestamp validation, UCF should not be ALWAYS, so I think it is okay. WDYT?

phoenix-core-client/src/main/java/org/apache/phoenix/schema/MetaDataClient.java
palashc: done

virajjasani (approved):
+1, pending jenkins build results

PR #1955, opened 2024-08-16
PHOENIX-7386 Override UPDATE_CACHE_FREQUENCY if table has disabled indexes

virajjasani:
Jira: PHOENIX-7386

If table has UPDATE_CACHE_FREQUENCY set to non-default i.e. anything other than ALWAYS, depending on the value, PTable objects are cached until that duration at the client side metadata cache. One of the common cases for creating index is to create index in CREATE_DISABLE state.

When any index is in DISABLE, CREATE_DISABLE or PENDING_ACTIVE state, Phoenix client does not include PTable of the corresponding index as part of the IndexMaintainer objects to server. Since indexes are expected to be either dropped or in building state from the above disabled states, it is crucial for Phoenix client to override the non-default value of UPDATE_CACHE_FREQUENCY on the base table. This would help avoid any data integrity issue because until the index state becomes BUILDING or ACTIVE, the client will continue to override UPDATE_CACHE_FREQUENCY to default value to let the UPSERT and/or SELECT queries to initiate getTable() RPC and override the client side metadata cache value.

### DIFF
--- a/phoenix-core-client/src/main/java/org/apache/phoenix/schema/MetaDataClient.java
+++ b/phoenix-core-client/src/main/java/org/apache/phoenix/schema/MetaDataClient.java
@@ -785,8 +785,12 @@ public class MetaDataClient {
             final long effectiveUpdateCacheFreq;
             final String ucfInfoForLogging; // Only used for logging purposes
 
-            boolean overrideUcfToDefault =
-                PIndexState.PENDING_DISABLE.equals(table.getIndexState());
+            boolean overrideUcfToDefault = false;
+            if (table.getType() == INDEX) {
+                overrideUcfToDefault =
+                        PIndexState.PENDING_DISABLE.equals(table.getIndexState()) ||
+                                !IndexMaintainer.sendIndexMaintainer(table);
+            }
             if (!overrideUcfToDefault && !table.getIndexes().isEmpty()) {
                 List<PTable> indexes = table.getIndexes();
                 List<PTable> maintainedIndexes =
--- a/phoenix-core-client/src/main/java/org/apache/phoenix/schema/MetaDataClient.java
+++ b/phoenix-core-client/src/main/java/org/apache/phoenix/schema/MetaDataClient.java
@@ -785,19 +785,33 @@ public class MetaDataClient {
             final long effectiveUpdateCacheFreq;
             final String ucfInfoForLogging; // Only used for logging purposes
 
+            boolean overrideUcfToDefault =
+                PIndexState.PENDING_DISABLE.equals(table.getIndexState());
+            if (!overrideUcfToDefault && !table.getIndexes().isEmpty()) {
+                List<PTable> indexes = table.getIndexes();
+                List<PTable> maintainedIndexes =
+                        Lists.newArrayList(IndexMaintainer.maintainedIndexes(indexes.iterator()));
+                // The maintainedIndexes contain only the indexes that are used by clients
+                // while generating the mutations. If all the indexes are usable by clients,
+                // we don't need to override UPDATE_CACHE_FREQUENCY. However, if any index is
+                // not in usable state by the client mutations, we should override
+                // UPDATE_CACHE_FREQUENCY to default value so that we make getTable() RPC calls
+                // until all index states change to ACTIVE, BUILDING or other usable states.
+                overrideUcfToDefault = indexes.size() != maintainedIndexes.size();
+            }
+
             // What if the table is created with UPDATE_CACHE_FREQUENCY explicitly set to ALWAYS?
             // i.e. explicitly set to 0. We should ideally be checking for something like
             // hasUpdateCacheFrequency().
 
             //always fetch an Index in PENDING_DISABLE state to retrieve server timestamp
             //QueryOptimizer needs that to decide whether the index can be used
-            if (PIndexState.PENDING_DISABLE.equals(table.getIndexState())) {
+            if (overrideUcfToDefault) {
                 effectiveUpdateCacheFreq =
-                        (Long) ConnectionProperty.UPDATE_CACHE_FREQUENCY.getValue(
-                            connection.getQueryServices().getProps().get(
-                  QueryServices.UPDATE_CACHE_FREQUENCY_FOR_PENDING_DISABLED_INDEX,
-                  QueryServicesOptions.DEFAULT_UPDATE_CACHE_FREQUENCY_FOR_PENDING_DISABLED_INDEX));
-                ucfInfoForLogging = "pending-disable-index-level";
+                    (Long) ConnectionProperty.UPDATE_CACHE_FREQUENCY.getValue(
+                        connection.getQueryServices().getProps()
+                            .get(QueryServices.DEFAULT_UPDATE_CACHE_FREQUENCY_ATRRIB));
+                ucfInfoForLogging = "connection-level-default";
             } else if (table.getUpdateCacheFrequency()
                     != QueryServicesOptions.DEFAULT_UPDATE_CACHE_FREQUENCY) {
                 effectiveUpdateCacheFreq = table.getUpdateCacheFrequency();
--- a/phoenix-core/src/it/java/org/apache/phoenix/end2end/UpdateCacheFreqIT.java
+++ b/phoenix-core/src/it/java/org/apache/phoenix/end2end/UpdateCacheFreqIT.java
@@ -110,7 +110,7 @@ public class UpdateCacheFreqIT extends BaseTest {
             stmt.execute("CREATE TABLE " + tableName
                     + " (COL1 CHAR(10) NOT NULL, COL2 CHAR(5) NOT NULL, COL3 VARCHAR,"
                     + " COL4 VARCHAR CONSTRAINT pk PRIMARY KEY(COL1, COL2))"
-                    + " UPDATE_CACHE_FREQUENCY=10000");
+                    + " UPDATE_CACHE_FREQUENCY=20000");
             stmt.execute("CREATE INDEX " + indexName
                     + " ON " + tableName + " (COL3) INCLUDE (COL4)");
             stmt.execute("CREATE VIEW " + view01
--- a/phoenix-core/src/it/java/org/apache/phoenix/end2end/UpdateCacheFreqIT.java
+++ b/phoenix-core/src/it/java/org/apache/phoenix/end2end/UpdateCacheFreqIT.java
@@ -1,0 +1,230 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to you under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.phoenix.end2end;
+
+import com.google.protobuf.RpcCallback;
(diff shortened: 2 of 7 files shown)


phoenix-core/src/it/java/org/apache/phoenix/end2end/UpdateCacheFreqWithDDLValidationIT.java
```diff
@@ -0,0 +1,49 @@
...
+import org.apache.phoenix.coprocessorclient.BaseScannerRegionObserverConstants;
+import org.apache.phoenix.query.QueryServices;
+import org.apache.phoenix.thirdparty.com.google.common.collect.Maps;
+import org.apache.phoenix.util.ReadOnlyProps;
+import org.junit.BeforeClass;
+
+import java.util.Map;
+
+//@Category(NeedsOwnMiniClusterTest.class)
+// Uncomment as part of PHOENIX-7381
```
virajjasani, line 29: FYI @palashc

phoenix-core-client/src/main/java/org/apache/phoenix/schema/MetaDataClient.java
```diff
@@ -785,19 +785,33 @@ private boolean avoidRpcToGetTable(boolean alwaysHitServer, Long resolvedTimesta
             final long effectiveUpdateCacheFreq;
             final String ucfInfoForLogging; // Only used for logging purposes
 
+            boolean overrideUcfToDefault =
```
palashc, line 788: @virajjasani Should we override UCF if the ptable passed to this method is an index in any "unusable" state and not just PENDING_DISABLE?

phoenix-core/src/it/java/org/apache/phoenix/end2end/UpdateCacheFreqWithDDLValidationIT.java
palashc: When enabling last_ddl_timestamp validation, we would want to set default UCF to NEVER.

phoenix-core-client/src/main/java/org/apache/phoenix/schema/MetaDataClient.java
virajjasani: Sounds good, let me extend this support as well.

phoenix-core/src/it/java/org/apache/phoenix/end2end/UpdateCacheFreqWithDDLValidationIT.java
virajjasani: This is not ALWAYS, this is some non default positive value e.g. 20s or 10s.

virajjasani: My point is, when we uncomment this test as part of PHOENIX-7381, the tests will likely pass. Until PHOENIX-7381, these tests will not pass because even for 10/20s UCF, they still perform validate ddl timestamp and that makes getTable() call internally, hence the test fails as of now.

phoenix-core/src/it/java/org/apache/phoenix/end2end/UpdateCacheFreqIT.java
```diff
@@ -0,0 +1,230 @@
...
+import java.sql.Connection;
+import java.sql.DriverManager;
+import java.sql.ResultSet;
+import java.sql.Statement;
+import java.util.Map;
+
+import static org.apache.phoenix.query.QueryServices.DISABLE_VIEW_SUBTREE_VALIDATION;
+
+@Category(NeedsOwnMiniClusterTest.class)
+public class UpdateCacheFreqIT extends BaseTest {
```
palashc, line 56: @virajjasani We already have a test class `UpdateCacheIT`. Should we re-use that for these tests? If not, maybe we can have some different name for this test class?

virajjasani: The only reason why i could not use existing UpdateCacheIT is because we need to set these configs before the cluster comes up and we want to validate getTable() RPC calls by switching the coproc implementations between some queries. Achieving this with existing test did not seem feasible, hence the new class. And also, created `UpdateCacheFreqWithDDLValidationIT` to run these exact tests just by using metadata caching redesign configs.

But if you have any other test name in mind, i can use it. Maybe `UpdateCacheWithGetTableIT`?

phoenix-core/src/it/java/org/apache/phoenix/end2end/UpdateCacheFreqIT.java
```diff
@@ -0,0 +1,230 @@
...
+
+            TestUtil.removeCoprocessor(conn, "SYSTEM.CATALOG", MetaDataEndpointImpl.class);
+            TestUtil.addCoprocessor(conn, "SYSTEM.CATALOG", TestMetaDataEndpointImpl.class);
+
+            stmt.execute("UPSERT INTO " + tableName
+                    + " (col1, col2, col3, col4) values ('c011', "
+                    + "'c012', 'c013', 'c014')");
+            conn.commit();
+            throw new RuntimeException("Should not reach here");
+        } catch (PhoenixIOException e) {
```
palashc, line 219: @virajjasani Do you think we should also test that once a "disabled" index is in BUILDING or ACTIVE state on the server, client learns about it and then stops overriding the UCF value?

phoenix-core-client/src/main/java/org/apache/phoenix/schema/MetaDataClient.java
palashc: @virajjasani thanks, this looks good now!

phoenix-core/src/it/java/org/apache/phoenix/end2end/UpdateCacheFreqIT.java
palashc: Since this is mostly about indexes and positive UCF values, something on those lines would be great, whatever you decide should be good!

phoenix-core/src/it/java/org/apache/phoenix/end2end/UpdateCacheFreqIT.java
virajjasani: Except for Index jobs, do we any way to directly work on Index PTable object, bypassing data table PTable? If we rely on index jobs, it would be difficult to simulate when to switch over to new test coproc in the middle of job execution right?

phoenix-core/src/it/java/org/apache/phoenix/end2end/UpdateCacheFreqWithDDLValidationIT.java
palashc: Okay got it, I will look at these these tests as part of PHOENIX-7381.

phoenix-core/src/it/java/org/apache/phoenix/end2end/UpdateCacheFreqIT.java
palashc: What if in the middle of this test:
1. we change index state using ALTER command
2. switch to normal co-proc and do a query on the table so that the getTable goes through
3. then we switch to new coproc and try another query to see if there was a getTable or not

virajjasani: Query on the table will go through because the whole Data table PTable is in the client cache. We need query to Index table specifically.

phoenix-core/src/it/java/org/apache/phoenix/end2end/UpdateCacheFreqIT.java
virajjasani: Done

phoenix-core/src/it/java/org/apache/phoenix/end2end/UpdateCacheFreqIT.java
virajjasani: Done

phoenix-core-client/src/main/java/org/apache/phoenix/schema/MetaDataClient.java
```diff
@@ -785,19 +785,37 @@ private boolean avoidRpcToGetTable(boolean alwaysHitServer, Long resolvedTimesta
...
+            }
+
             // What if the table is created with UPDATE_CACHE_FREQUENCY explicitly set to ALWAYS?
             // i.e. explicitly set to 0. We should ideally be checking for something like
             // hasUpdateCacheFrequency().
 
             //always fetch an Index in PENDING_DISABLE state to retrieve server timestamp
             //QueryOptimizer needs that to decide whether the index can be used
-            if (PIndexState.PENDING_DISABLE.equals(table.getIndexState())) {
+            if (overrideUcfToDefault) {
```
palashc, line 813: @virajjasani Default today is ALWAYS so this would work but what if default is something else? Should we explicitly override to ALWAYS? 
This is why I had added a separate config for `DEFAULT_UPDATE_CACHE_FREQUENCY_FOR_PENDING_DISABLED_INDEX` and set that to ALWAYS.

phoenix-core/src/it/java/org/apache/phoenix/end2end/UpdateCacheFreqIT.java
palashc: @virajjasani Maybe I am missing something but the current tests still only check whether there are no getTable calls when all indexes are active and that there are getTable calls when there is some index in a "disabled" state. 

Should there be a test where index is created in CREATE_DISABLE state to begin with, client should be overriding UCF and doing getTable calls, then we change the index state to BUILDING OR ACTIVE - after one more getTable call, client should stop overriding UCF and not do any getTable calls. Since this is exactly what would happen for atleast our prod use-case, I think it would be good to have that tested. WDYT?

phoenix-core-client/src/main/java/org/apache/phoenix/schema/MetaDataClient.java
virajjasani: It seems we are not directly using ALWAYS anywhere but get connection level prop: `phoenix.default.update.cache.frequency`. While if this is not set, obviously it gets default value of `ALWAYS`, but if `phoenix.default.update.cache.frequency` is set at Connection level, it means we need to use it. It basically means that the client really does not care for `ALWAYS` at all, it is risky for client to even use this prop at Connection level but whoever uses it, must know what consequences they will deal with.

virajjasani: > This is why I had added a separate config for `DEFAULT_UPDATE_CACHE_FREQUENCY_FOR_PENDING_DISABLED_INDEX` and set that to ALWAYS.

Yes, i can remove this also with this PR.

phoenix-core/src/it/java/org/apache/phoenix/end2end/UpdateCacheFreqIT.java
virajjasani: The reason why i am using ALTER TABLE to set index state is because otherwise using async index means running jobs to get the index state updated right? We want to avoid that and directly check index table state to begin with. The testing is same but avoiding async index status update is better as it tends to make tests flaky. We don't know by when the index state will be updated.

phoenix-core-client/src/main/java/org/apache/phoenix/schema/MetaDataClient.java
palashc: For PHOENIX-6883, we wanted to set default UCF to NEVER and that is why this config was introduced. At this point, I am a little confused about how this UCF change and PHOENIX-6883 would work together in the future.

palashc (approved):
+1, pending test results
Thank you @virajjasani!

phoenix-core-client/src/main/java/org/apache/phoenix/schema/MetaDataClient.java
virajjasani: Exactly, we will need some level of re-iteration of PHOENIX-6883 with UCF. Ultimately, when client sets UCF to some value other than ALWAYS and NEVER, we need to ensure no getTable() or validateDdlTimestamp() RPC calls are initiated by clients for the duration of UCF. Its good that now we know this, so we will re-iterate on PHOENIX-6883 :)

PR #1964, opened 2024-08-29
PHOENIX-7382 Eliminating index building and treating max lookback as …

kadirozde:
…TTL for CDC Index

With one caveat: For CDC indexes on views, we do not treat max lookback as TTL in this PR. We can do that optimization later.


haridsv (commented):
LGTM, except for a few minor comments.

phoenix-core-server/src/main/java/org/apache/phoenix/coprocessor/GlobalIndexRegionScanner.java
```diff
@@ -273,11 +278,31 @@ public GlobalIndexRegionScanner(final RegionScanner innerScanner,
...
-
+    /**
+     * For CDC indexes we do not need to consider rows outside max lookback window or before
+     * the index create time. minTimestamp needs to be computed and used for CDC indexes always
+     * even when it is not set on the scan
+     */
+    private void computeMinTimestamp(Configuration config) throws IOException {
+        minTimestamp = scan.getTimeRange().getMin();
+        if (indexMaintainer.isCDCIndex()) {
+            minTimestamp = EnvironmentEdgeManager.currentTimeMillis() - maxLookBackInMills;
```
sanjeet006py, line 293: Currently, `maxLookback` is used only during Index verification and thus, table level max lookback is passed as scan attribute only when scan attributes: `INDEX_REBUILD_VERIFY_TYPE` and `REBUILD_INDEXES` are also set. But by shifting dependency on max lookback outside index verification we now need to pass table level max lookback as scan attribute wherever we are setting `REBUILD_INDEXES`. Example: In `GlobalIndexChecker.GlobalIndexScanner.repairIndexRows()`.

kadirozde: ServerBuildIndexCompiler is used for rebuild and verification and we set the scan attribute there. GlobalIndexChecker does one row repair for covered indexes. We do not need to set there as it is not used for uncovered indexes and thus CDC indexes.

sanjeet006py: Got it, thanks. I saw that `repairIndexRows()` is called for covered indexes only but missed the part that CDC indexes are uncovered. Thanks

sanjeet006py: `MetadataRegionObserver.BuildIndexSchedulerTask.run()` seems like the only other class which can be on Index rebuild path. This also sets `REBUILD_INDEXES` scan attribute but is not setting table level max lookback as scan attribute. Please correct me if I am wrong. Thanks

phoenix-core-client/src/main/java/org/apache/phoenix/schema/MetaDataClient.java
```diff
@@ -1946,27 +1946,32 @@ public MutationState createCDC(CreateCDCStatement statement) throws SQLException
...
-            indexProps.add("COLUMN_ENCODED_BYTES=" + columnEncodedBytes);
-        }
-        createIndexSql = createIndexSql + " " + String.join(", ", indexProps);
-        try {
-            pstmt.execute(createIndexSql);
-        } catch (SQLException e) {
+        Properties props = connection.getClientInfo();
+        props.put(INDEX_CREATE_DEFAULT_STATE, "ACTIVE");
+        try (Connection internalConnection = QueryUtil.getConnection(props, connection.getQueryServices().getConfiguration())) {
+            PhoenixStatement pstmt = new PhoenixStatement((PhoenixConnection) internalConnection);
```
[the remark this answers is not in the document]
tkhurana: @kadirozde There is an issue here if replication is enabled on the base table. If you enable CDC and it is active right away it will start getting writes which will be replicated to the DR. But if the corresponding create CDC statement has not executed on the DR it will block the replication on the DR because of TableNotFound exception.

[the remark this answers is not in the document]
kadirozde: The CDC index table should not be replicated. If needed, it can be rebuilt on the destination.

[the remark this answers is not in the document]
tkhurana: If that is the case, then we should enforce it by setting `REPLICATION_SCOPE to 0` when creating the CDC index table.

phoenix-core-server/src/main/java/org/apache/phoenix/coprocessor/GlobalIndexRegionScanner.java
kadirozde: That is for old indexes. The new indexes does use that code path.

phoenix-core/src/it/java/org/apache/phoenix/end2end/CDCQueryIT.java
```diff
@@ -701,12 +660,143 @@ public void testSelectCDCFailDataTableUpdate() throws Exception {
...
+            }
+
+            String tenantId = multitenant ? "1000" : null;
+            String[] tenantids = { tenantId };
+            if (multitenant) {
+                tenantids = new String[] { tenantId, "2000" };
+            }
+
+            long startTS = System.currentTimeMillis();
+            generateChanges(startTS, tenantids, tableFullName, tableFullName, COMMIT_SUCCESS, null);
```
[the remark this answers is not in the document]
kadirozde: I wanted to use and correct any issue with generateMutations.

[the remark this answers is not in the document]
haridsv: I am not clear what you mean, `generateMutations` is the newer method but you are using the older `generateChanges`.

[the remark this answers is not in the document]
kadirozde: I did want to follow the general structure in the IT and fix any issues within that. I  restructured it to leverage it for new tests and fixed/improved it. Please check the latest commit.

phoenix-core-server/src/main/java/org/apache/phoenix/coprocessor/GlobalIndexRegionScanner.java
sanjeet006py: @kadirozde I guess you mean that `MetadataRegionObserver.BuildIndexSchedulerTask.run()` isn't used anymore with new indexes? (Above sentence seems a bit confusing so, just confirming for my understanding).

phoenix-core/src/it/java/org/apache/phoenix/end2end/CDCBaseIT.java
```diff
@@ -428,8 +434,8 @@ protected List<ChangeRow> generateChanges(long startTS, String[] tenantids, Stri
                         }})
                 ));
                 committer.commit(conn);
-
-                changes.add(addChange(conn, tableName, new ChangeRow(tid, startTS += 100,
+                injectEdge.setValue(startTS += 100);
```
haridsv, line 448: This is already getting called by addChange, any specific reason to do this from here? This will get overwritten by next addChange call anyway.

phoenix-core/src/it/java/org/apache/phoenix/end2end/CDCQueryIT.java
```diff
@@ -701,8 +698,14 @@ public void testCDCIndexBuildAndVerification() throws Exception {
             }
 
             long startTS = System.currentTimeMillis();
-            generateChanges(startTS, tenantids, tableFullName, tableFullName, COMMIT_SUCCESS, null);
-            EnvironmentEdgeManager.reset();
+            List<ChangeRow> changes = generateChanges(startTS, tenantids, tableFullName,
+                    tableFullName, COMMIT_SUCCESS, null, 0);
+            // Make sure the timestamp of the mutations are not in the future
```
haridsv, line 705: Why not start with a startTS that is in the past instead of sleeping below?

phoenix-core/src/it/java/org/apache/phoenix/end2end/CDCBaseIT.java
kadirozde: I missed that it was called in addChange. I will remove them

phoenix-core-server/src/main/java/org/apache/phoenix/coprocessor/GlobalIndexRegionScanner.java
kadirozde: The old index design on an index write failure, transitions indexes to the PENDING_DISABLE state on the server side and after the client retries, and after retries are exhausted, disables indexes. MetadataRegionObserver rebuilds these indexes incrementally. The new design does not change the state of indexes on index write failure as these failures are handled by read repair. After upgrading to new indexes, phoenix.index.failure.handling.rebuild can be set to false to skip running BuildIndexScheduleTask. This is what we do in production.

phoenix-core/src/it/java/org/apache/phoenix/end2end/CDCQueryIT.java
kadirozde: If you do that then whenever we change the code in generateChanges, then we need to change the code here. Also, it is not straightforward to see how much time spent in generateChanges and thus it is not easy for a reader to see if we use the correct timestamp in the past. Further, I am sure if there will be some side effect if we rewind the time back.

phoenix-core-client/src/main/java/org/apache/phoenix/schema/MetaDataClient.java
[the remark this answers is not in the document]
tkhurana: @kadirozde Are you taking care of setting replication_scope to 0 as part of this PR ?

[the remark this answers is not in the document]
kadirozde: It is a good idea. Let me do that.

phoenix-core-client/src/main/java/org/apache/phoenix/schema/MetaDataClient.java
```diff
@@ -1634,7 +1636,9 @@ public MutationState createIndex(CreateIndexStatement statement, byte[][] splits
 
         Map<String,Object> tableProps = Maps.newHashMapWithExpectedSize(statement.getProps().size());
         Map<String,Object> commonFamilyProps = Maps.newHashMapWithExpectedSize(statement.getProps().size() + 1);
-        populatePropertyMaps(statement.getProps(), tableProps, commonFamilyProps, PTableType.INDEX);
+        populatePropertyMaps(statement.getProps(), tableProps, commonFamilyProps, PTableType.INDEX,
+                CDCUtil.isCDCIndex(SchemaUtil
+                        .getTableNameFromFullName(statement.getIndexTableName().toString())));
```
haridsv, line 1641: Do we need this special handling considering `createCDC` is explicitly adding `REPLICATION_SCOPE=0` property? We need only one of these right?

phoenix-core/src/it/java/org/apache/phoenix/end2end/CDCBaseIT.java
```diff
@@ -438,7 +445,6 @@ protected List<ChangeRow> generateChanges(long startTS, String[] tenantids, Stri
                         }})
                 ));
                 committer.commit(conn);
-
```
haridsv, line 448: nit: vertical spacing was meant to provide visual cue of the grouping for txn.

phoenix-core/src/it/java/org/apache/phoenix/end2end/CDCBaseIT.java
```diff
@@ -511,7 +517,6 @@ protected List<ChangeRow> generateChanges(long startTS, String[] tenantids, Stri
...
```
haridsv, line 521: nit: vertical spacing was meant to provide visual cue of the grouping for txn.

phoenix-core/src/it/java/org/apache/phoenix/end2end/CDCQueryIT.java
haridsv: > Further, I am sure if there will be some side effect if we rewind the time back.

Tried it in an IT, some future calls get stuck in sleep, I think sleep duration is getting miscalculated somewhere.

phoenix-core/src/test/java/org/apache/phoenix/util/TestUtil.java
```diff
@@ -972,24 +973,17 @@ public static void dumpTable(Table table) throws IOException {
     }
 
     public static int getRawRowCount(Table table) throws IOException {
+        dumpTable(table);
```
haridsv, line 976: Are you planning to merge with this change?

haridsv (approved):

PR #1965, opened 2024-08-29
PHOENIX-7343: Support ARRAY and JSON types as simple values in CDC

haridsv:
Jira: PHOENIX-7343

Added support for complex types ARRAY and JSON so that they are treated at simple values. As they are treated as simple values, changes with n the value will not be detected.


ranganathg (approved):

phoenix-core-server/src/main/java/org/apache/phoenix/coprocessor/CDCGlobalIndexRegionScanner.java
```diff
@@ -257,7 +252,7 @@ private Object getColumnValue(Cell cell, PDataType dataType) {
 
     private Object getColumnValue(byte[] cellValue, int offset, int length, PDataType dataType) {
         Object value;
-        if (dataType.getSqlType() == Types.BINARY) {
+        if (CDCUtil.isBinaryType(dataType)) {
```
virajjasani, line 255: This is an improvement with this PR, correct?

virajjasani (approved):
Left one question, +1 otherwise

phoenix-core-server/src/main/java/org/apache/phoenix/coprocessor/CDCGlobalIndexRegionScanner.java
haridsv: This is just a utility method to abstract out the 3 types that we are considering as binary type.

virajjasani: Oh i missed that it was Types.BINARY which anyways apply to them, i think i misunderstood earlier that we only had Binary datatype, and not VARBINARY.

virajjasani:
@haridsv i am about to merge the PR but just realized the PR has merge conflict, that most likely came from today's changes.

haridsv:
Yes, looks like there is a new binary type added to the condition, I will resolve the conflict.

virajjasani (approved):
+1, pending jenkins build

PR #1966, opened 2024-08-29
PHOENIX-7397 Optimize ClientAggregatePlan/ClientScanPlan when inner q…

comnetwork:
…uery is UnionPlan

The main modifications are as follows: 
In the `QueryCompiler.compileSingleQuery` method, if it is found that the inner subquery compilation results is a `UnionPlan`, the `UnionCompiler.optimizeUnionOrderByIfPossible `method is introduced to determine whether the inner `UnionPlan` meets the conditions for optimizing the sorting of the outer query：

-  If every subquery in `UnionPlan` is ordered, and `QueryPlan.getOutputOrderBys` of each subquery are equal(absolute equality or the same column name is unnecessary, just column types are compatible and columns count is same), and at the same time the outer query of `UnionPlan` has group by or order by, we would further examine whether
maintaining this order for the entire `UnionPlan` can compile out the outer query's group by or order by. 

- If it is sure, then `UnionPlan` just to perform a simple merge on the output of each subquery to ensure the overall order of the union all;
- otherwise, `UnionPlan` would not perform any special processing on the output of the subqueries.

### DIFF
--- a/phoenix-core-client/src/main/java/org/apache/phoenix/parse/OrderByNode.java
+++ b/phoenix-core-client/src/main/java/org/apache/phoenix/parse/OrderByNode.java
@@ -87,14 +87,14 @@ public final class OrderByNode {
     }
 
     public boolean isLiteral() {
-        return child instanceof LiteralParseNode &&
-                ((LiteralParseNode)child).getType() == PInteger.INSTANCE;
+        return child instanceof LiteralParseNode
+                && ((LiteralParseNode) child).getType() == PInteger.INSTANCE;
     }
 
     public Integer getIntValueIfLiteral() {
         if (!isLiteral()) {
             return null;
         }
-        return (Integer)((LiteralParseNode)child).getValue();
+        return (Integer) ((LiteralParseNode) child).getValue();
     }
 }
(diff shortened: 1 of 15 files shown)


phoenix-core-client/src/main/java/org/apache/phoenix/parse/OrderByNode.java
```diff
@@ -84,4 +85,16 @@ public void toSQL(ColumnResolver resolver, StringBuilder buf) {
         if (!orderAscending) buf.append(" DESC");
         if (nullsLast) buf.append(" NULLS LAST ");
     }
+
+    public boolean isLiteral() {
```
tkhurana, line 89: Maybe we could rename this method to `isIntegerLiteral`

tkhurana (approved):
One small comment but other than that looks good. Thanks

comnetwork:
@tkhurana, thank you very much for reviewing, already fixed the problem.

phoenix-core-client/src/main/java/org/apache/phoenix/parse/OrderByNode.java
comnetwork: Ok, already fixed.

PR #1967, opened 2024-08-30
PHOENIX-7398 New PhoenixStatement API to return row for Atomic/Conditional Upserts

virajjasani:
Jira: PHOENIX-7398

Phoenix supports Atomic Conditional Upserts with ON DUPLICATE KEY clause, allowing clients to either ignore the update if the row with the given primary key(s) already exist or conditionally update the row with the given primary key(s). Phoenix also supports returning 0 or 1 as updated status code for the given Atomic/Conditional Upserts, where 0 represents no updates to the row and 1 represents the updated row (when the condition is satisfied by the atomic upsert).

While standard JDBC APIs support Upserts with the status code returned as integer values 0 or 1, some client applications also require returning the current row state back to the client depending on whether the row is successfully updated. If the condition is satisfied by the atomic upsert and the row is updated, the client application should be able to get the updated row status as HBase Result object. Similarly, if the condition is not satisfied for the given atomic upsert and therefore, row is not updated, the client application should be able to get the old/current row status as HBase Result object.

This support is somewhat similar in nature to what HBase provides for Increment and Append mutations. Both operations return updated row as Result object. For instance,

```
/**
 * Appends values to one or more columns within a single row.
 * <p>
 * This operation guaranteed atomicity to readers. Appends are done under a single row lock, so
 * write operations to a row are synchronized, and readers are guaranteed to see this operation
 * fully completed.
 * @param append object that specifies the columns and values to be appended
 * @throws IOException e
 * @return values of columns after the append operation (maybe null)
 */
default Result append(final Append append) throws IOException {
```

HBase does not provide API to return row state for atomic updates with Put and Delete mutations. HBase API checkAndMutate() also supports returning Result for Append and Increment mutations only, not for Put and Delete mutations.

Phoenix supports batchMutate() with Put mutation(s) by making it Atomic. Hence, for client applications using Phoenix provided atomic upserts, it would be really beneficial to also provide new API at PhoenixStatement and PhoenixPreparedStatement layer that can return the row as Result object back to the client.

The proposed API signature:

```
/**
 *
 */
```

### DIFF
--- a/phoenix-core-client/src/main/java/org/apache/phoenix/execute/MutationState.java
+++ b/phoenix-core-client/src/main/java/org/apache/phoenix/execute/MutationState.java
@@ -187,6 +187,9 @@ public class MutationState implements SQLCloseable {
 
     private final boolean indexRegionObserverEnabledAllTables;
 
+    private ReturnResult returnResult;
+    private Result result;
+
     public static void resetAllMutationState(){
         allDeletesMutations = true;
         allUpsertsMutations = true;
@@ -480,6 +483,10 @@ public class MutationState implements SQLCloseable {
 
     public int getNumRows() {
         return numRows;
+    }
+
+    public Result getResult() {
+        return this.result;
     }
 
     private MultiRowMutationState getLastMutationBatch(Map<TableRef, List<MultiRowMutationState>> mutations, TableRef tableRef) {
@@ -847,6 +854,12 @@ public class MutationState implements SQLCloseable {
                     if (onDupKeyBytes != null) {
                         mutation.setAttribute(PhoenixIndexBuilderHelper.ATOMIC_OP_ATTRIB, onDupKeyBytes);
                     }
+                    if (this.returnResult != null) {
+                        if (this.returnResult == ReturnResult.ROW) {
+                            mutation.setAttribute(PhoenixIndexBuilderHelper.RETURN_RESULT,
+                                    PhoenixIndexBuilderHelper.RETURN_RESULT_ROW);
+                        }
+                    }
                 }
                 rowMutationsPertainingToIndex = rowMutations;
             }
@@ -1125,6 +1138,10 @@ public class MutationState implements SQLCloseable {
 
     public long getBatchCount() {
         return batchCount;
+    }
+
+    public void setReturnResult(ReturnResult returnResult) {
+        this.returnResult = returnResult;
     }
 
     public static final class MutationBytes {
@@ -1550,6 +1567,11 @@ public class MutationState implements SQLCloseable {
                                 numUpdatedRowsForAutoCommit = PInteger.INSTANCE.getCodec()
                                         .decodeInt(cell.getValueArray(), cell.getValueOffset(),
                                                 SortOrder.getDefault());
+                                if (this.returnResult != null) {
+                                    if (this.returnResult == ReturnResult.ROW) {
+                                        this.result = result;
+                                    }
+                                }
                             } else {
                                 numUpdatedRowsForAutoCommit = 1;
                             }
@@ -2418,4 +2440,9 @@ public class MutationState implements SQLCloseable {
     public boolean isEmpty() {
         return mutationsMap != null ? mutationsMap.isEmpty() : true;
     }
+
+    public enum ReturnResult {
+        ROW
+    }
+
 }
--- a/phoenix-core-client/src/main/java/org/apache/phoenix/execute/MutationState.java
+++ b/phoenix-core-client/src/main/java/org/apache/phoenix/execute/MutationState.java
@@ -187,6 +187,12 @@ public class MutationState implements SQLCloseable {
 
     private final boolean indexRegionObserverEnabledAllTables;
 
+    /**
+     * Return result back to client. To be used when client needs to read the whole row
+     * or some specific attributes of the row back. As of PHOENIX-7398, returning whole
+     * row is supported. This is used to allow client to set Mutation attribute that is read
+     * by server for it to atomically read the row and return it back.
+     */
     private ReturnResult returnResult;
     private Result result;
 
@@ -2027,6 +2033,7 @@ public class MutationState implements SQLCloseable {
         estimatedSize = 0;
         this.mutationsMap.clear();
         phoenixTransactionContext = PhoenixTransactionContext.NULL_CONTEXT;
+        this.returnResult = null;
     }
 
     private void resetTransactionalState() {
--- a/phoenix-core-client/src/main/java/org/apache/phoenix/jdbc/PhoenixStatement.java
+++ b/phoenix-core-client/src/main/java/org/apache/phoenix/jdbc/PhoenixStatement.java
@@ -72,6 +72,7 @@ import org.apache.hadoop.fs.Path;
 import org.apache.hadoop.hbase.Cell;
 import org.apache.hadoop.hbase.CellComparator;
 import org.apache.hadoop.hbase.client.Consistency;
+import org.apache.hadoop.hbase.client.Result;
 import org.apache.hadoop.hbase.client.Scan;
 import org.apache.hadoop.hbase.util.Pair;
 import org.apache.phoenix.call.CallRunner;
@@ -107,6 +108,7 @@ import org.apache.phoenix.exception.SQLExceptionInfo;
 import org.apache.phoenix.exception.StaleMetadataCacheException;
 import org.apache.phoenix.exception.UpgradeRequiredException;
 import org.apache.phoenix.execute.MutationState;
+import org.apache.phoenix.execute.MutationState.ReturnResult;
 import org.apache.phoenix.execute.visitor.QueryPlanVisitor;
 import org.apache.phoenix.expression.KeyValueColumnExpression;
 import org.apache.phoenix.expression.RowKeyColumnExpression;
@@ -576,11 +578,22 @@ public class PhoenixStatement implements PhoenixMonitoredStatement, SQLCloseable
     }
 
 
-    protected int executeMutation(final CompilableStatement stmt, final AuditQueryLogger queryLogger) throws SQLException {
-        return executeMutation(stmt, true, queryLogger);
+    protected int executeMutation(final CompilableStatement stmt,
+                                  final AuditQueryLogger queryLogger) throws SQLException {
+        return executeMutation(stmt, true, queryLogger, null).getFirst();
     }
 
-    private int executeMutation(final CompilableStatement stmt, final boolean doRetryOnMetaNotFoundError, final AuditQueryLogger queryLogger) throws SQLException {
+    Pair<Integer, Result> executeMutation(final CompilableStatement stmt,
+                                   final AuditQueryLogger queryLogger,
+                                   final ReturnResult returnResult) throws SQLException {
+        return executeMutation(stmt, true, queryLogger, returnResult);
+    }
+
+    private Pair<Integer, Result> executeMutation(final CompilableStatement stmt,
+                                                  final boolean doRetryOnMetaNotFoundError,
+                                                  final AuditQueryLogger queryLogger,
+                                                  final ReturnResult returnResult)
+            throws SQLException {
         if (connection.isReadOnly()) {
             throw new SQLExceptionInfo.Builder(
                 SQLExceptionCode.READ_ONLY_CONNECTION).
@@ -590,9 +603,9 @@ public class PhoenixStatement implements PhoenixMonitoredStatement, SQLCloseable
         try {
             return CallRunner
                    .run(
-                        new CallRunner.CallableThrowable<Integer, SQLException>() {
+                        new CallRunner.CallableThrowable<Pair<Integer, Result>, SQLException>() {
                         @Override
-                            public Integer call() throws SQLException {
+                            public Pair<Integer, Result> call() throws SQLException {
                             boolean success = false;
                             String tableName = null;
                             boolean isUpsert = false;
@@ -631,12 +644,15 @@ public class PhoenixStatement implements PhoenixMonitoredStatement, SQLCloseable
                                 // just max out at Integer.MAX_VALUE
                                 int lastUpdateCount = (int) Math.min(Integer.MAX_VALUE,
                                         lastState.getUpdateCount());
+                                Result result = null;
                                 if (connection.getAutoCommit()) {
+                                    state.setReturnResult(returnResult);
                                     connection.commit();
                                     if (isAtomicUpsert) {
                                         lastUpdateCount = connection.getMutationState()
                                                 .getNumUpdatedRowsForAutoCommit();
                                     }
+                                    result = connection.getMutationState().getResult();
                                 }
                                 setLastQueryPlan(null);
                                 setLastUpdateCount(lastUpdateCount);
@@ -651,7 +667,7 @@ public class PhoenixStatement implements PhoenixMonitoredStatement, SQLCloseable
                                 }
 
                                 success = true;
-                                return lastUpdateCount;
+                                return new Pair<>(lastUpdateCount, result);
                             }
                             //Force update cache and retry if meta not found error occurs
                             catch (MetaDataEntityNotFoundException e) {
@@ -661,7 +677,8 @@ public class PhoenixStatement implements PhoenixMonitoredStatement, SQLCloseable
                                     }
                                     if (new MetaDataClient(connection).updateCache(connection.getTenantId(),
                                         e.getSchemaName(), e.getTableName(), true).wasUpdated()) {
-                                        return executeMutation(stmt, false, queryLogger);
+                                        return executeMutation(stmt, false, queryLogger,
+                                                returnResult);
                                     }
                                 }
                                 throw e;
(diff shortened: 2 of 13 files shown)


phoenix-core-client/src/main/java/org/apache/phoenix/jdbc/PhoenixStatement.java
```diff
@@ -2376,17 +2393,42 @@ public ResultSet executeQuery(String sql) throws SQLException {
...
+     * auto-commit Connection. This makes the operation atomic.
+     * If the row is successfully updated, return the updated row, otherwise if the row
+     * cannot be updated, return non-updated row.
+     *
+     * @param sql The SQL DML statement, UPSERT or DELETE for Phoenix.
+     * @return The pair of int and Result, where int represents value 1 for successful row update
+     * and 0 for non-successful row update, and Result represents the state of the row.
+     * @throws SQLException If the statement cannot be executed.
+     */
+    public Pair<Integer, Result> executeUpdateReturnRow(String sql) throws SQLException {
```
tkhurana, line 2414: It will be much cleaner and simpler for the application users if this API returns a ResultSet object. This is an external facing API and it is not easy to use the Result interface.

virajjasani: That is correct but it requires one to one mapping of Result to ResultSet, which we don't have. I expect that we can take it up as separate improvement and change this later.

virajjasani: ResultSet implementation is built entirely on iterators.

virajjasani: Also, Result is single row result whereas ResultSet is not necessarily meant to be single row result because we can call `ResultSet#next` and it is meant to move the cursor forward.

tkhurana: Well, it is the standard way to return information back. Single row or multiple rows doesn't matter. For example aggregate query like SELECT count(*) results are also returned through the same interface.

phoenix-core-client/src/main/java/org/apache/phoenix/jdbc/PhoenixStatement.java
```diff
@@ -2376,17 +2393,42 @@ public ResultSet executeQuery(String sql) throws SQLException {
...
+    }
+
+    /**
+     * Executes the given SQL statement similar to JDBC API executeUpdate() but also returns the
+     * updated or non-updated row as Result object back to the client. This must be used with
+     * auto-commit Connection. This makes the operation atomic.
+     * If the row is successfully updated, return the updated row, otherwise if the row
+     * cannot be updated, return non-updated row.
+     *
+     * @param sql The SQL DML statement, UPSERT or DELETE for Phoenix.
```
tkhurana, line 2409: How does it work for DELETE ? I can delete multiple rows using a single delete query. Does it handle that case ?

virajjasani: DELETE support is coming as separate follow up Jira. I did not want to complicate this PR.

phoenix-core-client/src/main/java/org/apache/phoenix/jdbc/PhoenixStatement.java
virajjasani: Yes, i also agree with it, but what i meant is that the whole ResultSet interface in Phoenix is built on top of RowProjector and ResultIterator. We do not have support for 1:1 HBase Result to Phoenix ResultSet conversion support. It would be interesting to build one but i think it should not be scope of this work.

phoenix-core-client/src/main/java/org/apache/phoenix/jdbc/PhoenixStatement.java
virajjasani: But yes reg your question, since single DELETE can delete multiple rows, similar to UPSERT, we want to restrict returning value only if single row was deleted.

phoenix-core-client/src/main/java/org/apache/phoenix/jdbc/PhoenixStatement.java
tkhurana: At-least we could use `ResultTuple` which builds a nice wrapper over hbase Result

virajjasani: Yes, used `Tuple` interface in the latest revision.

phoenix-core-server/src/main/java/org/apache/phoenix/index/PhoenixIndexBuilder.java
```diff
@@ -102,7 +102,8 @@ public void batchStarted(MiniBatchOperationInProgress<Mutation> miniBatchOp, Ind
     
     @Override
     public boolean isAtomicOp(Mutation m) {
-        return m.getAttribute(PhoenixIndexBuilderHelper.ATOMIC_OP_ATTRIB) != null;
+        return m.getAttribute(PhoenixIndexBuilderHelper.ATOMIC_OP_ATTRIB) != null ||
```
tkhurana, line 105: I am slightly confused here. I was under the impression that RETURN_RESULT is only supported for atomic upserts. But this implies non atomic upserts could also get return result.

virajjasani: Yes that's correct but this is done to do some preparation to support DELETE in advance. Because with Delete, we don't have ON DUPLICATE KEY clause, nor do we want to support it.

tkhurana: If returning results needs to exist independently then `isAtomicOP` API shouldn't change. Rather the caller of the API should check `if (isAtomicOp() || hasReturnResult)`

tkhurana: Another reason is let say tomorrow I want to do something just for ON DUPLICATE KEY but not for return result. I cannot use the `isAtomicOp()` API anymore because when it returns true I cannot distinguish between the 2 cases.

virajjasani: I was thinking about that and then realized that even if we do it, we sort of are making returnResult as atomic usecase internally (if not for UPSERT, at least for DELETE), that's why i kept it here. WDYT? Shall we still keep it this way or make callers do extra call for where `if (isAtomicOp() || hasReturnResult)` is true?

phoenix-core-client/src/main/java/org/apache/phoenix/jdbc/PhoenixStatement.java
```diff
@@ -2376,17 +2394,42 @@ public ResultSet executeQuery(String sql) throws SQLException {
...
+     * @return The pair of int and Tuple, where int represents value 1 for successful row
+     * update and 0 for non-successful row update, and Tuple represents the state of the row.
+     * @throws SQLException If the statement cannot be executed.
+     */
+    public Pair<Integer, Tuple> executeUpdateReturnRow(String sql) throws SQLException {
```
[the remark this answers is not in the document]
virajjasani: Absolutely, will do

phoenix-core-server/src/main/java/org/apache/phoenix/index/PhoenixIndexBuilder.java
virajjasani: > Another reason is let say tomorrow I want to do something just for ON DUPLICATE KEY but not for return result.

That sounds valid, let me make them different.

tkhurana: I think we should keep it separate.

phoenix-core/src/it/java/org/apache/phoenix/end2end/OnDuplicateKey2IT.java
```diff
@@ -118,6 +137,148 @@ public void testIgnoreReturnValue() throws Exception {
...
+                            + "VALUES('pk000', -123.98, 'pk003', 0) ON DUPLICATE"
+                            + " KEY IGNORE";
+            validateReturnedRowAfterUpsert(conn, upsertSql, tableName, 1011.202, null, false,
+                    null, bsonDocument1, 123);
+
+            upsertSql =
+                    "UPSERT INTO " + tableName
+                            + " (PK1, PK2, PK3, COUNTER1, COUNTER2) VALUES('pk000', -123.98, "
+                            + "'pk003', 234, 'col2_000')";
+            validateReturnedRowAfterUpsert(conn, upsertSql, tableName, 234d, "col2_000", true,
```
tkhurana, line 177: In case of partial update, will the returned row be a full row ?

tkhurana: In the above example we are doing a regular upsert with returned row.

virajjasani: > In case of partial update, will the returned row be a full row ?

That's correct, full row version will be returned. That's the main purpose actually: client can perform partial row update and see what the entire row looks like.

> In the above example we are doing a regular upsert with returned row.

Yes, just an example of how DELETE will also look like. But the point is, the server allows returning row only if client uses new API to modify only single row.

phoenix-core-server/src/main/java/org/apache/phoenix/hbase/index/IndexRegionObserver.java
```diff
@@ -1146,14 +1155,18 @@ private void identifyMutationTypes(MiniBatchOperationInProgress<Mutation> miniBa
                                               BatchMutateContext context) {
         for (int i = 0; i < miniBatchOp.size(); i++) {
             Mutation m = miniBatchOp.getOperation(i);
-            if (this.builder.isAtomicOp(m)) {
+            if (this.builder.returnResult(m)) {
+                context.returnResult = true;
```
tkhurana, line 1159: Should we only set this to true if batch size is 1 ? I think that will simplify because I saw in multiple places we are doing both checks.

virajjasani: Good idea, will do

phoenix-core-server/src/main/java/org/apache/phoenix/hbase/index/IndexRegionObserver.java
```diff
@@ -1586,29 +1661,61 @@ public Void visit(KeyValueColumnExpression expression) {
...
       ImmutableBytesPtr rowKeyPtr = new ImmutableBytesPtr(rowKey);
       // Get the latest data row state
       Pair<Put, Put> dataRowState = context.dataRowStates.get(rowKeyPtr);
       Put currentDataRowState = dataRowState != null ? dataRowState.getFirst() : null;
 
+        // if result needs to be returned but the DML does not have ON DUPLICATE KEY present,
+        // perform the mutation and return the result.
+        if (opBytes == null) {
+            mutations.add(atomicPut);
+            updateCurrColumnCellExpr(atomicPut, currColumnCellExprMap);
```
tkhurana, line 1702: atomicPut will only contain columns which are being updated. So in case of partial update we will be returning only those columns instead of the full row.

virajjasani: But we already have colCellExprMap generated from the row that we scan with point lookup as part of atomic upsert. Then we update the only content (columns) that are being updated with the atomicPut.

phoenix-core-client/src/main/java/org/apache/phoenix/execute/MutationState.java
```diff
@@ -187,6 +187,9 @@ public class MutationState implements SQLCloseable {
 
     private final boolean indexRegionObserverEnabledAllTables;
 
+    private ReturnResult returnResult;
```
tkhurana, line 190: Can you add some comments here explaining what this fields mean

virajjasani: Sure

tkhurana: Also, do we need to re-initialize them when the mutation state is `reset` ? If somebody re-uses the same connection .

phoenix-core/src/it/java/org/apache/phoenix/end2end/OnDuplicateKey2IT.java
```diff
@@ -118,6 +139,177 @@ public void testIgnoreReturnValue() throws Exception {
...
```
virajjasani, line 179: @tkhurana this is example of partial row update where the API returns full row state.

phoenix-core-server/src/main/java/org/apache/phoenix/hbase/index/IndexRegionObserver.java
tkhurana: Can you point where is that happening ?

phoenix-core/src/it/java/org/apache/phoenix/end2end/OnDuplicateKey2IT.java
virajjasani: COL3 and COL4 becomes null because the UPSERT updated only COUNTER1 and COUNTER2

phoenix-core-server/src/main/java/org/apache/phoenix/hbase/index/IndexRegionObserver.java
```diff
@@ -1586,29 +1661,61 @@ public Void visit(KeyValueColumnExpression expression) {
...
+            if (context.returnResult) {
+                context.currColumnCellExprMap = currColumnCellExprMap;
```
virajjasani, line 1704: FYI @tkhurana, here we update the context level map with the current row state i.e. all columns of the current row are part of the context map as of this moment.

phoenix-core/src/it/java/org/apache/phoenix/end2end/OnDuplicateKey2IT.java
tkhurana: Why will COL3 and COL4 become null if they are not being updated ?

phoenix-core-server/src/main/java/org/apache/phoenix/hbase/index/IndexRegionObserver.java
tkhurana: AFAIK `atomicPut` will only have the partial update. The `currentDataRowState` will have the full row but I don't see that being used in this case.

virajjasani: There is one oversight that you caught. For ON DUPLICATE (atomic) cases, i have already addressed using `currentDataRowState`, only for non-atomic case i was not using `currentDataRowState` when applicable.

For atomic case, here is what this PR introduces for reading the entire row while building the map:
```
      if (PhoenixIndexBuilderHelper.isDupKeyIgnore(opBytes)) {
          if (currentDataRowState == null) {
              // new row
              mutations.add(atomicPut);
              updateCurrColumnCellExpr(atomicPut, currColumnCellExprMap); <========== if row does not exist
          } else {
              updateCurrColumnCellExpr(currentDataRowState, currColumnCellExprMap); <======== if row exist, use it to formulate currColumnCellExprMap
          }
          if (context.returnResult) {
              context.currColumnCellExprMap = currColumnCellExprMap;
          }
          return mutations;
      }
```

phoenix-core/src/it/java/org/apache/phoenix/end2end/OnDuplicateKey2IT.java
virajjasani: Let me fix this, good catch, i did not realize that for non-atomic (without using ON DUPLICATE clause) this was an oversight.

phoenix-core-client/src/main/java/org/apache/phoenix/execute/MutationState.java
virajjasani: We can re-set it as part of `reset` but i think it is not strictly necessary because `PhoenixStatement` sets it only for auto-commit based new API usage, but its still good practice to reset it as part of the `MutationState` reset.

phoenix-core-server/src/main/java/org/apache/phoenix/hbase/index/IndexRegionObserver.java
virajjasani: Similarly, for ON DUPLICATE KEY UPDATE clause,

```
      if (currentDataRowState == null) { // row doesn't exist
          updateCurrColumnCellExpr(atomicPut, currColumnCellExprMap);
...
...
      } else {
          // Base current state off of existing row
          flattenedCells = cells;
          // store all current cells from currentDataRowState
          updateCurrColumnCellExpr(currentDataRowState, currColumnCellExprMap);
      }

```

If the row already exists, we build the map based on the existing row, otherwise from atomicPut.

tkhurana (approved):

PR #1969, opened 2024-09-03
PHOENIX-7239: Fix issues when data and index have different salt buckets

haridsv:
This PR includes the following changes:
* Currently IndexMaintainer assumes that both data and index tables have the same number of salt buckets. This change adds a separate field for data table and uses it for constructing the data keys. This fixes the joins between uncovered index and data tables.
* There is a typo in the CDC mutation code path that uses data table to construct the index row key instead of the index table which was throwing off the logic that join the changes to the data rows.
* Test for time range queries has been rewritten to be simpler and also add coverage for the above special DF markers.
* Some of the existing debug code in the CDC IT test has been refactored to be less repetitive.

### DIFF
--- a/phoenix-core-client/src/main/java/org/apache/phoenix/index/IndexMaintainer.java
+++ b/phoenix-core-client/src/main/java/org/apache/phoenix/index/IndexMaintainer.java
@@ -397,6 +397,7 @@ public class IndexMaintainer implements Writable, Iterable<ColumnReference> {
     private RowKeyMetaData rowKeyMetaData;
     private byte[] indexTableName;
     private int nIndexSaltBuckets;
+    private int nDataTableSaltBuckets;
     private byte[] dataEmptyKeyValueCF;
     private ImmutableBytesPtr emptyKeyValueCFPtr;
     private int nDataCFs;
@@ -470,6 +471,7 @@ public class IndexMaintainer implements Writable, Iterable<ColumnReference> {
         this.immutableStorageScheme = index.getImmutableStorageScheme() == null ? ImmutableStorageScheme.ONE_CELL_PER_COLUMN : index.getImmutableStorageScheme();
         this.dataEncodingScheme = dataTable.getEncodingScheme() == null ? QualifierEncodingScheme.NON_ENCODED_QUALIFIERS : dataTable.getEncodingScheme();
         this.dataImmutableStorageScheme = dataTable.getImmutableStorageScheme() == null ? ImmutableStorageScheme.ONE_CELL_PER_COLUMN : dataTable.getImmutableStorageScheme();
+        this.nDataTableSaltBuckets = isDataTableSalted ? dataTable.getBucketNum() : PTable.NO_SALTING;
 
         byte[] indexTableName = index.getPhysicalName().getBytes();
         // Use this for the nDataSaltBuckets as we need this for local indexes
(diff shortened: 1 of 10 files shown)


phoenix-core-client/src/main/java/org/apache/phoenix/index/IndexMaintainer.java
```diff
@@ -470,6 +471,7 @@ private IndexMaintainer(final PTable dataTable, final PTable cdcTable, final PTa
         this.immutableStorageScheme = index.getImmutableStorageScheme() == null ? ImmutableStorageScheme.ONE_CELL_PER_COLUMN : index.getImmutableStorageScheme();
         this.dataEncodingScheme = dataTable.getEncodingScheme() == null ? QualifierEncodingScheme.NON_ENCODED_QUALIFIERS : dataTable.getEncodingScheme();
         this.dataImmutableStorageScheme = dataTable.getImmutableStorageScheme() == null ? ImmutableStorageScheme.ONE_CELL_PER_COLUMN : dataTable.getImmutableStorageScheme();
+        this.nDataTableSaltBuckets = isDataTableSalted ? dataTable.getBucketNum() : PTable.NO_SALTING;
```
tkhurana, line 474: Any reason why are you introducing the special value of -1 ? The field `nIndexSaltBuckets` is set to 0 if index is not salted. We should maintain consistency.

phoenix-core-client/src/main/java/org/apache/phoenix/index/IndexMaintainer.java
```diff
@@ -1789,6 +1788,9 @@ public static IndexMaintainer fromProto(ServerCachingProtos.IndexMaintainer prot
         } else {
             maintainer.isCDCIndex = false;
         }
+        if (proto.hasDataTableSaltBuckets()) {
+            maintainer.nDataTableSaltBuckets = proto.getDataTableSaltBuckets();
```
tkhurana, line 1792: If the proto doesn't have data table salt buckets then what is the nDataTableSaltBuckets field ? I think it will be 0 not -1. Seems like -1 is constant that is introduced is unnecessary.

phoenix-core-client/src/main/java/org/apache/phoenix/schema/PTable.java
```diff
@@ -63,6 +63,7 @@ public interface PTable extends PMetaDataEntity {
     public static final String IS_IMMUTABLE_ROWS_PROP_NAME = "IMMUTABLE_ROWS";
     public static final boolean DEFAULT_DISABLE_WAL = false;
     public static final boolean DEFAULT_IMMUTABLE_ROWS = false;
+    static final Integer NO_SALTING = -1;
```
tkhurana, line 66: I don't think we need this special -1 value.

tkhurana:
@haridsv It would have been better if you had only salt bucket fix in this PR and not CDC related changes.

phoenix-core-client/src/main/java/org/apache/phoenix/index/IndexMaintainer.java
haridsv: In the existing code, `PTableImpl.toProto()` translates `0` into `-1` and `PTableImpl.createFromProto()` leaves this value as `null` by checking for `-1`. To be consistent, I think we should change `nIndexSaltBuckets` to also `-1` instead of the other way around. I did a quick check, I think the existing code will work as is, as it checks for `> 0`. WDYT?

phoenix-core-client/src/main/java/org/apache/phoenix/index/IndexMaintainer.java
haridsv: Agree on the inconsistency, but I think it should be set to `-1` to be consistent than `0` for the same reason I [mentioned](https://github.com/apache/phoenix/pull/1969/files#r1770766770) in the other thread.

phoenix-core-client/src/main/java/org/apache/phoenix/index/IndexMaintainer.java
tkhurana: Sounds good @haridsv

tkhurana (approved):

PR #1970, opened 2024-09-03
Exclude libraries with CVEs from phoenix-client-embedded

kokosing:


kokosing:
FYI @stoty @virajjasani

stoty:
I have opened a [DISCUSS] thread on dev@p.a.o. on how to handle transitive dependecy CVEs.
Please join the discussion.

virajjasani:
Do we know if CVEs of both of these dependencies have been resolved with some recently released version of Hadoop?

stoty:
I have added dependencyManagement for these in a different commit.

PR #1973, opened 2024-09-09
PHOENIX-7402: Fix row getting expired partially even if updated within TTL

sanjeet006py:
The issues that were identified:
- The empty cells are not being persisted/collected during flushes and minor compactions which causes row to expire partially on next major compaction even if it was updated within TTL.
- Removed `preFlush` hook from using CompactionScanner as we would need to retrieve PTable object in `preFlush` to pass onto CompactionScanner to determine the max lookback age and empty column family. As we need to use latest max lookback age always so, we would need to retrieve PTable by bypassing client cache and make RPC to SYSCAT RS.
- For multi-column family, we need to handle the case that during major compactions we can be needed to retain empty column cells to close the gap b/w empty CF and non-empty CF. To know minimum number of empty column cells to retain for this purpose we need to do region level scan.
- For minor compactions, because actual TTL is not known as getting View level TTL requires running queries on SYSCAT thus, we retain the empty cells which can't be safely discarded.

### DIFF
--- a/phoenix-core-server/src/main/java/org/apache/phoenix/coprocessor/CompactionScanner.java
+++ b/phoenix-core-server/src/main/java/org/apache/phoenix/coprocessor/CompactionScanner.java
@@ -2181,7 +2181,7 @@ public class CompactionScanner implements InternalScanner {
             Cell cellAtMaxLookbackWindowStart = null;
             for (Cell cell : result) {
                 long maxLookbackWindowStart = rowTracker.getRowContext().getMaxLookbackWindowStart();
-                if (cell.getTimestamp() > maxLookbackWindowStart) {
+                if (cell.getTimestamp() >= maxLookbackWindowStart) {
                     retainedCells.add(cell);
                     if (cell.getTimestamp() == maxLookbackWindowStart) {
                         cellAtMaxLookbackWindowStart = cell;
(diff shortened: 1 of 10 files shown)


sanjeet006py:
@kadirozde @virajjasani @tkhurana @jpisaac The PR is ready for review. Please review. 

I plan to follow-up with a PR to avoid doing region level scan when major compacting empty CF of a multi-CF table. I have thought of an approach but need to do its feasibility analysis and get it more concrete so, maybe first we can merge this PR and then next I can follow-up with optimization. Thanks

phoenix-core-server/src/main/java/org/apache/phoenix/coprocessor/CompactionScanner.java
```diff
@@ -154,6 +155,7 @@ public class CompactionScanner implements InternalScanner {
     private long inputCellCount = 0;
     private long outputCellCount = 0;
     private boolean phoenixLevelOnly = false;
+    private static boolean forceMinorCompaction = false;
```
kadirozde, line 158: Adding a comment stating that this is used only for testing would help here.

kadirozde:
> @kadirozde @virajjasani @tkhurana @jpisaac The PR is ready for review. Please review.
> 
> I plan to follow-up with a PR to avoid doing region level scan when major compacting empty CF of a multi-CF table. I have thought of an approach but need to do its feasibility analysis and get it more concrete so, maybe first we can merge this PR and then next I can follow-up with optimization. Thanks

Sounds good.

kadirozde (approved):
+1, Thanks!

PR #1974, opened 2024-09-09
Build the HBase 2.5+ profiles with Hadoop 3.3.6

stoty:


stoty:
Need to run some tests to see if the minicluster binaries can handle the newer Hadoop (especially for 2.5.0 and 2.4)

stoty:
2.4 tests fail with Hadoop 3.3 because of the known minicluster incompatibilities, but mixing Hadoop 3.2 and 3.3 does not cause any test failures.

2.4 is EOL anyway.

virajjasani:
Do we need to run all tests for 2.5 and 2.6 profiles? or let precommit builds run it?

virajjasani (approved):
+1, pending Jira title

stoty:
I have run the test for 2.5 manually, and did not see more than the usual failures @virajjasani .
2.6 is tested by the precommit job.

stoty:
I did not realize I missed the JIRA ID from the commit message, thanks for catching that.

stoty:
merged manually

PR #1976, opened 2024-09-11
PHOENIX-7395 Metadata Cache metrics at server and client side

jinggou:
Jira: [PHOENIX-7395](https://issues.apache.org/jira/browse/PHOENIX-7395)


phoenix-core/src/it/java/org/apache/phoenix/end2end/MetaDataEndPointIT.java
```diff
@@ -54,4 +60,107 @@ public void testUpdateIndexState() throws Throwable {
...
+            expectedDropTableCount += 1;
+            IndexMetricsIT.verifyCounterWithValue(MetricsMetadataSource.ALTER_ADD_COLUMN_COUNT,
+                    registry, expectedDropTableCount);
+
+//            System.out.println(expectedCacheUsedSize);
+//            System.out.println(tableEstimatedSize);
+//            System.out.println(newTableEstimatedSize);
+//            expectedCacheUsedSize = expectedCacheUsedSize - tableEstimatedSize + newTableEstimatedSize;
+//            IndexMetricsIT.verifyCounterWithValue(MetricsMetadataSource.METADATA_CACHE_USED_SIZE,
+//                    registry, expectedCacheUsedSize);
```
jinggou, line 163: In this test, it’s hard to calculate `expectedCacheUsedSize` after altering table, because the size of newDeletedTableMarker(kv.getTimestamp()) is also included in the actual cache used size, and its value is hard to get, right? @virajjasani  WDYT?

virajjasani: Correct @jinggou, it is not required that we compare the exact cache used size metric, but what we can do is measure that the value of used cache is higher or lower from previous value.

phoenix-core-client/src/main/java/org/apache/phoenix/schema/metrics/MetricsMetadataSource.java
```diff
@@ -88,6 +88,24 @@ public interface MetricsMetadataSource {
     String DROP_FUNCTION_COUNT = "dropFunctionCount";
     String DROP_FUNCTION_COUNT_DESC = "Count of DROP FUNCTION DDL statements";
 
+    String METADATA_CACHE_USED_SIZE = "metadataCacheEstimatedUsedSize";
+    String METADATA_CACHE_USED_SIZE_DESC = "Estimated used size of the metadata cache";
```
virajjasani, line 92: nit: `METADATA_CACHE_ESTIMATED_USED_SIZE`

virajjasani:
@jinggou we also need client side metrics. For instance, we already have `CLIENT_METADATA_CACHE_HIT_COUNTER` and `CLIENT_METADATA_CACHE_MISS_COUNTER`. Now we can add remaining i.e. cache size used (estimated), eviction/removal and cache add metrics. Let's see how much we can add there, if it becomes too much work, we can do it on separate follow-up Jira also.

phoenix-core-client/src/main/java/org/apache/phoenix/schema/PMetaDataCache.java
```diff
@@ -52,6 +53,16 @@ public void onRemoval(RemovalNotification<PTableKey, PTableRef> notification) {
...
                                 + notification.getCause().name());
+                        if (notification.wasEvicted()) {
+                            GlobalClientMetrics.GLOBAL_CLIENT_METADATA_CACHE_EVICTION_COUNTER.increment();
+                        } else {
+                            GlobalClientMetrics.GLOBAL_CLIENT_METADATA_CACHE_REMOVAL_COUNTER.increment();
+                        }
+                        if (notification.getValue() != null) {
+                            long newSize =
+                                    GlobalClientMetrics.GLOBAL_CLIENT_METADATA_CACHE_ESTIMATED_USED_SIZE.getMetric().getValue() - notification.getValue().getEstimatedSize();
+                            GlobalClientMetrics.GLOBAL_CLIENT_METADATA_CACHE_ESTIMATED_USED_SIZE.update(newSize);
```
virajjasani, line 64: Can we do decrement here to be consistent with server side metrics computation?

jinggou: There is no `increment(long value)` method here as server side, but seems this `update(long value)` method is adding the given value to the current value. So the calculation here is wrong and we need to change to 
```
GlobalClientMetrics.GLOBAL_CLIENT_METADATA_CACHE_ESTIMATED_USED_SIZE.update(-notification.getValue().getEstimatedSize());
```
Is that okay?

virajjasani: Sounds good

virajjasani:
+1 for the core changes. Let's fix some of these checkstyle warnings?
- https://ci-hadoop.apache.org/job/Phoenix/job/Phoenix-PreCommit-GitHub-PR/job/PR-1976/6/PR_20General_20Check_20Report/
- https://ci-hadoop.apache.org/job/Phoenix/job/Phoenix-PreCommit-GitHub-PR/job/PR-1976/6/PR_20General_20Check_20Report/
- https://ci-hadoop.apache.org/job/Phoenix/job/Phoenix-PreCommit-GitHub-PR/job/PR-1976/6/PR_20General_20Check_20Report/

Could you also create backport PR against 5.2 branch? This is really nice change to get included with upcoming 5.2.1 release.
Thank you @jinggou!

PR #1977, opened 2024-09-13
PHOENIX-7362: Update owasp plugin to 10.0.2

szucsvillo:


stoty:
Thanks.

In the future, please do not put a colon after the JIRA ID, that's not the convention Phoenix uses.

PR #1978, opened 2024-09-13
PHOENIX-7406 : Fix partial index creation on a table which was created with column names in double quotes

palashc:

PR #1980, opened 2024-09-17
PHOENIX-7406 : Fix partial index creation on a table which was create…

palashc:
…d with column names in double quotes (#1978)


virajjasani (approved):
+1

PR #1985, opened 2024-09-25
PHOENIX-7411 Atomic Delete: PhoenixStatement API to return row if single row is atomically deleted

virajjasani:
Jira: PHOENIX-7411

PHOENIX-7398 introduces new PhoenixStatement API to return row for Atomic/Conditional Upserts. Phoenix already supports Upserts as Atomic operation by using ON DUPLICATE KEY clause. However, Deletes are not supported as atomic and single Delete query can also delete multiple rows based on the WHERE clause used.

HBase does not provide API to return row state for atomic updates with Put and Delete mutations. HBase API checkAndMutate() also supports returning Result for Append and Increment mutations only, not for Put and Delete mutations.

The purpose of this Jira is to introduce support for atomic delete of single row that can return the row only if it existed before executing the Delete mutation. If the row is already deleted, the API support is not expected to return any row. For single row delete to be atomic, IndexRegionObserver needs to take row lock for to be deleted row and also scan the row, similar to Atomic Put mutation(s).

 

PhoenixStatement API signature is same as of PHOENIX-7398:

```
/**
 *
 */
```

### DIFF
--- a/phoenix-core-client/src/main/java/org/apache/phoenix/jdbc/PhoenixStatement.java
+++ b/phoenix-core-client/src/main/java/org/apache/phoenix/jdbc/PhoenixStatement.java
@@ -647,13 +647,16 @@ public class PhoenixStatement implements PhoenixMonitoredStatement, SQLCloseable
                                         lastState.getUpdateCount());
                                 Result result = null;
                                 if (connection.getAutoCommit()) {
-                                    state.setReturnResult(returnResult);
+                                    if (isSingleRowUpdate(isUpsert, isDelete, plan)) {
+                                        state.setReturnResult(returnResult);
+                                    }
                                     connection.commit();
                                     if (isAtomicUpsert) {
                                         lastUpdateCount = connection.getMutationState()
                                                 .getNumUpdatedRowsForAutoCommit();
                                     }
                                     result = connection.getMutationState().getResult();
+                                    connection.getMutationState().clearResult();
                                 }
                                 setLastQueryPlan(null);
                                 setLastUpdateCount(lastUpdateCount);
@@ -753,6 +756,18 @@ public class PhoenixStatement implements PhoenixMonitoredStatement, SQLCloseable
             Throwables.propagate(e);
             throw new IllegalStateException(); // Can't happen as Throwables.propagate() always throws
         }
+    }
+
+    private static boolean isSingleRowUpdate(boolean isUpsert, boolean isDelete,
+                                             MutationPlan plan) {
+        boolean isSingleRowUpdate = false;
+        if (isUpsert) {
+            isSingleRowUpdate = true;
+        } else if (isDelete) {
+            isSingleRowUpdate =
+                    plan.getContext().getScanRanges().isPointLookup();
+        }
+        return isSingleRowUpdate;
     }
 
     protected static interface CompilableStatement extends BindableStatement {
--- a/phoenix-core-client/src/main/java/org/apache/phoenix/jdbc/PhoenixStatement.java
+++ b/phoenix-core-client/src/main/java/org/apache/phoenix/jdbc/PhoenixStatement.java
@@ -647,7 +647,7 @@ public class PhoenixStatement implements PhoenixMonitoredStatement, SQLCloseable
                                         lastState.getUpdateCount());
                                 Result result = null;
                                 if (connection.getAutoCommit()) {
-                                    if (isSingleRowUpdate(isUpsert, isDelete, plan)) {
+                                    if (isPointLookupPlan(isUpsert, isDelete, plan)) {
                                         state.setReturnResult(returnResult);
                                     }
                                     connection.commit();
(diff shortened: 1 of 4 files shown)


phoenix-core-client/src/main/java/org/apache/phoenix/jdbc/PhoenixStatement.java
```diff
@@ -755,6 +758,18 @@ public Pair<Integer, Tuple> call() throws SQLException {
...
     }
 
+    private static boolean isSingleRowUpdate(boolean isUpsert, boolean isDelete,
+                                             MutationPlan plan) {
+        boolean isSingleRowUpdate = false;
+        if (isUpsert) {
+            isSingleRowUpdate = true;
+        } else if (isDelete) {
+            isSingleRowUpdate =
+                    plan.getContext().getScanRanges().isPointLookup();
```
tkhurana, line 768: What about multi key point lookup ? It is possible for delete also using IN clause.

phoenix-core/src/it/java/org/apache/phoenix/end2end/OnDuplicateKey2IT.java
```diff
@@ -198,37 +199,85 @@ public void testReturnRowResult() throws Exception {
...
-                    .executeUpdateReturnRow(upsertSql);
-        }
-        assertEquals(success ? 1 : 0, resultPair.getFirst().intValue());
+        final Pair<Integer, Tuple> resultPair = ps.executeUpdateReturnRow();
         ResultTuple result = (ResultTuple) resultPair.getSecond();
+
         PTable table = conn.unwrap(PhoenixConnection.class).getTable(tableName);
 
+        if (!success && result.getResult() == null) {
+            return;
```
tkhurana, line 264: I couldn't understand when will the result be null ?

phoenix-core/src/it/java/org/apache/phoenix/end2end/OnDuplicateKey2IT.java
```diff
@@ -198,37 +199,85 @@ public void testReturnRowResult() throws Exception {
...
+        }
         Cell cell =
                 result.getResult()
                         .getColumnLatestCell(table.getColumns().get(3).getFamilyName().getBytes(),
                                 table.getColumns().get(3).getColumnQualifierBytes());
+        if (!success) {
+            assertNull(cell);
```
tkhurana, line 271: When will this happen ? The result is not null but the cell is null ?

phoenix-core-client/src/main/java/org/apache/phoenix/jdbc/PhoenixStatement.java
virajjasani: Even then we won't return any result because `MutationState` changes will ensure that only if we are creating PRow for Delete, we use result return attribute. Let me add a test to ensure multi point lookup with SkipScan filter is covered.

phoenix-core/src/it/java/org/apache/phoenix/end2end/OnDuplicateKey2IT.java
virajjasani: It will be null when Delete does not have strict point lookup.

phoenix-core/src/it/java/org/apache/phoenix/end2end/OnDuplicateKey2IT.java
virajjasani: This happens when single key point lookup Delete is successful but it does not really delete the row because the row did not exist in the first place when it tried to delete it atomically.

phoenix-core-client/src/main/java/org/apache/phoenix/jdbc/PhoenixStatement.java
tkhurana: Why will Delete use a SkipScan ?

tkhurana: I am talking about the case where multiple full row keys are provided. There shouldn't be any SkipScan.

virajjasani: Covered in the test, it still uses PRow for Delete only if single row is deleted with point lookup plan rather than multi key point lookup.

tkhurana: So in this case all the Delete mutations sent by the client won't have the return result attribute ?

virajjasani: That is correct. Only if single key point lookup is used with auto-commit and the new JDBC API, only then the attribute will be sent.

tkhurana (approved):

PR #1994, opened 2024-10-05
PHOENIX-7417 Remove commons-collections dependency

NihalJain:
- Remove imports and replace with equivalent code
- Add RestrictImports rule to restrict further usage of it in codebase

### DIFF
--- a/phoenix-core-client/src/main/java/org/apache/phoenix/util/InstanceResolver.java
+++ b/phoenix-core-client/src/main/java/org/apache/phoenix/util/InstanceResolver.java
@@ -17,10 +17,9 @@
  */
 package org.apache.phoenix.util;
 
-import org.apache.commons.collections4.IteratorUtils;
-
 import org.apache.phoenix.thirdparty.com.google.common.annotations.VisibleForTesting;
 
+import java.util.ArrayList;
 import java.util.Iterator;
 import java.util.List;
 import java.util.ServiceLoader;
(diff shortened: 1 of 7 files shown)


NihalJain:
We have same in query server also. Do we use same ticket for both PRs? Or raise new Jira for queryserver PR?

NihalJain:
Saw an import issue in code change. Went ahead and raised https://github.com/apache/phoenix/pull/1995/files as fixing manually is too cumbersome

stoty:
There is minimal functionality that we use here.

Wouldn't it be better to just get rid of this dependency ?

NihalJain:
> Wouldn't it be better to just get rid of this dependency ?

Sure let me see if I can remove the usage of commons-collections altogether

NihalJain:
> > Wouldn't it be better to just get rid of this dependency ?
> 
> Sure let me see if I can remove the usage of commons-collections altogether

Done!

phoenix-core-client/src/main/java/org/apache/phoenix/util/InstanceResolver.java
```diff
@@ -71,14 +70,22 @@ public static <T> T getSingleton(Class<T> clazz, T defaultInstance) {
...
+            defaultInstances.addAll(iteratorToList(iterator));
         } else {
-            defaultInstances = IteratorUtils.toList(iterator);
+            defaultInstances = iteratorToList(iterator);
         }
 
         return defaultInstances;
     }
 
+    private static <T> List<T> iteratorToList(Iterator<T> iterator) {
```
stoty, line 81: nit: While there is nothing wrong with this, here are a few solutions that do not need a new method:

https://stackoverflow.com/questions/10117026/convert-iterator-to-list

stoty: I mean the guava or lambd based ones.

stoty (approved):
+1 LGTM

You may want to address my nit, but it's fine either way.

phoenix-core-client/src/main/java/org/apache/phoenix/util/InstanceResolver.java
NihalJain: done

stoty (approved):
+1 LGTM

stoty:
please fix the checkstyle errors before commit (unless they are pre-existing)

NihalJain:
> please fix the checkstyle errors before commit (unless they are pre-existing)

/phoenix-core-client/src/main/java/org/apache/phoenix/util/QueryBuilder.java:150:    is new let me fix this

stoty:
Still +1

NihalJain:
Thank you @stoty I plan to fix this for other phoenix projects. PQS also depends on commons-collections. Raised PHOENIX-7430.

PR #1995, opened 2024-10-06
DRAFT: Code formatting using spotless in Phoenix

NihalJain:
This PR currently has 3 commits, each of which do independent tasks as follows:
- Add spotless plugin to format code in phoenix
  - Update dev/PhoenixCodeTemplate.xml to use latest format as in hbase. 
     - _Question to reviewers_: Also IMO we should rename this file to phoenix_eclipse_formatter.xml?
  - Fix file having misplaced package block as we do not want any manual code change in next commit where we run spotless
  -   Copied licnese-header from hbase, the one which phoenix was using was a little different
  -  _Question to reviewers_: Revisit the import order rules?
- Run spotless:apply to format code in phoenix: 
  - No manual code change in this commit
- Add spotless phoenix pre commit check

We could create 3 JIRA sub tasks and do each commit one at a time for ease of backporting.

TODO: Raise JIRAs as needed


pom.xml
```diff
@@ -668,6 +669,113 @@
...
+            <sortPom>
+              <expandEmptyElements>false</expandEmptyElements>
+            </sortPom>
+          </pom>
+          <formats>
+            <!-- you can define as many formats as you want, each is independent -->
+            <format>
+              <!-- define the files to apply to -->
+              <includes>
+                <include>**/*.xml</include>
```
NihalJain: include Dockerfile

dev/PhoenixCodeTemplate.xml
```diff
@@ -1,312 +1,418 @@
...
-<setting id="org.eclipse.jdt.core.formatter.insert_space_after_semicolon_in_for" value="insert"/>
-<setting id="org.eclipse.jdt.core.formatter.never_indent_line_comments_on_first_column" value="false"/>
-<setting id="org.eclipse.jdt.core.formatter.insert_space_before_semicolon_in_try_resources" value="do not insert"/>
-<setting id="org.eclipse.jdt.core.formatter.insert_space_before_opening_angle_bracket_in_type_arguments" value="do not insert"/>
-<setting id="org.eclipse.jdt.core.formatter.never_indent_block_comments_on_first_column" value="false"/>
-<setting id="org.eclipse.jdt.core.formatter.keep_then_statement_on_same_line" value="true"/>
-</profile>
-</profiles>
+<profiles version="22">
+    <profile kind="CodeFormatterProfile" name="'HBase Apache Current'" version="22">
```
NihalJain, line 22: rename to 'Phoenix Code Template' or 'Apache Phoenix Current'

NihalJain:
This code can be reviewed, just that not ready for merge. Need to add JIRA title and also decide whether should go as one commit or multi commit.

stoty:
There was already a DISCUSS thread on this back in Janury 2023.

I think that this is visible and high impact enough to warrant another [DISCUSS] thread.

I think we should do this, BTW.

pom.xml
```diff
@@ -167,6 +167,7 @@
     <exec-maven-plugin.version>3.1.1</exec-maven-plugin.version>
     <maven-checkstyle-plugin.version>3.3.0</maven-checkstyle-plugin.version>
     <mvel2.version>2.5.2.Final</mvel2.version>
+    <spotless.version>2.27.2</spotless.version>
```
lfrancke, line 162: This is a pretty old version (from 2022). Is there a reason you're not using the latest? (2.43.0)

lfrancke:
Thank you so much for picking this up. I just stumbled upon the [previous PR](https://github.com/apache/phoenix/pull/1548), [Jira issue](https://issues.apache.org/jira/browse/PHOENIX-6851) and the [previous mailing list thread](https://lists.apache.org/thread/9gbh5xs544h4q4w4xt64oy9osy8g8f81).

I do appreciate the work a lot and can try to help out where I can.

I also appreciate that you split the PR up in multiple commits.
If it's not too much work splitting this up in at least two PRs would help tremendously with reviewing it.
One that does all the "machinery" and another that runs the reformat etc.

NihalJain:
> There was already a DISCUSS thread on this back in Januray 2023, but nothing happened.
I think that this is visible and high impact enough to warrant another [DISCUSS] thread.
I think we should do this, BTW.

Sure @stoty Thank you for the reference. Let me chime into DISCUSS thread and put up what we plan to do here.


> Thank you so much for picking this up. I just stumbled upon the [previous PR](https://github.com/apache/phoenix/pull/1548), [Jira issue](https://issues.apache.org/jira/browse/PHOENIX-6851) and the [previous mailing list thread](https://lists.apache.org/thread/9gbh5xs544h4q4w4xt64oy9osy8g8f81).

Thank you @lfrancke for bringing this up again on DISCUSS, and for pointing out the JIRA for previous issue.

> I do appreciate the work a lot and can try to help out where I can.

Sure will be great if you can help with reviews. Will let you know if anything else if needed.

> I also appreciate that you split the PR up in multiple commits.
If it's not too much work splitting this up in at least two PRs would help tremendously with reviewing it.
One that does all the "machinery" and another that runs the reformat etc.

I plan to raise 3 PRs . This PR was just to demo the level of inconsistency we have in the codebase and seek feedback. Let me create appropriate JIRAs and raise 3 distinct PRs.

1. Add spotless plugin to format code in phoenix
2. Run spotless:apply to format code in phoenix
3. Add spotless phoenix pre commit check

Also need to break the "machinery" part into tasks: task 1 and task 3 as we may want to add to pre-commit only after all issues are fixed with task2. This way we can have ease of backporting as well since task 2 generated code may differ for all branches.

pom.xml
NihalJain: Ah yes you are right. I just kept this in sync with hbase but yes let me bump into hbase codebase as well and move to the latest one over here as well.

NihalJain:
work superceded at https://github.com/apache/phoenix/pull/2023

PR #1996, opened 2024-10-08
PHOENIX-7418 SystemExitRule errors out because of SecurityManager dep…

stoty:
…recation / removal

run tests with -Djava.security.manager=allow
Catch and log SecurityManager exceptions in SystemExitRule


gjacoby126 (approved):
+1, after reading the discussion here: https://openjdk.org/jeps/411 and here: https://bugs.openjdk.org/browse/JDK-8199704 . Note though that at some point the security manager will be fully removed and a different solution will need to be found.

PR #2002, opened 2024-10-12
Support masking for Conditional TTL

tkhurana:


jpisaac (approved):
LGTM +1

jpisaac:
@tkhurana We should consider converting the regular TTL masking to use the query rewrite mechanism. Just a thought, but that can be a separate JIRA

phoenix-core-client/src/main/java/org/apache/phoenix/compile/QueryCompiler.java
```diff
@@ -787,6 +790,33 @@ protected QueryPlan compileSingleFlatQuery(
         TableRef tableRef = context.getCurrentTable();
         PTable table = tableRef.getTable();
 
+        TTLExpression ttlExpr = table.getTTL();
```
tkhurana, line 793: @comnetwork Looking for feedback if this is the right way to re-write the where clause by appending a boolean expression. In my testing it is working so far.

PR #2009, opened 2024-10-16
PHOENIX-7427 PHOENIX-7418 breaks backwards compatibility tests

stoty:


stoty:
There is no universal java.security.value that works with all Java versions, so we run a dummy command, and only add the property if it doesn't cause errors.

On the JVM versions where the allow value is not accepted, it is not needed either.

stoty:
Yetus does not run the affected tests.
See #2010 for the actual test results.

stoty:
As seen on #2010, this does fix the BackwardCompatibility failures.

virajjasani (approved):
+1

PR #2013, opened 2024-10-18
PHOENIX-7310 : Add block bytes to scan metrics

palashc:


phoenix-core/src/it/java/org/apache/hadoop/hbase/regionserver/BlockBytesScannedMetricIT.java
```diff
@@ -0,0 +1,159 @@
...
+    }
+
+    private void clearBlockCache(String tableName) {
+        HRegionServer regionServer = utility.getMiniHBaseCluster().getRegionServer(0);
+        for (HRegion region : regionServer.getRegions(TableName.valueOf(tableName))) {
+            regionServer.clearRegionBlockCache(region);
+        }
+    }
+
+    private void flush(String tableName) throws IOException {
```
tkhurana, line 122: I think we already have a utility flush method in TestUtil class.

phoenix-core/src/it/java/org/apache/hadoop/hbase/regionserver/BlockBytesScannedMetricIT.java
```diff
@@ -0,0 +1,159 @@
...
+        // where clause has non PK column, will have to scan all rows
+        Assert.assertTrue(count2 > count1);
+
+        long count3 = countBlockBytesScannedFromSql(stmt, SELECT_ALL_QUERY);
+        Assert.assertTrue(count3 > 0);
+        // should be same as previous query which also scans all rows
+        Assert.assertEquals(count3, count2);
+    }
+
+    private void clearBlockCache(String tableName) {
```
tkhurana, line 115: I would add this method to the TestUtil class.

palashc:
This change needs hbase-2.6 as default.

PR #2015, opened 2024-10-22
PHOENIX-7425 Partitioned CDC Index for eliminating salting

kadirozde:
The CDC future (PHOENX-7001) uses a global uncovered time (PHOENIX_ROW_TIMESTAM()) based index. Such an index will likely create hot spotting during writes. This is because the same index region will keep getting updated as the row key of the index table is  PHOENIX_ROW_TIMESTAMP() + data table row key.

The same hot spotting can happen during reads as a small subset of index regions can be used for a given time range. For example, the most recent changes will be retrieved through one or two index regions.

To address these hot spotting issues, PHOENX-7001 suggests salting the index. There are three main issues with salting.

The first one is that the number of salt buckets is static and needs to be determined when the index is created.

The second is that salting does not work well with batch writes as it results in breaking a batch of writes into separate mini batches, one for each salt bucket. This leads to using more client threads and server RPC handlers, one for each salt bucket.

The last issue is that the salt buckets are not visible to applications and thus they cannot take advantage of the parallelism that comes with salting during reads. For example, there is no way for applications to use multiple threads, one thread for each salt bucket, for their queries.

To address all these issues that come with salting, this PR introduces a built-in function for CDC indexes called PARTITION_ID(). PARTITION_ID() will be the prefix of an index row key (= PARTITION_ID() + PHOENIX_ROW_TIMESTAMP() + data table row key). PARTITION_ID() will identify the data table region of the data table row key. PARTITION_ID() can be the encoded name of the data table region.

Like PHOENIX_ROW_TIMESTAMP(), PARTITION_ID() can be used in CDC index queries.

By including PARTITION_ID() in the row key of an index table, we essentially create the effect of local index such that all index mutations for a given data table region are written to one index region determined by the PARITION_ID(). However, here we will not have the local index problem with region splits where copying index rows during data table region splits is required.

### DIFF
--- a/phoenix-core-client/src/main/java/org/apache/phoenix/expression/function/PartitionIdFunction.java
+++ b/phoenix-core-client/src/main/java/org/apache/phoenix/expression/function/PartitionIdFunction.java
@@ -29,8 +29,8 @@ import org.apache.phoenix.schema.types.PDataType;
 import java.util.List;
 
 /**
- * Function to return the partition id which is the encoded data table region name. This function
- * is used only with CDC Indexes
+ * Function to return the partition id which is the encoded data table region name as the prefix
+ * of the CDC index row key. This function is used only with CDC Indexes
  */
 @BuiltInFunction(name = PartitionIdFunction.NAME,
         nodeClass= PartitionIdParseNode.class,
@@ -70,7 +70,6 @@ public class PartitionIdFunction extends ScalarFunction {
      */
     @Override
     public boolean evaluate(Tuple tuple, ImmutableBytesWritable ptr) {
-
         if (tuple == null) {
             return false;
         }
@@ -78,6 +77,7 @@ public class PartitionIdFunction extends ScalarFunction {
         if (ptr.getLength() < PARTITION_ID_LENGTH) {
             return false;
         }
+        // The partition id of a row is always the prefix of the row key
         ptr.set(ptr.get(), 0, PARTITION_ID_LENGTH);
         return true;
     }
--- a/phoenix-core-client/src/main/java/org/apache/phoenix/expression/function/PartitionIdFunction.java
+++ b/phoenix-core-client/src/main/java/org/apache/phoenix/expression/function/PartitionIdFunction.java
@@ -1,0 +1,110 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.phoenix.expression.function;
+
+import org.apache.hadoop.hbase.io.ImmutableBytesWritable;
+import org.apache.phoenix.expression.Determinism;
+import org.apache.phoenix.expression.Expression;
+import org.apache.phoenix.parse.FunctionParseNode.BuiltInFunction;
+import org.apache.phoenix.parse.PartitionIdParseNode;
+import org.apache.phoenix.schema.tuple.Tuple;
+import org.apache.phoenix.schema.types.PDataType;
+import org.apache.phoenix.schema.types.PVarchar;
+
+import java.util.List;
+
+import static org.apache.phoenix.query.QueryConstants.SEPARATOR_BYTE;
+
+/**
+ * Function to return the partition id which is the encoded data table region name. This function
+ * is used only with CDC Indexes
+ */
+@BuiltInFunction(name = PartitionIdFunction.NAME,
+        nodeClass= PartitionIdParseNode.class,
+        args = {})
+public class PartitionIdFunction extends ScalarFunction {
+    public static final String NAME = "PARTITION_ID";
+
+    public PartitionIdFunction() {
+    }
+
+    /**
+     *  @param children none
+     *  {@link org.apache.phoenix.parse.PartitionIdParseNode#create create}
+     *  will return the partition id of a given CDC index row.
+     */
+    public PartitionIdFunction(List<Expression> children) {
+        super(children);
+        if (!children.isEmpty()) {
+            throw new IllegalArgumentException(
+                    "PartitionIdFunction should not have any child expression"
+            );
+        }
+    }
+
+    @Override
+    public String getName() {
+        return NAME;
+    }
+
+    /**
+     * The evaluate method is called under the following conditions -
+     * 1. When PARTITION_ID() is evaluated in the projection list.
+     *
+     * 2. When PARTITION_ID() is evaluated in the backend as part of the where clause.
+     *
+     */
+    @Override
+    public boolean evaluate(Tuple tuple, ImmutableBytesWritable ptr) {
+
+        if (tuple == null) {
+            return false;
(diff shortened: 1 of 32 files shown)


phoenix-core-client/src/main/java/org/apache/phoenix/index/IndexMaintainer.java
```diff
@@ -183,7 +185,8 @@ public static Iterator<PTable> maintainedGlobalIndexesWithMatchingStorageScheme(
             @Override
             public boolean apply(PTable index) {
                 return sendIndexMaintainer(index) && IndexUtil.isGlobalIndex(index)
-                        && dataTable.getImmutableStorageScheme() == index.getImmutableStorageScheme();
+                        && (dataTable.getImmutableStorageScheme() == index.getImmutableStorageScheme()
+                        && !CDCUtil.isCDCIndex(index));
```
kadirozde, line 189: This is not necessary (reviewed by Tanuj)

phoenix-core-client/src/main/java/org/apache/phoenix/index/IndexMaintainer.java
```diff
@@ -706,10 +709,14 @@ public void setDataEncodingScheme(QualifierEncodingScheme sc) {
         this.dataEncodingScheme = sc;
     }
 
-    public byte[] buildRowKey(ValueGetter valueGetter, ImmutableBytesWritable rowKeyPtr, byte[] regionStartKey, byte[] regionEndKey, long ts)  {
+    public byte[] buildRowKey(ValueGetter valueGetter, ImmutableBytesWritable rowKeyPtr,
```
kadirozde, line 712: Do not change existing public API (reviewed by Tanuj)

phoenix-core-client/src/main/java/org/apache/phoenix/expression/function/PartitionIdFunction.java
```diff
@@ -0,0 +1,105 @@
...
+import org.apache.phoenix.parse.PartitionIdParseNode;
+import org.apache.phoenix.schema.tuple.Tuple;
+import org.apache.phoenix.schema.types.PChar;
+import org.apache.phoenix.schema.types.PDataType;
+
+import java.util.List;
+
+/**
+ * Function to return the partition id which is the encoded data table region name. This function
+ * is used only with CDC Indexes
```
kadirozde, line 33: Add comment to state that the partition id is always the prefix of the row key. This is also assumed in the evaluate method. (By Tanuj)

phoenix-core-client/src/main/java/org/apache/phoenix/index/IndexMaintainer.java
```diff
@@ -785,7 +792,11 @@ public byte[] buildRowKey(ValueGetter valueGetter, ImmutableBytesWritable rowKey
                 	dataColumnType = expression.getDataType();
                 	dataSortOrder = expression.getSortOrder();
                     isNullable = expression.isNullable();
-                	expression.evaluate(new ValueGetterTuple(valueGetter, ts), ptr);
+                    if (expression instanceof PartitionIdFunction) {
```
kadirozde, line 795: Add check for ensuring that the partition id expression is the prefix of the index table row key (by Tanuj)

phoenix-core-client/src/main/java/org/apache/phoenix/index/IndexMaintainer.java
```diff
@@ -1216,8 +1247,11 @@ private RowKeySchema getIndexRowKeySchema() {
         return indexRowKeySchema;
     }
 
-    public Put buildUpdateMutation(KeyValueBuilder kvBuilder, ValueGetter valueGetter, ImmutableBytesWritable dataRowKeyPtr, long ts, byte[] regionStartKey, byte[] regionEndKey, boolean verified) throws IOException {
```
kadirozde, line 1219: Do not change the existing method signature here (by Tanuj)

phoenix-core-client/src/main/java/org/apache/phoenix/optimize/QueryOptimizer.java
```diff
@@ -283,13 +283,17 @@ private List<QueryPlan> getApplicablePlansForSingleFlatQuery(QueryPlan dataPlan,
             plans.add(dataPlan);
             hintedPlan = getHintedQueryPlan(statement, translatedIndexSelect, indexes,
                     targetColumns, parallelIteratorFactory, plans);
-            if (hintedPlan != null) {
+            if (hintedPlan != null ) {
                 PTable index = hintedPlan.getTableRef().getTable();
-                if (stopAtBestPlan && hintedPlan.isApplicable() && (index.getIndexWhere() == null
-                        || isPartialIndexUsable(select, dataPlan, index))) {
-                    return Collections.singletonList(hintedPlan);
+                // Ignore any index hint with a CDC index
```
kadirozde, line 288: As it is a truncated index, it only includes index rows for the data table rows that have changed within the max lookback window.

phoenix-core-server/src/main/java/org/apache/phoenix/coprocessor/BaseScannerRegionObserver.java
```diff
@@ -160,6 +163,11 @@ public void preScannerOpen(org.apache.hadoop.hbase.coprocessor.ObserverContext<R
             TimeRange timeRange = scan.getTimeRange();
             scan.setTimeRange(timeRange.getMin(), Bytes.toLong(txnScn));
         }
+        byte[] cdcScan = scan.getAttribute(CDC_DATA_TABLE_DEF);
+        if (cdcScan != null) {
```
kadirozde, line 167: Check if the scan is already adjusted (by checking if the scan is already changed to a raw scan.

phoenix-core-server/src/main/java/org/apache/phoenix/coprocessor/UncoveredIndexRegionScanner.java
```diff
@@ -299,21 +299,28 @@ private boolean verifyIndexRowAndRepairIfNecessary(Result dataRow, byte[] indexR
...
-                Put indexPut = new Put(indexRowKey);
-                indexPut.addColumn(emptyCF, emptyCQ, indexTimestamp, QueryConstants.VERIFIED_BYTES);
-                if ((EnvironmentEdgeManager.currentTimeMillis() - indexTimestamp) > ageThreshold) {
-                    Delete indexDelete = indexMaintainer.buildRowDeleteMutation(indexRowKey,
-                            IndexMaintainer.DeleteType.SINGLE_VERSION, indexTimestamp);
-                    mutations = new Mutation[]{indexPut, indexDelete};
-                } else {
-                    mutations = new Mutation[]{indexPut};
+        if (indexMaintainer.isCDCIndex()) {
+            if (indexMaintainer.checkIndexRow(indexRowKey, dataRow.getRow(), put, viewConstants)
```
kadirozde, line 303: Do we really need to call checkIndexRow as it will always return true since the index row does not include any non PK data table columns?

phoenix-core/src/it/java/org/apache/phoenix/end2end/CDCQueryIT.java
```diff
@@ -135,6 +140,117 @@ private void cdcIndexShouldNotBeUsedForDataTableQueries(Connection conn, String
...
+            return true;
+        } else if (filter instanceof FilterList) {
+                return isDistinctPrefixFilterIncludedInFilterList((FilterList) filter);
+        }
+        return false;
+    }
+
+    private void checkIndexPartitionIdCount(Connection conn, String cdcName) throws Exception {
+        // Verify that we can use retrieve partition ids
+        ResultSet rs = conn.createStatement().executeQuery("SELECT PARTITION_ID() FROM "
```
kadirozde, line 166: Add ORDER BY PARTITION_ID() to make sure rows are returned orderly here (by Tanuj)

phoenix-core-client/src/main/java/org/apache/phoenix/index/IndexMaintainer.java
kadirozde: It turns out that it was required.

tkhurana (approved):

PR #2018, opened 2024-10-25
PHOENIX-7449 Handling already disabled table gracefully during dropTables execution

ritegarg:
See jira for details.


phoenix-core-client/src/main/java/org/apache/phoenix/query/ConnectionQueryServicesImpl.java
```diff
@@ -2249,7 +2249,9 @@ private boolean ensureViewIndexTableDropped(byte[] physicalTableName, long times
                     final ReadOnlyProps props = this.getProps();
                     final boolean dropMetadata = props.getBoolean(DROP_METADATA_ATTRIB, DEFAULT_DROP_METADATA);
                     if (dropMetadata) {
-                        admin.disableTable(physicalIndexTableName);
+                        if (admin.isTableEnabled(physicalIndexTableName)) {
+                            admin.disableTable(physicalIndexTableName);
```
tkhurana, line 2253: There is a race condition here because there is a window between the check if the table is enabled and the actual disable table call. It would be better if you make the disable table call itself resilient and gracefully handle the case where the table is already disabled. Calling disable on a table that is already disabled should be a no-op.

ritegarg: Updated

tkhurana: I was thinking we could have a disableTable function in the ConnectionQueryServicesImpl class which could internally handle this exception. Something like `disableTable(Admin, Tablename)`. I did a quick check there are 5 places in this file where we are doing disableTable. We could call this function everywhere.

tkhurana (approved):

tkhurana:
Only flappers failed, merging

PR #2020, opened 2024-10-29
PHOENIX-7424 Support toStringBinary/toBytesBinary conversion

szucsvillo:


phoenix-core-client/src/main/java/org/apache/phoenix/expression/function/DecodeBinaryFunction.java
```diff
@@ -0,0 +1,138 @@
...
+import org.apache.phoenix.exception.SQLExceptionInfo;
+import org.apache.phoenix.expression.Expression;
+import org.apache.phoenix.parse.FunctionParseNode;
+import org.apache.phoenix.schema.IllegalDataException;
+import org.apache.phoenix.schema.types.PDataType;
+import org.apache.phoenix.schema.types.PVarchar;
+import org.apache.phoenix.schema.types.PVarbinary;
+import org.apache.phoenix.schema.tuple.Tuple;
+
+/**
```
stoty, line 20: You have the names mixed up.

Encoding is BINARY -> STRING

Decoding is STRING -> BINARY

See the semantics in DecodeFunction.

phoenix-core-client/src/main/java/org/apache/phoenix/expression/function/EncodeBinaryFunction.java
```diff
@@ -0,0 +1,124 @@
...
+ * Usage:
+ * ENCODE_BINARY(<input_string>, <format>)
+ *
+ * @param input_string The string to be encoded.
+ * @param format       The encoding format (HEX, BASE64, HBASE).
+ */
+@FunctionParseNode.BuiltInFunction(name = EncodeBinaryFunction.NAME, args = {
+        @FunctionParseNode.Argument(allowedTypes = {PVarchar.class}),
+        @FunctionParseNode.Argument(enumeration = "EncodeFormat")})
+public class EncodeBinaryFunction extends ScalarFunction {
```
stoty, line 30: This is derived from and backward compatible with DecodeFunction.

(After fixing the names) instead of duplicating the code, they should share the same code via inheritance.
Adding the new functionality to the existing DECODE function does not break functionality, so DECODE and DECODE_BINARY should be basically aliases to each other.

(Unless the annotations somehow prevent inheritance from working properly)

stoty (changes requested):
Apart from using the wrong names, the functionality looks good.

phoenix-core-client/src/main/java/org/apache/phoenix/expression/function/EncodeFormat.java
```diff
@@ -22,5 +22,6 @@ public enum EncodeFormat {
 	HEX, //format for encoding HEX value to bytes
 	BASE62, //format for encoding a base 10 long value to base 62 string
 	BASE64, //format for encoding a base 10 long value to base 64 string
-	ASCII // Plain Text
+	ASCII, //Plain Text
+	HBASE //HBase-specific binary encoding format
```
stoty, line 26: nit: this is more of an escaping format than encoding.

phoenix-core-client/src/main/java/org/apache/phoenix/expression/function/DecodeBinaryFunction.java
```diff
@@ -0,0 +1,128 @@
...
+ * Usage:
+ * DECODE_BINARY(<input_string>, <format>)
+ *
+ * @param input_string The string to be decoded.
+ * @param format       The encoding format (HEX, BASE64, HBASE).
+ */
+@FunctionParseNode.BuiltInFunction(name = DecodeBinaryFunction.NAME, args = {
+        @FunctionParseNode.Argument(allowedTypes = {PVarchar.class}),
+        @FunctionParseNode.Argument(enumeration = "EncodeFormat")})
+public class DecodeBinaryFunction extends DecodeFunction {
```
stoty, line 45: This leaves us with a less capable DECODE() function, and the more capable DECODE_BINARY() function.

Since we are only extending the functionality, we could use the exact same implementation, adding more features to DECODE() is not a problem.

The only reason we are adding new functions at all is the "interesting" existing ENCODE() function.

stoty: Technically, we do not even need the new DECODE_BINARY function, we are only adding it so that the encoding/decoding functions have intuitive names.

phoenix-core-client/src/main/java/org/apache/phoenix/expression/function/DecodeFunction.java
```diff
@@ -1,11 +1,10 @@
 /*
- * Licensed to the Apache Software Foundation (ASF) under one
- * or more contributor license agreements.  See the NOTICE file
- * distributed with this work for additional information
- * regarding copyright ownership.  The ASF licenses this file
- * to you under the Apache License, Version 2.0 (the
- * "License"); you may not use this file except in compliance
- * with the License.  You may obtain a copy of the License at
+ * Licensed to the Apache Software Foundation (ASF) under one or more
```
stoty, line 2: Is this the canonical header ?
I guess the tests will complain if not.

phoenix-core-client/src/main/java/org/apache/phoenix/expression/function/DecodeFunction.java
```diff
@@ -71,19 +71,25 @@ public boolean evaluate(Tuple tuple, ImmutableBytesWritable ptr) {
...
-				break;
-			default:
-				throw new IllegalDataException("Unsupported encoding \"" + encoding + "\"");
-		}
+        switch (format) {
+            case HEX:
+                out = decodeHex(stringToDecode);
+                break;
+            case BASE64:
+				out = Base64.getDecoder().decode(stringToDecode);
```
stoty, line 85: whitespaces

phoenix-core-client/src/main/java/org/apache/phoenix/expression/function/EncodeBinaryFunction.java
```diff
@@ -0,0 +1,123 @@
...
+import org.apache.phoenix.expression.Expression;
+import org.apache.phoenix.parse.FunctionParseNode;
+import org.apache.phoenix.schema.IllegalDataException;
+import org.apache.phoenix.schema.types.PDataType;
+import org.apache.phoenix.schema.types.PVarchar;
+import org.apache.phoenix.schema.types.PVarbinary;
+import org.apache.phoenix.schema.tuple.Tuple;
+
+/**
+ * Encodes binary data into a string from various formats such as Hex, Base64, or HBase binary.
```
stoty, line 36: to, instead of from

phoenix-core-client/src/main/java/org/apache/phoenix/expression/function/EncodeBinaryFunction.java
```diff
@@ -0,0 +1,123 @@
...
+        String encodingFormat = ((String) type.toObject(ptr)).toUpperCase();
+        EncodeFormat format = EncodeFormat.valueOf(encodingFormat);
+        String encodedString;
+
+        switch (format) {
+            case HEX:
+                encodedString = toHex(bytesToEncode);
+                break;
+            case BASE64:
+                encodedString = Base64.getEncoder().encodeToString(bytesToEncode);
```
stoty, line 83: This uses Basic, which should be fine.

stoty: How heavy is Base64.getEncoder() ?
Would storing a static instance be a reasonable optimization ?

phoenix-core/src/it/java/org/apache/phoenix/end2end/DecodeBinaryFunctionIT.java
```diff
@@ -0,0 +1,202 @@
...
+
+@Category(ParallelStatsDisabledTest.class)
+public class DecodeBinaryFunctionIT extends ParallelStatsDisabledIT {
+
+    @Test
+    public void testDecodeHBaseFromEscapedHex() throws Exception {
+        Connection conn = DriverManager.getConnection(getUrl());
+
+        String testTable = generateUniqueName();
+        String ddl = "CREATE TABLE " + testTable + " (id INTEGER PRIMARY KEY, data VARCHAR)";
```
stoty, line 43: creating a separate table for each test will slow the tests down.
Consider creating a single table in @BeforeClass and re-using it.

phoenix-core/src/it/java/org/apache/phoenix/end2end/DecodeBinaryFunctionIT.java
```diff
@@ -0,0 +1,202 @@
...
+
+    @Test
+    public void testLongBase64Decoding() throws Exception {
+        Connection conn = DriverManager.getConnection(getUrl());
+
+        String testTable = generateUniqueName();
+        String ddl = "CREATE TABLE " + testTable + " (id INTEGER PRIMARY KEY, data VARCHAR)";
+        conn.createStatement().execute(ddl);
+
+        String base64Chunk = "SGVsbG9QaG9lbml4";
```
stoty, line 119: This works because we do not disable padding.

phoenix-core/src/it/java/org/apache/phoenix/end2end/DecodeBinaryFunctionIT.java
```diff
@@ -0,0 +1,202 @@
...
+        PreparedStatement ps =
+                conn.prepareStatement("UPSERT INTO " + testTable + " (id, data) VALUES (1, '48656C6C6F')");
+        ps.execute();
+        conn.commit();
+
+        ResultSet rs = conn.createStatement()
+                .executeQuery("SELECT ENCODE_BINARY(DECODE_BINARY(data, 'HEX'), 'HEX') FROM " + testTable);
+        assertTrue(rs.next());
+        String actualString = rs.getString(1);
+        assertEquals("48656c6c6f", actualString);
```
stoty, line 159: nit:Could use a constant on both ends.

phoenix-core/src/it/java/org/apache/phoenix/end2end/EncodeBinaryFunctionIT.java
```diff
@@ -0,0 +1,256 @@
...
+
+        PreparedStatement ps = conn.prepareStatement(
+                "UPSERT INTO " + testTable + " (id, data) VALUES (1, X'48656c6c6f50686f656e6978')");
+        ps.execute();
+        conn.commit();
+
+        ResultSet rs = conn.createStatement().executeQuery("SELECT ENCODE_BINARY(data, 'HEX') FROM " + testTable);
+        assertTrue(rs.next());
+        String actualString = rs.getString(1);
+        assertEquals("48656c6c6f50686f656e6978", actualString);
```
stoty, line 56: nit: use a java constant for these kinds of tests, with toLower() toUpper() if needed

phoenix-core-client/src/main/java/org/apache/phoenix/expression/function/EncodeBinaryFunction.java
szucsvillo: It is a standard Base64 encoder which is not particularly heavy as the encoding logic itself is relatively lightweight. Since the encoder is stateless, it does not gain much from being reused. I think the performance gains from this optimization would be minimal in most cases.

phoenix-core/src/it/java/org/apache/phoenix/end2end/DecodeBinaryFunctionIT.java
```diff
@@ -0,0 +1,202 @@
...
+        String ddl = "CREATE TABLE " + testTable + " (id INTEGER PRIMARY KEY, data VARCHAR)";
+
+        conn.createStatement().execute(ddl);
+
+        PreparedStatement ps = conn.prepareStatement("UPSERT INTO " + testTable + " (id, data) VALUES (1, 'data')");
+        ps.execute();
+        conn.commit();
+
+        ResultSet rs = conn.createStatement().executeQuery(
+                "SELECT DECODE_BINARY('\\\\x48\\\\x65\\\\x6C\\\\x6C\\\\x6F\\\\x50" +
```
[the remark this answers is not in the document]
szucsvillo: I wasn't entirely sure whether we needed both tests for fully escaped strings and hex literals. They both cover similar cases of encoding, so I've removed the duplicate test.

phoenix-core/src/it/java/org/apache/phoenix/end2end/DecodeBinaryFunctionIT.java
szucsvillo: You're right. I will also add a separate test case to verify Base64 decoding with padding.

phoenix-core-client/src/main/java/org/apache/phoenix/expression/function/EncodeBinaryFunction.java
stoty: Yes, it looks like a simple getter for a class member.

stoty: It's probably inlined anyway

phoenix-core/src/it/java/org/apache/phoenix/end2end/DecodeBinaryFunctionIT.java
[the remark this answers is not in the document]
stoty: I still don't get why this works.
I would expect this to result in \H\e\l\l\o\P\h\o\e\n\i\x

Why does the DECODE_BINARY swallow the extra backslashes ?

phoenix-core-client/src/main/java/org/apache/phoenix/expression/function/DecodeFunction.java
```diff
@@ -59,7 +59,7 @@ public boolean evaluate(Tuple tuple, ImmutableBytesWritable ptr) {
 			return true; // expression was evaluated, but evaluated to null
 		}
 
-		String stringToDecode = (String) PVarchar.INSTANCE.toObject(ptr);
+        String stringToDecode = (String) PVarchar.INSTANCE.toObject(ptr);
```
stoty, line 62: The indentation seems to be off.

Make sure to configure your IDE with the Phoenix standard code formatting setting
(then again, maybe this file is just misformatted.

In any case, try to minimize the whitespace changes, we're going to reformat the whole code base soon.

szucsvillo: I tried to follow that approach, however, the static code analyzer flagged some formatting issues, which I addressed to avoid any errors in the process.

PR #2022, opened 2024-11-04
PHOENIX-7449 Handling already disabled table gracefully during dropTables execution

ritegarg:


tkhurana:
@ritegarg Can you look at the compilation errors

PR #2025, opened 2024-11-09
PHOENIX-7458 : Create new SYSTEM tables for tracking CDC Stream metadata

palashc:


phoenix-core-client/src/main/java/org/apache/phoenix/query/QueryConstants.java
```diff
@@ -640,4 +650,35 @@ enum JoinType {INNER, LEFT_OUTER}
...
+
+    String CREATE_CDC_STREAM_METADATA = "CREATE TABLE " + SYSTEM_CATALOG_SCHEMA + ".\"" +
+            SYSTEM_CDC_STREAM_TABLE + "\"(\n" +
+            // PK columns
+            TABLE_NAME + " VARCHAR NOT NULL," +
+            STREAM_NAME + " VARCHAR NOT NULL," +
+            PARTITION_ID + " VARCHAR NOT NULL," +
+            // Non-PK columns
+            PARENT_PARTITION_ID + " VARCHAR," +
+            PARTITION_START_TIME + " BIGINT," +
```
palashc, line 675: @virajjasani Does BIGINT make sense for start/end time since we can use (region_id) which is timestamp that we plan to use as sequence number? Or does it have to be Date as per the design doc?

virajjasani: I see, in this case we can use LONG directly instead of DATE.

virajjasani (commented):
Overall looks good otherwise

virajjasani (approved):
+1

palashc:
There seems to be one test failure `ConnectionQueryServicesMetricsIT.testMultipleCQSIMetricsInParallel`, will look into that.

palashc:
@virajjasani I am hitting some timing related issues in some tests on this PR due to TaskRegionObserver running in the background. The connection created by it causes some tests to flap, depends on when it runs relative to the test code. I fixed `ConnectionQueryServicesMetricsIT` by disabling TaskRegionObserver. Now `ConfigurableCacheIT` fails but I cannot override the setup for that test to disable TaskRegionObserver. What should be done in this case?

virajjasani:
You can use configs:

```
props.put(QueryServices.TASK_HANDLING_INITIAL_DELAY_MS_ATTRIB, Long.toString(Long.MAX_VALUE));
```

palashc:
Latest test report:
https://ci-hadoop.apache.org/job/Phoenix/job/Phoenix-PreCommit-GitHub-PR/job/PR-2025/11/testReport/

PR #2027, opened 2024-11-15
PHOENIX-7464 :- Performance Regression in Connection Initialization due to Configuration Handling in ConnInfo

lokiore:
Creating this PR for one of the solution of JIRA (Caching configuration at ConnectionInfo level as each CQSI can have different confs)

### DIFF
--- a/phoenix-core-client/src/main/java/org/apache/phoenix/jdbc/ConnectionInfo.java
+++ b/phoenix-core-client/src/main/java/org/apache/phoenix/jdbc/ConnectionInfo.java
@@ -493,15 +493,8 @@ public abstract class ConnectionInfo {
                                 LOGGER.info("Trying to connect to a secure cluster as {} "
                                         + "with keytab {}",
                                     principal, keytab);
-                                Configuration newConfig = HBaseFactoryProvider.getConfigurationFactory()
-                                        .getConfiguration();
+                                final Configuration newConfig = getConfiguration(principal, keytab);
                                 // We are intentionally changing the passed in Configuration object
-                                if (null != principal) {
-                                    newConfig.set(QueryServices.HBASE_CLIENT_PRINCIPAL, principal);
-                                }
-                                if (null != keytab) {
-                                    newConfig.set(QueryServices.HBASE_CLIENT_KEYTAB, keytab);
-                                }
                                 UserGroupInformation.setConfiguration(newConfig);
                                 User.login(newConfig, QueryServices.HBASE_CLIENT_KEYTAB,
                                     QueryServices.HBASE_CLIENT_PRINCIPAL, null);
@@ -521,6 +514,28 @@ public abstract class ConnectionInfo {
             } else {
                 LOGGER.debug("Principal and keytab not provided, not attempting Kerberos login");
             }
+        }
+
+        private Configuration getConfiguration( String principal, String keytab) {
+            final Configuration config = HBaseFactoryProvider.getConfigurationFactory().getConfiguration();
+            // Add QueryServices properties
+            for (Entry<String,String> entry : props) {
+                config.set(entry.getKey(), entry.getValue());
+            }
+            // Add any user-provided properties (via DriverManager)
+            if (info != null) {
+                for (Object key : info.keySet()) {
+                    config.set((String) key, info.getProperty((String) key));
+                }
+            }
+            // Set the principal and keytab if provided from the URL (overriding those provided in Properties)
+            if (null != principal) {
+                config.set(QueryServices.HBASE_CLIENT_PRINCIPAL, principal);
+            }
+            if (null != keytab) {
+                config.set(QueryServices.HBASE_CLIENT_KEYTAB, keytab);
+            }
+            return config;
         }
 
         protected String normalizeHostsList(String quorum, Integer defaultPort)
--- a/phoenix-core-client/src/main/java/org/apache/phoenix/jdbc/ConnectionInfo.java
+++ b/phoenix-core-client/src/main/java/org/apache/phoenix/jdbc/ConnectionInfo.java
@@ -59,6 +59,14 @@ public abstract class ConnectionInfo {
     protected static final boolean HAS_MASTER_REGISTRY;
     protected static final boolean HAS_RPC_REGISTRY;
 
+    private static final class ConfigurationHolder {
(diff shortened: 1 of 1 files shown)


stoty:
As I mentioned on the ticket, ConnectionInfo modifies the kerberos principal and keytab properties.

We should look into whether that causes problem with caching, and if yes, how to solve that.

stoty:
I'll try to track this down on Monday.

lokiore:
@stoty yes forgot about that, previous to registry support, we were creating new config when user was login in, and as it happens once in a day or some hours, that is not a problem , I did the same thing in the PR.
If that is sufficient this should be good!?

stoty:
> or I was thinking we can just update the config with these params when we are logging in actual configuration object, but I am not sure if we need them there except when we are logging in.

Yes, that sounds like the correct solution. 
Your modified patch would still have the perf regression when kerberos is changed, though.

When I did the refactor, I copied this code here with the rest of the config checks, but we do not actually need to log into Kerberos to build ConnectionInfo, this was just a convenient place to do it along with the other normalization steps in the refactors. My intent was not to change kerberos-related functionality/calls.

I don't remember the exact thread/principal flow either, unfortunately.

We would need to dig into whether we correctly support working on Connections with different principals correctly before/after my patch.
My gut feeling is that that would require checking the current principal in every JDBC entry point, and logging into it if needed when being called, and I don't remember us doing that.

Even if didn't work correctly before, we don't need to fix that in this ticket, only solve the perf regression, but we need to understand the thread/principal flow to decide where we need to perform the kerberos login optimally.

It would be great if you could dig into this @lokiore. If you do not have the time/inclination, then let me know and I will take this up.

lokiore:
> Yes, that sounds like the correct solution.
@stoty from the discussion, there are couple of options for this ticket.
- Go with current updated patch (Your modified patch would still have the perf regression when kerberos is changed, though.) As far as my understanding, we want ConnectionInfo as immutable object and updating configuration at ConnectionInfo level will make it different from config in Builder and we will have to keep them in sync which can be done, but maybe we can just create a new config here as it was same before [PHOENIX-6523](https://issues.apache.org/jira/browse/PHOENIX-6523), 

and for perf regression here, assuming we only login for every some hours one time and kerberos change won't be for every connection creation? As it was there before as well and wasn't really a problem and can be improved with next ticket for below
> We would need to dig into whether we correctly support working on Connections with different principals correctly before/after my patch.
My gut feeling is that that would require checking the current principal in every JDBC entry point, and logging into it if needed when being called, and I don't remember us doing that.
Even if didn't work correctly before, we don't need to fix that in this ticket, only solve the perf regression, but we need to understand the thread/principal flow to decide where we need to perform the kerberos login optimally.
It would be great if you could dig into this @lokiore. If you do not have the time/inclination, then let me know and I will take this up.

And sure I would like to take this up and will create another ticket for same..

virajjasani:
> The Hadoop implementation does some clever caching and default handling, so that creating a new Hadoop Configuration object is nearly free.
Is there any reason HBase does not do that ?

Btw this is valid point, we can/should cache Config object in HBase too. This might be worth discussion in the upstream community. Would you like to start a thread on dev@hbase @stoty?

virajjasani:
> The Hadoop implementation does some clever caching and default handling, so that creating a new Hadoop Configuration object is nearly free.
Is there any reason HBase does not do that ?

Btw this is valid point, we can/should cache Config object in HBase too. Would you like to start a thread on dev@hbase @stoty?

stoty:
I got sidetracked, I may not be able to react here for a day or two.

stoty:
The kerberos login and the Configuration creation (even with the modified patch) happens every time when a connection is created (from JDBC or internally) with explicit principal/keytab.

Even if the UGI call is shortcut and quick when we are already logged in as the requested principal (did not check), the original HBaseConfiguration performance regression remains in this case.

See my comment on the code for a suggested quick-and-dirty workaround for the performance issue.

I think that ultimately (next patch) we should move the UGI stuff out of this class, in which case there is no need to modify or create a new HBaseConfiguration, which makes syncing it moot.

stoty:
Thanks @virajjasani .

I have too much on my plate right now to initiate and follow another discussion.
If you have the time, you're absolutely welcome to pick that topic up and start the discussion.

virajjasani:
> Thanks @virajjasani .
> 
> 
> 
> I have too much on my plate right now to initiate and follow another discussion.
> 
> If you have the time, you're absolutely welcome to pick that topic up and start the discussion.

Let me also get back to this in a couple of weeks.

virajjasani:
> Even if the UGI call is shortcut and quick when we are already logged in as the requested principal (did not check), the original HBaseConfiguration performance regression remains in this case.

Yes, but this is still fine for the given Jira right? In a follow-up Jira, we can remove UGI from the flow?

stoty:
> > Even if the UGI call is shortcut and quick when we are already logged in as the requested principal (did not check), the original HBaseConfiguration performance regression remains in this case.
> 
> Yes, but this is still fine for the given Jira right? In a follow-up Jira, we can fix new config + UGI setup if possible?

My quoted reply was wrong, I have edited it out in the meantime, @lokiore  was right.

Yes, we can use the current patch as is, or implement my suggested change to make sure that we never need to create a full HBaseConfiguration object.

virajjasani (approved):
+1, changes resolved perf regression. However, let's wait until we get final +1 from @stoty before merging the changes.

Either an immediate 5.2.2 or 5.3.0 in a couple of months should have the changes anyway.

stoty (approved):
+1 LGTM

This fixes the problem for some of the cases.

We should have a follow-up ticket to clean up and fix the case where the principal is specified in the URL.

lokiore:
As we already have a folow up JIRA (https://issues.apache.org/jira/browse/PHOENIX-7470), will merge this PR in 24 hours. and then work on that with above suggestions. Let me know if there are any concerns. Thanks

lokiore:
@virajjasani @stoty do we need this in 5.2 and 5.1 branch as well?
or just 5.2 as we are not releasing 5.1 anymore..

virajjasani:
+1 for 5.2, we can have 5.2.2 release.

stoty:
I don't remember a discussion to EOL 5.1, though it would match our past pattern of mostly releasing patches for the latest minor version.

PR #2028, opened 2024-11-21
PHOENIX-7468 NPE in SchemaExtractionTool on salted table

tkhurana:
A column which is a PK column was incorrectly classified as non PK column and calling `col.getFamilyName().getString()` will cause NPE because familyName is null for PK columns


jinggou (approved):
LGTM +1

stoty (approved):
+1 LGTM

PR #2029, opened 2024-11-24
PHOENIX-7469 Partitioned CDC Index for partitioned tables

kadirozde:


kadirozde:
The main change in this PR is to use the encoded PHOENIX_ROW_TIMESTAMP() value in the index row key as the timestamp for index row instead of using the cell timestamp. This makes sure that when the cell timestamp does not agree with the encoded value on the row key, we use the encoded value. This happens for delete makers. In order to extract the encoded value, we need to know if the index is a partitioned index (multi-tenanted or view index). This required making the corresponding attributes of IndexMaintainer accessible publicly. In addition to that I improved the test code and made sure including PARTITION_ID() and PHOENIX_ROW_TIMESTAMP() expressions in the WHERE clause of the CDC query is properly handled. It is a pretty short PR and hope you will be able to review quickly. @virajjasani @tkhurana @palashc

PR #2033, opened 2024-12-03
PHOENIX-7459 : Bootstrap stream metadata when CDC is enabled on a table

palashc:
Jira: PHOENIX-7459

### DIFF
--- a/phoenix-core-client/src/main/java/org/apache/phoenix/query/QueryConstants.java
+++ b/phoenix-core-client/src/main/java/org/apache/phoenix/query/QueryConstants.java
@@ -655,11 +655,11 @@ public interface QueryConstants {
             SYSTEM_CDC_STREAM_STATUS_TABLE + "\"(\n" +
             // PK columns
             TABLE_NAME + " VARCHAR NOT NULL," +
-            STREAM_STATUS + " VARCHAR NOT NULL," +
+            STREAM_NAME + " VARCHAR NOT NULL," +
             // Non-PK columns
-            STREAM_NAME + " VARCHAR,\n" +
+            STREAM_STATUS + " VARCHAR,\n" +
             "CONSTRAINT " + SYSTEM_TABLE_PK_NAME + " PRIMARY KEY (" +
-            TABLE_NAME + "," + STREAM_STATUS + "))\n" +
+            TABLE_NAME + "," + STREAM_NAME + "))\n" +
             HConstants.VERSIONS + "=%s,\n" +
             ColumnFamilyDescriptorBuilder.KEEP_DELETED_CELLS + "=%s,\n" +
             TRANSACTIONAL + "=" + Boolean.FALSE;
--- a/phoenix-core-client/src/main/java/org/apache/phoenix/schema/MetaDataClient.java
+++ b/phoenix-core-client/src/main/java/org/apache/phoenix/schema/MetaDataClient.java
@@ -22,6 +22,7 @@ import static org.apache.phoenix.exception.SQLExceptionCode.ERROR_WRITING_TO_SCH
 import static org.apache.phoenix.exception.SQLExceptionCode.SALTING_NOT_ALLOWED_FOR_CDC;
 import static org.apache.phoenix.exception.SQLExceptionCode.TABLE_ALREADY_EXIST;
 import static org.apache.phoenix.jdbc.PhoenixDatabaseMetaData.CDC_INCLUDE_TABLE;
+import static org.apache.phoenix.jdbc.PhoenixDatabaseMetaData.SYSTEM_CDC_STREAM_STATUS_NAME;
 import static org.apache.phoenix.jdbc.PhoenixDatabaseMetaData.TTL_NOT_DEFINED;
 import static org.apache.phoenix.jdbc.PhoenixDatabaseMetaData.STREAMING_TOPIC_NAME;
 import static org.apache.phoenix.jdbc.PhoenixDatabaseMetaData.SYSTEM_TASK_TABLE;
@@ -135,6 +136,7 @@ import static org.apache.phoenix.schema.PTableType.TABLE;
 import static org.apache.phoenix.schema.PTableType.VIEW;
 import static org.apache.phoenix.schema.types.PDataType.FALSE_BYTES;
 import static org.apache.phoenix.schema.types.PDataType.TRUE_BYTES;
+import static org.apache.phoenix.util.CDCUtil.CDC_STREAM_NAME_FORMAT;
 
 import java.io.BufferedReader;
 import java.io.File;
@@ -2039,7 +2041,55 @@ public class MetaDataClient {
         createTableInternal(tableStatement, null, dataTable, null, null, null, null,
                 null, null, false, null,
                 null, statement.getIncludeScopes(), tableProps, commonFamilyProps);
+        // for now, only track stream partition metadata for tables, TODO: updatable views
+        if (PTableType.TABLE.equals(dataTable.getType())) {
+            updateStreamPartitionMetadata(dataTableFullName);
+        }
         return new MutationState(0, 0, connection);
+    }
+
+    /**
+     * Trigger CDC Stream Partition metadata bootstrap for the given table in the background.
+     * Mark status as ENABLING in SYSTEM.CDC_STREAM_STATUS and add {@link CdcStreamPartitionMetadataTask}
+     * to SYSTEM.TASK which updates partition metadata based on table regions.
+     */
+    private void updateStreamPartitionMetadata(String tableName) throws SQLException {
+        long cdcIndexTimestamp = CDCUtil.getCDCCreationTimestamp(connection.getTableNoCache(tableName));
+        String streamStatusSQL = "UPSERT INTO " + SYSTEM_CDC_STREAM_STATUS_NAME + " VALUES (?, ?, ?)";
+        PreparedStatement ps = connection.prepareStatement(streamStatusSQL);
+        String streamName = String.format(CDC_STREAM_NAME_FORMAT, tableName, cdcIndexTimestamp);
+        ps.setString(1, tableName);
+        ps.setString(2, streamName);
+        ps.setString(3, CDCUtil.CdcStreamStatus.ENABLING.getSerializedValue());
+        ps.executeUpdate();
+        connection.commit();
+
+        try {
+            List<Mutation> sysTaskUpsertMutations = Task.getMutationsForAddTask(
+                    new SystemTaskParams.SystemTaskParamsBuilder()
+                    .setConn(connection)
+                    .setTaskType(PTable.TaskType.CDC_STREAM_PARTITION)
+                    .setTableName(tableName) //give full table name
+                    .setSchemaName(streamName) // use schemaName to pass streamName
+                    .build());
+            byte[] rowKey = sysTaskUpsertMutations
+                    .get(0).getRow();
+            MetaDataProtocol.MetaDataMutationResult metaDataMutationResult =
+                    Task.taskMetaDataCoprocessorExec(connection, rowKey,
+                            new TaskMetaDataServiceCallBack(sysTaskUpsertMutations));
+            if (MetaDataProtocol.MutationCode.UNABLE_TO_UPSERT_TASK.equals(
+                    metaDataMutationResult.getMutationCode())) {
+                throw new SQLExceptionInfo.Builder(SQLExceptionCode.UNABLE_TO_UPSERT_TASK)
+                        .setSchemaName(SYSTEM_SCHEMA_NAME)
+                        .setTableName(SYSTEM_TASK_TABLE).build().buildException();
+            }
+        } catch (IOException ioe) {
+            throw new SQLExceptionInfo.Builder(SQLExceptionCode.UNABLE_TO_UPSERT_TASK)
+                    .setRootCause(ioe)
+                    .setMessage(ioe.getMessage())
+                    .setSchemaName(SYSTEM_SCHEMA_NAME)
+                    .setTableName(SYSTEM_TASK_TABLE).build().buildException();
+        }
     }
 
     /**
--- a/phoenix-core-client/src/main/java/org/apache/phoenix/schema/MetaDataClient.java
+++ b/phoenix-core-client/src/main/java/org/apache/phoenix/schema/MetaDataClient.java
@@ -2008,7 +2008,10 @@ public class MetaDataClient {
             }
             throw e;
         }
-
+        // for now, only track stream partition metadata for tables, TODO: updatable views
+        if (PTableType.TABLE.equals(dataTable.getType())) {
+            updateStreamPartitionMetadata(dataTableFullName);
+        }
         List<PColumn> pkColumns = dataTable.getPKColumns();
         List<ColumnDef> columnDefs = new ArrayList<>();
         List<ColumnDefInPkConstraint> pkColumnDefs = new ArrayList<>();
@@ -2041,10 +2044,6 @@ public class MetaDataClient {
         createTableInternal(tableStatement, null, dataTable, null, null, null, null,
                 null, null, false, null,
                 null, statement.getIncludeScopes(), tableProps, commonFamilyProps);
-        // for now, only track stream partition metadata for tables, TODO: updatable views
-        if (PTableType.TABLE.equals(dataTable.getType())) {
-            updateStreamPartitionMetadata(dataTableFullName);
-        }
         return new MutationState(0, 0, connection);
     }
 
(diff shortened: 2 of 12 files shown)


phoenix-core-client/src/main/java/org/apache/phoenix/util/CDCUtil.java
```diff
@@ -150,4 +152,31 @@ public static boolean isBinaryType(PDataType dataType) {
...
+        public String getSerializedValue() {
+            return serializedValue;
+        }
+    }
+
+    public static long getCDCCreationTimestamp(PTable table) {
+        long ts = 0;
+        for (PTable index : table.getIndexes()) {
+            if (CDCUtil.isCDCIndex(index)) {
+                ts = index.getTimeStamp();
```
virajjasani, line 177: nit: return value directly from here?

phoenix-core-server/src/main/java/org/apache/phoenix/coprocessor/tasks/CdcStreamPartitionMetadataTask.java
```diff
@@ -0,0 +1,100 @@
...
+ */
+public class CdcStreamPartitionMetadataTask extends BaseTask  {
+
+    public static final Logger LOGGER = LoggerFactory.getLogger(CdcStreamPartitionMetadataTask.class);
+    public static String CDC_STREAM_STATUS_UPSERT_SQL
+            = "UPSERT INTO " + SYSTEM_CDC_STREAM_STATUS_NAME + " VALUES (?, ?, ?)";
+
+    // parent_partition_id will be null, set partition_end_time to -1
+    private static String CDC_STREAM_PARTITION_UPSERT_SQL
+            = "UPSERT INTO " + SYSTEM_CDC_STREAM_NAME + " VALUES (?,?,?,null,?,-1,?,?)";
```
virajjasani, line 35: nit: public static final

phoenix-core-client/src/main/java/org/apache/phoenix/schema/MetaDataClient.java
```diff
@@ -2039,9 +2041,57 @@ public MutationState createCDC(CreateCDCStatement statement) throws SQLException
...
         return new MutationState(0, 0, connection);
     }
 
+    /**
+     * Trigger CDC Stream Partition metadata bootstrap for the given table in the background.
+     * Mark status as ENABLING in SYSTEM.CDC_STREAM_STATUS and add {@link CdcStreamPartitionMetadataTask}
+     * to SYSTEM.TASK which updates partition metadata based on table regions.
+     */
+    private void updateStreamPartitionMetadata(String tableName) throws SQLException {
+        long cdcIndexTimestamp = CDCUtil.getCDCCreationTimestamp(connection.getTableNoCache(tableName));
```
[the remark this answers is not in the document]
palashc: I will check if the index ptable is updated in the data table ptable in the client's cache.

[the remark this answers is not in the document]
palashc: confirmed that getTable should work.

virajjasani (approved):
+1, pending clean build results

phoenix-core-client/src/main/java/org/apache/phoenix/schema/MetaDataClient.java
```diff
@@ -2039,9 +2041,57 @@ public MutationState createCDC(CreateCDCStatement statement) throws SQLException
...
+                        .setSchemaName(SYSTEM_SCHEMA_NAME)
+                        .setTableName(SYSTEM_TASK_TABLE).build().buildException();
+            }
+        } catch (IOException ioe) {
+            throw new SQLExceptionInfo.Builder(SQLExceptionCode.UNABLE_TO_UPSERT_TASK)
+                    .setRootCause(ioe)
+                    .setMessage(ioe.getMessage())
+                    .setSchemaName(SYSTEM_SCHEMA_NAME)
+                    .setTableName(SYSTEM_TASK_TABLE).build().buildException();
+        }
```
haridsv, line 2092: Shouldn't we do this before creating the CDC object itself to be technically correct and avoid the window when the CDC is active but the stream is not? Also, how do we plan to synchronize this with the ongoing splits and merges?

haridsv:
When will the DROP CDC be hooked up to change the stream status?

palashc:
@haridsv Since DROP CDC needs more changes like not dropping the index but only stopping the writes to the index, I was thinking we can do that part as a separate JIRA. I can create another subtask under PHOENIX-7456, @virajjasani wdyt?

phoenix-core-client/src/main/java/org/apache/phoenix/schema/MetaDataClient.java
palashc: @haridsv [PHOENIX-7460](https://issues.apache.org/jira/browse/PHOENIX-7460) for keeping region metadata in sync.

phoenix-core-client/src/main/java/org/apache/phoenix/schema/MetaDataClient.java
```diff
@@ -2039,9 +2041,57 @@ public MutationState createCDC(CreateCDCStatement statement) throws SQLException
...
+    private void updateStreamPartitionMetadata(String tableName) throws SQLException {
+        long cdcIndexTimestamp = CDCUtil.getCDCCreationTimestamp(connection.getTable(tableName));
+        String streamStatusSQL = "UPSERT INTO " + SYSTEM_CDC_STREAM_STATUS_NAME + " VALUES (?, ?, ?)";
+        PreparedStatement ps = connection.prepareStatement(streamStatusSQL);
+        String streamName = String.format(CDC_STREAM_NAME_FORMAT, tableName, cdcIndexTimestamp);
+        ps.setString(1, tableName);
+        ps.setString(2, streamName);
+        ps.setString(3, CDCUtil.CdcStreamStatus.ENABLING.getSerializedValue());
+        ps.executeUpdate();
+        connection.commit();
```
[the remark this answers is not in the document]
palashc: done

phoenix-core-server/src/main/java/org/apache/phoenix/coprocessor/tasks/CdcStreamPartitionMetadataTask.java
```diff
@@ -0,0 +1,127 @@
...
+            pconn = QueryUtil.getConnectionOnServer(env.getConfiguration()).unwrap(PhoenixConnection.class);
+            List<HRegionLocation> tableRegions = pconn.getQueryServices().getAllTableRegions(
+                    tableName.getBytes(), getTableRegionsTimeout);
+            upsertPartitionMetadata(pconn, tableName, streamName, tableRegions);
+            updateStreamStatus(pconn, tableName, streamName);
+            return new TaskRegionObserver.TaskResult(TaskRegionObserver.TaskResultCode.SUCCESS, "");
+        } catch (Throwable t) {
+            LOGGER.error("Exception while bootstrapping CDC Stream Partition Metadata for "
+                    + taskRecord.getTableName() + " and timestamp " + timestamp.toString(), t);
+            return new TaskRegionObserver.TaskResult(TaskRegionObserver.TaskResultCode.FAIL, t.toString());
```
palashc, line 78: TODO: Consider a RETRY strategy here.

phoenix-core-client/src/main/java/org/apache/phoenix/query/QueryConstants.java
```diff
@@ -655,11 +655,11 @@ enum JoinType {INNER, LEFT_OUTER}
             SYSTEM_CDC_STREAM_STATUS_TABLE + "\"(\n" +
             // PK columns
             TABLE_NAME + " VARCHAR NOT NULL," +
-            STREAM_STATUS + " VARCHAR NOT NULL," +
+            STREAM_NAME + " VARCHAR NOT NULL," +
             // Non-PK columns
-            STREAM_NAME + " VARCHAR,\n" +
+            STREAM_STATUS + " VARCHAR,\n" +
```
virajjasani, line 660: @palashc btw the query pattern is going to be:

```
Step 1: Get the active table stream from the metadata.

SELECT STREAM_NAME FROM SYSTEM.CDC_STREAM_STATUS WHERE TABLE_NAME = <table-name> AND STREAM_STATUS = 'ACTIVE'
```

How would keeping STREAM_NAME help here? I missed this earlier.

palashc: @virajjasani I changed it because the update pattern is 
1. UPSERT (table_name, stream_name, ENABLING)
2. UPSERT (table_name, stream_name, ENABLED)

palashc: For querying, since there should only be one stream per table, the query should still be efficient?
SELECT STREAM_NAME FROM SYSTEM.CDC_STREAM_STATUS WHERE TABLE_NAME = <table-name> AND STREAM_STATUS = 'ENABLED'

phoenix-core-server/src/main/java/org/apache/phoenix/coprocessor/tasks/CdcStreamPartitionMetadataTask.java
palashc: @haridsv @virajjasani Not sure under what kind of failures, we should be retrying the task here?

phoenix-core-client/src/main/java/org/apache/phoenix/query/QueryConstants.java
haridsv: Also for the fact that you can can have multiple streams in the same state, see my latest comment on the corresponding [thread](https://docs.google.com/document/d/157KXONp4ZoWmhUBsy-lK3LNAanCsuJ3cb7qYTmozgZo/edit?disco=AAABZrpxMfI) in the design doc.

virajjasani:
DROP CDC can be separate sub task IMO. Keeping individual sub tasks make it much simpler to follow the change history.
I think we should be good to go unless there is any significant concern.

virajjasani:
We even need more validation changes for CDC creation anyway e.g. allow only one CDC Index to be created per table, and therefore allow only one CDC object per table.
Multiple CDC objects are redundant entries in SYSTEM.CATALOG if they only point to single CDC Index.

I mean to say there will always be some improvements, so I believe we should follow up with separate sub tasks.

virajjasani:
LThough if the PR changes are manageable as single sub task, we can include them here as well. Not a problem. It's upto you @palashc.

phoenix-core-server/src/main/java/org/apache/phoenix/coprocessor/tasks/CdcStreamPartitionMetadataTask.java
palashc: done

phoenix-core-server/src/main/java/org/apache/phoenix/coprocessor/tasks/CdcStreamPartitionMetadataTask.java
```diff
@@ -0,0 +1,153 @@
...
+        int getTableRegionsTimeout = configuration.getInt(PHOENIX_STREAMS_GET_TABLE_REGIONS_TIMEOUT,
+                DEFAULT_PHOENIX_STREAMS_GET_TABLE_REGIONS_TIMEOUT);
+        PhoenixConnection pconn = null;
+        String tableName = taskRecord.getTableName();
+        String streamName = taskRecord.getSchemaName();
+        Timestamp timestamp = taskRecord.getTimeStamp();
+        try {
+            pconn = QueryUtil.getConnectionOnServer(env.getConfiguration()).unwrap(PhoenixConnection.class);
+            List<HRegionLocation> tableRegions = pconn.getQueryServices().getAllTableRegions(
+                    tableName.getBytes(), getTableRegionsTimeout);
```
haridsv, line 75: If there is a timeout due to the table having a large number of regions, it may continue to timeout even in retries. How about we query for the PARTITION_END_KEY of the last partition and use `getTableRegions` to start from that key? We could load any existing partition records for this stream first (which will be empty the first time so we will default to `HConstants.EMPTY_START_ROW`), sort them and get the highest end key. The end key for `getTableRegions` would also be `HConstants.EMPTY_START_ROW`.

phoenix-core-client/src/main/java/org/apache/phoenix/schema/MetaDataClient.java
```diff
@@ -3942,11 +4013,22 @@ public MutationState dropCDC(DropCDCStatement statement) throws SQLException {
...
+        String streamStatusSQL = "UPSERT INTO " + SYSTEM_CDC_STREAM_STATUS_NAME + " VALUES (?, ?, ?)";
+        String streamName = String.format(CDC_STREAM_NAME_FORMAT, parentTableName, cdcTableName, cdcIndexTimestamp);
+        try (PreparedStatement ps = connection.prepareStatement(streamStatusSQL)) {
+            ps.setString(1, parentTableName);
+            ps.setString(2, streamName);
+            ps.setString(3, CDCUtil.CdcStreamStatus.DISABLED.getSerializedValue());
+            ps.executeUpdate();
+            connection.commit();
+            LOGGER.info("Marked stream {} for table {} as DISABLED", streamName, parentTableName);
+        }
```
haridsv, line 4028: We should do this after the CDC drop is successful, may be even after the index drop.

phoenix-core-client/src/main/java/org/apache/phoenix/util/CDCUtil.java
```diff
@@ -150,4 +154,30 @@ public static boolean isBinaryType(PDataType dataType) {
...
+            return serializedValue;
+        }
+    }
+
+    public static long getCDCCreationTimestamp(PTable table) {
+        for (PTable index : table.getIndexes()) {
+            if (CDCUtil.isCDCIndex(index)) {
+                return index.getTimeStamp();
+            }
+        }
```
haridsv, line 180: This method is misleading, it is returning the timestamp of the first index it finds. Since currently we can have multiple CDC indexes, it may return different value at different times. In theory, this can make `updateStreamPartitionMetadata` reuse a previous stream name instead of assigning a new one and `dropCDC` try to update the status for a wrong stream.

I am thinking, we can do the following as a quick workaround:
1. We can change `CDC_STREAM_NAME_FORMAT` to just "<table name>-UUID"` and avoid depending on this method in `updateStreamPartitionMetadata`.
2. In `dropCDC`, query for the name of the ENABLED/ENABLING stream to change its status, rather than constructing the name from `CDC_STREAM_NAME_FORMAT`.
3. In the ITs, we would assert on the existence of a stream of specific status rather than on the specific name.

palashc: @haridsv As we discussed I added the check to only allow a single (enable/enabling) stream in createCDC. This also has the effect of only allowing a single CDC index since I throw a SQLException.

palashc: Please correct me if I am wrong but we want to move towards single stream and single CDC index, right?

haridsv: Yes you are right, since we are not allowing more than 1 CDC, the logic is harmless, sorry about the spam.

phoenix-core-server/src/main/java/org/apache/phoenix/coprocessor/tasks/CdcStreamPartitionMetadataTask.java
virajjasani: We can use existing method that is already used by getAllTableRegions() API.

virajjasani: @haridsv I just had offline discussion with @palashc. Since he is already using the existing API which is already used by `BaseResultIterators` for each query. The timeout is not for HBase ConnectionRegistry API but its for Phoenix client to internally figure out if meta RPC calls are exceeding some timeout.

For very large tables with full table scan, while going through the loop (CQSI#getTableRegions), the timeout can happen only if the region locations are not cached. Let's assume we have 500k regions with 2 min timeout set by the CDC Stream Metadata Task. Let's say none of those regions are cached at the moment. So, CQSI#getTableRegions starts RPC calls for each region and somehow meta is slow to respond and hence after 200k region locations, CQSI exceeds 2 min timeout and throws error. This means we already have 200k region locations cached at the client side. Hence, next time when the task is retried, CQSI#getTableRegions will go through 200k region location loop in few milliseconds because no RPC call is required. For the remaining regions, it will start creating RPC calls, until every region location is cached.

Given that this API is already being used by Phoenix client with every query, we are confident that it is already working fine and even for large tables, it can potentially timeout only few times, not all time.

virajjasani: We have seen issues with range scan queries timing out once or twice due to large num of regions. To handle them efficiently, precise region boundaries are now used: https://github.com/apache/phoenix/commit/0b9d8201875c27a8b8a1084680104057bcd46e60

However, even with large tables, CQSI#getTableRegions never timed out infinitely. It only used to timeout 1 or 2 times. This is because it never has to re-do the whole work, as every RPC call will be done only once for each region location. Once region location is retrieved, HBase client caches it.

virajjasani:
+1, pending jenkins build results.

palashc:
Latest Test Report: https://ci-hadoop.apache.org/job/Phoenix/job/Phoenix-PreCommit-GitHub-PR/job/PR-2033/13/testReport/
Failures look unrelated to code change and pass locally. 
```
org.apache.phoenix.end2end.AggregateQueryWithRegionMoves2IT.testSplitWithCachedMeta[AggregateQueryWithRegionMoves2IT_3]
org.apache.phoenix.end2end.PermissionNSDisabledWithCustomAccessControllerIT.testSuperUserCanChangePerms
```

PR #2041, opened 2024-12-13
PHOENIX-7484 Optimize mutation plan creation for upserts over tenant connection

sanjeet006py:
Upserts to multi-tenant tables over tenant connections were regressed by 5K - 6K % compared to non-tenant connections. But with this patch upserts over tenant connection are as good as non-tenant connection.

### DIFF
--- a/phoenix-core-client/src/main/java/org/apache/phoenix/jdbc/PhoenixConnection.java
+++ b/phoenix-core-client/src/main/java/org/apache/phoenix/jdbc/PhoenixConnection.java
@@ -761,13 +761,22 @@ public class PhoenixConnection implements MetaDataMutated, SQLCloseable, Phoenix
     }
 
     public PTable getTable(PTableKey key) throws SQLException {
-        PTable table;
-        try {
-            table = getTableRef(key).getTable();
-        } catch (TableNotFoundException e) {
-            table =  getTableNoCache(key.getName());
+        PTableRef tableRef = getTableRefOptimized(key);
+        if (tableRef == null && key.getTenantId() != null) {
+            tableRef = getTableRefOptimized(new PTableKey(null, key.getName()));
         }
-        return table;
+        if (tableRef == null) {
+            return getTableNoCache(key.getName());
+        }
+        return tableRef.getTable();
+    }
+
+    public PTableRef getTableRefOptimized(PTableKey key) {
+        PTableRef tableRef = getQueryServices().getMetaDataCache().getTableRefOptimized(key);
+        if (tableRef != null && prune(tableRef.getTable())) {
+            return null;
+        }
+        return tableRef;
     }
 
     public PTableRef getTableRef(PTableKey key) throws TableNotFoundException {
--- a/phoenix-core-client/src/main/java/org/apache/phoenix/jdbc/PhoenixConnection.java
+++ b/phoenix-core-client/src/main/java/org/apache/phoenix/jdbc/PhoenixConnection.java
@@ -761,9 +761,9 @@ public class PhoenixConnection implements MetaDataMutated, SQLCloseable, Phoenix
     }
 
     public PTable getTable(PTableKey key) throws SQLException {
-        PTableRef tableRef = getTableRefOptimized(key);
+        PTableRef tableRef = getTableRefNoTNFE(key);
         if (tableRef == null && key.getTenantId() != null) {
-            tableRef = getTableRefOptimized(new PTableKey(null, key.getName()));
+            tableRef = getTableRefNoTNFE(new PTableKey(null, key.getName()));
         }
         if (tableRef == null) {
             return getTableNoCache(key.getName());
@@ -771,8 +771,8 @@ public class PhoenixConnection implements MetaDataMutated, SQLCloseable, Phoenix
         return tableRef.getTable();
     }
 
-    public PTableRef getTableRefOptimized(PTableKey key) {
-        PTableRef tableRef = getQueryServices().getMetaDataCache().getTableRefOptimized(key);
+    public PTableRef getTableRefNoTNFE(PTableKey key) {
+        PTableRef tableRef = getQueryServices().getMetaDataCache().getTableRefNoTNFE(key);
         if (tableRef != null && prune(tableRef.getTable())) {
             return null;
         }
--- a/phoenix-core-client/src/main/java/org/apache/phoenix/schema/PMetaData.java
+++ b/phoenix-core-client/src/main/java/org/apache/phoenix/schema/PMetaData.java
@@ -29,6 +29,7 @@ public interface PMetaData extends MetaDataMutated, Iterable<PTable> {
     }
     public int size();
     public PTableRef getTableRef(PTableKey key) throws TableNotFoundException;
+    public PTableRef getTableRefOptimized(PTableKey key);
     public void pruneTables(Pruner pruner);
     public PFunction getFunction(PTableKey key) throws FunctionNotFoundException;
     public void pruneFunctions(Pruner pruner);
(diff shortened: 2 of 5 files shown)


phoenix-core-client/src/main/java/org/apache/phoenix/schema/PMetaDataImpl.java
```diff
@@ -83,14 +83,27 @@ private void updateGlobalMetric(PTableRef pTableRef) {
...
+    @Override
+    public PTableRef getTableRefOptimized(PTableKey key) {
+        try {
+            return getTableRef(key, true);
+        } catch (TableNotFoundException e) {
+            throw new IllegalStateException(e);
+        }
+    }
+
+    private PTableRef getTableRef(PTableKey key, boolean doNotThrowTableNotFoundException) throws TableNotFoundException {
```
virajjasani, line 98: With this, you also mean to cover PHOENIX-7486?

sanjeet006py: Actually, [PHOENIX-7486](https://issues.apache.org/jira/browse/PHOENIX-7486) is a more of generic JIRA for finding and fixing other occurrences of this anti pattern. I fixed one in here as that was contributing to the regressed numbers. W/o the change in `PMetadataImpl` I was able to bring down regression to 30-40% from 5K-4K% but need this change to bring it down to 0.

virajjasani: nit: use `skipTNFE` instead of `doNotThrowTableNotFoundException`

virajjasani:
Very nice find @sanjeet006py!

phoenix-core-client/src/main/java/org/apache/phoenix/schema/PMetaData.java
```diff
@@ -29,6 +29,7 @@ public static interface Pruner {
     }
     public int size();
     public PTableRef getTableRef(PTableKey key) throws TableNotFoundException;
+    public PTableRef getTableRefOptimized(PTableKey key);
```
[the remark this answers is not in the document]
sanjeet006py: > rather it is in selection of right method (the new code PhoenixConnection)

But the selection of different method is all about avoiding creating TNFE object so, the optimization is about not throwing and creating TNFE. Right?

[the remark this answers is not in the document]
haridsv: I think we are both saying the same thing, the optimization is about method selection, but my point is that the new method is just a variant that facilitates this optimization but not directly addressing the optimization itself. The optimization comes from the use case, e.g., say you have a use case that calls the same method but converts the `null` to TNFE (basically what the existing getTableRef does, but now this is the application logic), now this new application code won't be optimized just because it is using this new method. Not the best argument possible, but I hope it conveys the meaning.

haridsv (approved):

phoenix-core-client/src/main/java/org/apache/phoenix/jdbc/PhoenixConnection.java
```diff
@@ -761,13 +761,22 @@ public PTable getTable(String name) throws SQLException {
...
 
     public PTable getTable(PTableKey key) throws SQLException {
-        PTable table;
-        try {
-            table = getTableRef(key).getTable();
-        } catch (TableNotFoundException e) {
-            table =  getTableNoCache(key.getName());
+        PTableRef tableRef = getTableRefNoTNFE(key);
+        if (tableRef == null && key.getTenantId() != null) {
+            tableRef = getTableRefNoTNFE(new PTableKey(null, key.getName()));
```
tkhurana, line 766: Consider the case when there is a tenant view on the table and the tenant has the same name as the underlying table. I think Phoenix allows that. Then if the base table is in the cache but the tenant view is not in the cache you will return the base table object.

sanjeet006py: Then I think what we need to do address PHOENIX-4475. I can take up PHOENIX-4475 and address both PHOENIX-4475 and PHOENIX-7484 in one PR. If we don't use UCF and always hit SYSCAT then its a huge regression with tenant connection compared to non-tenant connection.

sanjeet006py: Another idea could be to prevent creating tenant views with same name as base tables?

phoenix-core-client/src/main/java/org/apache/phoenix/schema/PMetaDataImpl.java
```diff
@@ -83,14 +83,27 @@ private void updateGlobalMetric(PTableRef pTableRef) {
...
+    private PTableRef getTableRef(PTableKey key, boolean skipTNFE) throws TableNotFoundException {
         if (physicalNameToLogicalTableMap.containsKey(key.getName())) {
             key = physicalNameToLogicalTableMap.get(key.getName());
         }
         PTableRef ref = metaData.get(key);
         if (!key.getName().contains(QueryConstants.SYSTEM_SCHEMA_NAME)) {
             updateGlobalMetric(ref);
         }
-        if (ref == null) {
+        if (ref == null && ! skipTNFE) {
```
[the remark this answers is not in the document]
sanjeet006py: We can create a NULL_TABLE_REF static constant and return that instead of returning null.

phoenix-core-client/src/main/java/org/apache/phoenix/jdbc/PhoenixConnection.java
sanjeet006py: > Then if the base table is in the cache but the tenant view is not in the cache you will return the base table object.

This will happen now also with MetaDataClient#updateCache() but it relies on UCF to get cache refreshed eventually as per my understanding. Please correct me if I am wrong.

tkhurana: > Another idea could be to prevent creating tenant views with same name as base tables?
You can introduce a name conflict when creating a tenant view with another non related table also. It doesn't even need to be the base table.

sanjeet006py: > You can introduce a name conflict when creating a tenant view with another non related table also. It doesn't even need to be the base table.

Got it. So how about: [this one](https://github.com/apache/phoenix/pull/2041#discussion_r1887104412). I think this will work, right?

tkhurana: I also think it as an application problem. You are using a tenant connection but the entity you are writing to is a global object. Maybe you should be using a global connection in the first place.

sanjeet006py: > Maybe you should be using a global connection in the first place.

With global connection I feel one tenant can modify the records of other tenants by providing some other tenant Id in UPSERT. With tenant specific connection I think we able to impose tenant boundaries and prevent one tenant from modifying records of other tenants. I think on a multi-tenant base table we should be able to read and write using tenant connection. WDYT?

tkhurana: Well, that's a weak argument because if I can specify some other tenantid in UPSERT, what's stopping to create a tenant connection with someone else's tenantid.

sanjeet006py: Yeah, that's true. For queries we use tenant connection to query a multi-tenant base table but for writes we need to use global connection? There is nothing wrong in it as such but does seem odd to me.

tkhurana: > For queries we use tenant connection to query a multi-tenant base table 

Where is that mandated ? 

You need to think in terms of the entity you are reading and writing to. If the entity is a TENANT VIEW  use a tenant connection otherwise use a global connection.

sanjeet006py:
Closing this as PHOENIX-7490 will fix this: PHOENIX-7484 along with a bigger problem (functional bug) we have as highlighted in PHOENIX-7490.

PR #2044, opened 2024-12-18
PHOENIX-7492: Need to clear the dependencies once it is closed

abhradeepkundu:
PHOENIX-7492: Need to clear the dependencies once it is closed, to avoid null pointer in future

Change-Id: Ifba79d894e20d0abd12382f969cdc1dd5119bd36


richardantal (commented):
Nice catch!

Do you think we should call a `dependencies.clear() ` or `caches.clear()` similartly in
HashJoinPlan.java, LiteralResultIterationPlan.java ?

abhradeepkundu:
> Nice catch!
> 
> Do you think we should call a `dependencies.clear() ` or `caches.clear()` similartly in HashJoinPlan.java, LiteralResultIterationPlan.java ?

Yes I can make the changes.

richardantal (approved):

PR #2053, opened 2025-01-08
PHOENIX-7502 :- Decouple principal from HAGroupInfo

lokiore:
- Decoupling principal from HAGroupInfo so that we can have 1:1 mapping between HAGroupInfo -> HighAvailabilityGroup and 1:n mapping between HAGroupInfo -> HAURLInfo (principal)
- Storing HAURLInfo in connection's own context so it has correct info.
- Policy can access all users associated with given HAGroup in case of clusterRoleRecord change
- This means we can have a single HighAvailability group/ clusterRoleRecord for different principals connecting to it and we don't have to create separate watchers for every principal.

### DIFF
--- a/phoenix-core-client/src/main/java/org/apache/phoenix/jdbc/FailoverPhoenixConnection.java
+++ b/phoenix-core-client/src/main/java/org/apache/phoenix/jdbc/FailoverPhoenixConnection.java
@@ -105,7 +105,8 @@ public class FailoverPhoenixConnection implements PhoenixMonitoredConnection {
         this.context = context;
         this.policy = FailoverPolicy.get(context.getProperties());
         this.isClosed = false;
-        this.connection = context.getHAGroup().connectActive(context.getProperties(), context.getHAURLInfo());
+        this.connection = context.getHAGroup().connectActive(context.getProperties(),
+                context.getHAURLInfo());
     }
 
     /**
@@ -179,7 +180,8 @@ public class FailoverPhoenixConnection implements PhoenixMonitoredConnection {
         while (newConn == null &&
                 EnvironmentEdgeManager.currentTimeMillis() < startTime + timeoutMs) {
             try {
-                newConn = context.getHAGroup().connectActive(context.getProperties(), context.getHAURLInfo());
+                newConn = context.getHAGroup().connectActive(context.getProperties(),
+                        context.getHAURLInfo());
             } catch (SQLException e) {
                 cause = e;
                 LOG.info("Got exception when trying to connect to active cluster.", e);
@@ -222,7 +224,8 @@ public class FailoverPhoenixConnection implements PhoenixMonitoredConnection {
                 }
             }
         }
-        LOG.info("Connection {} failed over to {}", context.getHAGroup().getGroupInfo(), connection.getURL());
+        LOG.info("Connection {} failed over to {}", context.getHAGroup().getGroupInfo(),
+                connection.getURL());
     }
 
     /**
@@ -323,8 +326,9 @@ public class FailoverPhoenixConnection implements PhoenixMonitoredConnection {
     @VisibleForTesting
     <T> T wrapActionDuringFailover(SupplierWithSQLException<T> s) throws SQLException {
         checkConnection();
-        final long timeoutMs = Long.parseLong(context.getProperties().getProperty(FAILOVER_TIMEOUT_MS_ATTR,
-                String.valueOf(FAILOVER_TIMEOUT_MS_DEFAULT)));
+        final long timeoutMs = Long.parseLong(context.getProperties().
+                getProperty(FAILOVER_TIMEOUT_MS_ATTR,
+                        String.valueOf(FAILOVER_TIMEOUT_MS_DEFAULT)));
         int failoverCount = 0;
         while (true) {
             try {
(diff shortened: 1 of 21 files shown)


phoenix-core-client/src/main/java/org/apache/phoenix/jdbc/FailoverPhoenixContext.java
```diff
@@ -0,0 +1,48 @@
...
+ * limitations under the License.
+ */
+package org.apache.phoenix.jdbc;
+
+import java.util.Properties;
+
+/**
+ * FailoverPhoenixContext holds the properties and HAGroup Info for a failover phoenix connection.
+ */
+public class FailoverPhoenixContext {
```
kadirozde, line 25: HA Group is not used only with the Failover policy. Is FailoverPhoenixContext used only for the Failover policy?

phoenix-core-client/src/main/java/org/apache/phoenix/jdbc/HAURLInfo.java
```diff
@@ -0,0 +1,87 @@
...
+            return false;
+        }
+        if (other == this) {
+            return true;
+        }
+        if (other.getClass() != getClass()) {
+            return false;
+        }
+        HAURLInfo otherInfo = (HAURLInfo) other;
+        return new EqualsBuilder()
```
jpisaac, line 71: The equals logic should be compatible with the hashcode? Doesn't the principal == null scenario be taken into account

phoenix-core-client/src/main/java/org/apache/phoenix/jdbc/HighAvailabilityGroup.java
```diff
@@ -195,36 +198,104 @@ private HighAvailabilityGroup(HAGroupInfo info, Properties properties) {
...
-                    .setMessage(String.format("URL %s is not a valid HA connection string", url))
-                    .build()
-                    .buildException();
-        }
+    /**
+     * Get an instance of HAURLInfo given the HA connecting URL (with "|") and client properties.
+     * Here we do parsing of url and try to extract principal and other additional params
+     * @throws SQLException
+     */
+    public static HAURLInfo getUrlInfo(String url, Properties properties) throws SQLException {
```
jpisaac, line 206: Are there unit tests for various Url patterns and use cases?

phoenix-core-client/src/main/java/org/apache/phoenix/jdbc/HighAvailabilityGroup.java
```diff
@@ -785,26 +851,49 @@ public String getUrl2() {
             return urls.getSecond();
         }
 
-        public String getJDBCUrl(String zkUrl) {
-            Preconditions.checkArgument(zkUrl.equals(getUrl1()) || zkUrl.equals(getUrl2()),
+
+        //It applies only the current thread context's URL Info as we have 1:n mapping between
+        //HAURLInfo -> HAGroupInfo.
```
jpisaac, line 856: nit: Isn't the 1:n mapping from HAGroupInfo -> HAURLInfo?

phoenix-core-client/src/main/java/org/apache/phoenix/jdbc/HighAvailabilityPolicy.java
```diff
@@ -66,14 +67,19 @@ private void transitStandby(HighAvailabilityGroup haGroup, String zkUrl)
...
+                    cqs = PhoenixDriver.INSTANCE.getConnectionQueryServices(
+                            jdbcZKUrl, haGroup.getProperties());
+                    cqs.closeAllConnections(new SQLExceptionInfo
+                            .Builder(SQLExceptionCode.HA_CLOSED_AFTER_FAILOVER)
+                            .setMessage("Phoenix connection got closed due to failover")
+                            .setHaGroupInfo(haGroup.getGroupInfo().toString()));
+                    LOG.info("Closed all connections to cluster {} for HA group {}",
+                            jdbcZKUrl, haGroup.getGroupInfo());
+                }
+
```
jpisaac, line 82: Shouldn't the finally be inside the loop?

jpisaac (commented):
Took a first pass, Mostly looks good, will take a another pass

phoenix-core-client/src/main/java/org/apache/phoenix/jdbc/FailoverPhoenixConnection.java
```diff
@@ -103,13 +100,12 @@ public class FailoverPhoenixConnection implements PhoenixMonitoredConnection {
      */
     private Map<String, Map<MetricType, Long>> previousReadMetrics = new HashMap<>();
 
-    public FailoverPhoenixConnection(HighAvailabilityGroup haGroup, Properties properties)
+    public FailoverPhoenixConnection(FailoverPhoenixContext context)
```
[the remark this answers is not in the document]
lokiore: This is created on though policy and there we create context every time, that's why didn't add the check.

phoenix-core-client/src/main/java/org/apache/phoenix/jdbc/HAURLInfo.java
lokiore: EqualsBuilder handles the case of null in append function, will add assert in test for this

phoenix-core-client/src/main/java/org/apache/phoenix/jdbc/HighAvailabilityGroup.java
lokiore: I have added tests with different principles, I think can add more with without principle as well

phoenix-core-client/src/main/java/org/apache/phoenix/jdbc/HighAvailabilityGroup.java
lokiore: Forgot to remove comment it was with other approach, will update this

phoenix-core-client/src/main/java/org/apache/phoenix/jdbc/HighAvailabilityPolicy.java
lokiore: Yes Thanks for catching that

jpisaac (approved):

jpisaac (approved):
LGTM + 1 Please add some multi threaded test with different CQSI and principals, as discussed offline.

PR #2054, opened 2025-01-10
PHOENIX-7462 Single row Atomic Upsert/Delete returning result should return ResultSet

virajjasani:
[PHOENIX-7398](https://issues.apache.org/jira/browse/PHOENIX-7398) introduces new PhoenixStatement API to return row for Atomic/Conditional Upserts. Phoenix already supports Upserts as Atomic operation by using ON DUPLICATE KEY clause.

Similarly, [PHOENIX-7411](https://issues.apache.org/jira/browse/PHOENIX-7411) and [PHOENIX-7434](https://issues.apache.org/jira/browse/PHOENIX-7434) introduce Atomic Delete support for the single row deletion. It returns result (row) back to the client (using new API) only if the given client initiated Delete operation deleted the row when the row was alive. In the context of atomicity, this means that the delete operation is already idempotent and will continue to be idempotent, however the client that was able to successfully delete the row by taking lock on the live row (row without delete marker) is expected to be the only client that can retrieve the old row result back.

The purpose of this Jira is to update the API signature for the PhoenixStatement API such that the abstract level ResultSet object can be returned for the returned row instead of Tuple object.

Old API signature:

```
/**
 *
 */
```

New API signature:

```
/**
 * updated or non-updated row as ResultSet object back to the client. This must be used with
 *
 * @return The pair of int and ResultSet, where int represents value 1 for successful row
 * update and 0 for non-successful row update, and ResultSet represents the state of the row.
 */
public Pair<Integer, ResultSet> executeAtomicUpdateReturnRow(String sql) throws SQLException {
```

### DIFF
--- a/phoenix-core-client/src/main/java/org/apache/phoenix/jdbc/PhoenixPrefetchedResultSet.java
+++ b/phoenix-core-client/src/main/java/org/apache/phoenix/jdbc/PhoenixPrefetchedResultSet.java
@@ -1,0 +1,48 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to you under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.phoenix.jdbc;
+
+import java.sql.SQLException;
+import java.util.List;
+
+import org.apache.phoenix.compile.RowProjector;
+import org.apache.phoenix.compile.StatementContext;
+import org.apache.phoenix.schema.tuple.Tuple;
+
+/**
+ * Phoenix ResultSet implementation with prefetched rows. It is expected that the implementation
+ * does not need to use any iterators to make server side RPC call.
+ */
+public class PhoenixPrefetchedResultSet extends PhoenixResultSet {
+
+  private final List<Tuple> prefetchedRows;
+  private int prefetchedRowsIndex;
+
+  public PhoenixPrefetchedResultSet(RowProjector rowProjector,
+      StatementContext ctx, List<Tuple> prefetchedRows) throws SQLException {
+    super(null, rowProjector, ctx);
+    this.prefetchedRows = prefetchedRows;
+    this.prefetchedRowsIndex = 0;
+  }
+
+  @Override
+  protected Tuple getCurrentRowImpl() {
+    return prefetchedRows.get(prefetchedRowsIndex++);
+  }
+
+}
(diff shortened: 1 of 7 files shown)


phoenix-core-client/src/main/java/org/apache/phoenix/util/TupleUtil.java
```diff
@@ -210,4 +219,35 @@ public static int write(Tuple result, DataOutput out) throws IOException {
...
+     * @param toProject Tuple to be projected.
+     * @param tableName Table name.
+     * @param conn Phoenix Connection object.
+     * @return ResultSet for the give single row.
+     * @throws SQLException If any SQL operation fails.
+     */
+    public static ResultSet getResultSet(Tuple toProject, String tableName, Connection conn)
+        throws SQLException {
+        try (PhoenixResultSet resultSet = (PhoenixResultSet) conn.createStatement()
+            .executeQuery("SELECT * FROM " + tableName)) {
```
palashc, line 236: For my understanding, will this query actually execute? If yes, should we add a `LIMIT 1`?

palashc (commented):
Just one comment, LGTM otherwise.

phoenix-core-client/src/main/java/org/apache/phoenix/util/TupleUtil.java
palashc: Got it, since the real execution of it happens during a `next` call?

virajjasani: Yes and also we are using Prefetched ResultSet so we are providing Tuple already. The new ResultSet does not even have ResultIterator so no real scan will happen.

palashc (approved):

phoenix-core-client/src/main/java/org/apache/phoenix/jdbc/PhoenixPrefetchedResultSet.java
```diff
@@ -0,0 +1,48 @@
...
+  public PhoenixPrefetchedResultSet(RowProjector rowProjector,
+      StatementContext ctx, List<Tuple> prefetchedRows) throws SQLException {
+    super(null, rowProjector, ctx);
+    this.prefetchedRows = prefetchedRows;
+    this.prefetchedRowsIndex = 0;
+  }
+
+  @Override
+  protected Tuple getCurrentRowImpl() {
+    return prefetchedRows.get(prefetchedRowsIndex++);
```
haridsv, line 45: When a row is deleted, there will only be a single row returned correct? Is there a scenario in which the size of `prefetchedRows` is more than 1?

In any case, JDBC contract for `next()` is to return a `false` when there are no more rows, but looking at the implementation of `getCurrentRow` in the base class, it looks like user wlll see an `IndexOutOfBoundsException`? We would also need test coverage for this.

virajjasani: Yes I was planning to return null but thought of avoiding the extra if condition because this implementation is only meant to be used by single row Tuple projection to ResultSet conversion for the purpose of atomic update returning single row. Atomic update/delete can not return more than single row or even status (1 for successful update and 0 for failure to update) for more than single row so the usecase itself is limited. Also, usually the application of prefetched result set should know how many rows they are expecting in advance. In this case, it's only one.

This is good point @haridsv, let me put this comment to make it clear. Avoiding if condition was just meant to make this implementation simpler and clean.

haridsv: I understand the reasoning, but I still think breaking the JDBC contract is not a good idea. May be you can find an alternative to honor the contract without adding any overhead?

PR #2058, opened 2025-01-16
PHOENIX-7510 VARBINARY_ENCODED should support using hex format in WHERE clause

virajjasani:
Jira: PHOENIX-7510

With VARBINARY_ENCODED, we support using PreparedStatement#setBytes to set the byte array which needs to be compared with WHERE clause. However, using hex format for the comparison fails. The purpose of this PR is to support using hex format for VARBINARY_ENCODED columns with WHERE clause.


stoty (approved):
+1 LGTM

PR #2059, opened 2025-01-17
PHOENIX-7509 : Metadata Cache should handle tables which have LAST_DDL_TIMESTAMP column null in syscat

palashc:


virajjasani (approved):
+1, pending build results

tkhurana (approved):

PR #2064, opened 2025-01-26
PHOENIX-7515: Add metric for count of Phoenix client batches used by a commit call

sanjeet006py:
Summary of the changes:
- Added a metric to track no. of mutation batches successful per commit call.

### DIFF
--- a/phoenix-core-client/src/main/java/org/apache/phoenix/execute/MutationState.java
+++ b/phoenix-core-client/src/main/java/org/apache/phoenix/execute/MutationState.java
@@ -1425,6 +1425,7 @@ public class MutationState implements SQLCloseable {
                 List<Mutation> currentMutationBatch = null;
                 boolean areAllBatchesSuccessful = false;
                 Object[] resultObjects = null;
+                int mutationBatchSuccessCounter = 0;
 
                 try {
                     if (table.isTransactional()) {
@@ -1530,7 +1531,7 @@ public class MutationState implements SQLCloseable {
                         // REPLAY_ONLY_INDEX_WRITES for first batch
                         // only in case of 1121 SQLException
                         itrListMutation.remove();
-
+                        mutationBatchSuccessCounter++;
                         batchCount++;
                         if (LOGGER.isDebugEnabled())
                             LOGGER.debug("Sent batch of " + mutationBatch.size() + " for "
@@ -1624,7 +1625,7 @@ public class MutationState implements SQLCloseable {
                                     numMutations,
                                     numFailedMutations,
                                     numFailedPhase3Mutations,
-                                    mutationCommitTime);
+                                    mutationCommitTime, mutationBatchSuccessCounter);
                     // Combine failure mutation metrics with committed ones for the final picture
                     committedMutationsMetric.combineMetric(failureMutationMetrics);
                     mutationMetricQueue.addMetricsForTable(htableNameStr, committedMutationsMetric);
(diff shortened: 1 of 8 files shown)


phoenix-core-client/src/main/java/org/apache/phoenix/execute/MutationState.java
```diff
@@ -1530,7 +1531,7 @@ public List<Mutation> getMutationList() {
                         // REPLAY_ONLY_INDEX_WRITES for first batch
                         // only in case of 1121 SQLException
                         itrListMutation.remove();
-
+                        mutationBatchSuccessCounter++;
```
tkhurana, line 1534: Can't we just use `mutationBatchList.size()`

phoenix-core/src/it/java/org/apache/phoenix/monitoring/PhoenixMetricsIT.java
```diff
@@ -508,11 +534,17 @@ public void testMetricsForUpsert() throws Exception {
                     assertEquals("Zero failed mutations expected", 0, metricValue);
                     mutationBatchFailedPresent = true;
                 }
+                else if (metricType.equals(MetricType.MUTATION_BATCH_SUCCESS_COUNTER)) {
+                    assertTrue("Mutation batch success count should be greater than zero",
+                            metricValue > 0);
```
tkhurana, line 539: You can do a much tighter check here by checking for the exact value of mutation batches.

tkhurana:
You should also add a test where you modify the upsert batch size property and then verify the value of this metric.

phoenix-core-client/src/main/java/org/apache/phoenix/execute/MutationState.java
sanjeet006py: Then to know the successful mutation batches, I would have to first keep total batches initially and have to compute the diff to get batches successfully committed. This I feel is a bit indirect way to compute same thing. WDYT?

tkhurana: I feel most of the value is in knowing how many batches we created per commit. That is where we can tune the various parameters to improve performance. If the commit succeeds the `mutationBatchSuccessCounter` will be the same as size of the list. If the commit fails, they might be different but that doesn't seem very interesting from metrics perspective. There is also a `batchCount` variable which seems to be tracking the same thing but

sanjeet006py: Got it. I agree when commit call has failed then our interest will be in knowing why it failed instead of knowing batch count. 
Regarding existing `batchCount` variable, I didn't use it because its shared at MutationState object level. So, if multiple application threads are using one Phoenix connection then they all will increment it and we will end up double counting the number of batches. Though I don't think anyone should use Phoenix connection across multiple application threads but there is nothing stopping them.

I will change to capturing no. of batches computed per commit call. Thanks

tkhurana (approved):

phoenix-core/src/it/java/org/apache/phoenix/monitoring/PhoenixTableLevelMetricsIT.java
```diff
@@ -342,7 +346,7 @@ private static void assertMutationTableMetrics(final boolean isUpsert, final Str
             final long expectedUpsertOrDeleteAggregateSuccessCt,
             final long expectedUpsertOrDeleteAggregateFailureCt,
             final Map<MetricType, Long> writeMutMetrics, final Connection conn,
-            final boolean expectedSystemCatalogMetric)
+            final boolean expectedSystemCatalogMetric, final long expectedMutationBatchCount)
```
d-c-manning, line 349: nit: there's a javadoc for the method with parameters

d-c-manning (approved):

sanjeet006py:
@tkhurana can you please help with merging this PR. Thanks

tkhurana:
@sanjeet006py Can you address the checkstyle warnings please

PR #2067, opened 2025-01-29
PHOENIX-7170 Conditional TTL

tkhurana:


phoenix-core-client/src/main/java/org/apache/phoenix/schema/ConditionalTTLExpression.java
```diff
@@ -0,0 +1,501 @@
...
+    }
+
+    /**
+     * The cells of the row (i.e., result) read from HBase store are lexicographically ordered
+     * for tables using the key part of the cells which includes row, family, qualifier,
+     * timestamp and type. The cells belong of a column are ordered from the latest to
+     * the oldest. The method leverages this ordering and groups the cells into their columns
+     * based on the pair of family name and column qualifier.
+     */
+    private List<Cell> getLatestRowVersion(List<Cell> result) {
```
kadirozde, line 147: The TTL region scanner should not call this if the scan is not a raw scan as this computation is not necessary in that case.

phoenix-core-client/src/main/java/org/apache/phoenix/schema/CompiledConditionalTTLExpression.java
```diff
@@ -0,0 +1,307 @@
...
+
+    /**
+     * This logic has been borrowed from GlobalIndexChecker#removeOlderCells
+     * This is needed to correctly handle updates to the index row when some columns
+     * are set to null and thus are not updated but an older version of the same column
+     * might have a valid value. Since TTLRegionScanner evaluates the row before GlobalIndexChecker
+     * processes the row we need to mask such columns in TTLRegionScanner also for correctness.
+     * @return
+     */
+    private List<Cell> maskOlderCells(List<Cell> cellList) {
```
kadirozde, line 154: Instead of borrowing/copying, we can make this method to a utility method in ScanUtil.

phoenix-core-client/src/main/java/org/apache/phoenix/schema/CompiledConditionalTTLExpression.java
```diff
@@ -0,0 +1,307 @@
...
+        List<Cell> latestRowVersion = new ArrayList<>();
+        Cell currentColumnCell = null;
+        long maxDeleteFamilyTS = 0;
+        for (Cell cell : result) {
+            if (currentColumnCell == null ||
+                    !CellUtil.matchingColumn(cell, currentColumnCell)) {
+                // found a new column cell which has the latest timestamp
+                currentColumnCell = cell;
+                if (currentColumnCell.getType() == Cell.Type.DeleteFamily ||
+                        currentColumnCell.getType() == Cell.Type.DeleteFamilyVersion) {
```
kadirozde, line 129: DeleteFamilyVersion should not be treated as DeleteFamily here. In other words, DeleteFamilyVersion deletes only a specific version of the row such that the timestamp of the DeleteFamilyVersion equals to the timestamp of the row version.

kadirozde (approved):
Thank you for extensive testing and comments! It looks great!

PR #2073, opened 2025-02-06
PHOENIX-7501 GC issues in TTLRegionScanner when gap is more than TTL

tkhurana:


phoenix-core/src/it/java/org/apache/phoenix/end2end/TableTTLIT.java
```diff
@@ -338,15 +342,105 @@ public void testMultipleRowsWithUpdatesMoreThanTTLApart() throws Exception {
...
+                conn.commit();
+            }
+            // set the max timestamp so that gap is 4*ttl - 1
+            injectEdge.setValue(startTime + (4*ttl*1000) - 1);
+            updateColumn(conn, tableName, "a1", 2, "col2_max"); // max timestamp
+            conn.commit();
+            String dql = String.format("SELECT * from %s where id='%s'",
+                    tableName, "a1");
+            ResultSet rs = conn.createStatement().executeQuery(dql);
+            assertTrue(rs.next());
```
kadirozde, line 407: We need to make this test to cover more code paths. I suggest that we include more columns and update these in a way that they will create gaps but still some of them do not expired because of other column updates and others expire.

kadirozde (approved):
+1, Thanks!

PR #2083, opened 2025-02-25
PHOENIX-7533 Fix broken compatibility for Zookeeper based ConnectionInfo

virajjasani:
Jira: PHOENIX-7533

Before PHOENIX-6523, ConnectionInfo used to provide hbase.zookeeper.quorum with Zookeeper host names separated by comma. With PHOENIX-6523, for zk registry, ConnectionInfo provides hbase.client.zookeeper.quorum with Zookeeper hosts and port values combined, separated by comma (e.g. "host1:1234,host2:2181,host3:9876" etc).

For the purpose of creating HBase Connection, providing "hbase.client.zookeeper.quorum" is sufficient. However, Phoenix downstream applications that are using "hbase.zookeeper.quorum" value to determine which HTable threadpool to use for the given Connection can no longer function for cases like dual client where different HTable threadpools need to be used for different HBase connections.

Given that Phoenix 5.1 client used to provide "hbase.zookeeper.quorum", 5.2 should not break this compatibility even though it provides "hbase.client.zookeeper.quorum" as additional Connection level configuration.

### DIFF
--- a/phoenix-core-client/src/main/java/org/apache/phoenix/jdbc/ZKConnectionInfo.java
+++ b/phoenix-core-client/src/main/java/org/apache/phoenix/jdbc/ZKConnectionInfo.java
@@ -17,7 +17,6 @@
 package org.apache.phoenix.jdbc;
 
 import java.sql.SQLException;
-import java.util.Arrays;
 import java.util.Map;
 import java.util.Properties;
 import java.util.StringTokenizer;
@@ -60,17 +59,6 @@ public class ZKConnectionInfo extends ConnectionInfo {
         return zkPort;
     }
 
-    public String getZkHostsWithoutPorts() {
-        if (zkHosts == null) {
-            return null;
-        }
-        String[] hostsWithoutPorts = zkHosts.split(",");
-        String[] hosts = Arrays.stream(hostsWithoutPorts)
-                .map(hostsWithoutPort -> hostsWithoutPort.split(":")[0].trim())
-                .toArray(String[]::new);
-        return String.join(",", hosts);
-    }
-
     public String getZkRootNode() {
         return zkRootNode;
     }
@@ -94,10 +82,7 @@ public class ZKConnectionInfo extends ConnectionInfo {
         if (getZkHosts() != null) {
             //This has the highest priority
             connectionProps.put(HConstants.CLIENT_ZOOKEEPER_QUORUM, getZkHosts());
-        }
-        String zkHostsWithoutPorts = getZkHostsWithoutPorts();
-        if (zkHostsWithoutPorts != null) {
-            connectionProps.put(HConstants.ZOOKEEPER_QUORUM, zkHostsWithoutPorts);
+            connectionProps.put(HConstants.ZOOKEEPER_QUORUM, getZkHosts());
         }
         //Port is already normalized into zkHosts
         if (getZkRootNode() != null) {
--- a/phoenix-core-client/src/main/java/org/apache/phoenix/jdbc/ZKConnectionInfo.java
+++ b/phoenix-core-client/src/main/java/org/apache/phoenix/jdbc/ZKConnectionInfo.java
@@ -17,6 +17,7 @@
 package org.apache.phoenix.jdbc;
 
 import java.sql.SQLException;
+import java.util.Arrays;
 import java.util.Map;
 import java.util.Properties;
 import java.util.StringTokenizer;
@@ -59,6 +60,17 @@ public class ZKConnectionInfo extends ConnectionInfo {
         return zkPort;
     }
 
+    public String getZkHostsWithoutPorts() {
+        if (zkHosts == null) {
+            return null;
+        }
+        String[] hostsWithoutPorts = zkHosts.split(",");
+        String[] hosts = Arrays.stream(hostsWithoutPorts)
+                .map(hostsWithoutPort -> hostsWithoutPort.split(":")[0].trim())
+                .toArray(String[]::new);
+        return String.join(",", hosts);
+    }
+
     public String getZkRootNode() {
         return zkRootNode;
     }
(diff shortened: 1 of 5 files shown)


stoty:
So the problem is that HBase normalizes the quorum, and you need it to be string equivalent to the one specified by the user ?

This solution is not good, as it would break when you have "hostA:100,hostb:100,hostc:100' and 'hostA:200,hostB:200,hostC:200'

What we COULD do is use a custom propery for the normalized quorum that we use internally to distinguish the connections, and leave set the HBase properties to the verbatim value specified by the user.

virajjasani:
@stoty this change is just restoring the version of what old client used to return. This is for backward compatibility purpose. Imagine, a downstream application using dual client connection using both 5.1 and 5.2 clients and expecting CQSI config to return same value for `hbase.zookeeper.quorum` depending on the primary/secondary cluster pairs defined somewhere.

Custom property is fine but by not doing this, not only we break compatibility, we return incorrect value of zk quorum config because unless overriden by ZkConnectionInfo, the value will be derived from cached config in ConnectionInfo.

stoty:
I disagree with your assumption that hbase.zookeeper.quorum and hbase.zookeeper.client.quorum have different semantics, and should have different values. IIRC the HBase behaviour is that hbase.zookeeper.client.quorum is the primary property, and hbase.zookeeper.quorum is the fallback if the client one is undefined, there is no other semantic difference, and I think we should honor that.

Phoenix should still behave consistently is new-style quorums with heterogenous prots are used, it should not set invalid quorum properties in that case.

You have a point in that  Phoenix should not change the quorum properties in coming either from hbase-site.xml or specified via the URL, but I think that you are trying to fix that in the wrong way.

I think that right solution is to leave the hbase-site.xml properties alone, and just naively copy the strings from the URL into both hbase.client.zookeeper.quorum and hbase.zookeeper.quorum (and for old-style URLs also set the port properties), AND somehow store phoenix-specific property that we can use to select the correctl CQSI object. 

This would only require changing ConnectionInfo so that it sets all the quorum-related properties based on the non-normalized quorum properties/URL components.

stoty:
Note that I've just committed some ZK (mostly test) changes, you may need to rebase.

virajjasani:
> hbase.zookeeper.client.quorum is the primary property, and hbase.zookeeper.quorum is the fallback if the client one is undefined, there is no other semantic difference, and I think we should honor that.

Yes, this is correct, and even with this change, this will be honored. ZkConnectionInfo#asProps() will return both of them. `hbase.zookeeper.client.quorum` will contain port info whereas for the purpose of backward compatibility, we want to remove port info from `hbase.zookeeper.quorum`. From the viewpoint of establishing HBase connection, there is no problem for patterns like `host1:123,host2:234,host3:345` because all port info from the url will be available to `hbase.zookeeper.client.quorum` and HBase connection will use it.

However, on the other hand, let's assume we use dual-client without this patch. In that case, given that ConnectionInfo already has cached `static volatile Configuration configuration`, if we don't provide `hbase.zookeeper.quorum`, both CQSI connections (one for primary and another for secondary), even having right URL, will provide CQSI connection config value for `hbase.zookeeper.quorum` same, because it will be derived from cached config.

> I think that right solution is to leave the hbase-site.xml properties alone, and just naively copy the strings from the URL into both hbase.client.zookeeper.quorum and hbase.zookeeper.quorum (and for old-style URLs also set the port properties), AND somehow store phoenix-specific property that we can use to select the correctl CQSI object.

This is what I thought earlier, but any downstream application dealing with both 5.1 and 5.2 client should expect same behavior. For instance, reading `hbase.zookeeper.quorum` value from 5.1 client CQSI config and then processing some data, and now after upgrade re-reading `hbase.zookeeper.quorum` value from 5.2 client CQSI config and using it to find out earlier computed value (e.g. thread pools to be used for HTable) can easily lead into problems.

I don't mind keeping same value for both `hbase.zookeeper.quorum` and `hbase.zookeeper.client.quorum`. It's the backward compatibility use case that makes me feel comfortable returning same value by both 5.1 and 5.2 clients.

Again, just to re-iterate, this has no impact w.r.t HBase connection creation because as you rightly mentioned above, HBase uses `hbase.zookeeper.client.quorum` if present and ignores `hbase.zookeeper.quorum`.

virajjasani:
In fact, since we return `hbase.zookeeper.client.quorum` with port info already, HBase ignores both `hbase.client.zookeeper.property.clientPort` and `hbase.zookeeper.property.clientPort`. So, overall HBase connection creation ignores three properties (including `hbase.zookeeper.quorum`) if `hbase.zookeeper.client.quorum` is provided with full url (i.e. `host:port` format separated by comma).

stoty:
I was planning on revisiting ConnectionInfo before 5.2.2 anyway.
Is this urgent ? Can we postpone thi to sometime next week when I will have refreshed my mind on ConnectionInfo issues ?

phoenix-core-client/src/main/java/org/apache/phoenix/jdbc/ZKConnectionInfo.java
```diff
@@ -83,6 +95,10 @@ public ReadOnlyProps asProps() {
             //This has the highest priority
             connectionProps.put(HConstants.CLIENT_ZOOKEEPER_QUORUM, getZkHosts());
         }
+        String zkHostsWithoutPorts = getZkHostsWithoutPorts();
+        if (zkHostsWithoutPorts != null) {
+            connectionProps.put(HConstants.ZOOKEEPER_QUORUM, zkHostsWithoutPorts);
```
jpisaac, line 100: @virajjasani We should be keeping this property the same as the  CLIENT_ZOOKEEPER_QUORUM, Right!, 
Since that's the ZK spec and should have the right format/spec to make the connection. The only backward compatibility issue was that it was not set earlier causing upstream application to misbehave.

stoty: If that's enough, that would certainly make life MUCH easier.

virajjasani: The backward compatibility issue is that 5.1 client never returned port number from ConnectionInfo whereas here if we don't do this, 5.2 client will always have port info.

virajjasani: Alright, let's keep them same with release note that downstreamer will have to take care of this.

tkhurana (approved):

virajjasani:
Thanks @tkhurana!
Merging this, considering the latest change as +1 from @jpisaac and @stoty as well.

PR #2085, opened 2025-02-26
PHOENIX-7470 Simplify Kerberos login in ConnectionInfo

stoty:


phoenix-core-client/src/main/java/org/apache/phoenix/jdbc/ConnectionInfo.java
```diff
@@ -493,11 +493,7 @@ protected void handleKerberosAndLogin() throws SQLException {
                                 LOGGER.info("Trying to connect to a secure cluster as {} "
                                         + "with keytab {}",
                                     principal, keytab);
-                                final Configuration newConfig = getConfiguration(principal, keytab);
-                                // We are intentionally changing the passed in Configuration object
-                                UserGroupInformation.setConfiguration(newConfig);
```
stoty, line 498: The added the possibility to configure the Hadoop UserGroupInformation from the connection URL / info objects.

That doesn't look like a good idea, but it is technically an incompatible change.

It is also a JVM global config, which makes it extra scary.

stoty:
This breaks some tests, I will look into that.

lokiore (approved):
LGTM, 
It did look scary that we have whitelisted UserGroupInformation from urls.

PR #2097, opened 2025-03-26
PHOENIX-7543 Wrong result returned when query is served by index and some columns are null

tkhurana:
Add DeleteColumn cells to index rows when a column is set to null in an upsert statement.


phoenix-core/src/it/java/org/apache/phoenix/end2end/index/GlobalIndexCheckerIT.java
```diff
@@ -1223,6 +1245,166 @@ public void testUnverifiedIndexRowWithFirstKeyOnlyFilter() throws Exception {
...
+                PhoenixResultSet prs = rs.unwrap(PhoenixResultSet.class);
+                String explainPlan = QueryUtil.getExplainPlan(prs.getUnderlyingIterator());
+                assertTrue(explainPlan.contains(indexName));
+                assertTrue(rs.next());
+                assertEquals("abc", rs.getString("val2"));
+            }
+
+            // update row ('a', 'ab', 'abc', null) -> ('a', 'ac', null, null)
+            conn.createStatement().execute(
+                    "upsert into " + dataTableName + " values ('a', 'ac', null, null)");
```
kadirozde, line 1288: val 3 was null already. We should skip updating it again.

kadirozde (approved):
Very minor change request on a test. Looks great!

PR #2105, opened 2025-04-09
PHOENIX-7574 Phoenix Compaction doesn't correctly handle DeleteFamily…

kadirozde:
…Version markers

### DIFF
--- a/phoenix-core-server/src/main/java/org/apache/phoenix/coprocessor/CompactionScanner.java
+++ b/phoenix-core-server/src/main/java/org/apache/phoenix/coprocessor/CompactionScanner.java
@@ -2179,6 +2179,7 @@ public class CompactionScanner implements InternalScanner {
         List<Cell> phoenixResult = new ArrayList<>();
         List<Cell> trimmedRow = new ArrayList<>();
         List<Cell> trimmedEmptyColumn = new ArrayList<>();
+        List<Cell> deleteFamilyVersionCellList = new ArrayList<>();
         private TTLTracker rowTracker;
 
         PhoenixLevelRowCompactor(TTLTracker rowTracker) {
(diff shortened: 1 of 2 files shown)


phoenix-core-server/src/main/java/org/apache/phoenix/coprocessor/CompactionScanner.java
```diff
@@ -2213,56 +2213,229 @@ class PhoenixLevelRowCompactor {
...
+                        lastRowVersion.add(cell);
+                    } else if (!major) {
+                        retainedCells.add(cell);
+                    }
+                    index = skipColumn(result, currentColumnCell, retainedCells, index);
+                    continue top;
+                }
+                if (cell.getType() == Cell.Type.Put) {
+                    lastRowVersion.add(cell);
+                    if (ScanUtil.isEmptyColumn(cell, emptyCF, emptyCQ)) {
```
kadirozde, line 2434: We do gap analysis for TTL processing and we expire rows only when the current compaction is a major compaction and so we should not track empty column cells when it is not a major compaction

phoenix-core-server/src/main/java/org/apache/phoenix/coprocessor/CompactionScanner.java
```diff
@@ -2213,56 +2213,229 @@ class PhoenixLevelRowCompactor {
...
+                while (cell.getType() == Cell.Type.Delete) {
+                    if (cell.getTimestamp() == maxLookbackWindowStart) {
                         lastRowVersion.add(cell);
+                    } else if (!major) {
+                        retainedCells.add(cell);
+                    }
+                    Cell nextCell = result.get(index + 1);
+                    if (!CellUtil.matchingColumn(currentColumnCell, nextCell)) {
+                        continue top;
+                    }
```
kadirozde, line 2415: At this moment, we have processed both the delete marker and the delete put cell so the index should be increment for both. So, we need to add index = index + 1 here.

phoenix-core-server/src/main/java/org/apache/phoenix/coprocessor/CompactionScanner.java
```diff
@@ -2213,56 +2213,229 @@ class PhoenixLevelRowCompactor {
...
+                    if (nextCell.getType() == Cell.Type.Put && cell.getTimestamp() == nextCell.getTimestamp()) {
```
kadirozde, line 2416: We need to consume this delete cell so index should be incremented here (review by @tkhurana)

tkhurana (approved):

PR #2111, opened 2025-04-14
PHOENIX-7170 Conditional TTL (addendum)

tkhurana:
Fix how we construct the latestRowVersion from the raw scan to evaluate the TTL expression.


kadirozde (approved):
+1, Thanks!

PR #2115, opened 2025-04-15
PHOENIX-7565 Replication log file format and replication log file writer and reader classes

apurtell:


apurtell:
I actually cannot build Phoenix master branch on the command line, I get this:

[ERROR] Failed to execute goal org.apache.maven.plugins:maven-dependency-plugin:3.1.1:analyze-only (enforce-dependencies) on project phoenix-core-server: Dependency problems found -> [Help 1]
org.apache.maven.lifecycle.LifecycleExecutionException: Failed to execute goal org.apache.maven.plugins:maven-dependency-plugin:3.1.1:analyze-only (enforce-dependencies) on project phoenix-core-server: Dependency problems found

without any indication of what the problem is. My IDE (Eclipse) can build and run the new code and new unit tests, which pass. 

And it seems like yetus/precommit here doesn't know that phoenix-core-server is a dependency of phoenix-core and so needs to be compiled first, in order for the phoenix-core tests to have the new package and classes available.

Any idea what is going on here? Is it just for me that Phoenix master branch won't build? @kadirozde @virajjasani

apurtell:
Rebased on latest master, changes on PR branch squashed.

phoenix-core-server/src/main/java/org/apache/phoenix/replication/log/LogFile.java
```diff
@@ -0,0 +1,236 @@
...
+        /** Current version of the replication log block header */
+        public static final byte VERSION = 1;
+        int getVersion();
+        Compression.Algorithm getCompression();
+        BlockHeader setCompression(Compression.Algorithm compression);
+        int getUncompressedSize();
+        BlockHeader setUncompressedSize(int uncompressedSize);
+        int getCompressedSize();
+        BlockHeader setCompressedSize(int compressedSize);
+        int getSerializedLength();
```
kadirozde, line 74: Adding comments for size and length methods will help here. Size methods are the block payload/data and the length method is for the block header. Maybe we can even rename them accordingly, for example getUncompressedDataSize, getSerializedHeaderLength

phoenix-core-server/src/main/java/org/apache/phoenix/replication/log/LogFile.java
```diff
@@ -0,0 +1,236 @@
...
+        Trailer setBlocksStartOffset(long offset);
+        long getTrailerStartOffset();
+        Trailer setTrailerStartOffset(long offset);
+        int getSerializedLength();
+        void readFields(DataInput in) throws IOException;
+        void write(DataOutput out) throws IOException;
+    }
+
+    /** Represents a single logical change */
+    public interface Record {
```
kadirozde, line 99: A record maps to an HBase mutation. I think it will be more efficient and simpler if the Phoenix coproc through ReplicationLogWriter passes an HBase mutation directly to create a record and also gets back a mutation when ReplicationLogReader reads a record. Then the in-memory representation of a record is composed of mainly an HBase mutation, commit id, and schema and there will be no need to translate a mutation to a record outside the log file implementation.

kadirozde: I agree with you on what you wrote. I did not mean to use the HBase code to serialize and deserialize mutations. Mutation is a public API object and our code should stay at that level. I observed that if the mutation is also the API for the log file library this PR builds then it would simplify the Phoenix code using this API and also simplify the library a bit, and at the same time it would make the solution more efficient. For example, the library would not need to redefine mutation type and would not need to keep track of columns using a separate map structure as mutation already has this map structure. This should reduce memory allocations.

apurtell: Ok, I am working on a commit for this PR that addresses this review feedback:
 - Document LogFile interfaces with javadoc
 - Directly accept HBase Mutation objects in the Writer
 - Directly provide HBase Mutation objects from the Reader

Additional changes:
 - Optimize the encoding and decoding of column family and qualifiers, to avoid string concat/splitting, changing the file format. Spec requires update!
 - Encode records with a single pass. Saves on lots of duplicate structure enumeration, requiring a small copy in trade.
 - LogFileCompressionTest test fix.
 
I will push it up soon.

apurtell: I pushed up the changes that resulted from this feedback as d34d4dc90 . 
I need to add more tests for serde of the various ways Put and Delete mutations might be constructed.
@kadirozde

phoenix-core-server/src/main/java/org/apache/phoenix/replication/log/LogFile.java
```diff
@@ -0,0 +1,579 @@
...
+        /**
+         * Writes the trailer fields to the provided DataOutput stream.
+         * @param out The DataOutput stream to write to.
+         * @throws IOException if an I/O error occurs.
+         */
+        void write(DataOutput out) throws IOException;
+    }
+
+    /** Represents a single logical change */
+    interface Record {
```
kadirozde, line 280: Can we simplify the record interface as follows? This will be the interface for the in-memory representation of the record. The on-disk representation of the record will be as described in the design document and will be used to serialize the in-memory representation.
```
    interface Record {
         /**
          * Gets the mutation this record represents.
          * @return The Mutation.
          */
         Mutation getMutation();
 
         /**
          * Sets the mutation this record represents.
          * @param mutation The Mutation to set.
          * @return This Record instance for chaining.
          */
         Record setMutation(Mutation mutation);
 
         /**
          * Gets the name of the HBase table this record pertains to.
          * @return The HBase table name.
          */
         String getHBaseTableName();
 
         /**
          * Sets the name of the HBase table this record pertains to.
          * @param tabletName The HBase table name to set.
          * @return This Record instance for chaining.
          */
         Record setHBaseTableName(String tableName);
 
         /**
          * Gets the commit ID or System Change Number (SCN) associated with this record.
          * @return The commit ID/SCN.
          */
         long getCommitId();
 
         /**
          * Sets the commit ID or System Change Number (SCN) for this record.
          * @param commitId The commit ID/SCN to set.
          * @return This Record instance for chaining.
          */
         Record setCommitId(long commitId);
  
         /**
          * Gets the total serialized length of this record in bytes, including any length prefixes
          * used by the codec. This value should be set by the codec after writing or reading.
          * @return The total serialized length of the record.
          */
         int getSerializedLength();
 
         /**
          * Sets the total serialized length of this record. This is typically called by the codec.
          * @param serializedLength The total serialized length in bytes.
          * @return This Record instance for chaining.
          */
         Record setSerializedLength(int serializedLength);
         }
```

phoenix-core-server/src/main/java/org/apache/phoenix/replication/log/LogFile.java
```diff
@@ -0,0 +1,579 @@
...
+         */
+        static Mutation toHBaseMutation(Record record) {
+            if (record == null) {
+                return null;
+            }
+            return LogFileRecord.toHBaseMutation(record);
+        }
+    }
+
+    /**
```
kadirozde, line 416: When we simplify the Record interface as in my previous comment, we do not need redefine the MutationType, I think.

phoenix-core-server/src/main/java/org/apache/phoenix/replication/log/LogFile.java
```diff
@@ -0,0 +1,579 @@
...
+        void init(LogFileWriterContext context) throws IOException;
+
+        /**
+         * Appends an HBase mutation to the log file. The log record may be buffered internally.
+         * @param schemaObjectName The schema object name
+         * @param commitId The commit identifier
+         * @param mutation The mutation to append.
+         * @throws IOException if an I/O error occurs during append.
+         */
+        void append(String schemaObjectName, long commitId, Mutation mutation) throws IOException;
```
kadirozde, line 480: schemaObjectName should be renamed as hbaseTableName or physicalTableName. This object is not a Phoenix object.

phoenix-core-server/src/main/java/org/apache/phoenix/replication/log/LogFileRecord.java
```diff
@@ -0,0 +1,364 @@
...
+import org.apache.hadoop.hbase.client.Delete;
+import org.apache.hadoop.hbase.client.Increment;
+import org.apache.hadoop.hbase.client.Mutation;
+import org.apache.hadoop.hbase.client.Put;
+import org.apache.hadoop.hbase.util.Bytes;
+import org.apache.phoenix.replication.log.LogFile;
+
+public class LogFileRecord implements LogFile.Record {
+
+    private LogFile.MutationType mutationType;
```
kadirozde, line 39: Similar the Record interface simplification, can we replace
```

    private LogFile.MutationType mutationType;
    private String schemaObjectName;
    private long commitId;
    private byte[] row;
    private long timestamp;
    // Nested map: Family -> Qualifier -> Value
    private Map<byte[], Map<byte[], byte[]>> familyMap;
    private int serializedLength; 
```

with 

```
    private Mutation mutation;
    private String hbaseTableName;
    private long commitId;
    private int serializedLength;
```

kadirozde:
I suggested further simplification. The idea behind this simplification is to directly translate Mutation to the on-disk representation of Record, and translate the on-disk representation back to Mutation without separately tracking columns, the row key, and the timestamp in memory.

apurtell:
To your point I can see substituting:

    private byte[] row;
    private long timestamp;
    // Nested map: Family -> Qualifier -> Value
    private Map<byte[], Map<byte[], byte[]>> familyMap;
    
 with
 
     private Mutation mutation
     
 
This is the gist of your suggestion, I think.

A goal has been to avoid using HBase types directly because when HBase changes them Phoenix will be impacted. However Mutation is a Public annotated class so it might be fine.

apurtell:
The failing unit tests do not involve this change. We know this because this change is for the moment totally decoupled from the rest of Phoenix and has its own unit tests.

virajjasani:
FYI @apurtell the feature branch: https://github.com/apache/phoenix/tree/PHOENIX-7562-feature

The target branch for this PR should be `PHOENIX-7562-feature` instead of `master`.

phoenix-core-server/src/main/java/org/apache/phoenix/replication/log/LogFileCodec.java
```diff
@@ -0,0 +1,269 @@
...
+import org.apache.hadoop.io.WritableUtils;
+
+/**
+ * Default Codec for encoding and decoding ReplicationLog Records within a block buffer.
+ * This implementation uses standard Java DataInput/DataOutput for serialization.
+ *
+ * Record Format within a block:
+ * - Total Record Length (vint)
+ * - Mutation Type (byte)
+ * - Schema Object Name Length (vint)
```
kadirozde, line 48: HBase Table Name

phoenix-core-server/src/main/java/org/apache/phoenix/replication/log/LogFileCodec.java
```diff
@@ -0,0 +1,269 @@
...
+            WritableUtils.writeVInt(recordOut, rowKey.length);
+            recordOut.write(rowKey);
+            recordOut.writeLong(mutation.getTimestamp());
+
+            Map<byte[], List<Cell>> familyMap = mutation.getFamilyCellMap();
+            int colCount = familyMap.size();
+            WritableUtils.writeVInt(recordOut, colCount);
+
+            for (Map.Entry<byte[], List<Cell>> entry: familyMap.entrySet()) {
+                byte[] column = entry.getKey();
```
kadirozde, line 119: This is a column family, so we should rename it to say cf or columnFamily

phoenix-core-server/src/main/java/org/apache/phoenix/replication/log/LogFileCodec.java
```diff
@@ -0,0 +1,269 @@
...
+ * Record Format within a block:
+ * - Total Record Length (vint)
+ * - Mutation Type (byte)
+ * - Schema Object Name Length (vint)
+ * - Schema Object Name Bytes (byte[])
+ * - Commit ID / SCN (vlong)
+ * - Row Key Length (vint)
+ * - Row Key Bytes (byte[])
+ * - Timestamp (long)
+ * - Number of Columns (vint)
```
kadirozde, line 54: Need to update to include a column family name before column name-values.

phoenix-core-server/src/main/java/org/apache/phoenix/replication/log/LogFileCodec.java
```diff
@@ -0,0 +1,269 @@
...
+            WritableUtils.writeVInt(recordOut, nameBytes.length);
+            recordOut.write(nameBytes);
+            WritableUtils.writeVLong(recordOut, record.getCommitId());
+            byte[] rowKey = mutation.getRow();
+            WritableUtils.writeVInt(recordOut, rowKey.length);
+            recordOut.write(rowKey);
+            recordOut.writeLong(mutation.getTimestamp());
+
+            Map<byte[], List<Cell>> familyMap = mutation.getFamilyCellMap();
+            int colCount = familyMap.size();
```
kadirozde, line 115: This should renamed to cfCount

phoenix-core-server/src/main/java/org/apache/phoenix/replication/log/AsyncFSDataOutput.java
```diff
@@ -0,0 +1,150 @@
...
+    @Override
+    public void writeShort(int v) throws IOException {
+        writeByte((v >>> 8) & 0xFF);
+        writeByte((v >>> 0) & 0xFF);
+    }
+
+    @Override
+    public void writeChar(int v) throws IOException {
+        writeByte((v >>> 8) & 0xFF);
+        writeByte((v >>> 0) & 0xFF);
```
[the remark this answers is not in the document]
apurtell: This is what DataOutputStream does in the JRE.

apurtell:
The target branch for this PR should be PHOENIX-7562-feature instead of master.

Too late, I'd have to open a new PR. How do I find out about feature branches?

apurtell:
Hey, I'm going to have to open a new PR because this is targeting the wrong branch.

apurtell:
New PR 
https://github.com/apache/phoenix/pull/2133

@kadirozde I will reflect your latest feedback over there.

PR #2125, opened 2025-04-22
Revert: PHOENIX-7006: Configure maxLookbackAge at table level

sanjeet006py:


virajjasani (approved):
+1

virajjasani:
Needs rebase :)

sanjeet006py:
> One downside is that maxlookback for indexes is not being inherited from base table. This could be a problem for CDC indexes. We will have to explicitly set maxlookback on indexes.

Yeah, I thought we were fine this limitation while reverting this change.

In general I see two limitations with reverting (noticed these in ITs):
- Indexes are not inheriting table level max lookback and for CDC indexes we need to set it explicitly on a hidden index.
- QueryCompiler doesn't prevent a user from setting SCN lesser than table level max lookback as it doesn't read HBase table attributes.

 Shall I create a JIRA for extending setting table level max lookback in HBase table attribute as a complete feature?

virajjasani:
While for other indexes, we can manually set HBase TableDescriptor level maxLookback config. For CDC index, we usually don't make it as visible as other indexes. I think having a follow-up Jira to set the CDC index TTL at HBase TableDescriptor level from base table maxLookback will be useful, which can be later extended to other indexes also.
WDYT @tkhurana @sanjeet006py?

virajjasani:
Test failures on the build don't seem relevant. This is good to go?

tkhurana (approved):

virajjasani:
I hope we don't need to backport this to 5.2.

sanjeet006py:
@virajjasani we don't need to backport this to 5.2. Thanks

PR #2128, opened 2025-04-24
PHOENIX-7586 :- Handle Role transitions for ActiveToStanby role in Failover HAPolicy

lokiore:


ritegarg (approved):

phoenix-core-client/src/main/java/org/apache/phoenix/jdbc/HighAvailabilityPolicy.java
```diff
@@ -47,6 +48,23 @@ public Connection provide(HighAvailabilityGroup haGroup, Properties info,
             return new FailoverPhoenixConnection(context);
         }
 
+        /**
+         * Cluster Role Transitions for Failover High Availability Policy, Here we are trying to
+         * close connections late to allow existing reads to continue during Failover.
+         * ACTIVE --> ACTIVE_TO_STANDBY (Doing Nothing as we are in process of moving the current
```
shahrs87, line 54: how about state transition from ACTIVE_TO_STANDBY --> ACTIVE? We will need this to rollback state change from ACTIVE --> ACTIVE_TO_STANDBY.

lokiore: ACTIVE --> ACTIVE_TO_STANDBY is not doing anything here, as we just want to block writes and nothing else, So if we go to ACTIVE_TO_STANDBY then rolling back should be same (nothing required), 
I see I missed commenting that will add the comment :v:

phoenix-core/src/it/java/org/apache/phoenix/jdbc/FailoverPhoenixConnection2IT.java
```diff
@@ -470,6 +474,75 @@ public void testAllWrappedConnectionsNotClosedAfterStandbyURLChange() throws Exc
...
+
+        //Creating some connections to ACTIVE cluster
+        short numberOfConnections = 10;
+        //Create FailoverPhoenixConnections with default urls
+        List<Connection> connectionList = new ArrayList<>(numberOfConnections);
+        for (short i = 0; i < numberOfConnections; i++) {
+            connectionList.add(createFailoverConnection());
+        }
+
+        //Transit Role1 ACTIVE --> ACTIVE_TO_STANDBY
```
shahrs87, line 501: I would just add one more test case: Check when  we transition a cluster from ATS --> Active. It should not close any connections in this case also.

shahrs87 (approved):
Just one comment. Other than that, looks good.
Thank you @lokiore  for the patch.

phoenix-core/src/it/java/org/apache/phoenix/jdbc/FailoverPhoenixConnection2IT.java
lokiore: Thanks, I @shahrs87 I have added one more test for early rollback Active --> ATS --> Active which should not affect connections at all

PR #2133, opened 2025-04-29
PHOENIX-7565 Replication log file, and replication log file writer and reader classes

apurtell:


phoenix-core-server/src/main/java/org/apache/phoenix/replication/log/LogFileRecord.java
```diff
@@ -0,0 +1,181 @@
...
+        if (obj == null) {
+            return false;
+        }
+        if (getClass() != obj.getClass()) {
+            return false;
+        }
+        LogFileRecord other = (LogFileRecord) obj;
+        return tableName.equals(other.tableName)
+            && commitId == other.commitId
+            && mutation.toString().equals(other.toString());
```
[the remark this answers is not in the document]
apurtell: If only HBase mutation had a reasonable equals, this would not be necessary! Fortunately this is good enough.

apurtell:
Just to be clear about this...

Copilot says
`HDFSDataOutput.java`, `FSDataInput.java`, `AsyncFSDataOutput.java` : Provide concrete implementations for HDFS and asynchronous I/O support.


`AsyncFSDataOutput` uses HBase's `hbase-asyncfs` to **synchronously** write logs using `hbase-asyncfs`'s drastically simpler (and therefore, more robust) custom HDFS writer. There is nothing _async_ about this in Phoenix. Nonetheless the name makes sense to me because what we are using here as the understream is `AsyncFSOutput` from `hbase-asyncfs`.

apurtell:
For checkstyle findings like this:

    ./phoenix-core-server/src/main/java/org/apache/phoenix/replication/log/AsyncFSDataOutput.java:2: * Licensed to the Apache Software Foundation (ASF) under one: Line does not match expected header line of ' * Licensed to the Apache Software Foundation (ASF) under one or more'. [Header]

This rule needs updating. A far as I can tell based on some spot checks all of the Phoenix source files do not conform to this rule. I myself cut and pasted the license header for the new files from adjacent existing Phoenix source.

phoenix-core-server/src/main/java/org/apache/phoenix/replication/log/LogFileCodec.java
```diff
@@ -0,0 +1,288 @@
...
+        private LogFileRecord current = null;
+
+        RecordDecoder(DataInput in) {
+            this.in = in;
+        }
+
+        @Override
+        public boolean advance(LogFile.Record reuse) throws IOException {
+            try {
+                int recordDataLength = WritableUtils.readVInt(in);
```
kadirozde, line 182: Instead of using an EOFException to determine the end of a block, should we keep track of the position within a block to improve CPU utilization?

apurtell: How would that improve CPU utilization?

phoenix-core-server/src/main/java/org/apache/phoenix/replication/log/LogFile.java
```diff
@@ -0,0 +1,439 @@
...
+         * @param offset The starting offset of the first block.
+         * @return This Trailer instance for chaining.
+         */
+        Trailer setBlocksStartOffset(long offset);
+
+        /**
+         * Gets the byte offset within the file where the trailer itself begins.
+         * @return The starting offset of the trailer.
+         */
+        long getTrailerStartOffset();
```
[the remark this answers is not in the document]
apurtell: The trailer does begin with metadata, but the offset is to the start of the entire trailer, not only its metadata.

phoenix-core-server/src/main/java/org/apache/phoenix/replication/log/LogFileReaderContext.java
```diff
@@ -0,0 +1,158 @@
...
+    /** Default for skipping corrupt blocks */
+    public static final boolean DEFAULT_LOGFILE_SKIP_CORRUPT_BLOCKS = true;
+
+    private final Configuration conf;
+    private FileSystem fs;
+    private Path path;
+    private LogFileCodec codec;
+    private long fileSize = -1;
+    private boolean isSkipCorruptBlocks;
+    private AtomicLong blocksRead = new AtomicLong();
```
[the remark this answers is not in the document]
apurtell: I had this idea that the context might be visible to multiple threads even if the reader is single threaded. But sure I can make this change.

phoenix-core-server/src/main/java/org/apache/phoenix/replication/log/LogFileCodec.java
kadirozde: I could not see the check for the end of the current block. Where is it?

phoenix-core-server/src/main/java/org/apache/phoenix/replication/log/LogFile.java
```diff
@@ -0,0 +1,439 @@
...
+         * Returns the current length of the log file being written, in bytes.
+         * This typically reflects the position of the underlying output stream.
+         * @return The current file length.
+         * @throws IOException if the length cannot be determined.
+         */
+        long getLength() throws IOException;
+    }
+
+    /** Interface for reading replication logs */
+    interface Reader extends Closeable, Iterable<Mutation> {
```
kadirozde, line 353: It is not sufficient just to read the mutation part of a record. Phoenix needs to know the HBase table name for this mutation.

apurtell: That's why originally I had this returning Record :-)

apurtell: Let me put that back...

phoenix-core-server/src/main/java/org/apache/phoenix/replication/log/LogFileCodec.java
apurtell: This is the Codec, block level concepts are not available here, like current block start and end offsets. A codec is given a stream and it reads from that stream a structure that is self describing in the sense that internally any field has a length preceding it. At any point, any read may throw EOFException though, that cannot be removed.

kadirozde: I have looked there too. There is a check for the end of the data offset. That is essentially checking the end of file essentially. There we read the block header and payload. If the payload is compressed, we decompressed it. The block now in memory as a `ByteBuffer`. We know the buffer size and can check if the end of buffer is reached before we call `advance`. However, we do not do this check.

apurtell:
I am adding unit tests for edge cases. I'll have a new push up this afternoon with many more test units.

apurtell:
I added more unit tests:

- Tests for likely edge cases in LogFileCodecTest.
- Tests for more corruption cases in LogFileFormatTest
- LogFileWriterSyncTest: Verify that our sync() really does cause FSDataOutputStream to call hsync on the underlying HDFS output. Ensures we get the durability we expect from our stack of indirection around output streams and sync().

apurtell:
I added more unit tests:

- Tests for likely edge cases in LogFileCodecTest.
- Tests for more corruption cases in LogFileFormatTest
- New unit LogFileWriterSyncTest: Verify that our sync() really does cause FSDataOutputStream to call hsync on the underlying HDFS output. Ensures we get the durability we expect from our stack of indirection around output streams and sync().

phoenix-core-server/src/main/java/org/apache/phoenix/replication/log/LogFileFormatReader.java
```diff
@@ -0,0 +1,355 @@
...
+
+    private static final Logger LOG = LoggerFactory.getLogger(LogFileFormatReader.class);
+
+    private LogFileReaderContext context;
+    private LogFile.Codec.Decoder decoder;
+    private SeekableDataInput input;
+    private LogFile.Header header;
+    private LogFile.Trailer trailer = null;
+    private long currentPosition = 0;
+    private ByteBuffer currentBlockBuffer = null;
```
kadirozde, line 48: We can track the position within the current block and use that position tracking to decide if we reached the end of the current block. If so, we eliminate an advance call and thus an end of file exception when the current block is consumed.

apurtell: > eliminate an advance call and thus an end of file exception when the current block is consumed.

Sure.

The Codec layer doesn't know about Blocks, so the EOFException handling in Codec should remain for the code there to be properly defensive. However my understanding is that is not the issue, it is the extra advance() call that is the issue, as you have indicated.

kadirozde (approved):
+1, Thanks!

apurtell:
There is one relevant precommit test failure here. 

    [ERROR] Tests run: 18, Failures: 0, Errors: 1, Skipped: 0, Time elapsed: 21.00 s <<< FAILURE! -- in org.apache.phoenix.replication.log.LogFileCodecTest
    [ERROR] org.apache.phoenix.replication.log.LogFileCodecTest.testCodecWithVeryLargeRecord -- Time elapsed: 17.58 s <<< ERROR!
    java.lang.OutOfMemoryError: Java heap space

I also saw NullValueTest OOME so I think the JVM heap is quite limited and I will just remove this test case. It's not really necessary. It is testing an unrealistic edge case.

apurtell:
The unit tests for this change pass. An unrelated unit test is failing for something to do with CDC. The remaining checkstyle errors are false positive findings related to ASF license headers. I think this is good to merge.

virajjasani:
Before merging, good to sync the feature branch with master branch, in case this helps with conflict resolution in advance before merging the PR.

apurtell:
Rebase complete. Merging this.

PR #2134, opened 2025-04-29
PHOENIX-7593: Enable CompactionScanner for flushes

sanjeet006py:


phoenix-core-server/src/main/java/org/apache/phoenix/coprocessor/CompactionScanner.java
```diff
@@ -416,11 +417,32 @@ private void postProcessForConditionalTTL(List<Cell> result) {
...
+
     @Override
     public boolean next(List<Cell> result) throws IOException {
         boolean hasMore = storeScanner.next(result);
         inputCellCount += result.size();
         if (!result.isEmpty()) {
+            // This will happen only during flushes as then we don't pass PTable object
+            // to determine emptyCF and emptyCQ
+            if (emptyCQ == EMPTY_BYTE_ARRAY) {
+                determineEmptyCfCq(result);
```
tkhurana, line 444: For not emptycfstore aren't we doing this check on every row

phoenix-core/src/it/java/org/apache/phoenix/end2end/TableTTLIT.java
```diff
@@ -273,10 +273,12 @@ public void testMinorCompactionShouldNotRetainCellsWhenMaxLookbackIsDisabled()
                         Thread.sleep(1);
                     }
                     flush(TableName.valueOf(tableName));
-                    // Flushes dump and retain all the cells to HFile.
-                    // Doing MAX_COLUMN_INDEX + 1 to account for empty cells
-                    assertEquals(TestUtil.getRawCellCount(conn, TableName.valueOf(tableName), row),
-                            rowUpdateCounter * (MAX_COLUMN_INDEX + 1));
+                    // At every flush, extra cell versions should be removed.
```
kadirozde, line 276: We need to rename this test method to testMinorCompactionAndFlushShouldNotRetainCellsWhenMaxLookbackIsDisabled

phoenix-core-server/src/main/java/org/apache/phoenix/coprocessor/CompactionScanner.java
sanjeet006py: Yeah, will fix that. Thanks

phoenix-core/src/it/java/org/apache/phoenix/end2end/MaxLookbackExtendedIT.java
```diff
@@ -610,6 +610,44 @@ public void testRetainingLastRowVersion() throws Exception {
...
+            conn.commit();
+            injectEdge.incrementValue(timeIntervalBetweenTwoUpserts * 1000);
+            stmt.execute("upsert into " + tableName + " values ('a', 'ab2')");
+            conn.commit();
+            injectEdge.incrementValue(timeIntervalBetweenTwoUpserts * 1000);
+            stmt.execute("upsert into " + tableName + " values ('a', 'ab3')");
+            conn.commit();
+            injectEdge.incrementValue(MAX_LOOKBACK_AGE * 1000);
+            TestUtil.dumpTable(conn, dataTableName);
+            TestUtil.flush(utility, dataTableName);
```
kadirozde, line 637: Let's use TestUtil.getRawCellCount and verify that extra row versions are removed.

phoenix-core/src/it/java/org/apache/phoenix/end2end/MaxLookbackExtendedIT.java
```diff
@@ -610,6 +610,44 @@ public void testRetainingLastRowVersion() throws Exception {
...
+            injectEdge.incrementValue(1);
+            TestUtil.dumpTable(conn, dataTableName);
+            majorCompact(dataTableName);
+            injectEdge.incrementValue(1);
+            TestUtil.dumpTable(conn, dataTableName);
+            ResultSet rs = stmt.executeQuery("select * from " + dataTableName + " where id = 'a'");
+            while(rs.next()) {
+                assertNotNull(rs.getString(3));
```
kadirozde, line 645: Let's verify the column values are equal to the expected values here.

phoenix-core-server/src/main/java/org/apache/phoenix/coprocessor/CompactionScanner.java
```diff
@@ -180,13 +181,13 @@ public CompactionScanner(RegionCoprocessorEnvironment env,
         // Empty column family and qualifier are always needed to compute which all empty cells to retain
         // even during minor compactions. If required empty cells are not retained during
         // minor compactions then we can run into the risk of partial row expiry on next major compaction.
-        this.emptyCF = SchemaUtil.getEmptyColumnFamily(table);
-        this.emptyCQ = SchemaUtil.getEmptyColumnQualifier(table);
+        this.emptyCF = table != null ? SchemaUtil.getEmptyColumnFamily(table) : EMPTY_BYTE_ARRAY;
+        this.emptyCQ = table != null ? SchemaUtil.getEmptyColumnQualifier(table) : EMPTY_BYTE_ARRAY;
```
virajjasani, line 185: Why not keep emptyCF and emptyCQ as null if PTable is null, so that we can also incorporate this logic?

Instead of
```
                    if (ScanUtil.isEmptyColumn(cell, emptyCF, emptyCQ)) {
                        index = addEmptyColumn(result, currentColumnCell, index, emptyColumn);
                    } else {
                        index = skipColumn(result, currentColumnCell, retainedCells, index);
                    }
```

this
```
                    if (emptyCF != null && emptyCQ != null && ScanUtil.isEmptyColumn(cell, emptyCF,
                            emptyCQ)) {
                        index = addEmptyColumn(result, currentColumnCell, index, emptyColumn);
                    } else {
                        index = skipColumn(result, currentColumnCell, retainedCells, index);
                    }
```

and similarly, `if (emptyCQ == EMPTY_BYTE_ARRAY)` too will be simple null check

sanjeet006py: By keeping `emptyCF` and `emptyCQ` values as `null` are we trying to optimize the `if` check? I actually kept it empty byte array to avoid null handling and nothing will match empty byte array.

sanjeet006py: > I don't think EMPTY_BYTE_ARRAY is allowed as CF:CQ, but while debugging, null check will be more readable rather than using incorrect values of emptyCF and emptyCQ for ScanUtil.isEmptyColumn?

Got it, will change to storing null values. I agree this improves readability. Thanks

virajjasani:
Not related to this PR, but as a general improvement, this method should not be named as `isEmptyColumn()` because it does not perform any empty column related check, all it checks for is whether the given cell has matching CF and CQ:
```
    public static boolean isEmptyColumn(Cell cell, byte[] emptyCF, byte[] emptyCQ) {
        return CellUtil.matchingFamily(cell, emptyCF, 0, emptyCF.length) &&
               CellUtil.matchingQualifier(cell, emptyCQ, 0, emptyCQ.length);
    }
```

We should remove the above utility because HBase CellUtil already provides exactly the same:
```
  public static boolean matchingColumn(final Cell left, final byte[] fam, final byte[] qual) {
    return matchingFamily(left, fam) && matchingQualifier(left, qual);
  }
```

sanjeet006py:
> (worth doing as separate Jira/PR though)

Created JIRA: https://issues.apache.org/jira/browse/HBASE-29278

phoenix-core-server/src/main/java/org/apache/phoenix/coprocessor/UngroupedAggregateRegionObserver.java
```diff
@@ -611,6 +612,29 @@ private boolean areMutationsInSameTable(Table targetHTable, Region region) {
                 region.getTableDescriptor().getTableName().getName()) == 0);
     }
 
+    @Override
+    public InternalScanner preFlush(ObserverContext<RegionCoprocessorEnvironment> c, Store store,
+                                    InternalScanner scanner, FlushLifeCycleTracker tracker)
+            throws IOException {
+        if (!isPhoenixTableTTLEnabled(c.getEnvironment().getConfiguration())) {
```
tkhurana, line 619: We are using the same config to control both flushing and compaction. Will there be a scenario where we want to disable Phoenix compaction on flushing but still continue to use Phoenix compaction for major/minor compaction ?

tkhurana:
@sanjeet006py Can you also do a perf study to rule out any performance degradation that can get introduced in the flushing path. We have some metrics at the regionserver like `hbase.regionserver.FlushTime`

phoenix-core-server/src/main/java/org/apache/phoenix/coprocessor/UngroupedAggregateRegionObserver.java
sanjeet006py: > Will there be a scenario where we want to disable Phoenix compaction on flushing but still continue to use Phoenix compaction for major/minor compaction ?

I think in general it will be good to have this flexibility. Shall I introduce a new config to enable disable `preFlush` hook separately?

virajjasani: This flexibility will be helpful. I think we can go ahead with merging the changes after config is added and perf can be done later. The config will anyways be helpful to turn off the feature, WDYT @tkhurana?

sanjeet006py: If we are fine with waiting for a day or 2 more then I can post the perf results. For single column family its done. I am currently, doing perf analysis of multi-CF. Thanks

sanjeet006py: @virajjasani perf analysis for multi-CF will take some time as I got to know that HBase flushTime metrics for multi-CF sometimes case could end up tracking combined time for flushing multiple CFs and other times only 1 CF. So, need to directly track time taken by StoreFlusher. 
One idea is we wait for perf analysis before merging this PR but 5.3 release can go on or if we want this PR in 5.3 then for now we can disable preFlush hook via config and only enable it after perf analysis. I am bit inclined towards first approach. WDYT @virajjasani @tkhurana?

tkhurana: Also, we don't need absolute numbers but just a comparison to make sure nothing is regressed.

sanjeet006py: > Does that have the same problem too ?

Yes as this log line only tells us how much time all the stores which were to be flushed took to flush data as total. I added a log line in StoreFlusher and will test with that now.

sanjeet006py:
@tkhurana @virajjasani the perf analysis is done: https://docs.google.com/document/d/1oQzEMP4LXOFxLHlKt1SZ5uvRLd3Vk90x39gn1hVBn0Y/edit?tab=t.0#heading=h.32xuccojgowv. Overall I see the overhead of enabling CompactionScanner for flushes will have some overhead (as expected) but no big enough to cause performance degradation. Thanks

phoenix-core-server/src/main/java/org/apache/phoenix/coprocessor/UngroupedAggregateRegionObserver.java
sanjeet006py: Added the config to disable CompactionScanner for flushes.

PR #2141, opened 2025-05-07
PHOENIX-7517 Allow BSON as primary key column

virajjasani:
Jira: PHOENIX-7517

PHOENIX-7330 introduced BSON as Document data type in Phoenix. Due to the complexity of the data it represents in the document format, the data type is restricted to be part of primary key columns. However, this restriction is not necessary as BSON could provide some value by being part of last PK column of the composite PK columns. One use case that can possibly have such schema is time-series database where metric tags could be provided in the document format and they can be stored together based on the partition key (rowkey prefix, non-document data type).

The purpose of this Jira is to allow BSON data type to be part of primary key columns.


phoenix-core/src/it/java/org/apache/phoenix/end2end/Bson4IT.java
```diff
@@ -426,6 +428,127 @@ public void testConditionalUpsertReturnRow() throws Exception {
...
+      assertTrue(rs.next());
+      assertEquals("1010", rs.getString(2));
+      BsonDocument document2 = (BsonDocument) rs.getObject(1);
+      assertEquals(bsonDocument2, document2);
+
+      assertFalse(rs.next());
+
+      // TODO : Fix this separately, using CAST with PK column results into full table scan due
+      //  to bug.
+      // validateExplainPlan(stmt, query, tableName, "POINT LOOKUP ON 1 KEY ");
```
virajjasani, line 489: This is an existing bug with using CAST against pk column, it turns out to be full table scan. I will file separate Jira to fix this.

kadirozde (approved):
+1, Thanks!

PR #2144, opened 2025-05-09
PHOENIX-7567 Replication Log Writer (Synchronous mode)

apurtell:


apurtell:
This PR is still in Draft state because I want to add more unit tests.

apurtell:
Ok, at 2ed948cf2 this is a known good state with all existing unit tests passing. Now I need to add more tests.

phoenix-core-server/src/main/java/org/apache/phoenix/replication/ReplicationLogWriter.java
```diff
@@ -0,0 +1,400 @@
...
+
+    /**
+     * Append a mutation to the log. This method is non-blocking and returns quickly, unless the
+     * ring buffer is full. The actual write happens asynchronously. We expect multiple append()
+     * calls followed by a sync(). The appends will be batched by the Disruptor. Should the ring
+     * buffer become full, which is not expected under normal operation but could (and should)
+     * happen if the log file writer is unable to make progress, due to a HDFS level disruption.
+     * Should we enter that condition this method will block until the append can be inserted.
+     * @param tableName The name of the HBase table the mutation applies to.
+     * @param commitId  The commit identifier (e.g., SCN) associated with the mutation.
```
tkhurana, line 127: Is this basically the timestamp the coproc assigns to the mutation ?

apurtell: No. The mutation and the cells in the mutation carry that timestamp. The "commit identifier" is a Phoenix level transaction id. It can be whatever Phoenix needs it to be. If we don't need this concept at this layer it can be removed.

phoenix-core-server/src/main/java/org/apache/phoenix/replication/ReplicationLogManager.java
```diff
@@ -0,0 +1,579 @@
...
+
+    /**
+     * Gets the singleton instance of the ReplicationLogManager using the lazy initializer pattern.
+     * Initializes the instance if it hasn't been created yet.
+     * @param conf Configuration object.
+     * @return The singleton ReplicationLogManager instance.
+     * @throws IOException If initialization fails.
+     */
+    public static ReplicationLogManager getInstance(Configuration conf,
+            RegionServerServices rsServices) throws IOException {
```
tkhurana, line 129: The IndexRegionObserver coproc has the RegionCoprocessorEnvironment interface. I didn't find any way to get RegionServerServices from that interface. https://github.com/apache/phoenix/blob/master/phoenix-core-server/src/main/java/org/apache/phoenix/hbase/index/IndexRegionObserver.java#L407

apurtell: Ok, see 2bd6d010a . Instead of aborting the regionserver if there is a Disruptor level problem we force close the writer. Instead of having the writer do a server abort the responsibility is pushed up to the higher layers, the users of the writer, which will receive IOExceptions because the writer has been closed. This provides more flexibility. (We can try to roll the writer first?) Although the challenge of aborting the regionserver if that must happen falls to the callers.

apurtell:
More tests and test fixes. I have a couple more test cases to add and then I will take this out of draft status.

phoenix-core-server/src/main/java/org/apache/phoenix/replication/ReplicationLogWriter.java
```diff
@@ -0,0 +1,393 @@
...
+    private final Configuration conf;
+    private final ReplicationLogManager logManager;
+    private final int ringBufferSize;
+    private final long syncTimeoutMs;
+    private Disruptor<LogEvent> disruptor;
+    private RingBuffer<LogEvent> ringBuffer;
+    private final AtomicBoolean started = new AtomicBoolean(false);
+    private final AtomicBoolean closed = new AtomicBoolean(false);
+
+    public ReplicationLogWriter(Configuration conf, ReplicationLogManager logManager) {
```
tkhurana, line 93: Is it expected that each rpc thread will instantiate its own ReplicationLogWriter() object ?

apurtell: > Is it expected that each rpc thread will instantiate its own ReplicationLogWriter() object ?

A RPC thread gets the `ReplicationLogManager` instance with `ReplicationLogManager.getInstance()` and then call its getWriter method.

    ReplicationLogManager manager = ReplicationLogManager.getInstance();
    Log.Writer writer = manager.getWriter();

I've tried to make this simple. `writer` has a close method but it's not for callers, it is not a closeable resource. It has an independent lifecycle. You use it until you can't use it and then call `manager.getWriter();` to get another. The methods of the `Writer` , `append` and `sync`, are multi-threading safe.

So the expected usage is like:

    ReplicationLogManager manager = ReplicationLogManager.getInstance();
    while (retrying) {
        Log.Writer writer;
        try {
            writer = manager.getWriter();
            writer.append(...);
            writer.append(...);
            writer.append(...);
            writer.sync();
        } catch (IOException e) {
            // Decide what to do about the exception. Retry, probably. Or abort, possibly.
            // Too many retries?
        }
    }

apurtell:
I squashed the javadoc, spotbugs, and checkstyle fixes into a single commit and rebased.

phoenix-core-server/src/main/java/org/apache/phoenix/replication/ReplicationLogWriter.java
tkhurana: Thanks @apurtell I ok with option 1.

tkhurana: I am still confused as to how the ReplicationLogWriter is being used. I don't see how ReplicationLogManager is using ReplicatonLogWriter.

apurtell: `ReplicationLogManager` returns `ReplicatonLogWriter` from its `getWriter` method. 

Callers of this stuff will use`ReplicationLogManager.getInstance().getWriter()` to get a ReplicatonLogWriter to use.

apurtell:
I cleaned up and simplified `ReplicationLogWriter` significantly.

phoenix-core-server/src/main/java/org/apache/phoenix/replication/ReplicationLogWriter.java
apurtell: Yeah, this is still in draft for that reason. It will make more sense soon.

apurtell:
I had in mind two classes with separation of concerns, *Manager (lifecycle) and *Writer (disruptor), but my thinking has evolved since I started hacking on this. Actually it has been drifting for a couple of days but it didn't come together until now. For now there will be a single class `ReplicationLogWriter`. This will make more sense.

    ReplicationLogWriter writer = ReplicationLogWriter.getInstance();
    while (retrying) {
        writer.append(...);
        writer.append(...);
        writer.append(...);
        writer.sync();
        return;
    } catch (IOException e) {
        // Decide what to do about the exception. Retry, probably. Or abort, possibly.
        // Too many retries?
    }

apurtell:
[d56a9ae](https://github.com/apache/phoenix/pull/2144/commits/d56a9aef3a6ccfbf471d5e1fc55f3666807e58a5) unifies the code into `ReplicationLog` and the new `ReplicationLogTest` contains the combined unit tests.

apurtell:
I think the basic test coverage is pretty good now and I will probably take this out of draft status today. Let me think if there are a couple more test cases that would make sense. 

Here is the test coverage to date:

## Basic Append and Sync:

`testAppendAndSync()`: Verifies that multiple appends are correctly batched and written to the underlying LogFileWriter upon sync(), and that the order of operations (appends then sync) is maintained.

## Failure Handling and Retry (Append/Sync):

`testAppendFailureAndRetry()`: Simulates an append failure on the `LogFileWriter`. It checks that the `ReplicationLog` correctly rotates to a new writer and retries the append (and subsequent sync) on the new writer.

`testSyncFailureAndRetry()`: Simulates a sync failure on the `LogFileWriter`. It verifies that after an append, a failed sync leads to a writer rotation, and the original batch (including the failed append) is replayed and synced on the new writer.

## Backpressure (Ring Buffer Full):

`testBlockingWhenRingFull()`: Tests the backpressure mechanism. It simulates a slow consumer (inner `LogFileWriter`) to fill the Disruptor ring buffer and confirms that subsequent append() calls block until space is available, then eventually succeed.

## Sync Timeout:

`testSyncTimeout()`: Simulates a `LogFileWriter.sync()` operation that takes longer than the configured timeout. It verifies that the `ReplicationLog.sync()` call throws an `IOException` caused by a `TimeoutException`.

## Concurrent Producers:

`testConcurrentProducers()`: Tests concurrent appends from multiple threads, ensuring all records are eventually processed and written by the LogFileWriter in the correct order (verified by commitId).

## Time-Based Rotation:

`testTimeBasedRotation()`: Verifies that the log rotates automatically after the configured time interval, even if no new appends trigger a size check.

`testRotationTask()`: Specifically tests the LogRotationTask runnable, ensuring it rotates the log when the time threshold is met.

## Size-Based Rotation:

`testSizeBasedRotation()`: Checks that the log rotates when the current writer's file size exceeds the configured size threshold.

## Close Behavior:

`testClose()`: Ensures that `close()` flushes pending appends, closes the underlying writer, and prevents further operations. Also checks for idempotency of `close()`.

## Error Handling during Rotation/Initialization:

`testFailedRotation()`: Simulates a failure during the creation of a new writer during rotation. It verifies that the ReplicationLog continues to use the old writer for subsequent appends and syncs, rather than failing completely.

`testEventProcessingException()`: Simulates a RuntimeException during the LogEventHandler's processing of an append event. It verifies that the `LogExceptionHandler` catches this, closes the ReplicationLog (and its writer), and subsequent operations fail.

`testSyncFailureAllRetriesExhausted()`: Simulates persistent sync failures across multiple retries (and rotations). It verifies that the sync operation eventually fails with a timeout after exhausting retries.

`testFsInitFailure()`: Simulates a failure during FileSystem.mkdirs() when initializing the ReplicationLog. Verifies that `init()` throws an `IOException`.

## Rotation During Batch:

`testRotationDuringBatch()`: Tests the scenario where a rotation occurs while a batch of appends is in-flight (i.e., published to the ring buffer but not yet processed by the event handler). It verifies that the in-flight batch is correctly replayed to the new writer after rotation.

apurtell:
I moved this out of Draft status. I may still add more tests. When the precommit checks come back with results will update for that too.

apurtell:
Pushed up fixes for precommit findings. 

There are false positive issues with ASF license.

There are also false positives for "codespell":

    phoenix-core/src/test/java/org/apache/phoenix/replication/ReplicationLogTest.java:29: atLeast ==> at least
    phoenix-core/src/test/java/org/apache/phoenix/replication/ReplicationLogTest.java:676: atLeast ==> at least
    
`atLeast` is the valid name of a Mockito matcher.

apurtell:
To aid in your review I asked Gemini to summarize this patch:

**Summary of Changes in PHOENIX-7567**

1.  **`ReplicationLog` Class (New):**
    *   **Singleton Pattern:** Implements a singleton pattern (`get` method) to manage a single `ReplicationLog` instance per server.
    *   **Configuration:** Reads various configuration parameters related to log rotation (time, size), HDFS URLs (standby, fallback), sharding, ring buffer size, and sync timeouts.
    *   **FileSystem Initialization:** Initializes `FileSystem` objects for both standby and fallback HDFS URLs. Creates the necessary base directories if they don't exist.
    *   **Sharding:** Implements logic to create sharded subdirectories under the main log directory. The shard is determined by a hash of the server name and the current timestamp to distribute logs.
    *   **Log Rotation:**
        *   Manages the current `LogFileWriter` instance.
        *   Implements `shouldRotate()` to check time and size-based rotation triggers.
        *   Implements `rotateLog()` to close the current writer and open a new one. This method is reentrant lock protected.
        *   Uses a `ScheduledExecutorService` (`rotationExecutor`) to periodically check for time-based rotation via `LogRotationTask`.
    *   **Disruptor Integration:**
        *   Initializes an LMAX Disruptor (`disruptor` and `ringBuffer`) with a configurable ring buffer size, using `ProducerType.MULTI` and `YieldingWaitStrategy`.
        *   Defines an inner `LogEvent` class to represent data and sync operations in the ring buffer.
        *   Defines an inner `LogEventHandler` to process events from the Disruptor. This handler is responsible for:
            *   Getting the current `LogFileWriter`.
            *   Handling writer rotation: If the writer's generation has changed (indicating a rotation), it replays the `currentBatch` of in-flight appends to the new writer.
            *   Appending data events to the writer and adding them to `currentBatch`.
            *   Processing sync events by calling `writer.sync()`, clearing `currentBatch`, and completing the associated `CompletableFuture`.
            *   Basic retry logic for `append` and `sync` operations: if an `IOException` occurs, it forces a log rotation and retries the operation on the new writer up to `maxRetries`.
        *   Defines an inner `LogExceptionHandler` to handle critical errors in the Disruptor lifecycle by calling `closeOnError()`.
    *   **`append()` Method:** Publishes a data event (mutation) to the Disruptor's ring buffer. This method is non-blocking unless the ring buffer is full.
    *   **`sync()` Method:** Publishes a sync event to the Disruptor's ring buffer and blocks (using `CompletableFuture.get()`) until the `LogEventHandler` processes this event and completes the future. Implements a timeout for the sync operation.
    *   **`close()` and `closeOnError()` Methods:** Handle graceful shutdown and error-induced shutdown of the Disruptor and writers.
    *   **Replication Mode:** Introduces a `ReplicationMode` enum (SYNC, STORE_AND_FORWARD, SYNC_AND_FORWARD) and a `currentMode` field, though the store-and-forward logic is not yet implemented (marked as TODO).
    *   **Dependencies:** Adds `com.lmax:disruptor` as a dependency in `phoenix-core-server/pom.xml`.

2.  **`LogFileWriter` Class (Modified):**
    *   Adds a `generation` field (long) and corresponding getter/setter. This generation number is incremented by `ReplicationLog` each time a new writer is created (during rotation or initialization). The `LogEventHandler` uses this to detect if the writer it holds is stale due to a rotation.

3.  **New Package:** Introduces `org.apache.phoenix.replication` package.

**Alignment with the Design Document**

*   **Replication Log Writer (High-Level):** The `ReplicationLog` class is the high-level "Replication Log Writer" component. It orchestrates log creation, rotation, and the asynchronous writing of records via the Disruptor.
*   **LMAX Disruptor:** The design explicitly mentions leveraging the LMAX Disruptor for its ring buffer implementation to decouple producers from the writer and handle batching. This is directly implemented.
*   **Log File Management (Rotation):** The time-based and size-based log rotation mechanisms are implemented as described.
*   **Synchronous Writes (Foundation):** The `append()` method itself is fast (publishes to ring buffer), but the `sync()` method provides the synchronous guarantee by blocking until the `LogEventHandler` confirms the underlying `LogFileWriter` has synced. This is the core of the "synchronous replication" part.
*   **Error Handling (Initial):** The `LogEventHandler` includes basic retry logic for `append` and `sync` by attempting to roll the writer. The `LogExceptionHandler` provides a mechanism to shut down on critical Disruptor errors.
*   **Replication Mode (Placeholder):** The `ReplicationMode` enum and `currentMode` field are introduced, laying the groundwork for future implementation of store-and-forward logic.
*   **Sharding:** The sharding mechanism based on server name and timestamp hash is implemented for distributing log files.

**How this Diff Builds Upon PHOENIX-7565.**

*   PHOENIX-7565 focused on the *format* of the log files (headers, blocks, trailers, records) and the low-level I/O classes (`LogFileFormatWriter/Reader`, `LogFileCodec`, stream adapters).
*   PHOENIX-7567 builds *on top* of that by creating the `ReplicationLog` class, which is the *manager* or *orchestrator* that uses the `LogFileWriter` (and by extension, the format writer and codec) to actually write data.
*   It introduces the asynchronous processing pipeline using LMAX Disruptor, which was a key part of the "Replication Log Writer" component's internal design for high throughput and low latency appends.
*   It implements the logic for log rotation, which is essential for managing log file lifecycle.
*   It provides the `sync()` mechanism that gives the synchronous guarantee to the callers (e.g., the coprocessor hooks).

phoenix-core-server/src/main/java/org/apache/phoenix/replication/ReplicationLog.java
```diff
@@ -0,0 +1,857 @@
...
+            closeWriter(currentWriter);
+            currentWriter = null;
+        }
+        // Directly halt the disruptor. shutdown() would wait for events to drain. We are expecting
+        // that will not work.
+        disruptor.halt();
+    }
+
+    /** Closes the log. */
+    public void close() {
```
tkhurana, line 669: It seems like this method is not threadsafe.

apurtell: Addressed in a5a81cecd

phoenix-core/src/test/java/org/apache/phoenix/replication/ReplicationLogTest.java
```diff
@@ -0,0 +1,1229 @@
...
+        doThrow(new IOException("Simulated append failure"))
+            .when(writerBeforeRoll).append(anyString(), anyLong(), any(Mutation.class));
+
+        // Append data
+        logWriter.append(tableName, commitId, put);
+        logWriter.sync();
+
+        // Get the inner writer we rolled to.
+        LogFileWriter writerAfterRoll = logWriter.getWriter();
+        assertNotNull("Initial writer should not be null", writerBeforeRoll);
```
tkhurana, line 192: Should we be checking `writerAfterRoll` ?

phoenix-core-server/src/main/java/org/apache/phoenix/replication/ReplicationLog.java
```diff
@@ -0,0 +1,908 @@
...
+     * are batched and provided to the underlying LogWriter, which will then be synced. If there is
+     * a problem syncing the LogWriter we will retry, up to the retry limit, rolling the writer for
+     * each retry.
+     * <p>
+     * NOTE: When the ReplicationLogManager is capable of switching between synchronous and
+     * fallback (store-and-forward) writers, then this will be pretty bullet proof. Right now we
+     * will still try to roll the synchronous writer a few times before giving up.
+     * @throws IOException If the sync operation fails after retries, or if interrupted.
+     */
+    public void sync() throws IOException {
```
tkhurana, line 408: What will be the behavior of sync when multiple rpc handler threads will do sync in parallel on the same underlying ReplicationLog ?

tkhurana: If I understand this correctly, we will publish a SYNC event for every sync call by the rpc handler and will do a filesystem sync for every SYNC event. Moreover, each SYNC will also create a new block

apurtell: Yes, unless there are no pending in-flight appends. I think it has to be this way. Any caller of sync expects when the sync returns that all of its in-flight appends have been persisted and also the sync has been propagated through to the underlying filesystem. 

My focus now is on correctness and simplicity in implementation is part of that. If the callers have a requirement that sync calls are optimized, this concern for now is pushed to the callers. I do think there are opportunities to do sync merging/consolidation, like how HBase WAL does, as a follow on.

apurtell: Addressed in 2a3b0f620

apurtell:
Some of the Yetus findings are no longer valid after an interim change was reverted. A tool introducing new dependencies was removed, yet the build errors about the new dependencies persist, and there was a markdown file removed by the revert  yet there are still "markdownlint" results.  I will try rebasing the PR branch to clean this up.

apurtell:
Squashed and force pushed, hope that cleans things up.

tkhurana (approved):

apurtell:
Rebased PHOENIX-7562-feature , rebased this PR on the rebased PHOENIX-7562-feature

PR #2149, opened 2025-05-13
PHOENIX-7609 : CDC creation fails when data table has case sensitive name

palashc:


haridsv:
The way CDC ITs are designed, the CDCDefinitionIT test only tests the DDL, so the verification is done not by running queries but by checking the metadata. As such, I would recommend having a separate test in CDCQueryIT that uses quoted identifiers. Also, I suggest including a test with quoted schema name.

virajjasani (approved):
+1

haridsv (commented):
The name generation in the CDCQueryIT is repetitive, can you please look into the possibility of adding utility methods and simplify the code?

haridsv (approved):

PR #2150, opened 2025-05-13
PHOENIX-7610 Using CAST() on pk columns always result in full table scan

virajjasani:
Jira: PHOENIX-7610

For the range scan or point lookup on the single row, it is required to provide either full pk columns with equal comparison (point lookup) or prefix pk columns with equal comparison operator (range scan).

However, when we use SQL CAST() function on pk columns to convert the type of the given column into another data type, the plan generated by where optimizer always performs full table scan instead of range scan or point lookup.


kadirozde (approved):
+1, Thanks!

PR #2152, opened 2025-05-15
PHOENIX-7605 Addendum Fixing IT where table was used after t.close(), passing pool to hbase client based on value.

ritegarg:
We use Connection.getTable in org.apache.phoenix.query.HTableFactory. There are 2 flavors of this getTable method and the issue comes from the implementation of this method.
getTable(tableName) -> This one has a default implementation which calls getTable(tableName, null) The null here is executorService.
But in org.apache.hadoop.hbase.client.ConnectionImplementation implementation of getTable(tableName) it calls getTable(tableName, getBatchPool()) . This pool is managed by hbase code and it cleans up the pool on close() -> [link](https://github.com/enis/hbase-sal/blob/master/hbase-client/src/main/java/org/apache/hadoop/hbase/client/HTable.java#L171-L176)

Note that in existing implementation of this change we were calling HTable.close() which was closing the threadpool managed by HTable. Ideally the table should be closed only after using it but in this case the scanner was working as ConnectionImplementation was creating the batchpool [here](https://git.soma.salesforce.com/bigdata-packaging/hbase/blob/2.5.10-sfdc-13/hbase-client/src/main/java/org/apache/hadoop/hbase/client/ConnectionImplementation.java#L454) and passing it to HTable which was not closing it when HTable.close() was called.
Refer [link](https://github.com/enis/hbase-sal/blob/master/hbase-client/src/main/java/org/apache/hadoop/hbase/client/HTable.java#L171-L176)

Now, we are safeguarding from this behavior by checking if pool passed to ConnectionImplementation is null or not.

### DIFF
--- a/phoenix-core/src/it/java/org/apache/phoenix/end2end/MappingTableDataTypeIT.java
+++ b/phoenix-core/src/it/java/org/apache/phoenix/end2end/MappingTableDataTypeIT.java
@@ -58,11 +58,9 @@ public class MappingTableDataTypeIT extends ParallelStatsDisabledIT {
         final TableName tableName = TableName.valueOf(mtest);
         Properties props = PropertiesUtil.deepCopy(TEST_PROPERTIES);
         PhoenixConnection conn = DriverManager.getConnection(getUrl(), props).unwrap(PhoenixConnection.class);
-        
 
-        // We should only close the table after we are done using it.
         try(Table t = conn.getQueryServices().getTable(Bytes.toBytes(mtest));
-            Admin admin = conn.getQueryServices().getAdmin();) {
+            Admin admin = conn.getQueryServices().getAdmin()) {
             // Create table then get the single region for our new table.
             TableDescriptorBuilder builder = TableDescriptorBuilder.newBuilder(tableName);
             builder.setColumnFamily(ColumnFamilyDescriptorBuilder.of(Bytes.toBytes("cf1")))
--- a/phoenix-core/src/it/java/org/apache/phoenix/end2end/MappingTableDataTypeIT.java
+++ b/phoenix-core/src/it/java/org/apache/phoenix/end2end/MappingTableDataTypeIT.java
@@ -100,6 +100,7 @@ public class MappingTableDataTypeIT extends ParallelStatsDisabledIT {
             assertEquals("Column Value", "value2", Bytes.toString(kvs.get(0).getValueArray(), kvs.get(0).getValueOffset(), kvs.get(0).getValueLength()));
             assertNull("Expected single row", results.next());
         } finally {
+            // We should close the table only after we are done using it.
             if(t != null) {
                 t.close();
             }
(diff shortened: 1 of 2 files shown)


phoenix-core/src/it/java/org/apache/phoenix/end2end/MappingTableDataTypeIT.java
```diff
@@ -60,15 +60,15 @@ public void testMappingHbaseTableToPhoenixTable() throws Exception {
         PhoenixConnection conn = DriverManager.getConnection(getUrl(), props).unwrap(PhoenixConnection.class);
         
         Admin admin = conn.getQueryServices().getAdmin();
+        Table t = null;
         try {
```
NihalJain, line 64: why not simply use try-with-resources?

phoenix-core/src/it/java/org/apache/phoenix/end2end/MappingTableDataTypeIT.java
```diff
@@ -60,15 +60,15 @@ public void testMappingHbaseTableToPhoenixTable() throws Exception {
         PhoenixConnection conn = DriverManager.getConnection(getUrl(), props).unwrap(PhoenixConnection.class);
         
         Admin admin = conn.getQueryServices().getAdmin();
-        try {
+        // We should only close the table after we are done using it.
+        try(Table t = conn.getQueryServices().getTable(Bytes.toBytes(mtest));) {
```
NihalJain, line 64: can do same for admin. also is the message is relevant now?

ritegarg: Done. I wanted to be explicit about why I changed this. I'll put it in PR description.

phoenix-core/src/it/java/org/apache/phoenix/end2end/MappingTableDataTypeIT.java
ritegarg: Done

ritegarg:
@virajjasani @tkhurana can you also review? I have added the reasoning in the description.

virajjasani (approved):
+1

NihalJain (approved):

jpisaac (approved):

palashc:
@ritegarg I saw a couple of test failures on another PR of mine after this change was merged. Also failing on the [master build](https://ci-hadoop.apache.org/job/Phoenix/job/Phoenix-mulitbranch/job/master/845/). 
```
org.apache.phoenix.query.ConnectionQueryServicesImplTest.testGetSysMutexTableWithNamespace
org.apache.phoenix.query.ConnectionQueryServicesImplTest.testGetSysMutexTableWithName
```

virajjasani:
Fixing this is more critical because this will not let any end-to-end IT tests run.

@ritegarg please take a look.

ritegarg:
Did you add these changes in your PR? 
https://github.com/apache/phoenix/commit/20e3bded8adabbb501d41a7d7cb8d7782e7cba0e#diff-806ba52058360c7a5cb13b945394e8caeece35f6a1aa6a6710715281f502f189R378

virajjasani:
The tests are failing on master build as Palash shared above. The latest commit on this build refers to this PR changes.

ritegarg:
Fix https://github.com/apache/phoenix/pull/2155/files

PR #2154, opened 2025-05-16
PHOENIX-7612 Fix Cell references in IndexRegionObserver

tkhurana:


phoenix-core-client/src/main/java/org/apache/phoenix/util/IndexUtil.java
```diff
@@ -856,6 +856,40 @@ public static void removeEmptyColumn(Mutation m, byte[] emptyCF, byte[] emptyCQ)
...
     }
 
+    /**
+     * Creates a true deep copy of the Put Mutation, including deep copies of all cells if the
+     * cells are backed by off-heap. The Mutation(Mutation source) constructor always does a
+     * shallow copy of the cells.
+     * @param original The original Put Mutation to copy
+     * @return A new Put Mutation with deep copies of all fields and Cells
+     */
+    public static Put copyPut(Put original) throws IOException {
```
virajjasani, line 866: While this might only be used by IndexRegionObserver for now, this can be used by others also in the future. How about keeping it as separate utility class e.g. `CellUtil`?

phoenix-core-server/src/main/java/org/apache/phoenix/hbase/index/IndexRegionObserver.java
```diff
@@ -910,7 +911,12 @@ private void applyPendingPutMutations(MiniBatchOperationInProgress<Mutation> min
                 context.dataRowStates.put(rowKeyPtr, dataRowState);
             }
             Put nextDataRowState = dataRowState.getSecond();
-            dataRowState.setSecond((nextDataRowState != null) ? applyNew((Put) m, nextDataRowState) : new Put((Put) m));
+            // Need to deep copy the references to the cells in the mutation
+            Put copied = IndexUtil.copyPut((Put) m);
```
virajjasani, line 915: In addition to this, two more places use such clone c'tor with `context.dataRowStates.put()`. I think this would apply to both of them because both are also setting values in the `BatchMutateContext`.

tkhurana: Which places are you talking about ? I think we need to do this only at places where we are setting values in BatchMutateContext for the first time. That can happen when we are reading from disk in `readDataTableRows` where we get the cells from `RegionScanner.next()` or when reading from the Mutation in the miniBatchOp.  Once we have our copies, regular copy is fine.

virajjasani: > I think we need to do this only at places where we are setting values in BatchMutateContext for the first time.

Yes I was talking about this:
```
    private void readDataTableRows(ObserverContext<RegionCoprocessorEnvironment> c,
                                   BatchMutateContext context, Scan scan) throws IOException {
        try (RegionScanner scanner = c.getEnvironment().getRegion().getScanner(scan)) {
            boolean more = true;
            while (more) {
                List<Cell> cells = new ArrayList<Cell>();
                more = scanner.next(cells);
                if (cells.isEmpty()) {
                    continue;
                }
                byte[] rowKey = CellUtil.cloneRow(cells.get(0));
                Put put = new Put(rowKey);
                for (Cell cell : cells) {
                    put.add(cell);
                }
                context.dataRowStates.put(new ImmutableBytesPtr(rowKey), new Pair<Put, Put>(put, new Put(put)));
            }
        }
    }
```

I was thinking about `new Put(put)` but looks like we don't need it because you cloned cells that are being added to Put mutation. We are good I think.

tkhurana:
There are some test failures, looking at it.

stoty:
Does this only affect 5.3 @tkhurana  ?

If not, then looks important enough to be backported.

PR #2163, opened 2025-05-20
PHOENIX-7616 NPE when there are Conditional expressions on indexed columns

tkhurana:
NPE when the row key had a column with a null value and that column was part of a conditional TTL expression.


phoenix-core-client/src/main/java/org/apache/phoenix/schema/CompiledConditionalTTLExpression.java
```diff
@@ -216,7 +216,7 @@ public boolean isExpired(List<Cell> result, boolean isRaw) {
             return false;
         }
         Object value = PBoolean.INSTANCE.toObject(ptr);
-        return value.equals(Boolean.TRUE);
+        return value != null ? value.equals(Boolean.TRUE) : false;
```
virajjasani, line 219: simplify `Boolean.TRUE.equals(value)`?

virajjasani (approved):
+1, pending build results

PR #2164, opened 2025-05-21
PHOENIX-7618 Missing jul-to-slf4j.jar in the lib folder of the assemb…

meszinorbi:
…led tar file


stoty:
The test failures are a Yetus limitation.

stoty:
Don't add a Signed-off-by: line for yourself, that is for not for the author but the reviewers.
(Also, we don't use that in Phoenix)

PR #2165, opened 2025-05-21
 PHOENIX-7600 Replace commons-logging with slf4j

meszinorbi:


phoenix-mapreduce-byo-shaded-hbase/pom.xml
```diff
@@ -578,6 +578,12 @@
       <artifactId>commons-configuration2</artifactId>
       <version>${commons-configuration2.excludeonly.version}</version>
       <scope>provided</scope>
+      <exclusions>
```
stoty, line 581: Are you sure this is necessary here ?
Provided should transitively apply to transitive dependencies, if the only user set to provided.

phoenix-server/pom.xml
```diff
@@ -545,6 +545,12 @@
       <artifactId>commons-configuration2</artifactId>
       <version>${commons-configuration2.excludeonly.version}</version>
       <scope>provided</scope>
+      <exclusions>
```
stoty, line 548: Same here

phoenix-mapreduce-byo-shaded-hbase/pom.xml
meszinorbi: mvn dependency:tree fooled me, after building without the exclusions, the produced jars did not contain commons-logging

virajjasani:
@stoty @meszinorbi are you able to run sqlline.py from the latest master branch HEAD?
I am facing this issue, not sure if any of the recent dependency changes might have broken this?

```
$ python3 bin/sqlline.py 
2025-06-13T13:23:56,862 WARN  [main] impl.MetricsConfig: Cannot locate configuration: tried hadoop-metrics2-phoenix.properties,hadoop-metrics2.properties
2025-06-13T13:23:56,968 WARN  [main] util.NativeCodeLoader: Unable to load native-hadoop library for your platform... using builtin-java classes where applicable
2025-06-13T13:23:57,026 ERROR [main] query.GuidePostsCacheProvider: Could not load/instantiate class org.apache.phoenix.query.DefaultGuidePostsCacheFactory
org.apache.phoenix.exception.PhoenixNonRetryableRuntimeException: Could not load/instantiate class org.apache.phoenix.query.DefaultGuidePostsCacheFactory
sqlline version 1.9.0
...
...
```

It does not start Phoenix system schema deployment.

stoty:
I've just built master, started phoenix_sandbox.py and the sqlline, I don't see any problems @virajjasani .
I tried JDK 8, 11 and 17. python 3.8 and 3.12.

I don't think it's the logging changes, if there are problems with the classpath phoenix fails before trying to loadGuidePostsCacheProvider.

I did change the GuidePostsCacheProvider loading code to improve classloader compatibility, but that was more than a year ago.

Have you re-run `mvn package -am -pl phoenix-core` to get the new dependencies downloaded into the /lib dir ?

PR #2166, opened 2025-05-21
PHOENIX-7591 Concurrent updates to tables with indexes can cause data inconsistency

tkhurana:
Added a new phase which tracks if we were able to acquire the row locks or not and accordingly decrement the reference count in pending rows.


kadirozde (approved):
+!, Thanks!

PR #2168, opened 2025-05-27
PHOENIX-7620 Sequence cache is ignored when performing SELECT NEXT N VALUES

tmnd1991:


stoty:
This patch doesn't work in itself, as show by the failing SequenceBulkAllocationIT .

While the idea is solid, the client side code also needs to be updated to handle the caching in this case, and backwards compatibility also needs to be maintained, so this would be a larger patch.

PR #2173, opened 2025-06-03
PHOENIX-7629 :- Invalidate already closed CQSI in case of URL change for Phoenix HA Connections

lokiore:


phoenix-core-client/src/main/java/org/apache/phoenix/jdbc/HighAvailabilityPolicy.java
```diff
@@ -273,6 +275,23 @@ private static void closeConnections(HighAvailabilityGroup haGroup, String url,
         }
     }
 
+    /**
+     * Utility to invalidate CQS cache for a given url of a haGroup
```
jpisaac: nit: may be good add comments around the benefits of invalidating vs just closing

jpisaac (approved):
LGTM +1, request for some more comment

phoenix-core-client/src/main/java/org/apache/phoenix/jdbc/HighAvailabilityPolicy.java
```diff
@@ -161,12 +164,7 @@ private void transitStandby(HighAvailabilityGroup haGroup, String url,
         private void transitActive(HighAvailabilityGroup haGroup, String url,
                                    ClusterRoleRecord.RegistryType registryType) throws SQLException {
             // Invalidate CQS cache if any that has been closed but has not been cleared
-                for (HAURLInfo haurlInfo : HighAvailabilityGroup.URLS.get(haGroup.getGroupInfo())) {
```
ritegarg, line 164: For line 161 -> If we change standby from cluster1 to cluster2, close the connection but not invalidate the cache. If we rollback, will we observe connection closed?

lokiore: You mean url1, STANDBY to url2, STANDBY ?
that will for to change in URL and we will invalidate cache there. 
161 is for change in role for a cluster from ACTIVE -> STANDBY

ritegarg: Clarified that this only applies to Failover connection and not parallel. +1

PR #2174, opened 2025-06-03
PHOENIX-7586 (ADDENDUM) :- Handle Role transitions for ActiveToStanby role in Failover HAPolicy

lokiore:
Closing connections when we transition to either Standby or Active_To_Standby instead of closing it at Standby only.

### DIFF
--- a/phoenix-core-client/src/main/java/org/apache/phoenix/jdbc/HighAvailabilityPolicy.java
+++ b/phoenix-core-client/src/main/java/org/apache/phoenix/jdbc/HighAvailabilityPolicy.java
@@ -75,20 +75,20 @@ public enum HighAvailabilityPolicy {
                         "Doing nothing for Cluster Role Change");
                 return;
             }
-            if ((oldRecord.getRole1() == ACTIVE || oldRecord.getRole1() == ACTIVE_TO_STANDBY)
-                    && newRecord.getRole1() == STANDBY) {
-                transitStandby(haGroup, oldRecord.getUrl1(), oldRecord.getRegistryType());
+            if (oldRecord.getRole1() == ACTIVE &&
+                    (newRecord.getRole1() == STANDBY || newRecord.getRole1() == ACTIVE_TO_STANDBY)) {
+                transitStandby(haGroup, oldRecord.getUrl1(), oldRecord.getRegistryType(),
+                        newRecord.getRole1());
             }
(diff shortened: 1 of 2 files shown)


phoenix-core-client/src/main/java/org/apache/phoenix/jdbc/HighAvailabilityPolicy.java
```diff
@@ -75,20 +75,20 @@ void transitClusterRole(HighAvailabilityGroup haGroup, ClusterRoleRecord oldReco
                         "Doing nothing for Cluster Role Change");
                 return;
             }
-            if ((oldRecord.getRole1() == ACTIVE || oldRecord.getRole1() == ACTIVE_TO_STANDBY)
-                    && newRecord.getRole1() == STANDBY) {
-                transitStandby(haGroup, oldRecord.getUrl1(), oldRecord.getRegistryType());
+            if (oldRecord.getRole1() == ACTIVE &&
+                    (newRecord.getRole1() == STANDBY || newRecord.getRole1() == ACTIVE_TO_STANDBY)) {
```
ritegarg, line 79: If we change standby only, we'll be going from cluster1 -> STANDBY to cluster1-> ACTIVE_TO_STANDBY  to cluster2 -> STANDBY. Should we close connection when we move from STANDBY -> ACTIVE_TO_STANDBY for that ParallelConnection use case?

lokiore: For Parallel we allow connection creation at any state except OFFLINE/UNKNOWN. These state changes are only for Failover, for parallel we don't do anything to connections when there is a state change. What usecase you are thinking here?

ritegarg (approved):

phoenix-core-client/src/main/java/org/apache/phoenix/jdbc/HighAvailabilityPolicy.java
shahrs87: Do we want to check if oldRole is Active? Can we just call transitStandby if the new role is Standby or Active to Standby.

lokiore: @shahrs87 we need to detect a change right? otherwise we could just close connections if there is only a version change or one role can affect the other role..

shahrs87 (approved):

PR #2179, opened 2025-06-05
PHOENIX-7573 (ADDENDUM) Handles upgrade for tables with alternate default column family

jpisaac:


phoenix-core-client/src/main/java/org/apache/phoenix/util/UpgradeUtil.java
```diff
@@ -1579,6 +1588,12 @@ public static void moveHBaseLevelTTLToSYSCAT(PhoenixConnection oldMetaConnection
                                         "in progress => numOfTableHasTTLMoved: %d",
                                 numOfTableThatHasTTLMoved));
 
+                    } catch (SQLException e) {
+                        LOGGER.error(String.format(
+                                "Failed moving ttl value to TTL" +
+                                        " column for %s with Exception :",
+                                currentTableName), e);
+
```
virajjasani: We want to propagate the exception from here I assume

virajjasani: Also, to log the error for any exception (e.g. NPE that we came across), how about catching generic exception and logging the current table name?

```
                    } catch (Exception e) {
                        LOGGER.error("Failed moving ttl value to TTL"
                                + " column for {} with Exception", currentTableName, e);
                        throw new RuntimeException(e);
                    }
```

phoenix-core-client/src/main/java/org/apache/phoenix/util/UpgradeUtil.java
```diff
@@ -1530,15 +1531,23 @@ public static void moveHBaseLevelTTLToSYSCAT(PhoenixConnection oldMetaConnection
...
+                            byte[] emptyCF = SchemaUtil.getEmptyColumnFamily(pTable);
                             TableDescriptor tableDesc = admin.getDescriptor(SchemaUtil.getPhysicalTableName(
                                     fullTableName, readOnlyProps));
-                            int ttl = tableDesc.getColumnFamily(DEFAULT_COLUMN_FAMILY_BYTES).
-                                    getTimeToLive();
+                            ColumnFamilyDescriptor cfd = tableDesc.getColumnFamily(emptyCF);
+                            if (cfd == null) {
+                                LOGGER.warn("Not upgrading HBase level TTL definition for table {}", fullTableName);
+                                continue;
+                            }
```
virajjasani: How about?

```
if (cfd == null) {
    LOGGER.warn("No emptyCF found. Not upgrading HBase level TTL definition for table {}", fullTableName);
    continue;
}
```

phoenix-core-client/src/main/java/org/apache/phoenix/util/UpgradeUtil.java
```diff
@@ -1530,15 +1531,23 @@ public static void moveHBaseLevelTTLToSYSCAT(PhoenixConnection oldMetaConnection
                             byte[] tableName = rowKeyMetaData[PhoenixDatabaseMetaData.TABLE_NAME_INDEX];
 
                             String fullTableName = SchemaUtil.getTableName(schemaName, tableName);
+                            currentTableName = fullTableName;
                             if (SchemaUtil.isSystemTable(SchemaUtil.getTableNameAsBytes(schemaName, tableName))) {
                                 //We do not support system table ttl through phoenix ttl, and it will be moved to a
                                 //constant value in future commit.
                                 continue;
                             }
+                            PTable pTable = oldMetaConnection.getTableFromServerNoCache(schemaName, tableName);
```
virajjasani: As all we want is emptyCF, we don't need to make syscat call if PTable is already in cache? So, we can use `oldMetaConnection.getTable(null, fullTableName)` instead?

phoenix-core-client/src/main/java/org/apache/phoenix/util/UpgradeUtil.java
virajjasani: Eventually top level caller `CQSI#upgradeSystemTables()` anyways take care of wrapping any non-SQL exceptions into SQLException.

jpisaac: Was trying to complete as many batches as possible by letting it run

jpisaac:
@virajjasani Addressed comments

phoenix-core-client/src/main/java/org/apache/phoenix/util/UpgradeUtil.java
```diff
@@ -1530,15 +1531,23 @@ public static void moveHBaseLevelTTLToSYSCAT(PhoenixConnection oldMetaConnection
                             byte[] tableName = rowKeyMetaData[PhoenixDatabaseMetaData.TABLE_NAME_INDEX];
 
                             String fullTableName = SchemaUtil.getTableName(schemaName, tableName);
+                            currentTableName = fullTableName;
                             if (SchemaUtil.isSystemTable(SchemaUtil.getTableNameAsBytes(schemaName, tableName))) {
                                 //We do not support system table ttl through phoenix ttl, and it will be moved to a
                                 //constant value in future commit.
                                 continue;
                             }
+                            PTable pTable = oldMetaConnection.getTable(null, fullTableName);
```
virajjasani, line 1540: I wonder if this could cause too many getTable() calls to syscat, though should be manageable I assume. WDYT @jpisaac @tkhurana?

virajjasani (approved):
+1 overall

phoenix-core-client/src/main/java/org/apache/phoenix/util/UpgradeUtil.java
jpisaac: @virajjasani  Do not think it will be a lot ~ 200 calls during upgrade. The SYSCAT has been used to large number of getTable calls during client startup anyways.

virajjasani: Sounds good!

PR #2181, opened 2025-06-09
PHOENIX-7619 Excess HFiles are being read to look for more than required column versions

sanjeet006py:
Using ByteBuddy's runtime byte-code manipulation in IT to count the `StoreFileScanner#seek()` and `StoreFileScanner#reseek()` calls done for a table. HBase code already uses ByteBuddy indirectly via Mockito and is using the same version that I am adding. We can't use Mockito here as it is not able to mock/spy `StoreFileScanner` instances in mini-cluster as they are created in a separate thread different from main client thread. Thus, using ByteBuddy directly.

### DIFF
--- a/phoenix-core-server/src/main/java/org/apache/phoenix/coprocessor/BaseScannerRegionObserver.java
+++ b/phoenix-core-server/src/main/java/org/apache/phoenix/coprocessor/BaseScannerRegionObserver.java
@@ -430,7 +430,8 @@ abstract public class BaseScannerRegionObserver implements RegionObserver {
 
         Configuration conf = c.getEnvironment().getConfiguration();
         if (isPhoenixCompactionEnabled(conf)) {
-            setScanOptionsForFlushesAndCompactions(options);
+            options.setKeepDeletedCells(KeepDeletedCells.TTL);
+            options.setTTL(Long.MAX_VALUE);
             return;
         }
         if (!storeFileScanDoesntNeedAlteration(options)) {
--- a/phoenix-core/src/it/java/org/apache/phoenix/end2end/PreStoreScannerOpenIT.java
+++ b/phoenix-core/src/it/java/org/apache/phoenix/end2end/PreStoreScannerOpenIT.java
@@ -1,0 +1,260 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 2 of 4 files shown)


phoenix-core-server/src/main/java/org/apache/phoenix/coprocessor/BaseScannerRegionObserver.java
```diff
@@ -430,7 +430,8 @@ public void preStoreScannerOpen(ObserverContext<RegionCoprocessorEnvironment> c,
 
         Configuration conf = c.getEnvironment().getConfiguration();
         if (isPhoenixCompactionEnabled(conf)) {
-            setScanOptionsForFlushesAndCompactions(options);
+            options.setKeepDeletedCells(KeepDeletedCells.TTL);
+            options.setTTL(Long.MAX_VALUE);
```
virajjasani, line 434: This looks good, i was wondering if we could use `store.getColumnFamilyDescriptor().getTimeToLive()` value

sanjeet006py: Using `store.getColumnFamilyDescriptor().getTimeToLive()` can cause Phoenix row to expire partially as say one cell has HBase TTL expired but another cell of same row but different column doesn't have TTL expired then from Phoenix point of view all the cells should be visible. Please correct me if I am wrong.

phoenix-core/src/it/java/org/apache/phoenix/end2end/PreStoreScannerOpenIT.java
```diff
@@ -0,0 +1,260 @@
...
+    private static final int TTL_IN_DAYS = 30;
+    // The table name is used to filter the seek/reseek calls
+    public static final String TABLE_NAME = generateUniqueName();
+
+    // Counters for tracking seek/reseek calls
+    public static final AtomicLong SEEK_COUNT = new AtomicLong(0);
+    public static final AtomicLong RESEEK_COUNT = new AtomicLong(0);
+
+    // ByteBuddy advice classes for intercepting method calls
+    public static class SeekAdvice {
```
virajjasani, line 65: That's interesting usage. I hope the library version is stable, CVE-free.

sanjeet006py: I used same version used in HBase. Didn't check for CVE status. Let me check. Thanks

sanjeet006py: Mockito framework uses this same library i.e. `byte-buddy` for mocking constructors.

sanjeet006py: No vuln. are there:

- https://ossindex.sonatype.org/component/pkg:maven/net.bytebuddy/byte-buddy-agent@1.15.11
- https://ossindex.sonatype.org/component/pkg:maven/net.bytebuddy/byte-buddy@1.15.11

sanjeet006py:
There is only one already know flapper test failing: `org.apache.hadoop.hbase.regionserver.wal.WALRecoveryRegionPostOpenIT.testRecoveryRegionPostOpen`, rest all are succeeding. Thanks

PR #2183, opened 2025-06-10
PHOENIX-7633 Tuple API to provide serialized bytes size for the given Tuple

virajjasani:
Jira: PHOENIX-7633

### DIFF
--- a/phoenix-core-client/src/main/java/org/apache/phoenix/execute/TupleProjector.java
+++ b/phoenix-core-client/src/main/java/org/apache/phoenix/execute/TupleProjector.java
@@ -323,7 +323,7 @@ public class TupleProjector {
         }
 
         @Override
-        public long getKeyValueBytesSize() {
+        public long getSerializedSize() {
             long size = keyPtr.getLength() + projectedValue.getLength();
             if (keyValue != null) {
                 size += keyValue.getFamilyLength() + keyValue.getQualifierLength();
--- a/phoenix-core-client/src/main/java/org/apache/phoenix/execute/TupleProjector.java
+++ b/phoenix-core-client/src/main/java/org/apache/phoenix/execute/TupleProjector.java
@@ -323,6 +323,15 @@ public class TupleProjector {
         }
 
         @Override
+        public long getKeyValueBytesSize() {
+            long size = keyPtr.getLength() + projectedValue.getLength();
+            if (keyValue != null) {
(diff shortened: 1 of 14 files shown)


phoenix-core-client/src/main/java/org/apache/phoenix/execute/TupleProjector.java
```diff
@@ -322,6 +322,15 @@ public void getKey(ImmutableBytesWritable ptr) {
             ptr.set(keyPtr.get(), keyPtr.getOffset(), keyPtr.getLength());
         }
 
+        @Override
+        public long getKeyValueBytesSize() {
+            long size = keyPtr.getLength() + projectedValue.getLength();
+            if (keyValue != null) {
+                size += keyValue.getFamilyLength() + keyValue.getQualifierLength();
```
jpisaac, line 329: Don't you need the length of the actual value object?

phoenix-core-client/src/main/java/org/apache/phoenix/schema/tuple/ResultTuple.java
```diff
@@ -101,4 +101,16 @@ public boolean getValue(byte[] family, byte[] qualifier,
...
         return true;
     }
+
+    @Override
+    public long getKeyValueBytesSize() {
+        if (result == null || result.isEmpty()) {
+            return 0;
+        }
+        long totalSize = 0;
+        for (Cell cell : result.rawCells()) {
```
jpisaac, line 111: nit: There seems to be a utility method - getTotalSizeOfCells on the Result class

jpisaac (approved):
LGTM +1 Couple of comments

kadirozde (approved):
+1, Thanks!

phoenix-core-client/src/main/java/org/apache/phoenix/execute/TupleProjector.java
virajjasani: For projection, we are taking projectedValue's length, which will cover actual value.

phoenix-core-client/src/main/java/org/apache/phoenix/schema/tuple/ResultTuple.java
virajjasani: I tried that but it has lot of overhead values added, i see quite a diff in the cell values. It is using some additional heap bytes fixed overhead which is not serialized size that client application needs to worry about for memory tracking purpose.

PR #2187, opened 2025-06-11
PHOENIX-7637 Added jcl-over-slf4j dependency to phoenix-perf pom

meszinorbi:


stoty:
Please remove the colon from the commit message.
Also make sure that the JIRA description and commit message are in sync (in this case it'S better to update the jira description)

stoty:
Thanks.
Let's wait for the test results (though it rarely gets as far as the pherf tests)

stoty:
Looks like you need to add an exclusuion for the dependency check @meszinorbi (same as for the other modules)

> [ERROR] Failed to execute goal org.apache.maven.plugins:maven-dependency-plugin:3.1.1:analyze-only (enforce-dependencies) on project phoenix-pherf: Dependency problems found -> [Help 1]

PR #2188, opened 2025-06-11
Phoenix-7632 Adding ReplicationLogProcessor Component

Himanshu-g81:
JIRA - [PHOENIX-7632](https://issues.apache.org/jira/browse/PHOENIX-7632)

Adding the ReplicationLogProcessor component, that accepts a single replication log file, process mutations in configurable batch size and apply them to HBase table via async client.
The Replication Log Discovery (to be added as part of [PHOENIX-7568](https://issues.apache.org/jira/browse/PHOENIX-7568)) would list the log files and use ReplicationLogProcessor to read and apply the mutations to respective HBase table.

The Discovery and Tracking (Marking file in-progrss / deletion after successful processing) will be done as part of [PHOENIX-7568](https://issues.apache.org/jira/browse/PHOENIX-7568).

### DIFF
--- a/phoenix-core-server/src/main/java/org/apache/phoenix/replication/reader/ReplicationLogProcessor.java
+++ b/phoenix-core-server/src/main/java/org/apache/phoenix/replication/reader/ReplicationLogProcessor.java
@@ -29,7 +29,6 @@ import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
 import java.util.concurrent.Future;
 
-import com.google.common.util.concurrent.ThreadFactoryBuilder;
 import org.apache.hadoop.conf.Configuration;
 import org.apache.hadoop.fs.FileSystem;
 import org.apache.hadoop.fs.Path;
@@ -46,6 +45,7 @@ import org.apache.phoenix.replication.log.LogFileReader;
 import org.apache.phoenix.replication.log.LogFileReaderContext;
 import org.apache.phoenix.replication.metrics.MetricsReplicationLogProcessor;
 import org.apache.phoenix.replication.metrics.MetricsReplicationLogProcessorImpl;
+import org.apache.phoenix.thirdparty.com.google.common.util.concurrent.ThreadFactoryBuilder;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -152,7 +152,8 @@ public class ReplicationLogProcessor implements Closeable {
      * @return ReplicationLogProcessor instance
      */
     public static ReplicationLogProcessor get(Configuration conf, String haGroupId) {
-        return INSTANCES.computeIfAbsent(haGroupId, k -> new ReplicationLogProcessor(conf, haGroupId));
+        return INSTANCES.computeIfAbsent(haGroupId, 
+            k -> new ReplicationLogProcessor(conf, haGroupId));
     }
 
     /**
@@ -175,7 +176,10 @@ public class ReplicationLogProcessor implements Closeable {
                 DEFAULT_REPLICATION_STANDBY_LOG_REPLAY_THREAD_POOL_SIZE);
         decorateConf();
         this.metrics = createMetricsSource();
-        this.executorService = Executors.newFixedThreadPool(threadPoolSize, new ThreadFactoryBuilder().setNameFormat("Phoenix-Replication-Log-Processor-" + haGroupId + "-%d").build());
+        this.executorService = Executors.newFixedThreadPool(threadPoolSize, 
+            new ThreadFactoryBuilder()
+                .setNameFormat("Phoenix-Replication-Log-Processor-" + haGroupId + "-%d")
+                .build());
     }
 
     /**
@@ -252,7 +256,8 @@ public class ReplicationLogProcessor implements Closeable {
 
     /**
      * Creates a LogFileReader for the specified file path.
-     * Validates that the file exists and initializes the reader with the given file system and path.
+     * Validates that the file exists and initializes the reader with the given 
+     * file system and path.
      * @param fs The file system to use for reading
      * @param filePath The path to the log file
      * @return A configured LogFileReader instance
@@ -307,7 +312,7 @@ public class ReplicationLogProcessor implements Closeable {
 
         long startTime = System.currentTimeMillis();
 
-        while(attempt <= batchRetryCount && !currentOperations.isEmpty()) {
+        while (attempt <= batchRetryCount && !currentOperations.isEmpty()) {
             if (attempt > 0) {
                 LOG.warn("Retrying failed batch operations, attempt {} of {}",
                         attempt, batchRetryCount);
@@ -371,7 +376,7 @@ public class ReplicationLogProcessor implements Closeable {
     protected ApplyMutationBatchResult applyMutations(
             Map<TableName, List<Mutation>> tableMutationMap) throws IOException {
 
-        if(tableMutationMap == null || tableMutationMap.isEmpty()) {
+        if (tableMutationMap == null || tableMutationMap.isEmpty()) {
             return new ApplyMutationBatchResult(Collections.emptyMap(), null);
         }
 
@@ -437,12 +442,12 @@ public class ReplicationLogProcessor implements Closeable {
     public void close() throws IOException {
         synchronized (this) {
             // Close the async connection
-            if(asyncConnection != null && !asyncConnection.isClosed()) {
+            if (asyncConnection != null && !asyncConnection.isClosed()) {
                 asyncConnection.close();
             }
             asyncConnection = null;
             // Shutdown the executor service
-            if(executorService != null && !executorService.isShutdown()) {
+            if (executorService != null && !executorService.isShutdown()) {
                 executorService.shutdownNow();
             }
             // Remove the instance from cache
@@ -497,8 +502,10 @@ public class ReplicationLogProcessor implements Closeable {
         private final Map<TableName, List<Mutation>> failedMutations;
         private final Exception exception;
 
-        public ApplyMutationBatchResult(Map<TableName, List<Mutation>> failedMutations, Exception exception) {
-            this.failedMutations = failedMutations != null ? failedMutations : Collections.emptyMap();
+        public ApplyMutationBatchResult(Map<TableName, List<Mutation>> failedMutations, 
+                Exception exception) {
+            this.failedMutations = failedMutations != null ? 
+                failedMutations : Collections.emptyMap();
             this.exception = exception;
         }
 
--- a/phoenix-core-server/src/main/java/org/apache/phoenix/replication/reader/ReplicationLogProcessor.java
+++ b/phoenix-core-server/src/main/java/org/apache/phoenix/replication/reader/ReplicationLogProcessor.java
@@ -1,0 +1,288 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.phoenix.replication.reader;
+
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Future;
+
+import org.apache.hadoop.conf.Configuration;
+import org.apache.hadoop.fs.FileSystem;
+import org.apache.hadoop.fs.Path;
+import org.apache.hadoop.hbase.HBaseConfiguration;
+import org.apache.hadoop.hbase.HConstants;
+import org.apache.hadoop.hbase.TableName;
+import org.apache.hadoop.hbase.client.AsyncConnection;
+import org.apache.hadoop.hbase.client.AsyncTable;
+import org.apache.hadoop.hbase.client.ConnectionFactory;
+import org.apache.hadoop.hbase.client.Mutation;
+import org.apache.hadoop.hbase.client.RetriesExhaustedException;
+import org.apache.hadoop.hbase.util.FutureUtils;
+import org.apache.phoenix.replication.log.LogFile;
+import org.apache.phoenix.replication.log.LogFileReader;
+import org.apache.phoenix.replication.log.LogFileReaderContext;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class ReplicationLogProcessor {
+
+    private static final Logger LOG = LoggerFactory.getLogger(ReplicationLogProcessor.class);
+
+    /**
+     * The maximum count of mutations to process in single batch while reading replication log file
+     */
+    public static final String REPLICATION_STANDBY_LOG_REPLAY_BATCH_SIZE =
+            "phoenix.replication.log.standby.replay.batch.size";
+
+    /**
+     * The default batch size for reading the replication log file.
+     * Assuming each log record to be 10 KB (un-compressed) and allow at max 64 MB of
+     * in-memory records to be processed
+     */
+    public static final int DEFAULT_REPLICATION_STANDBY_LOG_REPLAY_BATCH_SIZE = 6400;
+
+    /**
+     * The maximum number of retries for HBase client operations while applying the mutations
+     */
+    public static final String REPLICATION_STANDBY_HBASE_CLIENT_RETRIES_COUNT =
+            "phoenix.replication.standby.hbase.client.retries.number";
+
+    /**
+     * The default number of retries for HBase client operations while applying the mutations.
+     */
+    public static final int DEFAULT_REPLICATION_STANDBY_HBASE_CLIENT_RETRIES_COUNT = 4;
+
+    /**
+     * The timeout for HBase client operations while applying the mutations.
+     */
+    public static final String REPLICATION_STANDBY_HBASE_CLIENT_OPERATION_TIMEOUT_MS =
+            "phoenix.replication.standby.hbase.client.operations.timeout";
+
+    /**
+     * The default timeout for HBase client operations while applying the mutations.
+     */
+    public static final int DEFAULT_REPLICATION_STANDBY_HBASE_CLIENT_OPERATION_TIMEOUT_MS = 10000;
+
+    private final Configuration conf;
+
+    private final ExecutorService executorService;
+
+    /**
+     * This {@link AsyncConnection} is used for handling mutations
+     */
+    private volatile AsyncConnection asyncConnection;
+
+    private final Object asyncConnectionLock = new Object();
+
+    private final int batchSize;
+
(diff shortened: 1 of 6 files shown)


phoenix-core-server/src/main/java/org/apache/phoenix/replication/reader/ReplicationLogProcessor.java
```diff
@@ -0,0 +1,288 @@
...
+
+    private final int batchSize;
+
+    /**
+     * Creates a new ReplicationLogProcessor with the given configuration and executor service.
+     * @param conf The configuration to use
+     * @param executorService The executor service for processing mutations
+     * @throws IOException if initialization fails
+     */
+    public ReplicationLogProcessor(final Configuration conf,
```
Himanshu-g81: Injecting executor service via constructor so that each HA group can have independent set of threads to process the mutations.

phoenix-core-server/src/main/java/org/apache/phoenix/replication/reader/ReplicationLogProcessor.java
```diff
@@ -0,0 +1,288 @@
...
+    /**
+     * The maximum number of retries for HBase client operations while applying the mutations
+     */
+    public static final String REPLICATION_STANDBY_HBASE_CLIENT_RETRIES_COUNT =
+            "phoenix.replication.standby.hbase.client.retries.number";
+
+    /**
+     * The default number of retries for HBase client operations while applying the mutations.
+     */
+    public static final int DEFAULT_REPLICATION_STANDBY_HBASE_CLIENT_RETRIES_COUNT = 4;
```
apurtell, line 72: Why 4? Why not 3? or 5? Does this align with other retry defaults, like in the code we added on the writer side? (It does not, fwiw..)
These are just starting points for users, sure, but should work well without being changed. The default number of retries and the default timeouts work together to establish a time budget for transient issues. After this budget is exhausted the issues are raised by exception to higher layers, which may in some cases abort the server.
Let's think them through.

phoenix-core-server/src/main/java/org/apache/phoenix/replication/reader/ReplicationLogProcessor.java
```diff
@@ -0,0 +1,288 @@
...
+    /**
+     * The timeout for HBase client operations while applying the mutations.
+     */
+    public static final String REPLICATION_STANDBY_HBASE_CLIENT_OPERATION_TIMEOUT_MS =
+            "phoenix.replication.standby.hbase.client.operations.timeout";
+
+    /**
+     * The default timeout for HBase client operations while applying the mutations.
+     */
+    public static final int DEFAULT_REPLICATION_STANDBY_HBASE_CLIENT_OPERATION_TIMEOUT_MS = 10000;
```
apurtell: So the total time we'd wait before throwing an exception up to the higher layers if consistently timing out is 50 seconds (10s + (4 retries * 10s))? Why 50? Should it be 60? or 30? or 120? or 300? We are applying edits on the standby, we can certainly be more tolerant than on the write site, where timeouts/retries need to be short to fail fast. We don't need to fail fast in the reader. On the other hand we don't want to wait so long as to complicate failover or miss a failover SLA.

phoenix-core-server/src/main/java/org/apache/phoenix/replication/reader/ReplicationLogProcessor.java
```diff
@@ -0,0 +1,288 @@
...
+    private final Configuration conf;
+
+    private final ExecutorService executorService;
+
+    /**
+     * This {@link AsyncConnection} is used for handling mutations
+     */
+    private volatile AsyncConnection asyncConnection;
+
+    private final Object asyncConnectionLock = new Object();
```
apurtell: Do you need this? Taking a lock on this to lazily initialize the `asyncConnection` is equivalent to `synchronized (this)` so just use that?

phoenix-core-server/src/main/java/org/apache/phoenix/replication/reader/ReplicationLogProcessor.java
```diff
@@ -0,0 +1,288 @@
...
+        if (tableMutationMap == null || tableMutationMap.isEmpty()) {
+            return;
+        }
+
+        List<Future<?>> futures = new ArrayList<>();
+        for (Map.Entry<String, List<Mutation>> entry : tableMutationMap.entrySet()) {
+            String tableName = entry.getKey();
+            List<Mutation> mutations = entry.getValue();
+            AsyncTable<?> table = getAsyncConnection()
+                    .getTable(TableName.valueOf(tableName), executorService);
```
apurtell: How about `Map<TableName, List<Mutation>>` instead?

phoenix-core-server/src/main/java/org/apache/phoenix/replication/reader/ReplicationLogProcessor.java
```diff
@@ -0,0 +1,288 @@
...
+            futures.add(table.batchAll(mutations));
+        }
+
+        IOException error = null;
+
+        for (Future<?> future : futures) {
+            try {
+                FutureUtils.get(future);
+            } catch (RetriesExhaustedException e) {
```
apurtell: Is this the only exception type `get()`might throw here? Should we be catching `IOException` instead?

phoenix-core/src/test/java/org/apache/phoenix/replication/reader/ReplicationLogProcessorTest.java
```diff
@@ -0,0 +1,720 @@
...
+            assertTrue("Error message should mention file does not exist and file path name",
+                    exception.getMessage().contains("TableNotFoundException"));
+        }
+    }
+
+    /**
+     * Tests behavior when HBase operations fail (simulated by disabling table).
+     */
+    @Test
+    public void testProcessReplicationLogBatchWithHBaseFailure() throws Exception {
```
apurtell: It would also be useful to have a test case where only part of a batch fails. This means making a table with multiple regions, building a set of mutations that would be applied to all regions, and taking one of those regions offline.

apurtell: Right now we'd have an equivalent result... a thrown IOException. However as the implementation evolves we might have a more granular response. Resubmission of only the mutations involved with failed futures, etc. Do you want to try to tackle this now, even? It's much more likely a regionserver is down than a table is disabled, so batches are likely to only partially fail.

phoenix-core/src/test/java/org/apache/phoenix/replication/reader/ReplicationLogProcessorTest.java
```diff
@@ -0,0 +1,720 @@
...
+        // Validate that the batch size is correctly set
+        assertEquals("Batch size should be set correctly", batchSize, replicationLogProcessor.getBatchSize());
+        ReplicationLogProcessor spyProcessor = Mockito.spy(replicationLogProcessor);
+        // Mock the processReplicationLogBatch method
+        Mockito.doNothing().when(spyProcessor).processReplicationLogBatch(Mockito.any(Map.class));
+        // Process the log file
+        spyProcessor.processLogFile(localFs, multiTableBatchFilePath);
+        // Verify processReplicationLogBatch was called the expected number of times
+        Mockito.verify(spyProcessor, Mockito.times(expectedBatchCalls))
+            .processReplicationLogBatch(Mockito.any(Map.class));
```
apurtell: You can also use your spy to check if expected mutations were found in each invocation of processReplicationLogBatch, put1 in the first, put2 in the second, as a sanity check.

phoenix-core/src/test/java/org/apache/phoenix/replication/reader/ReplicationLogProcessorTest.java
```diff
@@ -0,0 +1,720 @@
...
+            replicationLogProcessor.processLogFile(localFs, emptyFilePath);
+            // If we reach here, the empty file was processed successfully
+            assertTrue("Processing empty log file should complete without errors", true);
+        } catch (Exception e) {
+            fail("Processing empty log file should not throw exception: " + e.getMessage());
+        }
+    }
+
+    /**
+     * Tests processing of log files that were not closed, ensuring it's successf.
```
apurtell: Spelling

phoenix-core/src/test/java/org/apache/phoenix/replication/reader/ReplicationLogProcessorTest.java
```diff
@@ -0,0 +1,720 @@
...
+        // Process the file without closing - should not throw any exceptions
+        ReplicationLogProcessor spyProcessor = Mockito.spy(new ReplicationLogProcessor(conf, executorService));
+        Mockito.doNothing().when(spyProcessor).processReplicationLogBatch(Mockito.any(Map.class));
+
+        spyProcessor.processLogFile(localFs, emptyFilePath);
+
+        // Verify processReplicationLogBatch was called the expected number of times
+        Mockito.verify(spyProcessor, Mockito.times(1))
+                .processReplicationLogBatch(Mockito.any(Map.class));
+    }
```
apurtell, line 1033: Can also verify here that the expected mutation was seen, `put` with sequence `3L`.

phoenix-core/src/test/java/org/apache/phoenix/replication/reader/ReplicationLogProcessorTest.java
```diff
@@ -0,0 +1,720 @@
...
+        // Mock the processReplicationLogBatch method to do nothing but track calls
+        Mockito.doNothing().when(spyProcessor).processReplicationLogBatch(Mockito.any(Map.class));
+
+        // Process the log file
+        spyProcessor.processLogFile(localFs, batchTestFilePath);
+
+        // Verify processReplicationLogBatch was called the expected number of times
+        Mockito.verify(spyProcessor, Mockito.times(expectedBatchCalls))
+                .processReplicationLogBatch(Mockito.any(Map.class));
+    }
```
apurtell, line 1580: You can also verify that the expected mutations for the batch were found in the batch.

phoenix-core-server/src/main/java/org/apache/phoenix/replication/reader/ReplicationLogProcessor.java
apurtell: Alternatively, if you add an init() method then the async connection can be set up there once by the code that has just constructed the object.

    // no exception from constructor, good coding practice
    ReplicationLogProcessor processor = new ReplicationLogProcessor(...);
    // now init can create the async connection, do other setups which might throw, ...
    processor.init();
    ...

phoenix-core-server/src/main/java/org/apache/phoenix/replication/reader/ReplicationLogProcessor.java
```diff
@@ -0,0 +1,288 @@
...
+import org.apache.hadoop.hbase.client.Mutation;
+import org.apache.hadoop.hbase.client.RetriesExhaustedException;
+import org.apache.hadoop.hbase.util.FutureUtils;
+import org.apache.phoenix.replication.log.LogFile;
+import org.apache.phoenix.replication.log.LogFileReader;
+import org.apache.phoenix.replication.log.LogFileReaderContext;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class ReplicationLogProcessor {
```
apurtell: This should be a singleton, right? 
So you should implement the singleton pattern... private constructor, class method for getting the singleton instance, lazy initialization of the singleton instance, ... ?

phoenix-core-server/src/main/java/org/apache/phoenix/replication/reader/ReplicationLogProcessor.java
Himanshu-g81: I thought of keeping it less than one round time (1 min), hence kept it 50. With latest commit (including the suggestion for having retries only for failed batch as well) and updating the retry timeout to 8 sec, we would max wait for higher layers of close to 2 mins (i.e 127 seconds (10 + 4*8) * 3 + (1 + 2 + 4 - expo backoff during retries for failed mutations).
If we agree that 2 - 3 mins is max delay that we want on target cluster (when we are sure that file is written to target cluster) beyond which we can raise an alert in failure scenarios.

Please note, all these values (hbase operation retries, timeout, failed batch retry) are exposed via configurations so we can tune them as per need.

phoenix-core-server/src/main/java/org/apache/phoenix/replication/reader/ReplicationLogProcessor.java
Himanshu-g81: Sure, I usually keep single lock object per attribute that should be explicitly locked before access (instead of taking lock over the object itself), which might be useful in case we have more such attributes.
But in this case since we only have single attribute, using object itself would also have same effect. Thanks for pointing this.

> if you add an init()
I thought if multiple threads at upper layer are using this object and one of them closed it, others should not be impacted. It's not expected but just added it via getConnection() method to be on safe side.

phoenix-core-server/src/main/java/org/apache/phoenix/replication/reader/ReplicationLogProcessor.java
Himanshu-g81: Thanks, it can throw IOException. Fixed in next commit.

phoenix-core/src/test/java/org/apache/phoenix/replication/reader/ReplicationLogProcessorTest.java
Himanshu-g81: In a batch, one future is submitted per table. So mutations of single table would be retried together.
I have added retries (configurable) only for failed futures as well in latest commit ([6b3fc83](https://github.com/apache/phoenix/pull/2188/commits/6b3fc83b0d626ea35a416bd172353d0784d26acd)) and also the test cases when region is down temporarily but comes up during retry ensuring we have resiliance for temporariy down regions.

phoenix-core/src/test/java/org/apache/phoenix/replication/reader/ReplicationLogProcessorTest.java
Himanshu-g81: Sure, added in latest commit ([6b3fc83](https://github.com/apache/phoenix/pull/2188/commits/6b3fc83b0d626ea35a416bd172353d0784d26acd)).

phoenix-core-server/src/main/java/org/apache/phoenix/replication/reader/ReplicationLogProcessor.java
Himanshu-g81: Thanks for the disscusion over slack.
As discussed we will have the top level class (which would initialize the individual replication compoenent per HA group) would be singleton and sub-componenets would be HA group level, so keeping it non-singleton for now.

phoenix-core-server/src/main/java/org/apache/phoenix/replication/reader/ReplicationLogProcessor.java
```diff
@@ -0,0 +1,417 @@
...
+        // Check results and track failures
+        for (Map.Entry<TableName, Future<?>> entry : futures.entrySet()) {
+            TableName tableName = entry.getKey();
+            Future<?> future = entry.getValue();
+            try {
+                FutureUtils.get(future);
+            } catch (IOException e) {
+                // Add failed mutations to retry list
+                failedOperations.put(tableName, tableMutationMap.get(tableName));
+                LOG.warn("Failed to apply mutations for table {}: {}", tableName, e.getMessage());
```
apurtell: This will probably be pretty noisy? Can we do something else? I would recommend the detail log at DEBUG and a metric for failed mutation application we can track instead of a likely flood of WARN level messages every time there is an issue.

phoenix-core-server/src/main/java/org/apache/phoenix/replication/reader/ReplicationLogProcessor.java
```diff
@@ -0,0 +1,417 @@
...
+                Map<TableName, List<Mutation>> failedOperations = applyMutations(currentOperations);
+                
+                // If no failures, we're done
+                if (failedOperations.isEmpty()) {
+                    return;
+                }
+                
+                // Update current operations for next retry
+                currentOperations = failedOperations;
+                lastError = new IOException("Failed to apply the mutations");
```
apurtell: No details?

phoenix-core-server/src/main/java/org/apache/phoenix/replication/reader/ReplicationLogProcessor.java
apurtell: Speaking of metrics, you should have a metrics source for the log processor that includes at least the failed mutation count but also consider latency histograms and other details useful for monitoring and debugging runtime issues.

Himanshu-g81: Sure, I have made it debug mode. I was planning to add metrics separately for all the replay components, but I have added the initial set for ReplicationLogProcessor (failedMutationCount, failedBatchCount, logFileReplayFailureCount, logFileReplaySuccessCount, batchReplayTime and logFileReplayTime) in recent commit.

phoenix-core-server/src/main/java/org/apache/phoenix/replication/reader/ReplicationLogProcessor.java
Himanshu-g81: Exception was logged in applyMutations method previously, however as we plan to move to debug level, have added custom return type for applyMutations method to include both failedMutations and exception observed.
Same exception is propagated now.

Himanshu-g81:
Rebased after update of PHOENIX-7562-feature branch.

phoenix-core-server/src/main/java/org/apache/phoenix/replication/reader/ReplicationLogProcessor.java
```diff
@@ -0,0 +1,524 @@
...
+
+    private final int batchSize;
+
+    private final int batchRetryCount;
+
+    private final long maxRetryDelayMs;
+
+    private final MetricsReplicationLogProcessor metrics;
+
+    /** Cache of ReplicationLogGroup instances by HA Group ID */
```
kadirozde: ID -> Name

phoenix-core-server/src/main/java/org/apache/phoenix/replication/reader/ReplicationLogProcessor.java
```diff
@@ -0,0 +1,524 @@
...
+            // Create the LogFileReader for given path
+            logFileReader = createLogFileReader(fs, filePath);
+
+            for (LogFile.Record record : logFileReader) {
+                final TableName tableName = TableName.valueOf(record.getHBaseTableName());
+                final Mutation mutation = record.getMutation();
+
+                tableToMutationsMap.computeIfAbsent(tableName, k -> new ArrayList<>())
+                        .add(mutation);
+                currentBatchSize++;
```
kadirozde, line 246: I think we need two batch sizes, one with using the number of mutations and the other using total byte size, that is, batchSize vs batchSizeBytes. For calculating batchSizeBytes, we can use mutation.size().

Himanshu-g81: Sure, added in latest commit (including test case for the same).
Please note, have used [mutation.heapSize()](https://github.com/apache/hbase/blob/8ddf925daac7af48a5b624c6192bd2cdc45f7955/hbase-client/src/main/java/org/apache/hadoop/hbase/client/Mutation.java#L412-L436) to get the in memory size of this mutation as I observed [mutation.size()](https://github.com/apache/hbase/blob/8ddf925daac7af48a5b624c6192bd2cdc45f7955/hbase-client/src/main/java/org/apache/hadoop/hbase/client/Mutation.java#L397-L403) returns count of key values pairs.

PR #2191, opened 2025-06-12
PHOENIX-7639 Improve error handling in PhoenixMasterObserver

virajjasani:
Jira: PHOENIX-7639

If the cluster is upgraded to Phoenix 5.3 without metadata upgrade, new system tables SYSTEM.CDC_STREAM_STATUS and SYSTEM.CDC_STREAM will not be created yet.

Any region merges or splits with PhoenixMasterObserver coproc hooks can encounter errors as without the system tables, we cannot persist stream partition metadata related to region splits and merges. We need to handle this error.

Moreover, it is important to not miss the updates to SYSTEM.CDC_STREAM table as part of postCompletedSplitRegionAction() and postCompletedMergeRegionsAction(). The purpose of this Jira is to introduce maximum retries possible to make the upserts successful.

### DIFF
--- a/phoenix-core-server/src/main/java/org/apache/phoenix/coprocessor/PhoenixMasterObserver.java
+++ b/phoenix-core-server/src/main/java/org/apache/phoenix/coprocessor/PhoenixMasterObserver.java
@@ -82,7 +82,8 @@ public class PhoenixMasterObserver implements MasterObserver, MasterCoprocessor 
             = "UPSERT INTO " + SYSTEM_CDC_STREAM_NAME + " (TABLE_NAME, STREAM_NAME, PARTITION_ID, "
             + "PARENT_PARTITION_ID, PARTITION_END_TIME) VALUES (?,?,?,?,?)";
 
-    private static final int MAX_RETRY_COUNT = 20;
+    public static final String PHOENIX_MASTER_MAX_RETRY_COUNT = "phoenix.master.max.retry.count";
+    public static final int DEFAULT_PHOENIX_MASTER_MAX_RETRY_COUNT = 20;
 
     @Override
     public Optional<MasterObserver> getMasterObserver() {
@@ -109,6 +110,8 @@ public class PhoenixMasterObserver implements MasterObserver, MasterCoprocessor 
                                                final RegionInfo regionInfoA,
                                                final RegionInfo regionInfoB) throws IOException {
         Configuration conf = c.getEnvironment().getConfiguration();
+        int maxRetryCount =
+                conf.getInt(PHOENIX_MASTER_MAX_RETRY_COUNT, DEFAULT_PHOENIX_MASTER_MAX_RETRY_COUNT);
         int tries = 0;
         Exception caughtException;
         do {
@@ -158,10 +161,10 @@ public class PhoenixMasterObserver implements MasterObserver, MasterCoprocessor 
                 metricSource.incrementPostSplitPartitionUpdateFailureCount();
                 caughtException = e;
             }
-        } while (tries++ < MAX_RETRY_COUNT);
+        } while (tries++ < maxRetryCount);
         // All retries were exhausted
         throw new IOException(
-                "Failed to update CDC Stream Partition metadata after " + MAX_RETRY_COUNT +
+                "Failed to update CDC Stream Partition metadata after " + maxRetryCount +
                         " retries during split. Daughter regions: " +
                         regionInfoA.getEncodedName() + " " +
                         regionInfoB.getEncodedName(), caughtException);
@@ -180,6 +183,7 @@ public class PhoenixMasterObserver implements MasterObserver, MasterCoprocessor 
                                                 final RegionInfo[] regionsToMerge,
                                                 final RegionInfo mergedRegion) throws IOException {
         Configuration conf = c.getEnvironment().getConfiguration();
+        int maxRetryCount = conf.getInt(PHOENIX_MASTER_MAX_RETRY_COUNT, DEFAULT_PHOENIX_MASTER_MAX_RETRY_COUNT);
         int tries = 0;
         Exception caughtException;
         do {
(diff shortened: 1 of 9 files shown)


phoenix-core/src/it/java/org/apache/phoenix/end2end/CDCStream2IT.java
```diff
@@ -0,0 +1,174 @@
...
+        assertTrue(mergedDaughter.stream().anyMatch(
+                d -> Objects.equals(d.parentPartitionId, splitParents.get(0).partitionId)));
+        assertTrue(mergedDaughter.stream().anyMatch(
+                d -> Objects.equals(d.parentPartitionId, splitParents.get(1).partitionId)));
+        for (PartitionMetadata splitDaughter : splitParents) {
+            Assert.assertEquals(mergedDaughter.get(0).startTime, splitDaughter.endTime);
+        }
+    }
+
+    private void createTableAndEnableCDC(Connection conn, String tableName) throws Exception {
```
palashc: Should we move the util methods to a common place and re-use in both test classes?

phoenix-core-server/src/main/java/org/apache/phoenix/coprocessor/PhoenixMasterObserver.java
```diff
@@ -74,11 +82,19 @@ public class PhoenixMasterObserver implements MasterObserver, MasterCoprocessor
             = "UPSERT INTO " + SYSTEM_CDC_STREAM_NAME + " (TABLE_NAME, STREAM_NAME, PARTITION_ID, "
             + "PARENT_PARTITION_ID, PARTITION_END_TIME) VALUES (?,?,?,?,?)";
 
+    private static final int MAX_RETRY_COUNT = 20;
```
palashc: Can this be in the config with default value of 20?

palashc (approved):
Couple of nits, lgtm otherwise.

phoenix-core/src/it/java/org/apache/phoenix/end2end/CDCStream2IT.java
```diff
@@ -0,0 +1,174 @@
...
+        assertEquals("Post split partition update failures should be 0 initially",
+                0, METRICS_SOURCE.getPostSplitPartitionUpdateFailureCount());
+        assertEquals("Post merge partition update failures should be 0 initially",
+                0, METRICS_SOURCE.getPostMergePartitionUpdateFailureCount());
+
+        // Perform a split operation - this will fail 24 times before succeeding
+        TestUtil.splitTable(conn, tableName, Bytes.toBytes("m"));
+
+        // Verify split metric is 24
+        assertEquals("Post split partition update failures should be 15 after retries",
```
palashc: typo in errorMsg in assert?

phoenix-core-server/src/main/java/org/apache/phoenix/coprocessor/PhoenixMasterObserver.java
virajjasani: We could but probably good to avoid config here. I verified that HBase will also retry when coproc throws IOE here anyways.
20 or 40, eventually we need some retries here, otherwise error will anyways trigger Split/Merge procedures to retry.

phoenix-core/src/it/java/org/apache/phoenix/end2end/CDCStream2IT.java
virajjasani: Nice catch, let me update

phoenix-core/src/it/java/org/apache/phoenix/end2end/CDCStream2IT.java
virajjasani: Yes we can do that but here, i am trying to use default behavior i.e. let the TaskObserver take care of updating the Stream status rather than trigger it manually:
```
        TaskRegionObserver.SelfHealingTask task =
                new TaskRegionObserver.SelfHealingTask(
                        TaskRegionEnvironment, QueryServicesOptions.DEFAULT_TASK_HANDLING_MAX_INTERVAL_MS);
        task.run();
```

virajjasani: Let me however refactor by creating common utility.

palashc: Yeah like `PartitionMetadata` and `createTableAndEnableCDC` etc can be shared

virajjasani: Done

phoenix-core-server/src/main/java/org/apache/phoenix/coprocessor/PhoenixMasterObserver.java
virajjasani: Added config for flexibility

phoenix-core-server/src/main/java/org/apache/phoenix/coprocessor/PhoenixMasterObserver.java
```diff
@@ -82,7 +82,8 @@ public class PhoenixMasterObserver implements MasterObserver, MasterCoprocessor
             = "UPSERT INTO " + SYSTEM_CDC_STREAM_NAME + " (TABLE_NAME, STREAM_NAME, PARTITION_ID, "
             + "PARENT_PARTITION_ID, PARTITION_END_TIME) VALUES (?,?,?,?,?)";
 
-    private static final int MAX_RETRY_COUNT = 20;
+    public static final String PHOENIX_MASTER_MAX_RETRY_COUNT = "phoenix.master.max.retry.count";
```
palashc: `phoenix.master.observer.max.retry.count` ?

virajjasani: I doubt if we use `observer` in any configs, Observer is internal coproc concept. Let me check if we use it in config keys already.

virajjasani: We don't use it but nothing wrong with including `observer` in the config key name. Let me do it.

PR #2192, opened 2025-06-14
PHOENIX-7601 Use synchronous replication in Phoenix coprocs on the write path

tkhurana:


phoenix-core-server/src/main/java/org/apache/phoenix/hbase/index/IndexRegionObserver.java
```diff
@@ -448,6 +515,23 @@ public void start(CoprocessorEnvironment e) throws IOException {
           BloomType bloomFilterType = tableDescriptor.getColumnFamilies()[0].getBloomFilterType();
           // when the table descriptor changes, the coproc is reloaded
           this.useBloomFilter = bloomFilterType == BloomType.ROW;
+          byte[] tableName = env.getRegionInfo().getTable().getName();
+          boolean isSystemTable = SchemaUtil.isSystemTable(tableName);
```
kadirozde: I think we should move this check into a method in SchemaUtil. There we need to state that it is assumed that DDL statements that are executed within global (not tenant specific) connections are executed by an admin or a admin process in each clusters separately and thus are not replicated via the HBase or Phoenix replication. We also need to go through for each system table and explain why we do not replicate them. I feel that we may need to replicate them. It is okay to have a TO DO for them in this PR and revisit them later in a separate Jira.

phoenix-core/src/it/java/org/apache/phoenix/replication/ReplicationLogGroupIT.java
```diff
@@ -0,0 +1,339 @@
...
+            expected.put(indexName2, rowCount * 2);
+            // we didn't create any tenant views so no change in the syscat entries
+            expected.put(SYSTEM_CATALOG_NAME, 0);
+            expected.put(SYSTEM_CHILD_LINK_NAME, 0);
+            verifyReplication(conn, expected);
+        }
+    }
+
+    @Test
+    public void testPreWALRestoreSkip() throws Exception {
```
kadirozde: This test name should be testWALRestore, I think. We should add a high level description of what this test does, something like this test simulates region server crashes in the middle of write transactions.

phoenix-core/src/it/java/org/apache/phoenix/replication/ReplicationLogGroupIT.java
```diff
@@ -0,0 +1,339 @@
...
+            String ddl = String.format("create table %s (id1 integer not null, " +
+                    "id2 integer not null, val1 varchar, val2 varchar " +
+                    "constraint pk primary key (id1, id2))", tableName);
+            conn.createStatement().execute(ddl);
+            ddl = String.format("create index %s on %s (val1) include (val2)",
+                    indexName, tableName);
+            conn.createStatement().execute(ddl);
+            conn.commit();
+        }
+        JVMClusterUtil.RegionServerThread rs2 = cluster.startRegionServer();
```
kadirozde, line 290: It will be useful to add a comment here to explain why this is a second region server, something like, a mini cluster comes with a single region server and here we add a second region server to the cluster.

phoenix-core/src/it/java/org/apache/phoenix/replication/ReplicationLogGroupIT.java
```diff
@@ -0,0 +1,339 @@
...
+    public void testSystemTables() throws Exception {
+        createViewHierarchy();
+        Map<String, List<Mutation>> logsByTable = groupLogsByTable();
+        dumpTableLogCount(logsByTable);
+        Map<String, List<Mutation>> systemTables = logsByTable.entrySet().stream()
+                .filter(entry -> entry.getKey().startsWith(QueryConstants.SYSTEM_SCHEMA_NAME))
+                .collect(Collectors.toMap(Map.Entry::getKey, Map.Entry::getValue));
+        assertEquals(2, systemTables.size());
+        assertEquals(1, getCountForTable(systemTables, SYSTEM_CHILD_LINK_NAME));
+        assertTrue(getCountForTable(systemTables, SYSTEM_CATALOG_NAME) > 0);
```
kadirozde, line 337: Should this be an equality check, = 1 ?

phoenix-core/src/it/java/org/apache/phoenix/replication/ReplicationLogGroupIT.java
```diff
@@ -0,0 +1,339 @@
...
+            // atleast 1 log entry for syscat
+            expected.put(SYSTEM_CATALOG_NAME, 1);
+            verifyReplication(conn, expected);
+        }
+    }
+
+    @Test
+    public void testSystemTables() throws Exception {
+        createViewHierarchy();
+        Map<String, List<Mutation>> logsByTable = groupLogsByTable();
```
kadirozde, line 340: We should have a test at for a system table that is not the catalog or child link table and show that it is not replicated.

kadirozde (approved):
+1, Thanks!

PR #2193, opened 2025-06-17
PHOENIX-7641 Support placeholder for document field keys in BSON condition expression

virajjasani:
Jira: PHOENIX-7641

BSON supports various non-alphanumeric values as document field keys which are case sensitive. Not all can be used in the BSON_CONDITION_EXPRESSION() antlr grammar as the field key can have duplicate meaning with operator. e.g. ">" is valid field key, therefore condition expression "> > :val" is valid because the expression tries to evaluate whether the value of ">" key is greater than the value represented by ":val" value placeholder. However, the antlr grammar needs to recognize ">" as only "greater-than" symbol to avoid ambiguity.

For complex key names like this, it is recommended to provide them as key alias placeholders as part of the condition expression document argument, which already consists of value alias placeholders.

Example:
```
{
  "$EXPR": "#press = :press AND #track[0].#shot[2][0].#city.#standard[50] = :softly",
  "$VAL": {
    ":press": "beat",
    ":softly": "softly"
  },
  "$KEYS": {
    "#press": "press",
    "#track": "track",
    "#shot": "shot",
    "#city": "city",
    "#standard": "standard"
  }
}
```


virajjasani:
Build results look good.

PR #2196, opened 2025-06-18
PHOENIX-7644 : CDC Stream improvements

palashc:
1. Remove CDC Object from SYSTEM.CATALOG when table is dropped
2. Set UCF=2h on CDC streams system tables.
3. try-with-resource for PhoenixMasterObserver

### DIFF
--- a/phoenix-core-client/src/main/java/org/apache/phoenix/util/CDCUtil.java
+++ b/phoenix-core-client/src/main/java/org/apache/phoenix/util/CDCUtil.java
@@ -99,6 +99,19 @@ public class CDCUtil {
         return indexName.startsWith(CDC_INDEX_PREFIX);
     }
 
+    public static boolean isCDCIndex(byte[] indexNameBytes) {
+        String indexName = Bytes.toString(indexNameBytes);
+        return indexName.startsWith(CDC_INDEX_PREFIX);
+    }
+
+    public static byte[] getCdcObjectName(byte[] cdcIndexName) {
+        byte[] result = new byte[cdcIndexName.length - CDC_INDEX_PREFIX.length()];
+        for (int i = 0; i < cdcIndexName.length - CDC_INDEX_PREFIX.length(); i++) {
+            result[i] = cdcIndexName[i + CDC_INDEX_PREFIX.length()];
+        }
+        return result;
+    }
+
     public static boolean isCDCIndex(PTable indexTable) {
         return isCDCIndex(indexTable.getTableName().getString());
     }
--- a/phoenix-core-client/src/main/java/org/apache/phoenix/util/CDCUtil.java
+++ b/phoenix-core-client/src/main/java/org/apache/phoenix/util/CDCUtil.java
@@ -22,6 +22,7 @@ import java.sql.SQLException;
 import java.sql.Types;
 import java.text.DateFormat;
 import java.text.SimpleDateFormat;
+import java.util.Arrays;
 import java.util.Base64;
 import java.util.Date;
 import java.util.HashSet;
@@ -39,7 +40,6 @@ import org.apache.phoenix.exception.SQLExceptionInfo;
 import org.apache.phoenix.execute.DescVarLengthFastByteComparisons;
 import org.apache.phoenix.schema.PTable;
 import org.apache.phoenix.schema.types.PDataType;
-import org.apache.phoenix.schema.types.PVarchar;
 import org.bson.RawBsonDocument;
 
 public class CDCUtil {
@@ -101,15 +101,11 @@ public class CDCUtil {
 
     public static boolean isCDCIndex(byte[] indexNameBytes) {
         String indexName = Bytes.toString(indexNameBytes);
-        return indexName.startsWith(CDC_INDEX_PREFIX);
+        return isCDCIndex(indexName);
     }
 
     public static byte[] getCdcObjectName(byte[] cdcIndexName) {
-        byte[] result = new byte[cdcIndexName.length - CDC_INDEX_PREFIX.length()];
-        for (int i = 0; i < cdcIndexName.length - CDC_INDEX_PREFIX.length(); i++) {
-            result[i] = cdcIndexName[i + CDC_INDEX_PREFIX.length()];
-        }
-        return result;
+        return Arrays.copyOfRange(cdcIndexName, CDC_INDEX_PREFIX.length(), cdcIndexName.length);
     }
 
     public static boolean isCDCIndex(PTable indexTable) {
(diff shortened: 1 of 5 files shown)


phoenix-core/src/it/java/org/apache/phoenix/end2end/CDCDefinitionIT.java
```diff
@@ -303,6 +306,44 @@ public void testDropCDCIndex () throws SQLException {
...
+        try (ResultSet rs = conn.createStatement().executeQuery("SELECT cdc_include FROM " +
+                "system.catalog WHERE table_name = '" + cdcName +
+                "' AND column_name IS NULL and column_family IS NULL")) {
+            assertEquals(false, rs.next());
+        }
+
+        String drop_cdc_sql = "DROP CDC " + cdcName + " ON " + tableName;
+        try {
+            conn.createStatement().execute(drop_cdc_sql);
+            fail("Expected to fail as cdc table doesn't exist");
```
virajjasani, line 362: Without the PR changes, it fails here i.e. it is able to execute `DROP CDC` successfully?

palashc: It fails before that, in the previous assert 
`//cdc object should have been dropped`

virajjasani: Nice, one more test with schema and table name, or schema + table name with `"` for case-sensitive table names would also be helpful

palashc: done

virajjasani:
+1, pending build results

phoenix-core-client/src/main/java/org/apache/phoenix/util/CDCUtil.java
```diff
@@ -99,6 +99,19 @@ public static boolean isCDCIndex(String indexName) {
         return indexName.startsWith(CDC_INDEX_PREFIX);
     }
 
+    public static boolean isCDCIndex(byte[] indexNameBytes) {
+        String indexName = Bytes.toString(indexNameBytes);
+        return indexName.startsWith(CDC_INDEX_PREFIX);
```
haridsv: You can call the `isCDCIndex(String)` variant here.

phoenix-core-client/src/main/java/org/apache/phoenix/util/CDCUtil.java
```diff
@@ -99,6 +99,19 @@ public static boolean isCDCIndex(String indexName) {
...
+    }
+
+    public static byte[] getCdcObjectName(byte[] cdcIndexName) {
+        byte[] result = new byte[cdcIndexName.length - CDC_INDEX_PREFIX.length()];
+        for (int i = 0; i < cdcIndexName.length - CDC_INDEX_PREFIX.length(); i++) {
+            result[i] = cdcIndexName[i + CDC_INDEX_PREFIX.length()];
+        }
```
haridsv: You can use one of the existing byte copy methods such as `Arrays.copyOfRange`.

phoenix-core-server/src/main/java/org/apache/phoenix/coprocessor/MetaDataEndpointImpl.java
```diff
@@ -3333,6 +3333,11 @@ private MetaDataMutationResult doDropTable(byte[] key, byte[] tenantId, byte[] s
 
         // Recursively delete indexes
         for (byte[] indexName : indexNames) {
+            if (CDCUtil.isCDCIndex(indexName)) {
+                byte[] cdcKey = SchemaUtil.getTableKey(tenantId, schemaName, CDCUtil.getCdcObjectName(indexName));
+                Delete deleteCdc = new Delete(cdcKey, clientTimeStamp);
+                catalogMutations.add(deleteCdc);
+            }
```
haridsv, line 3340: I think in the order of delete, it should be CDC index first and then the CDC object, basically following the same philosophy of deleting the indexes before the table.

phoenix-core/src/it/java/org/apache/phoenix/end2end/CDCDefinitionIT.java
```diff
@@ -145,10 +145,21 @@ public void testCreateCaseSensitiveTable() throws Exception {
         String cdc_sql = "CREATE CDC " + cdcName + " ON " + tableName;
         conn.createStatement().execute(cdc_sql);
         conn.createStatement().executeQuery("SELECT * FROM " + cdcName);
+
+        String drop_sql = forView ? "DROP VIEW " + tableName : "DROP TABLE " + tableName;
```
haridsv, line 149: So, CASCADE option is not required?

virajjasani: CASCADE is not required for Indexes also IIRC. It is needed for view metadata cleanup.

phoenix-core-server/src/main/java/org/apache/phoenix/coprocessor/MetaDataEndpointImpl.java
palashc: @haridsv I followed what is done during DROP CDC where the virtual table is dropped first and then the uncovered index.

phoenix-core-client/src/main/java/org/apache/phoenix/util/CDCUtil.java
palashc: done

phoenix-core-client/src/main/java/org/apache/phoenix/util/CDCUtil.java
palashc: done

phoenix-core-server/src/main/java/org/apache/phoenix/coprocessor/MetaDataEndpointImpl.java
haridsv: The thought process there was that it is OK to go ahead and drop the CDC object (main user intention), even if the index gets left behind. However, in this case, the user intention is to drop the table and if for some reason dropping index fails, it would have left the table in tact, but would have dropped the CDC. 

On the other hand, reordering the drop operations is only a slight improvement, as a failure in dropping the table can already cause a partial cleanup. Besides, since your logic is detecting the CDC via the index, if the index is dropped first and a failure occurs before CDC object is dropped, then a retry will not be able to discover it, so on a second thought, the current order seems better. You may want to leave a quick comment  on why we should drop CDC first.

palashc: > Besides, since your logic is detecting the CDC via the index, if the index is dropped first and a failure occurs before CDC object is dropped, then a retry will not be able to discover it

Very good point, I missed adding the comment, I will make sure to add it with a future PR.

PR #2197, opened 2025-06-18
PHOENIX-7640 Refactor ReplicationLog for HA Groups

apurtell:


phoenix-core-server/src/main/java/org/apache/phoenix/replication/ReplicationLogGroup.java
```diff
@@ -0,0 +1,312 @@
...
+    }
+
+    /**
+     * Switch the writer implementation (e.g., from synchronous to store-and-forward). This method
+     * is thread-safe and ensures proper cleanup of the old writer.
+     *
+     * @param writer The new writer implementation
+     * @throws IOException if the switch fails
+     */
+    protected void switchWriter(ReplicationLogGroupWriter writer) throws IOException {
```
tkhurana: @apurtell The way I was envisioning was that we will always first write to the standby regardless of the mode we are currently in. That will also serve as a connectivity check. So basically we will have 2 writer object instances in the ReplicationLogGroup class.

kadirozde: @tkhurana, I agree that we should attempt to write synchronously when we are in the store-and-forward and so we need to have two writers. However, we do not have to attempt for every mutation, otherwise we will increase write latency significantly when we are in the store-and-forward mode. Instead, for example, we can attempt every n seconds where n can be configurable with the default value 5 seconds.

apurtell: Sure, we can do it that way.

apurtell: Addressed in 5a1b952c8

apurtell:
Rebased after update of `PHOENIX-7562-feature` branch.

phoenix-core-server/src/main/java/org/apache/phoenix/replication/ReplicationLogGroup.java
```diff
@@ -0,0 +1,352 @@
...
+        localWriter = null;
+        LOG.info("Started ReplicationLogGroup for HA Group: {}", haGroupId);
+    }
+
+    /**
+     * Get the HA Group ID managed by this instance.
+     *
+     * @return HA Group ID
+     */
+    public String getHaGroupId() {
```
kadirozde: I think we need to use Name instead of ID. ID implies that an HA group has also a name. The ID also implies that it is generated instead of given. An HA group has a name and the name is given by the user. The HA feature is not responsible for the uniqueness of HA group names among Phoenix clusters.

apurtell: Addressed in 943239154

PR #2198, opened 2025-06-20
PHOENIX-7598 Add ITs for client write behavior when they encounter ATS state

ritegarg:


phoenix-core-client/src/main/java/org/apache/phoenix/exception/MutationBlockedIOException.java
```diff
@@ -0,0 +1,33 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.phoenix.exception;
+
+import java.io.IOException;
+
+
```
shahrs87: Add javadoc

phoenix-core/src/it/java/org/apache/phoenix/end2end/index/IndexRegionObserverMutationBlockingIT.java
```diff
@@ -0,0 +1,292 @@
...
+            
+            // Mutation should work
+            conn.createStatement().execute("UPSERT INTO " + dataTableName + 
+                " VALUES ('1', 'David', 40)");
+            conn.commit();
+            
+            // Transition to ACTIVE_TO_STANDBY (blocking state)
+            crr = new ClusterRoleRecord("transition_test",
+                    HighAvailabilityPolicy.FAILOVER, 
+                    haAdmin.getZkUrl(),
```
shahrs87: can we add a test case where we change haAdmin.getZkUrl() back to Active from ATS state and make sure write to this cluster succeeds? This is required in case we have to rollback the state change.

phoenix-core/src/it/java/org/apache/phoenix/end2end/index/IndexRegionObserverMutationBlockingIT.java
```diff
@@ -0,0 +1,292 @@
...
+                    haAdmin.getZkUrl(), 
+                    ClusterRoleRecord.ClusterRole.STANDBY,
+                    "standby-zk-url", 
+                    ClusterRoleRecord.ClusterRole.ACTIVE, 
+                    3L);
+            haAdmin.createOrUpdateDataOnZookeeper(crr);
+            Thread.sleep(ZK_CURATOR_EVENT_PROPAGATION_TIMEOUT_MS);
+            
+            // Mutations should work again
+            conn.createStatement().execute("UPSERT INTO " + dataTableName +
```
shahrs87: How come this works? Are we sure that conn object talks to cluster with URL "standby-zk-url" which  is currently active? If yes, where is that cluster? We are re-using the connection object that is pointing to first cluster.

shahrs87 (changes requested):
requested some test related changes.

phoenix-core/src/it/java/org/apache/phoenix/end2end/index/IndexRegionObserverMutationBlockingIT.java
```diff
@@ -0,0 +1,292 @@
...
+            }
+        }
+    }
+
+    @Test
+    public void testMutationBlockingTransition() throws Exception {
+        String dataTableName = generateUniqueName();
+        String indexName = generateUniqueName();
+        
+        try (Connection conn = DriverManager.getConnection(getUrl())) {
```
shahrs87, line 187: Why are we creating connection with a specific URL? Shouldn't we use some generic getConnection method that will lookup in CRR which cluster is active and create connection to that cluster?

phoenix-core/src/it/java/org/apache/phoenix/end2end/index/IndexRegionObserverMutationBlockingIT.java
ritegarg: Currently we only block the mutations when current cluster role is ATS. In any other case we don't block. ParallelConnection can still write to a cluster in Standby State

phoenix-core/src/it/java/org/apache/phoenix/end2end/index/IndexRegionObserverMutationBlockingIT.java
ritegarg: I don't think it should matter with current logic of blocking mutations. Only ATS state is blocked, any other state is not blocked.

phoenix-core/src/it/java/org/apache/phoenix/end2end/index/IndexRegionObserverMutationBlockingIT.java
ritegarg: You are assuming this is using Failover Connection and in FailoverConnectionIT, this is simple connection using URL we have.

phoenix-core/src/it/java/org/apache/phoenix/end2end/index/IndexRegionObserverMutationBlockingIT.java
ritegarg: Added tests for [ParallelPhoenixConnectionIT.java](https://github.com/apache/phoenix/pull/2198/files#diff-70800ab3f4d05a909c26f4a905f6830b5161a7b3a7aa69991ac4412d3cdd0ec5)

phoenix-core-server/src/main/java/org/apache/phoenix/hbase/index/IndexRegionObserver.java
```diff
@@ -540,8 +541,9 @@ public void preBatchMutate(ObserverContext<RegionCoprocessorEnvironment> c,
           final Configuration conf = c.getEnvironment().getConfiguration();
           final HAGroupStoreManager haGroupStoreManager = HAGroupStoreManager.getInstance(conf);
           if (haGroupStoreManager.isMutationBlocked()) {
-              throw new IOException("Blocking Mutation as Some CRRs are in ACTIVE_TO_STANDBY "
-                      + "state and CLUSTER_ROLE_BASED_MUTATION_BLOCK_ENABLED is true");
+               throw new MutationBlockedIOException("Blocking Mutation as Some CRRs "
```
shahrs87: should the comment be like: 
`Blocking Mutation as this cluster is in ACTIVE_TO_STANDBY state and CLUSTER_ROLE_BASED_MUTATION_BLOCK_ENABLED is true`

phoenix-core/src/it/java/org/apache/phoenix/jdbc/ParallelPhoenixConnectionIT.java
```diff
@@ -250,6 +257,48 @@ public void testCluster1OfflineRole() throws Exception {
...
 
+    /**
+     * Test Phoenix connection creation and basic operations with HBase both cluster is ACTIVE_TO_STANDBY role.
+     */
+    @Test
+    public void bothClusterATSRole() throws Exception {
+        CLUSTERS.transitClusterRole(haGroup, ClusterRole.ACTIVE_TO_STANDBY, ClusterRole.ACTIVE_TO_STANDBY);
+        try (Connection conn = getParallelConnection()) {
+            doTestBasicOperationsWithConnection(conn, tableName, haGroupName);
+        } catch (SQLException e) {
```
shahrs87, line 377: We need to have assert.fail statement either in try or catch block? I assume since both the clusters are in ATS state, these operations will fail so we need to have assert.fail statement in try block?

shahrs87: Do we have any checks whether both the clusters can be in ATS state? I know we have check where both the clusters cannot be in ACTIVE state.

phoenix-core/src/it/java/org/apache/phoenix/jdbc/ParallelPhoenixConnectionIT.java
```diff
@@ -250,6 +257,48 @@ public void testCluster1OfflineRole() throws Exception {
...
+    /**
+     * Test Phoenix connection creation and basic operations with HBase one cluster is ACTIVE_TO_STANDBY role.
+     */
+    @Test
+    public void oneClusterATSRole() throws Exception {
+        CLUSTERS.transitClusterRole(haGroup, ClusterRole.ACTIVE_TO_STANDBY, ClusterRole.ACTIVE);
+        int countExceptions = 0;
+        try (Connection conn = getParallelConnection()) {
+            doTestBasicOperationsWithConnection(conn, tableName, haGroupName);
+        } catch (SQLException e) {
```
shahrs87, line 408: similar comment as above.

phoenix-core-server/src/main/java/org/apache/phoenix/hbase/index/IndexRegionObserver.java
ritegarg: IMO, this is more specific. There can be multiple CRRs. Current logic is that even if one of the CRR is in ATS state we block the mutation. Currently this means that cluster itself is in CRR state but in future if we can identify traffic based on HAGroup(or CRR) we can selectively block mutations for that HAGroup as well.

shahrs87: > There can be multiple CRRs

For a given cluster on the server side, it will always watch its own local ZK's CRR. Even if this cluster is marked ATS on some other cluster's CRR, how will this cluster get notified when we change the role in other cluster's CRR?

ritegarg: That is not in scope for changes in PHOENIX-7493. This will be handled in Consistent Failover.

phoenix-core-client/src/main/java/org/apache/phoenix/exception/MutationBlockedIOException.java
ritegarg: Done

phoenix-core/src/it/java/org/apache/phoenix/jdbc/ParallelPhoenixConnectionIT.java
ritegarg: This is part of client code which is not under test. Removed these tests for now as per our discussion

phoenix-core/src/it/java/org/apache/phoenix/jdbc/ParallelPhoenixConnectionIT.java
ritegarg: Same as above

phoenix-core-server/src/main/java/org/apache/phoenix/hbase/index/IndexRegionObserver.java
shahrs87: >  This will be handled in Consistent Failover.

Didn't understand this comment?  In the current implementation can a regionserver have a watcher pointing to different clusters CRR? @ritegarg

ritegarg: No, that is not in current implementation. Currently it only watches the current cluster and not peer.

shahrs87 (approved):
+1 ltgm.

phoenix-core-server/src/main/java/org/apache/phoenix/hbase/index/IndexRegionObserver.java
jpisaac: nit: some (lower case) instead on Some

phoenix-core/src/it/java/org/apache/phoenix/end2end/index/IndexRegionObserverMutationBlockingIT.java
jpisaac: @ritegarg Do we have similar tests in Failover and ParallelConnectionIT?

phoenix-core-server/src/main/java/org/apache/phoenix/hbase/index/IndexRegionObserver.java
ritegarg: fixed

jpisaac (approved):

PR #2199, opened 2025-06-21
PHOENIX-7646 New PhoenixStatement API to return old row state in Atomic Updates

virajjasani:
Jira: PHOENIX-7646

PHOENIX-7462, PHOENIX-7398, PHOENIX-7434, PHOENIX-7411 and PHOENIX-7630 Jiras have introduced support for PhoenixStatement to return ResultSet for updated new row version.

The purpose of this PR is to add support for returning the old row state (before update) in atomic update operations. This enhancement allows applications to retrieve the previous state of a row when performing conditional updates, enabling better audit trails.

New API signature:
```
/**
 * old row (before update) as Result object back to the client. This must be used with
 * If the row is successfully updated, return the old row (state before update), otherwise
 * if the row cannot be updated, return non-updated (old) row.
 *
 * @return The pair of int and ResultSet, where int represents value 1 for successful row
 * update and 0 for non-successful row update, and ResultSet represents the old state of the
 * row.
 */
public Pair<Integer, ResultSet> executeAtomicUpdateReturnOldRow(String sql)
        throws SQLException {
```

### DIFF
--- a/phoenix-core-client/src/main/java/org/apache/phoenix/execute/MutationState.java
+++ b/phoenix-core-client/src/main/java/org/apache/phoenix/execute/MutationState.java
@@ -775,12 +775,12 @@ public class MutationState implements SQLCloseable {
                     }
                 }
                 if (this.returnResult != null) {
-                    if (this.returnResult == ReturnResult.ROW) {
+                    if (this.returnResult == ReturnResult.NEW_ROW_ON_SUCCESS) {
                         for (Mutation mutation : rowMutations) {
                             mutation.setAttribute(PhoenixIndexBuilderHelper.RETURN_RESULT,
                                     PhoenixIndexBuilderHelper.RETURN_RESULT_ROW);
                         }
-                    } else if (this.returnResult == ReturnResult.OLD_ROW) {
+                    } else if (this.returnResult == ReturnResult.OLD_ROW_ALWAYS) {
                         for (Mutation mutation : rowMutations) {
                             mutation.setAttribute(PhoenixIndexBuilderHelper.RETURN_RESULT,
                                     PhoenixIndexBuilderHelper.RETURN_RESULT_OLD_ROW);
@@ -807,10 +807,10 @@ public class MutationState implements SQLCloseable {
                         mutation.setAttribute(PhoenixIndexBuilderHelper.ATOMIC_OP_ATTRIB, onDupKeyBytes);
                     }
                     if (this.returnResult != null) {
-                        if (this.returnResult == ReturnResult.ROW) {
+                        if (this.returnResult == ReturnResult.NEW_ROW_ON_SUCCESS) {
                             mutation.setAttribute(PhoenixIndexBuilderHelper.RETURN_RESULT,
                                     PhoenixIndexBuilderHelper.RETURN_RESULT_ROW);
-                        } else if (this.returnResult == ReturnResult.OLD_ROW) {
+                        } else if (this.returnResult == ReturnResult.OLD_ROW_ALWAYS) {
                             mutation.setAttribute(PhoenixIndexBuilderHelper.RETURN_RESULT,
                                     PhoenixIndexBuilderHelper.RETURN_RESULT_OLD_ROW);
                         }
@@ -1527,8 +1527,8 @@ public class MutationState implements SQLCloseable {
                                         .decodeInt(cell.getValueArray(), cell.getValueOffset(),
                                                 SortOrder.getDefault());
                                 if (this.returnResult != null) {
-                                    if (this.returnResult == ReturnResult.ROW ||
-                                            this.returnResult == ReturnResult.OLD_ROW) {
+                                    if (this.returnResult == ReturnResult.NEW_ROW_ON_SUCCESS ||
+                                            this.returnResult == ReturnResult.OLD_ROW_ALWAYS) {
                                         this.result = result;
                                     }
                                 }
@@ -2402,8 +2402,8 @@ public class MutationState implements SQLCloseable {
     }
 
     public enum ReturnResult {
-        ROW,
-        OLD_ROW
+        NEW_ROW_ON_SUCCESS,
+        OLD_ROW_ALWAYS
     }
 
 }
--- a/phoenix-core-client/src/main/java/org/apache/phoenix/execute/MutationState.java
+++ b/phoenix-core-client/src/main/java/org/apache/phoenix/execute/MutationState.java
@@ -780,6 +780,11 @@ public class MutationState implements SQLCloseable {
                             mutation.setAttribute(PhoenixIndexBuilderHelper.RETURN_RESULT,
                                     PhoenixIndexBuilderHelper.RETURN_RESULT_ROW);
                         }
+                    } else if (this.returnResult == ReturnResult.OLD_ROW) {
+                        for (Mutation mutation : rowMutations) {
+                            mutation.setAttribute(PhoenixIndexBuilderHelper.RETURN_RESULT,
+                                    PhoenixIndexBuilderHelper.RETURN_RESULT_OLD_ROW);
+                        }
                     }
                 }
                 // The DeleteCompiler already generates the deletes for indexes, so no need to do it again
@@ -805,6 +810,9 @@ public class MutationState implements SQLCloseable {
                         if (this.returnResult == ReturnResult.ROW) {
                             mutation.setAttribute(PhoenixIndexBuilderHelper.RETURN_RESULT,
                                     PhoenixIndexBuilderHelper.RETURN_RESULT_ROW);
+                        } else if (this.returnResult == ReturnResult.OLD_ROW) {
+                            mutation.setAttribute(PhoenixIndexBuilderHelper.RETURN_RESULT,
+                                    PhoenixIndexBuilderHelper.RETURN_RESULT_OLD_ROW);
                         }
                     }
                 }
@@ -1519,7 +1527,8 @@ public class MutationState implements SQLCloseable {
                                         .decodeInt(cell.getValueArray(), cell.getValueOffset(),
                                                 SortOrder.getDefault());
                                 if (this.returnResult != null) {
-                                    if (this.returnResult == ReturnResult.ROW) {
+                                    if (this.returnResult == ReturnResult.ROW ||
+                                            this.returnResult == ReturnResult.OLD_ROW) {
                                         this.result = result;
                                     }
                                 }
@@ -2393,7 +2402,8 @@ public class MutationState implements SQLCloseable {
     }
 
     public enum ReturnResult {
-        ROW
+        ROW,
+        OLD_ROW
     }
 
 }
--- a/phoenix-core-client/src/main/java/org/apache/phoenix/jdbc/PhoenixPreparedStatement.java
+++ b/phoenix-core-client/src/main/java/org/apache/phoenix/jdbc/PhoenixPreparedStatement.java
@@ -219,8 +219,8 @@ public class PhoenixPreparedStatement extends PhoenixStatement implements Phoeni
      * Executes the given SQL statement similar to JDBC API executeUpdate() but also returns the
      * old row (before update) as Result object back to the client. This must be used with
      * auto-commit Connection. This makes the operation atomic.
-     * If the row is successfully updated, return the old row (state before update), otherwise if
-     * the row cannot be updated, return non-updated (old) row.
+     * Return the old row (state before update) regardless of whether the update is
+     * successful or not.
      *
      * @return The pair of int and ResultSet, where int represents value 1 for successful row update
      * and 0 for non-successful row update, and ResultSet represents the old state of the row.
--- a/phoenix-core-client/src/main/java/org/apache/phoenix/jdbc/PhoenixPreparedStatement.java
+++ b/phoenix-core-client/src/main/java/org/apache/phoenix/jdbc/PhoenixPreparedStatement.java
@@ -226,6 +226,7 @@ public class PhoenixPreparedStatement extends PhoenixStatement implements Phoeni
      * and 0 for non-successful row update, and ResultSet represents the old state of the row.
      * @throws SQLException If the statement cannot be executed.
      */
+    // Note: Do Not remove this, it is expected to be used by downstream applications
     public Pair<Integer, ResultSet> executeAtomicUpdateReturnOldRow() throws SQLException {
         if (!connection.getAutoCommit()) {
             throw new SQLExceptionInfo.Builder(SQLExceptionCode.AUTO_COMMIT_NOT_ENABLED).build()
@@ -247,6 +248,7 @@ public class PhoenixPreparedStatement extends PhoenixStatement implements Phoeni
      * and 0 for non-successful row update, and ResultSet represents the state of the row.
      * @throws SQLException If the statement cannot be executed.
      */
+    // Note: Do Not remove this, it is expected to be used by downstream applications
     public Pair<Integer, ResultSet> executeAtomicUpdateReturnRow() throws SQLException {
         if (!connection.getAutoCommit()) {
             throw new SQLExceptionInfo.Builder(SQLExceptionCode.AUTO_COMMIT_NOT_ENABLED).build()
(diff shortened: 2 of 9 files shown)


phoenix-core-client/src/main/java/org/apache/phoenix/jdbc/PhoenixPreparedStatement.java
```diff
@@ -215,6 +215,28 @@ private void preExecuteUpdate() throws SQLException {
         }
     }
 
+    /**
+     * Executes the given SQL statement similar to JDBC API executeUpdate() but also returns the
+     * old row (before update) as Result object back to the client. This must be used with
+     * auto-commit Connection. This makes the operation atomic.
+     * If the row is successfully updated, return the old row (state before update), otherwise if
+     * the row cannot be updated, return non-updated (old) row.
```
haridsv: Do you mean "Return the old row (state before update) whether the update is successful or not"?

phoenix-core-server/src/main/java/org/apache/phoenix/hbase/index/IndexRegionObserver.java
```diff
@@ -1656,7 +1665,13 @@ public void postBatchMutateIndispensably(ObserverContext<RegionCoprocessorEnviro
...
                       cells.add(cell);
 
-                      addCellsIfResultReturned(miniBatchOp, context, cells);
+                      if (!context.returnOldRow) {
+                          addCellsIfResultReturned(miniBatchOp, context.returnResult, cells,
+                                  context.currColumnCellExprMap, false);
+                      } else {
+                          addCellsIfResultReturned(miniBatchOp, context.returnResult, cells,
+                                  context.oldRowColumnCellExprMap, true);
+                      }
```
haridsv, line 1674: Why not return both old and new row state and provide a single interface?

phoenix-core-client/src/main/java/org/apache/phoenix/jdbc/PhoenixPreparedStatement.java
virajjasani: Sure, let me update

haridsv:
My assumption was that `miniBatchOp.setOperationStatus()` can be called for index `1` to return a new `Result` even if it is for the same row. Are you saying this won't work because they will have the same row key?

Regarding data transfer overhead, I wasn't suggesting that we should return both versions unconditionally. We could take in a param for the caller to indicate what row versions they are interested in.

virajjasani:
> My assumption was that `miniBatchOp.setOperationStatus()` can be called for index `1` to return a new `Result` even if it is for the same row. Are you saying this won't work because they will have the same row key?

The index must match the num of operations in the miniBatch, therefore for the single row update, we cannot use index other than 0 IIRC.

> Regarding data transfer overhead, I wasn't suggesting that we should return both versions unconditionally. We could take in a param for the caller to indicate what row versions they are interested in.

Got it, that's what this PR is doing, but we cannot use any index other than 0 for the single row atomic update.

virajjasani:
I just skimmed through the logic to cross verify. HBase throws ArrayIndexOutOfBoundsException if we use `setOperationStatus()` with index > 0 for single row.

```
  private int getAbsoluteIndex(int index) {
    if (index < 0 || this.firstIndex + index >= this.lastIndexExclusive) {
      throw new ArrayIndexOutOfBoundsException(index);
    }
    return this.firstIndex + index;
  }
```

phoenix-core-client/src/main/java/org/apache/phoenix/jdbc/PhoenixStatement.java
```diff
@@ -2501,6 +2503,33 @@ public Pair<Integer, ResultSet> executeAtomicUpdateReturnRow(String sql) throws
...
+     * successful or not.
+     *
+     * @param sql The SQL DML statement, UPSERT or DELETE for Phoenix.
+     * @return The pair of int and ResultSet, where int represents value 1 for successful row
+     * update and 0 for non-successful row update, and ResultSet represents the old state of the
+     * row.
+     * @throws SQLException If the statement cannot be executed.
+     */
+    // Note: Do Not remove this, it is expected to be used by downstream applications
+    public Pair<Integer, ResultSet> executeAtomicUpdateReturnOldRow(String sql)
```
haridsv, line 2521: The only difference between this and the existing `executeAtomicUpdateReturnRow` is the parameter `ReturnResult` correct? Why not refactor to avoid the duplication?

virajjasani: Yes, I was thinking about it, but I thought this would be more clear.
Using Phoenix internal Enum for the end user API might be bit odd? WDYT?

phoenix-core-client/src/main/java/org/apache/phoenix/jdbc/PhoenixPreparedStatement.java
```diff
@@ -215,6 +215,28 @@ private void preExecuteUpdate() throws SQLException {
...
+     * auto-commit Connection. This makes the operation atomic.
+     * Return the old row (state before update) regardless of whether the update is
+     * successful or not.
+     *
+     * @return The pair of int and ResultSet, where int represents value 1 for successful row update
+     * and 0 for non-successful row update, and ResultSet represents the old state of the row.
+     * @throws SQLException If the statement cannot be executed.
+     */
+    // Note: Do Not remove this, it is expected to be used by downstream applications
+    public Pair<Integer, ResultSet> executeAtomicUpdateReturnOldRow() throws SQLException {
```
haridsv, line 229: Same question about the feasibility to refactor. In fact, why not expose the ReturnResult as a param and have a single public method? We can leave the current one as is and expose a new overloaded method that takes ReturnResult. The current one can call the new one with `ReturnResult.ROW` as the default.

phoenix-core-client/src/main/java/org/apache/phoenix/execute/MutationState.java
```diff
@@ -2393,7 +2402,8 @@ public boolean isEmpty() {
     }
 
     public enum ReturnResult {
-        ROW
+        ROW,
+        OLD_ROW
```
haridsv: May be these should be renamed like this?
```suggestion
        NEW_ROW_ON_SUCCESS,
        OLD_ROW_ALWAYS
```

phoenix-core-client/src/main/java/org/apache/phoenix/jdbc/PhoenixPreparedStatement.java
virajjasani: Given that this can also have two methods anyways, and the possibilities of returning the results that we want to support are only either new row version or old row version, shall we keep the current approach? It would be cleaner than exposing internal Enum to JDBC clients?

virajjasani: In current state, the client application only has to convert Statement to PhoenixStatement but it does not have to import extra Enum in the application.

phoenix-core-client/src/main/java/org/apache/phoenix/jdbc/PhoenixStatement.java
haridsv: This is anyway an internal API (i.e., non-JDBC API) that we are exposing, so it is just a matter of committing the enum to be public and stable, so I don't see that as an issue.

phoenix-core-client/src/main/java/org/apache/phoenix/jdbc/PhoenixPreparedStatement.java
haridsv: Exposing the enum itself doesn't add any additional burden or limitation correct? It comes from the same classpath and jar file.

virajjasani: I meant to say, the application has to unnecessarily import `org.apache.phoenix.execute.MutationState.ReturnResult` and besides old and new row are the only use case we are planning to support.

phoenix-core-client/src/main/java/org/apache/phoenix/jdbc/PhoenixStatement.java
virajjasani: That's right, however, the only part that Phoenix clients usually take thick client dependency (for imports) of is `PhoenixConnection` or `PhoenixStatement` or `PhoenixPreparedStatement`. Everything else comes from standard JDBC classes only, starting from `DriverManager`.

phoenix-core-client/src/main/java/org/apache/phoenix/execute/MutationState.java
virajjasani: Sounds good, let me update them.

phoenix-core-client/src/main/java/org/apache/phoenix/jdbc/PhoenixPreparedStatement.java
haridsv: An import on PhoenixStatement or PhoenixPreparedStatement is already needed right? Is there a difference between a single import vs two imports? Since you are confident of having only two types, an alternative is to use a boolean param which can then be translated to the enum inside the method.

phoenix-core-client/src/main/java/org/apache/phoenix/jdbc/PhoenixStatement.java
haridsv: True, but see my other [comment](https://github.com/apache/phoenix/pull/2199#discussion_r2163102267) which is applicable here too.

phoenix-core-client/src/main/java/org/apache/phoenix/jdbc/PhoenixPreparedStatement.java
virajjasani: For the import, I only meant that it is usually expected for Phoenix downstreamers to import PhoenixConnection or PhoenixStatement or PhoenixPreparedStatement. However, anything else e.g. Tuple, MutationState etc is not expected to be imported or used in anyway.

> An import on PhoenixStatement or PhoenixPreparedStatement is already needed right? Is there a difference between a single import vs two imports? Since you are confident of having only two types, an alternative is to use a boolean param which can then be translated to the enum inside the method.

That's also nice idea, but my intention was to try to follow the method signature of executeUpdate(). As part of Kadir's suggestion, the method signature was changed to `executeAtomicUpdateReturnRow()` from `executeUpdateReturnRow()` with PHOENIX-7462. The intention was to keep as close to `executeUpdate()` as possible :)

phoenix-core-client/src/main/java/org/apache/phoenix/execute/MutationState.java
virajjasani: Let me make this change tomorrow.

phoenix-core-client/src/main/java/org/apache/phoenix/jdbc/PhoenixPreparedStatement.java
haridsv: Once we digress from JDBC API I don't see any value in matching the signature of this method to existing JDBC method. Users will not be able to discover these APIs without going through custom docs written specifically for Phoenix anyway so we are talking about documenting 2 methods vs one with extra parameter, and IMHO one method is simpler but may be it is just me.

phoenix-core-server/src/main/java/org/apache/phoenix/hbase/index/IndexRegionObserver.java
```diff
@@ -1316,6 +1320,11 @@ private void identifyMutationTypes(MiniBatchOperationInProgress<Mutation> miniBa
             Mutation m = miniBatchOp.getOperation(i);
             if (this.builder.returnResult(m) && miniBatchOp.size() == 1) {
                 context.returnResult = true;
+                byte[] returnResult = m.getAttribute(PhoenixIndexBuilderHelper.RETURN_RESULT);
+                if (returnResult != null && Arrays.equals(returnResult,
+                        PhoenixIndexBuilderHelper.RETURN_RESULT_OLD_ROW)) {
```
haridsv, line 1325: Considering there is existing build.returnResult(), why not add one more returnOldResult() and use it?

I also noticed an existing usage at line 798 that I think can be replaced with builder.returnResult() like this:

```
--- phoenix-core-server/src/main/java/org/apache/phoenix/hbase/index/IndexRegionObserver.java
+++ phoenix-core-server/src/main/java/org/apache/phoenix/hbase/index/IndexRegionObserver.java
@@ -1864,8 +1864,7 @@ public class IndexRegionObserver implements RegionCoprocessor, RegionObserver {
             throws IOException {
       List<Mutation> mutations = Lists.newArrayListWithExpectedSize(2);
       byte[] opBytes = atomicPut.getAttribute(ATOMIC_OP_ATTRIB);
-        byte[] returnResult = atomicPut.getAttribute(RETURN_RESULT);
-        if ((opBytes == null && returnResult == null) ||
+        if ((opBytes == null && this.builder.returnResult(atomicPut)) ||
                 (opBytes == null && miniBatchOp.size() != 1)) {
             // Unexpected
             // Either mutation should be atomic by providing non-null ON DUPLICATE KEY, or
```

phoenix-core-server/src/main/java/org/apache/phoenix/hbase/index/IndexRegionObserver.java
```diff
@@ -1892,6 +1911,13 @@ private List<Mutation> generateOnDupMutations(BatchMutateContext context,
       Pair<Put, Put> dataRowState = context.dataRowStates.get(rowKeyPtr);
       Put currentDataRowState = dataRowState != null ? dataRowState.getFirst() : null;
```
haridsv, line 1912: The above 4 lines seem to have incorrect indent level?

phoenix-core-client/src/main/java/org/apache/phoenix/jdbc/PhoenixPreparedStatement.java
virajjasani: I think the API contract is very clean with this. Yes boolean based single method is also not bad, but the method name would not be self explanatory, one will have to look up the meaning of boolean. It Enum was exposed, probably would have been better than boolean.

Given that we only need new row and old row contracts, this is not expected to change and the method names are quite self explanatory. One has to make intentionally bad decision in using oldRow() for new row behaviour or newRow() for old row behaviour. Whereas boolean based method can be copied from one place to another and one might miss checking the meaning of boolean.

I don't mind making single method, it's not difficult either, but with two methods instead of one, I am finding it much cleaner.

virajjasani:
```
 -          } else if (context.returnResult) {
 +          } else if (context.returnResult && !context.returnOldRow) {
                Map<ColumnReference, Pair<Cell, Boolean>> currColumnCellExprMap = new HashMap<>();
                byte[] rowKey = m.getRow();
                ImmutableBytesPtr rowKeyPtr = new ImmutableBytesPtr(rowKey);
```
This means DELETE cannot return old row even if someone uses oldRow return API.
Anyways, these too are debatable mechanical changes that we can do but I am bit hesitant only because the logic becomes more complicated to understand for anyone. When someone looks at separate map, it becomes way easier to track it's limited usage and find out how the server is returning old row vs new row, rather than debugging through 5 extra conditional checks.

phoenix-core-server/src/main/java/org/apache/phoenix/hbase/index/IndexRegionObserver.java
virajjasani: Yes, would be better to address in separate PR i guess, but yes this is annoying

haridsv:
In fact, there is one less if condition now because there is only one variable. The change is based on two facts, 1) there is only one version that you are returning, so why have 2 variables, especially that only one of them is ever going to be non-null, 2) chronologically, you come across the old version first so take a note of it and use it unless a new version is available. To me this feels way more cleaner even without any verbs to wrap the if conditions (there are simple enough with just one or two conditions only), but I will leave the decision up to you.

phoenix-core-client/src/main/java/org/apache/phoenix/jdbc/PhoenixPreparedStatement.java
haridsv: Sure, I guess no big deal.

haridsv: You could still consider deduping the code though.

virajjasani:
> > This means DELETE cannot return old row even if someone uses oldRow return API.
> > Is there a test case to catch this? All the existing tests passed with my change.

I can add a test but given a couple more expected upcoming features, i was thinking of including test with it.

PR #2202, opened 2025-06-24
PHOENIX-7647 : BSON_UPDATE_EXPRESSION() set value if field key does not exist

palashc:


virajjasani (approved):
+1, pending build results

PR #2203, opened 2025-06-25
PHOENIX-7648 Introduce new Atomic Operation - ON DUPLICATE KEY UPDATE_ONLY

virajjasani:
Jira: PHOENIX-7648

Phoenix provides two types of Atomic Update operations: `ON DUPLICATE KEY IGNORE` and `ON DUPLICATE KEY UPDATE`.

The purpose of `ON DUPLICATE KEY IGNORE` is to perform the insertion of new row i.e. if the row does not exist, add the row, else ignore the update operation. The purpose of `ON DUPLICATE KEY UPDATE` is to provide atomic update based on the expressions provided on the various non-pk columns. For the conditional updates that include `CASE` expression, the atomic update goes through successfully only if the condition is satisfied on the row.

However, if the row does not already exist, `ON DUPLICATE KEY UPDATE` performs the insertion anyways regardless of the condition provided. There are valid use cases that consider the absence of the row as condition not being satisfied i.e. if the row does not exist, the condition provided with `CASE` statement is not satisfied anyways.

Therefore, the purpose of this PR is to introduce new type of atomic update operation with statement `ON DUPLICATE KEY UPDATE_ONLY`, that performs the condition evaluation similar to `ON DUPLICATE KEY UPDATE`, however if the row does not exist, `UPDATE_ONLY` is not expected to perform any insertion.

### DIFF
--- a/phoenix-core-client/src/main/java/org/apache/phoenix/execute/MutationState.java
+++ b/phoenix-core-client/src/main/java/org/apache/phoenix/execute/MutationState.java
@@ -91,6 +91,7 @@ import org.apache.phoenix.monitoring.MutationMetricQueue.MutationMetric;
 import org.apache.phoenix.monitoring.MutationMetricQueue.NoOpMutationMetricsQueue;
 import org.apache.phoenix.monitoring.ReadMetricQueue;
 import org.apache.phoenix.monitoring.TableMetricsManager;
+import org.apache.phoenix.parse.UpsertStatement.OnDuplicateKeyType;
 import org.apache.phoenix.query.QueryConstants;
 import org.apache.phoenix.query.QueryServices;
 import org.apache.phoenix.query.QueryServicesOptions;
@@ -2306,24 +2307,26 @@ public class MutationState implements SQLCloseable {
         @Nonnull
         private final RowTimestampColInfo rowTsColInfo;
         private byte[] onDupKeyBytes;
-        private boolean isUpdateOnly;
+        private OnDuplicateKeyType onDupKeyType;
         private long colValuesSize;
 
         public RowMutationState(@Nonnull Map<PColumn, byte[]> columnValues, long colValuesSize, int statementIndex,
                 @Nonnull RowTimestampColInfo rowTsColInfo, byte[] onDupKeyBytes) {
-            this(columnValues, colValuesSize, statementIndex, rowTsColInfo, onDupKeyBytes, false);
+            this(columnValues, colValuesSize, statementIndex, rowTsColInfo, onDupKeyBytes,
+                    OnDuplicateKeyType.NONE);
         }
 
         public RowMutationState(@Nonnull Map<PColumn, byte[]> columnValues, long colValuesSize,
                                 int statementIndex, @Nonnull RowTimestampColInfo rowTsColInfo,
-                                byte[] onDupKeyBytes, boolean isUpdateOnly) {
+                                byte[] onDupKeyBytes,
+                                OnDuplicateKeyType onDupKeyType) {
             checkNotNull(columnValues);
             checkNotNull(rowTsColInfo);
             this.columnValues = columnValues;
             this.statementIndexes = new int[] { statementIndex };
             this.rowTsColInfo = rowTsColInfo;
             this.onDupKeyBytes = onDupKeyBytes;
-            this.isUpdateOnly = isUpdateOnly;
+            this.onDupKeyType = onDupKeyType;
             this.colValuesSize = colValuesSize;
         }
 
@@ -2337,7 +2340,7 @@ public class MutationState implements SQLCloseable {
         }
 
         boolean isUpdateOnly() {
-            return isUpdateOnly;
+            return onDupKeyType == OnDuplicateKeyType.UPDATE_ONLY;
         }
 
         public Map<PColumn, byte[]> getColumnValues() {
@@ -2375,7 +2378,9 @@ public class MutationState implements SQLCloseable {
             // Concatenate ON DUPLICATE KEY bytes to allow multiple
             // increments of the same row in the same commit batch.
             this.onDupKeyBytes = PhoenixIndexBuilderHelper.combineOnDupKey(this.onDupKeyBytes, newRow.onDupKeyBytes);
-            this.isUpdateOnly = this.isUpdateOnly || newRow.isUpdateOnly;
+            if (newRow.onDupKeyType == OnDuplicateKeyType.UPDATE_ONLY) {
+                this.onDupKeyType = OnDuplicateKeyType.UPDATE_ONLY;
+            }
             statementIndexes = joinSortedIntArrays(statementIndexes, newRow.getStatementIndexes());
             return true;
         }
--- a/phoenix-core-client/src/main/java/org/apache/phoenix/compile/UpsertCompiler.java
+++ b/phoenix-core-client/src/main/java/org/apache/phoenix/compile/UpsertCompiler.java
@@ -62,6 +62,7 @@ import org.apache.phoenix.parse.ParseNode;
 import org.apache.phoenix.parse.SelectStatement;
 import org.apache.phoenix.parse.SequenceValueParseNode;
 import org.apache.phoenix.parse.UpsertStatement;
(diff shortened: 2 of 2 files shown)


palashc (approved):

phoenix-core-client/src/main/java/org/apache/phoenix/execute/MutationState.java
```diff
@@ -804,7 +804,13 @@ private void generateMutations(final TableRef tableRef, final long mutationTimes
                 // TODO: use our ServerCache
                 for (Mutation mutation : rowMutations) {
                     if (onDupKeyBytes != null) {
-                        mutation.setAttribute(PhoenixIndexBuilderHelper.ATOMIC_OP_ATTRIB, onDupKeyBytes);
+                        mutation.setAttribute(PhoenixIndexBuilderHelper.ATOMIC_OP_ATTRIB,
+                                onDupKeyBytes);
+                        if (rowEntry.getValue().isUpdateOnly()) {
+                            mutation.setAttribute(
```
tkhurana, line 811: Instead of creating a new mutation attribute, one other way could have been to extend the onDupKeyBytes and send this extra information there.

phoenix-core-server/src/main/java/org/apache/phoenix/index/PhoenixIndexBuilder.java
```diff
@@ -152,6 +152,21 @@ public List<Mutation> executeAtomicOp(Increment inc) throws IOException {
                         : Collections.<Mutation>emptyList();
             }
         }
+
```
tkhurana: Is this change needed ? Does the client ever create an Increment mutation ?

phoenix-core-client/src/main/java/org/apache/phoenix/execute/MutationState.java
virajjasani: I thought this would clearly distinguish new client with the ability to use UPDATE_ONLY vs old clients.

phoenix-core-server/src/main/java/org/apache/phoenix/index/PhoenixIndexBuilder.java
virajjasani: Yeah you are right, it's not needed, we don't have test coverage also for this. I just thought of changing this because i saw reference to `PhoenixIndexBuilderHelper.isDupKeyIgnore(opBytes)` above. Shall we keep this? Eventually, as separate Jira, someone can cleanup this whole hook?

phoenix-core-client/src/main/java/org/apache/phoenix/execute/MutationState.java
```diff
@@ -2300,16 +2306,24 @@ public static class RowMutationState {
...
         private byte[] onDupKeyBytes;
+        private boolean isUpdateOnly;
         private long colValuesSize;
 
         public RowMutationState(@Nonnull Map<PColumn, byte[]> columnValues, long colValuesSize, int statementIndex,
                 @Nonnull RowTimestampColInfo rowTsColInfo, byte[] onDupKeyBytes) {
+            this(columnValues, colValuesSize, statementIndex, rowTsColInfo, onDupKeyBytes, false);
+        }
+
+        public RowMutationState(@Nonnull Map<PColumn, byte[]> columnValues, long colValuesSize,
```
tkhurana, line 2319: @virajjasani This is in reference to my earlier comment. `isUpdateOnly` is relevant only in the context of onDupKeyBytes but that is not obvious looking at this implementation. If it is handled within onDupKeyBytes then the RowMutationState class doesn't need to change.

phoenix-core-server/src/main/java/org/apache/phoenix/index/PhoenixIndexBuilder.java
tkhurana: My suggestion will be to not make any change in this code.

virajjasani: Okk let me revert this then

phoenix-core-client/src/main/java/org/apache/phoenix/execute/MutationState.java
virajjasani: `onDupKeyBytes` remains same for both `UPDATE` and `UPDATE_ONLY` types. Only `IGNORE` does not have it. This is why we needed to introduce enum and let the grammar pass it to the client.

tkhurana: Grammar change is fine. I am not talking about that. Just like `IGNORE` vs `UPDATE` is differentiated by the contents of opDupKeyBytes. The same pattern should be followed for `UPDATE_ONLY`.

virajjasani: I see, but that might cause backward compatibility issue right? New server expecting certain bytes pattern might still get them from old client depending on `onDupKeyBytes` derived by the old client?

tkhurana: We can always add new fields when serializing the opDupKey object. It is similar to how we add optional fields in protobuf. We did something similar when serializing skipScanFilter on the client side when we added an additional field https://github.com/apache/phoenix/blob/master/phoenix-core-client/src/main/java/org/apache/phoenix/filter/SkipScanFilter.java#L655-L661

virajjasani: Compared to ON DUPLICATE KEY cases, SkipScanFilter seems quite straightforward.
For instance, here we also combine multiple onDupKeyBytes:
```
            // Concatenate ON DUPLICATE KEY bytes to allow multiple
            // increments of the same row in the same commit batch.
            this.onDupKeyBytes = PhoenixIndexBuilderHelper.combineOnDupKey(this.onDupKeyBytes, newRow.onDupKeyBytes);
```

Appending additional byte at the end would need more changes, and probably create more chances of failure.

Whereas current approach has just one additional attribute with single byte value, it is handled in a way none of the onDupKeyBytes logic needs to change at either client or server side. I think this is far safer approach.

virajjasani: Throwing and catching Exception while deserializing is also bit of an compute overhead, which will not be present with current approach.

virajjasani: Also, I think already having `break` while catching EOFException here can make our case even more complicated

```
          try {
              int nExpressions = WritableUtils.readVInt(input);
              List<Expression> expressions = Lists.newArrayListWithExpectedSize(nExpressions);
              for (int i = 0; i < nExpressions; i++) {
                  Expression expression = ExpressionType.values()[WritableUtils.readVInt(input)].newInstance();
                  expression.readFields(input);
                  expressions.add(expression);
                  expression.accept(visitor);
              }
              PTableProtos.PTable tableProto = PTableProtos.PTable.parseDelimitedFrom(input);
              PTable table = PTableImpl.createFromProto(tableProto);
              operations.add(new Pair<>(table, expressions));
          } catch (EOFException e) {
              break;
          }
```

tkhurana: Ideally, given a serialized onDupKeyBytes array you should be able to deserialize  it and then determine the type of duplicate key operation that needs to be performed without resorting to other mechanisms like mutation attributes. This is what is happening today but it won't after your change.
But I understand your concern around added complexity in adding additional field to the onDupKeyBytes array.  So what I would suggest you is instead of passing the boolean isUpdateOnly, you pass the enum `OnDuplicateKeyType` around and store that in the RowMutationState object. This way the code will be more readable and it will be clearer that the duplicate type is not part of the serialized array.

virajjasani: Done, the build results are clean, known flaky tests only.

PR #2206, opened 2025-06-28
PHOENIX-7626: Add metrics to capture HTable thread pool utilization and contention (addendum: fix incompatibility for HBase 2.4 build)

sanjeet006py:


virajjasani (approved):
+1

PR #2211, opened 2025-07-03
PHOENIX-7584 Conditional TTL on SYSTEM.CDC_STREAM

palashc:
Jira: PHOENIX-7584


phoenix-core/src/it/java/org/apache/phoenix/end2end/CDCStreamTTLIT.java
```diff
@@ -0,0 +1,72 @@
...
+        createTableAndEnableCDC(conn, tableName, true);
+        TestUtil.splitTable(conn, tableName, Bytes.toBytes("m"));
+        String sql = "SELECT PARTITION_END_TIME FROM SYSTEM.CDC_STREAM WHERE TABLE_NAME='" + tableName + "'";
+        ResultSet rs = conn.createStatement().executeQuery(sql);
+        int count = 0;
+        while (rs.next()) {
+            count++;
+        }
+        Assert.assertEquals(3, count);
+        Thread.sleep(11000);
```
virajjasani: Instead of sleep, using `ManualEnvironmentEdge` would be much better approach.
Even if we don't configure `PHOENIX_CDC_STREAM_PARTITION_EXPIRY_MIN_AGE_MS` for the test, we can still increment the clock manually by more than 30 hr and without sleep, we can verify that the parent closed partition is no longer visible.

virajjasani:
`testSettingTTLForViewsOnTableWithTTL()` failures also likely relevant

phoenix-core/src/it/java/org/apache/phoenix/end2end/CDCStreamTTLIT.java
palashc: @virajjasani done

palashc:
@virajjasani @jpisaac I have fixed the test failures.

phoenix-core-client/src/main/java/org/apache/phoenix/schema/MetaDataClient.java
```diff
@@ -1236,7 +1237,7 @@ private byte[][] processSplits(Map<String, Object> tableProperties, byte[][] spl
      * @throws SQLException
      */
     private void populatePropertyMaps(ListMultimap<String,Pair<String,Object>> statementProps, Map<String, Object> tableProps,
-            Map<String, Object> commonFamilyProps, PTableType tableType, boolean isCDCIndex) throws SQLException {
+            Map<String, Object> commonFamilyProps, PTableType tableType, ViewType viewType, String tableName, boolean isCDCIndex) throws SQLException {
```
virajjasani: viewType is not being used

phoenix-core/src/it/java/org/apache/phoenix/end2end/CDCStreamIT.java
```diff
@@ -208,6 +209,37 @@ public void testStreamMetadataWhenTableIsDropped() throws SQLException {
...
+        int newCount = 0;
+        while (rs.next()) {
+            // parent partition row with non-zero end time should have expired
+            if (rs.getLong(1) > 0) {
+                Assert.fail("Closed partition should have expired after TTL.");
+            }
+            newCount++;
+        }
+        Assert.assertEquals(2, newCount);
+        EnvironmentEdgeManager.reset();
```
virajjasani: Let's wrap this in try-finally:

```
        try {
            ManualEnvironmentEdge injectEdge = new ManualEnvironmentEdge();
            long t = System.currentTimeMillis() +
                    QueryServicesOptions.DEFAULT_PHOENIX_CDC_STREAM_PARTITION_EXPIRY_MIN_AGE_MS;
            t = (t / 1000) * 1000;
            EnvironmentEdgeManager.injectEdge(injectEdge);
            injectEdge.setValue(t);
            rs = conn.createStatement().executeQuery(sql);
            int newCount = 0;
            while (rs.next()) {
                // parent partition row with non-zero end time should have expired
                if (rs.getLong(1) > 0) {
                    Assert.fail("Closed partition should have expired after TTL.");
                }
                newCount++;
            }
            Assert.assertEquals(2, newCount);
        } finally {
            EnvironmentEdgeManager.reset();
        }
```

virajjasani:
@palashc the changes look good overall, left few minor comments above.

virajjasani (approved):
+1, pending build results

PR #2215, opened 2025-07-08
PHOENIX-7658 CDC event for TTL_DELETE to exclude pre-image if PRE scope is not selected

virajjasani:
Jira: PHOENIX-7658

As a follow-up to PHOENIX-7653 which introduced TTL_DELETE as new CDC event type, we should not return pre-built pre-image with CDC JSON if the SELECT query does not have PRE scope.

### DIFF
--- a/phoenix-core-server/src/main/java/org/apache/phoenix/coprocessor/CDCCompactionUtil.java
+++ b/phoenix-core-server/src/main/java/org/apache/phoenix/coprocessor/CDCCompactionUtil.java
@@ -47,7 +47,6 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import java.io.IOException;
-import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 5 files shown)


phoenix-core-server/src/main/java/org/apache/phoenix/coprocessor/CDCCompactionUtil.java
```diff
@@ -130,7 +129,6 @@ private static Map<String, Object> createTTLDeleteCDCEvent(Put expiredRowPut, PT
             }
         }
         cdcEvent.put(QueryConstants.CDC_PRE_IMAGE, preImage);
-        cdcEvent.put(QueryConstants.CDC_POST_IMAGE, Collections.emptyMap());
```
virajjasani, line 133: We don't keep post-image if the event is DELETE, therefore same needs to apply to TTL_DELETE also

haridsv (commented):
Looks fine to me. I guess you went with removal later while reading because you can't check if it is in scope while building?

virajjasani:
Yes exactly @haridsv!

PR #2222, opened 2025-07-14
PHOENIX-7198 support for multi row constructors in single upsert query

richardantal:


virajjasani:
This could be a nice feature!
I am curious about the use case: Is this required for auto-commit connections from sqlline?

tkhurana:
@richardantal Is there performance gain from using multiple values in a single query ? I don't think so since we can achieve the same by batching multiple upsert statements and then doing a commit.

phoenix-core/src/it/java/org/apache/phoenix/end2end/MultipleUpsertIT.java
```diff
@@ -0,0 +1,93 @@
...
+        String ddl = "CREATE TABLE " + tableName + "(K VARCHAR NOT NULL PRIMARY KEY, INT INTEGER, INT2 INTEGER)";
+        conn.createStatement().execute(ddl);
+
+
+        conn.createStatement().execute("UPSERT INTO " + tableName + " VALUES ('A', 11, 12)");
+        conn.createStatement().execute("UPSERT INTO " + tableName + "(K, INT) VALUES ('B', 2)");
+        conn.createStatement().execute("UPSERT INTO " + tableName + "(K, INT, INT2) VALUES ('E', 5, 5),('F', 61, 6)");
+        conn.createStatement().execute("UPSERT INTO " + tableName + " VALUES ('C', 31, 32),('D', 41, 42)");
+        conn.createStatement().execute("UPSERT INTO " + tableName + " VALUES ('G', 7, 72),('H', 8)");
+        conn.createStatement().execute("UPSERT INTO " + tableName + " VALUES ('I', 9),('I', 10)");
```
stoty: I don't think this is valid standard SQL (i.e. not all columns are specified, but no column list either) , but if this works for the single value upsert then it's fine.

phoenix-core/src/test/java/org/apache/phoenix/parse/QueryParserTest.java
```diff
@@ -803,6 +803,26 @@ public void testValidUpsertSelectHint() throws Exception {
...
+    }
+
+    @Test
+    public void testValidMultipleUpsert2() throws Exception {
+        String sql = "upsert into t(a,b) VALUES(1,2),(3,4)";
+        parseQuery(sql);
+    }
+
+    @Test
+    public void testValidMultipleUpsert3() throws Exception {
```
stoty: A few more failure cases to exercise the parser would be nice.
i.e. no value list at all, no commas between, not closing some parentheses and similar.

phoenix-core/src/it/java/org/apache/phoenix/end2end/MultipleUpsertIT.java
```diff
@@ -0,0 +1,93 @@
...
+    @Test
+    public void testUpsertMultiple() throws Exception {
+        Properties props = new Properties();
+        Connection conn = DriverManager.getConnection(getUrl(), props);
+        String tableName = generateUniqueName();
+        String ddl = "CREATE TABLE " + tableName + "(K VARCHAR NOT NULL PRIMARY KEY, INT INTEGER, INT2 INTEGER)";
+        conn.createStatement().execute(ddl);
+
+
+        conn.createStatement().execute("UPSERT INTO " + tableName + " VALUES ('A', 11, 12)");
```
stoty: I'd like to also have some tests for expressions and mixed expressions and literals.
i.e.
VALUES (substring('APPLE',0,1), 2*2)

phoenix-core-client/src/main/java/org/apache/phoenix/compile/UpsertCompiler.java
```diff
@@ -1286,75 +1302,83 @@ public MutationState execute() throws SQLException {
...
-                if (constantExpression.getDataType() != null) {
-                    value = constantExpression.getDataType().toObject(ptr, constantExpression.getSortOrder(),
-                            constantExpression.getMaxLength(), constantExpression.getScale());
-                    if (!constantExpression.getDataType().isCoercibleTo(column.getDataType(), value)) {
-                        throw TypeMismatchException.newException(
-                            constantExpression.getDataType(), column.getDataType(), "expression: "
-                                    + constantExpression.toString() + " in column " + column);
+
+            int index = -1;
+            for (byte[][] valuesListIem : valuesList) {
```
stoty: Keeping the two lists in sync like this awkward.
Using an old-style for() would be more readable to me:
for(int index=0; index<valuesList.length();i++)
and then use the index to get both the valuesListItem and constantExpression.

stoty: Also a typo in valuesListIem

stoty:
> @richardantal Is there performance gain from using multiple values in a single query ? I don't think so since we can achieve the same by batching multiple upsert statements and then doing a commit.

My take:
- Avoids network RTT for Query Server
- Avoids multiple query parse and compilation phases

also:
- Adds an previously unimplemented SQL feature.
- Improves compatibility with existing SQL tools that expect thisfeature (though in reality, most of those tools don't handle upserts anyway)

richardantal:
Thanks for the reviews, I added a bit more tests

I also uploaded a simple performance test for this feature to the ticket as an attachment.
For me, inserting simple rows took 65 sec, inserting 2 at once 59, 4: 57, 8: 56 (With the same number of commits)
So there is a minor performance improvement using this feature but it is not the major point of this feature.

stoty:
10% improvement for bulk loading is nothing to sneeze at @richardantal .

phoenix-core-client/src/main/antlr3/PhoenixSQL.g
```diff
@@ -864,9 +864,10 @@ finally{ contextStack.pop(); }
 
 // Parse a full upsert expression structure.
 upsert_node returns [UpsertStatement ret]
+@init{List<List<ParseNode>> v = new ArrayList<List<ParseNode>>(); }
     :   UPSERT (hint=hintClause)? INTO t=from_table_name
         (LPAREN p=upsert_column_refs RPAREN)?
-        ((VALUES LPAREN v=one_or_more_expressions RPAREN ( ON DUPLICATE KEY ( ig=IGNORE |
+        ((VALUES LPAREN e = one_or_more_expressions {v.add(e);} RPAREN (COMMA LPAREN e = one_or_more_expressions {v.add(e);} RPAREN )* ( ON DUPLICATE KEY ( ig=IGNORE |
```
stoty: nit:
Would extracting `LPAREN e = one_or_more_expressions {v.add(e);} RPAREN` imto a function make sense ?

stoty (approved):
+1 LGTM

richardantal:
I rebased the change to resolve conflict with spotless change.
Lets wait for a test run

phoenix-core-client/src/main/java/org/apache/phoenix/compile/UpsertCompiler.java
```diff
@@ -571,7 +574,7 @@ public MutationPlan compile(UpsertStatement upsert) throws SQLException {
       }
     }
     boolean isAutoCommit = connection.getAutoCommit();
-    if (valueNodes == null) {
+    if (valueNodesList.isEmpty()) {
```
virajjasani, line 579: Should this not also consider `valueNodesList` being null? like we consider above with the for-loop?

phoenix-core-client/src/main/java/org/apache/phoenix/compile/UpsertCompiler.java
```diff
@@ -557,8 +558,12 @@ public MutationPlan compile(UpsertStatement upsert) throws SQLException {
         columnsBeingSet.set(columnIndexesToBe[i] = rowTimestampCol.getPosition());
         pkSlotIndexesToBe[i] = table.getRowTimestampColPos();
         targetColumns.add(rowTimestampCol);
-        if (valueNodes != null && !valueNodes.isEmpty()) {
-          valueNodes.add(getNodeForRowTimestampColumn(rowTimestampCol));
+        if (valueNodesList != null) {
```
richardantal, line 561: Probably this check is not needed, but better to be safe

richardantal:
@virajjasani @kadirozde Can I proceed to merging this?

virajjasani:
Sounds good. Based on @stoty's review, feel free to merge.

Just one thing to confirm: could you run ConcurrentMutationsExtendedIT a couple of times to ensure no test is failing? In the jenkins build, concurrent mutations test is failing for the last two builds. Just want to make sure it is flaky and locally it is passing at least two times.

Thank you @richardantal!

PR #2223, opened 2025-07-15
PHOENIX-7666 Index query failure with SkipScanFilter

tkhurana:


phoenix-core-server/src/main/java/org/apache/phoenix/index/GlobalIndexChecker.java
```diff
@@ -452,6 +452,7 @@ private void repairIndexRows(byte[] indexRowKey, long ts, List<Cell> row) throws
                 // The row is "unverified". Rewind the scanner and let the row be scanned again
                 // so that it can be repaired
                 scanner.close();
+                indexScan.withStartRow(CellUtil.cloneRow(row.get(0)), true);
```
kadirozde: The comments starting on line 452 should read as follows
We got a different index row. This means either the unverified index row has been deleted by the single row rebuild as part of the read repair process, or the filters on the scan object skip the rebuilt index row.  If this new row is "verified" (or empty) then we are good to go.

PR #2224, opened 2025-07-15
PHOENIX-7566 [1/n] HAGroupStoreClient/Manager changes for Consistent HA

ritegarg:
**Summary of Changes:-**

1. Created HAGroupStoreRecord to hold cluster role information for single cluster. This will be used for consistent HA instead of existing ClusterRoleRecord.
2. Created 2 vertical stacks of classes (HAGroupStoreManagerImpl -> HAGroupStoreClient) and (HAGroupStoreManagerV1Impl -> HAGroupStoreClientV1) to support HAGroupStoreRecord and ClusterRoleRecord respectively.
Note that previously the CRR handling classes were named as HAGroupStoreManager and HAGroupStoreClient but now I have renamed them to V1 as eventually we want to deprecate the V1 and keep the ones without V1
3.New Features in HAGroupStoreManagerImpl -> HAGroupStoreClient
i. Created singleton instance of HAGroupStoreClient for each HAGroupName
ii. Added watcher for PeerZK. If Peer ZK url is present in HAGroupStoreClient, it will watch that in addition to local ZK
iii. Added support for checkAndPut for records in ZK for serialization of concurrent updates to HAGroupStoreRecord
iv. Added new methods to be used by ReplicationLogWriter
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;a. getHAGroupStoreRecord(haGroupName)
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;b. setStoreAndForwardMode(haGroupName)
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;c. setSyncMode(haGroupName)
v. Added ITs for new classes and new methods introduced.

**Changes in ClusterRoleRecord**
i. Added new states needed for consistent HA
ii. Added transition allowList for selectively allowing transitions between roles
iii. Adding transition wait time before transitioning between specific roles(cache entry has to be older than wait time for update to succeed)

**Next PRs**
Adding support for getting HAGroupStoreRecord from System table for recovery.
Adding more ITs for parallel connection with one cluster supporting CRR and other supporting HAGroupStoreRecord
Add Parallel and Failover tests with new HAGroupStoreRecord

### DIFF
--- a/phoenix-core-client/src/main/java/org/apache/phoenix/jdbc/ClusterRoleRecord.java
+++ b/phoenix-core-client/src/main/java/org/apache/phoenix/jdbc/ClusterRoleRecord.java
@@ -23,7 +23,10 @@ import com.fasterxml.jackson.annotation.JsonIgnore;
 import com.fasterxml.jackson.annotation.JsonProperty;
 import org.apache.commons.lang3.builder.EqualsBuilder;
 import org.apache.commons.lang3.builder.HashCodeBuilder;
+import org.apache.phoenix.exception.InvalidClusterRoleTransitionException;
 import org.apache.phoenix.thirdparty.com.google.common.base.Preconditions;
+
+import org.apache.phoenix.thirdparty.com.google.common.collect.ImmutableSet;
 import org.apache.phoenix.util.JDBCUtil;
 import org.apache.phoenix.util.JacksonUtil;
 import org.slf4j.Logger;
@@ -33,6 +36,13 @@ import java.io.IOException;
 import java.nio.charset.StandardCharsets;
 import java.util.Arrays;
 import java.util.Optional;
+import java.util.Set;
+import org.apache.hadoop.conf.Configuration;
+
+import static org.apache.phoenix.query.QueryServices.HA_STORE_AND_FORWARD_MODE_REFRESH_INTERVAL_MS;
+import static org.apache.phoenix.query.QueryServices.HA_SYNC_MODE_REFRESH_INTERVAL_MS;
+import static org.apache.phoenix.query.QueryServicesOptions.DEFAULT_HA_STORE_AND_FORWARD_MODE_REFRESH_INTERVAL_MS;
+import static org.apache.phoenix.query.QueryServicesOptions.DEFAULT_HA_SYNC_MODE_REFRESH_INTERVAL_MS;
 
 /**
  * Immutable class of a cluster role record for a pair of HBase clusters.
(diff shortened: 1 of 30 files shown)


phoenix-core-client/src/main/java/org/apache/phoenix/jdbc/ClusterRoleRecord.java
```diff
@@ -55,18 +65,136 @@
...
+                throws InvalidClusterRoleTransitionException {
+            if (!allowedTransitions.contains(targetRole)) {
+                throw new InvalidClusterRoleTransitionException("Cannot transition from " + this + " to " + targetRole);
+            }
+
+            // Read wait times from configuration based on the transition
+            switch (this) {
+                case ACTIVE_NOT_IN_SYNC:
+                    if (targetRole == ACTIVE) {
+                        return conf.getLong(HA_SYNC_MODE_REFRESH_INTERVAL_MS, DEFAULT_HA_SYNC_MODE_REFRESH_INTERVAL_MS);
```
jpisaac, line 156: Does it need to check the conf object everytime?

phoenix-core-server/src/main/java/org/apache/phoenix/coprocessor/PhoenixRegionServerEndpoint.java
```diff
@@ -119,10 +121,15 @@ public void invalidateHAGroupStoreClient(RpcController controller,
             RegionServerEndpointProtos.InvalidateHAGroupStoreClientRequest request,
             RpcCallback<RegionServerEndpointProtos.InvalidateHAGroupStoreClientResponse> done) {
         LOGGER.info("PhoenixRegionServerEndpoint invalidating HAGroupStoreClient");
-        HAGroupStoreManager haGroupStoreManager;
         try {
-            haGroupStoreManager = HAGroupStoreManager.getInstance(conf);
-            haGroupStoreManager.invalidateHAGroupStoreClient();
+            Optional<HAGroupStoreManager> haGroupStoreManagerOptional
+                    = HAGroupStoreManagerFactory.getInstance(conf);
```
jpisaac, line 126: Can we try and reduce the conf lookups? The necessary parameters can be looked up once and stored during start

phoenix-core-server/src/main/java/org/apache/phoenix/hbase/index/IndexRegionObserver.java
```diff
@@ -621,10 +623,16 @@ public void preBatchMutate(ObserverContext<RegionCoprocessorEnvironment> c,
...
-              throw new IOException("Blocking Mutation as Some CRRs are in ACTIVE_TO_STANDBY "
-                      + "state and CLUSTER_ROLE_BASED_MUTATION_BLOCK_ENABLED is true");
+          final Optional<HAGroupStoreManager> haGroupStoreManagerOptional
+                  = HAGroupStoreManagerFactory.getInstance(conf);
+          if (!haGroupStoreManagerOptional.isPresent()) {
+              throw new IOException("HAGroupStoreManager is null "
+                      + "for current cluster, check configuration");
+          }
+          if (haGroupStoreManagerOptional.get().isMutationBlocked(conf)) {
+              throw new MutationBlockedIOException("Blocking Mutation as Some CRRs are in "
```
jpisaac, line 633: nit: lowercase "some"

phoenix-core-client/src/main/java/org/apache/phoenix/jdbc/HighAvailabilityGroup.java
```diff
@@ -435,9 +448,28 @@ static Optional<String> getFallbackCluster(String url, Properties properties) th
...
+     * @param jdbcUrl    the ZK endpoint host:port or the JDBC connection String host:port:/hbase
+     * @param properties the properties defining time out values and retry count
+     * @param namespace  the ZooKeeper namespace to use, defaults to PHOENIX_HA_ZOOKEEPER_ZNODE_NAMESPACE if null
+     * @return a new Curator framework client
+     */
+    @SuppressWarnings("UnstableApiUsage")
+    public static CuratorFramework getCurator(String jdbcUrl, Properties properties, String namespace)
+            throws IOException {
+        // Use namespace as part of cache key to avoid conflicts between different namespaces
+        String effectiveNamespace = namespace != null ? namespace : PHOENIX_HA_ZOOKEEPER_ZNODE_NAMESPACE;
```
jpisaac, line 468: Why the need for null check on namespace, can we require and expect namespace to be passed in if not throw an exception?

phoenix-core-client/src/main/java/org/apache/phoenix/jdbc/HighAvailabilityGroup.java
```diff
@@ -498,7 +535,7 @@ private static CuratorFramework createCurator(String jdbcUrl, Properties propert
         CuratorFramework curator = CuratorFrameworkFactory
                 .builder()
                 .connectString(zkUrl)
-                .namespace(PHOENIX_HA_ZOOKEEPER_ZNODE_NAMESPACE)
+                .namespace(namespace != null ? namespace : PHOENIX_HA_ZOOKEEPER_ZNODE_NAMESPACE)
```
jpisaac, line 538: Same thoughts here, namespace argument should not be null

phoenix-core-client/src/main/java/org/apache/phoenix/jdbc/HAGroupStoreRecord.java
```diff
@@ -0,0 +1,192 @@
...
+    }
+
+    public static byte[] toJson(HAGroupStoreRecord record) throws IOException {
+        return JacksonUtil.getObjectWriter().writeValueAsBytes(record);
+    }
+
+    /**
+     * @return true if this record is newer than the given record.
+     */
+    public boolean isNewerThan(HAGroupStoreRecord other) {
```
jpisaac, line 95: Would it make sense to implement the Comparable interface?

jpisaac (commented):
Took a first pass at it, Left some general comments. Will take a another look later.

phoenix-core-client/src/main/java/org/apache/phoenix/jdbc/HAGroupStoreManager.java
```diff
@@ -18,63 +18,78 @@
...
+public interface HAGroupStoreManager {
 
     /**
-     * Creates/gets an instance of HAGroupStoreManager.
+     * Checks whether mutation is blocked or not across all HA groups.
+     * If any HAGroupStoreClient instance is not created, it will be created.
+     * If any HAGroup mutation is blocked, it will return true.
+     * @throws IOException when HAGroupStoreClient is not healthy.
+     */
+    boolean isMutationBlocked(Configuration conf) throws IOException;
```
jpisaac, line 39: "conf" can be a implementation property/attribute passed in via the constructor. The api would be cleaner without needing the conf parameter everytime. Ideally the conf is inspected once in the constructor and be done with it.

phoenix-core-server/src/main/java/org/apache/phoenix/hbase/index/IndexRegionObserver.java
ritegarg: Fixed

phoenix-core-client/src/main/java/org/apache/phoenix/jdbc/ClusterRoleRecord.java
ritegarg: Extracted required values in constructor for HAGroupStoreClient and now only checking if transition is valid or not.

PR #2225, opened 2025-07-15
PHOENIX-7667 Strict vs Relaxed TTL

virajjasani:
Jira: PHOENIX-7667

Types of TTLs supported by Phoenix:

1. Literal TTL: A simple numeric value specifying the TTL in seconds.
e.g.
`CREATE TABLE T1 (id VARCHAR PRIMARY KEY, COL1 VARCHAR, COL2 INTEGER) TTL = 86400`

Literal TTL values:

- NONE / TTL_EXPRESSION_NOT_DEFINED: TTL is not specified for the table.
- FOREVER / TTL_EXPRESSION_FOREVER: TTL is set to not expire for the rows.
- TTL_EXPRESSION_DEFINED_IN_TABLE_DESCRIPTOR: Clients older than Phoenix 5.3 sets TTL value to the HBase TableDescriptor.
- User provided TTL: Literal value of TTL in seconds.

2. Conditional TTL: A boolean expression to determine the row expiration based on column values.
e.g.
`CREATE TABLE T1 (id VARCHAR PRIMARY KEY, status VARCHAR) TTL = 'status = ''EXPIRED'' OR TO_NUMBER(CURRENT_TIME()) - TO_NUMBER(PHOENIX_ROW_TIMESTAMP()) >= 108000000'`

As of Phoenix 5.3.0 (client and server), both types of TTLs are stored in SYSTEM.CATALOG table.

The default behavior of conditional TTL includes parsing and compilation of the TTL expression, serializing the compiled expression into bytes and sending the serialized bytes as the Scan attribute. The scan attribute for Conditional TTL is deserialized and used by many region coprocessors and scanner implementations including TTLRegionScanner, GlobalIndexRegionScanner and IndexRegionObserver.

In order to provide strict TTL view for the users, the region observers perform extra computation on the row to determine whether the row has already expired and therefore should not be processed. For instance, IndexRegionObserver performs read for the given upsert to identify whether the row has already expired. While the extra cost incurred by the region observers help provide strict TTL expiration, some use case might not require strict TTL expiry.

Let’s define the types of TTL use cases:

- Strict TTL expiration: As soon as the TTL expires for the given row, the row must not be visible to the user queries.
- Relaxed TTL expiration: After the TTL expiration, it can take several hours to days for the row to not be visible to the user queries.

Costs involved to achieve strict Conditional TTL expiry:

- TTLRegionScanner achieves masking of the row (Literal and Conditional TTL)
- IndexRegionObserver performs read for each update operation (No additional cost when the table has covered index)
- GlobalIndexRegionScanner evaluates TTL expression for each data table row during rebuild

For users that do not care about the immediate row masking after the TTL expiry, we can provide optional configuration to avoid the extra cost associated with making the strict TTL expiration. The relaxed TTL expiration is expected to rely only on the Major compaction. After the major compaction successfully expires or deletes the given row, the client will no longer be able to read the expired row.

### DIFF
--- a/phoenix-core-client/src/main/java/org/apache/phoenix/query/QueryServices.java
+++ b/phoenix-core-client/src/main/java/org/apache/phoenix/query/QueryServices.java
@@ -383,6 +383,8 @@ public interface QueryServices extends SQLCloseable {
     public static final String WAL_EDIT_CODEC_ATTRIB = "hbase.regionserver.wal.codec";
     //Property to know whether TTL at View Level is enabled
     public static final String PHOENIX_VIEW_TTL_ENABLED = "phoenix.view.ttl.enabled";
+    // TTL strict mode - when false, TTL is considered relaxed
+    public static final String PHOENIX_TTL_STRICT = "phoenix.ttl.strict";
 
     public static final String PHOENIX_VIEW_TTL_TENANT_VIEWS_PER_SCAN_LIMIT = "phoenix.view.ttl.tenant_views_per_scan.limit";
     // Block mutations based on cluster role record
--- a/phoenix-core-server/src/main/java/org/apache/phoenix/hbase/index/IndexRegionObserver.java
+++ b/phoenix-core-server/src/main/java/org/apache/phoenix/hbase/index/IndexRegionObserver.java
@@ -147,6 +147,8 @@ import static org.apache.phoenix.coprocessorclient.BaseScannerRegionObserverCons
 import static org.apache.phoenix.hbase.index.util.IndexManagementUtil.rethrowIndexingException;
 import static org.apache.phoenix.index.PhoenixIndexBuilderHelper.ATOMIC_OP_ATTRIB;
 import static org.apache.phoenix.index.PhoenixIndexBuilderHelper.RETURN_RESULT;
+import static org.apache.phoenix.query.QueryServices.PHOENIX_TTL_STRICT;
+import static org.apache.phoenix.query.QueryServicesOptions.DEFAULT_PHOENIX_TTL_STRICT;
 import static org.apache.phoenix.util.ByteUtil.EMPTY_BYTE_ARRAY;
 
 /**
@@ -400,6 +402,7 @@ public class IndexRegionObserver implements RegionCoprocessor, RegionObserver {
   private static final int DEFAULT_ROWLOCK_WAIT_DURATION = 30000;
   private static final int DEFAULT_CONCURRENT_MUTATION_WAIT_DURATION_IN_MS = 100;
   private byte[] encodedRegionName;
+  private boolean isTTLStrict;
 
   @Override
   public Optional<RegionObserver> getRegionObserver() {
@@ -410,6 +413,8 @@ public class IndexRegionObserver implements RegionCoprocessor, RegionObserver {
   public void start(CoprocessorEnvironment e) throws IOException {
       try {
         final RegionCoprocessorEnvironment env = (RegionCoprocessorEnvironment) e;
+        this.isTTLStrict = env.getConfiguration().getBoolean(PHOENIX_TTL_STRICT,
+                DEFAULT_PHOENIX_TTL_STRICT);
         encodedRegionName = env.getRegion().getRegionInfo().getEncodedNameAsBytes();
         String serverName = env.getServerName().getServerName();
         if (env.getConfiguration().getBoolean(CHECK_VERSION_CONF_KEY, true)) {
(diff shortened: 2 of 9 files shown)


phoenix-core-client/src/main/java/org/apache/phoenix/query/QueryServices.java
```diff
@@ -383,6 +383,8 @@ public interface QueryServices extends SQLCloseable {
     public static final String WAL_EDIT_CODEC_ATTRIB = "hbase.regionserver.wal.codec";
     //Property to know whether TTL at View Level is enabled
     public static final String PHOENIX_VIEW_TTL_ENABLED = "phoenix.view.ttl.enabled";
+    // TTL strict mode - when false, TTL is considered relaxed
+    public static final String PHOENIX_TTL_STRICT = "phoenix.ttl.strict";
```
tkhurana: Don't we need to support strict/relaxed semantics on views ? IMO this a gap which will be difficult to support later because then we could run into backward compatibility issues.

virajjasani: Yes, it should be supported on views too by default with this PR, unless i am missing something. The primary goal of relaxed TTL is to ensure we don't mask the rows while reading. It seems there is no view specific logic in TTLRegionScanner. View specific TTL logic is available only in CompactionScanner, which anyways is not expected to change regardless of strict or relaxed TTL.

FYI @jpisaac if i missed something with view TTL masking logic. Do we have anything specific?

tkhurana: @virajjasani But this setting is at the table level. Shouldn't it be supported at the view level just like different views have different TTL

virajjasani: Yes storing it as syscat/PTable property is not bad idea, but it can be done outside of this Jira. To keep it simpler, i believe we should scope this as server side change i.e. configure at cluster level or table level and allow customers to use it using CREATE TABLE or ALTER TABLE DDLs.

I can create follow-up jira to store this attribute at syscat level.

virajjasani: Added it as PTable property, please review

phoenix-core-client/src/main/java/org/apache/phoenix/schema/MetaDataClient.java
```diff
@@ -3669,6 +3681,12 @@ public boolean isViewReferenced() {
                 tableUpsert.setBytes(37, rowKeyMatcher);
             }
 
+            if (isStrictTTLProp == null) {
+                tableUpsert.setNull(38, Types.BOOLEAN);
```
sanjeet006py: How about setting this to `PTable.DEFAULT_IS_STRICT_TTL`? This way we can consistent in treating NULL for `isStrictTTLProp`.

phoenix-core-client/src/main/java/org/apache/phoenix/schema/PTableImpl.java
```diff
@@ -2039,6 +2051,10 @@ public static PTable createFromProto(PTableProtos.PTable table) {
         if (table.hasChangeDetectionEnabled()) {
             isChangeDetectionEnabled = table.getChangeDetectionEnabled();
         }
+        boolean isStrictTTL = true;
```
sanjeet006py: For consistency maybe good to use the `DEFAULT_IS_STRICT_TTL`?

phoenix-core-server/src/main/java/org/apache/phoenix/hbase/index/IndexRegionObserver.java
```diff
@@ -1500,7 +1528,7 @@ && isPartialUncoveredIndexMutation(indexMetaData, miniBatchOp))) {
             onDupCheckTime += (EnvironmentEdgeManager.currentTimeMillis() - start);
         }
 
-        if (context.hasConditionalTTL) {
+        if (context.hasConditionalTTL && isStrictTTLEnabled(miniBatchOp)) {
```
sanjeet006py: Do we need this check as I see in the method call also we check if strictTTL is enabled upfront?

phoenix-core-client/src/main/java/org/apache/phoenix/schema/MetaDataClient.java
virajjasani: Good point, i think we can directly use `tableUpsert.setBoolean(38, isStrictTTL)` because isStrictTTL already contains default value?

sanjeet006py (approved):
LGTM

phoenix-core-client/src/main/java/org/apache/phoenix/util/ScanUtil.java
```diff
@@ -1788,6 +1816,10 @@ public static void annotateMutationWithConditionalTTL(
         byte[] ttl = ttlExpr.serialize();
         for (Mutation mutation : mutations) {
             mutation.setAttribute(BaseScannerRegionObserverConstants.TTL, ttl);
+            if (!table.isStrictTTL()) {
+                mutation.setAttribute(BaseScannerRegionObserverConstants.IS_STRICT_TTL,
+                        PBoolean.INSTANCE.toBytes(table.isStrictTTL()));
+            }
```
virajjasani, line 1822: This is improvement: we don't need to send the mutation and scan attribute `_IS_STRICT_TTL` if the TTL is already strict.

PR #2226, opened 2025-07-16
PHOENIX-7651 Support RETURNING * with UPSERT and DELETE

haridsv:
This addresses the requirements in [PHOENIX-7651](https://issues.apache.org/jira/browse/PHOENIX-7651)


phoenix-core-client/src/main/java/org/apache/phoenix/jdbc/PhoenixStatement.java
```diff
@@ -825,14 +826,14 @@ private ExecutableSelectStatement(TableNode from, HintNode hint, boolean isDisti
...
             super(from, hint, isDistinct, select, where, groupBy, having, orderBy, limit, offset, bindCount, isAggregate, hasSequence, selects, udfParseNodes);
         }
-        
+
         private ExecutableSelectStatement(ExecutableSelectStatement select) {
             this(select.getFrom(), select.getHint(), select.isDistinct(), select.getSelect(), select.getWhere(),
                     select.getGroupBy(), select.getHaving(), select.getOrderBy(), select.getLimit(), select.getOffset(), select.getBindCount(),
                     select.isAggregate(), select.hasSequence(), select.getSelects(), select.getUdfParseNodes());
         }
-
```
virajjasani: Too many formatting changes, can we avoid it? :)

haridsv: I would like to keep them as they will fix a lot of Yetus warnings. One general argument is that such changes should be done in a dedicated commit, but that never happens. You can however ignore whitespace changes:

image

Here is the link for your convenience: https://github.com/apache/phoenix/pull/2226/files?diff=unified&w=1

Honestly, I don't know how to revert just the whitespace changes, they were done by the editor settings to strip trailing whitespace while saving. Let me know if you still feel strongly about it, I will try to figure out how to do it.

phoenix-core/src/it/java/org/apache/phoenix/end2end/OnDuplicateKey2IT.java
```diff
@@ -591,8 +598,8 @@ private static void validateReturnedRowBeforeUpsert(Connection conn, String upse
       updateCount = resultPair.getFirst();
       resultSet = resultPair.getSecond();
     }
-    boolean isOnDuplicateKey = upsertSql.toUpperCase().contains("ON DUPLICATE KEY");
-    if (conn.getAutoCommit() && isOnDuplicateKey) {
+    boolean isReturningRow = upsertSql.toUpperCase().contains("RETURNING *");
```
virajjasani: We also need to assert values for old row API, so how about?
```
    boolean isReturningRow = upsertSql.toUpperCase().contains("RETURNING *") ||
            upsertSql.toUpperCase().contains("ON DUPLICATE KEY");
```

phoenix-core-client/src/main/java/org/apache/phoenix/util/TupleUtil.java
```diff
@@ -226,14 +226,16 @@ public static int write(Tuple result, DataOutput out) throws IOException {
...
+   * @param conn         Phoenix Connection object.
+   * @param withPrefetch When {@code true}, the returned ResultSet is prefetched, otherwise one
+   *                     needs to call next() on it.
    * @return ResultSet for the give single row.
    * @throws SQLException If any SQL operation fails.
    */
-  public static ResultSet getResultSet(Tuple toProject, TableName tableName, Connection conn)
-    throws SQLException {
+  public static ResultSet getResultSet(Tuple toProject, TableName tableName, Connection conn,
+    boolean withPrefetch) throws SQLException {
```
virajjasani, line 244: I wonder if we need `withPrefetch` argument. The purpose of the utility is to always return the prefetched result

virajjasani (approved):
Few minor comments, +1 otherwise

phoenix-core-client/src/main/java/org/apache/phoenix/util/TupleUtil.java
haridsv: > The purpose of the utility is to always return the prefetched result

Not from the point of JDBC standard. JDBC always expects the client to invoke next() before the row is made available. The prefetch here is kind of a misnomer, it is not really a fetch from the server but just indicating the iterator semantics.

phoenix-core/src/it/java/org/apache/phoenix/end2end/OnDuplicateKey2IT.java
haridsv: Right now `ON DUPLICATE KEY` alone is expected to return a row unless `RETURNING *` is specified, in fact that is the very reason we introduced this new syntax, to avoid performance overhead when the user is not really interested to fetch the row.

virajjasani: That is correct. However, the problem here is that some tests using this method don't have `RETURNING *` but we are testing `executeAtomicUpdateReturnOldRow()` which will anyways return row.

virajjasani: I think we should remove this check entirely and if the connection is auto-commit, always assert resultSet values because `executeAtomicUpdateReturnOldRow()` will always return row

haridsv: > That is correct. However, the problem here is that some tests using this method don't have RETURNING * but we are testing executeAtomicUpdateReturnOldRow() which will anyways return row.

This is to facilitate the following variations:
- Statement without `RETURNING *` executed via standard JDBC. In this case no, there will be no result set.
- Statement with `RETURNING *` executed via standard JDBC. This results in a result set that will confirm to the JDBC standard, i.e., client would need to call `next()` before accessing the (first and only) row.
- Statement with or without `RETURNING *` executed via one of executeAtomicUpdateReturn* methods. This results in a result set that is prefetched, i.e., `next()` has already been called. This maintains backward compatibility with the existing behavior.

The reason for the check to support the 1st case above when the test is executing statements that don't return a RS and these are existing test cases.

virajjasani: I understand but now I wonder how we can make sure that we assert resultSet values for executeAtomicUpdateReturnOldRow()?

haridsv: There are some dedicated tests that still use executeAtomicUpdateReturnOldRow(), are they not enough?

virajjasani: The point is these asserts are for executeAtomicUpdateReturnOldRow() only and that will always return some old row, so the logic to ensure whether the statement has `RETURNING *` is almost not useful right? If we were not asserting the results of executeAtomicUpdateReturnOldRow(), then the logic would hage surely made sense. WDYT?

haridsv: I see what you are saying, ya that makes sense. I think I just blindly replaced the check for ON DUPLICATE KEY with that of RETURNING *, but even the preexising check was unnecessary. I have now removed this check, please take another look.

haridsv:
No new changes can be found after running `mvn spotless:apply`.

haridsv:
Interesting... I run `spotless:apply` and see no new changes, but I can go through the errors pointed out by the PR validation build and fix them. I am not sure why local run is not finding them.

haridsv:
The below waring is in the existing code, not a new change in this PR:

> org.apache.phoenix.util.TupleUtil.getResultSet(Tuple, TableName, Connection, boolean) may fail to clean up java.sql.Statement on checked exception
> --
>   | Bug type OBL_UNSATISFIED_OBLIGATION_EXCEPTION_EDGE (click for details)In class org.apache.phoenix.util.TupleUtilIn method org.apache.phoenix.util.TupleUtil.getResultSet(Tuple, TableName, Connection, boolean)Reference type java.sql.Statement1 instances of obligation remainingObligation to clean up resource created at TupleUtil.java:[line 243] is not dischargedRemaining obligations: {Statement x 1}
> 
> <br class="Apple-interchange-newline">org.apache.phoenix.util.TupleUtil.getResultSet(Tuple, TableName, Connection, boolean) may fail to clean up java.sql.Statement on checked exception
> [Bug type OBL_UNSATISFIED_OBLIGATION_EXCEPTION_EDGE (click for details)](https://ci-hadoop.apache.org/job/Phoenix/job/Phoenix-PreCommit-GitHub-PR/job/PR-2226/10/artifact/yetus-general-check/output/new-spotbugs-phoenix-core-client.html#OBL_UNSATISFIED_OBLIGATION_EXCEPTION_EDGE)
> In class org.apache.phoenix.util.TupleUtil
> In method org.apache.phoenix.util.TupleUtil.getResultSet(Tuple, TableName, Connection, boolean)
> Reference type java.sql.Statement
> 1 instances of obligation remaining
> Obligation to clean up resource created at TupleUtil.java:[line 243] is not discharged
> Remaining obligations: {Statement x 1}

However, the fix is simple, so I can take care of it.

haridsv:
I see the output is archived, but it is not clear what the error is from the below output (seems garbled:

> [ERROR] Failed to execute goal com.diffplug.spotless:spotless-maven-plugin:2.30.0:check (default-cli) on project phoenix-core-client: The following files had format violations:
> [ERROR]     src/main/java/org/apache/phoenix/schema/MetaDataClient.java
> [ERROR]         @@ -4745,9 +4745,9 @@
> [ERROR]          Â·Â·Â·Â·Â·Â·Â·Â·Â·Â·/**
> [ERROR]          Â·Â·Â·Â·Â·Â·Â·Â·Â·Â·Â·*Â·ToÂ·checkÂ·ifÂ·TTLÂ·isÂ·definedÂ·atÂ·anyÂ·ofÂ·theÂ·childÂ·belowÂ·weÂ·areÂ·checkingÂ·itÂ·at
> [ERROR]          Â·Â·Â·Â·Â·Â·Â·Â·Â·Â·Â·*Â·{@linkÂ·org.apache.phoenix.coprocessor.MetaDataEndpointImpl#mutateColumn(List,Â·ColumnMutator,Â·int,Â·PTable,Â·PTable,Â·boolean)}
> [ERROR]         -Â·Â·Â·Â·Â·Â·Â·Â·Â·Â·Â·*Â·levelÂ·whereÂ·inÂ·functionÂ·{@linkÂ·org.apache.phoenix.coprocessor.MetaDataEndpointImpl#
> [ERROR]         -Â·Â·Â·Â·Â·Â·Â·Â·Â·Â·Â·*Â·validateIfMutationAllowedOnParent(PTable,Â·List,Â·PTableType,Â·long,Â·byte[],Â·byte[],
> [ERROR]         -Â·Â·Â·Â·Â·Â·Â·Â·Â·Â·Â·*Â·byte[],Â·List,Â·int)}Â·weÂ·areÂ·alreadyÂ·traversingÂ·throughÂ·allDescendantViews.
> [ERROR]         +Â·Â·Â·Â·Â·Â·Â·Â·Â·Â·Â·*Â·levelÂ·whereÂ·inÂ·function
> [ERROR]         +Â·Â·Â·Â·Â·Â·Â·Â·Â·Â·Â·*Â·{@linkÂ·org.apache.phoenix.coprocessor.MetaDataEndpointImpl#Â·validateIfMutationAllowedOnParent(PTable,Â·List,Â·PTableType,Â·long,Â·byte[],Â·byte[],Â·byte[],Â·List,Â·int)}
> [ERROR]         +Â·Â·Â·Â·Â·Â·Â·Â·Â·Â·Â·*Â·weÂ·areÂ·alreadyÂ·traversingÂ·throughÂ·allDescendantViews.
> [ERROR]          Â·Â·Â·Â·Â·Â·Â·Â·Â·Â·Â·*/
> [ERROR]          Â·Â·Â·Â·Â·Â·Â·Â·}

virajjasani:
It's fine, let's merge it!

haridsv:
I see the reason, reverted it and added an annotation to suppress the warning.

PR #2229, opened 2025-07-16
PHOENIX-7582 Accept phoenix.skip.system.tables.existence.check from h…

meszinorbi:
…base-site.xml


meszinorbi:
I've added the IT, but another used the SkipSystemTablesExistenceCheckIT name which is rather a result iterator check, so I renamed it.

virajjasani:
Let's backport to 5.3?

meszinorbi:
It should be okay to backport in my opinion. I'll create a PR for it.

PR #2230, opened 2025-07-17
PHOENIX-7671 Sync maxLookback from data table to indexes

virajjasani:
Co-authored-by: Palash Chauhan <palashc983@gmail.com>

Jira: PHOENIX-7671

TTL of the CDC index is same as the maxLookback configured at the cluster level. If the data table has different value of maxLookback than cluster level, the table level maxLookback is not used as TTL by the cdc index of the same data table.

The purpose of this PR is to sync the value of maxLookback from data table to index table's TableDescriptor while creating the index. Similarly, when the maxLookback value of the data table is altered using ALTER TABLE DDL, same value should be applied to all indexes.
The major compaction of CDC index should therefore consider data table's maxLookback value (from data table's descriptor) as it's TTL value.

### DIFF
--- a/phoenix-core-client/src/main/java/org/apache/phoenix/schema/MetaDataClient.java
+++ b/phoenix-core-client/src/main/java/org/apache/phoenix/schema/MetaDataClient.java
@@ -17,6 +17,7 @@
  */
 package org.apache.phoenix.schema;
 
+import static org.apache.phoenix.coprocessorclient.BaseScannerRegionObserverConstants.PHOENIX_MAX_LOOKBACK_AGE_CONF_KEY;
 import static org.apache.phoenix.exception.SQLExceptionCode.CANNOT_SET_CONDITIONAL_TTL_ON_TABLE_WITH_MULTIPLE_COLUMN_FAMILIES;
 import static org.apache.phoenix.exception.SQLExceptionCode.CANNOT_TRANSFORM_TRANSACTIONAL_TABLE;
 import static org.apache.phoenix.exception.SQLExceptionCode.CDC_ALREADY_ENABLED;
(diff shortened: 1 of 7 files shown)


phoenix-core/src/it/java/org/apache/phoenix/end2end/CDCStreamIT.java
```diff
@@ -95,6 +99,146 @@ public static synchronized void doSetup() throws Exception {
...
+     * Test to verify CDC events with index/table level maxLookback and TTL.
+     */
+    @Test
+    public void testCdcWithMaxLookbackAndCompaction() throws Exception {
+        Connection conn = newConnection();
+        String tableName = generateUniqueName();
+        String cdcName = generateUniqueName();
+        String cdcDdl = "CREATE CDC " + cdcName + " ON " + tableName;
+        // maxLookback 27 hr
+        conn.createStatement().execute(
```
tkhurana, line 115: Can you add a testcase for ALTER TABLE where the maxlookback on data table changes

virajjasani: Do we update `KEEP_DELETED_CELLS` and `REPLICATION_SCOPE` of index table if one updates data table using ALTER TABLE? I saw UpgradeUtil has some utility to keep them in sync but that utility is only in use for version less than 4.15 (PHOENIX-3955).

virajjasani: We don't have sync for them at TableDescriptor props level. But for maxLookback, let me add the sync during ALTER TABLE using TableDescriptor rather than CF level, good point!

tkhurana: I think REPLICATION_SCOPE and other properties remain in sync on ALTER

virajjasani: The earlier properties are chosen to be synced at CF level because indexes always have only default CF?

phoenix-core-client/src/main/java/org/apache/phoenix/schema/MetaDataClient.java
```diff
@@ -1251,6 +1250,13 @@ private void populatePropertyMaps(ListMultimap<String,Pair<String,Object>> state
                             .setMessage("Property: " + prop.getFirst()).build()
                             .buildException();
                 }
+                if (tableType == PTableType.INDEX
```
palashc, line 1253: Should we add a test to verify this exception being thrown?

virajjasani: Yes, added in latest commits

PR #2232, opened 2025-07-18
PHOENIX-7663 : BSON Condition Function field_type()

palashc:


phoenix-core-client/src/main/java/org/apache/phoenix/expression/util/bson/SQLComparisonExpressionUtils.java
```diff
@@ -636,6 +648,69 @@ private static boolean contains(final String fieldKey, final String containsValu
...
+      case "BOOL":
+        return fieldValue.isBoolean();
+      case "NULL":
+        return fieldValue.isNull();
+      case "L":
+        return fieldValue.isArray();
+      case "M":
+        return fieldValue.isDocument();
+      case "SS":
+        return CommonComparisonExpressionUtils.isBsonSet(fieldValue)
```
virajjasani: For sets, it is possible that there might be no element in the map. In that case, we need to return false.

phoenix-core-client/src/main/java/org/apache/phoenix/expression/util/bson/SQLComparisonExpressionUtils.java
```diff
@@ -636,6 +648,69 @@ private static boolean contains(final String fieldKey, final String containsValu
...
+                && ((BsonArray) ((BsonDocument) fieldValue).get("$set")).get(0).isString();
+      case "NS":
+        return CommonComparisonExpressionUtils.isBsonSet(fieldValue)
+                && ((BsonArray) ((BsonDocument) fieldValue).get("$set")).get(0).isNumber();
+      case "BS":
+        return CommonComparisonExpressionUtils.isBsonSet(fieldValue)
+                && ((BsonArray) ((BsonDocument) fieldValue).get("$set")).get(0).isBinary();
+      default:
+        throw new IllegalArgumentException("Unsupported type in field_type() for BsonConditionExpression: " + type);
```
virajjasani: We should return false or throw Exception?

palashc: If it is not one of the supported types, there should be an exception ,right? 
We will return false if the field is not found in the document.

virajjasani: Sounds good, let's throw `BsonConditionInvalidArgumentException`

phoenix-core-client/src/main/java/org/apache/phoenix/expression/util/bson/SQLComparisonExpressionUtils.java
palashc: done

phoenix-core/src/test/java/org/apache/phoenix/util/bson/ComparisonExpressionUtilsTest.java
```diff
@@ -1954,4 +1954,272 @@ private static RawBsonDocument getDocumentValue() {
     return RawBsonDocument.parse(json);
   }
 
+  @Test
+  public void testFieldTypeFunction() {
+    RawBsonDocument rawBsonDocument = getFieldTypeTestDocument();
+    RawBsonDocument compareValues = getFieldTypeCompareValDocument();
+
+    // Test String type (S)
+    assertTrue(SQLComparisonExpressionUtils.evaluateConditionExpression(
```
virajjasani, line 1901: nit: how about at least one test for array element at some index?

palashc: done

virajjasani (approved):
+1, pending build results

phoenix-core-client/src/main/java/org/apache/phoenix/parse/DocumentFieldTypeParseNode.java
```diff
@@ -0,0 +1,44 @@
+package org.apache.phoenix.parse;
```
virajjasani, line 18: missing ASF license

virajjasani:
Looks like now before pushing any changes, we need to run `mvn spotless:apply` to fix any spotless issues

virajjasani (approved):
+1, pending build results

virajjasani:
Build results look good

PR #2238, opened 2025-07-19
PHOENIX-7536 Capture Query parsing time for Read/Write path

rahulLiving:
JIRA https://issues.apache.org/jira/browse/PHOENIX-7536


rahulLiving:
@NihalJain Thanks for the PR. Checkstyle is passing with the latest change

PR #2239, opened 2025-07-21
PHOENIX-7659: Leverage = ANY() instead of big IN list to do huge number of point lookups in a single query

sanjeet006py:
JIRA: [PHOENIX-7659](https://issues.apache.org/jira/browse/PHOENIX-7659)

### DIFF
--- a/phoenix-core-client/src/main/java/org/apache/phoenix/compile/WhereOptimizer.java
+++ b/phoenix-core-client/src/main/java/org/apache/phoenix/compile/WhereOptimizer.java
@@ -1985,7 +1985,7 @@ public class WhereOptimizer {
      *                       are to be extracted
      * @return true if the scan ranges can be extracted, false otherwise
      */
-    private boolean prepareForArrayAnyComparisonKeySlots(ArrayAnyComparisonExpression node,
+    private boolean shouldExtractKeyRangesForArrayAnyExpr(ArrayAnyComparisonExpression node,
       List<Expression> keyExpressions) {
       // {@link ArrayAnyComparisonExpression} has two children, and the second child is
       // comparison expression
@@ -2023,6 +2023,8 @@ public class WhereOptimizer {
         }
         // Capture {@link RowKeyColumnExpression} for the generation of key slots.
         keyExpressions.add(rhs);
+      } else {
+        return false;
       }
       return true;
     }
@@ -2030,7 +2032,7 @@ public class WhereOptimizer {
     @Override
     public Iterator<Expression> visitEnter(ArrayAnyComparisonExpression node) {
       ArrayList<Expression> keyExpressions = new ArrayList<>();
-      if (prepareForArrayAnyComparisonKeySlots(node, keyExpressions)) {
+      if (shouldExtractKeyRangesForArrayAnyExpr(node, keyExpressions)) {
         return keyExpressions.iterator();
       }
       // If the scan ranges cannot be extracted, we return an empty iterator
@@ -2043,7 +2045,7 @@ public class WhereOptimizer {
         return null;
       }
       // Doing type casting is safe here as we won't have reached here unless the expression
-      // tree is of the form expected by the method prepareForArrayAnyComparisonKeySlots.
+      // tree is of the form expected by the method shouldExtractKeyRangesForArrayAnyExpr.
       Expression arrayExpr = node.getChildren().get(0);
       PhoenixArray arr = (PhoenixArray) ((LiteralExpression) arrayExpr).getValue();
       int numElements = arr.getDimensions();
--- a/phoenix-core/src/it/java/org/apache/phoenix/end2end/WhereOptimizerForArrayAnyIT.java
+++ b/phoenix-core/src/it/java/org/apache/phoenix/end2end/WhereOptimizerForArrayAnyIT.java
@@ -681,6 +681,95 @@ public class WhereOptimizerForArrayAnyIT extends BaseTest {
     }
   }
 
+  @Test
+  public void testArrayAnyComparisonWithIndexPKColumn() throws Exception {
+    String tableName = generateUniqueName();
+    String createTableDdl = "CREATE TABLE " + tableName + " (" + "pk1 INTEGER NOT NULL, "
+      + "pk2 VARCHAR(3), " + "col1 VARCHAR, " + "CONSTRAINT pk PRIMARY KEY (pk1, pk2)" + ")";
+    String createIndexDdl = "CREATE INDEX idx_pk1 ON " + tableName + " (col1)";
+    try (Connection conn = DriverManager.getConnection(getUrl())) {
+      try (Statement stmt = conn.createStatement()) {
+        stmt.execute(createTableDdl);
+        conn.commit();
+        stmt.execute(createIndexDdl);
+        conn.commit();
+      }
+    }
+    insertData(tableName, 1, "a", "b");
+    try (Connection conn = DriverManager.getConnection(getUrl())) {
+      String selectSql = "SELECT * FROM " + tableName + " WHERE col1 = ANY(?)";
+      Array arr = conn.createArrayOf("VARCHAR", new String[] { "a", "b", "c" });
+      try (PreparedStatement stmt = conn.prepareStatement(selectSql)) {
+        stmt.setArray(1, arr);
+        try (ResultSet rs = stmt.executeQuery()) {
+          assertTrue(rs.next());
+          assertEquals(1, rs.getInt(1));
+          assertEquals("a", rs.getString(2));
+          assertEquals("b", rs.getString(3));
+        }
+        assertSkipScanIsGenerated(stmt, 3);
+      }
+    }
+  }
+
+  @Test
+  public void testArrayAnyComparisonWithRowKeyPrefix() throws Exception {
+    String tableName = generateUniqueName();
+    String ddl = "CREATE TABLE " + tableName + " (" + "pk1 INTEGER NOT NULL, " + "pk2 VARCHAR(3), "
+      + "col1 VARCHAR, " + "CONSTRAINT pk PRIMARY KEY (pk1, pk2)" + ")";
+    try (Connection conn = DriverManager.getConnection(getUrl())) {
+      try (Statement stmt = conn.createStatement()) {
+        stmt.execute(ddl);
+        conn.commit();
+      }
+    }
+    insertData(tableName, 1, "a", "b");
+    try (Connection conn = DriverManager.getConnection(getUrl())) {
+      String selectSql = "SELECT * FROM " + tableName + " WHERE pk1 = ANY(?)";
+      Array arr = conn.createArrayOf("INTEGER", new Integer[] { 1, 2, 3 });
+      try (PreparedStatement stmt = conn.prepareStatement(selectSql)) {
+        stmt.setArray(1, arr);
+        try (ResultSet rs = stmt.executeQuery()) {
+          assertTrue(rs.next());
+          assertEquals(1, rs.getInt(1));
+          assertEquals("a", rs.getString(2));
+          assertEquals("b", rs.getString(3));
+        }
+        assertSkipScanIsGenerated(stmt, 3);
+      }
+    }
+  }
+
+  @Test
+  public void testArrayAnyComparisonWhenRowKeyColumnExpressionIsNotTopLevelExpression()
+    throws Exception {
+    String tableName = generateUniqueName();
+    String ddl = "CREATE TABLE " + tableName + " (" + "pk1 INTEGER NOT NULL, " + "pk2 VARCHAR(3), "
+      + "col1 VARCHAR, " + "CONSTRAINT pk PRIMARY KEY (pk1, pk2)" + ")";
+    try (Connection conn = DriverManager.getConnection(getUrl())) {
+      try (Statement stmt = conn.createStatement()) {
+        stmt.execute(ddl);
+        conn.commit();
+      }
+    }
+    insertData(tableName, 1, "a", "b");
+    try (Connection conn = DriverManager.getConnection(getUrl())) {
+      String selectSql =
+        "SELECT * FROM " + tableName + " WHERE CAST (pk1 as BIGINT) = ANY(?) AND pk2 = 'a'";
+      Array arr = conn.createArrayOf("BIGINT", new Long[] { 1L, 2L, 3L });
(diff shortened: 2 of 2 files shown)


phoenix-core-client/src/main/java/org/apache/phoenix/compile/WhereOptimizer.java
```diff
@@ -1945,6 +1948,150 @@ public KeySlots visitLeave(InListExpression node, List<KeySlots> childParts) {
...
+              || keyRange == KeyRange.IS_NULL_RANGE
+          ) {
+            // Skip null range along with empty range as null check is done via IS NULL as
+            // per SQL standards
+            continue;
+          }
+          keyRanges.add(keyRange);
+        }
+      } catch (Exception e) {
+        LOGGER.warn(String.format(
```
virajjasani: Let's avoid using `String.format()` for loggers, we can use parameterized log format with `{}` for each object that needs to be logged

phoenix-core/src/it/java/org/apache/phoenix/end2end/WhereOptimizerForArrayAnyIT.java
```diff
@@ -0,0 +1,807 @@
...
+          assertEquals(col1Value, rs.getString(4));
+        }
+      }
+      try (PreparedStatement stmt = conn.prepareStatement("EXPLAIN " + selectSql)) {
+        stmt.setArray(1, binaryArr);
+        stmt.setArray(2, varbinaryArr);
+        stmt.setArray(3, varbinaryArr);
+        try (ResultSet rs = stmt.executeQuery()) {
+          String explainPlan = QueryUtil.getExplainPlan(rs);
+          assertTrue(explainPlan.contains("POINT LOOKUP ON"));
```
virajjasani: For each of these places, since we just need to scan type, let's use `ExplainPlanAttributes` based assertion:
```
    ExplainPlan plan = stmt.unwrap(PhoenixPreparedStatement.class).optimizeQuery().getExplainPlan();
    ExplainPlanAttributes planAttributes = plan.getPlanStepsAsAttributes();
    assertEquals("POINT LOOKUP ON N KEY ", planAttributes.getExplainScanType());
```

sanjeet006py: Sure. One doubt: Can you please tell why is this better than existing approach? Thanks

virajjasani: ExplainPlan V2 APIs were built to verify values of specific attributes instead of converting to full string and compare using contains()

sanjeet006py: Got it, thanks.

phoenix-core-client/src/main/java/org/apache/phoenix/compile/WhereOptimizer.java
```diff
@@ -1945,6 +1948,150 @@ public KeySlots visitLeave(InListExpression node, List<KeySlots> childParts) {
...
+     * This will help in saving the query parsing time as on using IN list query parsing time
+     * increases with the size of IN list but in case of ANY expression it is constant. Below we
+     * account for cases where COL is on the LHS or RHS of the comparison expression.
+     * @param node           {@link ArrayAnyComparisonExpression} node for which scan ranges are to
+     *                       be extracted
+     * @param keyExpressions {@link RowKeyColumnExpression} for the PK column for which scan ranges
+     *                       are to be extracted
+     * @return true if the scan ranges can be extracted, false otherwise
+     */
+    private boolean prepareForArrayAnyComparisonKeySlots(ArrayAnyComparisonExpression node,
```
tkhurana: This method can be renamed. It is not at all descriptive. Generally boolean returning methods are like `is..` or `should...`. It could be `shouldExtractKeyRangesForAnyExpr`.

phoenix-core-client/src/main/java/org/apache/phoenix/compile/WhereOptimizer.java
```diff
@@ -1945,6 +1948,150 @@ public KeySlots visitLeave(InListExpression node, List<KeySlots> childParts) {
...
+     * <pre>
+     * COL = ANY(ARR)
+     * </pre>
+     *
+     * then we can extract the scan ranges for the COL, given COL is a PK column. This syntactical
+     * pattern can be used as a replacement for a IN expression. So, instead of following IN
+     * expression:
+     *
+     * <pre>
+     * COL IN (VAL1, VAL2, ... VALN)
```
tkhurana, line 1963: I don't think we can use this for RVC list like (COL1, COL2) IN ( (val11,val12), (val21, val22))

phoenix-core-client/src/main/java/org/apache/phoenix/compile/WhereOptimizer.java
```diff
@@ -1945,6 +1948,150 @@ public KeySlots visitLeave(InListExpression node, List<KeySlots> childParts) {
...
+
+      // {@link ComparisonExpression} will have two children in this case, we need to make
+      // sure that one of them is a {@link RowKeyColumnExpression} and the other is a {@link
+      // ArrayElemRefExpression}. Further, the first child of {@link ArrayElemRefExpression}
+      // must be a {@link LiteralExpression}. The first child of {@link
+      // ArrayElemRefExpression} is same as the first child of {@link
+      // ArrayAnyComparisonExpression}.
+      Expression lhs = comparisonExpr.getChildren().get(0);
+      Expression rhs = comparisonExpr.getChildren().get(1);
+      if (lhs instanceof RowKeyColumnExpression && rhs instanceof ArrayElemRefExpression) {
```
tkhurana, line 2011: What if the RowKeyColumnExpression is embedded inside another expression like CoerceExpression. That should not have any impact on the scan range generated.

phoenix-core/src/it/java/org/apache/phoenix/end2end/WhereOptimizerForArrayAnyIT.java
```diff
@@ -0,0 +1,792 @@
...
+    setUpTestDriver(new ReadOnlyProps(new HashMap<String, String>()));
+  }
+
+  @Test
+  public void testArrayAnyComparisonForNonPkColumn() throws Exception {
+    String tableName = generateUniqueName();
+    createTableASCPkColumns(tableName);
+    insertData(tableName, 1, "x", "a");
+    try (Connection conn = DriverManager.getConnection(getUrl())) {
+      String selectSql = "SELECT * FROM " + tableName + " WHERE col1 = ANY(?)";
```
tkhurana, line 66: Can you also add a testcase where there is an index on the table and the ANY is on a column which is part of index pk and not data table pk.

phoenix-core-client/src/main/java/org/apache/phoenix/compile/WhereOptimizer.java
tkhurana: In other words, can we do this only if the RowKeyColumnExpression is the top level expression ?

phoenix-core/src/it/java/org/apache/phoenix/end2end/WhereOptimizerForArrayAnyIT.java
```diff
@@ -0,0 +1,792 @@
...
+  }
+
+  @Test
+  public void testArrayAnyComparisonWithLiteralArray() throws Exception {
+    String tableName = generateUniqueName();
+    createTableASCPkColumns(tableName);
+    insertData(tableName, 1, "x", "a");
+    try (Connection conn = DriverManager.getConnection(getUrl())) {
+      String selectSql =
+        "SELECT * FROM " + tableName + " WHERE pk1 = 1 AND pk2 = ANY(ARRAY['x', 'y'])";
```
tkhurana, line 130: Can you also add a testcase where only a prefix of the primary key is available

phoenix-core-client/src/main/java/org/apache/phoenix/compile/WhereOptimizer.java
```diff
@@ -1945,6 +1948,150 @@ public KeySlots visitLeave(InListExpression node, List<KeySlots> childParts) {
...
+
+      } else if (lhs instanceof ArrayElemRefExpression && rhs instanceof RowKeyColumnExpression) {
+        ArrayElemRefExpression arrayElemRefExpr = (ArrayElemRefExpression) lhs;
+        if (!(arrayElemRefExpr.getChildren().get(0) instanceof LiteralExpression)) {
+          return false;
+        }
+        // Capture {@link RowKeyColumnExpression} for the generation of key slots.
+        keyExpressions.add(rhs);
+      }
+      return true;
```
tkhurana, line 2029: Should there be an else condition here ? What is the expected behavior when none of the if-else-if conditions are met ?

tkhurana:
With inlist we have some special handling to prevent key space explosion https://github.com/apache/phoenix/blob/master/phoenix-core-client/src/main/java/org/apache/phoenix/compile/WhereOptimizer.java#L252-L262 We can have similar key space explosion with ANY().  Don't we need to handle that ?

phoenix-core-client/src/main/java/org/apache/phoenix/compile/WhereOptimizer.java
sanjeet006py: Yes, using Array Any we can't use RVC list as we don't support 2D arrays. I tried this.

phoenix-core-client/src/main/java/org/apache/phoenix/compile/WhereOptimizer.java
sanjeet006py: Yes, currently I have scoped the changes such that only if RowKeyColumnExpression is a top level expression or from customer point of view, if they directly specify a condition on PK column (w/o wrapping those in a expression), only then KeyRange will be generated for Array Any. 
We can later extend it to cover more cases as and when needed.

phoenix-core-client/src/main/java/org/apache/phoenix/compile/WhereOptimizer.java
sanjeet006py: Thanks!! I missed it and there is no test case even checking that. Will fix it and add test case.

virajjasani (approved):

phoenix-core-client/src/main/java/org/apache/phoenix/compile/WhereOptimizer.java
```diff
@@ -1945,6 +1948,152 @@ public KeySlots visitLeave(InListExpression node, List<KeySlots> childParts) {
...
+     *
+     * <pre>
+     * try (Connection conn = DriverManager.getConnection(url)) {
+     *      conn.createArrayOf("CHAR", new String[] {"VAL1", "VAL2", ... "VALN"});
+     *      try (PreparedStatement stmt = conn.prepareStatement(
+     *          "SELECT ... FROM TABLE WHERE COL = ANY(?)")) {
+     *          stmt.setArray(1, arr);
+     *          ResultSet rs = stmt.executeQuery();
+     *      }
+     * }
```
haridsv, line 1976:
```suggestion
     * we can use the above ANY expression, like this:
     *
     * <pre>
     * try (Connection conn = DriverManager.getConnection(url)) {
     *      Array arr = conn.createArrayOf("CHAR", new String[] {"VAL1", "VAL2", ... "VALN"});
     *      try (PreparedStatement stmt = conn.prepareStatement(
     *          "SELECT ... FROM TABLE WHERE COL = ANY(?)")) {
     *          stmt.setArray(1, arr);
     *          ResultSet rs = stmt.executeQuery();
     *      }
     * }

File: phoenix-core-client/src/main/java/org/apache/phoenix/compile/WhereOptimizer.java [hunk 10]
```diff
@@ -1945,6 +1948,152 @@ public KeySlots visitLeave(InListExpression node, List<KeySlots> childParts) {
...
+     * </pre>
+     *
+     * This will help in saving the query parsing time as on using IN list query parsing time
```
haridsv, line 1979: It will also make the `PreparedStatement` highly reusable unlike in case of `IN` list where you need the correct number of bind parameters for one to be reusable.

phoenix-core-client/src/main/java/org/apache/phoenix/compile/WhereOptimizer.java
```diff
@@ -1945,6 +1948,152 @@ public KeySlots visitLeave(InListExpression node, List<KeySlots> childParts) {
...
+     * increases with the size of IN list but in case of ANY expression it is constant. Below we
+     * account for cases where COL is on the LHS or RHS of the comparison expression.
+     * @param node           {@link ArrayAnyComparisonExpression} node for which scan ranges are to
+     *                       be extracted
+     * @param keyExpressions {@link RowKeyColumnExpression} for the PK column for which scan ranges
+     *                       are to be extracted
+     * @return true if the scan ranges can be extracted, false otherwise
+     */
+    private boolean shouldExtractKeyRangesForArrayAnyExpr(ArrayAnyComparisonExpression node,
```
haridsv, line 1988: Nit: Method  name feels like it is a readonly operation, doesn't clearly indicate that `keyExpressions` will be modified.

phoenix-core-client/src/main/java/org/apache/phoenix/compile/WhereOptimizer.java
```diff
@@ -1945,6 +1948,152 @@ public KeySlots visitLeave(InListExpression node, List<KeySlots> childParts) {
...
+        keyExpressions.add(lhs);
+
+      } else if (lhs instanceof ArrayElemRefExpression && rhs instanceof RowKeyColumnExpression) {
+        ArrayElemRefExpression arrayElemRefExpr = (ArrayElemRefExpression) lhs;
+        if (!(arrayElemRefExpr.getChildren().get(0) instanceof LiteralExpression)) {
+          return false;
+        }
+        // Capture {@link RowKeyColumnExpression} for the generation of key slots.
+        keyExpressions.add(rhs);
+      } else {
```
haridsv, line 2026: Nit: you can avoid repetitiveness using 2 pairs of (lhs, rhs) and (rhs, lhs).

PR #2244, opened 2025-07-23
PHOENIX-7566 [1/n] HAGroupStoreClient/Manager changes for Consistent HA

ritegarg:


phoenix-core-client/src/main/java/org/apache/phoenix/query/QueryServices.java
```diff
@@ -140,7 +140,15 @@ public interface QueryServices extends SQLCloseable {
...
-    public static final String HA_GROUP_NAME_ATTRIB = "phoenix.ha.group";
+    public static final String HA_IMPLEMENTATION = "phoenix.ha.implementation";
+    // TODO: Revisit this default value based on standard ZK update time
+    // and time it takes for an update to reach regionserver cache.
+    // Refresh interval for store and forward mode status updates in milliseconds
+    public static final String HA_STORE_AND_FORWARD_MODE_REFRESH_INTERVAL_MS = "phoenix.ha.store.forward.mode.refresh.interval.ms";
+    // TODO: Revisit this default value based on standard ZK update time
+    // and time it takes for an update to reach regionserver cache.
+    // Refresh interval for sync mode status updates in milliseconds
+    public static final String HA_SYNC_MODE_REFRESH_INTERVAL_MS = "phoenix.ha.sync.mode.refresh.interval.ms";
```
ritegarg: Plan to remove this in next iteration. We will use "zookeeper.session.timeout" with a multiplier(default 1.1) for this value. Reason is that regionservers create ephemeral node in ZK once it starts and if RS is not connected to ZK for "zookeeper.session.timeout" , RegionServerTracker will abort the regionserver for which ephemeral node was deleted.

phoenix-core-server/src/main/java/org/apache/phoenix/coprocessor/UngroupedAggregateRegionObserver.java
```diff
@@ -1055,20 +1055,4 @@ public void preClose(ObserverContext<RegionCoprocessorEnvironment> c, boolean ab
     protected boolean isRegionObserverFor(Scan scan) {
         return scan.getAttribute(BaseScannerRegionObserverConstants.UNGROUPED_AGG) != null;
     }
-
-    @Override
-    public void preBatchMutate(ObserverContext<RegionCoprocessorEnvironment> c,
```
ritegarg, line 1060: Discussed with @tkhurana and concluded that this is not needed anymore. We will be relying on IndexRegionObserver for mutation

kadirozde:
HA group store manager, client, etc. are currently included in phoenix-core-client. Should they be included in phoenix-core-server and possibly with their own package HAGroupStore? I think the HA group store client can be package private as it should not be used outside the package. @tkhurana, what do you think?

phoenix-core-client/src/main/java/org/apache/phoenix/jdbc/HAGroupStoreClient.java
```diff
@@ -17,191 +17,640 @@
...
      *
      * @param conf configuration
+     * @param haGroupName name of the HA group. Only specified HA group is tracked.
+     * @param zkUrl zkUrl to use for the client. Prefer providing this parameter to avoid
+     *             the overhead of getting the local zkUrl from the configuration.
      * @return HAGroupStoreClient instance
      */
-    public static HAGroupStoreClient getInstance(Configuration conf) {
-        if (haGroupStoreClientInstance == null || !haGroupStoreClientInstance.isHealthy) {
+    public static HAGroupStoreClient getInstance(Configuration conf, String haGroupName,
```
kadirozde: Can we have two methods, getInstance and getInstanceForLocalZK?

phoenix-core-client/src/main/java/org/apache/phoenix/jdbc/HAGroupStoreClient.java
```diff
@@ -17,191 +17,640 @@
...
+     */
+    public void rebuild(boolean broadcastUpdate) throws Exception {
+        if (!isHealthy) {
+            throw new IOException("HAGroupStoreClient is not healthy");
         }
-        return null;
+        LOGGER.info("Rebuilding HAGroupStoreClient for HA group {} with broadcastUpdate {}",
+                haGroupName, broadcastUpdate);
+        if (broadcastUpdate) {
+            // TODO: Add a lock for the row in system table.
```
kadirozde: The lock should be taken on the region where the row is. So, it needs to be done in a coproc hook.

phoenix-core-client/src/main/java/org/apache/phoenix/jdbc/HAGroupStoreClient.java
```diff
@@ -17,191 +17,640 @@
...
+            && newHAGroupState == HAGroupState.ACTIVE_NOT_IN_SYNC) {
+                //This is to avoid extra requests to ZK for updates.
+                waitTime = waitTimeForStoreAndForwardModeInMs;
+            }
+        } else {
+            throw new InvalidClusterRoleTransitionException("Cannot transition from "
+                    + currentHAGroupState + " to " + newHAGroupState);
         }
-        return ImmutableList.copyOf(clusterRoleToCRRMap.getOrDefault(clusterRole, new ConcurrentHashMap<>()).values());
+        return ((System.currentTimeMillis() - currentHAGroupStoreRecordMtime) > waitTime);
```
kadirozde, line 638: It is not safe to skip an update because a refresh interval has not expired. The caller would not know if the update is skipped or not necessary. Do we really need this refresh interval?

ritegarg: I added it to reduce unnecessary updates to ZK. But I think it is not required. I have removed the config and its usage

phoenix-core-client/src/main/java/org/apache/phoenix/jdbc/HAGroupStoreClient.java
ritegarg: Updated

phoenix-core-client/src/main/java/org/apache/phoenix/jdbc/HAGroupStoreRecord.java
```diff
@@ -0,0 +1,265 @@
...
+     * Enum representing the HA group state with each state having a corresponding ClusterRole.
+     */
+    public enum HAGroupState {
+        ABORT_TO_ACTIVE,
+        ABORT_TO_STANDBY,
+        ACTIVE_IN_SYNC,
+        ACTIVE_NOT_IN_SYNC,
+        ACTIVE_NOT_IN_SYNC_TO_STANDBY,
+        ACTIVE_NOT_IN_SYNC_WITH_OFFLINE_PEER,
+        ACTIVE_TO_STANDBY,
```
kadirozde: I think ACTIVE_TO_STANDBY should be renamed to ACTIVE_IN_SYNC_TO_STANDBY

phoenix-core-client/src/main/java/org/apache/phoenix/jdbc/HAGroupStoreRecord.java
```diff
@@ -0,0 +1,265 @@
...
+public class HAGroupStoreRecord {
+
+    private static final Logger LOG = LoggerFactory.getLogger(HAGroupStoreRecord.class);
+    public static final String DEFAULT_PROTOCOL_VERSION = "1.0";
+
+    /**
+     * Enum representing the HA group state with each state having a corresponding ClusterRole.
+     */
+    public enum HAGroupState {
+        ABORT_TO_ACTIVE,
```
kadirozde: I assume the admin will be able to abort the failover when the active cluster in state ACTIVE_NOT_IN_SYNC_TO_STANDBY.  In that case, we need to have two abort states, ABORT_TO_ACTIVE_IN_SYNC and ABORT_TO_ACTIVE_NOT_IN_SYNC instead of just ABORT_TO_ACTIVE.

ritegarg: Replaced ABORT_TO_ACTIVE with  ABORT_TO_ACTIVE_IN_SYNC  and added ABORT_TO_ACTIVE_NOT_IN_SYNC. Also added their default role and allowed transitions

phoenix-core-client/src/main/java/org/apache/phoenix/jdbc/HAGroupStoreClient.java
ritegarg: The plan of action is:-

1. Intercept the mutation to System.HA_Group table in IndexRegionObserver.
2. Obtain read and write lock on row to maintain serialization
3. If attribute changed is in worth broadcasting an update. eg. we don't need to broadcast update for clusterRole because ZNode will contain the state already.
                   a. If broadcast worthy, we use same mechanism as MetadataCache to invalidate. We create a new RPC                  to send updated attributes to all other HAGroupStoreClient and each HAGroupStoreClient updates its attributes. This has 1 retry built in.
                   b. Once complete for all RS, we let mutation complete. If mutation failed for even 1 RS, we fail the mutation and operator can retry.
4. Release the lock

phoenix-core-client/src/main/java/org/apache/phoenix/jdbc/HAGroupStoreClient.java
ritegarg: Adding row locking and broadcast system table update to all other RS’s HAGroupStoreClient instance
1. Intercept the mutation to System.HA_Group table in IndexRegionObserver.
2. Obtain read and write lock on row to maintain serialization
3. If attribute changed is in worth broadcasting an update. eg. we don't need to broadcast update for clusterRole because ZNode will contain the state already.
             a. If broadcast worthy, we use same mechanism as MetadataCache to invalidate. We create a new RPC to send updated attributes to all other HAGroupStoreClient and each HAGroupStoreClient updates its attributes. This has 1 retry built in.
             b. Once complete for all RS, we let mutation complete. If mutation failed for even 1 RS, we fail the mutation and operator can retry.
4. Release the lock


**NOTE**: Need to check if we need to obtain Scan lock as well as it might be possible we retrieve the locked entry in scans as well while the mutation of row is in progress and HAGroupStoreClient might have inconsistent data. Currently we have 2 use cases for the table 
1. During initialization we retrieve table entry for HAGroupName via where clause for HAGroupName which is the primary key.
2. We have an option to retrieve all the HAGroupNames associated with local cluster where we scan this table. This output should not change unless we delete the HAGroupName entry
**NOTE**: Need to handle delete case.

kadirozde (approved):
+1, Thanks!

PR #2247, opened 2025-07-25
PHOENIX-7677 TTL_DELETE CDC event to use batch mutation

virajjasani:
Jira: PHOENIX-7677

Inserting TTL_DELETE CDC event from CompactionScanner for single expired row at a time can cause slight degradation to the write workload to the CDC index table.

To improve the performance of inserting TTL_DELETE events, CompactionScanner can accumulate 50 mutations by default and perform the batchMutate(). Any mutations that are not inserted into the CDC index should be inserted with batchMutate() while closing the CompactionScanner.

### DIFF
--- a/phoenix-core-server/src/main/java/org/apache/phoenix/coprocessor/CDCCompactionUtil.java
+++ b/phoenix-core-server/src/main/java/org/apache/phoenix/coprocessor/CDCCompactionUtil.java
@@ -67,7 +67,7 @@ public final class CDCCompactionUtil {
   private static final Logger LOGGER = LoggerFactory.getLogger(CDCCompactionUtil.class);
 
   // Shared cache for row images across all CompactionScanner instances in the JVM.
-  // Entries expire after 20 minutes by default.
+  // Entries expire after 1200 seconds (20 minutes) by default.
   // The JVM level cache helps merge the pre-image for the row with multiple CFs.
   // The key of the cache contains (regionId + data table rowkey).
   // The value contains pre-image that needs to be directly inserted in the CDC index.
@@ -87,12 +87,12 @@ public final class CDCCompactionUtil {
     if (sharedTtlImageCache == null) {
       synchronized (CDCCompactionUtil.class) {
         if (sharedTtlImageCache == null) {
-          int expiryMinutes = config.getInt(QueryServices.CDC_TTL_SHARED_CACHE_EXPIRY_MINUTES,
-            QueryServicesOptions.DEFAULT_CDC_TTL_SHARED_CACHE_EXPIRY_MINUTES);
+          int expirySeconds = config.getInt(QueryServices.CDC_TTL_SHARED_CACHE_EXPIRY_SECONDS,
+            QueryServicesOptions.DEFAULT_CDC_TTL_SHARED_CACHE_EXPIRY_SECONDS);
           sharedTtlImageCache =
-            CacheBuilder.newBuilder().expireAfterWrite(expiryMinutes, TimeUnit.MINUTES).build();
-          LOGGER.info("Initialized shared CDC row image cache with expiry of {} minutes",
-            expiryMinutes);
+            CacheBuilder.newBuilder().expireAfterWrite(expirySeconds, TimeUnit.SECONDS).build();
+          LOGGER.info("Initialized shared CDC row image cache with expiry of {} seconds",
+            expirySeconds);
         }
       }
     }
--- a/phoenix-core-server/src/main/java/org/apache/phoenix/coprocessor/CDCCompactionUtil.java
+++ b/phoenix-core-server/src/main/java/org/apache/phoenix/coprocessor/CDCCompactionUtil.java
@@ -20,18 +20,17 @@ package org.apache.phoenix.coprocessor;
 import static org.apache.phoenix.query.QueryConstants.NAME_SEPARATOR;
 
 import java.io.IOException;
+import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Properties;
+import java.util.concurrent.TimeUnit;
+import org.apache.hadoop.conf.Configuration;
 import org.apache.hadoop.hbase.Cell;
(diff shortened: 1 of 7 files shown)


phoenix-core-server/src/main/java/org/apache/phoenix/coprocessor/CDCCompactionUtil.java
```diff
@@ -53,19 +54,206 @@
...
+    public void flushBatch() throws Exception {
+      if (pendingMutations.isEmpty()) {
+        return;
+      }
+
+      Exception lastException = null;
+      for (int retryCount = 0; retryCount < cdcTtlMutationMaxRetries; retryCount++) {
+        try (Table cdcIndexTable =
+          env.getConnection().getTable(TableName.valueOf(cdcIndex.getPhysicalName().getBytes()))) {
+          cdcIndexTable.put(new ArrayList<>(pendingMutations.values()));
```
palashc: For my understanding, what are the semantics here for partial failures when trying to apply batch mutations? We will log failure even if some mutations from the batch were applied after all retries?

phoenix-core/src/it/java/org/apache/phoenix/end2end/TableTTLIT.java
```diff
@@ -782,6 +784,130 @@ private void majorCompact(TableName table) throws Exception {
     TestUtil.majorCompact(getUtility(), table);
   }
 
+  /**
+   * Test CDC batch mutations for TTL expired rows. This test creates a table with TTL and CDC
+   * index, inserts 82 rows (to test batching: 25+25+25+7), lets them expire via TTL, and verifies
+   * that all 82 rows have CDC TTL_DELETE events recorded with correct pre-image data.
+   */
+  @Test
+  public void testCDCBatchMutationsForTTLExpiredRows() throws Exception {
```
palashc, line 793: Is there a way to verify the actual batching and retry logic from this IT? Do you think a metric for this operation would be helpful?

phoenix-core-server/src/main/java/org/apache/phoenix/coprocessor/CDCCompactionUtil.java
```diff
@@ -53,19 +54,206 @@
...
- * that are written directly to CDC index tables.
+ * that are written directly to CDC index tables using batch mutations.
  */
 public final class CDCCompactionUtil {
 
   private static final Logger LOGGER = LoggerFactory.getLogger(CDCCompactionUtil.class);
 
+  // Shared cache for row images across all CompactionScanner instances in the JVM.
+  // Entries expire after 1200 seconds (20 minutes) by default.
+  // The JVM level cache helps merge the pre-image for the row with multiple CFs.
```
tkhurana, line 72: Is it guaranteed that for multiple CFs all compactionscanner instances will run in the same JVM ? What if the RS aborts between compaction on two CFs ?

phoenix-core-server/src/main/java/org/apache/phoenix/coprocessor/CDCCompactionUtil.java
```diff
@@ -53,19 +54,206 @@
...
+  // The key of the cache contains (regionId + data table rowkey).
```
tkhurana, line 72: Is there any advantage to have a global cache with the region id as the key ? Couldn't this be a per region object just like CompactionScanner ?

phoenix-core-server/src/main/java/org/apache/phoenix/coprocessor/CDCCompactionUtil.java
```diff
@@ -53,19 +54,206 @@
...
   private CDCCompactionUtil() {
     // empty
   }
 
+  /**
+   * Gets the shared row image cache, initializing it lazily with configuration.
+   * @param config The Hadoop configuration to read cache expiry from
+   * @return the shared cache instance
+   */
+  static Cache<ImmutableBytesPtr, Map<String, Object>>
```
tkhurana, line 85: If its per region then we don't need all of this code because there is nothing to share.

phoenix-core-server/src/main/java/org/apache/phoenix/coprocessor/CDCCompactionUtil.java
virajjasani: Since we are using batchMutate() on the table where all the rowkeys are most likely expected to go to same region, either all mutations will be successful or none

phoenix-core/src/it/java/org/apache/phoenix/end2end/TableTTLIT.java
virajjasani: Metric would be helpful but initiating retry is not that simple, certainly should be kept separately

phoenix-core-server/src/main/java/org/apache/phoenix/coprocessor/CDCCompactionUtil.java
virajjasani: Yes, this is best case scenario only. Let me add a note here. Generating TTL_DELETE event for multi-CF is not guaranteed to be accurate due to the edge case with server crash.

virajjasani: Rather than comment in the code, let me clarify this in Jira

virajjasani: Let me also evaluate doing batch writes only during CompactionScanner close(). This way, we might be able to guarantee no incorrect values for the CDC event. Let me think about this.

phoenix-core/src/it/java/org/apache/phoenix/end2end/TableTTLIT.java
virajjasani: @palashc i have added bunch of logs, do you think we still need metrics?

tkhurana (approved):

PR #2252, opened 2025-07-29
PHOENIX-7670 Region level threadpool for uncovered index to scan da…

palashc:
…ta table


phoenix-core-server/src/main/java/org/apache/phoenix/coprocessor/UncoveredGlobalIndexRegionScanner.java
```diff
@@ -97,10 +124,16 @@ public UncoveredGlobalIndexRegionScanner(final RegionScanner innerScanner, final
...
-        .setCoreTimeout(INDEX_WRITER_KEEP_ALIVE_TIME_CONF_KEY), env));
+    try {
+      pool = getRegionThreadPoolCache().get(region,
+        () -> new WaitForCompletionTaskRunner(ThreadPoolManager
+          .getExecutor(new ThreadPoolBuilder("Uncovered Global Index", env.getConfiguration())
+            .setMaxThread(NUM_CONCURRENT_INDEX_THREADS_CONF_KEY, DEFAULT_CONCURRENT_INDEX_THREADS)
+            .setCoreTimeout(INDEX_WRITER_KEEP_ALIVE_TIME_CONF_KEY), env)));
+    } catch (ExecutionException e) {
+      throw new IOException("Failed to create thread pool for UncoveredGlobalIndexRegionScanner",
+        e.getCause());
```
virajjasani: Can we keep `e` as is here?

phoenix-core-server/src/main/java/org/apache/phoenix/coprocessor/UncoveredGlobalIndexRegionScanner.java
```diff
@@ -97,10 +124,16 @@ public UncoveredGlobalIndexRegionScanner(final RegionScanner innerScanner, final
...
 
-    pool = new WaitForCompletionTaskRunner(ThreadPoolManager
-      .getExecutor(new ThreadPoolBuilder("Uncovered Global Index", env.getConfiguration())
-        .setMaxThread(NUM_CONCURRENT_INDEX_THREADS_CONF_KEY, DEFAULT_CONCURRENT_INDEX_THREADS)
-        .setCoreTimeout(INDEX_WRITER_KEEP_ALIVE_TIME_CONF_KEY), env));
+    try {
+      pool = getRegionThreadPoolCache().get(region,
+        () -> new WaitForCompletionTaskRunner(ThreadPoolManager
+          .getExecutor(new ThreadPoolBuilder("Uncovered Global Index", env.getConfiguration())
+            .setMaxThread(NUM_CONCURRENT_INDEX_THREADS_CONF_KEY, DEFAULT_CONCURRENT_INDEX_THREADS)
```
virajjasani: Do we still want to keep 16 as default max threads for uncovered index performing scan to data table? With this PR, we have opportunity to increase it? How about 128 or 256?

cc @tkhurana

virajjasani (commented):
The core logic looks good

phoenix-core-server/src/main/java/org/apache/phoenix/coprocessor/UncoveredGlobalIndexRegionScanner.java
palashc: makes sense

phoenix-core-server/src/main/java/org/apache/phoenix/coprocessor/UncoveredGlobalIndexRegionScanner.java
```diff
@@ -178,7 +167,8 @@ protected void submitTasks(TaskBatch<Boolean> tasks) throws IOException {
     Pair<List<Boolean>, List<Future<Boolean>>> resultsAndFutures = null;
     try {
       LOGGER.debug("Waiting on index tasks to complete...");
-      resultsAndFutures = this.pool.submitUninterruptible(tasks);
+      resultsAndFutures =
+        PhoenixRegionServerEndpoint.getUncoveredIndexThreadPool(env).submitUninterruptible(tasks);
```
palashc: @virajjasani how should I handle the case where PRSE is not loaded?

palashc: Should I add this before the access?
```
try {
    Class.forName("org.apache.phoenix.coprocessor.PhoenixRegionServerEndpoint");
} catch (ClassNotFoundException e) {
    throw new IOException("PhoenixRegionServerEndpoint should be loaded to use Uncovered Indexes.");
}
```

virajjasani: This would be complex check, we don't need it. All we need to check is whether the threadpool is initialized or not. If not initialized, regionserver coproc was also not initialized, so throw error.

palashc: @virajjasani but I am initializing the thread pool lazily when a scanner needs it, not during PRSE#start since ThreadPoolManager#getExecutor needs a RegionCoprocessorEnvironment which we do not have during PRSE#start

phoenix-core-server/src/main/java/org/apache/phoenix/coprocessor/PhoenixRegionServerEndpoint.java
```diff
@@ -133,4 +151,24 @@ public ServerMetadataCache getServerMetadataCache() {
     return ServerMetadataCacheImpl.getInstance(conf);
   }
 
+  public static TaskRunner getUncoveredIndexThreadPool(RegionCoprocessorEnvironment env) {
+    if (uncoveredIndexThreadPool == null) {
+      synchronized (PhoenixRegionServerEndpoint.class) {
```
virajjasani: Why do we need to start the pool lazily? Why not start it upfront in init() specifically when we can set lower core pool size than max pool size. Core pool size can be 10 and max pool size 512 (default). Also, setting idle thread timeout will help terminate threads if they stay idle (i.e. no one uses uncovered index)

palashc: I wanted to use `ThreadPoolManager` but that needs a `RegionCoprocessorEnvironment`

virajjasani:
16:17:38  |  -1  |       patch  |   0m  5s   | 
16:17:38  |      |              |            | https://github.com/apache/phoenix/pull/2252
16:17:38  |      |              |            | does not apply to master. Rebase required?
16:17:38  |      |              |            | Wrong Branch? See
16:17:38  |      |              |            | https://yetus.apache.org/documentation/in-pr
16:17:38  |      |              |            | ogress/precommit-patchnames for help.

virajjasani:
The PR needs rebase

phoenix-core-client/src/main/java/org/apache/phoenix/query/QueryServicesOptions.java
```diff
@@ -501,6 +501,9 @@ public class QueryServicesOptions {
   public static final long DEFAULT_PHOENIX_CDC_STREAM_PARTITION_EXPIRY_MIN_AGE_MS =
     30 * 60 * 60 * 1000; // 30 hours
 
+  public static final int DEFAULT_PHOENIX_UNCOVERED_INDEX_MAX_POOL_SIZE = 512;
+  public static final int DEFAULT_PHOENIX_UNCOVERED_INDEX_KEEP_ALIVE_TIME_MS = 60 * 1000; // 1min
```
virajjasani: This is in seconds, so it should be `60`

virajjasani:
```
    ShutdownOnUnusedThreadPoolExecutor pool =
      new ShutdownOnUnusedThreadPoolExecutor(maxThreads, maxThreads, keepAliveTime,
        TimeUnit.SECONDS, new LinkedBlockingQueue<Runnable>(),
        new ThreadFactoryBuilder().setNameFormat(builder.getName() + "-pool-%d")
          .setUncaughtExceptionHandler(Threads.LOGGING_EXCEPTION_HANDLER).build(),
        builder.getName());
```

phoenix-core-client/src/main/java/org/apache/phoenix/query/QueryServices.java
```diff
@@ -627,6 +627,10 @@ public interface QueryServices extends SQLCloseable {
   String PHOENIX_CDC_STREAM_PARTITION_EXPIRY_MIN_AGE_MS =
     "phoenix.cdc.stream.partition.expiry.min.age.ms";
 
+  String PHOENIX_UNCOVERED_INDEX_MAX_POOL_SIZE = "phoenix.uncovered.index.threads.max";
+  String PHOENIX_UNCOVERED_INDEX_KEEP_ALIVE_TIME_MS =
+    "phoenix.uncovered.index.threads.keepalive.ms";
```
virajjasani: With the same change, let's also convert this to `phoenix.uncovered.index.threads.keepalive.sec`

virajjasani:
We also need to co-ordinate with this PR https://github.com/apache/phoenix/pull/2273/files#diff-42d39a0ca68a86ad3d86a6d3bad511af8a273a520986ce269ab0e920da6596a7R58

It is also introducing stop() in regionserver coproc

cc @jpisaac

phoenix-core-client/src/main/java/org/apache/phoenix/query/QueryServices.java
palashc: done

phoenix-core-client/src/main/java/org/apache/phoenix/query/QueryServicesOptions.java
palashc: good catch, done

PR #2255, opened 2025-07-31
PHOENIX-7684 Introduce Segment Scan

virajjasani:
Jira: PHOENIX-7684

HBase stores rows of data in tables. Tables are split into groups of lexicographically adjacent rows. These groups are called regions. By lexicographically adjacent, all rows in the table that sort between the region’s start row key and end row key are stored in the same region. Large table with multi tera bytes or peta bytes of data can have their data spread across hundreds of thousands of regions.

Depending on the size of the table, full table scan can take from few seconds to several hours. When client executes "SELECT * FROM <table>" query using Phoenix JDBC client, single client is responsible for retrieving all the rows from all the regions of the given table. Although, single Phoenix client does divide the scan range into the table region ranges and submit the scans in parallel. Single client application can still become bottleneck for the end user when the table size is very large, as the single client application always have limited memory, CPU and IO capabilities.

The purpose of this Jira is to introduce segment scan. The segment is logical chunk of the given table, similar to HBase table region. While scanning the full table, the client application does not have any insight into how the table data are distributed among the regions. The concept of segment scan allows application to define the number of segments into which the table data can be divided. A new function TOTAL_SEGMENTS() can be used to retrieve the scan boundary of each segment. The retrieved scan boundaries can be provided to individual client worker (thread or VM) so that a given worker only performs scan of a segment of the table. Additional functions SCAN_START_KEY() and SCAN_END_KEY() are used to retrieve the segment boundaries and then later on use the boundaries to submit the segment scan request.

```
SELECT SCAN_START_KEY(), SCAN_END_KEY() FROM T1 WHERE TOTAL_SEGMENTS() = 10 
```
The above SELECT query is meant to retrieve total 10 segment boundaries from the table T1 by bucketing the table region boundaries into 10 ranges.

e.g.

- If the table has 12 regions and client needs 4 segments: each segment will contain 3 regions
- If the table has 10 regions and client needs 3 segments: you get segments of sizes 4, 3, and 3 regions
- If the table has 3 regions and client needs 10 segments: you get 3 segments (one per region)

One of the advantages of the segment scan: External frameworks like Spark, Trino or MapReduce can now ask for "give me exactly 100 work units" instead of dealing with an unknown number of HBase regions. The segment scan approach provides horizontal scaling of the client workers.

After retrieving the above segment boundaries, any of the segment can be scanned using
```
SELECT * FROM <table> WHERE SCAN_START_KEY() = ? AND SCAN_END_KEY() = ?
```
The above query takes param values of SCAN_START_KEY() and SCAN_END_KEY() as the VARBINARY values earlier retrieved from the first query. With the scan boundaries as the given segment boundary, only the specific segment worth of data will be scanned by the query.

Overall, this feature provides deterministic data partitioning for parallel processing, making Phoenix more suitable for integration with modern big data processing frameworks.

### DIFF
--- a/phoenix-core-client/src/main/java/org/apache/phoenix/compile/WhereCompiler.java
+++ b/phoenix-core-client/src/main/java/org/apache/phoenix/compile/WhereCompiler.java
@@ -38,6 +38,7 @@ import java.util.Set;
 import org.apache.hadoop.hbase.CompareOperator;
 import org.apache.hadoop.hbase.client.Scan;
 import org.apache.hadoop.hbase.filter.Filter;
+import org.apache.hadoop.hbase.io.ImmutableBytesWritable;
 import org.apache.hadoop.hbase.util.Bytes;
 import org.apache.hadoop.io.WritableUtils;
 import org.apache.phoenix.coprocessorclient.BaseScannerRegionObserverConstants;
@@ -75,7 +76,11 @@ import org.apache.phoenix.expression.SubtractExpression;
 import org.apache.phoenix.expression.function.ArrayAnyComparisonExpression;
 import org.apache.phoenix.expression.function.ArrayElemRefExpression;
 import org.apache.phoenix.expression.function.ScalarFunction;
+import org.apache.phoenix.expression.function.ScanEndKeyFunction;
+import org.apache.phoenix.expression.function.ScanStartKeyFunction;
 import org.apache.phoenix.expression.function.SingleAggregateFunction;
+import org.apache.phoenix.expression.function.TotalSegmentsFunction;
+import org.apache.phoenix.expression.visitor.CloneExpressionVisitor;
 import org.apache.phoenix.expression.visitor.KeyValueExpressionVisitor;
 import org.apache.phoenix.expression.visitor.TraverseAllExpressionVisitor;
 import org.apache.phoenix.filter.MultiCFCQKeyValueComparisonFilter;
@@ -85,6 +90,7 @@ import org.apache.phoenix.filter.RowKeyComparisonFilter;
 import org.apache.phoenix.filter.SingleCFCQKeyValueComparisonFilter;
 import org.apache.phoenix.filter.SingleCQKeyValueComparisonFilter;
 import org.apache.phoenix.parse.ColumnParseNode;
+import org.apache.phoenix.parse.ComparisonParseNode;
 import org.apache.phoenix.parse.FilterableStatement;
 import org.apache.phoenix.parse.HintNode;
 import org.apache.phoenix.parse.ParseNode;
@@ -188,10 +194,25 @@ public class WhereCompiler {
     }
 
     Set<Expression> extractedNodes = Sets.<Expression> newHashSet();
-    WhereExpressionCompiler whereCompiler = new WhereExpressionCompiler(context);
+    ScanBoundaryExtractingCompiler whereCompiler = new ScanBoundaryExtractingCompiler(context);
     Expression expression = where == null
       ? LiteralExpression.newConstant(true, PBoolean.INSTANCE, Determinism.ALWAYS)
       : where.accept(whereCompiler);
+
+    if (whereCompiler.getScanStartKey() != null || whereCompiler.getScanEndKey() != null) {
+      // Remove scan boundary functions from the expression
+      ScanBoundaryRemovalVisitor removalVisitor = new ScanBoundaryRemovalVisitor();
+      expression = expression.accept(removalVisitor);
+    }
+
+    // Check for TOTAL_SEGMENTS function
+    if (whereCompiler.hasTotalSegments()) {
+      context.setTotalSegmentsFunction(true);
+      context.setTotalSegmentsValue(whereCompiler.getTotalSegmentsValue());
+      // Remove TOTAL_SEGMENTS function from the expression
+      ScanBoundaryRemovalVisitor removalVisitor = new ScanBoundaryRemovalVisitor();
+      expression = expression.accept(removalVisitor);
+    }
     if (whereCompiler.isAggregate()) {
       throw new SQLExceptionInfo.Builder(SQLExceptionCode.AGGREGATE_IN_WHERE).build()
         .buildException();
@@ -222,13 +243,22 @@ public class WhereCompiler {
       expression = WhereOptimizer.pushKeyExpressionsToScan(context, hints, expression,
         extractedNodes, minOffset);
     }
+    if (whereCompiler.getScanStartKey() != null || whereCompiler.getScanEndKey() != null) {
+      Scan scan = context.getScan();
+      if (scan.getStartRow().length == 0 && whereCompiler.getScanStartKey() != null) {
+        scan.withStartRow(whereCompiler.getScanStartKey());
+      }
+      if (scan.getStopRow().length == 0 && whereCompiler.getScanEndKey() != null) {
+        scan.withStopRow(whereCompiler.getScanEndKey());
+      }
+    }
     setScanFilter(context, statement, expression, whereCompiler.disambiguateWithFamily);
 
     return expression;
   }
 
   public static class WhereExpressionCompiler extends ExpressionCompiler {
-    private boolean disambiguateWithFamily;
+    protected boolean disambiguateWithFamily;
 
     public WhereExpressionCompiler(StatementContext context) {
       super(context, true);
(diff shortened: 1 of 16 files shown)


phoenix-core-client/src/main/java/org/apache/phoenix/iterate/SegmentResultIterator.java
```diff
@@ -0,0 +1,124 @@
...
+
+  public static List<Segment> getSegments(List<HRegionLocation> regions, int numBuckets) {
+    int size = regions.size();
+    List<Segment> segments = new ArrayList<>();
+    if (size <= numBuckets) {
+      for (HRegionLocation region : regions) {
+        segments.add(new Segment(region.getRegion().getStartKey(), region.getRegion().getEndKey()));
+      }
+    } else {
+      int q = size / numBuckets;
```
palashc, line 98: Where do we handle 0 or negative values for TOTAL_SEGMENTS?

phoenix-core/src/it/java/org/apache/phoenix/end2end/ScanBoundaryFunction2IT.java
```diff
@@ -0,0 +1,310 @@
...
+          assertEquals(8, results.size());
+          assertEquals("2:KEY_008", results.get(0));
+          assertEquals("2:KEY_029", results.get(7));
+        }
+      }
+    }
+  }
+
+  @Test
+  public void testScanEndWithPkFilters2() throws Exception {
```
palashc, line 238: nit: Should we also add test with both start and end key, along with the PK filters?

phoenix-core/src/it/java/org/apache/phoenix/end2end/ScanBoundaryFunctionIT.java
```diff
@@ -0,0 +1,584 @@
...
+    }
+  }
+
+  @Test
+  public void testScanBoundariesMixedWithOrConditionShouldWork() throws Exception {
+    Properties props = PropertiesUtil.deepCopy(TestUtil.TEST_PROPERTIES);
+    try (Connection conn = DriverManager.getConnection(getUrl(), props)) {
+      // Test scan boundaries combined with other OR conditions - this should work
+      // The scan boundaries should be applied, and the OR condition should be evaluated as a filter
+      String sql = "SELECT PK, COL1, COL2 FROM " + fullTableName
```
palashc, line 540: nit: we can add tests where query has SCAN_END_KEY() before SCAN_START_KEY()

phoenix-core/src/it/java/org/apache/phoenix/end2end/ScanBoundaryFunctionIT.java
```diff
@@ -0,0 +1,584 @@
...
+import org.apache.phoenix.compile.QueryPlan;
+import org.apache.phoenix.jdbc.PhoenixPreparedStatement;
+import org.apache.phoenix.util.PropertiesUtil;
+import org.apache.phoenix.util.TestUtil;
+import org.junit.Before;
+import org.junit.Test;
+import org.junit.experimental.categories.Category;
+
+@Category(ParallelStatsDisabledTest.class)
+public class ScanBoundaryFunctionIT extends ParallelStatsDisabledIT {
```
palashc, line 43: Should we also add tests for VARBINARY/VARBINARY_ENCODED data types?

phoenix-core/src/test/java/org/apache/phoenix/compile/ScanBoundaryFunctionTest.java
```diff
@@ -0,0 +1,217 @@
...
+    assertArrayEquals(Bytes.toBytes("bindend"), scan.getStopRow());
+    assertArrayEquals(new byte[0], scan.getStartRow());
+  }
+
+  @Test
+  public void testScanEndKeyWithBindParameter2() throws Exception {
+    Properties props = new Properties();
+    PhoenixConnection conn = (PhoenixConnection) DriverManager.getConnection(getUrl(), props);
+
+    String sql = "SELECT * FROM SYSTEM.CATALOG WHERE SCAN_START_KEY() = ? AND SCAN_END_KEY() = ?";
```
palashc, line 149: nit: we can add test with SCAN_END_KEY() before SCAN_START_KEY()

phoenix-core/src/it/java/org/apache/phoenix/end2end/TotalSegmentsFunctionIT.java
```diff
@@ -0,0 +1,752 @@
...
+import org.apache.phoenix.query.ConnectionQueryServices;
+import org.apache.phoenix.util.PropertiesUtil;
+import org.apache.phoenix.util.TestUtil;
+import org.junit.Before;
+import org.junit.Test;
+
+/**
+ * Tests for TOTAL_SEGMENTS()
+ */
+public class TotalSegmentsFunctionIT extends ParallelStatsDisabledIT {
```
palashc: We should add tests where TOTAL_SEGMENTS is provided as 0 or negative value. Any invalid data type will be handled during parsing I guess?

phoenix-core/src/it/java/org/apache/phoenix/end2end/ScanBoundaryFunction2IT.java
virajjasani: Ideally that should not be used as it's conflicting. The primary purpose of Segment scan is full table scan with or without non-pk column filters

phoenix-core/src/it/java/org/apache/phoenix/end2end/ScanBoundaryFunctionIT.java
virajjasani: Sounds good

phoenix-core/src/it/java/org/apache/phoenix/end2end/TotalSegmentsFunctionIT.java
virajjasani: Yeah we should throw error if value is <= 0

phoenix-core/src/it/java/org/apache/phoenix/end2end/ScanBoundaryFunction2IT.java
palashc: Okay that makes sense.

phoenix-core/src/test/java/org/apache/phoenix/compile/ScanBoundaryFunctionTest.java
virajjasani: Covered with `ScanBoundaryFunctionIT`, here we can't verify result anyways

phoenix-core/src/it/java/org/apache/phoenix/end2end/ScanBoundaryFunctionIT.java
virajjasani: Done

phoenix-core-client/src/main/java/org/apache/phoenix/iterate/SegmentResultIterator.java
virajjasani: This is covered in WhereCompiler to fail-fast

phoenix-core/src/it/java/org/apache/phoenix/end2end/ScanBoundaryFunctionIT.java
palashc: @virajjasani can you point me to the test for varbinary pk where scan boundary functions are used in the query?

phoenix-core-client/src/main/java/org/apache/phoenix/compile/WhereCompiler.java
```diff
@@ -294,6 +325,152 @@ protected ColumnRef resolveColumn(ColumnParseNode node) throws SQLException {
...
+      }
+      return null;
+    }
+
+    private Integer getTotalSegmentsVal(LiteralExpression literal) throws SQLException {
+      ImmutableBytesWritable ptr = new ImmutableBytesWritable();
+      if (literal.evaluate(null, ptr)) {
+        Integer value = (Integer) PInteger.INSTANCE.toObject(ptr);
+        if (value != null && value <= 0) {
+          throw new SQLExceptionInfo.Builder(SQLExceptionCode.INVALID_TOTAL_SEGMENTS_VALUE).build()
```
palashc, line 396: Do we have a test for this?

phoenix-core-client/src/main/java/org/apache/phoenix/iterate/SegmentResultIterator.java
palashc: Do we have a test for this?

phoenix-core/src/test/java/org/apache/phoenix/compile/ScanBoundaryFunctionTest.java
palashc: @virajjasani what I meant was `SELECT * FROM TABLE WHERE SCAN_END_KEY() = ? AND SCAN_START_KEY() = ?` just to test a different placement of the clauses in the query.

phoenix-core/src/it/java/org/apache/phoenix/end2end/ScanBoundaryFunctionIT.java
palashc: I see it now

phoenix-core-client/src/main/java/org/apache/phoenix/compile/WhereCompiler.java
virajjasani: Yes, we do in `testTotalSegmentsWithInvalidValues()`

phoenix-core/src/test/java/org/apache/phoenix/compile/ScanBoundaryFunctionTest.java
virajjasani: I see, the ordering of the functions is not a big deal because ANTLR grammar is known to take care of it. However, using inverted scan boundaries is good test case

virajjasani:
> checkstyle/spotless fixes etc

It's same spotless that keeps showing up on multiple PRs for one of the javadocs in MetaDataClient. Given that this is big PR already, let's take care of spotless in separate commit

PR #2273, opened 2025-08-11
PHOENIX-7638 Creating a large number of views leads to OS thread exhaustion

jpisaac:


phoenix-core-server/src/main/java/org/apache/phoenix/coprocessor/PhoenixRegionServerEndpoint.java
```diff
@@ -54,6 +55,12 @@ public void start(CoprocessorEnvironment env) throws IOException {
       MetricsPhoenixCoprocessorSourceFactory.getInstance().getMetadataCachingSource();
   }
 
+  @Override
+  public void stop(CoprocessorEnvironment env) throws IOException {
```
palashc, line 69: @jpisaac FYI I also added the stop() hook in this endpoint as part of this commit https://github.com/apache/phoenix/commit/20e1f62e52df8de3f5ebdc940479aeae393c1def.

virajjasani:
@jpisaac could you please rebase with master one time? just one final build and we should be good to go

PR #2274, opened 2025-08-14
PHOENIX-7566 HAGroupState subscription feature and state management for ReplicationLogReader

ritegarg:


phoenix-core-client/src/main/java/org/apache/phoenix/jdbc/HAGroupStoreClient.java
```diff
@@ -293,10 +308,29 @@ public void setHAGroupStatusIfNeeded(HAGroupStoreRecord.HAGroupState haGroupStat
...
+                // NOT ACTIVE or ACTIVE_TO_STANDBY
+                // set the time to null to mark that we are current(or we don't have any reader).
+                // TODO: Verify that for reader this is the correct approach.
+                Long lastSyncTimeInMs = currentHAGroupStoreRecord
+                        .getLastSyncStateTimeInMs();
+                ClusterRole clusterRole = haGroupState.getClusterRole();
+                if (currentHAGroupStoreRecord.getHAGroupState()
+                        == HAGroupStoreRecord.HAGroupState.ACTIVE_IN_SYNC
+                        && haGroupState == HAGroupStoreRecord.HAGroupState.ACTIVE_NOT_IN_SYNC) {
+                    lastSyncTimeInMs = System.currentTimeMillis();
```
kadirozde, line 320: This should be a filed in the HA Store record in Zookeeper. This field is updated when the cluster switches from ACTIVE_IN_SYNC to ACTIVE_NOT_IN_SYNC. When a region server of the standby cluster starts and the active cluster is in ACTIVE_NOT_IN_SYNC, the region server needs to know lastSyncTimeInMs for establishing the consistency point for the standby cluster.

phoenix-core-client/src/main/java/org/apache/phoenix/jdbc/HAGroupStoreClient.java
```diff
@@ -638,4 +692,189 @@ private boolean isUpdateNeeded(HAGroupStoreRecord.HAGroupState currentHAGroupSta
...
+
+    /**
+     * Subscribe to be notified when a specific state transition occurs.
+     *
+     * @param fromState the state to transition from
+     * @param toState the state to transition to
+     * @param clusterType whether to monitor local or peer cluster
+     * @param listener the listener to notify when the transition occurs
+     */
+    public void subscribeToTransition(HAGroupState fromState, HAGroupState toState,
```
kadirozde: I feel that we do not need to complicate matters by having subscriptions at the state transition level. I think subscriptions at <toState, clusterType> should be good enough. A subscriber does not need to know the details of the state transitions as it mostly cares about the target state. Also, subscription at the transition level creates coupling between subscribers and the HA state machine, which is not necessary.

phoenix-core-client/src/main/java/org/apache/phoenix/jdbc/HAGroupStoreClient.java
ritegarg: Yes, I have added that new field 
https://github.com/apache/phoenix/pull/2274/files#diff-8d6c9a772a5a7947b4f4f3c98f460924658742acf4bbb22d53f45c6d241f3cdaR166
It is updated here https://github.com/apache/phoenix/pull/2274/files#diff-61fd251b01270c601f6cc2851770e91d36d12eafec0aa709782649be2a286cfbR340

phoenix-core-client/src/main/java/org/apache/phoenix/jdbc/HAGroupStoreClient.java
ritegarg: Removed specific transition methods and also updated test cases

kadirozde (approved):
+1, Thanks!

PR #2278, opened 2025-08-26
PHOENIX-7568 - Adding Replication Log Replay Implementation

Himanshu-g81:
High Level Description of major Replication Replay Componenets added in this PR

1. **ReplicationLogReplayService** - A singleton class that has single thread which gets all the HA groups and start Replication Replay for each using ReplicationReplay.get(conf, replicationGroup).startReplay(); every 60 seconds (configurable). Note that startReplay() of ReplicationReplay is idempotent. This is hooked into RS start / stop path in PhoenixRegionServerEndpoint.java

2. **ReplicationLogReplay** - Responsible for handling replication replay lifecycle for single HA Group. It initialize the file system from which replay needs to be done for this HA Group. The init method also initialize ReplicationReplayLogDiscovery (and respective ReplicationLogFileTrackerReplay for the group) - Responsibilites of these 2 componenets are described below.

3. **ReplicationLogFileTracker** - Abstract class to deal with all file system interactions (getNewFiles, markInProgress, markCompleted, markInProgress). It has one implementation currently for standby cluster (ReplicationLogReplayFileTracker) overriding directory (IN directory) and metric source. Similar implementation can be added for store and forward mode (with OUT as directory) and custom metric source.

4. **ReplicationLogDiscovery** - Abstract class responsible for logic of processing files round by round. It contains ReplicationLogFileTracker. It creates a thread pool (with all properties configurable, i.e. thread count, scheduling interval, etc) to process the log files round by round ([details](https://docs.google.com/document/d/1usap8PCYFU0Z4orznUPvk0tnSv0X-vnbgD_QZejrnv0/edit?tab=t.0#bookmark=id.kahc07qfainp) - this is salesforce internal doc, will update once design is published). The process method is abstract and implemented for standby cluster replay (as ReplicationReplayLogDiscovery) to apply mutations on target. Similar implementation can be added for store and forward mode - to just copy file to standby cluster.

5. **ReplicationShardDirectoryManager** - Encapsulates the logic of shard directory management (for both active and DR cluster). Only root directory needs to be given during initialization. Changes to leverage it in source are also part of this PR.

**Simplified Sequence Diagram**
sequence_diagram

### DIFF
--- a/phoenix-core-server/src/main/java/org/apache/phoenix/replication/ReplicationLogDiscovery.java
+++ b/phoenix-core-server/src/main/java/org/apache/phoenix/replication/ReplicationLogDiscovery.java
@@ -144,7 +144,7 @@ public abstract class ReplicationLogDiscovery {
             scheduler.scheduleAtFixedRate(() -> {
                 try {
                     replay();
-                } catch (IOException e) {
+                } catch (Exception e) {
                     LOG.error("Error during replay", e);
                 }
             }, 0, getReplayIntervalSeconds(), TimeUnit.SECONDS);
--- a/phoenix-core-server/src/main/java/org/apache/phoenix/replication/ReplicationLogTracker.java
+++ b/phoenix-core-server/src/main/java/org/apache/phoenix/replication/ReplicationLogTracker.java
@@ -98,10 +98,6 @@ public class ReplicationLogTracker {
      * exist.
      */
     public void init() throws IOException {
-//        Path newFilesDirectory = new Path(new Path(rootURI.getPath(), haGroupName),
-//            getNewLogSubDirectoryName());
-//        this.replicationShardDirectoryManager = new ReplicationShardDirectoryManager(conf,
-//            newFilesDirectory);
         this.inProgressDirPath = new Path(getReplicationShardDirectoryManager().getRootDirectoryPath().getParent(),
             getInProgressLogSubDirectoryName());
         createDirectoryIfNotExists(inProgressDirPath);
--- a/phoenix-core/src/test/java/org/apache/phoenix/replication/ReplicationLogDiscoveryTest.java
+++ b/phoenix-core/src/test/java/org/apache/phoenix/replication/ReplicationLogDiscoveryTest.java
@@ -69,7 +69,6 @@ public class ReplicationLogDiscoveryTest {
     private TestableReplicationLogTracker fileTracker;
     private Configuration conf;
     private FileSystem localFs;
-    private URI rootURI;
     private Path testFolderPath;
     private static final String haGroupName = "testGroup";
     private static final MetricsReplicationLogTracker metricsLogTracker = new MetricsReplicationLogTrackerReplayImpl(haGroupName);
@@ -79,7 +78,7 @@ public class ReplicationLogDiscoveryTest {
     public void setUp() throws IOException {
         conf = HBaseConfiguration.create();
         localFs = FileSystem.getLocal(conf);
-        rootURI = new Path(testFolder.getRoot().toString()).toUri();
+        URI rootURI = new Path(testFolder.getRoot().toString()).toUri();
         testFolderPath = new Path(testFolder.getRoot().getAbsolutePath());
         Path newFilesDirectory = new Path(new Path(rootURI.getPath(), haGroupName), ReplicationLogReplay.IN_DIRECTORY_NAME);
         ReplicationShardDirectoryManager replicationShardDirectoryManager =
@@ -939,9 +938,10 @@ public class ReplicationLogDiscoveryTest {
         allInProgressFiles.addAll(inProgressFiles0004);
         allInProgressFiles.addAll(inProgressFiles0102);
 
-        // Mock processFile to throw exception for specific files (files 1 and 3) - using prefix matching
+        // Mock processFile to throw exception for specific files (files 1 and 3) only on first call, succeed on retry
         String file1Prefix = allInProgressFiles.get(1).getName().substring(0, allInProgressFiles.get(1).getName().lastIndexOf("_"));
         Mockito.doThrow(new IOException("Processing failed for file 1"))
+            .doCallRealMethod()
             .when(discovery).processFile(Mockito.argThat(path -> {
                 String pathName = path.getName();
                 String withoutExtension = pathName.substring(0, pathName.lastIndexOf("."));
@@ -950,6 +950,7 @@ public class ReplicationLogDiscoveryTest {
             }));
         String file3Prefix = allInProgressFiles.get(3).getName().substring(0, allInProgressFiles.get(3).getName().lastIndexOf("_"));
         Mockito.doThrow(new IOException("Processing failed for file 3"))
+                .doCallRealMethod()
                 .when(discovery).processFile(Mockito.argThat(path -> {
                     String pathName = path.getName();
                     String withoutExtension = pathName.substring(0, pathName.lastIndexOf("."));
@@ -960,13 +961,16 @@ public class ReplicationLogDiscoveryTest {
         // Process in-progress directory
         discovery.processInProgressDirectory();
 
-        // Verify that markInProgress was called 5 times (for all in-progress files initially)
-        Mockito.verify(fileTracker, Mockito.times(5)).markInProgress(Mockito.any(Path.class));
+        // Verify that markInProgress was called 7 times (5 initially + 2 for retries)
+        Mockito.verify(fileTracker, Mockito.times(7)).markInProgress(Mockito.any(Path.class));
         
         // Verify that markInProgress was called for each expected file
-        for (Path expectedFile : allInProgressFiles) {
+        // Files 1 and 3 are called twice (initial attempt + retry), others once
+        for (int i = 0; i < allInProgressFiles.size(); i++) {
+            Path expectedFile = allInProgressFiles.get(i);
             String expectedPrefix = expectedFile.getName().substring(0, expectedFile.getName().lastIndexOf("_"));
-            Mockito.verify(fileTracker, Mockito.times(1)).markInProgress(
+            int expectedTimes = (i == 1 || i == 3) ? 2 : 1; // Files 1 and 3 are retried
+            Mockito.verify(fileTracker, Mockito.times(expectedTimes)).markInProgress(
                     Mockito.argThat(path -> {
                         String pathName = path.getName();
                         String prefix = pathName.substring(0, pathName.lastIndexOf("_"));
@@ -978,9 +982,12 @@ public class ReplicationLogDiscoveryTest {
         Mockito.verify(discovery, Mockito.times(7)).processFile(Mockito.any(Path.class));
 
         // Verify that processFile was called for each specific file (using prefix matching)
-        for (Path expectedFile : allInProgressFiles) {
+        // Files 1 and 3 should be called twice (fail once, succeed on retry), others once
+        for (int i = 0; i < allInProgressFiles.size(); i++) {
+            Path expectedFile = allInProgressFiles.get(i);
             String expectedPrefix = expectedFile.getName().substring(0, expectedFile.getName().lastIndexOf("_"));
-            Mockito.verify(discovery, Mockito.times(1)).processFile(
+            int expectedTimes = (i == 1 || i == 3) ? 2 : 1; // Files 1 and 3 are called twice (fail + retry success)
+            Mockito.verify(discovery, Mockito.times(expectedTimes)).processFile(
                 Mockito.argThat(path -> {
                     String pathName = path.getName();
                     String withoutExtension = pathName.substring(0, pathName.lastIndexOf("."));
--- a/phoenix-core-server/src/main/java/org/apache/phoenix/replication/reader/ReplicationLogDiscoveryReplay.java
+++ b/phoenix-core-server/src/main/java/org/apache/phoenix/replication/reader/ReplicationLogDiscoveryReplay.java
@@ -459,7 +459,9 @@ public class ReplicationLogDiscoveryReplay extends ReplicationLogDiscovery {
 
     protected boolean shouldTriggerFailover() throws IOException {
         return failoverPending.get() && lastRoundInSync.equals(lastRoundProcessed)
-                && replicationLogTracker.getInProgressFiles().isEmpty();
+                && replicationLogTracker.getInProgressFiles().isEmpty()
+                ;
+        // TODO: Check for in files of lastRoundProcessed, lastRoundProcessed + roundTime as well - because there can be new files for upcoming round
     }
 
     protected void triggerFailover() throws IOException, InvalidClusterRoleTransitionException {
--- a/phoenix-core/src/test/java/org/apache/phoenix/replication/reader/ReplicationLogReplayTest.java
+++ b/phoenix-core/src/test/java/org/apache/phoenix/replication/reader/ReplicationLogReplayTest.java
@@ -17,6 +17,9 @@
  */
 package org.apache.phoenix.replication.reader;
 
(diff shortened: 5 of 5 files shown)


apurtell:
Looks like there are lots of blank, checkstyle, and spotbugs issues. Can we fix those first?

phoenix-core/src/test/java/org/apache/phoenix/replication/reader/ReplicationLogDiscoveryReplayTest.java
```diff
@@ -0,0 +1,1175 @@
...
+        }
+        
+        @Override
+        protected HAGroupStoreRecord getHAGroupRecord() {
+            return haGroupStoreRecord;
+        }
+        
+        @Override
+        protected void processRound(ReplicationRound replicationRound) throws IOException {
+            System.out.println("Processing Round: " + replicationRound);
```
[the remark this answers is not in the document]
apurtell: Definitely don't do this.

phoenix-core/src/test/java/org/apache/phoenix/replication/reader/ReplicationLogDiscoveryReplayTest.java
```diff
@@ -0,0 +1,1175 @@
...
+            // Track processed rounds
+            processedRounds.add(replicationRound);
+            
+            // Simulate state change by listener after certain number of rounds
+            roundsProcessed++;
+            if (stateChangeAfterRounds > 0 && roundsProcessed == stateChangeAfterRounds && newStateAfterRounds != null) {
+                System.out.println("Rounds Processed: " + roundsProcessed + " - " + newStateAfterRounds);
```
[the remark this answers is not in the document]
apurtell: Same.

phoenix-core/src/test/java/org/apache/phoenix/replication/ReplicationLogDiscoveryTest.java
```diff
@@ -0,0 +1,1467 @@
...
+
+        // 7. Ensure only current round new files (3) are processed (Total 3, no in-progress files)
+        List<Path> processedFiles = discovery.getProcessedFiles();
+        assertEquals("Invalid number of files processed", 3, processedFiles.size());
+
+        System.out.println("Processed files");
+        for (Path file : processedFiles) {
+            System.out.println(file);
+        }
+
```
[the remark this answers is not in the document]
apurtell: Same

apurtell (changes requested):
Address checkstyle, spotbugs, and copilot findings first.

phoenix-core-server/src/main/java/org/apache/phoenix/replication/StoreAndForwardLogGroupWriter.java
```diff
@@ -68,6 +68,11 @@ protected void initializeFileSystems() throws IOException {
         // TODO
     }
 
+    @Override
+    protected void initializeReplicationShardDirectoryManager() {
+        // TODO
```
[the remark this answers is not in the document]
Himanshu-g81: To be done via Store & Forward mode implementation.

Himanshu-g81:
> Address checkstyle, spotbugs, and copilot findings, please.

Sure, addressed all of those (except false positives of spotbugs).

phoenix-core-server/src/main/java/org/apache/phoenix/replication/ReplicationLogTracker.java
```diff
@@ -0,0 +1,518 @@
...
+        return this.metrics;
+    }
+
+    protected String getInProgressLogSubDirectoryName() {
+        return getNewLogSubDirectoryName() + "_progress";
+    }
+
+    /**
+     * Initializes the file tracker by setting up new files directory and in-progress directory.
+     * Creates the in-progress directory (rootURI/<group-name>/[in/out]_progress) if it doesn't
```
tkhurana, line 97: In ReplicationShardDirectoryManager the format specified is `rootURI/group-name/in/shard/000` or `/shard/001` but I don't see reference to `shard` in here

tkhurana: I see, you are constructing the ReplicationShardDirectoryManager which will internally append the `shard` to the path.

tkhurana: I would suggest that ReplicationLogTracker should take this path as input `rootURI/groupName/in` or `rootURI/groupName/out` as input. You don't need that enum DirectoryType. The ReplicationLogTracker doesn't need to know about it. It should take a path as input and just work with it.

phoenix-core-server/src/main/java/org/apache/phoenix/replication/ReplicationLogTracker.java
```diff
@@ -0,0 +1,518 @@
...
+        private final String name;
+
+        DirectoryType(final String name) {
+            this.name = name;
+        }
+
+        public String getName() {
+            return this.name;
+        }
+    }
```
tkhurana: I think we should get rid of this enum. It is not really an enum. What if we use a different terminology. These are just string constants but they don't need to be defined here but at a higher level. The ReplicationLogTracker doesn't need to know about it. It should take a path as input and just work with it.

phoenix-core-server/src/main/java/org/apache/phoenix/replication/ReplicationLogTracker.java
tkhurana: Whatever path is given as input to you, you can create a new progress directory as `<input-path>_progress`

tkhurana: Ideally, you should take the `ReplicationShardDirectoryManager` as an input. The `ReplicationLogGroupWriter` already has a shard directory manager object. When we are in StoreAndForward mode , we can simply use that to create a ReplicationLogTracker. Any information on paths can be fetched from the shard directory manager. This way all path construction logic lives inside the shard directory manager.

phoenix-core-server/src/main/java/org/apache/phoenix/replication/metrics/MetricsReplicationLogDiscovery.java
```diff
@@ -0,0 +1,75 @@
...
+package org.apache.phoenix.replication.metrics;
+
+import org.apache.hadoop.hbase.metrics.BaseSource;
+
+public interface MetricsReplicationLogDiscovery extends BaseSource {
+
+    String METRICS_NAME = "MetricsReplicationLogDiscovery";
+    String METRICS_CONTEXT = "phoenix";
+    String METRICS_DESCRIPTION = "Metrics about Replication Log File Discovery for an HA Group";
+    String METRICS_JMX_CONTEXT = "RegionServer,sub=" + METRICS_NAME;
```
tkhurana, line 27: I don't think this should be defined in the interface. Concrete implementations will define it anyways.

tkhurana: Never mind, I see this is a pattern that is being followed else where to.

phoenix-core-server/src/main/java/org/apache/phoenix/replication/metrics/MetricsReplicationLogDiscoveryReplayImpl.java
```diff
@@ -0,0 +1,39 @@
...
+        "Metrics about Replication Replay Log Discovery for an HA Group";
+    private static final String METRICS_JMX_CONTEXT = "RegionServer,sub=" + METRICS_NAME;
+
+    public MetricsReplicationLogDiscoveryReplayImpl(final String haGroupName) {
+        super(MetricsReplicationLogDiscoveryReplayImpl.METRICS_NAME,
+            MetricsReplicationLogDiscoveryReplayImpl.METRICS_DESCRIPTION,
+            MetricsReplicationLogDiscoveryImpl.METRICS_CONTEXT,
+            MetricsReplicationLogDiscoveryReplayImpl.METRICS_JMX_CONTEXT
+                    + ",haGroup=" + haGroupName);
+        super.groupMetricsContext =
```
tkhurana: Can this also be moved to the constructor instead of a standalone call ?

phoenix-core-server/src/main/java/org/apache/phoenix/replication/metrics/MetricsReplicationLogTrackerReplayImpl.java
```diff
@@ -0,0 +1,39 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.phoenix.replication.metrics;
+
+public class MetricsReplicationLogTrackerReplayImpl
```
tkhurana, line 20: Can you rename this class to `MetricsReplicationLogReplayTrackerImpl`

phoenix-core-server/src/main/java/org/apache/phoenix/replication/metrics/MetricsReplicationLogDiscoveryReplayImpl.java
```diff
@@ -0,0 +1,39 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.phoenix.replication.metrics;
+
+/** Implementation of metrics source for ReplicationLogDiscoveryReplay operations. */
+public class MetricsReplicationLogDiscoveryReplayImpl
+    extends MetricsReplicationLogDiscoveryImpl {
+
+    private static final String METRICS_NAME = "ReplicationLogDiscoveryReplay";
```
tkhurana, line 24: also rename this to ReplicationLogReplayDiscovery

phoenix-core-server/src/main/java/org/apache/phoenix/replication/metrics/MetricsReplicationLogDiscoveryReplayImpl.java
```diff
@@ -0,0 +1,39 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.phoenix.replication.metrics;
+
+/** Implementation of metrics source for ReplicationLogDiscoveryReplay operations. */
+public class MetricsReplicationLogDiscoveryReplayImpl
```
tkhurana, line 21: rename this class to MetricsReplicationLogReplayDiscoveryImpl

phoenix-core-server/src/main/java/org/apache/phoenix/replication/metrics/MetricsReplicationLogTrackerReplayImpl.java
tkhurana: The thought is that I can use `MetricsReplicationLogForwarderTrackerImpl`

phoenix-core-server/src/main/java/org/apache/phoenix/replication/metrics/MetricsReplicationLogDiscoveryReplayImpl.java
tkhurana: Similarly, I can use `MetricsReplicationLogForwarderDiscoveryImpl`

phoenix-core-server/src/main/java/org/apache/phoenix/replication/metrics/MetricsReplicationLogTrackerReplayImpl.java
tkhurana: Actually, let's keep it this way.

phoenix-core-server/src/main/java/org/apache/phoenix/replication/metrics/MetricsReplicationLogDiscoveryReplayImpl.java
tkhurana: Ignore. Keep it the same.

phoenix-core-server/src/main/java/org/apache/phoenix/replication/metrics/MetricsReplicationLogDiscoveryReplayImpl.java
tkhurana: Ignore, keep it the same

phoenix-core-server/src/main/java/org/apache/phoenix/replication/ReplicationLogDiscovery.java
```diff
@@ -0,0 +1,501 @@
...
+     * getReplicationRoundFromEndTime(),
+     * which rounds down to the nearest round boundary to ensure we start from a complete round.
+     *
+     * @throws IOException if there's an error reading file timestamps
+     */
+    protected void initializeLastRoundProcessed() throws IOException {
+        Optional<Long> minTimestampFromInProgressFiles =
+                getMinTimestampFromInProgressFiles();
+        if (minTimestampFromInProgressFiles.isPresent()) {
+            LOG.info("Initializing lastRoundProcessed from IN PROGRESS files with minimum "
```
tkhurana: IN should not be hardcoded here

phoenix-core-server/src/main/java/org/apache/phoenix/replication/ReplicationLogDiscovery.java
```diff
@@ -0,0 +1,501 @@
...
+     * @param files - List of files from which to select and process one randomly
+     */
+    private void processOneRandomFile(final List<Path> files) throws IOException {
+        // Pick a random file and process it
+        Path file = files.get(ThreadLocalRandom.current().nextInt(files.size()));
+        Optional<Path> optionalInProgressFilePath = Optional.empty();
+        try {
+            optionalInProgressFilePath = replicationLogTracker.markInProgress(file);
+            if (optionalInProgressFilePath.isPresent()) {
+                processFile(file);
```
tkhurana: I think we need to pass `optionalInProgressFilePath.get()` here since markInProgress can rename the file

Himanshu-g81: Done, thank you.

phoenix-core-server/src/main/java/org/apache/phoenix/replication/ReplicationLogTracker.java
Himanshu-g81: Done in subsequent commit.

phoenix-core-server/src/main/java/org/apache/phoenix/replication/ReplicationLogTracker.java
```diff
@@ -64,36 +63,33 @@ public class ReplicationLogTracker {
...
         return this.metrics;
     }
 
+    protected String getInSubDirectoryName() {
+        return getReplicationShardDirectoryManager().getRootDirectoryPath().getName();
+    }
+
     protected String getInProgressLogSubDirectoryName() {
-        return getNewLogSubDirectoryName() + "_progress";
+        return getInSubDirectoryName() + "_progress";
```
tkhurana, line 92: Shouldn't this be a more generic name instead of `getInSubDirectoryName`

PR #2286, opened 2025-09-12
PHOENIX-7700 Update default values for schema change retry policy

virajjasani:
Co-authored-by: David Manning <david.manning@salesforce.com>

Jira: PHOENIX-7700

### DIFF
--- a/phoenix-core-client/src/main/java/org/apache/phoenix/query/ConnectionQueryServicesImpl.java
+++ b/phoenix-core-client/src/main/java/org/apache/phoenix/query/ConnectionQueryServicesImpl.java
@@ -1748,7 +1748,7 @@ public class ConnectionQueryServicesImpl extends DelegateQueryServices
         }
       }
       numTries++;
-      if (!success) {
+      if (numTries < maxRetries && !success) {
         Thread.sleep(sleepInterval);
       }
     } while (numTries < maxRetries && !success);
--- a/phoenix-core-client/src/main/java/org/apache/phoenix/query/ConnectionQueryServicesImpl.java
+++ b/phoenix-core-client/src/main/java/org/apache/phoenix/query/ConnectionQueryServicesImpl.java
@@ -1748,7 +1748,9 @@ public class ConnectionQueryServicesImpl extends DelegateQueryServices
         }
       }
       numTries++;
-      Thread.sleep(sleepInterval);
+      if (!success) {
+        Thread.sleep(sleepInterval);
+      }
     } while (numTries < maxRetries && !success);
 
     watch.stop();
--- a/phoenix-core-client/src/main/java/org/apache/phoenix/query/ConnectionQueryServicesImpl.java
+++ b/phoenix-core-client/src/main/java/org/apache/phoenix/query/ConnectionQueryServicesImpl.java
@@ -1730,7 +1730,7 @@ public class ConnectionQueryServicesImpl extends DelegateQueryServices
       ConnectionQueryServicesImpl.this.props.getLong(QueryServices.DELAY_FOR_SCHEMA_UPDATE_CHECK,
         QueryServicesOptions.DEFAULT_DELAY_FOR_SCHEMA_UPDATE_CHECK);
     boolean success = false;
-    int numTries = 1;
+    int numTries = 0;
     PhoenixStopWatch watch = new PhoenixStopWatch();
     watch.start();
     do {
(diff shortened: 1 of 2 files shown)


phoenix-core-client/src/main/java/org/apache/phoenix/query/QueryServicesOptions.java
```diff
@@ -314,8 +314,8 @@ public class QueryServicesOptions {
   public static final boolean DEFAULT_EXPLAIN_CHUNK_COUNT = true;
   public static final boolean DEFAULT_EXPLAIN_ROW_COUNT = true;
   public static final boolean DEFAULT_ALLOW_ONLINE_TABLE_SCHEMA_UPDATE = true;
-  public static final int DEFAULT_RETRIES_FOR_SCHEMA_UPDATE_CHECK = 10;
-  public static final long DEFAULT_DELAY_FOR_SCHEMA_UPDATE_CHECK = 5 * 1000; // 5 seconds.
+  public static final int DEFAULT_RETRIES_FOR_SCHEMA_UPDATE_CHECK = 50;
+  public static final long DEFAULT_DELAY_FOR_SCHEMA_UPDATE_CHECK = 1000; // 1 second.
```
d-c-manning, line 318: > @virajjasani Shouldn't we go to sleep only when success is false https://github.com/apache/phoenix/blob/master/phoenix-core-client/src/main/java/org/apache/phoenix/query/ConnectionQueryServicesImpl.java#L1751

Right, I'm not intimately familiar with this part of the code, but it seemed like it was just a bug that it was sleeping even in the success case. Maybe a 5 second sleep is fine if there was an exception... especially if the exception code path is truly exceptional and usually we just succeed.

In that case, you may not even have to change these values.

phoenix-core-client/src/main/java/org/apache/phoenix/query/ConnectionQueryServicesImpl.java
```diff
@@ -1748,7 +1748,9 @@ private void checkAndRetry(RetriableOperation op) throws InterruptedException, T
         }
       }
       numTries++;
-      Thread.sleep(sleepInterval);
+      if (!success) {
+        Thread.sleep(sleepInterval);
+      }
     } while (numTries < maxRetries && !success);
```
d-c-manning, line 1754: I mean... while you're here... this seems wrong too?

The first time we come into this while condition, `numTries` will be `2`. So only if your retries are 3+ will you attempt even 1 retry? The default value of 10 for `NUM_RETRIES_FOR_SCHEMA_UPDATE_CHECK` means you will have done 9 attempts (8 retries). It's misleading.

virajjasani: Yeah, this whole logic seems flaky, it is also not updated anytime soon

virajjasani: Maybe Java should flag error for any do-while usage that is not appropriate :)

phoenix-core-client/src/main/java/org/apache/phoenix/query/ConnectionQueryServicesImpl.java
```diff
@@ -1730,16 +1730,16 @@ private void checkAndRetry(RetriableOperation op) throws InterruptedException, T
...
     PhoenixStopWatch watch = new PhoenixStopWatch();
     watch.start();
     do {
       try {
         success = op.checkForCompletion();
       } catch (Exception ex) {
-        // If we encounter any exception on the first or last try, propagate the exception and fail.
+        // If we encounter any exception on the first try, propagate the exception and fail.
         // Else, we swallow the exception and retry till we reach maxRetries.
-        if (numTries == 1 || numTries == maxRetries) {
```
virajjasani, line 1742: This logic was also not right, `numTries == maxRetries` is never going to be the case

phoenix-core-client/src/main/java/org/apache/phoenix/query/ConnectionQueryServicesImpl.java
virajjasani: I kept `numTries = 0` so let's say maxRetries = 1, we will execute this only once and error out. Shall we treat maxRetries as total num of attempts or total num of retries?

phoenix-core-client/src/main/java/org/apache/phoenix/query/QueryServicesOptions.java
virajjasani: @d-c-manning given that DDL operations in normal cases take 500-800ms to complete, it is likely that first time `op.checkForCompletion()` might return false and we might have to sleep temporarily only once but sleeping for 5s seems quite time consuming, so maybe 1s is not bad I think

PR #2288, opened 2025-09-18
PHOENIX-7562 :- Enhance Dual Cluster Client for consistent failover

lokiore:


phoenix-core-client/src/main/java/org/apache/phoenix/exception/SQLExceptionCode.java
```diff
@@ -280,6 +280,10 @@ public SQLException newException(SQLExceptionInfo info) {
     HA_READ_FROM_CLUSTER_FAILED_ON_NULL(1986, "F1Q86", "Unable to read from cluster for null."),
     HA_INVALID_PROPERTIES(1987, "F1Q87", "Invalid properties to get a Phoenix HA connection."),
     HA_CLUSTER_CAN_NOT_CONNECT(1988, "F1Q88", "Cluster can not serve any requests for this HA group"),
+    CLUSTER_ROLE_RECORD_NOT_FOUND(1989, "F1Q89", "Cluster role record not found for this HA group"),
+    STALE_CRR_RETHROW_AFTER_REFRESH(1991, "F1Q91", "Stale CRR found during operation refreshing it"),
```
apurtell: Wouldn't the client transparently refresh the CRR and retry? Why does the application need to handle it?

phoenix-core-client/src/main/java/org/apache/phoenix/exception/SQLExceptionCode.java
```diff
@@ -280,6 +280,10 @@ public SQLException newException(SQLExceptionInfo info) {
...
+    STALE_CRR_RETHROW_AFTER_REFRESH_FAILED(1992, "F1Q92", "Stale CRR found during operation, HAGroup refresh failed"),
```
apurtell: For the case where the refresh failed, this one makes sense to throw.

phoenix-core-client/src/main/java/org/apache/phoenix/exception/StaleClusterRoleRecordException.java
```diff
@@ -0,0 +1,44 @@
...
+ */
+package org.apache.phoenix.exception;
+
+import java.io.IOException;
+
+/**
+ * Exception thrown when attempting to do operation on STANDBY Cluster of HA Connection,
+ * indicating that the client has slate Cluster Role Record info for the HAGroup.
+ */
+public class StaleClusterRoleRecordException extends IOException {
```
apurtell, line 26: A disallowed operation doesn't definitively mean the client has a stale record. The client's record may be up to date. The error only means the client has tried to do an operation that is not supported in the current mode. It should refresh its understanding of current state to be sure, but this does not automatically imply things are "stale"

phoenix-core-client/src/main/java/org/apache/phoenix/jdbc/FailoverPhoenixConnection.java
```diff
@@ -333,10 +339,31 @@ <T> T wrapActionDuringFailover(SupplierWithSQLException<T> s) throws SQLExceptio
...
+                    //refreshing the ClusterRoleRecord and state transitions if required
+
+                    if (!context.getHAGroup().refreshClusterRoleRecord()) {
+                        throw new SQLExceptionInfo.Builder(SQLExceptionCode.STALE_CRR_RETHROW_AFTER_REFRESH_FAILED)
+                                .setMessage(String.format("Error while running operation Stale ClusterRoleRecord for HAGroup %s" +
+                                                " found with version %s, refreshing HAGroup failed :- %s",
+                                        context.getHAGroup(), context.getHAGroup().getRoleRecord().getVersion(), e)).build().buildException();
+                    }
+
+                    throw new SQLExceptionInfo.Builder(SQLExceptionCode.STALE_CRR_RETHROW_AFTER_REFRESH)
```
apurtell: What is the difference between these two cases? The error messages and error code names do not provide good context.

apurtell: What is the difference between these two cases? The error messages and error code names do not provide good context.

phoenix-core-client/src/main/java/org/apache/phoenix/jdbc/HAGroupStoreClient.java
```diff
@@ -689,6 +727,40 @@ private boolean isUpdateNeeded(HAGroupStoreRecord.HAGroupState currentHAGroupSta
...
     }
 
+    /**
+     * Combine two versions into a single long value to be used for comparison and sending back to client
+     * @param version1 HAGroupRecord version for cluster 1
+     * @param version2 HAGroupRecord version for cluster 2
+     * @return combined version
+     */
+
+    private long combineCanonicalVersions(long version1, long version2) {
```
apurtell, line 737: It's a bit weird to represent two values in java as bitfields. Why not int[2]?

phoenix-core-client/src/main/java/org/apache/phoenix/jdbc/HAGroupStoreManager.java
```diff
@@ -100,6 +137,17 @@ public boolean isMutationBlocked(String haGroupName) throws IOException {
         return false;
     }
 
+    public boolean isHAGroupOnClientStale(String haGroupName) throws IOException, SQLException {
+        if (checkStaleCRRForEveryMutation) {
+            HAGroupStoreClient haGroupStoreClient = getHAGroupStoreClient(haGroupName);
+            //If local cluster is not ACTIVE/ACTIVE_TO_STANDBY, it means the Failover CRR is stale on client
+            //As they are trying to write/read from a STANDBY cluster.
+            return haGroupStoreClient.getPolicy() == HighAvailabilityPolicy.FAILOVER &&
```
apurtell, line 153: This doesn't allow for future evolution where the CRR is valid but the local cluster is simply in the wrong state for a given operation. In other words, just because a cluster is in the wrong state for a given operation it doesn't mean the client does not have the most up to date CRR. 

In the future some operations may be allowed on the STANDBY side. At least, this should not foreclose the possibility.

In that future case, simply assuming the latest state is "stale" because of disallowed operation, doing a refresh, and getting the same state back, then fetching again because the state did not change (it is actually up to date), would cause looping?

What I am suggesting is getting away from thinking "we are stale" (definitive, we will always assume we are not up to date) to "we have done something unexpectedly invalid, let's double check we are using the correct policy" (not definitive, maybe we are up to date or maybe not).

phoenix-core-client/src/main/java/org/apache/phoenix/query/QueryConstants.java
```diff
@@ -708,7 +709,8 @@ enum JoinType {INNER, LEFT_OUTER}
             CLUSTER_ROLE_1 + " VARCHAR," + // Role for peer cluster might not be accurate, we will use only local role for recovery if needed
             CLUSTER_ROLE_2 + " VARCHAR," + // Role for peer cluster might not be accurate, we will use only local role for recovery if needed
             POLICY + " VARCHAR," +  // There should be only one policy for an HA group
-            VERSION + " BIGINT," +  // Version should be incremented for Admin Updates, only for CLUSTER_URLs and REGISTRY_TYPE
+            VERSION_1 + " BIGINT," +  // Version should be incremented for Admin Updates, only for CLUSTER_URLs and REGISTRY_TYPE
```
apurtell: _1 and _2 designate the version of two separate clusters, but that isn't clear. VERSION_1 and VERSION_2 may be two version constants or who knows.  

VERSION_CLUSTER_1 or VERSION_CLUSTER_2 are possible names that make it clear. 

Getting this right matters because there will be compatibility concerns if these are changed later.

phoenix-core-client/src/main/java/org/apache/phoenix/query/QueryServices.java
```diff
@@ -386,6 +386,9 @@ public interface QueryServices extends SQLCloseable {
     public static final String PHOENIX_VIEW_TTL_TENANT_VIEWS_PER_SCAN_LIMIT = "phoenix.view.ttl.tenant_views_per_scan.limit";
     // Block mutations based on cluster role record
     public static final String CLUSTER_ROLE_BASED_MUTATION_BLOCK_ENABLED = "phoenix.cluster.role.based.mutation.block.enabled";
+    // Check HAGroup is Stale for mutations
+    public static final String HA_GROUP_STALE_FOR_MUTATION_CHECK_ENABLED = "phoenix.ha.group.stale.for.mutation.check.enabled";
```
apurtell, line 390: Why wouldn't we always check if the CRR is up to date if we have a question about it, i.e. if we tried an operation and got back an unexpected failure?

phoenix-core-client/src/main/java/org/apache/phoenix/util/GetClusterRoleRecordUtil.java
```diff
@@ -0,0 +1,227 @@
...
+    public static ClusterRoleRecord fetchClusterRoleRecord(String url, String haGroupName, HighAvailabilityGroup haGroup, long pollerInterval, Properties properties) throws SQLException {
+        ClusterRoleRecord clusterRoleRecord = getClusterRoleRecord(url, haGroupName, true, properties);
+
+        if (clusterRoleRecord.getPolicy() == HighAvailabilityPolicy.FAILOVER &&
+                !clusterRoleRecord.getRole1().isActive() && !clusterRoleRecord.getRole2().isActive()) {
+            LOGGER.info("Non-active ClusterRoleRecord found for HA group {}. Scheduling poller to check every {} seconds," +
+            "until we find an ACTIVE CRR", haGroupName, pollerInterval);
+            //Schedule a poller to fetch ClusterRoleRecord every 5 seconds (or configured value)
+            //until we get an Active ClusterRoleRecord and return the Non-Active CRR
+            schedulePoller(url, haGroupName, haGroup, pollerInterval, properties);
```
apurtell, line 146: How expensive is the fetch of the CRR? 

If every server and client in a cluster or pointing to the cluster must contact a server on the cluster every 5 seconds, that would be expensive, that server would be pretty hot

Also, a 5 second polling interval means every client will see an issue if there is a blip >= 5 seconds in the availability of whatever is serving the CRR. That is not failure tolerant. If something is temporarily down or unavailable but you don't need it it is not actually an availability problem. Tolerance is increased.

A poll is only waste and a potential availability problem whenever we aren't doing anything or if the CRR didn't actually change. I think polling is the wrong model for this. 

Should this be notification based and not polled? Fetch it during initialization. Then, we would be "notified" when some operation we expect to succeed is instead failed due to some policy check (on the remote side). And at that time, the up to date CRR could be fetched. 

Let the client assume the CRR it has is up to date. Let it try something. If the thing it tries fails because the local CRR is possibly out of sync, then fetch the latest CRR and potentially retry the operation if the cached CRR and newly fetched CRR are different. And cache the newly fetched CRR if it changed.

phoenix-core-server/src/main/java/org/apache/phoenix/hbase/index/IndexRegionObserver.java
```diff
@@ -634,12 +635,24 @@ public void preBatchMutate(ObserverContext<RegionCoprocessorEnvironment> c,
...
-                  throw new MutationBlockedIOException("Blocking Mutation as Some CRRs are in "
-                          + "ACTIVE_TO_STANDBY state and "
-                          + "CLUSTER_ROLE_BASED_MUTATION_BLOCK_ENABLED is true");
-              }
+                //TODO: Below approach might be slow need to figure out faster way, slower part is
+                //getting haGroupStoreClient We can also cache roleRecord (I tried it and still its
+                //slow due to haGroupStoreClient initialization) and caching will give us old result
+                //in case one cluster is unreachable instead of UNKNOWN. Check if mutation's
+                //haGroup is stale
+                boolean isHAGroupOnClientStale = haGroupStoreManager
```
apurtell: If there is a local uncertainty about state, either because state is missing or not initialized yet or if the remote gave back an error suggesting client understanding is not current, refresh the CRR, then decide what to do next. I think this is the way.

phoenix-core-client/src/main/java/org/apache/phoenix/jdbc/PhoenixHAAdmin.java
```diff
@@ -151,6 +151,12 @@ public static String getLocalZkUrl(Configuration conf) {
         return String.format("%s:%d:%s", localZkQuorum, port, localZkRoot);
     }
 
+    public static String getLocalMasterUrl(Configuration conf) {
+        String local = conf.get(HConstants.MASTER_ADDRS_KEY).replaceAll(":", "\\\\:");
+        LOG.info("LOCAL MASTERRRRRRR {}", local);
```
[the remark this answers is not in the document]
apurtell: Yes let's fix this :-)

phoenix-core-client/src/main/java/org/apache/phoenix/exception/SQLExceptionCode.java
lokiore: Client is transparently refreshing the CRR when it receives exception that the CRR is stale, but we are not retrying the mutations, as we are trying to detect Staleness when Failover event happens and with that the connections will be closed after refresh, we decided to not retry them and let application retry with new ACTIVE.
There are 2 codes, one if refresh failed and one if it passes for more visibility and debugging.

phoenix-core-client/src/main/java/org/apache/phoenix/jdbc/HAGroupStoreClient.java
lokiore: After our recent discussion, CRR version implementation will change a bit on server side, we will have version for admin changes only. It'll come with @ritegarg's next change, so either on top of it, or I'll have to pull and resolve conflicts

phoenix-core-client/src/main/java/org/apache/phoenix/jdbc/HAGroupStoreManager.java
lokiore: > In that future case, simply assuming the latest state is "stale" because of disallowed operation, doing a refresh, and getting the same state back, then fetching again because the state did not change (it is actually up to date), would cause looping?

Client will have to have some notification that it is sending mutations(btw this method is only used for mutations but with recent discussion might be used for point lookups as well) to a wrong cluster, If cluster is not Active, then client should refresh the CRR cache and retry, refreshing will make this cluster STANDBY in client's CRR and it won't send the request to this cluster instead it will go to the dr/secondary cluster which will be in active state.

> In the future some operations may be allowed on the STANDBY side. At least, this should not foreclose the possibility.

We will have to define what operations and in what scenarios that will be allowed and then handle accordingly. As currently client only give back connection to Active cluster only.


For client to determine that the operation is send to wrong cluster, we will have to check on server if it should receive that operation (We could think of 2 ways either checking the local state/role or comparing versions) we (@kadirozde and me) chose checking state, as a way forward, as for versions we will have to send client version with every operation and there are possibilities that version change happens but role change won't happen at that time, and now there might be possibility that version change might not happen with some role change scenarios.

phoenix-core-client/src/main/java/org/apache/phoenix/query/QueryServices.java
lokiore: Sorry for confusion on this one, will update the comment, this is only there for some to pass which don't need the check on server. like LogginHAConnectionLimiter, from debugging I could pinpoint it to getting HAGroupStoreClient in HAGroupStoreManager for checking anything that takes a bit of time and which leads to test failures, right now we still don't have autonomous changes required for testing like (updating system table or ZNode syncs, which we have to do explicitly for testing)

WIll add comment to remove this when everything is available and we won't need to initialize haGroupStoreClient as many times as we are doing today for testing.

phoenix-core-client/src/main/java/org/apache/phoenix/util/GetClusterRoleRecordUtil.java
lokiore: > Should this be notification based and not polled? Fetch it during initialization. Then, we would be "notified" when some operation we expect to succeed is instead failed due to some policy check (on the remote side). And at that time, the up to date CRR could be fetched.

That is the current behavior, this polling will happen only when client receives a CRR for a HAGroup, which has both roles as non-active, as client doesn't have any active cluster, connection creation can't happen then we will go into polling mode to fetch the CRR until we have an ACTIVE cluster and client can resume creating connections and poller will stop.

Right now client assumes the CRR is up to date, but if CRR received doesn't have ACTIVE cluster, client will have to poll as it doesn't know which cluster to create connection to.

phoenix-core-client/src/main/java/org/apache/phoenix/exception/SQLExceptionCode.java
kadirozde: What is the difference between HA_NO_ACTIVE_CLUSTER and STALE_CRR_RETHROW_AFTER_REFRESH?

phoenix-core-client/src/main/java/org/apache/phoenix/exception/StaleClusterRoleRecordException.java
kadirozde: This exception will be generated only when the operation includes the HA group name as an operation attribute. Inclusion of HA group name means the operation should be executed only if the role of the cluster is active for mutations, and active or active-to-standby for scans. If the cluster role is not what was expected then the client cache is possibly stale now and needs to be refreshed.

kadirozde: More accurately (1) if the cluster role is not active or active-to-standby then StaleClusterRoleRecordException is thrown regardless of the operation type and (2) if the operation type is mutation and the cluster role is active-to-standby then MutationBlockedIOException is thrown.

phoenix-core-server/src/main/java/org/apache/phoenix/hbase/index/IndexRegionObserver.java
kadirozde: The Phoenix client will refresh if it gets a StaleClusterRoleRecordException. We (@lokiore and I) decided that the client should not retry in this case, instead it will return a SQL exception to the client and in turn the client will close the current connection. This is because if the current cluster is not the right cluster for this mutation, then a new Phoenix connection needs to be opened using the right HBase connection. In this case the application needs to close the old connection and open a new connection and continue to execute the operations using the new Phoenix connection context. In future, we may try to improve this behavior but it is not straightforward to switch an application connection from one cluster to another and resubmit the pending operation.

kadirozde (approved):
+1, Thanks!

PR #2289, opened 2025-09-18
PHOENIX-7701 UngroupedAggregateRegionScanner doesn't release rpc handler on page timeouts

tkhurana:


phoenix-core-server/src/main/java/org/apache/phoenix/coprocessor/UngroupedAggregateRegionScanner.java
```diff
@@ -702,6 +708,10 @@ public boolean next(List<Cell> resultsToReturn, ScannerContext scannerContext)
           } while (
             hasMore && (EnvironmentEdgeManager.currentTimeMillis() - startTime) < pageSizeMs
```
kadirozde, line 709: We need to add here and to every other loop ScannerContextUtil.checkTimeLimit(scannerContext). This is to make sure that when a lower Phoenix region scanner requests to return immediately, all the region scanners above that break their loop.

tkhurana: @kadirozde That was something I was planning to address in a separate PR which ensures that we page out across multiple `next()` calls.

kadirozde (approved):

virajjasani:
Let's cherry-pick the commit in 5.3 branch also

PR #2291, opened 2025-09-24
PHOENIX-7731: Add phoenix-hbase-compat module for HBase 2.6.4

sanjeet006py:
- JIRA: [PHOENIX-7731](https://issues.apache.org/jira/browse/PHOENIX-7731)
- The PR has added a new `phoenix-hbase-compat` module for HBase 2.6.4. [Link to the Phoenix dev mailing list thread](https://lists.apache.org/thread/7hjn3f7hrglsfg9yqshvlcdq936qx37h) with alignment on creating new `phoenix-hbase-compat` module.
- The new `phoenix-hbase-compat-2.6.4` module is same as `phoenix-hbase-compat-2.6.0` except for class: `CompatIndexHalfStoreFileReader` where new API in 2.6.4 is being used: `StoreFileInfo.createStoreFileInfoForHFile()`.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -2148,8 +2148,7 @@
       <id>phoenix-hbase-compat-2.5.12-default</id>
       <activation>
         <property>
-          <name>hbase.profile</name>
-          <value>2.5</value>
+          <name>!hbase.profile</name>
         </property>
       </activation>
       <properties>
@@ -2210,8 +2209,8 @@
       <id>phoenix-hbase-compat-2.6.4</id>
       <activation>
         <property>
-          <name>!hbase.profile</name>
-          <!-- <value>2.6</value> -->
+          <name>hbase.profile</name>
+          <value>2.6</value>
         </property>
       </activation>
       <properties>


pom.xml
```diff
@@ -2188,11 +2195,26 @@
...
+      <id>phoenix-hbase-compat-2.6.3</id>
+      <activation>
+        <property>
+          <name>hbase.profile</name>
+          <value>2.6.3</value>
+        </property>
+      </activation>
+      <properties>
+        <hbase.profile>2.6.3</hbase.profile>
+        <hadoop.version>3.4.1</hadoop.version>
```
stoty: Why did you add this ?
While 2.6.3 is built with 3.4.1, in my tests it ran fine with 3.4.2.
We generally want the latest Hadoop that the Hbase version works with (and can run the tests)

stoty (changes requested):

pom.xml
sanjeet006py: When I initially started on PR then Phoenix by default was using hadoop 3.4.1 so, I also used same for pinning hadoop version for HBase 2.6.3. But I see now HBase & Phoenix are using 3.4.2 so will update the same. Thanks

stoty: You should just remove the line and fall back to the default 3.4.2

sanjeet006py:
> @sanjeet006py we still don't have full build results with 2.6 profile

Yes, one of my changes in the PR broke regions splits for local indexes. And, Phoenix version has been updated to 5.4.0 so, that also broke PR's build as I didn't sync latest from master.

virajjasani:
My bad, i thought 2.6 default profile was changed to 2.5. I misunderstood with 5.4 change

pom.xml
```diff
@@ -54,6 +55,12 @@
     <module>phoenix-pherf</module>
     <module>phoenix-tracing-webapp</module>
     <!-- shaded artifact and assembly modules are added in shade-and-assembly profile -->
+    <module>phoenix-client-parent</module>
+    <module>phoenix-client-parent/phoenix-client-embedded</module>
+    <module>phoenix-client-parent/phoenix-client-lite</module>
+    <module>phoenix-server</module>
+    <module>phoenix-mapreduce-byo-shaded-hbase</module>
+    <module>phoenix-assembly</module>
```
virajjasani: Curious, we need this change for checkstyle? It seems we should have had these modules defined here right? @stoty wdyt?

stoty: These shouldn't be listed here.
Adding these here will disable the build optimizations that enable quicker builds by skipping the shading steps.

stoty: See the comment above.

sanjeet006py: Actually, I am seeing Javadoc failure after patch is applied saying above extra modules listed under `shade-and-assembly` profile for `hbase-2.6` are not found. I thought adding them here might help me get past that but doesn't seem to be case. So, reverting this change. I will be posting a summary of unexpected failures I am seeing and would need help on that. Thanks

stoty:
Yes, go ahead @sanjeet006py .

PR #2292, opened 2025-09-27
PHOENIX-7705 Support for a row size function

kadirozde:


phoenix-core-client/src/main/java/org/apache/phoenix/expression/function/RowSizeFunction.java
```diff
@@ -0,0 +1,82 @@
...
+import org.apache.phoenix.parse.FunctionParseNode.BuiltInFunction;
+import org.apache.phoenix.parse.RowSizeParseNode;
+import org.apache.phoenix.schema.tuple.Tuple;
+import org.apache.phoenix.schema.types.*;
+
+/**
+ * Function to return the total size of the HBase cells that constitute a given row
+ */
+@BuiltInFunction(name = RowSizeFunction.NAME, nodeClass = RowSizeParseNode.class, args = {})
+public class RowSizeFunction extends ScalarFunction {
```
ujjawal4046, line 34: Are the scalar function evaluated on server side as well (or only on client side) ? If it's client side, then we need to fetch the whole row back to client for size computation ?

phoenix-core-client/src/main/java/org/apache/phoenix/compile/QueryCompiler.java
```diff
@@ -752,6 +753,16 @@ protected QueryPlan compileSingleQuery(StatementContext context, SelectStatement
   protected QueryPlan compileSingleFlatQuery(StatementContext context, SelectStatement select,
     boolean asSubquery, boolean allowPageFilter, QueryPlan innerPlan, boolean inJoin,
     boolean inUnion) throws SQLException {
+    for (AliasedNode node : select.getSelect()) {
+      if (node.getNode() instanceof RowSizeParseNode) {
```
ujjawal4046, line 757: I am not clear with the usage. Does it mean we can't use query where row_size needs to be fetched for each row (e.g. `select row_size() from table` or  `select row_size() from table group by tenant_id`)

phoenix-core-client/src/main/java/org/apache/phoenix/iterate/BaseResultIterators.java
```diff
@@ -361,12 +361,19 @@ private static void initializeScan(QueryPlan plan, Integer perScanLimit, Integer
       if (optimizeProjection) {
         optimizeProjection(context, scan, table, statement);
       }
+      if (context.hasRowSizeFunction()) {
+        scan.getFamilyMap().clear();
+        ScanUtil.removePageFilter(scan);
```
ujjawal4046: I see the need to remove family map so that it accounts for cells across all column families. 

Why do we need to remove page filter ?

ujjawal4046 (commented):
@tkhurana has already asked on how to account for all versions/delete markers in a row in a comment on the [jira](https://issues.apache.org/jira/browse/PHOENIX-7705). 

@kadirozde A related question to this - 
- If we introduce a separate RAW_ROW_SIZE for such case, can it be used in conjunction with other functions (e.g. `select raw_row_size,count(*) from table group by tenant_id;`). This is considering raw_row_size would need to do a raw scan which may not be compatible with other queries (which assumes that scan only always return most recent version)

phoenix-core-client/src/main/java/org/apache/phoenix/iterate/BaseResultIterators.java
kadirozde: Page filter removal was not intentional. Good catch!

phoenix-core-client/src/main/java/org/apache/phoenix/expression/function/RowSizeFunction.java
kadirozde: It can be evaluated on the client size to check if the where clause evaluates to true on an empty tuple and, or when it is specified as a top level expression node in a select clause. This PR does not allow the row_size function to be a top level node in a select clause. In this PR, a row is never returned to the client; only its size is returned as part of an aggregation function result.

phoenix-core-client/src/main/java/org/apache/phoenix/compile/QueryCompiler.java
kadirozde: Please see the exception message and also the row size test to see how to get individual row sizes.

ujjawal4046 (approved):

tkhurana:
@kadirozde The IT tests are timing out and being aborted. https://ci-hadoop.apache.org/job/Phoenix/job/Phoenix-PreCommit-GitHub-PR/view/change-requests/job/PR-2292/

tkhurana:
`java.net.SocketTimeoutException: callTimeout=60000, callDuration=68935: java.io.IOException: org.apache.phoenix.expression.OrExpression cannot be cast to org.apache.phoenix.expression.function.SingleAggregateFunction
	at org.apache.hadoop.hbase.ipc.RpcServer.call(RpcServer.java:479)
	at org.apache.hadoop.hbase.ipc.CallRunner.run(CallRunner.java:124)
	at org.apache.hadoop.hbase.ipc.RpcHandler.run(RpcHandler.java:102)
	at org.apache.hadoop.hbase.ipc.RpcHandler.run(RpcHandler.java:82)
Caused by: java.lang.ClassCastException: org.apache.phoenix.expression.OrExpression cannot be cast to org.apache.phoenix.expression.function.SingleAggregateFunction
	at org.apache.phoenix.expression.aggregator.ServerAggregators.deserialize(ServerAggregators.java:118)
	at org.apache.phoenix.coprocessor.UngroupedAggregateRegionScanner.next(UngroupedAggregateRegionScanner.java:613)
	at org.apache.phoenix.coprocessor.UngroupedAggregateRegionScanner.nextRaw(UngroupedAggregateRegionScanner.java:595)
	at org.apache.phoenix.coprocessor.DelegateRegionScanner.next(DelegateRegionScanner.java:108)
	at org.apache.phoenix.coprocessor.DelegateRegionScanner.nextRaw(DelegateRegionScanner.java:77)
	at org.apache.phoenix.coprocessor.DelegateRegionScanner.next(DelegateRegionScanner.java:108)
	at org.apache.phoenix.coprocessor.DelegateRegionScanner.nextRaw(DelegateRegionScanner.java:77)
	at org.apache.phoenix.coprocessor.BaseScannerRegionObserver$RegionScannerHolder.nextRaw(BaseScannerRegionObserver.java:266)
	at org.apache.hadoop.hbase.regionserver.RSRpcServices.scan(RSRpcServices.java:3403)
	at org.apache.hadoop.hbase.regionserver.RSRpcServices.scan(RSRpcServices.java:3669)
	at org.apache.hadoop.hbase.shaded.protobuf.generated.ClientProtos$ClientService$2.callBlockingMethod(ClientProtos.java:43508)
	at org.apache.hadoop.hbase.ipc.RpcServer.call(RpcServer.java:415)
	... 3 more`

kadirozde:
@tkhurana, Thank you for pointing out the test failure. It should be fixed now.

PR #2411, opened 2026-04-16
PHOENIX-7799 Coalesce splits by region server to avoid hotspotting from concurrent mappers

rahulLiving:
https://issues.apache.org/jira/browse/PHOENIX-7799

PhoenixSyncTableTool creates one MapReduce InputSplit per HBase region, causing each split to spawn its own mapper task. When multiple regions reside on the same RegionServer, concurrent mappers hit the same server simultaneously, leading to hotspotting and resource contention (noisy neighbor problem). This degrades performance and can cause timeouts.

Implement locality-aware split coalescing that groups all InputSplits from the same RegionServer into a single coalesced split. This ensures only one mapper processes regions from each server sequentially, eliminating concurrent requests and hotspotting. The feature will be controlled by the configuration property phoenix.sync.table.split.coalescing (default: false). For a table with 100 regions across 5 RegionServers, this reduces mapper count from 100 to 5, eliminating server-side contention while maintaining data locality.


phoenix-core-server/src/main/java/org/apache/phoenix/mapreduce/PhoenixSyncTableInputFormat.java
```diff
@@ -211,4 +242,93 @@ List<InputSplit> filterCompletedSplits(List<InputSplit> allSplits,
...
+    Map<String, List<PhoenixInputSplit>> splitsByServer = new HashMap<>();
+    for (InputSplit split : splits) {
+      PhoenixInputSplit pSplit = (PhoenixInputSplit) split;
+      KeyRange keyRange = pSplit.getKeyRange();
+      HRegionLocation regionLocation =
+        regionLocator.getRegionLocation(keyRange.getLowerRange(), false);
+      if (regionLocation == null) {
+        throw new IOException("Could not determine region location for key: "
+          + Bytes.toStringBinary(keyRange.getLowerRange()));
+      }
```
haridsv, line 295: I am thinking that this should never see a `null` if all regions are fully online and client cache is up to date. However (per what AI says), if the region the key maps to is currently offline due to a RIT event (say, in the middle of a split), then the return value will be null, so shouldn't we retry to ride over such RITs, just like how hbase-client does during data path such as scans? Otherwise, there is a high chance of hitting a RIT and treating it as an error, correct?

haridsv (commented):
Currently, the mapper doesn't report meaningful progress during execution, it would effectively shows 0% until completion, then jump to 100%, isn't it? With coalesced splits the issue sort of gets worse, as each mapper runs for much longer. Before, there is at least some visibility into the progress based on the %ge of mappers completed. It would be helpful to include progress reporting as each range is completed. One idea is to make `PhoenixNoOpSingleRecordReader` more sophisticated to return one dummy record for each range and have `getProgress` return a ratio based on how many ranges completed. You would of course rename it to not imply "SingleRecord".

rahulLiving:
>Currently, the mapper doesn't report meaningful progress during execution, it would effectively shows 0% until completion, then jump to 100%, isn't it? 

No that is not the case currently. From my past execution, I saw progress being reported incrementally from 0->100. I think progress is being reported based on how many mapper pending/in-progress/completed. But you are right that, any mapper which is running will be considered as 0/1 for job completion. There won't be a middle ground. 

I see the formulae at mapper level, `Progress = Number of bytes read/Total bytes in Input Split.`
Will have to see if we can update bytes read manually in RecordReader. Do you think we should take this up as follow up ? @haridsv

phoenix-core-server/src/main/java/org/apache/phoenix/mapreduce/PhoenixSyncTableInputFormat.java
rahulLiving: Even during RIT, I believe meta would return cached location which might be outdated. It could give a false colocation but that is expected for some of regions as sync tool runs.

phoenix-core-server/src/main/java/org/apache/phoenix/mapreduce/PhoenixSyncTableInputFormat.java
```diff
@@ -94,15 +101,36 @@ public List<InputSplit> getSplits(JobContext context) throws IOException, Interr
...
+      conf.getBoolean(PhoenixSyncTableTool.PHOENIX_SYNC_TABLE_SPLIT_COALESCING,
+        PhoenixSyncTableTool.DEFAULT_PHOENIX_SYNC_TABLE_SPLIT_COALESCING);
+    LOGGER.info("Split coalescing enabled: {}, for table {}", enableSplitCoalescing, tableName);
+
+    if (enableSplitCoalescing && unprocessedSplits.size() > 1) {
+      try (Connection conn = ConnectionUtil.getInputConnection(conf)) {
+        PhoenixConnection pConn = conn.unwrap(PhoenixConnection.class);
+        byte[] physicalTableName = pConn.getTable(tableName).getPhysicalName().getBytes();
+        org.apache.hadoop.hbase.client.Connection hbaseConn =
+          pConn.getQueryServices().getAdmin().getConnection();
```
tkhurana, line 119: There is a resource leak here. Admin object is not being closed. Also hconnection is also not being closed. We already have similar APIs available to us in ConnectionQueryServicesImpl class. My suggestion would be to use `getTableRegionLocation`

phoenix-core-server/src/main/java/org/apache/phoenix/mapreduce/PhoenixInputSplit.java
```diff
@@ -39,6 +39,7 @@ public class PhoenixInputSplit extends InputSplit implements Writable {
 
   private List<Scan> scans;
   private KeyRange keyRange;
+  private List<KeyRange> keyRanges;
```
tkhurana, line 42: keyRange is redundant now. It's just the span from `keyRanges.get(0).getLowerRange()` to `keyRanges.get(last).getUpperRange()`. You can keep only keyRanges and derive the overall bounds on demand:


```suggestion
  public KeyRange getKeyRange() {
      return KeyRange.getKeyRange(
          keyRanges.get(0).getLowerRange(),
          keyRanges.get(keyRanges.size() - 1).getUpperRange());
  }
```

tkhurana:
With coalescing each map task will span multiple table regions which means that each mapper will run longer. We could run the risk of map task timing out and getting killed. To get around that we might need to set a higher timeout value.

phoenix-core-server/src/main/java/org/apache/phoenix/mapreduce/PhoenixSyncTableMapper.java
```diff
@@ -88,15 +87,15 @@ public enum SyncCounters {
...
-  private byte[] mapperRegionEnd;
+  private List<KeyRange> regionKeyRanges;
   private PhoenixSyncTableOutputRepository syncTableOutputRepository;
   private Timestamp mapperStartTime;
 
   @Override
   protected void setup(Context context) throws InterruptedException {
     try {
       super.setup(context);
+      context.getCounter(SyncCounters.TASK_CREATED).increment(1);
```
tkhurana, line 98: Isn't this just the number of mappers created ? YARN already maintains that information. We can simply use the corresponding counter from YARN rather than maintaining our own.

rahulLiving:
>With coalescing each map task will span multiple table regions which means that each mapper will run longer. We could run the risk of map task timing out and getting killed. To get around that we might need to set a higher timeout value.

We have `phoenix.sync.table.query.timeout` config to override MRJobConfig.TASK_TIMEOUT in case we are coalescing. Does that sounds good enough ?

phoenix-core-server/src/main/java/org/apache/phoenix/mapreduce/PhoenixSyncTableInputFormat.java
haridsv: It is possible that the split generation logic ended up caching the metadata with some of the regions in transition so when we try to locate them, it would end up using the same cached metadata and fail to lookup. If we introduce at least 1 retry with a metadata reload, it can help avoid the IOException in majority of the scenarios. WDYT?

haridsv:
> I saw progress being reported incrementally from 0->100. I think progress is being reported based on how many mapper pending/in-progress/completed. **But you are right that, any mapper which is running will be considered as 0/1 for job completion. There won't be a middle ground.**

This (the highlighted) is exactly what I am saying! Also, we wouldn't know in advance how many bytes would be encountered and so there is noe meaningful way to give a percentage based on bytes processed, but this simplification was fine before this change, but not anymore.

> I see the formulae at mapper level, Progress = Number of bytes read/Total bytes in Input Split.
> Will have to see if we can update bytes read manually in RecordReader.

My point is, we know how many splits will be processed by the mapper so we can easily generate a progress based on the number of splits completed.

> Do you think we should take this up as follow up ?

If you mean, you want to address it in the very next PR, I am fine with it, but I don't think this should be considered as optional.

phoenix-core-server/src/main/java/org/apache/phoenix/mapreduce/PhoenixSyncTableInputFormat.java
rahulLiving: I see with @tkhurana suggestion to use [ConnectionQueryServiceImpl.getTableRegionLocation](https://github.com/apache/phoenix/blob/master/phoenix-core-client/src/main/java/org/apache/phoenix/query/ConnectionQueryServicesImpl.java#L6773), we will have one retry by default

rahulLiving:
>If you mean, you want to address it in the very next PR, I am fine with it, but I don't think this should be considered as optional.

yes as part of very next PR, we have couple of more PR incoming for repair phase.

phoenix-core-server/src/main/java/org/apache/phoenix/mapreduce/PhoenixSyncTableInputFormat.java
haridsv: It is only retrying on exception case (not on null value), besides, the retry doesn't happen with reload=true either.

rahulLiving: I was thinking if we should integrate retry as part of `ConnectionQueryServicesImpl.getTableRegionLocation` itself with reload=true, I see there are 4 callers for same ?

rahulLiving: How about to maintain a UNKNOWN_SERVER split for all such occurence of locating region fails. And this could be assigned as one mapper ?

haridsv: Yes, grouping them together actually makes sense to me and any RIT recovery would be handled by hbase-client as part of mapper execution itself and help improve the overall tool resiliency.

tkhurana (approved):

haridsv (approved):

PR #2419, opened 2026-04-22
PHOENIX-7751: Feature to validate table data using PhoenixSyncTable tool b/w source and target cluster

rahulLiving:


phoenix-core-server/src/main/java/org/apache/phoenix/mapreduce/PhoenixSyncTableOutputRepository.java
```diff
@@ -223,15 +225,15 @@ public List<PhoenixSyncTableCheckpointOutputRow> getProcessedChunks(String table
...
       if (hasStartBoundary) {
-        ps.setBytes(paramIndex, mapperRegionStart);
+        ps.setString(paramIndex, bytesToHex(mapperRegionStart));
       }
       try (ResultSet rs = ps.executeQuery()) {
         while (rs.next()) {
-          byte[] rawStartKey = rs.getBytes("START_ROW_KEY");
-          byte[] endRowKey = rs.getBytes("END_ROW_KEY");
+          byte[] rawStartKey = hexToBytes(rs.getString("START_ROW_KEY"));
+          byte[] endRowKey = hexToBytes(rs.getString("END_ROW_KEY"));
```
haridsv, line 236: I would recommend using base64 encoding which is a lot more efficient than hex (always doubles the size).

phoenix-core/src/it/java/org/apache/phoenix/end2end/PhoenixSyncTableToolIT.java
```diff
@@ -207,80 +208,36 @@ public void testSyncTableWithDeletedRowsOnTarget() throws Exception {
...
     flushAndMajorCompact(CLUSTERS.getHBaseCluster2(), uniqueTableName);
 
-    long sourceCountPostCompact = TestUtil.getRowCount(sourceConnection, uniqueTableName);
-    long targetCountPostCompact = TestUtil.getRowCount(targetConnection, uniqueTableName);
-    assertEquals("Source should see 10 live rows", 10, sourceCountPostCompact);
-    assertEquals("Target should see 7 live rows", 7, targetCountPostCompact);
-
-    // We shouldn't see expired rows even with --raw-scan flag
-    Job job2 = runSyncTool(uniqueTableName, "--raw-scan");
+    // Run sync tool after compaction - should detect mismatch (source has rows, target doesn't)
```
haridsv: Shouldn't we handle this filtering in the sync tool coproc? Otherwise, we will end up seeing a lot of false +ves.

phoenix-core-server/src/main/java/org/apache/phoenix/mapreduce/PhoenixSyncTableOutputRepository.java
rahulLiving: base64 wouldn't work with varchar comparison.

haridsv: > base64 wouldn't work with varchar comparison.

Can you explain? Which comparison code are you referring to?

rahulLiving: Image

phoenix-core/src/it/java/org/apache/phoenix/end2end/PhoenixSyncTableToolIT.java
rahulLiving: Yeah, we don't want to read expired rows. Thanks.

phoenix-core-server/src/main/java/org/apache/phoenix/mapreduce/PhoenixSyncTableOutputRepository.java
haridsv: Oh, you mean to preserve the order? Good point!

phoenix-core-server/src/main/java/org/apache/phoenix/mapreduce/PhoenixSyncTableMapper.java
```diff
@@ -395,7 +396,7 @@ private ChunkScannerContext createChunkScanner(Connection conn, byte[] startKey,
     scan.setAttribute(BaseScannerRegionObserverConstants.SYNC_TABLE_CHUNK_FORMATION, TRUE_BYTES);
     scan.setAttribute(BaseScannerRegionObserverConstants.SKIP_REGION_BOUNDARY_CHECK, TRUE_BYTES);
     scan.setAttribute(BaseScannerRegionObserverConstants.UNGROUPED_AGG, TRUE_BYTES);
-    ScanUtil.setScanAttributesForPhoenixTTL(scan, pTable, phoenixConn);
```
rahulLiving: This seems to be useful for PhoenixTTLRegionScanner which is for view TTL, so not required.

haridsv (approved):
As discussed offline, exposing `setEmptyColumnScanAttributes` and calling it from the tool is not very intuitive, instead this should be made private and we should have a new public util method like `setScanAttributesForSyncTool` that combines all the scan attributes specific logic. However, it is fine to do this change in the upcoming PR, so approving this PR.

PR #2423, opened 2026-04-24
PHOENIX-7792 New file age check before processing during replay

Himanshu-g81:
During replay, we now wait before processing a new file to ensure it's no longer being written by the source cluster. A file must be at least roundTime + configurable % old (default: 15%) before it becomes eligible. For the default 60s round time, that's 69s.                                                                                   

When some files are old enough, we process those immediately. When all remaining files are too young, we sleep for exactly the time needed until the oldest one becomes eligible.

### DIFF
--- a/phoenix-core-server/src/main/java/org/apache/phoenix/replication/ReplicationLogDiscovery.java
+++ b/phoenix-core-server/src/main/java/org/apache/phoenix/replication/ReplicationLogDiscovery.java
@@ -18,6 +18,7 @@
 package org.apache.phoenix.replication;
 
 import java.io.IOException;
+import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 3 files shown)


phoenix-core-server/src/main/java/org/apache/phoenix/replication/ReplicationLogDiscovery.java
```diff
@@ -283,19 +295,54 @@ protected boolean shouldProcessInProgressDirectory() {
...
       haGroupName);
     long startTime = EnvironmentEdgeManager.currentTime();
-    List<Path> files = replicationLogTracker.getNewFilesForRound(replicationRound);
-    LOG.info("Number of new files for round {} is {}", replicationRound, files.size());
-    while (!files.isEmpty()) {
-      processOneRandomFile(files);
-      files = replicationLogTracker.getNewFilesForRound(replicationRound);
+    List<Path> allFiles = replicationLogTracker.getNewFilesForRound(replicationRound);
+    LOG.info("Number of new files for round {} is {}", replicationRound, allFiles.size());
+    long minAgeMs = getNewFileMinAgeMs();
```
tkhurana, line 311: This won't be changing right from one round to another right. We can avoid computing it everytime

phoenix-core-server/src/main/java/org/apache/phoenix/replication/ReplicationLogDiscovery.java
```diff
@@ -283,19 +295,54 @@ protected boolean shouldProcessInProgressDirectory() {
...
+    while (!allFiles.isEmpty()) {
```
tkhurana, line 312: I highly recommend adding isRunning check here.
```suggestion
    while (!allFiles.isEmpty() && isRunning) {
```

tkhurana: @Himanshu-g81 As discussed offline we may not need this PR but I still think we should add the isRunning check to prevent stalling RS stop operation.

Himanshu-g81: Sure, added in this PR itself - https://github.com/apache/phoenix/pull/2432/changes.

PR #2432, opened 2026-04-29
PHOENIX-7813 Align ReplicationLogDiscovery scheduler to round boundaries to fix file distribution imbalance across RS

Himanshu-g81:


phoenix-core-server/src/main/java/org/apache/phoenix/replication/ReplicationLogDiscovery.java
```diff
@@ -68,7 +68,7 @@ public abstract class ReplicationLogDiscovery {
   /**
    * Default interval in seconds between replay operations
    */
-  private static final long DEFAULT_REPLAY_INTERVAL_SECONDS = 10;
+  private static final long DEFAULT_REPLAY_INTERVAL_SECONDS = 60;
```
tkhurana: Do we need a separate replay interval config  or can we use just use the replay round value ? One less config to manage.

tkhurana (approved):

PR #2435, opened 2026-05-01
PHOENIX-7817 Resolve Omid transaction manager via provider singleton view

apurtell:
`OmidTransactionClient` captures the provider's `HBaseTransactionManager` at construction time. `OmidTransactionContext` in turn caches that same reference in a private field. When a prior test class shuts down OMID's TSO together with its mini-cluster, the next test class reuses the stale wrapper and is dispatched to the dead executor. The symptom is a `RejectedExecutionException` originating from the OMID TSOClient state-machine executor.
 
In production the singleton's manager is set exactly once and never replaced, but the test framework exercises the swap path between classes within a single Surefire/Failsafe fork.
 
Fix:
 
`OmidTransactionClient` becomes a thin view onto the singleton's current `HBaseTransactionManager`. It resolves
`provider.transactionManager` on every call to `getTransactionClient()`. `transactionManager` is marked volatile so test-framework injection is published to other threads without synchronization.

In production the dynamic lookup is functionally identical to caching.

### DIFF
--- a/phoenix-core-client/src/main/java/org/apache/phoenix/transaction/OmidTransactionContext.java
+++ b/phoenix-core-client/src/main/java/org/apache/phoenix/transaction/OmidTransactionContext.java
@@ -44,19 +44,24 @@ public class OmidTransactionContext implements PhoenixTransactionContext {
 
   private static final Logger LOGGER = LoggerFactory.getLogger(OmidTransactionContext.class);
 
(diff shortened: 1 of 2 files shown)


palashc:
https://ci-hadoop.apache.org/job/Phoenix/job/Phoenix-PreCommit-GitHub-PR/job/PR-2435/

PR #2437, opened 2026-05-01
PHOENIX-7819 HighAvailabilityGroup.URLS inner Set must be thread-safe

apurtell:
When multiple threads register HA URL infos for the same `HAGroupInfo`, `computeIfAbsent()` returns the same HashSet instance to every thread, which then concurrently call `add()` on it. `java.util.HashSet` is not thread-safe. Concurrent `add()` is undefined behavior and can lose entries.


palashc:
https://ci-hadoop.apache.org/job/Phoenix/job/Phoenix-PreCommit-GitHub-PR/job/PR-2437/3/

lokiore (approved):

PR #2439, opened 2026-05-01
PHOENIX-7822 MaxConcurrentConnectionsIT poll for async connection counts

apurtell:
`PhoenixConnection.close()` is asynchronous because internal child connections are reaped on a background thread. Counters are decremented on the close path with updates that may not be visible immediately to the caller after `close()` returns. Tests that assert against these counters immediately after the close call were racy. The fix is to use `HBaseTestingUtility.waitFor()` to poll until the counters reach the desired condition.


apurtell:
Rebased on current master.

palashc:
https://ci-hadoop.apache.org/job/Phoenix/job/Phoenix-PreCommit-GitHub-PR/job/PR-2439/3/

palashc (approved):
+1, pending build results

PR #2440, opened 2026-05-01
PHOENIX-7823 AuditLoggingIT replace hardcoded table names with generateUniqueName()

apurtell:
`AuditLoggingIT` uses hardcoded table names `test1`..`test6` a shared mini-cluster, causing `CREATE TABLE` collisions on CI retries and allowing `SYSTEM.LOG` queries to match stale entries from prior runs. The fix is to replace those names with `generateUniqueName()`everywhere they appear in SQL, and while in the file also rename the misspelled local `createqQery` -> `createQuery` and drop a stray `System.out.println` debug print.


palashc (approved):

palashc:
https://ci-hadoop.apache.org/job/Phoenix/job/Phoenix-PreCommit-GitHub-PR/job/PR-2440/2/

PR #2442, opened 2026-05-01
PHOENIX-7825 ConcurrentMutationsIT make MyClock.shouldAdvance volatile

apurtell:
`ConcurrentMutationsIT` installs a custom `EnvironmentEdge`, and the test thread calls `setAdvance(false)` to freeze time so two mutations share a timestamp. However, `MyClock.shouldAdvance` is not volatile, so the JMM does not require regionserver handler threads in the same JVM to observe the freeze, and in practice they keep advancing time, causing mutations expected to share a timestamp to diverge and producing non-deterministic index-scrutiny results, with the shared `ParallelStatsDisabled` mini-cluster amplifying the race.


palashc (approved):

palashc:
https://ci-hadoop.apache.org/job/Phoenix/job/Phoenix-PreCommit-GitHub-PR/job/PR-2442/2/

PR #2443, opened 2026-05-01
PHOENIX-7826 MetadataGetTableReadLockIT wait for coprocessor swap to land

apurtell:
The `testBlockedReadDoesNotBlockAnotherRead` test installs a `BlockingMetaDataEndpointImpl` coprocessor on `SYSTEM.CATALOG` and expects a worker thread's query to hit it and signal a latch the main thread is waiting on, but `TestUtil.addCoprocessor()` only updates the master `TableDescriptor` and doesn't wait for region servers to reopen the region with the new coprocessor, so if the query reaches the old `MetaDataEndpointImpl` first the latch is never signaled and the main thread hangs until the Surefire/Failsafe fork times out. The fix is to add a helper that polls each `SYSTEM.CATALOG` region's coprocessor host (via `findCoprocessor` for up to 10 seconds and call it after `addCoprocessor` so the coprocessor swap is guaranteed visible before the worker thread starts.

### DIFF
--- a/phoenix-core/src/it/java/org/apache/phoenix/end2end/MetadataGetTableReadLockIT.java
+++ b/phoenix-core/src/it/java/org/apache/phoenix/end2end/MetadataGetTableReadLockIT.java
@@ -23,6 +23,8 @@ import java.sql.DriverManager;
 import java.util.List;
 import java.util.Map;
 import java.util.concurrent.CountDownLatch;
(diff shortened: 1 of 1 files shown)


palashc (approved):

palashc:
https://ci-hadoop.apache.org/job/Phoenix/job/Phoenix-PreCommit-GitHub-PR/job/PR-2443/2/

PR #2445, opened 2026-05-01
PHOENIX-7828 De-flake UCFWithDisabledIndexIT

apurtell:
`UCFWithDisabledIndexIT` was flaky due to two independent races.

First, the tests swap `MetaDataEndpointImpl` on `SYSTEM.CATALOG` with a `TestMetaDataEndpointImpl` that throws `DoNotRetryIOException("Not allowed")` on every getTable RPC, but `TestUtil.addCoprocessor()` only polls until the master `TableDescriptor` is updated and does not wait for region servers to reopen `SYSTEM.CATALOG` with the new coprocessor. This creates five race sites where client work runs before the swap is visible.

Second, `testUcfWithNoGetTableCalls` created its data table with `UPDATE_CACHE_FREQUENCY=20000`, then performed 7 DDLs and 7 UPSERTs before a final `SELECT` that is expected to be served entirely from the client metadata cache. Whenever wall-clock delay between `CREATE TABLE` and the `SELECT` exceeded 20 s the cache expired, a getTable RPC was issued, and the stub rejected it.

Add a `waitForCoprocessorOnSystemCatalog(Class)` helper that polls `region.getCoprocessorHost().findCoprocessor(className)` on every `SYSTEM.CATALOG` region (up to 10 s) and call it after each of the five swaps.

Change `testUcfWithNoGetTableCalls`'s `CREATE TABLE` from `UPDATE_CACHE_FREQUENCY=20000` to `UPDATE_CACHE_FREQUENCY=NEVER`, which decouples the assertion from wall-clock time.

### DIFF
--- a/phoenix-core/src/it/java/org/apache/phoenix/end2end/UCFWithDisabledIndexIT.java
+++ b/phoenix-core/src/it/java/org/apache/phoenix/end2end/UCFWithDisabledIndexIT.java
@@ -27,9 +27,12 @@ import java.sql.DriverManager;
 import java.sql.ResultSet;
 import java.sql.SQLException;
 import java.sql.Statement;
+import java.util.List;
 import java.util.Map;
 import java.util.Properties;
 import org.apache.hadoop.hbase.DoNotRetryIOException;
+import org.apache.hadoop.hbase.TableName;
+import org.apache.hadoop.hbase.regionserver.HRegion;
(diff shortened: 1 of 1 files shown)


palashc (approved):

palashc:
https://ci-hadoop.apache.org/job/Phoenix/job/Phoenix-PreCommit-GitHub-PR/job/PR-2445/3/

PR #2448, opened 2026-05-01
PHOENIX-7831 PhoenixQueryTimeoutIT wait for DelayedRegionObserver to load

apurtell:
`testScanningResultIteratorQueryTimeoutForPagingWithVeryLowTimeout` attaches a `DelayedRegionObserver` to its data table so the test query reliably times out, but `addCoprocessor` only invokes `admin.modifyTable`, which returns as soon as the master `TableDescriptor` is updated and does not wait for region servers to reopen the affected regions with the new coprocessor. The query then runs immediately against the old observer chain, completes too quickly, and the timeout assertion fails. The fix polls `region.getCoprocessorHost().findCoprocessor()` on every region of the test table after `addCoprocessor` until the coprocessor is loaded.


palashc (approved):

palashc:
https://ci-hadoop.apache.org/job/Phoenix/job/Phoenix-PreCommit-GitHub-PR/job/PR-2448/2/

PR #2450, opened 2026-05-02
PHOENIX-7833 CDCBaseIT/CDCQueryIT keep manual edge installed after applyMutations

apurtell:
`CDCBaseIT.applyMutations` installs a `ManualEnvironmentEdge` so each `addChange` pins `EnvironmentEdgeManager.currentTimeMillis()` to the row's `changeTS`, then calls `committer.reset()` whose default implementation is `EnvironmentEdgeManager.reset()`, reverting `currentTimeMillis()` to the wall clock. The rest of the CDC test body easily takes longer than `CDCQueryIT.MAX_LOOKBACK_AGE = 10s`, so any SCN-bounded read whose smallest SCN is the earliest `changeTS` trips `QueryCompiler.verifySCN` with `ERROR 538 "Cannot use SCN to look further back in the past beyond the configured max lookback age"`. The `IndexTool` mapper shares the same `EnvironmentEdgeManager` static state in the in-process mini-cluster and `PhoenixServerBuildIndexInputFormat` hits the same `ERROR 538` while compiling its SCN-bounded plan. The fix stops calling `committer.reset()` at the end of `applyMutations`, instead pinning `injectEdge` the highest `changeTS` seen across all batches so subsequent SCN-bounded reads of the just-applied mutations stay inside the max-lookback window, and adds a `CDCQueryIT.afterTest()` that calls `EnvironmentEdgeManager.reset()` to preserve per-test isolation.

### DIFF
--- a/phoenix-core/src/it/java/org/apache/phoenix/end2end/CDCBaseIT.java
+++ b/phoenix-core/src/it/java/org/apache/phoenix/end2end/CDCBaseIT.java
@@ -396,15 +396,25 @@ public class CDCBaseIT extends ParallelStatsDisabledIT {
     String datatableName, String tid, List<Set<ChangeRow>> batches, String cdcName)
     throws Exception {
     EnvironmentEdgeManager.injectEdge(injectEdge);
+    long latestChangeTS = injectEdge.currentTime();
     try (Connection conn = committer.getConnection(tid)) {
       for (Set<ChangeRow> batch : batches) {
         for (ChangeRow changeRow : batch) {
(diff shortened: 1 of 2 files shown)


haridsv (approved):
Applied this along with #2451 and #2475 on my local workspace and ran for up to 50 iterations in a loop and didn't encounter any failure.

palashc:
https://ci-hadoop.apache.org/job/Phoenix/job/Phoenix-PreCommit-GitHub-PR/job/PR-2450/2/

PR #2453, opened 2026-05-02
PHOENIX-7836 UncoveredGlobalIndexRegionScanner2IT relax per-test query timeout

apurtell:
`testCount` upserts 100,000 rows, builds an uncovered global index, sets `phoenix.scanner.disable.uncovered.regions.move=false`, and runs an aggregate count that forces every region scan through the post-dummy region-move callback. For the `uncovered=false, salted=true` parameterization the per-region work is heavier and the global `phoenix.query.timeoutMs` set by `QueryServicesTestImpl` to `DEFAULT_THREAD_TIMEOUT_MS` (5 minutes) can be exhausted. The fix overrides the timeout only for this test's connection by passing a `Properties` with `phoenix.query.timeoutMs = 20 * 60 * 1000`.


palashc:
https://ci-hadoop.apache.org/job/Phoenix/job/Phoenix-PreCommit-GitHub-PR/job/PR-2453/2/

palashc (approved):
+1, pending build results

PR #2460, opened 2026-05-05
PHOENIX-7842 Reduce the shading contribution to build time

apurtell:
Phoenix's build time is dominated by four sibling shade executions (`phoenix-client-embedded`, `phoenix-client-lite`, `phoenix-server`, `phoenix-mapreduce-byo-shaded-hbase`). This change upgrades `maven-shade-plugin` to 3.6.2, which declares thread safety and fixes bugs, so `mvn -T 1C` parallelizes the four shades and they complete successfully, and turns off source-jar shading by default and reenables it under `-Prelease`.

It also makes `phoenix-mapreduce-byo-shaded-hbase` release-only. The module is moved out of the `shade-and-assembly profile` and into `-Prelease`, and `phoenix-assembly` is refactored so the mapreduce dependency is only pulled in only under `-Prelease`. `BUILDING.md` documents the new fast paths.

Verified end-to-end. 

`mvn -T 1C clean package` builds 15 modules with no mapreduce jar in the tarball

`mvn -T 1C -Prelease clean package` builds all 16 with the mapreduce uberjar bundled.

The build time is reduced from 26:37 to 11:27 on my M4 MacBook, a 2.32x speedup.

### DIFF
--- a/BUILDING.md
+++ b/BUILDING.md
@@ -69,6 +69,25 @@ and optionally, to just skip all the tests and build the jars:
 
 Note: javadocs are generated in target/apidocs
 
+Faster builds
+-------------
+
+The following options can dramatically reduce wall-clock time:
+
+* The four shade modules are reactor-siblings and the shade plugin is thread-safe, so they can build concurrently:
+
+  `$ mvn -T 1C clean package -DskipTests`
+
+  (`-T 1C` uses one build thread per CPU core. `-T 4` is also fine.)
+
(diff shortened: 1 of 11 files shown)


richardantal (approved):
LGTM
Its always nice to have a faster build time.

apurtell:
@virajjasani lmk if you have review comments

virajjasani (approved):

PR #2465, opened 2026-05-07
PHOENIX-7769 Initialize and Close replication log groups as part of RS startup and shutdown

tkhurana:
## Summary
- Eagerly initialize `ReplicationLogGroup` instances during RS startup (prewarm phase) when synchronous replication is enabled, so they are ready before the first mutation arrives.
- Close all `ReplicationLogGroup` instances belonging to a server during RS shutdown via a new `ReplicationLogGroup.closeAll(serverName)` method.
- Extract `serverName` capture from the test subclass into the parent `PhoenixRegionServerEndpoint`, removing redundant code.
- Fix flaky `ReplicationLogDiscoveryTest` by injecting an advancing clock so rename timestamps are always strictly before the threshold on the next loop iteration.
- Refactor `ReplicationLogBaseTest` to support configurable aligned vs full-round initial rotation delay, enabling `ReplicationLogDiscoveryForwarderTest` to use production-aligned rotation.

## Test plan
- [x] Existing unit tests pass: `ReplicationLogGroupTest`, `ReplicationLogDiscoveryTest`, `ReplicationLogDiscoveryForwarderTest`
- [x] Ran all HA related integration tests (HABaseIT subclasses)
- [x] Verify RS shutdown logs show `Closing all ReplicationLogGroup instances`
- [x] Verify failed eager init logs a warning and does not block startup

### DIFF
--- a/phoenix-core-server/src/main/java/org/apache/phoenix/coprocessor/PhoenixRegionServerEndpoint.java
+++ b/phoenix-core-server/src/main/java/org/apache/phoenix/coprocessor/PhoenixRegionServerEndpoint.java
@@ -17,6 +17,9 @@
  */
 package org.apache.phoenix.coprocessor;
 
+import static org.apache.phoenix.query.QueryServices.SYNCHRONOUS_REPLICATION_ENABLED;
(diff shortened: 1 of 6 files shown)


tkhurana:
@Himanshu-g81 fixed the flapper test

Himanshu-g81 (approved):
Thanks!

PR #2466, opened 2026-05-07
PHOENIX-7845 ReplicationLogGroup initialization resilience to standby cluster unavailability

tkhurana:
## Summary

- Defer peer shard manager creation from `ReplicationLogGroup.init()` to individual mode consumers (`SyncModeImpl`, `SyncAndForwardModeImpl`, `ReplicationLogDiscoveryForwarder`)
- When the standby namenode is unavailable at startup, the group gracefully degrades from SYNC to STORE_AND_FORWARD via the existing `updateModeOnFailure()` path
- Change `ReplicationLogGroup.get()` to throw `IOException` instead of `RuntimeException`, so callers in `IndexRegionObserver` get proper error classification (not misclassified as `IndexBuildingFailureException`)
- Forwarder lazily creates its peer shard manager with automatic retry on subsequent rounds
- Replace `createStandbyLog()`/`createFallbackLog()` with single `createReplicationLog(shardManager)` factory

## Test plan

- [x] `testInitDegradesToSafWhenPeerUnavailable` — peer unavailable at startup → SAF mode, writes succeed
- [x] `testInitFailsWhenLocalUnavailable` — local FS unavailable → init fails with IOException
- [x] `testForwarderRetriesPeerCreation` — forwarder retries peer shard manager on next round after initial failure
- [x] All 41 existing `ReplicationLogGroupTest` tests pass
- [x] All 3 `ReplicationLogDiscoveryForwarderTest` tests pass
- [x] All 2 `ReplicationLogTest` tests pass
- [x] HABaseIT integration tests (running)

### DIFF
--- a/phoenix-core-server/src/main/java/org/apache/phoenix/replication/ReplicationLogGroup.java
+++ b/phoenix-core-server/src/main/java/org/apache/phoenix/replication/ReplicationLogGroup.java
@@ -180,7 +180,7 @@ public class ReplicationLogGroup {
   protected final String haGroupName;
   protected final HAGroupStoreManager haGroupStoreManager;
   protected final MetricsReplicationLogGroupSource metrics;
-  protected ReplicationShardDirectoryManager peerShardManager;
+  protected HAGroupStoreRecord haGroupStoreRecord;
   protected ReplicationShardDirectoryManager localShardManager;
   protected ReplicationLogDiscoveryForwarder logForwarder;
   protected long syncTimeoutMs;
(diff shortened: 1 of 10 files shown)


phoenix-core-server/src/main/java/org/apache/phoenix/replication/ReplicationLogGroup.java
```diff
@@ -783,21 +800,58 @@ private ReplicationShardDirectoryManager createShardManager(String uri, String l
...
+          try {
+            return createPeerShardManager();
+          } catch (IOException e) {
+            throw new UncheckedIOException(e);
+          }
+        });
+      try {
+        peerShardManager = future.get(peerInitTimeoutMs, TimeUnit.MILLISECONDS);
+        return peerShardManager;
+      } catch (UncheckedIOException e) {
```
apurtell: Nit. future.get() won't throw UncheckedIOException. When the supplier throws unchecked, the future fails and Future.get() wraps the cause in ExecutionException. You can unwrap the UncheckedIOException from ExecutionException where you need it.

phoenix-core-server/src/main/java/org/apache/phoenix/replication/ReplicationLogGroup.java
```diff
@@ -783,21 +800,58 @@ private ReplicationShardDirectoryManager createShardManager(String uri, String l
...
+        throw e.getCause();
+      } catch (ExecutionException e) {
+        if (e.getCause() instanceof IOException) {
+          throw (IOException) e.getCause();
+        }
+        throw new IOException("Failed to create peer shard manager", e.getCause());
+      } catch (TimeoutException e) {
+        future.cancel(true);
```
apurtell: future.cancel(true) does not actually free the worker thread. (Per CompletableFuture Javadoc, mayInterruptIfRunning has no effect i.e. interrupts are not used to control processing.) The stall is moved to the worker pool, which is better, but the stall still holds resources the entire time the underlying HDFS call is blocked. Every subsequent call to this method spawns another supplyAsync task while the previous one is still hung, so this issue can compound and potentially exhaust the worker pool. 

I asked the robot about this and it recommended: _Use a dedicated ExecutorService (single-thread, bounded queue) so future.cancel(true) can actually interrupt, and so leakage is bounded._

phoenix-core-server/src/main/java/org/apache/phoenix/replication/ReplicationLogGroup.java
```diff
@@ -783,21 +800,58 @@ private ReplicationShardDirectoryManager createShardManager(String uri, String l
...
+   * creation is cached; subsequent calls return the cached instance. Bounded by
+   * {@link #REPLICATION_LOG_PEER_INIT_TIMEOUT_MS_KEY} to prevent blocking the disruptor handler
+   * thread on a peer NN outage.
+   */
+  protected ReplicationShardDirectoryManager getOrCreatePeerShardManager() throws IOException {
+    ReplicationShardDirectoryManager cached = peerShardManager;
+    if (cached != null) {
+      return cached;
+    }
+    synchronized (this) {
```
apurtell, line 820: With both the disruptor handler and the forwarder thread involved we get a mutual exclusion here that may be longer than 10s.

apurtell (approved):
+1 with some minor comments

PR #2467, opened 2026-05-11
PHOENIX-7848. Use ZK TLS properties from HBase config if present

anmolnar:

Parse ZooKeeper TLS properties from `hbase-site.xml` and add them to `sqlline` command line args.

Phoenix `sqlline` thick client is unable to connect to TLS-only ZooKeeper ensemble.

No.

Locally.

No.

### DIFF
--- a/bin/phoenix_utils.py
+++ b/bin/phoenix_utils.py
@@ -129,20 +129,27 @@ def setPath():
             # Try to provide something valid
             hbase_conf_dir = '.'
 
-    global zk_tls_args
-    root = ET.parse(os.path.join(hbase_conf_dir, "hbase-site.xml")).getroot()
-    zk_hbase_prefix = "hbase.zookeeper.property."
-    zkcfg = {
-        prop.find("name").text[len(zk_hbase_prefix):]: prop.find("value").text
-        for prop in root.findall("property")
-        if prop.find("name").text.startswith(zk_hbase_prefix)
-    }
-    if zkcfg.get('client.secure').lower() == 'true':
-        zk_tls_args = '-Dzookeeper.client.secure=true ' + \
-            '-Dzookeeper.clientCnxnSocket=' + zkcfg['clientCnxnSocket'] + ' ' + \
-            '-Dzookeeper.ssl.trustStore.location=' + zkcfg['ssl.trustStore.location'] + ' ' + \
-            '-Dzookeeper.ssl.trustStore.type=' + zkcfg['ssl.trustStore.type'] + ' ' + \
-            '-Dzookeeper.ssl.trustStore.password=' + zkcfg['ssl.trustStore.password'] + ' '
+    hbase_site_file = os.path.join(hbase_conf_dir, "hbase-site.xml")
+    try:
+        global zk_tls_args
+        root = ET.parse(hbase_site_file).getroot()
+        zk_hbase_prefix = "hbase.zookeeper.property."
+        zkcfg = {
+            prop.find("name").text[len(zk_hbase_prefix):]: prop.find("value").text
+            for prop in root.findall("property")
+            if prop.find("name").text.startswith(zk_hbase_prefix)
+        }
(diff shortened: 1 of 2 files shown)


bin/phoenix_utils.py
```diff
@@ -126,6 +127,21 @@ def setPath():
...
+        for prop in root.findall("property")
+        if prop.find("name").text.startswith(zk_hbase_prefix)
+    }
+    if zkcfg.get('client.secure').lower() == 'true':
+        zk_tls_args = '-Dzookeeper.client.secure=true ' + \
+            '-Dzookeeper.clientCnxnSocket=' + zkcfg['clientCnxnSocket'] + ' ' + \
+            '-Dzookeeper.ssl.trustStore.location=' + zkcfg['ssl.trustStore.location'] + ' ' + \
+            '-Dzookeeper.ssl.trustStore.type=' + zkcfg['ssl.trustStore.type'] + ' ' + \
+            '-Dzookeeper.ssl.trustStore.password=' + zkcfg['ssl.trustStore.password'] + ' '
+
```
richardantal, line 153: Can you please add Exception handling to this?
We want to continue if there is some problem with parsing the zookeeper parameters.

Please add a default "" value to zk_tls_args in case we don't have hbase.zookeeper.property.client.secure property.

bin/sqlline.py
```diff
@@ -94,6 +94,7 @@ def kill_child():
 
 java_cmd = phoenix_utils.java + ' ' + phoenix_utils.jvm_module_flags + \
     ' ' + opts + \
+    ' ' + getattr(phoenix_utils, "zk_tls_args", "") + \
```
richardantal: If we make sure zk_tls_args has some value, we could have `phoenix_utils.zk_tls_args` here

richardantal (changes requested):

bin/sqlline.py
anmolnar: No, we cannot. If it doesn't get any value in the utils, it will be undefined and this code would throw an exception (tested).

bin/phoenix_utils.py
anmolnar: Default "" doesn't make any difference for this field. The field will be undefined. We have two options: use the `getattr` call as below, or define `zk_tls_args` at top level of this file.

Certainly I can add exception handler, but there's no exception handling anywhere else in this file. Does it make sense?

richardantal: I think so, yes 
There are much more thinks that could go wrong here.

richardantal: I prefer the later option, but I other variables are not defined at top level of the file, just with `global`

anmolnar:
@richardantal I moved `zk_tls_args` to module level in order to be defined at all times.

I'm not sure about the exception handler. What should be the logic inside the exception handler? I don't want the error to be swallowed and ignored, because if there's anything wrong with the configuration, the tool should refuse to work.

bin/phoenix_utils.py
```diff
@@ -126,6 +129,21 @@ def setPath():
...
 
+    global zk_tls_args
+    root = ET.parse(os.path.join(hbase_conf_dir, "hbase-site.xml")).getroot()
+    zk_hbase_prefix = "hbase.zookeeper.property."
+    zkcfg = {
+        prop.find("name").text[len(zk_hbase_prefix):]: prop.find("value").text
+        for prop in root.findall("property")
+        if prop.find("name").text.startswith(zk_hbase_prefix)
+    }
+    if zkcfg.get('client.secure').lower() == 'true':
```
richardantal: We should check if '`client.secure' in zkcfg` too

bin/phoenix_utils.py
richardantal: There could be better error when some of the keys are missing

anmolnar: This is done @richardantal

richardantal (approved):
LGTM

Thanks Andor for this change.

PR #2474, opened 2026-05-13
PHOENIX-7854 Use Abortable interface to trigger RS aborts

tkhurana:
## Summary
  - Replace RuntimeException-based abort with the HBase Abortable interface for triggering real RS shutdowns on fatal replication errors
  - Annotate PhoenixRegionServerEndpoint and IndexRegionObserver with @CoreCoprocessor to access HasRegionServerServices and obtain an Abortable (RegionServerServices), threaded into ReplicationLogGroup
  - On unrecoverable errors the Disruptor event handler sets a fatalException field and fails pending sync futures, but never calls abort() or close() itself — the
  producer thread (RPC handler) that receives the failed future initiates the abort
  - Wrap onEvent() in a top-level catch (Throwable) so exceptions never escape to LogExceptionHandler, which becomes a safety net
  - Simplify close(boolean graceful) into a single close() that always drains with a timeout — when the handler has a fatal exception the drain completes instantly since each event hits the short-circuit path;

## Test plan
- [x] Unit tests in `ReplicationLogGroupTest` updated to expect `IOException` / `PhoenixWALSyncTimeoutException` instead of `RuntimeException`
- [x] All 18 HABaseIT integration tests pass
- [x] Verify graceful close drains the Disruptor with a bounded timeout before halting

### DIFF
--- a/phoenix-core-server/src/main/java/org/apache/phoenix/coprocessor/PhoenixRegionServerEndpoint.java
+++ b/phoenix-core-server/src/main/java/org/apache/phoenix/coprocessor/PhoenixRegionServerEndpoint.java
@@ -84,6 +84,7 @@ public class PhoenixRegionServerEndpoint extends
     if (env instanceof RegionServerCoprocessorEnvironment) {
       this.serverName = ((RegionServerCoprocessorEnvironment) env).getServerName();
     }
+    // @CoreCoprocessor guarantees HasRegionServerServices, but guard for testability
     if (env instanceof HasRegionServerServices) {
       this.abortable = ((HasRegionServerServices) env).getRegionServerServices();
     }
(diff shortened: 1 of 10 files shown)


phoenix-core-server/src/main/java/org/apache/phoenix/coprocessor/PhoenixRegionServerEndpoint.java
```diff
@@ -79,6 +84,9 @@ public void start(CoprocessorEnvironment env) throws IOException {
     if (env instanceof RegionServerCoprocessorEnvironment) {
       this.serverName = ((RegionServerCoprocessorEnvironment) env).getServerName();
     }
+    if (env instanceof HasRegionServerServices) {
```
kadirozde, line 88: Is this check necessary? Are there cases where env is not an instance of RegionServerCoprocessorEnvironment here or is this just defensive coding? Can we add a comment here to clarify?

kadirozde (approved):
+1, Thank you!

PR #2482, opened 2026-05-19
PHOENIX-7859 Make ParallelPhoenixConnectionFallbackIT deterministic

mnpoonia:
## What is the purpose of this change?

Fix intermittent timeout in `ParallelPhoenixConnectionFallbackIT.testParallelConnectionBackoff` by increasing `waitFor` timeout ceilings to tolerate CI host CPU starvation.

## Background

The test uses `waitFor()` with a 5s ceiling throughout. On a busy Yetus worker, newly created thread pools may not be scheduled for several seconds, causing the condition to never be observed within the deadline. The 5s limit was chosen by eye and offers very little margin under resource contention.

`waitFor()` exits as soon as the condition becomes true, so raising the ceiling costs nothing on fast machines — it only provides headroom when the CI host is under load.

## What changes did I make?

Raised all `waitFor` timeout ceilings from 5000ms to 60000ms.

## Related Issues

- Fixes: [PHOENIX-7859](https://issues.apache.org/jira/browse/PHOENIX-7859)
- Related: [PHOENIX-6840](https://issues.apache.org/jira/browse/PHOENIX-6840) (flaky ParallelPhoenix tests)

### DIFF
--- a/phoenix-core/src/it/java/org/apache/phoenix/jdbc/ParallelPhoenixConnectionFallbackIT.java
+++ b/phoenix-core/src/it/java/org/apache/phoenix/jdbc/ParallelPhoenixConnectionFallbackIT.java
@@ -23,19 +23,16 @@ import static org.apache.phoenix.jdbc.HighAvailabilityGroup.PHOENIX_HA_GROUP_ATT
 import static org.apache.phoenix.jdbc.HighAvailabilityPolicy.PARALLEL;
 import static org.apache.phoenix.jdbc.HighAvailabilityTestingUtility.HBaseTestingUtilityPair;
 import static org.apache.phoenix.jdbc.HighAvailabilityTestingUtility.doTestBasicOperationsWithConnection;
-import static org.junit.Assert.assertFalse;
 import static org.junit.Assert.assertTrue;
 
 import java.sql.Connection;
 import java.sql.DriverManager;
 import java.sql.SQLException;
-import java.util.List;
 import java.util.Properties;
 import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
 import java.util.concurrent.Future;
-import java.util.concurrent.ThreadPoolExecutor;
 import java.util.function.Supplier;
 import org.apache.phoenix.end2end.NeedsOwnMiniClusterTest;
 import org.junit.AfterClass;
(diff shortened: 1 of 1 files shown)


mnpoonia:
@virajjasani

phoenix-core/src/it/java/org/apache/phoenix/jdbc/ParallelPhoenixConnectionFallbackIT.java
```diff
@@ -108,9 +111,27 @@ public void testParallelConnectionBackoff() throws Exception {
...
+    waitFor(() -> {
+      List<PhoenixHAExecutorServiceProvider.PhoenixHAClusterExecutorServices> services =
+          PhoenixHAExecutorServiceProvider.get(PROPERTIES);
+      int queueSize1 = ((ThreadPoolExecutor) services.get(0).getExecutorService()).getQueue().size();
+      int queueSize2 = ((ThreadPoolExecutor) services.get(1).getExecutorService()).getQueue().size();
+
+      LOG.debug("Waiting for queues to fill: cluster1 queue={}, cluster2 queue={}",
+          queueSize1, queueSize2);
+
+      return queueSize1 >= 1 && queueSize2 >= 1;
```
apurtell: If any future change adjusts `HA_MAX_QUEUE_SIZE` this test may pass when it should fail (the wait completes earlier than the fallback signal actually flips) or fail in confusing ways. Consider deriving the trigger from the configured values, e.g.:

```java
int maxQueue = Integer.parseInt(
    PROPERTIES.getProperty(PhoenixHAExecutorServiceProvider.HA_MAX_QUEUE_SIZE));
double threshold = Double.parseDouble(
    PROPERTIES.getProperty(
        PhoenixHAExecutorServiceProvider.HA_THREADPOOL_QUEUE_BACKOFF_THRESHOLD));
int trigger = (int) Math.ceil(threshold * maxQueue);
```

... or just keep `hasCapacity()` as the wait predicate

apurtell (changes requested):
The stated root cause is at least partly wrong? 

Even if one `hasCapacity()` call returns a stale answer because of the non-atomic read pair, the next poll will see the update. The test polls 50 times over 5s. Maybe detection is delayed by one tick (~100ms). It cannot cause a 5s timeout. There must be some other reason?

Be sure to run `mvn spotless:apply` before posting an update or this will fail precommit or cause precommits to fail post merge.

apurtell:
FWIW @mnpoonia I put this to my robot and it suggests CI host CPU starvation is the cause because the timeouts are too tight. 

> On a busy Yetus worker, the brand-new single-thread pool created on line 100 of the test may not be scheduled for several seconds. The 5s timeout is just too tight. It was almost certainly chosen by eye, not by measurement, and offers very little margin.

mnpoonia:
Thanks @apurtell In that case i think we should increase timeout here. We are checking every 100 ms so it will not impact normal runs but will help in overloaded CI case.

mnpoonia:
@apurtell A gentle reminder.

PR #2489, opened 2026-05-27
PHOENIX-7863 Add replication consistency point guard to compaction

Himanshu-g81:
On standby clusters, compaction can prematurely purge delete markers that have timestamps newer than the replication consistency point. This causes permanent stale data because when replay eventually catches up, the delete markers that should have been applied are already gone.

This PR adds a compaction guard that floors `maxLookbackWindowStart` to the minimum consistency point across all HA groups, ensuring delete markers newer than the consistency point are retained until replay has processed them.

The guard is controlled by `phoenix.replication.compaction.guard.enabled` (default `true`). It activates automatically when `phoenix.replication.replay.enabled=true` — no additional configuration needed on standby clusters. The separate flag allows operators to disable the guard independently if needed without turning off replay entirely.


apurtell:
> `phoenix.replication.compaction.guard.enabled`

Given this is a critical safety feature to prevent data desynchronization, when would it ever be disabled?

phoenix-core-server/src/main/java/org/apache/phoenix/replication/reader/ReplicationLogReplayService.java
```diff
@@ -229,6 +251,37 @@ protected long getConsistencyPoint() throws IOException, SQLException {
...
+      LOG.warn("Replication guard enabled but consistency point unavailable for table={} store={}."
+        + " Retaining all delete markers.", tableName, columnFamilyName, e);
+      return 0L;
+    }
+  }
+
+  @VisibleForTesting
+  static long adjustMaxLookbackWindowStart(long currentMaxLookbackWindowStart,
+    long consistencyPoint, String tableName, String columnFamilyName) {
+    long adjusted = Math.min(currentMaxLookbackWindowStart, consistencyPoint);
```
apurtell: You are changing the store-level `maxLookbackWindowStart` but the value actually used for retention is computed per row as the max of this and the `ttlWindowStart` of the Phoenix compactor's `RowContext`. When `ttlWindowStart` in the `RowContext` is greater than `consistencyPoint` the boundary snaps to `ttlWindowStart` and delete markers between `consistencyPoint` and `ttlWindowStart` get purged anyway, which I think you are trying to prevent.

The robot pointed me to this issue with this text:
"The guard is effective only when replay lag stays within the TTL window. Concretely: table TTL = 1h, replay lag = 2h ⇒ consistencyPoint = now-2h, guard sets store start to now-2h, but the row uses max(now-1h, now-2h) = now-1h, so markers in the [now-2h, now-1h] band are dropped. Tests happen to pass only because the test table has no TTL."

phoenix-core-server/src/main/java/org/apache/phoenix/replication/reader/ReplicationLogReplayService.java
```diff
@@ -229,6 +251,37 @@ protected long getConsistencyPoint() throws IOException, SQLException {
...
 
+  /**
+   * Applies the replication replay consistency point as a floor on maxLookbackWindowStart. On
+   * standby clusters, this prevents compaction from dropping delete markers that have timestamps
+   * newer than the consistency point.
+   */
+  public static long applyReplicationConsistencyGuard(long currentMaxLookbackWindowStart,
+    Configuration conf, String tableName, String columnFamilyName) {
+    try {
+      long consistencyPoint = getInstance(conf).getConsistencyPoint();
```
apurtell: This will iterate every HA group and in SYNC state each call hits the filesystem. That is at least an `exists()` + `listStatus()` RPC to the NameNode per HA group, executed once per store per compaction, synchronously, inside the `CompactionScanner` constructor. 

The consistency point moves slowly relative to compaction frequency, so consider caching it with a short TTL of a few seconds and generally using the cached value, rather than recomputing on every scanner construction.

phoenix-core-server/src/main/java/org/apache/phoenix/replication/reader/ReplicationLogReplayService.java
```diff
@@ -50,6 +51,17 @@ public class ReplicationLogReplayService {
    */
   public static final boolean DEFAULT_REPLICATION_REPLAY_ENABLED = false;
 
+  /**
+   * Configuration key for enabling/disabling the replication compaction guard
+   */
+  public static final String REPLICATION_COMPACTION_GUARD_ENABLED =
+    "phoenix.replication.compaction.guard.enabled";
```
apurtell: I commented on this elsewhere, but Claude called this a "foot gun", lol.

>  I'd argue the flag is a foot-gun: setting `phoenix.replication.compaction.guard.enabled=false` on a standby silently re-introduces exactly the data-desync bug this PR wants to fix.

apurtell (changes requested):

phoenix-core-server/src/main/java/org/apache/phoenix/replication/reader/ReplicationLogReplayService.java
```diff
@@ -229,6 +251,37 @@ protected long getConsistencyPoint() throws IOException, SQLException {
...
+
+  @VisibleForTesting
+  static long adjustMaxLookbackWindowStart(long currentMaxLookbackWindowStart,
+    long consistencyPoint, String tableName, String columnFamilyName) {
+    long adjusted = Math.min(currentMaxLookbackWindowStart, consistencyPoint);
+    if (adjusted < currentMaxLookbackWindowStart) {
+      LOG.info(
+        "Replication guard: table={} store={} maxLookbackWindowStart adjusted from {} to {}"
+          + " (consistencyPoint={})",
+        tableName, columnFamilyName, currentMaxLookbackWindowStart, adjusted, consistencyPoint);
```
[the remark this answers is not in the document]
apurtell: Fix this!!

Himanshu-g81:
> > `phoenix.replication.compaction.guard.enabled`
> 
> Given this is a critical safety feature to prevent data desynchronization, when would it ever be disabled?

The idea was if we have any issue in consistency point calculation, etc and we need to disable this check specificailly to avoid issues with compaction temporarily (since max-look back anyway hold the delete markers upto 3 days, or we update that config and disable this one to remediate the issue).
So this is just to have more control over compaction behaviour.

phoenix-core-server/src/main/java/org/apache/phoenix/replication/reader/ReplicationLogReplayService.java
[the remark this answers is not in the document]
Himanshu-g81: Done in subsequent commit

phoenix-core-server/src/main/java/org/apache/phoenix/replication/reader/ReplicationLogReplayService.java
```diff
@@ -105,6 +117,16 @@ public static ReplicationLogReplayService getInstance(Configuration conf) throws
...
 
+  @VisibleForTesting
+  public static void setInstanceForTesting(ReplicationLogReplayService testInstance) {
+    instance = testInstance;
+  }
+
+  @VisibleForTesting
+  public static void resetInstanceForTesting() {
+    instance = null;
+  }
```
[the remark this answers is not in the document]
Himanshu-g81: It's used in phoenix-core module since all the tests are in same module.

phoenix-core-server/src/main/java/org/apache/phoenix/replication/reader/ReplicationLogReplayService.java
Himanshu-g81: Fixed, thanks!

phoenix-core-server/src/main/java/org/apache/phoenix/replication/reader/ReplicationLogReplayService.java
Himanshu-g81: https://github.com/apache/phoenix/pull/2489#issuecomment-4680562199

phoenix-core-server/src/main/java/org/apache/phoenix/replication/reader/ReplicationLogReplayService.java
```diff
@@ -77,14 +91,33 @@ public class ReplicationLogReplayService {
...
+      try {
+        return getConsistencyPoint();
+      } catch (IOException | SQLException e) {
+        throw new RuntimeException("Failed to fetch consistency point", e);
+      }
+    }, CONSISTENCY_POINT_CACHE_TTL_SECONDS, TimeUnit.SECONDS);
+  }
+
+  private ReplicationLogReplayService(long fixedConsistencyPoint) {
+    this.cachedConsistencyPoint = () -> fixedConsistencyPoint;
```
tkhurana, line 119: Shouldn't the final field conf been initialized ?

phoenix-core-server/src/main/java/org/apache/phoenix/replication/reader/ReplicationLogReplayService.java
```diff
@@ -77,14 +91,33 @@ public class ReplicationLogReplayService {
...
+  }
+
+  private ReplicationLogReplayService(Supplier<Long> supplier) {
+    this.cachedConsistencyPoint = Suppliers.memoizeWithExpiration(supplier,
```
tkhurana: Shouldn't the final field conf been initialized ?

tkhurana:
When the cache is cold or expired, the delegate calls getConsistencyPoint() →
  ReplicationLogDiscoveryReplay.getConsistencyPoint(), which reads lastRoundInSync (ReplicationLogDiscoveryReplay.java:570-574, 584-585):

  private ReplicationRound lastRoundInSync;   // line 106 — plain field, NOT volatile

  That field is written by the background replay thread (setLastRoundInSync at :328, and init at :242/:249) and has no volatile modifier and no synchronization (:438-448).
  Before this PR, getConsistencyPoint() was only ever called from the single replay scheduler thread (REPLICATION_REPLAY_SERVICE_EXECUTOR_THREAD_COUNT = 1, invoked at
  :287/:349), so the field was effectively thread-confined.

  This PR makes HBase compaction threads (the large/small compaction pools — commonly >1, so multiple CompactionScanner constructors run in parallel on one RS) call
  getConsistencyPoint() for the first time from outside that thread. The synchronized(this) inside the Guava supplier serializes the compaction threads against each other,
  but it locks the supplier object — it establishes no happens-before with the replay thread's writes to lastRoundInSync. Consequences:

  - Visibility: a compaction thread can observe a stale lastRoundInSync, or even null after the replay thread has set it — which trips the throw new
  IOException("...lastRoundInSync is not initialized") at :577/:588 → resolveConsistencyPoint catches it → returns 0L → retain everything (the silent fail-closed path from
  the main review). So the race doesn't corrupt data, but it can spuriously and silently disable purging.
  - replicationReplayState is an AtomicReference (:109) so that read is fine; lastRoundInSync is the unprotected one.

phoenix-core-server/src/main/java/org/apache/phoenix/replication/reader/ReplicationLogReplayService.java
```diff
@@ -77,14 +91,33 @@ public class ReplicationLogReplayService {
    */
   public static final int DEFAULT_REPLICATION_REPLAY_SERVICE_EXECUTOR_SHUTDOWN_TIMEOUT_SECONDS = 30;
 
+  private static final long CONSISTENCY_POINT_CACHE_TTL_SECONDS = 30;
+
   private static volatile ReplicationLogReplayService instance;
 
   private final Configuration conf;
```
apurtell, line 95: The two test-only constructors in `ReplicationLogReplayService` do not initialize the blank-final field `conf`. This will break the build. 

```
ReplicationLogReplayService.java:116: error: variable conf might not have been initialized
ReplicationLogReplayService.java:121: error: variable conf might not have been initialized
```

phoenix-core-server/src/main/java/org/apache/phoenix/replication/reader/ReplicationLogReplayService.java
```diff
@@ -77,14 +91,33 @@ public class ReplicationLogReplayService {
...
   private ScheduledExecutorService scheduler;
   private volatile boolean isRunning = false;
+  private final Supplier<Long> cachedConsistencyPoint;
 
   private ReplicationLogReplayService(final Configuration conf) {
     this.conf = conf;
+    this.cachedConsistencyPoint = Suppliers.memoizeWithExpiration(() -> {
```
apurtell, line 108: `Suppliers.memoizeWithExpiration` uses `synchronized(this)` to serialize loads, but that lock is on the supplier object. This establishes no happens-before relationship with the replay thread's writes to `lastRoundInSync`.

So you may end up advancing the consistency point too far back, or even getting a a `null` back from the supplier even after the replay thread has already set it, spuriously throwing an `IOException`.  The net effect is sometimes a major compaction won't drop delete markers when it should.

*Something* needs to enforce the happens-before relationship. Perhaps `volatile ReplicationRound lastRoundInSync` , or consider `AtomicReference<ReplicationRound>`. Or redo this.

phoenix-core-server/src/main/java/org/apache/phoenix/replication/reader/ReplicationLogReplayService.java
apurtell: This is unnecessary configuration and an accident waiting to happen.

apurtell (changes requested):
I highlighted a build break and consistency issue. Both must be fixed. In addition the tests still don't cover the TTL case that motivated the rewrite.

I asked the robot to suggest necessary test improvements:

Add back a unit test on `computeRowMaxLookbackWithGuard(...)` covering at least:
- `ttlWindowStart` > `maxLookbackWindowStart` > `consistencyPoint` (TTL-dominated, guard caps)
- `maxLookbackWindowStart` > `ttlWindowStart` > `consistencyPoint` (lookback-dominated, guard caps)
- `consistencyPoint` > both (guard inactive)
- `consistencyPoint` = 0 (fallback)
- `consistencyPoint` = `Long.MAX_VALUE` (guard disabled at call site)

Add an IT case to `CompactionReplicationGuardIT` that creates the table with TTL = N (using HBase CF TTL via ALTER or Phoenix TTL property), sets `consistencyPoint` = `beforeDeleteTime - 1`, advances past TTL by M >> N, and asserts the delete marker is retained. Without this, the row-level fix is only theoretically tested.

phoenix-core-server/src/main/java/org/apache/phoenix/replication/reader/ReplicationLogReplayService.java
```diff
@@ -77,14 +81,35 @@ public class ReplicationLogReplayService {
...
+      try {
+        return getConsistencyPoint();
+      } catch (IOException | SQLException e) {
+        throw new RuntimeException("Failed to fetch consistency point", e);
+      }
+    }, CONSISTENCY_POINT_CACHE_TTL_SECONDS, TimeUnit.SECONDS);
+  }
+
+  private ReplicationLogReplayService(long fixedConsistencyPoint) {
+    this.conf = null;
```
apurtell: If we need a conf to vary our behavior, it should be passed in by a higher layer. If we need a conf because some API we call needs it, we should create it with `new Configuration()`.

phoenix-core-server/src/main/java/org/apache/phoenix/replication/reader/ReplicationLogReplayService.java
```diff
@@ -77,14 +81,35 @@ public class ReplicationLogReplayService {
...
+    this.cachedConsistencyPoint = () -> fixedConsistencyPoint;
+  }
+
+  private ReplicationLogReplayService(Supplier<Long> supplier) {
+    this.conf = null;
```
apurtell: If we need a conf to vary our behavior, it should be passed in by a higher layer. If we need a conf because some API we call needs it, we should create it with `new Configuration()`.

apurtell (approved):
Looks like the substantiative issues have been addressed. There is one nit I'd like to see improved but this is looking good to me, so I have approved it.

phoenix-core-server/src/main/java/org/apache/phoenix/replication/reader/ReplicationLogReplayService.java
apurtell: The robot says

Any code path on the test singleton that touches conf will NPE:

start() reads conf.getBoolean(...) (line 155)
stop() reads conf.getBoolean(...) (line 190)
getReplicationGroups() calls HAGroupStoreManager.getInstance(conf) (line 295)
getReplicationLogReplay(...) calls ReplicationLogReplay.get(conf, ...) (line 299)

Today this is safe only because the test ITs and UTs go straight through cachedConsistencyPoint.get() and never call the lifecycle methods.

phoenix-core-server/src/main/java/org/apache/phoenix/replication/reader/ReplicationLogReplayService.java
apurtell: See above

tkhurana (approved):

phoenix-core-server/src/main/java/org/apache/phoenix/replication/reader/ReplicationLogReplayService.java
Himanshu-g81: Sure, removed i subsequent commit

PR #2491, opened 2026-05-28
PHOENIX-7871 :- Fail-fast batched mutation block via DNRIOE inheritance

lokiore:

The two HA failover signaling exception classes — `MutationBlockedIOException` and `StaleClusterRoleRecordException` — now extend `org.apache.hadoop.hbase.DoNotRetryIOException` instead of `java.io.IOException`. This is sufficient to deliver fail-fast on the batched mutation path; no Phoenix-side intercept is needed.

JIRA: https://issues.apache.org/jira/browse/PHOENIX-7871

Pre-fix, the HA failover exceptions extended `IOException`. When the server-side mutation-block gate fired, HBase's RPC retry layers had no signal to fail fast — the failure was wrapped in `RetriesExhaustedWithDetailsException` only after the client retry budget was exhausted (default 16 retries × per-retry timeout = tens of seconds). For mutations issued during a brief mutation-block window or against a now-STANDBY cluster, this manifested as multi-second tail latencies that should have been single-RTT failures.

With the inheritance change, server-side rehydration (via `ProtobufUtil.toException`) delivers a real `MutationBlockedIOException` instance to HBase's batched-RPC retry layer (`AsyncRequestFutureImpl.manageError`); the `instanceof DoNotRetryIOException` check at line 749 returns true post-inheritance, so no retries fire and the failure surfaces immediately.

An earlier draft of this PR included a Phoenix-side intercept in `MutationState.send` (a `findHaFailoverCauseInChain` helper + a config flag `phoenix.ha.failfast.failover.exceptions.enabled`). That intercept was empirically redundant — it never fired in practice because HBase already short-circuits via the inheritance path. This is the slimmed-down version: inheritance-only.

No

The exception-inheritance change is internal — both exception classes still satisfy `instanceof IOException`, so any existing `catch (IOException)` site continues to match. The user-visible behavior change is reduced tail latency on the failover path, which is a strict improvement.

**New unit tests** (`HaFailoverExceptionInheritanceTest`, 3 tests, all PASS) — verifies the inheritance change:
- `mutationBlockedExtendsDoNotRetryIOException`
- `staleClusterRoleRecordExtendsDoNotRetryIOException`
- `staleClusterRoleRecordTwoArgConstructorPreserved`

**IT** (`IndexRegionObserverMutationBlockingIT`, 7 tests, all PASS):
- 4 baseline mutation-blocking tests (data table with index, allowed-when-not-blocked, transition, system HA group exemption)
- `testMutationBlockedFailsFastWithDNRIOE` — wall-clock <10s assertion under default retry config
- `testMutationBlockedFailsFastUnderElevatedRetries` — wall-clock <10s assertion under `hbase.client.retries.number=16` + `hbase.client.pause=100`
- `testMutationBlockedFailsFastViaInheritanceAlone` — crisp empirical proof of inheritance-alone fail-fast under elevated retry config

Local mvn output:
```
[INFO] --- failsafe:3.2.2:integration-test (default-cli) @ phoenix-core ---
[INFO] Running org.apache.phoenix.end2end.index.IndexRegionObserverMutationBlockingIT
[INFO] Running org.apache.phoenix.exception.HaFailoverExceptionInheritanceTest
[INFO] Tests run: 3, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 0.034 s
[INFO] Tests run: 7, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 107.3 s
[INFO] Tests run: 10, Failures: 0, Errors: 0, Skipped: 0
[INFO] BUILD SUCCESS
[INFO] Total time:  01:55 min
```

Generated-by: Claude Code (Opus 4.7)

### DIFF
--- a/phoenix-core/src/it/java/org/apache/phoenix/end2end/index/IndexRegionObserverMutationBlockingIT.java
+++ b/phoenix-core/src/it/java/org/apache/phoenix/end2end/index/IndexRegionObserverMutationBlockingIT.java
@@ -277,12 +277,222 @@ public class IndexRegionObserverMutationBlockingIT extends HABaseIT {
     }
   }
 
+  /**
+   * Walks the cause chain of a {@link CommitException} looking for a
+   * {@link MutationBlockedIOException}. Handles two paths:
+   * <ul>
+   *   <li>Direct chain — {@code MutationBlockedIOException} now extends
+   *       {@link org.apache.hadoop.hbase.DoNotRetryIOException}, so HBase's RPC retry layers
+   *       fail-fast and propagate the exception without wrapping it in
+   *       {@link RetriesExhaustedWithDetailsException}.</li>
+   *   <li>REWDE-wrapped — the legacy retry path; preserved here for any caller that still
+   *       reaches it (defense-in-depth).</li>
+   * </ul>
+   */
   private boolean containsMutationBlockedException(CommitException e) {
-    Throwable cause = e.getCause();
-    if (cause instanceof RetriesExhaustedWithDetailsException) {
-      RetriesExhaustedWithDetailsException re = (RetriesExhaustedWithDetailsException) cause;
-      return re.getCause(0) instanceof MutationBlockedIOException;
+    Throwable t = e.getCause();
+    while (t != null) {
+      if (t instanceof MutationBlockedIOException) {
+        return true;
+      }
+      if (t instanceof RetriesExhaustedWithDetailsException) {
+        RetriesExhaustedWithDetailsException re = (RetriesExhaustedWithDetailsException) t;
+        for (int i = 0; i < re.getNumExceptions(); i++) {
+          if (re.getCause(i) instanceof MutationBlockedIOException) {
+            return true;
+          }
+        }
+      }
+      t = t.getCause();
     }
     return false;
   }
+
+  /**
+   * Regression test for the fail-fast fix on the batched-mutation path. The fix is the inheritance
+   * change on {@link MutationBlockedIOException} — it now extends
+   * {@code DoNotRetryIOException}, signaling intent to fail fast.
+   *
+   * <p>Empirical verification: server-side rehydration (via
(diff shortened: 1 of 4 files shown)


phoenix-core/src/it/java/org/apache/phoenix/end2end/index/IndexRegionObserverMutationBlockingIT.java
```diff
@@ -277,12 +277,276 @@ public void testSystemHAGroupTableMutationsAllowedDuringActiveToStandby() throws
...
+        assertTrue("Expected fail-fast under elevated retry settings: failed commit took "
+          + durationMs + "ms which exceeds the bound of " + MAX_FAIL_FAST_DURATION_MS
+          + "ms. Under hbase.client.retries.number=16 and hbase.client.pause=100, the absence "
+          + "of the DNRIOE inheritance change would surface a multi-second retry tail. "
+          + "If this assertion fails, the inheritance change is not in place or HBase has "
+          + "started retrying despite DoNotRetryIOException — investigate the cause chain.",
+          durationMs < MAX_FAIL_FAST_DURATION_MS);
+      }
+    }
+  }
```
tkhurana: testMutationBlockedFailsFastViaInheritanceAlone and testMutationBlockedFailsFastUnderElevatedRetries look very similar. Either have one test or differentiate them by what they actually exercise.

lokiore: Fair point — dropped `testMutationBlockedFailsFastViaInheritanceAlone`. The test was redundant with `testMutationBlockedFailsFastUnderElevatedRetries`, which already exercises the same elevated-retry inheritance-alone path. Updated the Javadoc on the kept test to make the empirical-proof framing explicit. Amended commit pushed.

tkhurana (approved):

PR #2492, opened 2026-06-02
PHOENIX-7877 More granular metrics for ReplicationLogWriter

tkhurana:
## Summary

Adds finer-grained histograms in `MetricsReplicationLogGroupSource` to give better visibility into where time is spent inside `ReplicationLogGroup`:

- **`fsSyncTime`**  — wall time of the underlying filesystem sync (`fsync`) inside `processPendingSyncs`. 
- **`pendingSyncWaitTime`** — per-pending-sync wait between consumer pickup of a SYNC event and the start of `fsync`. Helps quantify within-batch drain (events the consumer processes between picking up the SYNC marker and reaching `endOfBatch`).
- **`pendingSyncCount`** — number of sync futures coalesced into a single fsync. Indicates how well the disruptor batches sync calls.
- **`batchSize`** — number of events drained per disruptor batch. Counter is reset in a `finally` block on `endOfBatch` so the value never leaks across batches when an exception path bypasses `processPendingSyncs`.
- **`rotationTime`** — was declared but never populated; now instrumented in `ReplicationLog.LogRotationTask.run` and recorded on both success and failure paths.

## Test plan
- [x] `mvn test -pl phoenix-core -Dtest=ReplicationLogGroupTest` passes
- [x] `mvn verify -pl phoenix-core -Dit.test=ReplicationLogGroupIT` passes
- [x] `mvn spotless:check` passes
- [ ] After deployment, verify new histograms appear under `RegionServer,sub=ReplicationLogGroup,haGroup=...` JMX context with non-zero values during sustained workload

### DIFF
--- a/phoenix-core-server/src/main/java/org/apache/phoenix/replication/ReplicationLogGroup.java
+++ b/phoenix-core-server/src/main/java/org/apache/phoenix/replication/ReplicationLogGroup.java
@@ -962,12 +962,27 @@ public class ReplicationLogGroup {
   }
 
   /**
-   * Handles events from the Disruptor,
+   * Holder for a pending SYNC event captured at consumer pickup time. Stored separately from the
+   * LogEvent because Disruptor reuses ring buffer slots — retaining a LogEvent reference past the
+   * end of its onEvent invocation is unsafe.
    */
+  private static class PendingSync {
+    final CompletableFuture<Void> future;
+    final long pickupTimeNs;
+
+    PendingSync(CompletableFuture<Void> future, long pickupTimeNs) {
+      this.future = future;
+      this.pickupTimeNs = pickupTimeNs;
+    }
+  }
+
+  /** Handles events from the Disruptor. */
   protected class LogEventHandler implements EventHandler<LogEvent>, LifecycleAware {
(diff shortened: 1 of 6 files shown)


phoenix-core-server/src/main/java/org/apache/phoenix/replication/ReplicationLogGroup.java
```diff
@@ -1028,16 +1043,27 @@ private void updateModeOnFailure(IOException e) throws IOException {
...
      */
     private void processPendingSyncs(long sequence) throws IOException {
-      if (pendingSyncFutures.isEmpty()) {
+      int pendingSyncCount = pendingSyncs.size();
+      if (pendingSyncCount == 0) {
         return;
       }
+      metrics.updatePendingSyncCount(pendingSyncCount);
+      // Record per-event wait between SYNC pickup and fsync start.
+      long syncStartNs = System.nanoTime();
```
apurtell, line 1052: Move store to `syncStartNs` below the histogram updates to only time the actual sync. Just a nit.

apurtell (approved):

PR #2493, opened 2026-06-03
PHOENIX-7878 CDC perf improvement - skip redundant cell versions on data table scans

virajjasani:
Jira: PHOENIX-7878

CDC perf improvement - skip redundant cell versions on data table scans

When a CDC query runs with pre, post, and/or change scopes, it scans the data table to reconstruct each change event (the change image plus the pre-image, and for the consumer path the full data-row state). Today that data table scan is a raw, all-versions scan, so for every data row we read back every version of every column - even though, for a given batch of changes, we only need two cells per column per change: the cell at the change timestamp, and the most recent cell just below it (the pre-image). On rows that are updated frequently this means we read, transfer, and process far more cells than the event reconstruction actually uses, which adds CPU, memory, and network overhead to CDC reads.

The purpose of this Jira is to add new CDCVersionFilter, in addition to SkipScanFilter on the data table scans. For each row it is given the set of change timestamps from the current batch and keeps only the cells that matter: the cell at each change timestamp, the first cell below each change timestamp (the pre-image), and all DeleteFamily markers (needed for deletion tracking), other cells are skipped to avoid redundant data transfer.

This is performance improvement

UT and IT tests

Claude Opus 4.8

### DIFF
--- a/phoenix-core-server/src/main/java/org/apache/phoenix/coprocessor/CDCGlobalIndexRegionScanner.java
+++ b/phoenix-core-server/src/main/java/org/apache/phoenix/coprocessor/CDCGlobalIndexRegionScanner.java
@@ -28,9 +28,11 @@ import java.util.Base64;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.HashMap;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import java.util.TreeSet;
 import org.apache.hadoop.hbase.Cell;
 import org.apache.hadoop.hbase.CellBuilder;
 import org.apache.hadoop.hbase.CellBuilderFactory;
@@ -40,6 +42,8 @@ import org.apache.hadoop.hbase.client.Put;
 import org.apache.hadoop.hbase.client.Result;
 import org.apache.hadoop.hbase.client.Scan;
 import org.apache.hadoop.hbase.coprocessor.RegionCoprocessorEnvironment;
+import org.apache.hadoop.hbase.filter.Filter;
+import org.apache.hadoop.hbase.filter.FilterList;
 import org.apache.hadoop.hbase.io.ImmutableBytesWritable;
 import org.apache.hadoop.hbase.regionserver.Region;
 import org.apache.hadoop.hbase.regionserver.RegionScanner;
@@ -49,6 +53,7 @@ import org.apache.phoenix.coprocessor.generated.IndexMutationsProtos;
 import org.apache.phoenix.execute.TupleProjector;
 import org.apache.phoenix.expression.Expression;
 import org.apache.phoenix.expression.SingleCellColumnExpression;
+import org.apache.phoenix.filter.CDCVersionFilter;
 import org.apache.phoenix.hbase.index.util.ImmutableBytesPtr;
 import org.apache.phoenix.index.CDCTableInfo;
 import org.apache.phoenix.index.IndexMaintainer;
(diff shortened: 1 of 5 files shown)


virajjasani:
Build: https://ci-hadoop.apache.org/job/Phoenix/job/Phoenix-PreCommit-GitHub-PR/job/PR-2493/

phoenix-core/src/it/java/org/apache/phoenix/end2end/CDCQueryIT.java
```diff
@@ -434,6 +434,86 @@ public void testSelectCDC() throws Exception {
...
+        put("V1", "INTEGER");
+        put("V2", "INTEGER");
+        put("V3", "INTEGER");
+        put("B.VB", "INTEGER");
+      }
+    };
+    String cdcFullName = SchemaUtil.getTableName(schemaName, cdcName);
+    try (Connection conn = newConnection(tenantId)) {
+      // For debug: uncomment to see the exact results logged to console.
+      dumpCDCResults(conn, cdcName, new TreeMap<String, String>() {
```
palashc, line 491: Was this meant to be commented out?

phoenix-core-server/src/main/java/org/apache/phoenix/coprocessor/CDCGlobalIndexRegionScanner.java
```diff
@@ -123,21 +128,60 @@ protected Scan prepareDataTableScan(Collection<byte[]> dataRowKeys) throws IOExc
...
-    // stopTimeRange = PLong.INSTANCE.getCodec().decodeLong(
-    // scan.getStopRow(), 0, SortOrder.getDefault());
-    // }
     Scan dataScan = prepareDataTableScan(dataRowKeys, true);
     if (dataScan == null) {
       return null;
     }
-    return CDCUtil.setupScanForCDC(dataScan);
+    CDCUtil.setupScanForCDC(dataScan);
+    Map<ImmutableBytesPtr, long[]> timestampMap = buildDataRowTimestampMap(dataRowKeys);
```
palashc: Can we avoid building this timestamp map in every task and precompute beforehand? But maybe it is okay since number of tasks will usually be small - based on number of regions involved and number of rowkeys?

palashc (approved):
LGTM +1, couple of nits

phoenix-core-server/src/main/java/org/apache/phoenix/coprocessor/CDCGlobalIndexRegionScanner.java
virajjasani: It is possible but it is far bigger refactor, worth doing as separate PR because this PR is already too big. However, dataRowKeys are already available so generating map would not be that time consuming.

phoenix-core/src/it/java/org/apache/phoenix/end2end/CDCQueryIT.java
virajjasani: Sure, removing

tkhurana (approved):

phoenix-core/src/it/java/org/apache/phoenix/end2end/CDCQueryIT.java
palashc: do we need to dumpCDCResults in CI test runs? I though you would comment that out.

virajjasani: It's just being used in all tests, it's fine either way, there are not many rows to print but I can also remove it

haridsv:
Just capturing the below feedback  I gave offline as the context for the latest change:

Question on the per-row long[] of change timestamps: why enumerate each timestamp rather than send a per-row [min, max] range?

For the common CDC time-range query, a page covers a contiguous timestamp band within a partition, so a row's cells in [min, max] are exactly its changes for that page — a range and the enumerated array select the same cells. The pre-image below min is still covered: the filter already includes the first cell below the lowest change and then NEXT_COLs, so "one cell past min" needs no extra machinery, and the builder tolerates any surplus in-band versions anyway. A range is also cheaper to build (no per-row TreeSet/array) and smaller to serialize.

Is there a case where enumeration prunes meaningfully more than a range would? Want to make sure I'm not missing something the per-timestamp encoding is specifically handling.

I believe range covers splits/merges correctly too. A range never under-fetches: every change is in [min, max], and every pre-image is the first cell below its change, which the scan still reaches (the lowest change's pre-image is just the "one cell below min").

A PARTITION_ID is the (hashed) encoded region name, so when a row's region splits/merges, its later changes land under a different PARTITION_ID that can sort before or after the old one — i.e. a single row's change blocks aren't necessarily in time order in the index. A page straddling two such blocks may therefore see a wider [min, max] and pull in a few extra in-band versions — but those are exactly the ones CDCChangeBuilder already discards, so it's bounded over-fetch, not a correctness gap.

PR #2496, opened 2026-06-05
PHOENIX-7880 Do not treat literal string SET values as arithmetic in BSON UpdateExpression

palashc:

`UpdateExpressionUtils.getNewFieldValue` interpreted any `BsonString` SET value
containing `" + "` or `" - "` as an arithmetic expression (splitting on
whitespace and resolving each token as a numeric field/literal). This change
removes that string-content heuristic: every plain string is now treated as a
literal. Arithmetic continues to be evaluated only from the explicit document
forms (`$ADD` / `$SUBTRACT` / `$IF_NOT_EXISTS`), which is what the
UpdateExpression parser already emits.

The now-unused `stringToNumber` helper and its imports (`Pattern`, `Matcher`,
`NumberFormat`, `ParseException`) were removed. `getNewFieldValue` has a single
caller (`executeSetExpression`).

The heuristic is fundamentally ambiguous: a literal value such as
`"Foo - Bar Baz Service"` is indistinguishable from an arithmetic expression
like `"fieldA - fieldB"`. Any DynamoDB `UpdateItem` (legacy `AttributeUpdates`
PUT or `UpdateExpression` `SET attr = :val`) whose string value contained
`" + "` / `" - "` failed with:
```
java.lang.IllegalArgumentException: Operand ... does not exist at org.apache.phoenix.expression.util.bson.UpdateExpressionUtils.getNewFieldValue(UpdateExpressionUtils.java:630)
```

Genuine arithmetic is unaffected because the parser encodes it as
`$ADD`/`$SUBTRACT` documents, not as raw strings.
Yes. Previously, a `SET` string value containing `" + "` or `" - "` was
(mis)interpreted as arithmetic — typically throwing `Operand ... does not
exist`. Now such values are stored verbatim as literals. This only affects the
unreleased branch behavior of `BSON_UPDATE_EXPRESSION`; explicit
`$ADD`/`$SUBTRACT` arithmetic is unchanged.
- Updated `UpdateExpressionUtilsTest`:
  - `testMixedSetExpressions` and `testUpdateExpression` now assert literal
    storage for the former string-arithmetic cases.
  - Added `testLiteralStringWithArithmeticOperatorsIsStoredVerbatim`.
  - All 25 tests pass.
- End-to-end verified against the DynamoDB REST adapter (phoenix-adapters)
  `UpdateItemIT` with new ITs covering literal strings containing `" - "`/`" + "`
  via both legacy `AttributeUpdates` and `UpdateExpression :val` paths.
Generated-by: Cursor (Claude Opus 4.8)

### DIFF
--- a/phoenix-core-client/src/main/java/org/apache/phoenix/expression/util/bson/UpdateExpressionUtils.java
+++ b/phoenix-core-client/src/main/java/org/apache/phoenix/expression/util/bson/UpdateExpressionUtils.java
@@ -18,14 +18,10 @@
 package org.apache.phoenix.expression.util.bson;
 
 import java.math.BigDecimal;
-import java.text.NumberFormat;
-import java.text.ParseException;
 import java.util.ArrayList;
 import java.util.HashSet;
 import java.util.Map;
 import java.util.Set;
-import java.util.regex.Matcher;
-import java.util.regex.Pattern;
 import org.bson.BsonArray;
 import org.bson.BsonDecimal128;
 import org.bson.BsonDocument;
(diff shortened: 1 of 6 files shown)


virajjasani (approved):
+1, pending build results

virajjasani:
https://ci-hadoop.apache.org/job/Phoenix/job/Phoenix-PreCommit-GitHub-PR/job/PR-2496/

PR #2497, opened 2026-06-05
PHOENIX-7881 Refactor UTs and ITs to assert on ExplainPlanAttributes

apurtell:
Refactors ~70 UTs and ITs to assert on `ExplainPlanAttributes` via a fluent `assertPlan` API instead of hardcoding exact `EXPLAIN` text matches. Expands `ExplainPlanAttributes` with the additional fields the assertion API needs and reorganizes its attribute display order for better logical grouping. 

No changes to the actual `EXPLAIN` text output format. Coverage is preserved.  Some exact text matching is retained in `QueryPlanTest` so that coverage is not lost. Further refactoring toward full `assertPlan` reliance is deferred until the new `EXPLAIN` logic and output-grammar validation land.

Generated-by: Claude Opus 4.8[1m] <noreply@anthropic.com>


apurtell:
Going to switch to a feature branch.

PR #2518, opened 2026-06-11
PHOENIX-7566 HAGroupStore state machine: ANISTS non-blocking, DS->STANDBY_TO_ACTIVE

ritegarg:
HAGroupStore state-machine fixes for consistent failover (PHOENIX-7566).

- In `HAGroupStoreRecord.getClusterRole()`, map `ACTIVE_NOT_IN_SYNC_TO_STANDBY` to the `ACTIVE` role (non-blocking). Only `ACTIVE_IN_SYNC_TO_STANDBY` remains `ACTIVE_TO_STANDBY` (mutation-blocking).
- Add `STANDBY_TO_ACTIVE` to `DEGRADED_STANDBY.allowedTransitions`, so a degraded standby can promote directly on failover (matches the TLA+ model).

On the not-in-sync failover path, an active that is still draining its backlog was treated as mutation-blocked, even though it should keep serving writes until it is in sync and ready to hand off. Separately, a `DEGRADED_STANDBY` had no allowed transition to `STANDBY_TO_ACTIVE`, so it could not promote during failover. Together these prevented a clean failover when the active is behind.

No. Internal HA state-machine semantics only; no API, config, or CLI changes.

- `HAGroupStoreRecordTest` (17/17): role mapping + mutation-block coverage (`ACTIVE_NOT_IN_SYNC_TO_STANDBY` non-blocking, `ACTIVE_IN_SYNC_TO_STANDBY` blocking) and the `DEGRADED_STANDBY → STANDBY` / `STANDBY_TO_ACTIVE` transitions.
- `HAGroupStoreManagerIT` (17/17): per-group mutation blocking.

Yes — co-authored with Cursor.

### DIFF
--- a/phoenix-core-client/src/main/java/org/apache/phoenix/jdbc/HAGroupStoreRecord.java
+++ b/phoenix-core-client/src/main/java/org/apache/phoenix/jdbc/HAGroupStoreRecord.java
@@ -76,11 +76,11 @@ public class HAGroupStoreRecord {
         case ABORT_TO_ACTIVE_NOT_IN_SYNC:
         case ACTIVE_IN_SYNC:
         case ACTIVE_NOT_IN_SYNC:
+        case ACTIVE_NOT_IN_SYNC_TO_STANDBY:
         case ACTIVE_NOT_IN_SYNC_WITH_OFFLINE_PEER:
         case ACTIVE_WITH_OFFLINE_PEER:
           return ClusterRoleRecord.ClusterRole.ACTIVE;
         case ACTIVE_IN_SYNC_TO_STANDBY:
-        case ACTIVE_NOT_IN_SYNC_TO_STANDBY:
           return ClusterRoleRecord.ClusterRole.ACTIVE_TO_STANDBY;
         case ABORT_TO_STANDBY:
         case DEGRADED_STANDBY:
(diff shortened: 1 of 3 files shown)


phoenix-core-client/src/main/java/org/apache/phoenix/jdbc/HAGroupStoreRecord.java
```diff
@@ -114,7 +114,7 @@ public ClusterRoleRecord.ClusterRole getClusterRole() {
       ACTIVE_IN_SYNC_TO_STANDBY.allowedTransitions =
         ImmutableSet.of(ABORT_TO_ACTIVE_IN_SYNC, STANDBY);
       STANDBY_TO_ACTIVE.allowedTransitions = ImmutableSet.of(ABORT_TO_STANDBY, ACTIVE_IN_SYNC);
-      DEGRADED_STANDBY.allowedTransitions = ImmutableSet.of(STANDBY);
+      DEGRADED_STANDBY.allowedTransitions = ImmutableSet.of(STANDBY, STANDBY_TO_ACTIVE);
```
apurtell, line 117: Looks good. Fixes a DS side deadlock at peer AISTS.

phoenix-core-client/src/main/java/org/apache/phoenix/jdbc/HAGroupStoreRecord.java
```diff
@@ -76,11 +76,11 @@ public ClusterRoleRecord.ClusterRole getClusterRole() {
         case ABORT_TO_ACTIVE_NOT_IN_SYNC:
         case ACTIVE_IN_SYNC:
         case ACTIVE_NOT_IN_SYNC:
+        case ACTIVE_NOT_IN_SYNC_TO_STANDBY:
```
apurtell, line 79: Looks good. Reduces unavailability during failover. The safety argument shifts a bit, but the transition tables do not allow the peer to promote, so we still have only a single writer in all states.

apurtell (approved):
Approved.
I did the review myself, but did have the robots help with considering TLA model updates after this goes in. (I'll have some fun there.)

Then I asked for suggestions on improving test coverage because the current coverage seems a bit weak. These were the suggestions, fyi

1. Drive (ANIS, DS) -> (S, AIS) through the real ZK + FailoverManagementListener path in something like FailoverPhoenixConnectionIT or HAGroupStoreManagerIT, and assert both clusters converge to (S, AIS). The current unit test only pins the transition table.

2. Nit in HAGroupStoreRecordTest: Add an assertFalse(DEGRADED_STANDBY.isTransitionAllowed(SOMETHING_NOT_IN_SET)) to lock down the allowed set.

ritegarg:
> Approved. I did the review myself, but did have the robots help with considering TLA model updates after this goes in. (I'll have some fun there.)
> 
> Then I asked for suggestions on improving test coverage because the current coverage seems a bit weak. These were the suggestions, fyi
> 
> 1. Drive (ANIS, DS) -> (S, AIS) through the real ZK + FailoverManagementListener path in something like FailoverPhoenixConnectionIT or HAGroupStoreManagerIT, and assert both clusters converge to (S, AIS). The current unit test only pins the transition table.
> 2. Nit in HAGroupStoreRecordTest: Add an assertFalse(DEGRADED_STANDBY.isTransitionAllowed(SOMETHING_NOT_IN_SET)) to lock down the allowed set.

1. It will not work with this change, @Himanshu-g81 will work on [PHOENIX-7920](https://issues.apache.org/jira/browse/PHOENIX-7920) and we can add a test then.
2. I have added that test.

PR #2519, opened 2026-06-11
PHOENIX-7566 HAGroupStore admin tool: HDFS URLs, URL validation, failover recovery guidance

ritegarg:
HAGroupStore admin tool (`PhoenixHAAdminTool`) fixes for consistent failover (PHOENIX-7566).

* **HDFS URLs in the tool:** `get`/`list` print `HDFS URL` / `Peer HDFS URL`; `update` accepts `--hdfs-url` / `--peer-hdfs-url` (registered, counted in the "at least one field" guard, shown in proposed changes).
* **Canonical `SYSTEM.HA_GROUP` writes:** `create`/`update` write the slot columns (including `HDFS_URL_1/2`) in a canonical order keyed on the formatted ZK URL, so each slot's `ZK`/`CLUSTER`/`ROLE`/`HDFS` stay paired and both clusters persist identical rows (matching the periodic ZK→`SYSTEM.HA_GROUP` sync). Previously `update` wrote local-first and never wrote HDFS, which could leave `ZK_URL_n` unpaired from `HDFS_URL_n`.
* **URL validation on write, tolerance on read:** `create`/`update` validate URL fields against the registry type the read path uses (ZK quorum vs RPC/master). **ZK URLs are always validated** (they identify the HA pair); `--force` stores only malformed *cluster* URLs as-is. `HAGroupStoreClient.getHAGroupNames` skips + WARNs a row whose ZK URL won't parse instead of failing enumeration for all callers. `get-cluster-role-record` renders an unparseable stored cluster URL as `<invalid>` (surfacing the cause) instead of throwing.
* **Failover recovery guidance:** on `initiate-failover` / `abort-failover` timeout, prints manual-recovery steps (inspect both clusters, restore connectivity, abort on the standby, or force a steady state).

The tool could not display or set HDFS URLs, and `update` could persist a row whose ZK/CLUSTER/ROLE/HDFS slots were unpaired. A single malformed stored URL could crash `list` / `get-cluster-role-record` and break HA-group enumeration for server-side callers. A timed-out failover left the operator with no recovery direction.

### Does this PR introduce any user-facing change?
Yes — CLI only: new `--hdfs-url` / `--peer-hdfs-url` and HDFS URLs in `get`/`list`; canonical, paired `SYSTEM.HA_GROUP` rows; malformed cluster URLs rejected on `create`/`update` (overridable with `--force`; ZK URLs always validated); clearer `get-cluster-role-record` and failover-timeout output. No API or wire changes.

* `PhoenixHAAdminToolIT` (18/18) and `PhoenixHAAdminIT` (9/9).

Yes — co-authored with Cursor.

### DIFF
--- a/phoenix-core-client/src/main/java/org/apache/phoenix/jdbc/PhoenixHAAdminTool.java
+++ b/phoenix-core-client/src/main/java/org/apache/phoenix/jdbc/PhoenixHAAdminTool.java
@@ -565,34 +565,6 @@ public class PhoenixHAAdminTool extends Configured implements Tool {
   }
 
   /**
-   * Render a degraded cluster-role-record view from the raw HAGroupStoreRecord when normalization
-   * of a stored cluster URL fails, marking the offending URL {@code <invalid>} so the operator can
-   * repair it (e.g. {@code update -g <group> -c <good-url> -av}, adding {@code --force} if needed).
-   */
-  private void printClusterRoleRecordWithInvalidUrls(HAGroupStoreManager manager,
-    String haGroupName) {
-    System.out.println("\nCluster Role Record:");
-    System.out.println("  HA Group Name:     " + haGroupName);
-    try {
-      HAGroupStoreRecord record = manager.getHAGroupStoreRecord(haGroupName).orElse(null);
-      if (record != null) {
-        System.out.println("  Policy:            " + record.getPolicy());
-        System.out.println("  Cluster 1 URL:     "
-          + describeUrl(record.getClusterUrl(), ClusterRoleRecord.RegistryType.RPC));
-        System.out.println("  Cluster 1 Role:    " + record.getClusterRole());
-        System.out.println("  Cluster 2 URL:     "
-          + describeUrl(record.getPeerClusterUrl(), ClusterRoleRecord.RegistryType.RPC));
-      }
-    } catch (Exception e) {
-      LOG.warn("Could not read raw HA group record for " + haGroupName, e);
-    }
-    System.out.println("  \u26a0 A stored cluster URL is invalid and could not be normalized.");
-    System.out.println(
-      "    Repair with: update -g " + haGroupName + " -c <good-url> (or -pc) -av  [--force]");
-    System.out.println();
-  }
-
-  /**
    * Initiates failover on active cluster, transitioning it to standby while peer becomes active.
    * <p>
    * <b>Architecture:</b> Failover is a coordinated state transition where the currently ACTIVE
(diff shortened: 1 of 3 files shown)


phoenix-core-client/src/main/java/org/apache/phoenix/jdbc/PhoenixHAAdminTool.java
```diff
@@ -807,9 +872,24 @@ private int executeGetClusterRoleRecord(String[] args) throws Exception {
...
+        System.out.println("\nCluster Role Record for '" + haGroupName
+          + "' could not be built (likely a malformed stored URL): " + e.getMessage());
+        if (raw != null) {
+          System.out.println("  Cluster URL:       "
+            + describeUrl(raw.getClusterUrl(), ClusterRoleRecord.RegistryType.RPC));
+          System.out.println("  Peer Cluster URL:  "
+            + describeUrl(raw.getPeerClusterUrl(), ClusterRoleRecord.RegistryType.RPC));
+        }
+        System.out.println(
+          "  Repair with: update -g " + haGroupName + " -c <good-url> (or -pc) -av  [--force]");
```
tkhurana: Should this go to System.err to match other existing error output in the file

phoenix-core-client/src/main/java/org/apache/phoenix/jdbc/PhoenixHAAdminTool.java
```diff
@@ -807,9 +872,24 @@ private int executeGetClusterRoleRecord(String[] args) throws Exception {
...
+      }
 
       return RET_SUCCESS;
```
tkhurana, line 894: The exception code path falls down and returns RET_SUCCESS

tkhurana (changes requested):

phoenix-core-client/src/main/java/org/apache/phoenix/jdbc/PhoenixHAAdminTool.java
ritegarg: Good catch — moved the diagnostic to System.err to match handleCommandError and the rest of the tool's error output. Done in 79e7b1e.

phoenix-core-client/src/main/java/org/apache/phoenix/jdbc/PhoenixHAAdminTool.java
ritegarg: Fixed — the catch now returns RET_UPDATE_ERROR (same code handleCommandError uses) instead of falling through to RET_SUCCESS, so a malformed stored URL surfaces a non-zero exit code. Updated testGetClusterRoleRecordMarksInvalidUrl to assert RET_UPDATE_ERROR and the stderr <invalid> diagnostic. Done in 79e7b1e.

tkhurana (approved):

PR #2540, opened 2026-06-22
PHOENIX-7931 Coalesce per-batch mutations into a single record

tkhurana:
## PHOENIX-7931: Coalesce per-batch replication appends into a single record

Targets the `PHOENIX-7562-feature-new` feature branch (not master).

### What changed
On the write path, a server-side batch's mutations for a given table are now coalesced into a **single** replication log record carrying a flat cell stream, rather than one record per mutation. On the replay path, `MutationCellGrouper` reconstructs Put/Delete mutations from that cell stream on the row + put-vs-delete boundary, mirroring the algorithm HBase's `ReplicationSink` uses for WALEdits.

This cuts ring-buffer pressure substantially (≈5× fewer Disruptor events at high batch sizes) without changing the fsync floor — coalescing already minimized inner syncs. The on-disk record format changes accordingly (per-cell row/family/qualifier/timestamp/type/value, with a per-record table name + commitId + attributes header).

### Key files
- `IndexRegionObserver` — per-(table, batch) cell aggregation before a single `append`
- `MutationCellGrouper` (new) — replay-side cell-stream → mutation regrouping
- `LogFileCodec` / `LogFileRecord` / `LogFile` — new coalesced record format
- `ReplicationLogGroup` / `ReplicationLog` — cell-oriented append API

### Testing
- New `MutationCellGrouperTest` (10 cases): empty input, single Put/Delete, contiguous same-row/type coalescing, row-change splits, Put→Delete same-row type-boundary split, three-row Put/Delete/Put, adjacent DeleteColumn/DeleteFamily subtype split, non-consecutive same-row partitioning, and cell preservation.
- New `testAppendAndSyncSingleBatchRecordCount` IT pins the coalesced record-count contract with cross-cluster cell-level equality.
- New `testSyncMetricsEmitted` unit test — deterministic replacement for the flaky IT metrics assertion (injects an fsync delay to clear ms-truncation; reads metrics only after a graceful close that joins the consumer).
- Expanded `LogFileCodecTest` round-trip coverage for the new format (multi-family ordering, empty-record rejection, all cell types).
- `testReplicationSyncPathSimulator` (opt-in) for sync-coalescing measurement under contention.

### Verification
- `phoenix-core-server` compiles clean; `MutationCellGrouperTest`, `LogFileCodecTest`, and the replication unit tests pass.
- `mvn spotless:check` passes on both `phoenix-core-server` and `phoenix-core`.

### DIFF
--- a/phoenix-core-server/src/main/java/org/apache/phoenix/replication/log/LogFileCodec.java
+++ b/phoenix-core-server/src/main/java/org/apache/phoenix/replication/log/LogFileCodec.java
@@ -26,55 +26,56 @@ import java.io.EOFException;
 import java.io.IOException;
 import java.nio.ByteBuffer;
 import java.nio.charset.StandardCharsets;
+import java.util.ArrayList;
+import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import org.apache.hadoop.hbase.Cell;
 import org.apache.hadoop.hbase.KeyValue;
-import org.apache.hadoop.hbase.client.Delete;
-import org.apache.hadoop.hbase.client.Mutation;
-import org.apache.hadoop.hbase.client.Put;
 import org.apache.hadoop.hbase.io.ByteBuffInputStream;
 import org.apache.hadoop.hbase.nio.ByteBuff;
 import org.apache.hadoop.hbase.util.Bytes;
 import org.apache.hadoop.io.WritableUtils;
 
 /**
- * Default Codec for encoding and decoding ReplicationLog Records within a block buffer. This
- * implementation uses standard Java DataInput/DataOutput for serialization. Record Format within a
- * block:
+ * Default Codec for encoding and decoding ReplicationLog Records within a block buffer. The on-disk
+ * record is cell-oriented (mirroring HBase's WALEdit): a record carries a flat ordered list of
+ * cells across one or more rows. Mutation reconstruction (grouping by row+type) is the consumer's
+ * responsibility and lives in {@link org.apache.phoenix.replication.MutationCellGrouper}.
  *
  * <pre>
  *   +--------------------------------------------+
  *   | RECORD LENGTH (vint)                       |
  *   +--------------------------------------------+
  *   | RECORD HEADER                              |
- *   |  - Mutation type (byte)                    |
  *   |  - HBase table name length (vint)          |
  *   |  - HBase table name (byte[])               |
- *   |  - Transaction/SCN or commit ID (vint)     |
+ *   |  - Commit ID (vlong)                       |
  *   +--------------------------------------------+
- *   | ROW KEY LENGTH (vint)                      |
- *   | ROW KEY (byte[])                           |
+ *   | NUMBER OF ATTRIBUTES (vint)                |
  *   +--------------------------------------------+
- *   | MUTATION TIMESTAMP (vint)                  |
- *   +--------------------------------------------+
- *   | NUMBER OF COLUMN FAMILIES CHANGED (vint)   |
- *   +--------------------------------------------+
- *   | PER-FAMILY DATA (repeated)                 |
+ *   | PER-ATTRIBUTE (repeated)                   |
  *   |   +--------------------------------------+ |
- *   |   | COLUMN FAMILY NAME LENGTH (vint)     | |
- *   |   | COLUMN FAMILY NAME (byte[])          | |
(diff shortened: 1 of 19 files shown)


phoenix-core-server/src/main/java/org/apache/phoenix/replication/log/LogFileCodec.java
```diff
@@ -26,55 +26,56 @@
...
 import org.apache.hadoop.hbase.io.ByteBuffInputStream;
 import org.apache.hadoop.hbase.nio.ByteBuff;
 import org.apache.hadoop.hbase.util.Bytes;
 import org.apache.hadoop.io.WritableUtils;
 
 /**
- * Default Codec for encoding and decoding ReplicationLog Records within a block buffer. This
- * implementation uses standard Java DataInput/DataOutput for serialization. Record Format within a
- * block:
+ * Default Codec for encoding and decoding ReplicationLog Records within a block buffer. The on-disk
```
apurtell, line 41: This is the most interesting change (to me). I suppose it's of a piece with the evolution of HBase WAL with WALEdit to group cells and on the RPC side "cell blocks" instead of individually described cells in the IDL. 

It was probably inevitable that this grouping was necessary to bring down overheads but I wanted to go with simple and atomic first if we could get away with it. 

We may want to restructure the ascii art here further

```text
RECORD LENGTH (vint)
RECORD HEADER
  Table name length (vint)
  Table name bytes
  Commit id (vlong)
ATTRIBUTES COUNT (vint)            -- NEW: record-level attribute map
PER-ATTRIBUTE (repeated):
  Key length (vint)
  Key bytes (UTF-8)
  Value length (vint)
  Value bytes
CELL COUNT (vint)
PER-CELL (repeated):
  Row length (vint)                -- row written on EVERY cell
  Row bytes
  Family length (vint)             -- family written on EVERY cell
  Family bytes
  Qualifier length (vint)
  Qualifier bytes
  Cell timestamp (long, 8 bytes)
  Cell type byte (1 byte)
  Value length (vint)
  Value bytes
```

phoenix-core-server/src/main/java/org/apache/phoenix/replication/reader/ReplicationLogProcessor.java
```diff
@@ -246,21 +246,19 @@ public void processLogFile(FileSystem fs, Path filePath) throws IOException {
...
-          totalProcessed += currentBatchSize;
-          tableToMutationsMap.clear();
-          currentBatchSize = 0;
-          currentBatchSizeBytes = 0;
+        for (Mutation mutation : record.getMutations()) {
+          tableToMutationsMap.computeIfAbsent(tableName, k -> new ArrayList<>()).add(mutation);
+          currentBatchSize++;
+          currentBatchSizeBytes += mutation.heapSize();
+
+          // Process when we reach either the batch count or size limit
```
apurtell, line 257: Before this change a record was always a single Mutation, so batches always split between records. Now a single record's mutations can be split across two `processReplicationLogBatch` invocations. I think it's fine but future maintainers cannot assume a certain atomicity here.

phoenix-core/src/test/java/org/apache/phoenix/replication/ReplicationLogGroupTest.java
```diff
@@ -1429,59 +1456,15 @@ public void testRetryPicksUpStagedWriter() throws Exception {
...
   }
 
-  /**
-   * Tests the idle-then-lease-recovery scenario: after a sync clears currentBatch, the system goes
-   * idle. A rotation tick stages a pending writer. The reader performs HDFS lease recovery,
-   * breaking the old writer's stream. When events resume, apply() drains the healthy staged writer
-   * before the action — the broken writer is never touched. No replay needed (empty batch).
-   */
-  @Test
-  public void testIdleLeaseRecoveryDrainsStagedWriter() throws Exception {
```
apurtell, line 1446: I think this test is a real concurrency concern. After a sync clears `currentBatch`, the system goes idle, a rotation tick stages a pending writer, invalidating the old writer. When events resume we need to ensure `apply()` drains to the new and healthy writer before the action and should ensure the dead writer is never touched.

The body of this test only needs some simple updates (any(Mutation.class) → any(List.class), eq(put) → eq(LogFileTestUtil.cellsOf(put))) Let's restore and migrate it.

apurtell (approved):

PR #2546, opened 2026-06-25
PHOENIX-7937 Trim whitespace in EXPLAIN FORMAT JSON scan attributes

apurtell:
Trim scan attribute string values before populating `ExplainPlanAttributes`. Text plan is unchanged. The JSON field values are canonicalized.


apurtell:
This one is annoying because it had a much larger silly impact on tests than expected, and sometimes you just have to grab the IDE and do it by hand. I think I have a handle on it now.

PR #2547, opened 2026-06-25
PHOENIX-7562 HAGroupStore peer cache: fail-closed replay on peer loss

ritegarg:

Make a `STANDBY` RegionServer's replication replay **fail closed** while its peer cluster is not visible, and move all peer-connection handling out of `HAGroupStoreClient` into a dedicated `PeerClusterWatcher`.

* **`PeerClusterWatcher` (new)** — owns the peer `PathChildrenCache` + `PhoenixHAAdmin` for one HA group and reports peer state via a `PeerStateListener` (`onPeerStateChanged` / `onPeerVisible` / `onPeerBlind`):
   * _Retry when peer ZK is down_ (e.g. unreachable at startup): if the cache can't be built the watcher goes `BLIND`, and a per-client daemon — scheduled lazily only once a peer is configured, so an unused watcher starts no thread — retries the build on `phoenix.ha.group.store.peer.cache.retry.interval.seconds` until the peer returns, then goes `VISIBLE` with no restart needed.
   * _Forced redelivery after reconnect_: peer records are de-duplicated by znode version, with exactly one forced redelivery on reconnect so no transition is missed across the disconnect.
   * _Concurrency_: lock order `transitionLock → stateLock`; the blocking cache build and listener callbacks run outside `stateLock`, and visible/blind transitions are delivered atomically with their notifications.
* **`HAGroupStoreCacheUtil` (new)** — shared helpers to parse a znode into `(record, stat)` and build/start a `PathChildrenCache` (init latch released in `finally`).
* **`HAGroupStoreClient`** — delegates peer handling to `PeerClusterWatcher` via a `PeerStateListener`; adds the in-memory effective-state overlay `getEffectiveHAGroupStoreRecord()` (reports a local `STANDBY` as `DEGRADED_STANDBY` whenever the peer is blind — whether the peer drops while already `STANDBY` or the role reaches `STANDBY` after the peer is blind — never persisted), suppresses a redundant real `STANDBY` while degraded, and serializes the local degrade/recover notifications so they cannot reorder.
* **`HAGroupStoreManager` / `HAGroupStoreRecord`** — add `getEffectiveHAGroupStoreRecord(haGroupName)` and `withHAGroupState(...)` (immutable copy for the overlay), and document `DEGRADED_STANDBY`'s dual nature. The `DEGRADED_STANDBY` state and the subscription framework are pre-existing.
* **Replication** — the replay consumers (`ReplicationLogGroup.init`, `ReplicationLogDiscoveryReplay.getHAGroupRecord`, `ReplicationLogReplay.init`) read the **effective** record; mode mapping is otherwise unchanged.
* **Config** — new `phoenix.ha.group.store.peer.cache.retry.interval.seconds` (default 60s; `0` disables retry), with jittered retry and rate-limited `WARN` (1st + every 10th attempt).

When the peer ZK is unreachable — including down at startup — a `STANDBY` can't reliably determine peer state, so replay risked proceeding as if in sync (fail-open). It must instead **fail closed** (`STORE_AND_FORWARD`) until the peer is reachable, then recover automatically. Extracting the peer lifecycle into `PeerClusterWatcher` decouples replay from peer connectivity and keeps `HAGroupStoreClient` focused on the effective HA view.

Yes, within the unreleased consistent-failover feature branch (no change vs released Phoenix):

* New config `phoenix.ha.group.store.peer.cache.retry.interval.seconds` (default 60s).
* A `STANDBY` whose peer ZK is unreachable presents an effective `DEGRADED_STANDBY` (in-memory only, never written to ZK), so replay fails closed until the peer is visible again. Persisted wire format is unchanged.

New and existing unit/integration tests: `HAGroupStoreClientIT`, `HAGroupStoreManagerIT`, `HAGroupStateSubscriptionIT`, `ReplicationLogDiscoveryReplayTestIT`, `ReplicationLogGroupIT`, `ReplicationLogGroupTest`, `HAGroupStoreRecordTest`, `PeerClusterWatcherTest`.

Key cases: peer-loss degrade/recover; the role entering `STANDBY` while the peer is already blind, and cold start with peer ZK down, both present `DEGRADED_STANDBY` while the persisted record stays `STANDBY`; peer ZK down at startup then retry rebuilds the cache; lazy retry scheduling (a never-configured watcher starts no retry thread); in-memory-only overlay; `STANDBY` re-entry suppressed while peer-blind; forced reconnect redelivery; visible/blind serialization; `close()` idempotency; replayer `degrade → abort → recover`; and a listener throwing on the cache `INITIALIZED` event still initializes healthy.

Generated-by: Cursor

### DIFF
--- a/phoenix-core-client/src/main/java/org/apache/phoenix/jdbc/HAGroupStoreClient.java
+++ b/phoenix-core-client/src/main/java/org/apache/phoenix/jdbc/HAGroupStoreClient.java
@@ -59,7 +59,6 @@ import java.util.Objects;
 import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.CopyOnWriteArraySet;
-import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.Executors;
 import java.util.concurrent.RejectedExecutionException;
 import java.util.concurrent.ScheduledExecutorService;
@@ -110,7 +109,6 @@ public class HAGroupStoreClient implements Closeable {
   // Exclusive upper bound for initial-delay jitter on the periodic reconciler (0..30s).
   private static final long LEGACY_CRR_SYNC_JOB_MAX_JITTER_SECONDS = 31;
   private PhoenixHAAdmin phoenixHaAdmin;
-  private PhoenixHAAdmin peerPhoenixHaAdmin;
   // Admin + NodeCache on /phoenix/ha; null when feature disabled.
   private volatile PhoenixHAAdmin legacyHaAdmin;
   private volatile NodeCache legacyCrrNodeCache;
@@ -122,23 +120,36 @@ public class HAGroupStoreClient implements Closeable {
   private final String haGroupName;
   // PathChildrenCache for current cluster and HAGroupName
   private PathChildrenCache pathChildrenCache = null;
-  // PathChildrenCache for peer cluster and HAGroupName
-  private PathChildrenCache peerPathChildrenCache = null;
+  // Watches the peer cluster's HA record; owns the peer cache/admin, retry, and visibility.
+  private final PeerClusterWatcher peerWatcher;
+  // True while this RegionServer presents a local-only DEGRADED_STANDBY. The shared HA record is
+  // not changed; this only affects the effective local view and de-dupes the synthetic
+  // degrade/recover notifications.
+  private volatile boolean localDegradedStandbyActive = false;
   // Whether the client is healthy
   private volatile boolean isHealthy = false;
   // Configuration
   private final Configuration conf;
   // ZK URL for the current cluster and HAGroupName
   private String zkUrl;
-  // Peer Custom Event Listener
-  private final PathChildrenCacheListener peerCustomPathChildrenCacheListener;
   // Wait time for sync mode
   private final long waitTimeForSyncModeInMs;
   // Rotation time for a log
   private final long rotationTimeMs;
   // State tracking for transition detection
   private volatile HAGroupState lastKnownLocalState;
+  // Last peer state delivered to subscribers; used only to populate the "from" state in PEER
+  // notifications. Written from the peer watcher's single delivery thread.
   private volatile HAGroupState lastKnownPeerState;
+  // Last peer ZK URL applied to the peer watcher; used to skip a reconcile when it is unchanged.
+  private volatile String lastConfiguredPeerZKUrl;
+  // Guards the local-only DEGRADED_STANDBY flag; short critical sections only, and the lock
+  // getEffectiveHAGroupStoreRecord() uses. Notifications run outside this lock.
+  private final Object localDegradedStandbyLock = new Object();
+  // Serializes a LOCAL transition with its notification (present/clear/handleLocalStateChange),
+  // held across the notify so degrade/recover/state notifications cannot reorder. Unlike the lock
+  // above, this is not taken by getEffectiveHAGroupStoreRecord().
+  private final Object localTransitionLock = new Object();
 
   // Subscription storage for HA group state change notifications per client instance
   // Map key format: "clusterType:targetState" -> Set<Listeners>
@@ -175,11 +186,13 @@ public class HAGroupStoreClient implements Closeable {
     HAGroupStoreClient result =
       instances.getOrDefault(localZkUrl, new ConcurrentHashMap<>()).getOrDefault(haGroupName, null);
     if (result == null || !result.isHealthy) {
+      HAGroupStoreClient replaced = null;
       synchronized (HAGroupStoreClient.class) {
         result = instances.getOrDefault(localZkUrl, new ConcurrentHashMap<>())
           .getOrDefault(haGroupName, null);
         if (result == null || !result.isHealthy) {
-          result = new HAGroupStoreClient(conf, null, null, haGroupName, zkUrl);
+          HAGroupStoreClient stale = result;
+          result = new HAGroupStoreClient(conf, null, haGroupName, zkUrl);
           if (!result.isHealthy) {
             result.close();
             result = null;
@@ -193,8 +206,15 @@ public class HAGroupStoreClient implements Closeable {
               v.put(haGroupName, created);
               return v;
             });
+            replaced = stale;
           }
         }
+      }
+      // Reclaim the replaced unhealthy client's threads and caches. Outside the monitor because
+      // close() can block; safe because deregisterFromInstances() is value-based and will not
+      // remove the replacement we just registered.
+      if (replaced != null) {
+        replaced.close();
       }
     }
     return result;
@@ -244,10 +264,17 @@ public class HAGroupStoreClient implements Closeable {
     return result;
   }
 
+  /**
+   * Visible-for-testing constructor. A non-null {@code pathChildrenCacheListener} replaces the
+   * default LOCAL cache listener only; it does not affect peer handling. There is no peer-listener
+   * injection point anymore - peer cache events are owned by {@link PeerClusterWatcher}. Tests that
+   * need peer-event visibility should subscribe with
+   * {@code subscribeToTargetState(state, ClusterType.PEER, listener)} or use the package-private
+   * {@code peerWatcher} field directly.
+   */
   @VisibleForTesting
   HAGroupStoreClient(final Configuration conf,
-    final PathChildrenCacheListener pathChildrenCacheListener,
-    final PathChildrenCacheListener peerPathChildrenCacheListener, final String haGroupName,
+    final PathChildrenCacheListener pathChildrenCacheListener, final String haGroupName,
     final String zkUrl) {
     this.conf = conf;
     this.haGroupName = haGroupName;
@@ -258,21 +285,23 @@ public class HAGroupStoreClient implements Closeable {
       QueryServicesOptions.DEFAULT_REPLICATION_LOG_ROTATION_TIME_MS);
     this.legacyCrrSyncEnabled = conf.getBoolean(PHOENIX_HA_LEGACY_CRR_SYNC_ENABLED,
       DEFAULT_PHOENIX_HA_LEGACY_CRR_SYNC_ENABLED);
-    // Custom Event Listener
-    this.peerCustomPathChildrenCacheListener = peerPathChildrenCacheListener;
+    this.peerWatcher = new PeerClusterWatcher(conf, haGroupName,
+      ZK_CONSISTENT_HA_GROUP_RECORD_NAMESPACE, new PeerListener());
     try {
-      // Initialize Phoenix HA Admin
       this.phoenixHaAdmin =
         new PhoenixHAAdmin(this.zkUrl, conf, ZK_CONSISTENT_HA_GROUP_RECORD_NAMESPACE);
-      // Initialize local cache
+      long initTimeoutMs = conf.getLong(PHOENIX_HA_GROUP_STORE_CLIENT_INITIALIZATION_TIMEOUT_MS,
+        DEFAULT_HA_GROUP_STORE_CLIENT_INITIALIZATION_TIMEOUT_MS);
+      PathChildrenCacheListener localListener =
+        pathChildrenCacheListener != null ? pathChildrenCacheListener : localCacheListener();
       this.pathChildrenCache =
-        initializePathChildrenCache(phoenixHaAdmin, pathChildrenCacheListener, ClusterType.LOCAL);
-      // Initialize ZNode if not present in ZK
+        HAGroupStoreCacheUtil.startCache(phoenixHaAdmin.getCurator(), localListener, initTimeoutMs);
       initializeZNodeIfNeeded();
       if (this.pathChildrenCache != null) {
         this.isHealthy = true;
-        // Initialize peer cache
-        maybeInitializePeerPathChildrenCache();
+        HAGroupStoreRecord local = getHAGroupStoreRecord();
+        this.lastConfiguredPeerZKUrl = local != null ? local.getPeerZKUrl() : null;
+        peerWatcher.reconfigure(this.lastConfiguredPeerZKUrl);
       } else {
         LOGGER.error("PathChildrenCache is not initialized, HAGroupStoreClient for " + haGroupName
           + " is unhealthy");
@@ -311,14 +340,7 @@ public class HAGroupStoreClient implements Closeable {
     if (pathChildrenCache != null) {
       pathChildrenCache.rebuild();
     }
-    if (peerPathChildrenCache != null) {
-      try {
-        peerPathChildrenCache.rebuild();
-      } catch (Exception e) {
-        LOGGER.error("Unexpected error occurred while rebuilding peerPathChildrenCache, continuing",
-          e);
-      }
-    }
+    peerWatcher.rebuild();
     LOGGER.info("Rebuild Complete for HAGroupStoreClient for HA group {}", haGroupName);
   }
 
@@ -331,8 +353,7 @@ public class HAGroupStoreClient implements Closeable {
     if (!isHealthy) {
       throw new IOException("HAGroupStoreClient is not healthy");
     }
-    return fetchCacheRecordAndPopulateZKIfNeeded(this.pathChildrenCache, ClusterType.LOCAL)
-      .getLeft();
+    return fetchLocalRecordAndPopulateZKIfNeeded().getLeft();
   }
 
   /**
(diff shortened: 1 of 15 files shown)


phoenix-core-client/src/main/java/org/apache/phoenix/jdbc/PeerClusterWatcher.java
```diff
@@ -0,0 +1,428 @@
...
+    }
+    closeConnection();
+  }
+
+  private void retryIfBlind() {
+    boolean needsBuild;
+    String desiredUrl;
+    long attempt = 0L;
+    synchronized (stateLock) {
+      needsBuild = !watcherClosed && StringUtils.isNotBlank(peerZkUrl) && cache == null;
```
apurtell, line 241: Recovery depends on the `cache` field being null, which could be problematic.

`CuratorFramework` is configured with `HighAvailabilityGroup.RETRY_POLICY`, an exponential backoff retry, 5 retries, 10s max sleep, roughly a minute. If a peer ZK outage exceeds that time the watcher could stay blind indefinitely with cache != null and thus no path to healing.

I think generally Curator will manage to reconnect and the watcher will recover, but there could be prolonged disruptions. We may want to consider a watchdog that, when blind for longer than some threshold, responds by tearing down the peer admin/cache and building it again.

phoenix-core-client/src/main/java/org/apache/phoenix/jdbc/HAGroupStoreClient.java
```diff
@@ -122,23 +120,36 @@ public class HAGroupStoreClient implements Closeable {
...
   // State tracking for transition detection
   private volatile HAGroupState lastKnownLocalState;
+  // Last peer state delivered to subscribers; used only to populate the "from" state in PEER
+  // notifications. Written from the peer watcher's single delivery thread.
   private volatile HAGroupState lastKnownPeerState;
+  // Last peer ZK URL applied to the peer watcher; used to skip a reconcile when it is unchanged.
+  private volatile String lastConfiguredPeerZKUrl;
+  // Guards the local-only DEGRADED_STANDBY flag; short critical sections only, and the lock
+  // getEffectiveHAGroupStoreRecord() uses. Notifications run outside this lock.
+  private final Object localDegradedStandbyLock = new Object();
```
apurtell, line 148: localDegradedStandbyLock... A stuck peer rebuild can stall peer state delivery on the local side. 

`getEffectiveHAGroupStoreRecord()` takes localDegradedStandbyLock and then  `fetchLocalRecordAndPopulateZKIfNeeded` -> `rebuild()` -> `peerWatcher.rebuild()`, which does a blocking `PathChildrenCache.rebuild()` against the peer ZK. 

Concurrently meanwhile a peer event delivering `setBlind`/`setVisible` could call `presentLocalDegradedStandbyIfStandby` or `clearLocalDegradedStandbyIfStillStandby`, both of which need to take `localDegradedStandbyLock`.

Consider these mitigations

In `getEffectiveHAGroupStoreRecord`, snapshot the local record into a local first, then take `localDegradedStandbyLock` only for making the decision.

OR

In `fetchLocalRecordAndPopulateZKIfNeeded`, skip `peerWatcher.rebuild()` , because the local cache is the only thing that needs to be rebuilt.

phoenix-core-client/src/main/java/org/apache/phoenix/jdbc/PeerClusterWatcher.java
```diff
@@ -0,0 +1,428 @@
...
+    }
+    try {
+      ex.execute(() -> reconfigure(url));
+    } catch (RejectedExecutionException e) {
+      LOGGER.debug("Peer reconfigure skipped for HA group {}: watcher closing", haGroupName);
+    }
+  }
+
+  /** Current peer record, or null when the peer is not visible. */
+  HAGroupStoreRecord getCurrentPeerRecord() {
```
apurtell, line 174: Returning a stale value can have downstream effects. In HAGroupStoreClient#getHAGroupStoreRecordFromPeer() → getClusterRoleRecord(), the state state is visible to external clients building/refreshing routing decisions. In syncZKToSystemTable() we can persist a stale peer role into SYSTEM.HA_GROUP. In syncLegacyCRRIfRoleChanged() we can publish a stale combined CRR . In setHAGroupStatusIfNeeded() we might update post transition state with the stale peer role.

This introduces fail open cases.

I think you just need to add

```java
  if (isBlind()) {
    return null;
  }
```

at the top of the method.

phoenix-core-client/src/main/java/org/apache/phoenix/jdbc/HAGroupStoreClient.java
```diff
@@ -866,214 +885,203 @@ private void syncZKToSystemTable() {
...
+        if (!localDegradedStandbyActive) {
+          return;
+        }
+        local = readLocalRecordQuietly();
+        localDegradedStandbyActive = false;
+        recover = local != null && local.getHAGroupState() == HAGroupState.STANDBY;
+      }
+      if (recover) {
+        LOGGER.warn("Peer visible again for HA group {}; clearing local DEGRADED_STANDBY "
+          + "(reason=peer-blind)", haGroupName);
```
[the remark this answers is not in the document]
apurtell: This seems like a good suggestion.

phoenix-core-client/src/main/java/org/apache/phoenix/jdbc/HAGroupStoreClient.java
```diff
@@ -866,214 +885,203 @@ private void syncZKToSystemTable() {
...
+          return false;
+        }
+        localDegradedStandbyActive = true;
       }
-      if (peerPhoenixHaAdmin != null) {
-        peerPhoenixHaAdmin.close();
-        peerPhoenixHaAdmin = null;
+      LOGGER.warn("Peer not visible for HA group {}; presenting local DEGRADED_STANDBY "
+        + "(reason=peer-blind)", haGroupName);
+      notifySubscribers(HAGroupState.STANDBY, HAGroupState.DEGRADED_STANDBY,
```
apurtell: Consider passing the prior effective state explicitly.

phoenix-core-client/src/main/java/org/apache/phoenix/jdbc/HAGroupStoreClient.java
```diff
@@ -866,214 +885,203 @@ private void syncZKToSystemTable() {
...
+          return;
+        }
+        local = readLocalRecordQuietly();
+        localDegradedStandbyActive = false;
+        recover = local != null && local.getHAGroupState() == HAGroupState.STANDBY;
+      }
+      if (recover) {
+        LOGGER.warn("Peer visible again for HA group {}; clearing local DEGRADED_STANDBY "
+          + "(reason=peer-blind)", haGroupName);
+        notifySubscribers(HAGroupState.DEGRADED_STANDBY, HAGroupState.STANDBY,
```
apurtell, line 1094: Same concern as above. This hard codes a from state of DEGRADED_STANDBY regardless of what the listener actually observed.

phoenix-core-client/src/main/java/org/apache/phoenix/jdbc/PeerClusterWatcher.java
```diff
@@ -0,0 +1,428 @@
...
+ * Watches one peer cluster's {@link HAGroupStoreRecord} over the peer's ZooKeeper, for a single HA
+ * group on one RegionServer. Owns the peer cache + admin and a background retry, and reports peer
+ * state changes (de-duplicated by znode version, with one forced redelivery after a reconnect) and
+ * visible&lt;-&gt;blind transitions. Thread-safe; the (possibly blocking) cache build runs on the
+ * caller's thread for the initial {@link #reconfigure} and on the retry executor afterwards, never
+ * holding {@link #stateLock}. Listener callbacks fire outside the lock.
+ */
+final class PeerClusterWatcher implements Closeable {
+
+  /**
```
apurtell, line 54: This javadoc is subtly wrong. onPeerVisible and onPeerBlind are invoked under lock, but onPeerStateChanged is invoked from deliver(), which is not holding transitionLock, only stateLock briefly, and the listener fires outside it.

phoenix-core-client/src/main/java/org/apache/phoenix/jdbc/HAGroupStoreClient.java
```diff
@@ -866,214 +885,203 @@ private void syncZKToSystemTable() {
...
+   * {@link #getHAGroupStoreRecord()} except that, while this RegionServer cannot see the peer, a
+   * local STANDBY is reported as DEGRADED_STANDBY. The DEGRADED_STANDBY overlay is in-memory only
+   * and is never written to the shared HA record; peer connectivity is never exposed directly. The
+   * overlay is applied only on top of a successfully read local record, so this requires the LOCAL
+   * client to be healthy: a lost LOCAL connection is bounded (Curator reconnects, or the
+   * RegionServer is aborted on prolonged ZK loss), so no stale-record fallback is provided here.
+   * @throws IOException if the LOCAL client is not healthy (same contract as
+   *                     {@link #getHAGroupStoreRecord()})
+   */
+  public HAGroupStoreRecord getEffectiveHAGroupStoreRecord() throws IOException {
```
apurtell, line 984: getHAGroupStoreRecord() can call rebuild(), which calls getHAGroupStoreRecordInZooKeeper, then a query from SYSTEM.HA_GROUP, then createHAGroupStoreRecordInZooKeeper, all while synchronized on localDegradedStandbyLock. The peer watcher's onPeerBlind and onPeerVisible listeners also need to acquire localDegradedStandbyLock  so a slow rebuild will pile up peer visibility callbacks behind it. 

Consider snapshotting under the lock and rebuilding outside it.

phoenix-core-client/src/main/java/org/apache/phoenix/jdbc/PeerClusterWatcher.java
```diff
@@ -0,0 +1,428 @@
...
+  }
+
+  /** Current peer record, or null when the peer is not visible. */
+  HAGroupStoreRecord getCurrentPeerRecord() {
+    // Read the cache under the lock: close() nulls the field and closes the cache only after
+    // acquiring this lock, so the O(1) in-memory read here always sees a live cache.
+    synchronized (stateLock) {
+      return HAGroupStoreCacheUtil.recordAndStatAt(cache, toPath(haGroupName)).getLeft();
+    }
+  }
```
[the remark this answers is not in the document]
ritegarg: Good catch — fixed in 0ed09baa. getCurrentPeerRecord() now returns null while the
watcher is blind. The Curator PathChildrenCache keeps serving its last-known data
across a CONNECTION_SUSPENDED/LOST (it isn't torn down on a blind transition), so
returning it would leak a stale peer role into ClusterRoleRecord derivation and the
SYSTEM.HA_GROUP / legacy CRR sync. Failing closed here is consistent with the
effective-state overlay. Added HAGroupStoreClientIT#testPeerRecordIsNullWhileBlind
(peer visible → record exposed; peer ZK down → null while blind; peer back → exposed
again).

phoenix-core-client/src/main/java/org/apache/phoenix/jdbc/HAGroupStoreClient.java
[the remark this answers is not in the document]
ritegarg: Fixed in 0ed09baa — the recover path now logs reason=peer-visible (was mislabeled),
matching reason=peer-blind on the degrade path.

phoenix-core-client/src/main/java/org/apache/phoenix/jdbc/PeerClusterWatcher.java
ritegarg: I dug into this and don't think a watchdog is needed for an already-built cache. The retry path only guards the cache == null case (peer never built at startup). Once the cache exists, a blind state self-heals: the shared Curator client is long-lived and is not torn down on a blind transition (only setBlind() runs), so its background reconnect keeps trying indefinitely — independent of the bounded ExponentialBackoffRetry, which bounds per-operation RetryLoops, not the client's reconnect. When peer ZK returns, CONNECTION_RECONNECTED fires and the watcher goes visible again.

Added HAGroupStoreClientIT#testProlongedPeerOutageBeyondRetryBudgetRecovers, which keeps peer ZK down well past the derived retry budget (PHOENIX_HA_ZK_RETRY_MAX_DEFAULT * PHOENIX_HA_ZK_RETRY_MAX_SLEEP_MS_DEFAULT) and confirms recovery with no watchdog. Happy to add one if you still think the cache == null startup
window warrants it.

phoenix-core-client/src/main/java/org/apache/phoenix/jdbc/HAGroupStoreClient.java
ritegarg: Fixed in 0ed09baa. getEffectiveHAGroupStoreRecord() now fetches the local record (which may rebuild: SYSTEM.HA_GROUP query + local/peer ZK) outside localDegradedStandbyLock, and takes the lock only for the I/O-free overlay decision. A slow rebuild can no longer stall the peer-visibility callbacks (present/clear) that need the same lock.

phoenix-core-client/src/main/java/org/apache/phoenix/jdbc/PeerClusterWatcher.java
ritegarg: Good catch — fixed in https://github.com/apache/phoenix/commit/0ed09baae2a4c9a6c1aa127c819e5ab4c87abc31. getCurrentPeerRecord() now returns null while the watcher is blind. The Curator PathChildrenCache keeps serving its last-known data across a CONNECTION_SUSPENDED/LOST (it isn't torn down on a blind transition), so returning it would leak a stale peer role into ClusterRoleRecord derivation and the SYSTEM.HA_GROUP / legacy CRR sync. Failing closed here is consistent with the effective-state overlay. Added HAGroupStoreClientIT#testPeerRecordIsNullWhileBlind (peer visible → record exposed; peer ZK down → null while blind; peer back → exposed again).

phoenix-core-client/src/main/java/org/apache/phoenix/jdbc/HAGroupStoreClient.java
ritegarg: Fixed in 0ed09baa. Added a single effectiveLocalState(rawState, degradedActive) helper (STANDBY + overlay → DEGRADED_STANDBY, else unchanged) used by both the read path and the LOCAL transition notifications, so fromState reflects what subscribers last saw instead of a hardcoded value — a transition while degraded now reports DEGRADED_STANDBY → STANDBY_TO_ACTIVE. Covered by HAGroupStoreClientIT#testTransitionWhileDegradedReportsDegradedStandbyAsFromState.

phoenix-core-client/src/main/java/org/apache/phoenix/jdbc/HAGroupStoreClient.java
ritegarg: Fixed now

phoenix-core-client/src/main/java/org/apache/phoenix/jdbc/PeerClusterWatcher.java
ritegarg: Fixed the javadoc in 0ed09baa: onPeerVisible/onPeerBlind fire while the watcher holds transitionLock (outside stateLock), whereas onPeerStateChanged fires from deliver() without transitionLock (only a brief stateLock for the version de-dup, and the callback runs outside it). Either way, implementations must not re-enter the watcher and should offload blocking work.

phoenix-core-client/src/main/java/org/apache/phoenix/jdbc/HAGroupStoreClient.java
ritegarg: Fixed in https://github.com/apache/phoenix/commit/0ed09baae2a4c9a6c1aa127c819e5ab4c87abc31. getEffectiveHAGroupStoreRecord() now fetches the local record (which may rebuild: SYSTEM.HA_GROUP query + local/peer ZK) outside localDegradedStandbyLock, and takes the lock only for the I/O-free overlay decision. A slow rebuild can no longer stall the peer-visibility callbacks (present/clear) that need the same lock.

apurtell (approved):
Thanks, this lgtm.

I had the robot look at it also and it gave this reasonable suggestion, but I think it is minor

Tighten the fail-closed check in `getCurrentPeerRecord`: There is a small TOCTOU: `isBlind()` reads the volatile visibility outside `stateLock`. A concurrent `setBlind()` from the Curator event thread could flip visibility between the check and entering `synchronized (stateLock)`, at which point we return the (now-stale) cache.

ritegarg:
> Thanks, this lgtm.
> 
> I had the robot look at it also and it gave this reasonable suggestion, but I think it is minor
> 
> Tighten the fail-closed check in `getCurrentPeerRecord`: There is a small TOCTOU: `isBlind()` reads the volatile visibility outside `stateLock`. A concurrent `setBlind()` from the Curator event thread could flip visibility between the check and entering `synchronized (stateLock)`, at which point we return the (now-stale) cache.

Thanks for the review and the approval! Good call on the robot's suggestion — tightened
it in 3771d223: the isBlind() guard now sits inside synchronized (stateLock), and since
setBlind() flips visibility under that same lock, a concurrent blind transition can no
longer slip between the check and the cache read. No other behavior change;
HAGroupStoreClientIT (53) still green.

PR #2475, opened 2026-05-14
PHOENIX-7855 Fix for flapper root cause: java.lang.IllegalArgumentException: bound must be positive

haridsv:
[PHOENIX-7855](https://issues.apache.org/jira/browse/PHOENIX-7855)

Tested by applying this along with #2450 and #2451 on my local workspace and ran for up to 50 iterations in a loop and didn't encounter any failure.

Generated-by: Claude Code 2.1.128 with Model: us.anthropic.claude-opus-4-6-v1[1m]

PR #2557, opened 2026-06-29
PHOENIX-7946 CDCQueryIT IllegalArgumentException in testSelectWithTimeRange

apurtell:
`CDCQueryIT` can fail with `IllegalArgumentException: bound must be positive at CDCQueryIT.java:786`, where the test calls `rand.nextInt(lastDeletionTSpos - 1)` to pick a random index into the unique timestamp list. `lastDeletionTSpos` is only assigned when the synthesized CDC change row stream contains at least one deletion encountered at a unique timestamp. Under the parameterized test matrix it can stay null, or be 0 or 1. The fix is to skip the random sub-range exercise when the available window is < 2.


palashc:
@apurtell I see @haridsv has a fix for this in https://github.com/apache/phoenix/pull/2475, if we are keeping that one, should we close this PR?

apurtell:
Sure we can close this.

PR #2558, opened 2026-06-29
PHOENIX-7947 PhoenixConnection.getTable NPE when auto-unboxing

apurtell:
`GlobalConnectionTenantTableIT` has three deterministically failing tests that all throw the same NPE ("Cannot invoke "java.lang.Long.longValue()" because "timestamp" is null at PhoenixConnection.java:643"). The root cause is in `PhoenixConnection.getTable(@Nullable String tenantId, String fullTableName, @Nullable Long timestamp)`. Its documented contract says a null timestamp means "use the cache, fall back to a server lookup on miss," and the cache hit path honors that, but the cache miss catch dispatches to `getTableNoCache(PName, String, long)` whose third parameter is a primitive long, so when `timestamp == null` the JVM auto-unboxes the null `Long` and throws a NPE.

### DIFF
--- a/phoenix-core-client/src/main/java/org/apache/phoenix/jdbc/PhoenixConnection.java
+++ b/phoenix-core-client/src/main/java/org/apache/phoenix/jdbc/PhoenixConnection.java
@@ -640,7 +640,13 @@ public class PhoenixConnection
         throw new TableNotFoundException(fullTableName);
       }
     } catch (TableNotFoundException e) {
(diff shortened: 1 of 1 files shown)


palashc (approved):
+1, pending build

palashc:
https://ci-hadoop.apache.org/job/Phoenix/job/Phoenix-PreCommit-GitHub-PR/job/PR-2558/1/

PR #2560, opened 2026-06-29
PHOENIX-7949 IndexScrutinyWithMaxLookbackIT scrutiny at SCN beyond max lookback counts updated rows as VALID

apurtell:
`IndexScrutinyWithMaxLookbackIT` updated only the covered column of a row and expected scrutiny to count it as `BEYOND_MAX_LOOKBACK`, but an in-place covered column update leaves the indexed column and thus the index row key unchanged, so the index row is never deleted. Major compaction retains the pre-update version as the row version visible at the lower edge of the max lookback window, and `IndexScrutinyTool` refuses to scan at an SCN older than max lookback. The fix adds `@Category(NeedsOwnMiniClusterTest.class)` and replaces the two invalid test cases with `testScrutinyOnUpdatedRowsWithinMaxLookbackAreValid` and `testScrutinyOnUpdatedViewIndexRowsWithinMaxLookbackAreValid`, which assert the correct invariants with the rationale documented in inline comments, while leaving `testScrutinyOnDeletedRowsBeyondMaxLookBack` unchanged.

### DIFF
--- a/phoenix-core/src/it/java/org/apache/phoenix/end2end/IndexScrutinyWithMaxLookbackIT.java
+++ b/phoenix-core/src/it/java/org/apache/phoenix/end2end/IndexScrutinyWithMaxLookbackIT.java
@@ -52,9 +52,11 @@ import org.apache.phoenix.util.TestUtil;
 import org.junit.Before;
 import org.junit.BeforeClass;
 import org.junit.Test;
+import org.junit.experimental.categories.Category;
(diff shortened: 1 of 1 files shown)


palashc:
https://ci-hadoop.apache.org/job/Phoenix/job/Phoenix-PreCommit-GitHub-PR/job/PR-2560/2/

PR #2562, opened 2026-06-30
PHOENIX-7945 - Retain orphaned delete markers during Phoenix compaction

ujjawal4046:
## Summary

Orphaned delete markers (DeleteFamily markers without corresponding puts) are dropped during major compaction before Phoenix CompactionScanner can process them.

## Root Cause

HBase `DropDeletesCompactionScanQueryMatcher.tryDropDelete()` drops delete markers whose timestamp < `earliestPutTs` when `KeepDeletedCells=TTL` and `timeToPurgeDeletes` is 0. Since `earliestPutTs` is a **global minimum across ALL HFiles** being compacted, a put in any other row HFile can cause orphaned markers to be dropped before Phoenix CompactionScanner ever sees them.

## Fix

Set `timeToPurgeDeletes = Long.MAX_VALUE` in `setScanOptionsForFlushesAndCompactions()`. This short-circuits `tryDropDelete()` so HBase never purges delete markers -- Phoenix CompactionScanner then applies its standard max-lookback logic.

## How CompactionScanner Preserves Orphan Markers Within Max-Lookback

Once `timeToPurgeDeletes` ensures HBase delivers all cells to CompactionScanner, the existing max-lookback logic handles orphan delete markers correctly:

In `getLastRowVersionInMaxLookbackWindow()`, the very first check is:
```java
if (cell.getTimestamp() > maxLookbackWindowStart) {
    retainedCells.add(cell);  // unconditionally retained regardless of cell type
    continue;
}
```

This retains ALL cells within the max-lookback window -- puts, delete markers, orphaned markers alike. No special handling needed. The marker goes directly into `retainedCells` and survives compaction.

For markers **outside** max-lookback: they fall through to the column-formation logic where `deleteFamilyCell` is set but `lastRowVersion` stays empty (no puts exist). Back in `retainCellsForMaxLookback`, `lastRowVersion.isEmpty()` returns true with empty `retainedCells` -- the marker is purged. This matches the behavior of normal deleted rows outside max-lookback.

## Orphan Delete Marker Lifecycle (with fix)

Same as normal deleted rows:

| Time Zone | Behavior |
|-----------|----------|
| Within max-lookback | Retained |
| Outside max-lookback (but within TTL) | **Purged** |
| Outside TTL | Purged |

Users who need markers retained beyond max-lookback (for replication lag) can use the per-table max-lookback override to extend it up to TTL.

## Comparison with HBase (without Phoenix)

| Configuration | Behavior for orphan markers |
|--------------|----------------------------|
| `KeepDeletedCells=FALSE` (default) | Always dropped (SKIP) |
| `KeepDeletedCells=TTL` | Still dropped via `earliestPutTs` bug |
| `timeToPurgeDeletes > 0` | Only reliable mechanism -- marker retained until `now - markerTs > timeToPurgeDeletes` |

Phoenix already sets `KeepDeletedCells=TTL` for its compaction scans, but that alone is insufficient because `tryDropDelete()` runs before the cell reaches CompactionScanner. Setting `timeToPurgeDeletes = Long.MAX_VALUE` is the equivalent of telling HBase "never purge markers -- Phoenix will handle it."

## Negative Test Validation

Verified that removing the `timeToPurgeDeletes` fix causes the within-max-lookback test to fail:

```
java.lang.AssertionError: Orphaned delete marker within max-lookback should be retained
after major compaction. Before=1, After=0
  at OrphanDeleteMarkerCompactionIT.testOrphanedDeleteMarkerRetainedWithinMaxLookback(OrphanDeleteMarkerCompactionIT.java:160)
```

This confirms:
- Without the fix, HBase drops the orphaned marker at the StoreScanner layer (`tryDropDelete`) even when the marker is well within the max-lookback window
- The existing CompactionScanner max-lookback logic is correct but never gets a chance to run because it never receives the cell

Co-authored-by: Claude Opus 4.8[1m] [noreply@anthropic.com](mailto:noreply@anthropic.com)

### DIFF
--- a/phoenix-core-server/src/main/java/org/apache/phoenix/coprocessor/BaseScannerRegionObserver.java
+++ b/phoenix-core-server/src/main/java/org/apache/phoenix/coprocessor/BaseScannerRegionObserver.java
@@ -383,6 +383,14 @@ abstract public class BaseScannerRegionObserver implements RegionObserver {
     options.setTTL(HConstants.FOREVER);
     options.setMaxVersions(Integer.MAX_VALUE);
     options.setMinVersions(Integer.MAX_VALUE);
+    // Prevent HBase from purging orphaned delete markers (those without corresponding
+    // puts) during compaction. Without this, DropDeletesCompactionScanQueryMatcher drops
+    // delete markers whose timestamp < earliestPutTs, before Phoenix CompactionScanner
+    // can see them. Once delivered to CompactionScanner, orphaned markers within the
+    // max-lookback window are retained by getLastRowVersionInMaxLookbackWindow (which
+    // unconditionally retains all cells with timestamp > maxLookbackWindowStart), and
+    // markers outside max-lookback are purged — same lifecycle as normal deleted rows.
+    options.setTimeToPurgeDeletes(Long.MAX_VALUE);
   }
 
   @Override
--- a/phoenix-core-server/src/main/java/org/apache/phoenix/coprocessor/CompactionScanner.java
+++ b/phoenix-core-server/src/main/java/org/apache/phoenix/coprocessor/CompactionScanner.java
@@ -2154,9 +2154,12 @@ public class CompactionScanner implements InternalScanner {
       top: for (int index = 0; index < result.size(); index++) {
         Cell cell = result.get(index);
         if (cell.getTimestamp() > maxLookbackWindowStart) {
-          // All cells within the max lookback window are retained. Here we retain all
-          // except the ones at the lower edge of the window. Those will be included in
-          // the last row version in the rest of the body of the loop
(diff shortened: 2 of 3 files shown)


virajjasani:
https://ci-hadoop.apache.org/job/Phoenix/job/Phoenix-PreCommit-GitHub-PR/job/PR-2562/

virajjasani:
https://ci-hadoop.apache.org/job/Phoenix/job/Phoenix-PreCommit-GitHub-PR/job/PR-2562/1/PR_20General_20Check_20Report/

```
-1 | spotless | 0m 38s |   | patch has 63 errors when running spotless:check, run spotless:apply to fix.
-- | -- | -- | -- | --

```

virajjasani (approved):
+1

PR #2566, opened 2026-07-06
PHOENIX-7874 Eliminate replicating index mutations; regenerate on standby

tkhurana:
## Summary

Index tables no longer generate their own replication-log records. The active ships only the data-table record plus a per-`(row, ts)` **PRE_IMAGE** (the prior row state observed at lock time), and the standby regenerates every index entry from that record on the replay path.

This targets the `PHOENIX-7562-feature-new` feature branch (builds on the per-batch coalescing / cell-oriented log format, #2540).

## Why

- **Out-of-order replay safety.** Because each replicated batch carries the exact prior-row snapshot it was built against, the standby derives each `(row, ts)` group's next state from its own shipped pre-image instead of a region scan. Shard files can be replayed in any order or in parallel; each group is an independent reproduction of the active `preBatchMutate`.
- **Cross-cluster rowkeys.** CDC/global/local index rowkeys can embed the encoded data-table region name (`PARTITION_ID`), which differs between clusters. Replicating index cells verbatim would carry the active's region name; the standby instead rebuilds the rowkey with its own.

## Main changes

- **Cell-oriented replication format** (codec + record + `MutationCellGrouper`) carrying data cells plus one METAFAMILY pre-image cell per row.
- **`IndexRegionObserver`**: active-side pre-image capture; a forked standby replay path (`prepareReplicatedIndexMutations`) that groups by `(row, ts)` and builds index updates from the shipped pre-image.
- **`PreImageLocalTable`**: a `LocalHBaseState` serving the local-index builder's prior-row-state from the shipped pre-image instead of a region scan; new `IndexBuildManager.getIndexUpdates` overload accepts it.
- **`ReplicationLogProcessor`**: replay a record as one indivisible batch; do not retry `DoNotRetryIOException`.

## Test coverage

Global, local, uncovered, atomic, conditional-TTL, and CDC (plain and behind an EVENTUAL secondary index) index regeneration; concurrent/out-of-order replay for both global and local indexes; unit tests for `(row, ts)` grouping isolation and Put/Delete decomposition.

## Notes

- Squashed into a single commit.
- Please squash-merge.

### DIFF
--- a/phoenix-core/src/it/java/org/apache/phoenix/replication/ReplicationLogGroupEventualIndexIT.java
+++ b/phoenix-core/src/it/java/org/apache/phoenix/replication/ReplicationLogGroupEventualIndexIT.java
@@ -17,6 +17,9 @@
  */
 package org.apache.phoenix.replication;
 
+import static org.apache.phoenix.hbase.index.IndexCDCConsumer.INDEX_CDC_CONSUMER_POLL_INTERVAL_MS;
+import static org.apache.phoenix.hbase.index.IndexCDCConsumer.INDEX_CDC_CONSUMER_STARTUP_DELAY_MS;
+import static org.apache.phoenix.hbase.index.IndexCDCConsumer.INDEX_CDC_CONSUMER_TIMESTAMP_BUFFER_MS;
 import static org.apache.phoenix.hbase.index.IndexRegionObserver.PHOENIX_INDEX_CDC_CONSUMER_ENABLED;
 import static org.apache.phoenix.query.BaseTest.generateUniqueName;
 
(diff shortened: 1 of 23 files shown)


phoenix-core/src/it/java/org/apache/phoenix/replication/ReplicationLogGroupEventualIndexIT.java
```diff
@@ -0,0 +1,144 @@
...
+
+  /**
+   * Disables the IndexCDCConsumer on both clusters, then starts them. CDC-index regeneration is
+   * verified directly and the consumer's downstream secondary-index convergence is out of scope, so
+   * it stays off. The {@code serializeCDCMutations} subclass calls this after setting its own
+   * toggle so the consumer-disable lives in one place.
+   */
+  protected static void setupEventualIndexClusters() throws Exception {
+    conf1.setBoolean(PHOENIX_INDEX_CDC_CONSUMER_ENABLED, false);
+    conf2.setBoolean(PHOENIX_INDEX_CDC_CONSUMER_ENABLED, false);
```
virajjasani: do we have coverage for checking if eventually consistent indexes also have rows matching after waiting for some time?

virajjasani (approved):
+1, left one question

phoenix-core/src/it/java/org/apache/phoenix/replication/ReplicationLogGroupEventualIndexIT.java
tkhurana: @virajjasani Added the test for eventually consistent index

PR #2569, opened 2026-07-07
PHOENIX-7955 CREATE TABLE IF NOT EXISTS should not mutate existing HBase table properties

PDavid:
#### Problem

When CREATE TABLE IF NOT EXISTS was executed against an already-existing table, Phoenix would overwrite the HBase-level table and column family properties (e.g. REGION_REPLICATION, VERSIONS, TTL) with the values from the new DDL statement. This is unexpected - IF NOT EXISTS should be a no-op if the table already exists.

(Please see the ticket about reproduction steps.)

#### Fix

Added new method parameter `boolean createIfNotExists` to `ConnectionQueryServices.createTable()` method to signal in CREATE statement was issued with IF NOT EXISTS. 

- When CREATE TABLE IF NOT EXISTS was executed and table already exists:
  - and the table is already a Phoenix-managed table we return early, don't modify HBase table-level properties.
  - and plain HBase table being mapped we fall through, add coprocessors via modifyTable.
- When CREATE TABLE IF NOT EXISTS was executed and table DOES NOT exists, no change.
- When CREATE TABLE without IF NOT EXISTS is executed, no change.

No.

Added new test in CreateTableIT which reproduces the problem and verifies the fix.

Generated-by: Claude Opus 4.6 <noreply@anthropic.com>

### DIFF
--- a/phoenix-core-client/src/main/java/org/apache/phoenix/query/ConnectionQueryServicesImpl.java
+++ b/phoenix-core-client/src/main/java/org/apache/phoenix/query/ConnectionQueryServicesImpl.java
@@ -1992,6 +1992,11 @@ public class ConnectionQueryServicesImpl extends DelegateQueryServices
         }
         return null;
       } else {
+        if (createIfNotExists && doesPhoenixTableAlreadyExist(existingDesc)) {
+          // Already a Phoenix-managed table, skip remaining modifications entirely.
+          return existingDesc;
+        }
+        // We create phoenix table that maps to existing HBase table - we shall continue to add coprocessors.
         if (isMetaTable && !isUpgradeRequired()) {
           checkClientServerCompatibility(
             SchemaUtil.getPhysicalName(SYSTEM_CATALOG_NAME_BYTES, this.getProps()).getName());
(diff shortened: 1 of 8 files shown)


phoenix-core-client/src/main/java/org/apache/phoenix/query/ConnectionQueryServicesImpl.java
```diff
@@ -2052,8 +2053,8 @@ private TableDescriptor ensureTableCreated(byte[] physicalTableName,
           return null; // Indicate that no metadata was changed
         }
 
-        // Do not call modifyTable for SYSTEM tables
-        if (tableType != PTableType.SYSTEM) {
+        // Do not call modifyTable for: SYSTEM tables or when CREATE IF NOT EXISTS was used.
+        if (tableType != PTableType.SYSTEM && !createIfNotExists) {
```
richardantal: Couldn't we return earlier? When we realize that `tableExist` or similarly to `modifyExistingMetaData`?

PDavid:
I still need to re-run tests because some were failed as I remember and fix the tests.

PR #2570, opened 2026-07-08
PHOENIX-7938 Round-align the sync-replication replay consistency point

Himanshu-g81:
**JIRA**: https://issues.apache.org/jira/browse/PHOENIX-7938 (sub-task of PHOENIX-7562)
  
**Problem**:  In SYNC state, getConsistencyPoint() used the raw minimum timestamp of the IN-PROGRESS files. Since files within a round move to IN-PROGRESS in random order, that minimum can belong to a later file while an older sibling still waits in IN — advancing the consistency point past unreplayed data. Because the compaction guard treats it as an exclusive bound (consistencyPoint - 1), it could drop delete markers for unreplayed data (data-resurrection risk).
  
**Fix**: align the minimum IN-PROGRESS timestamp down to its round start. Every file in round N has ts >= roundStart(N) and earlier rounds are fully replayed first, so roundStart(N) is a safe, conservative bound requiring no IN-directory listing. Only the SYNC branch changes; the guard formula and shared helpers are untouched, and the configurable round duration is honored.

**Tests**: new unit tests (out-of-order, boundary, multi-round) and a mini-cluster IT. Pre-fix code returns the raw minimum and fails these assertions.

### DIFF
--- a/phoenix-core-server/src/main/java/org/apache/phoenix/replication/reader/ReplicationLogDiscoveryReplay.java
+++ b/phoenix-core-server/src/main/java/org/apache/phoenix/replication/reader/ReplicationLogDiscoveryReplay.java
@@ -584,7 +584,7 @@ public class ReplicationLogDiscoveryReplay extends ReplicationLogDiscovery {
           // round that are waiting in the IN directory (files are picked in random order).
           long minTimestampInProgress = optionalMinTimestampInProgressTimestamp.get();
           consistencyPoint = replicationLogTracker.getReplicationShardDirectoryManager()
-            .getReplicationRoundFromStartTime(minTimestampInProgress).getStartTime();
+            .getNearestRoundStartTimestamp(minTimestampInProgress);
         } else if (lastRoundInSync != null) {
           // Use lastRoundInSync end time if no in-progress files
           // Since we are in sync mode, both lastRoundProcessed and lastRoundInSync would be same.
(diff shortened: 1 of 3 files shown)


phoenix-core-server/src/main/java/org/apache/phoenix/replication/reader/ReplicationLogDiscoveryReplay.java
```diff
@@ -571,8 +579,12 @@ public long getConsistencyPoint() throws IOException {
         Optional<Long> optionalMinTimestampInProgressTimestamp =
           getMinTimestampFromInProgressFiles();
         if (optionalMinTimestampInProgressTimestamp.isPresent()) {
-          // Use minimum timestamp from in-progress files as consistency point
-          consistencyPoint = optionalMinTimestampInProgressTimestamp.get();
+          // Align the minimum in-progress timestamp down to the start of the round it belongs to,
+          // so the consistency point never advances past older, still-unreplayed files of the same
+          // round that are waiting in the IN directory (files are picked in random order).
+          long minTimestampInProgress = optionalMinTimestampInProgressTimestamp.get();
+          consistencyPoint = replicationLogTracker.getReplicationShardDirectoryManager()
```
tkhurana, line 586: Simpler to use `getNearestRoundStartTimestamp` ?

tkhurana (changes requested):

tkhurana (approved):

PR #2574, opened 2026-07-16
PHOENIX-7961 Secondary index diverges from data table after TTL expiry on partial-touch upserts

sanjeet006py:

This PR makes Phoenix's internal current-row read during secondary-index maintenance TTL-mask exactly like an ordinary client read, so a
  data table and its secondary index stay consistent after TTL expiry.

  Index maintenance in IndexRegionObserver.preBatchMutateWithExceptions reads the current on-disk row via getCurrentRowStates to rebuild
  the correct index entry. That read was opened directly through region.getScanner(scan), which bypasses the postScannerOpen coprocessor
  hook — the only place a scan is normally wrapped in TTLRegionScanner. As a result the internal read was not TTL-masked, so on a "partial
  touch" upsert (an UPSERT that does not re-write any index-referenced column) the index was rebuilt from
  logically-expired-but-not-yet-compacted cells, while the data table correctly expired them on read.

  Changes:

  1. Client side — ScanUtil.annotateMutationWithLiteralTTL (called from MutationState) threads onto each mutation exactly what the client
  read path sets as scan attributes: the empty-column CF/CQ (unconditionally for any literal-TTL table/view), a view's compiled literal
  TTL as the _TTL attribute (base tables rely on the CF-descriptor TTL), and IS_STRICT_TTL=false only when the table/view is non-strict.
  2. Server side — new ServerScanUtil (package org.apache.phoenix.coprocessor, module phoenix-core-server) reproduces the client read path
  server-side: setInternalScanAttributes / setInternalScanAttributesForPaging set the same empty-column/TTL/strict/paging attributes, and
  openRegionScanner wraps the scan in TTLRegionScanner + PagingRegionScanner exactly as postScannerOpen does. IndexRegionObserver
  captures the client-threaded attributes into the batch context (extractLiteralTTLForInternalScan) and passes them into
  getCurrentRowStates.
  3. Anchor the masking clock at batchTimestamp — batchTimestamp is computed before the current-row read, and the internal scan sets
  scan.setTimeRange(0, batchTimestamp) so TTLRegionScanner's "current time" equals the exact timestamp at which the index is rebuilt,
  closing a sub-millisecond boundary window between the scan-open wall clock and batchTimestamp. The half-open range drops nothing
  current, since every pre-existing locked-row cell was written at ts < batchTimestamp.
  4. Companion correctness fix — moving getBatchTimestamp earlier means a registered set could be structurally mutated outside its
  synchronized block; batchesWithLastTimestamp now registers a defensive TreeSet snapshot to avoid a ConcurrentModificationException,
  staying conservative for the sleep/timestamp invariant (at most an extra sleep, never a missed one).

  Production files touched: IndexRegionObserver, ScanUtil, new ServerScanUtil, MutationState, and a doc-comment update in MetaDataClient.

This is a silent data-integrity/correctness bug. On a table or view with a literal TTL and a secondary index, after TTL expiry the
  secondary index can return a column value the data table no longer returns — the two diverge with no error raised. The divergence
  surfaces after logical TTL expiry and before major compaction physically purges the expired cells.

  The root cause is that the internal current-row read bypassed TTLRegionScanner, so index rebuild saw expired cells the data-side read
  masks. The affected paths are: secondary indexes (global covered, global uncovered, and immutable indexes whose data/index storage
  schemes differ) and the no-index current-row reads on a literal-TTL table (atomic / ON DUPLICATE KEY upserts, returnResult upserts, and
  row deletes). Conditional-TTL, non-TTL, and non-strict-TTL tables are unaffected — masking is a no-op there and the read is
  byte-identical to before.

Yes — a bug fix (behavior change) relative to master and released versions, for tables/views with a literal TTL and a secondary index.

  - Before: on a partial-touch upsert near/after the TTL boundary, the secondary index could return a value that the data table no longer
  returned (index resurrected a logically-expired value); data and index disagreed.
  - After: the internal current-row read is TTL-masked identically to a client read and anchored at batchTimestamp, so the index is
  rebuilt from the same masked state the data table exposes, and the two stay consistent.

  No API, syntax, or configuration change. No new config flag is introduced.

New integration test IndexDataTableSyncIT (parameterized over column-encoded on/off), covering:

  - base-table covered-column re-sync;
  - uncovered-index indexed-column re-sync;
  - view covered-column re-sync;
  - non-strict table is not masked;
  - no-index atomic (ON DUPLICATE KEY) upsert masks an expired row;
  - UPSERT ... SELECT and UPSERT ... VALUES touches near the TTL boundary keep data and index consistent;
  - an immutable index with a differing storage scheme;
  - a concurrency regression (testConcurrentMajorCompactionDuringIndexWriteKeepsDataAndIndexConsistent) where a data-table major
  compaction runs during a slow index write — after the current-row read and after the data locks are released, but before the index write
  completes — and data/index must remain consistent.

Generated-by: Claude Code (Anthropic Claude Opus 4.8)


sanjeet006py:
PR validation: https://ci-hadoop.apache.org/job/Phoenix/job/Phoenix-PreCommit-GitHub-PR/job/PR-2574/

phoenix-core-client/src/main/java/org/apache/phoenix/execute/MutationState.java
```diff
@@ -1528,6 +1528,10 @@ private void sendMutations(Iterator<Entry<TableInfo, List<Mutation>>> mutationsI
         // no-op if table doesn't have Conditional TTL
         ScanUtil.annotateMutationWithConditionalTTL(connection, tableInfo.getPTable(),
           mutationList);
+        // no-op unless table/view has a literal TTL; threads the empty-column CF/CQ (plus a view's
+        // literal TTL and any non-strict flag) so the internal current-row scan masks like a client
+        // read
+        ScanUtil.annotateMutationWithLiteralTTL(connection, tableInfo.getPTable(), mutationList);
```
tkhurana, line 1534: Did you verify that conditional ttl doesn't have this problem ? I didn't see any test cases with conditional ttl

phoenix-core-server/src/main/java/org/apache/phoenix/hbase/index/IndexRegionObserver.java
```diff
@@ -1854,7 +1967,8 @@ public void preBatchMutateWithExceptions(ObserverContext<RegionCoprocessorEnviro
           || !context.immutableRows && context.hasUncoveredIndex
             && isPartialUncoveredIndexMutation(indexMetaData, miniBatchOp)
       ) {
-        getCurrentRowStates(c, context);
+        getCurrentRowStates(c, context, context.literalTTLForInternalScan,
+          isStrictTTLEnabled(miniBatchOp), batchTimestamp);
```
tkhurana: We can save is strict ttl in also in the context. You are already saving ttl and empty column family and qualifier. That will make the API cleaner.

phoenix-core-server/src/main/java/org/apache/phoenix/hbase/index/IndexRegionObserver.java
```diff
@@ -1228,20 +1270,41 @@ private void getCurrentRowStates(ObserverContext<RegionCoprocessorEnvironment> c
...
         if (cells.isEmpty()) {
           continue;
         }
+        // With server paging wired in (ServerScanUtil.setInternalScanAttributes),
+        // PagingRegionScanner
+        // returns a dummy result when a page is paged out; skip it and let the loop resume rather
+        // than build a Put from the dummy cell.
+        if (ScanUtil.isDummy(cells)) {
+          continue;
+        }
```
tkhurana: I am not sure what are we gaining with paging here ?

phoenix-core-server/src/main/java/org/apache/phoenix/coprocessor/ServerScanUtil.java
```diff
@@ -0,0 +1,134 @@
...
+    }
+    scan.setAttribute(BaseScannerRegionObserverConstants.SERVER_PAGE_SIZE_MS,
+      Bytes.toBytes(Long.valueOf(pageSizeMs)));
+    // Wrap the scan filter in a PagingFilter as the top-level filter, matching
+    // BaseScannerRegionObserver.preScannerOpen. PagingRegionScanner then detects when PagingFilter
+    // has paged the scan out and returns a dummy result; readDataTableRows skips those dummies.
+    if (!(scan.getFilter() instanceof PagingFilter)) {
+      scan.setFilter(new PagingFilter(scan.getFilter(), ScanUtil.getPageSizeMsForFilter(scan)));
+    }
+  }
```
tkhurana: This whole code has been duplicated from the client but there seems to be little benefit since we are doing a region local scan which doesn't go through the rpc path so what is paging benefit we are trying to achieve ?

tkhurana (changes requested):

phoenix-core-client/src/main/java/org/apache/phoenix/execute/MutationState.java
sanjeet006py: I observed that conditional TTL case was already covered by `IndexRegionObserver#updateMutationsForConditionalTTL`. There we check if a row is already expired and add delete markers.

I didn't explicitly verify via IT. Do you think I should add one?

phoenix-core-server/src/main/java/org/apache/phoenix/hbase/index/IndexRegionObserver.java
sanjeet006py: There is no explicit functional gain rather this is scan path and my understanding is we want to cover all scan paths via paging to ensure fair utilization of server resources/handler threads. Thus, added paging explicitly. This way internal scan done by IndexRegionObserver by directly opening region scanner won't bypass paging.

sanjeet006py: Oh, I didn't realize that if its not going through RPC path then it won't even consume additional handler threads. Yeah, there is no benefit of paging in here. Thanks, will remove it.

phoenix-core-server/src/main/java/org/apache/phoenix/coprocessor/ServerScanUtil.java
sanjeet006py: Thanks for highlighting this explicitly that no handler threads will be involved here.

phoenix-core-server/src/main/java/org/apache/phoenix/coprocessor/ServerScanUtil.java
```diff
@@ -78,57 +76,21 @@ public static void setInternalScanAttributes(Configuration conf, Scan scan, byte
...
-   * — so wrapping a non-TTL scan changes no behavior.
+   * Unlike the client read path this does not wrap in a {@code PagingRegionScanner}: the scan is
+   * region-local (opened directly on the {@link Region}, off the RPC path), so it holds no handler
+   * thread and has nothing for paging to protect.
    */
   public static RegionScanner openRegionScanner(RegionCoprocessorEnvironment env, Region region,
     Scan scan) throws IOException {
-    return new TTLRegionScanner(env, scan,
-      new PagingRegionScanner(region, region.getScanner(scan), scan));
+    return new TTLRegionScanner(env, scan, region.getScanner(scan));
```
tkhurana: It might not be safe in all cases to directly pass the hbase regionscanner as the delegate to TTLRegionScanner. There are code paths where we cast to DelegateRegionScanner.

phoenix-core-client/src/main/java/org/apache/phoenix/coprocessorclient/BaseScannerRegionObserverConstants.java
```diff
@@ -131,6 +131,12 @@ public static long getMaxLookbackInMillis(Configuration conf) {
   public static final String SKIP_REGION_BOUNDARY_CHECK = "_SKIP_REGION_BOUNDARY_CHECK";
   public static final String TX_SCN = "_TxScn";
   public static final String TTL = "_TTL";
+  // Literal TTL threaded per-mutation for the server-side internal current-row scan
```
tkhurana: I think it is better to use the same TTL attribute for both conditional and literal ttl. We already do that on the scan path. There is no reason we can't do that on the mutation path also.

sanjeet006py: I intentionally used a new mutation attribute for literal TTL as currently in `updateMutationsForConditionalTTL` there is a blind cast of TTL expression to Conditional TTL expression and moreover, in IndexRegionObserver there is implicit assumption that if `_TTL` attribute is set on a mutation then its conditional TTL. If client changes in this PR gets deployed earlier then server changes, then that can break mutation path by throwing ClassCastException.

phoenix-core-client/src/main/java/org/apache/phoenix/execute/MutationState.java
tkhurana: Its ok. There are already existing tests.

phoenix-core-client/src/main/java/org/apache/phoenix/coprocessorclient/BaseScannerRegionObserverConstants.java
tkhurana: The current check was done because it was a cheap way to do it but the server can deserialize the ttl expression and can easily determine if the expression is a literal ttl or conditional ttl. If the server is upgraded first which is typically the case it will remain backward compatible to any client using conditional ttl. I am not sure if we really need to worry about the case of the client getting deployed before server. That can cause all sorts of other issues.

sanjeet006py: Had offline discussion, will unify the attributes. Thanks

sanjeet006py: @tkhurana I have addressed the suggestion. PTAL

phoenix-core-server/src/main/java/org/apache/phoenix/hbase/index/IndexRegionObserver.java
```diff
@@ -1770,18 +1768,10 @@ private void extractLiteralTTLForInternalScan(MiniBatchOperationInProgress<Mutat
...
-      // Conditional TTL: leave the attribute in place for updateMutationsForConditionalTTL.
-      return;
-    }
-    context.literalTTLForInternalScan = ttlBytes;
-    for (int i = 0; i < miniBatchOp.size(); i++) {
-      miniBatchOp.getOperation(i).setAttribute(BaseScannerRegionObserverConstants.TTL, null);
+    CompiledTTLExpression ttlExpr = TTLExpressionFactory.create(ttlBytes);
+    context.ttlExpressionForBatch = ttlExpr;
+    if (context.isLiteralTTL()) {
+      context.literalTTLForInternalScan = ttlBytes;
```
tkhurana: Do we need a separate field for literal ttl ?

phoenix-core-server/src/main/java/org/apache/phoenix/hbase/index/IndexRegionObserver.java
```diff
@@ -794,18 +808,16 @@ private void addOnDupMutationsToBatch(MiniBatchOperationInProgress<Mutation> min
    */
   private void updateMutationsForConditionalTTL(MiniBatchOperationInProgress<Mutation> miniBatchOp,
     BatchMutateContext context) throws IOException {
-    // If TTL is not strict, skip conditional TTL processing
-    if (!context.isStrictTTLForInternalScan) {
+    if (!context.isConditionalTTL()) {
```
tkhurana: Are we mixing the two concepts here conditional ttl and literal ttl and strict vs non-strict ttl. These are independent properties and all 4 combinations are possible and permitted.

phoenix-core-server/src/main/java/org/apache/phoenix/hbase/index/IndexRegionObserver.java
sanjeet006py: Ideally no. But instead of reserializiing TTL Expression (which we have already extracted) for literal TTL, I thought why not store that also as bytes. Literal TTL in bytes form doesn't add much overhead.

phoenix-core-server/src/main/java/org/apache/phoenix/hbase/index/IndexRegionObserver.java
sanjeet006py: The invocation of method `updateMutationsForConditionalTTL` is gated by pre-existing `hasConditionalTTL` in `preBatchMutateWithExceptions`. `hasConditionalTTL` is true when mutation batch has conditionalTTL along with strict TTL being true. Given conditional TTL and strict TTL are two independent properties then gating by the caller of the method makes sense so, I removed the check from method itself. 

But separately added a check for conditionalTTL explicitly as this method should never be called when its not conditional TTL. Its more of defense in depth check as whole method is relevant for conditional TTL only.

So, I am not mixing two concepts.

sanjeet006py: If you think this can cause confusion then I can add the check for strictTTL back. This was a nit change so, I am fine either way.

phoenix-core-server/src/main/java/org/apache/phoenix/hbase/index/IndexRegionObserver.java
tkhurana: If you want to store raw bytes I would just store the raw ttl bytes all the time and rename the variable accordingly.

phoenix-core-server/src/main/java/org/apache/phoenix/hbase/index/IndexRegionObserver.java
tkhurana: There is confusion. We have `hasConditionalTTL` which is true when when mutation batch has conditionalTTL along with strict TTL being true. That is confusing naming wise. We also have `context.hasConditionalTTL` and `context.isConditionalTTL`.

tkhurana (approved):

tkhurana:
@sanjeet006py Thanks for making the changes. This looks neat and clean now.

PR #2576, opened 2026-07-17
PHOENIX-7562 Add HAGroupStore Metrics2 observability

ritegarg:
Expose focused per-HA-group observability without coupling metrics to HA decisions.

- Add a process-lifetime Hadoop Metrics2 source per HA group with 5 gauges, 6 counters, and 1 listener-latency histogram.
- Export zero-is-good local-cache and peer-visibility status gauges, stable non-ordinal HA state codes, and selected episode/failure counters.
- Tag each source by HA group and safely quote its JMX identity.
- Add focused unit and two-cluster integration coverage for the exported contract and runtime behavior.

HAGroupStore currently relies on logs for cache health, peer reachability, state divergence, fail-closed behavior, background failures, and slow synchronous listeners. These metrics provide alertable current state and targeted diagnostics while remaining independent of HA control flow.

Yes. RegionServers expose new HAGroupStore Metrics2/JMX metrics. There is no SQL, JDBC, or HA decision behavior change.

- Unit tests:
  `mvn -pl phoenix-core -am -DskipITs -Dtest=HAGroupStoreRecordTest,HAGroupStoreMetricsSourceImplTest -Dsurefire.failIfNoSpecifiedTests=false -DfailIfNoTests=false test`
- Integration tests:
  `mvn -pl phoenix-core -am -Dtest=NoSuchUnitTest -Dsurefire.failIfNoSpecifiedTests=false -DfailIfNoTests=false -Dit.test=HAGroupStoreMetricsIT -DskipParallelStatsEnabledTests=true -DskipParallelStatsDisabledTests=true verify`
- Result: 27 unit tests and 9 integration tests passed.

Generated-by: Cursor 3.9.16

### DIFF
--- a/phoenix-core-client/src/main/java/org/apache/phoenix/jdbc/HAGroupStoreClient.java
+++ b/phoenix-core-client/src/main/java/org/apache/phoenix/jdbc/HAGroupStoreClient.java
@@ -695,8 +695,21 @@ public class HAGroupStoreClient implements Closeable {
 
   // Update the system table on best effort basis for HA group
   // In case of failure, we will log the error and continue.
-  private void updateSystemTableHAGroupRecordSilently(String haGroupName,
-    SystemTableHAGroupRecord record) throws SQLException {
+  private boolean updateSystemTableHAGroupRecordSilently(String haGroupName,
+    SystemTableHAGroupRecord record) {
+    try {
+      updateSystemTableHAGroupRecord(haGroupName, record);
+      return true;
+    } catch (Exception e) {
+      metricsSource.incrementSystemTableSyncFailedCount();
+      LOGGER.error("Failed to update system table on best effort basis for HA group {}", haGroupName,
+        e);
+      return false;
+    }
+  }
+
+  private void updateSystemTableHAGroupRecord(String haGroupName, SystemTableHAGroupRecord record)
+    throws SQLException {
     StringBuilder updateQuery = new StringBuilder("UPSERT INTO " + SYSTEM_HA_GROUP_NAME + " (");
     StringBuilder valuesClause = new StringBuilder(" VALUES (");
     List<Object> parameters = new ArrayList<>();
@@ -803,11 +816,6 @@ public class HAGroupStoreClient implements Closeable {
 
       pstmt.executeUpdate();
       conn.commit();
-    } catch (Exception e) {
-      metricsSource.incrementSystemTableSyncFailedCount();
-      LOGGER.error(
-        "Failed to update system table on best" + "effort basis for HA group {}, error: {}",
-        haGroupName, e);
     }
   }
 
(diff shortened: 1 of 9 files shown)


phoenix-core-client/src/main/java/org/apache/phoenix/jdbc/HAGroupStoreClient.java
```diff
@@ -795,6 +804,7 @@ private void updateSystemTableHAGroupRecordSilently(String haGroupName,
       pstmt.executeUpdate();
       conn.commit();
     } catch (Exception e) {
+      metricsSource.incrementSystemTableSyncFailedCount();
```
tkhurana: So this metric is only accounting for write failures but it is ignoring all other failures which can happen before the call even reaches here.

phoenix-core-client/src/main/java/org/apache/phoenix/jdbc/HAGroupStoreClient.java
tkhurana, line 904: Just noticed that we are not logging the exception here

tkhurana: There are other reasons which can cause exceptions but the metric is not tracking that.

tkhurana (changes requested):

phoenix-core-client/src/main/java/org/apache/phoenix/jdbc/HAGroupStoreClient.java
ritegarg: Addressed in 64a625c417. Both synchronization error logs now pass the exception as the trailing SLF4J argument, so the full stack trace and root cause are preserved.

phoenix-core-client/src/main/java/org/apache/phoenix/jdbc/HAGroupStoreClient.java
ritegarg: Addressed in 64a625c417. The metric now covers the complete synchronization attempt: failures while preparing or writing SYSTEM.HA_GROUP are counted in the helper, while periodic failures before the write are counted by the outer sync handler. The paths are mutually exclusive, so each failed attempt increments once.

phoenix-core-client/src/main/java/org/apache/phoenix/jdbc/HAGroupStoreClient.java
ritegarg: Addressed in 64a625c417. Periodic synchronization now catches and counts ZK reads, peer reads, policy/URL preparation, system-table comparison, and write failures. Write-helper failures are swallowed after counting and return false, preventing double-counting and false success logs.

tkhurana (approved):

PR #2577, opened 2026-07-17
PHOENIX-7920 Add replay-side handling for direct DEGRADED_STANDBY → STANDBY_TO_ACTIVE transition (ANISTS → AISTS)

Himanshu-g81:
JIRA : https://issues.apache.org/jira/browse/PHOENIX-7920

### DIFF
--- a/phoenix-core-server/src/main/java/org/apache/phoenix/replication/reader/ReplicationLogDiscoveryReplay.java
+++ b/phoenix-core-server/src/main/java/org/apache/phoenix/replication/reader/ReplicationLogDiscoveryReplay.java
@@ -127,6 +127,13 @@ public class ReplicationLogDiscoveryReplay extends ReplicationLogDiscovery {
           clusterType == ClusterType.LOCAL
             && HAGroupStoreRecord.HAGroupState.DEGRADED_STANDBY.equals(toState)
         ) {
+          // Unconditional set (not compareAndSet): degradation is an authoritative, fail-closed
+          // signal. Whatever the current state (SYNC, SYNCED_RECOVERY, or even NOT_INITIALIZED if a
+          // degrade lands mid-init), replay must stop advancing the sync point. There is no single
+          // valid "from" state to CAS against, so a CAS would silently drop legitimate degrade
+          // signals arriving from the other states. Contrast triggerFailoverListner below, which is
+          // deliberately conditional (compareAndSet from DEGRADED only) so it cannot clobber a
+          // healthy SYNC or an already-pending SYNCED_RECOVERY.
           replicationReplayState.set(ReplicationReplayState.DEGRADED);
           LOG.info("Cluster degraded detected for {}. replicationReplayState={}", haGroupName,
             ReplicationReplayState.DEGRADED);
@@ -216,9 +223,11 @@ public class ReplicationLogDiscoveryReplay extends ReplicationLogDiscovery {
    * (lastRoundInSync from the minimum of lastSyncStateTimeInMs and the file frontier, so it
    * represents the last consistent point before degradation).</li>
    * <li>STANDBY_TO_ACTIVE: a restart while already mid-failover (e.g. after a direct
-   * DEGRADED_STANDBY -&gt; STANDBY_TO_ACTIVE transition). Sets replicationReplayState to
-   * SYNCED_RECOVERY and initializes both rounds the same way, so the first replay() rewinds to
-   * lastRoundInSync before failover can promote.</li>
+   * DEGRADED_STANDBY -&gt; STANDBY_TO_ACTIVE transition). Initializes both rounds from the last
+   * known good sync point, then sets replicationReplayState to SYNCED_RECOVERY only when a rewind
+   * is actually pending (lastRoundInSync behind lastRoundProcessed) so the first replay() rewinds
+   * before failover can promote; when the rounds are already equal it initializes directly to SYNC
+   * so promotion is not delayed by a round window.</li>
    * <li>Other states (e.g. STANDBY): sets replicationReplayState to SYNC, delegates to the parent
    * to initialize lastRoundProcessed, and sets lastRoundInSync equal to lastRoundProcessed.</li>
    * </ul>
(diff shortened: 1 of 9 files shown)


ritegarg (approved):
Approved from HAGroupStore perspective

phoenix-core/src/it/java/org/apache/phoenix/replication/CrossClusterReplicationTestUtil.java
```diff
@@ -0,0 +1,117 @@
...
+import org.apache.hadoop.hbase.client.ConnectionFactory;
+import org.apache.hadoop.hbase.client.Result;
+import org.apache.hadoop.hbase.client.ResultScanner;
+import org.apache.hadoop.hbase.client.Scan;
+import org.apache.hadoop.hbase.client.Table;
+import org.apache.phoenix.util.TestUtil;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/** Shared cross-cluster helpers for replication ITs (extracted from ReplicationLogGroupIT). */
```
tkhurana, line 41: Good to reduce code duplication by refactoring ReplicationLogGroupIT

phoenix-core-server/src/main/java/org/apache/phoenix/replication/reader/ReplicationLogDiscoveryReplay.java
```diff
@@ -139,7 +150,15 @@ public void init() throws IOException {
...
         ) {
-          replicationReplayState.set(ReplicationReplayState.SYNCED_RECOVERY);
+          // compareAndSet from DEGRADED only, not an unconditional set(): DEGRADED is the only
+          // state whose lastRoundInSync lags lastRoundProcessed, hence the only state a recovery
+          // to STANDBY must rewind from. A STANDBY event that lands while already SYNC (e.g.
+          // ABORT_TO_STANDBY -> STANDBY after a prior rewind, or a cache-reconnect redelivery)
+          // would otherwise flip SYNC -> SYNCED_RECOVERY and needlessly re-process the frontier
+          // round. Symmetric with triggerFailoverListner below; contrast degradedListener above,
+          // whose unconditional fail-closed set() is intentional.
+          replicationReplayState.compareAndSet(ReplicationReplayState.DEGRADED,
```
tkhurana: The return value of compareAndSet is ignored ?

phoenix-core-server/src/main/java/org/apache/phoenix/replication/reader/ReplicationLogDiscoveryReplay.java
```diff
@@ -151,6 +170,13 @@ public void init() throws IOException {
           clusterType == ClusterType.LOCAL
             && HAGroupStoreRecord.HAGroupState.STANDBY_TO_ACTIVE.equals(toState)
         ) {
+          // Direct DEGRADED_STANDBY -> STANDBY_TO_ACTIVE skips the STANDBY event that normally
+          // drives recovery. If we are DEGRADED, schedule the rewind so replay() re-syncs from
+          // lastRoundInSync before shouldTriggerFailover() (which gates on SYNC) can promote.
+          // compareAndSet, not set: a listener firing while already SYNC (healthy failover) or
+          // SYNCED_RECOVERY (rewind already pending) must not clobber a good state.
+          replicationReplayState.compareAndSet(ReplicationReplayState.DEGRADED,
```
tkhurana: The return value of compareAndSet is ignored

phoenix-core-server/src/main/java/org/apache/phoenix/replication/reader/ReplicationLogDiscoveryReplay.java
```diff
@@ -202,50 +228,80 @@ protected void processFile(Path path) throws IOException {
...
-    ) {
+    HAGroupStoreRecord.HAGroupState haGroupState = haGroupStoreRecord.getHAGroupState();
+    LOG.info("Found HA Group state during initialization as {} for haGroup: {}", haGroupState,
+      haGroupName);
+    // Each branch below sets the initial state with compareAndSet(NOT_INITIALIZED, ...), not a
+    // plain set: the LOCAL state listeners are subscribed in init() before this runs, so a
+    // concurrent LOCAL transition may have already advanced the state off NOT_INITIALIZED. When
+    // that happens the CAS no-ops and we deliberately keep the listener's value -- a live
+    // transition is more recent (and thus more authoritative) than the record snapshot read above.
+    if (HAGroupStoreRecord.HAGroupState.DEGRADED_STANDBY.equals(haGroupState)) {
```
tkhurana, line 272: Init-window race can leave a direct failover silently stuck (correctness / silent failure — narrow window, self-heals on RS restart)
  ReplicationLogDiscoveryReplay.java:266 interacting with triggerFailoverListner at :178-180

  Listeners are subscribed in init() before super.init() runs initializeLastRoundProcessed(), so a LOCAL transition can fire mid-init. The comment at :260-264 covers the
  interleaving where the listener wins, but not this one:
  1. getHAGroupRecord() reads (slightly stale) DEGRADED_STANDBY → picks the DEGRADED branch.
  2. Before CAS(NOT_INITIALIZED, DEGRADED) at :266, the record flips to STANDBY_TO_ACTIVE; triggerFailoverListner runs compareAndSet(DEGRADED, SYNCED_RECOVERY) while state
  is still NOT_INITIALIZED → no-op, then failoverPending.set(true).
  3. Init resumes: CAS(NOT_INITIALIZED, DEGRADED) succeeds → state DEGRADED.
  4. Trailing arm-check at :316 tests the local haGroupState (DEGRADED_STANDBY), so it doesn't reconcile.

  Result: state=DEGRADED + failoverPending=true, but nothing will move DEGRADED → SYNCED_RECOVERY (the record is already STANDBY_TO_ACTIVE and won't re-fire; a direct
  failover never revisits STANDBY). shouldTriggerFailover() gates on SYNC, so promotion never happens until the RS restarts (the STANDBY_TO_ACTIVE init branch heals it).
  Low-probability, but it's exactly the "legitimate signal silently dropped" case the degradedListener comment warns against.

  Suggested fix: after the init branches resolve, reconcile — if failoverPending.get() is true and resolved state is DEGRADED, promote to SYNCED_RECOVERY (or re-derive as
  the STANDBY_TO_ACTIVE branch does) and log a WARN.

phoenix-core-server/src/main/java/org/apache/phoenix/replication/reader/ReplicationLogDiscoveryReplay.java
Himanshu-g81: We just want to ensure we don't do any un-desirable state change and avoid race conditions. Have logged it also (in one of the existing log line)

phoenix-core-server/src/main/java/org/apache/phoenix/replication/reader/ReplicationLogDiscoveryReplay.java
Himanshu-g81: https://github.com/apache/phoenix/pull/2577#discussion_r3638948503

phoenix-core-server/src/main/java/org/apache/phoenix/replication/reader/ReplicationLogDiscoveryReplay.java
Himanshu-g81: Thanks, have updated in latest commit.

tkhurana (approved):

PR #2580, opened 2026-07-22
PHOENIX-7965 :- ConnectionQueryServicesImpl.getMetaDataCache() returns null after connection close, causing a bare NPE during concurrent query compilation

lokiore:

`ConnectionQueryServicesImpl.getMetaDataCache()` returned the `volatile latestMetaData` field directly, with no null guard:

```java
public PMetaData getMetaDataCache() {
  return latestMetaData;
}
```

This PR adds a guard so a closed connection surfaces a descriptive exception instead of a bare NPE. The read is done once into a local so the null-check and the returned reference are guaranteed to be the same value (no time-of-check/time-of-use re-read window):

```java
public PMetaData getMetaDataCache() {
  PMetaData cache = latestMetaData;
  if (cache == null) {
    throwConnectionClosedException();
  }
  return cache;
}
```

`throwConnectionClosedException()` is the same private helper that the existing `throwConnectionClosedIfNullMetaData()` (used by `addTable`, `removeTable`, `pruneTables`, the internal `getTable` path, etc.) delegates to — it throws `IllegalStateException("Connection to the cluster is closed")`. `getMetaDataCache()` was the one cache accessor missing this guard.

`close()` sets `latestMetaData = null`. When a connection is closed concurrently with query compilation (for example, connection-pool teardown or a cluster role transition), an in-flight compile reaches the cache via `PhoenixConnection.getTableRef()` / `getFunction()` / `getSchema()`, which call `getMetaDataCache().getTableRef(key)` (and friends) on the `null` return. The result is a **bare, causeless `NullPointerException`** with no message — difficult to attribute to a closed connection during triage.

Every other accessor/mutator on this class already guards this state and throws the descriptive `IllegalStateException`; `getMetaDataCache()` was inconsistent. This change makes the failure mode uniform and diagnosable.

No behavioral change on an open connection: when `latestMetaData` is present, the method returns exactly the same cache reference as before. The only change is the failure mode when the connection has been closed — an opaque `NullPointerException` is upgraded to a descriptive `IllegalStateException("Connection to the cluster is closed")`. No caller relies on a `null` return (every production caller immediately dereferences the result), so no caller is regressed.

Added unit tests in `ConnectionQueryServicesImplTest` exercising a **real** `ConnectionQueryServicesImpl` instance (not a mock, so the real guard executes):

- `testGetMetaDataCacheThrowsWhenClosed` — reflectively nulls `latestMetaData` (closed-connection state) and asserts `getMetaDataCache()` throws `IllegalStateException` with message `"Connection to the cluster is closed"`.
- `testGetMetaDataCacheReturnsCacheWhenOpen` — asserts the method returns the cache unchanged when present.

Both pass (`Tests run: 2, Failures: 0, Errors: 0`) via a reactor build (`mvn -pl phoenix-core -am test`). `mvn spotless:apply` reports the tree clean.

### DIFF
--- a/phoenix-core-client/src/main/java/org/apache/phoenix/query/ConnectionQueryServicesImpl.java
+++ b/phoenix-core-client/src/main/java/org/apache/phoenix/query/ConnectionQueryServicesImpl.java
@@ -1090,7 +1090,16 @@ public class ConnectionQueryServicesImpl extends DelegateQueryServices
   }
 
   public PMetaData getMetaDataCache() {
-    return latestMetaData;
+    // A concurrent close() (e.g. connection pool teardown or cluster failover) nulls out
+    // latestMetaData. Read it once into a local so the null-check and the returned reference
+    // are guaranteed to be the same value; a caller that immediately dereferences the result
+    // then surfaces the descriptive "connection closed" exception instead of a bare NPE,
+    // matching every other cache accessor/mutator on this class.
+    PMetaData cache = latestMetaData;
+    if (cache == null) {
+      throwConnectionClosedException();
+    }
+    return cache;
   }
 
   @Override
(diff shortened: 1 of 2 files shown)


palashc (approved):
+1, pending build
https://ci-hadoop.apache.org/job/Phoenix/job/Phoenix-PreCommit-GitHub-PR/job/PR-2580/3/

PR #2581, opened 2026-07-22
PHOENIX-7962 Fix ReplicationLogGroup mode initialization and handle failover deadlock

tkhurana:
## Summary

A `ReplicationLogGroup` is a **writer** and should exist only where the local cluster is active for the HA group. Previously it was created unconditionally — lazily on first mutation and eagerly at RS startup — which let non-active clusters start writers and re-ship (split-brain risk), and let a cutover writer deadlock standby failover by continuously rotating files into the peer's shard directory.

## Changes

- **Role-gated creation** — `ReplicationLogGroup.get(...)` now returns `Optional` and creates a group only when the local role is active. A non-active role returns empty and is **not cached**, so the next call after promotion re-checks. Live write path (`preBatchMutate`) throws on empty (split-brain guard); replay path (`preWALRestore`) skips re-ship on empty.
- **Skip eager init on non-active clusters** — `PhoenixRegionServerEndpoint` only eager-inits where the local role is active; others init lazily on promotion.
- **Failover deadlock fix** — a new `failoverPending` flag (set on entry to the `ACTIVE_IN_SYNC_TO_STANDBY` cutover gate, cleared on abort) suspends log rotation and turns a SYNC-write failure into an RS abort instead of the illegal cutover→SAF transition. `init()` seeds the flag on restart-in-cutover.
- **Consolidated state listeners** — all LOCAL state subscriptions moved into `subscribeToStateChanges()`, each paired with its unsubscribe so `close()` leaks no watchers; handles cutover suspend, `STANDBY`/`DEGRADED_STANDBY` close, and SYNC ↔ SYNC_AND_FORWARD flips.
- **Constructor overload** — `init()` reuses the record the role gate already read.

## Tests

9 new unit tests in `ReplicationLogGroupTest` plus 2 integration tests (`testWriterClosesOnDemotionToStandby`, `testCutoverSuspendsAndResumesRotation`).

## Note

Behavioral change: a sync-path write to a non-active cluster now throws rather than silently creating a writer — the intended split-brain guard.

### DIFF
--- a/phoenix-core-server/src/main/java/org/apache/phoenix/coprocessor/PhoenixRegionServerEndpoint.java
+++ b/phoenix-core-server/src/main/java/org/apache/phoenix/coprocessor/PhoenixRegionServerEndpoint.java
@@ -29,6 +29,7 @@ import java.util.ArrayList;
 import java.util.Collections;
 import java.util.Iterator;
 import java.util.List;
+import java.util.Optional;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
 import org.apache.hadoop.conf.Configuration;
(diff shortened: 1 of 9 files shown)


apurtell (commented):
+1 (belated)

PR #2583, opened 2026-07-24
PHOENIX-7966 :- Fix flaky HA IT transitClusterRole: expect the settled peer role

lokiore:

Fixes intermittent timeouts in the HA integration-test helper `transitClusterRole` (~1 in 5 runs) when a group transitions with both clusters live. The helper writes an expected `ClusterRoleRecord` to both clusters and then `waitFor`s the refreshed record to equal it. When the local cluster moves toward standby (`ACTIVE_TO_STANDBY`), a live peer's failover-management listener autonomously advances a `STANDBY` peer to `STANDBY_TO_ACTIVE`. Naming the transient pre-advance role in the wait races that autonomous transition and intermittently never matches, timing out at the 10s poll deadline.

Two complementary, deterministic fixes, each matching its call site's intent:

- **`FailoverPhoenixConnectionIT`, `HAGroupMetricsIT`** (setup-then-assert sites): the `{ACTIVE_TO_STANDBY, STANDBY}` transit is only a stepping stone to a later settled state the test actually asserts on. Keep the settle-wait but name the **settled** peer role `STANDBY_TO_ACTIVE` — a fixed point in the peer/local transition maps — so the equality wait converges deterministically within the settle window rather than racing the listener.
- **`ParallelPhoenixConnectionIT.testOneClusterATSRoleWithStandby`** (asserts on the transient pair itself): `{ACTIVE_TO_STANDBY, STANDBY}` is the state under test and is not reachable as a stable pair with both clusters live, so no settled role can be named without changing what the test exercises. Skip the settle-wait (`doRefresh=false`) instead; the parallel connection is still exercised against the `ACTIVE_TO_STANDBY` cluster pair and must complete without error. The now-unused single-argument `testOneClusterATSRole(ClusterRole)` overload is removed.

The `transitClusterRole` Javadoc and the cluster-down / cluster-restart variants are annotated to document the peer-advance exposure and why each is or is not subject to it.

The tests are flaky: the `waitFor` equality check against a role pair that a live peer autonomously advances past will intermittently never match, producing a `TimeoutException` unrelated to any product defect. This erodes trust in the HA IT suite and causes spurious CI failures.

No. Test-only change; no production code is touched.

- `mvn spotless:apply` clean; test-compile `BUILD SUCCESS`.
- `ParallelPhoenixConnectionIT#testOneClusterATSRoleWithStandby` run in isolation: `Tests run: 1, Failures: 0, Errors: 0`, zero `Timed out` occurrences. Verified non-vacuous — the log shows the group reaching `ACTIVE_IN_SYNC_TO_STANDBY` while the parallel connection ran.
- Loop of `testOneClusterATSRoleWithStandby` (5×) plus `testOneClusterATSRoleWithActive` and `testBothClusterATSRole`: all green, zero timeouts/failures/errors.

### DIFF
--- a/phoenix-core/src/it/java/org/apache/phoenix/jdbc/FailoverPhoenixConnectionIT.java
+++ b/phoenix-core/src/it/java/org/apache/phoenix/jdbc/FailoverPhoenixConnectionIT.java
@@ -196,7 +196,10 @@ public class FailoverPhoenixConnectionIT extends HABaseIT {
       doTestBasicOperationsWithConnection(conn, tableName, haGroupName);
     }
 
-    CLUSTERS.transitClusterRole(haGroup, ClusterRole.ACTIVE_TO_STANDBY, ClusterRole.STANDBY);
+    // role2 = the SETTLED peer role: local ACTIVE_TO_STANDBY drives the peer
+    // STANDBY -> STANDBY_TO_ACTIVE, so the waitFor must expect the settled state.
+    CLUSTERS.transitClusterRole(haGroup, ClusterRole.ACTIVE_TO_STANDBY,
+      ClusterRole.STANDBY_TO_ACTIVE);
     CLUSTERS.transitClusterRole(haGroup, ClusterRole.STANDBY, ClusterRole.STANDBY);
 
     try {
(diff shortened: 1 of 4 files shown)


apurtell (approved):
+1

The robot has a small javadoc related suggestion: The Javadoc says the settled naming makes the record "match on the first poll regardless of listener timing." That's true in the common case but not literally guaranteed. The behavior is race-free in terms of convergence, but "first poll" is aspirational. Consider softening to "matches deterministically within the settle window."

lokiore:
Thanks for the review and the +1 @apurtell. Good catch on the wording — softened both the `transitClusterRole` Javadoc and the PR description from "matches on the first poll regardless of listener timing" to "converges/matches deterministically within the settle window", which is the precise claim (race-free convergence, not a literal first-poll guarantee). Pushed as ce84187 (docs-only, no behavior change).

PR #2588, opened 2026-07-28
PHOENIX-7974 Preserve origin server identity when forwarding replication log files

tkhurana:
The forwarder computed the peer destination as getWriterPath(sourceTs, forwardingServerName), keying the dst on the forwarding RS's own name and discarding the origin writer identity. Two source files sharing a timestamp (across RegionServers) then collapsed onto one dst; with FileUtil.copy overwrite=false the second was permanently rejected with PathExistsException, holding the HA group in SYNC_AND_FORWARD long after data was replicated.

Key the dst on the origin server name instead, so the forwarded file is byte-identical to what the origin RS would have written natively and inherits the writer's existing (server, timestamp) uniqueness invariant. Add ReplicationLogTracker.getServerName() to extract the origin identity (dot-safe so IP/FQDN host names survive), and route the .plog extension through a shared ReplicationShardDirectoryManager.LOG_FILE_EXTENSION constant.

### DIFF
--- a/phoenix-core-server/src/main/java/org/apache/phoenix/replication/ReplicationLogTracker.java
+++ b/phoenix-core-server/src/main/java/org/apache/phoenix/replication/ReplicationLogTracker.java
@@ -368,16 +368,13 @@ public class ReplicationLogTracker {
         // Format: <ts>_<server>_<UUID>_<renameTs>.plog → extract prefix (first 2 parts)
         String[] parts = fileName.split("_");
         String prefix = parts[0] + "_" + parts[1];
-        newFileName = prefix + "_" + UUID.randomUUID() + "_" + EnvironmentEdgeManager.currentTime()
-          + ReplicationShardDirectoryManager.LOG_FILE_EXTENSION;
+        newFileName = buildInProgressFileName(prefix);
         targetDirectory = file.getParent();
       } else {
         // File is not in in-progress directory, add UUID + rename timestamp and move to
         // IN_PROGRESS directory
         String baseName = fileName.substring(0, fileName.lastIndexOf("."));
-        newFileName =
-          baseName + "_" + UUID.randomUUID() + "_" + EnvironmentEdgeManager.currentTime()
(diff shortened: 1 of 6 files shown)


tkhurana:
@Himanshu-g81 please review

phoenix-core-server/src/main/java/org/apache/phoenix/replication/ReplicationLogTracker.java
```diff
@@ -368,15 +368,16 @@ protected Optional<Path> markInProgress(final Path file) {
         // Format: <ts>_<server>_<UUID>_<renameTs>.plog → extract prefix (first 2 parts)
         String[] parts = fileName.split("_");
         String prefix = parts[0] + "_" + parts[1];
-        newFileName =
-          prefix + "_" + UUID.randomUUID() + "_" + EnvironmentEdgeManager.currentTime() + ".plog";
+        newFileName = prefix + "_" + UUID.randomUUID() + "_" + EnvironmentEdgeManager.currentTime()
```
Himanshu-g81: nit: can be a utility method and reused in the same file for in progress files.

phoenix-core-server/src/main/java/org/apache/phoenix/replication/ReplicationShardDirectoryManager.java
```diff
@@ -64,11 +64,14 @@ public class ReplicationShardDirectoryManager {
    */
   public static final String SHARD_DIR_FORMAT = "%03d";
 
+  /** File extension for replication log files. */
+  public static final String LOG_FILE_EXTENSION = ".plog";
```
Himanshu-g81, line 68: Good to replace all the usages? Example: https://github.com/apache/phoenix/blob/91db1394c69803c57a645bb2ca199dc1bf84d0c6/phoenix-core-server/src/main/java/org/apache/phoenix/replication/tool/LogFileAnalyzer.java#L208

Himanshu-g81 (approved):
Just 2 NITs, rest LGTM.

